PR #120, opened 2019-02-07
AppEngine: do not use debug mode for WSGIApplication.

Dor1s:
Debug mode enables surfacing stacktraces to the users and maybe some other stuff. I think we log everything into StackDriver, so it's better to return 500 Internal Server Error in case of an exception.


Dor1s:
Thanks for the pointer, I'll play more with it later.

PR #128, opened 2019-02-08
[install_deps.bash] Don't add ppa unless needed

jonathanmetzman:


local/install_deps_linux.bash
```diff
@@ -64,15 +64,15 @@ if [ "$distro_codename" != "rodete" ]; then
...
-  sudo apt-get update
 fi
 
 # Install JDK, prerequisite for bazel.
 if [ "$distro_codename" == "trusty" ]; then
+  sudo add-apt-repository -y ppa:webupd8team/java
+  sudo apt-get update
   sudo apt-get install -y oracle-java8-installer
 else
+  sudo apt-get update
```
inferno-chromium: This just repeats it twice with line 80, any ways around it ?

inferno-chromium: Probably move this after line 79 block, then this line can be removed.

local/install_deps_linux.bash
```diff
@@ -65,19 +65,19 @@ if [ "$distro_codename" != "rodete" ]; then
...
   sudo apt-get install -y openjdk-8-jdk
 fi
 
 # Install apt-get packages.
-sudo apt-get update
```
inferno-chromium: Don't like this removal since in future code fixes, people can just remove block above and forget.

inferno-chromium:
actually best to move package name into a $var that is assigned in the loop
and then sudo apt-get install that installs other packages also install this $var. my point is just to keep sudo apt-get update outside.

local/install_deps_linux.bash
```diff
@@ -67,16 +67,18 @@ if [ "$distro_codename" != "rodete" ]; then
...
+  java_package=oracle-java8-installer
 else
-  sudo apt-get update
-  sudo apt-get install -y openjdk-8-jdk
+  java_package=openjdk-8-jdk
 fi
 
+# Install Java
+sudo apt-get update
+sudo apt-get install -y $java_package
```
inferno-chromium: can club with ones on line 83.

PR #126, opened 2019-02-07
Use Python3 compatible syntax.

alex:
This code continues to run correctly under Python2.

Changes include:
- Modernized octal literals
- Print as a function
- 'as' syntax for handling exceptions
- Remove of 'L' long suffixes

PR #133, opened 2019-02-08
CI testing

oliverchang:


cclauss:
A test that would be worthwhile to add to the CI would be:
```
$ pip install flake8
# stop the build if there are Python syntax errors or undefined names
$ flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
```
When run on Python 3, it would find all the issues fixed in #126 plus many undefined names that will be required to achieve complete the Python 3 compatibility.

__E901,E999,F821,F822,F823__ are the "_showstopper_" [flake8](http://flake8.pycqa.org) issues that can halt the runtime with a SyntaxError, NameError, etc. These 5 are different from most other flake8 issues which are merely "style violations" -- useful for readability but they do not effect runtime safety.
* F821: undefined name `name`
* F822: undefined name `name` in `__all__`
* F823: local variable name referenced before assignment
* E901: SyntaxError or IndentationError
* E999: SyntaxError -- failed to compile a file into an Abstract Syntax Tree

Issue #134, opened 2019-02-08
local/install_deps.bash seems to fail mysteriously when there isn't enough memory


evverx:
I launched a VM where `install_deps.bash` failed with
```sh
Running: bazel run //local:create_gopath 2>/dev/null
| Return code is non-zero (1).
| Exit.
```
, which wasn't very helpful. Without `2 >/dev/null` I was able to figure out what happened but unfortunately additionally broke something else because apparently the output is used to set `GOPATH` :-) I think it would be great if the output wasn't sent to "/dev/null" here.

(I relaxed the limits a bit and got past `install_deps.bash` and now `python butler.py run_server --bootstrap` seems to be stuck trying to "build templates for App Engine". The machine is totally unresponsive so it's hard to tell what's going on. I'm going to tweak the limits until it works but it would probably be useful to mention in the documentation how a "typical" machine is supposed to look like, that is something like https://github.com/google/oss-fuzz/blob/master/docs/faq.md#what-are-the-specs-on-your-machines maybe).

evverx:
> The machine is totally unresponsive

Judging by "Out of memory: Kill process 2299 (node) score 115 or sacrifice child", it was the OOM killer that was busy destroying "polymer-bundler". I hope 32Gb will suffice.

Dor1s:
That's really weird, I don't think you need any beefy machine for that. Which OS are you trying this on?

evverx:
That was Ubuntu 18.04.1 LTS run on top of macOS Mojave with VirtualBox.

evverx:
Anyway, I kind of find an abandoned beefy machine :-) and everything seems to be working as far as I can tell. I don't think I'll start to dig deeper any time soon so I think the issue can safely be closed.

And, by the way, congratulations on open-sourcing ClusterFuzz!

inferno-chromium:
Reopening to improve the things mentioned in c#0.

inferno-chromium:
bazel does not support a quite mode, so everything gets dumped to stderr - https://groups.google.com/forum/#!topic/bazel-discuss/JVCotYRXeyk

till then i will fix this to make it more obvious where we are running this command, like this is run in 'src' subdir. maybe also update code with comment.

inferno-chromium:
polymer_bundler will be tracked by its own issue. bazel issue should be fixed now. Thanks for the report.

PR #138, opened 2019-02-08
Show output from bazel run for easier debugging (#134)

inferno-chromium:


src/local/butler/common.py
```diff
@@ -145,7 +145,10 @@ def _print(s):
     if print_output:
       print(s)
 
-  _print('Running: %s' % command)
+  print_string = 'Running: %s' % command
+  if cwd:
+    print_string += " (dir='%s')" % cwd
```
Dor1s: super nit: maybe `(cwd='%s')` ? `dir` is a bit ambiguous

inferno-chromium: Done!

evverx:
@inferno-chromium @Dor1s thank you!

Not that it matters but I was wondering what "/gcbrun" is and having followed all the links I ended up at https://console.cloud.google.com/cloud-build/builds/88085705-09b3-4724-adfd-dc8200182f1e?project=472119376969 where I was informed that I don't have access to view builds and triggers. Would it be possible to make it publicly available (which I think would be useful when external contributors opens PRs)?

Dor1s:
> Would it be possible to make it publicly available (which I think would be useful when external contributors open PRs)?

Not sure about making that public, as it can be abused. Ideally, contributors should run the tests locally, and we'll trigger the tests on the CI prior to merging a patch. @oliverchang may have a better perspective

evverx:
Ideally, yes, contributors should run the tests and PR won't be merged if the CI is unhappy, but if I hypothetically broke something, I'd have to wait for someone else to take a look and tell me what went wrong which doesn't seem to be as convenient (and fast) as immediate feedback I'd receive if I had access to the logs. But I agree that currently this is a hypothetical issue.

PR #141, opened 2019-02-08
Run the linter on CloudBuild CI

ammaraskar:
This was a quick solution I came up with to run the CI on CloudBuild. Please feel free to close if there's an easier or more straightforward way I'm missing. Here it is in action breaking on the lint step: https://github.com/ammaraskar/clusterfuzz/pull/4

Some things of note:

---

CloudBuild with Github integration actually doesn't pull over the `.git` folder. It simply downloads the source as a tarball (see here: https://github.com/GoogleCloudPlatform/cloud-builders/issues/401)

Thus, I took a note out of the kubernetes example here: https://github.com/GoogleCloudPlatform/marketplace-k8s-app-example/blob/master/cloudbuild.yaml and performed a git clone to get the history out as the first step.

---

For pull requests, I explicitly compare against master which may or may not be good. Unlike say travis which provides a TRAVIS_COMMIT_RANGE [variable](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables), CloudBuild has no way of providing what commits are part of the pull request.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -13,6 +13,26 @@
 # limitations under the License.
 
 steps:
+- id: Initialize git
+  name: gcr.io/cloud-builders/git
+  entrypoint: /bin/bash
+  args:
+  - -exc
+  - |
+    # Cloud Build x GitHub integration uses source archives to fetch
+    # the source, rather than Git source fetching, and as a consequence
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
Thanks a lot Ammar!
@oliverchang since you were working on this issue can you please take a look?

PR #142, opened 2019-02-08
Only build templates if needed, and remove parallel compilation.

mbarbella-chromium:
Even though parallel building is much faster, it feels pretty hacky. Instead of throttling it to only use a certain number of processes, this feels like a cleaner approach. It's quite a bit slower when a build is required (at least on a powerful machine), but builds shouldn't be needed very often with this change.


local/polymer_bundler.bash
```diff
@@ -19,10 +19,17 @@ build_one_file() {
   local deployment_directory=$3
   local deployment_file="${source_file/$source_directory/$deployment_directory}"
   mkdir -p `dirname $deployment_file`
-  polymer-bundler --inline-scripts --inline-css --strip-comments --out-file=$deployment_file $source_file &
+  polymer-bundler --inline-scripts --inline-css --strip-comments --out-file=$deployment_file $source_file
```
inferno-chromium: I had tested this, this is too slow without it :(

local/polymer_bundler.bash
```diff
@@ -19,10 +19,17 @@ build_one_file() {
...
 }
 
 cd $( dirname ${BASH_SOURCE[0]} )/../src/appengine
+mkdir templates 2> /dev/null
+
+if [[ $(find private -type f -printf "%T@\n" | sort -n | tail -1 | awk -F. '{ print $1 }') -le $(find templates -type f -printf "%T@\n" | sort -n | head -1 | awk -F. '{ print $1 }') ]]; then
```
inferno-chromium: This is too hacky. Lets write this file in python ?

mbarbella-chromium:
Discussed offline. Seems better to port this to Python to make it easier to maintain, but if that ends up being more difficult than expected we could consider landing this as a stopgap. Leaving this open for now.

mbarbella-chromium:
Abandoning in favor of https://github.com/google/clusterfuzz/pull/146

PR #146, opened 2019-02-09
Port polymer_bundler.bash to Python.

mbarbella-chromium:
PTAL when you have time. I realized late that this doesn't really need to be a standalone script, but until we have a better solution for updates while the dev appserver is running I think it's nice to have the option to run it directly to force it without a server restart.

It's definitely slower, but not as bad as doing it completely serially:

$ time ./polymer_bundler.bash 
Building templates for App Engine...
Finished building templates

real	0m10.616s
user	4m4.549s
sys	0m11.821s

$ touch ../src/appengine/private/templates/bots.html
$ time python polymer_bundler.py 
Building templates for App Engine...
App Engine templates built successfully!

real	0m26.773s
user	3m36.543s
sys	0m10.223s

$ time python polymer_bundler.py 
App Engine templates are up to date!

real	0m0.091s
user	0m0.046s
sys	0m0.050s


local/polymer_bundler.py
```diff
@@ -0,0 +1,77 @@
...
+
+from __future__ import print_function
+
+import os
+
+from multiprocessing import Pool
+
+APPENGINE_DIRECTORY = os.path.join(
+    os.path.dirname(__file__), os.pardir, 'src', 'appengine')
+PROCESS_COUNT = 5
```
inferno-chromium: how about max(multiprocessing.cpu_count() / 2, 1) ?

local/polymer_bundler.py
```diff
@@ -0,0 +1,77 @@
...
+    os.mkdir('templates')
+
+  template_names = os.listdir(os.path.join('private', 'templates'))
+  pool = Pool(PROCESS_COUNT)
+  result = pool.map(build_file, template_names)
+
+  if all(result):
+    print('App Engine templates built successfully!')
+  else:
+    print('Failed to build App Engine templates!')
```
inferno-chromium: exit with -1 or whatever common.execute does.

local/polymer_bundler.py
mbarbella-chromium: Good idea!

PR #147, opened 2019-02-09
Add comment on why patchelf is needed.

inferno-chromium:


inferno-chromium:
TBR simple comment pointed by @alex

PR #148, opened 2019-02-09
Update local_instance.md

kcwu:


docs/getting-started/local_instance.md
```diff
@@ -29,8 +29,8 @@ python butler.py run_server --bootstrap
 python butler.py run_server
 ```
 
-This may take a few seconds to start. Once you see output starting with
-`INFO:root` at the beginning, you should be able to navigate to
+This may take a few seconds to start. Once you see output line like
+`INFO <timestamp> instance.py:<num>] Instance PID`, you should be able to navigate to
```
inferno-chromium: Instance happens when cron runs after 60 sec.

Server actually starts as soon as this line
| INFO     2019-02-09 03:12:52,397 admin_server.py:150] Starting admin server at: http://localhost:9002
so localhost:9000 is usable after that line

kcwu: revised

PR #149, opened 2019-02-09
When using a recent Ubuntu/Debian, install patchelf with apt

alex:


inferno-chromium:
Looks ok to me, but @oliverchang added this code, so will let him review this. Also unsure if we want to  add even more if conditions based on distros, as it will get harder for maintainability.

PR #154, opened 2019-02-09
Suggest not to run -u/-v with parallel test execution.

inferno-chromium:


inferno-chromium:
basically `python butler.py py_unittest -t appengine -m -v` just hangs. i think we shouldn't both supporting -m mode with -v, all prints and other things get garbled together. -u/-v should be run in single process mode.

inferno-chromium:
Closing this as we have some clear repro steps for root cause.

Issue #155, opened 2019-02-09
AppEngine tests fail when -v switch is used.


inferno-chromium:
See https://github.com/google/clusterfuzz/issues/139#issuecomment-462076432.

Issue #139, opened 2019-02-08
polymer_bundler can be improved to avoid OOMs


inferno-chromium:
https://github.com/google/clusterfuzz/issues/134#issuecomment-461794436

inferno-chromium:
I think our background processes might need to be throttled, and need to support this on both linux and mac.

jfoote:
I think I may have run into a related issue where `polymer-bundler` was causing a lightweight system to hang (GCE n1-standard-1, 1 vCPU, 3.75 GB memory). I removed [the & here](https://github.com/google/clusterfuzz/blob/master/local/polymer_bundler.bash#L22) to cause the invocations to run serially as a workaround FWIW.

inferno-chromium:
this should be fixed soon, Marty is looking into this.

mbarbella-chromium:
This should be addressed by https://github.com/google/clusterfuzz/pull/146. Let me know if you still have any issues.

evverx:
Speaking of OOMs, the python unit tests run in verbose mode (`python butler.py py_unittest -t appengine -v`) seem to get stuck and eventually killed by the OOM killer on the machine I mentioned in https://github.com/google/clusterfuzz/issues/134:
```
ok
test_lease_reach_latest (tests.appengine.common.tasks_test.LeaseTaskTest)
Test leasing a task and reaching the maximum lease time. ... WARNING:root:suspended generator _get_tasklet(context.py:344) raised BadRequestError(Error code: INVALID_ARGUMENT. Message: Key path element must not be incomplete: [Heartbeat: ])
WARNING:root:suspended generator get(context.py:760) raised BadRequestError(Error code: INVALID_ARGUMENT. Message: Key path element must not be incomplete: [Heartbeat: ])
Killed
```
```
Out of memory: Kill process 12437 (python) score 471 or sacrifice child
Out of memory: Kill process 12437 (python) score 489 or sacrifice child
Out of memory: Kill process 12437 (python) score 607 or sacrifice child
```
I'm not sure if it makes sense to fix it though because in general it doesn't seems to be a very good idea to try to run ClusterFuzz on lightweight machines.

inferno-chromium:
2 things:

1. For fuzzing, ClusterFuzz always runs on lightweight machines. Our 25,000 vms are all n1-standard-1 (1core, 4gb) linux and win is like n1-standard-2 (2core, 8 gb). There is no issue there at all and it does not run anything heavy wight there.
[...]

PR #156, opened 2019-02-09
Mock data_handler.update_heartbeat in tasks_test.py (#155)

inferno-chromium:
* data_handler.update_heartbeat uses time.time() which is mocked in tasks_test and its side_effect
return values order gets messed up. So, mocking update_heartbeat since test just tests tasks's leasing behavior.

### DIFF
--- a/src/python/tests/appengine/common/tasks_test.py
+++ b/src/python/tests/appengine/common/tasks_test.py
@@ -142,6 +142,7 @@ class GetTaskTest(unittest.TestCase):
     self.assertEqual('test override job', task.payload())
 
 
+@test_utils.with_cloud_emulators('datastore')
 class LeaseTaskTest(unittest.TestCase):
   """Tests for leasing tasks."""
 
--- a/src/python/tests/appengine/common/tasks_test.py
+++ b/src/python/tests/appengine/common/tasks_test.py
@@ -142,7 +142,6 @@ class GetTaskTest(unittest.TestCase):
     self.assertEqual('test override job', task.payload())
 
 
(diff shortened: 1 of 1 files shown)


inferno-chromium:
This causes

Test leasing a task and reaching the maximum lease time. ... WARNING:root:suspended generator _get_tasklet(context.py:344) raised BadRequestError(Error code: INVALID_ARGUMENT. Message: Key path element must not be incomplete: [Heartbeat: ])
WARNING:root:suspended generator get(context.py:760) raised BadRequestError(Error code: INVALID_ARGUMENT. Message: Key path element must not be incomplete: [Heartbeat: ])

when test is run with -v, weird it didn't show without. This should have started happening after https://github.com/google/clusterfuzz/commit/6224ff6468e5eac4fdf14fa68e9f52b8d41c2fda#diff-bd5bc9b8475580e06b37ec4379612038

inferno-chromium:
The reason is 
    persistent_cache.set_value(
        HEARTBEAT_LAST_UPDATE_KEY, time.time(), persist_across_reboots=True)
it calls time.time() and tasks_test functions in e.g. test_lease_reach_latest
do mock time.time using side_effect and confuse the values order
    self.mock.time.side_effect = [
        0, 0, 0, 600, tasks.TASK_LEASE_SECONDS - 30, tasks.TASK_LEASE_SECONDS
    ]

inferno-chromium:
reporter confirms that the fix works - https://github.com/google/clusterfuzz/issues/139#issuecomment-462097541

PR #158, opened 2019-02-11
Remove depot_tools from CI.

oliverchang:


oliverchang:
TBR. this is causing issues since there is a pylint in depot_tools which has different behaviour from ours. 

We don't need this anymore anyway, so removing.

inferno-chromium:
LGTM nice to remove this.

Issue #159, opened 2019-02-11
Clean up unused libraries


oliverchang:
We may have a bunch of pip packages that we don't use anymore, such as pycrypto.

inferno-chromium:
yes @mbarbella-chromium was mentioning that as well. I think when we removed HTTPSServer support to serve testcases from, we probably forgot to remove pycrypto.

inferno-chromium:
had to revert ssl module in https://github.com/google/clusterfuzz/commit/c44bf3f81d19148548a913e6585101c5d9241585

inferno-chromium:
We can keep this open to check others. Wished there was a tool for this :)

PR #161, opened 2019-02-11
Remove unneeded pycrypto, pyOpenSSL and ssl packages (#159)

inferno-chromium:


inferno-chromium:
local testcase load on local webserver feature (only for blackbox fuzzers over things like browsers) is only supported through http. https was removed due to cert complexity with deployment and is unneeded for local bot to its local webserver.

inferno-chromium:
will let it run overnight on test project so make sure i am not missing anything.

PR #162, opened 2019-02-11
Run cloudbuild lint on Python instead of legacy Python

cclauss:
Related to #126


alex:
Since clusterfuzz is a python2 project, I don't think this makes much sense.

inferno-chromium:
yes we would transition sometime this year, but this is not applicable right now.

Issue #163, opened 2019-02-11
Document port forwarding for 900*, maybe other ports if server accessed from remote.


kcwu:
I'm running a local instance of clusterfuzz. I just run the server with default settings.
```
$ python butler.py run_server
```

Then I follows this guide https://google.github.io/clusterfuzz/setting-up-fuzzing/libfuzzer-and-afl/ ,

I can "add new job". But it failed if I upload custom fuzzer zip to the job.
```
| ERROR    2019-02-11 15:44:09,029 base_handler.py:198] Failed to upload.
| Traceback (most recent call last):
|   File "/usr/lib/google-cloud-sdk/platform/google_appengine/lib/webapp2-2.3/webapp2.py", line 545, in dispatch
|     return method(*args, **kwargs)
|   File "/home/kcwu/src/clusterfuzz/src/appengine/libs/handler.py", line 286, in wrapper
|     return func(self, *args, **kwargs)
|   File "/home/kcwu/src/clusterfuzz/src/appengine/libs/handler.py", line 413, in wrapper
|     return func(self, *args, **kwargs)
|   File "/home/kcwu/src/clusterfuzz/src/appengine/handlers/jobs.py", line 126, in post
|     blob_info = self.get_upload()
|   File "/home/kcwu/src/clusterfuzz/src/appengine/handlers/base_handler.py", line 248, in get_upload
|     raise helpers.EarlyExitException('Failed to upload.', 500)
| EarlyExitException: Failed to upload.
```

inferno-chromium:
can you refresh the `http://localhost:9000/fuzzers` page and retry. can you reliably reproduce this ? also can you try ctrl+c on the run_server and restarting the server again to reproduce ?

inferno-chromium:
Another question - How big is your fuzzer zip ?

kcwu:
Refreshing /jobs page and restarting the server don't help. And yes, always reproducible --- I haven't ever successfully uploaded fuzzers. The zip size is 12MB.

jonathanmetzman:
To clarify, @kcwu are you trying to upload a custom build to the job? 
You aren't trying to upload a fuzzer to the fuzzers page right?

jonathanmetzman:
Also, where did you get your copy of ClusterFuzz? Did you download the relase? Or did you clone the repo (if you cloned it, can you post the output of `git rev-parse HEAD` so we can know what commit you are on).

kcwu:
[...]

PR #170, opened 2019-02-12
Log when we fail to get a blob from a key

jonathanmetzman:


jonathanmetzman:
Should help debug #163

Issue #168, opened 2019-02-11
Add links to fuzzing resources in documentation


oliverchang:
We should add more links to tutorials for fuzzing to our documentation for people who may not be as familiar with it.

Sai-Adarsh:
@oliverchang Shall I work on this?

Dor1s:
Something like these? Mostly taken from OSS-Fuzz page + a new structure aware fuzzing one.

* [libFuzzer documentation](http://libfuzzer.info)
* [libFuzzer tutorial](http://tutorial.libfuzzer.info)
* [libFuzzer workshop](https://github.com/Dor1s/libfuzzer-workshop)
* [Chromium Fuzzing Page](https://chromium.googlesource.com/chromium/src/testing/libfuzzer/)
* [Structure-Aware Fuzzing with libFuzzer](https://github.com/google/fuzzer-test-suite/blob/master/tutorial/structure-aware-fuzzing.md)

Sai-Adarsh:
> Something like these? Mostly taken from OSS-Fuzz page + a new structure aware fuzzing one.
> 
> * [libFuzzer documentation](http://libfuzzer.info)
> * [libFuzzer tutorial](http://tutorial.libfuzzer.info)
> * [libFuzzer workshop](https://github.com/Dor1s/libfuzzer-workshop)
> * [Chromium Fuzzing Page](https://chromium.googlesource.com/chromium/src/testing/libfuzzer/)
> * [Structure-Aware Fuzzing with libFuzzer](https://github.com/google/fuzzer-test-suite/blob/master/tutorial/structure-aware-fuzzing.md)

looks good to me.

inferno-chromium:
actually wanted more like what fuzzing is and more overview stuff. above links are already in libfuzzer/afl documentation.

Sai-Adarsh:
> 
> 
> actually wanted more like what fuzzing is and more overview stuff. above links are already in libfuzzer/afl documentation.

Like these?

* [Fuzz testing: Beginner’s guide](https://medium.com/ouspg/fuzz-testing-beginners-guide-da2c9179caa7)
* [Complete Guide to Fuzzing](https://www.cybrary.it/2017/08/your-complete-guide-to-fuzzing/)
* [15 minute guide to fuzzing](https://www.mwrinfosecurity.com/our-thinking/15-minute-guide-to-fuzzing/)
* [Curated list of fuzzing resources](https://github.com/secfigo/Awesome-Fuzzing)

[...]

PR #174, opened 2019-02-12
Update README.md

Sai-Adarsh:
Updates README.md, solves #168 by adding more links on fuzzing resources in documentation.


Sai-Adarsh:
@inferno-chromium could you review my PR for the issue #168

PR #178, opened 2019-02-12
Warn about port forwarding

jonathanmetzman:


docs/getting-started/local_instance.md
```diff
@@ -29,9 +29,12 @@ python butler.py run_server --bootstrap
 python butler.py run_server
 ```
 
-This may take a few seconds to start. Once you see output line like
-`INFO <timestamp> admin_server.py:<num>] Starting admin server`, you should be able to navigate to
-[http://localhost:9000](http://localhost:9000) to view the web interface.
+This may take a few seconds to start. Once you see an output line like
+`INFO <timestamp> admin_server.py:<num>] Starting admin server`, you should be
+able to navigate to [http://localhost:9000](http://localhost:9000) to view the
+web interface. Note that the local instance may use ports other than 9000, such
```
inferno-chromium: nit: can you use Note to next line as

<star><star>Note<star><star>:

inferno-chromium: Can you hyperlink "other than 9000" to https://github.com/google/clusterfuzz/blob/master/src/local/butler/constants.py ? that has list of ports.

docs/getting-started/local_instance.md
```diff
@@ -29,9 +29,12 @@ python butler.py run_server --bootstrap
...
+as 9008, for things like uploading files. Therefore, using port forwarding to
+access a local instance may break.
```
inferno-chromium: "break if all the needed ports are not forwarded" or something like that ?

jonathanmetzman:
Should fix #163
cc @kcwu @charleyw

docs/getting-started/local_instance.md
jonathanmetzman: Done.

docs/getting-started/local_instance.md
jonathanmetzman: Done.

docs/getting-started/local_instance.md
jonathanmetzman: Done.

jonathanmetzman:
@inferno-chromium I made this about more than port forwarding since that is a pretty specific case.
I think a better fix is to fail loudly on the client side if the upload fails.

docs/getting-started/local_instance.md
```diff
@@ -32,9 +32,11 @@ python butler.py run_server
 This may take a few seconds to start. Once you see an output line like
 `INFO <timestamp> admin_server.py:<num>] Starting admin server`, you should be
 able to navigate to [http://localhost:9000](http://localhost:9000) to view the
-web interface. Note that the local instance may use ports other than 9000, such
-as 9008, for things like uploading files. Therefore, using port forwarding to
-access a local instance may break.
+web interface.
+Note: the local instance may use ports [other than 9000](https://github.com/google/clusterfuzz/blob/master/src/local/butler/constants.py),
+such as 9008, for things like uploading files. Therefore, using the local
```
kcwu: How about generalize this, ask users to tweak network rules like firewall and port forwarding.

jonathanmetzman: I think mentioning port forwarding and allowing through firewall is kind of too specific and I don't want to complicate the docs for the majority of users for whom this isn't an issue. What do you think of changing the last sentence to this:
>Therefore, using the local instance may break if the needed ports are unavailable or if you can only access some of the needed ports from your browser.

inferno-chromium: sgtm

jonathanmetzman: I ended up warning about port forwarding and firewall

PR #182, opened 2019-02-12
Fix linting and formatting errors

jonathanmetzman:


src/python/bot/minimizer/minimizer.py
```diff
@@ -368,20 +368,20 @@ def _attempt_merge(self, hypotheses, sibling_merge_succeeded=False):
...
+      return
 
     front = hypotheses[:len(hypotheses) / 2]
     back = hypotheses[len(hypotheses) / 2:]
 
     # If we could remove either one of two hypotheses, favor removing the first.
     front_merged_successfully = self._attempt_merge(front)
     self._attempt_merge(back, sibling_merge_succeeded=front_merged_successfully)
-    return False
+    return
```
mbarbella-chromium: Nit: this one can just be deleted

jonathanmetzman:
Doesn't fix issues in html_tokenizer.py which seemed more involved to fix.

src/python/google_cloud_utils/storage.py
```diff
@@ -510,7 +510,8 @@ def from_key(key):
     try:
       return GcsBlobInfo(blobs_bucket(), key)
     except Exception as exception:
-      logs.log_error('Failed to get blob from key %s.' % key)
+      logs.log_error(
+          'Failed to get blob from key %s.' % key, exception=exception)
```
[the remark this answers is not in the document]
oliverchang: I thought log_error did this automatically through sys.exc_info() ?

src/python/tests/core/bot/webserver/http_server_test.py
```diff
@@ -34,15 +34,15 @@ def __init__(self):
     def write(self, data):
       self.contents += data
 
-  def __init__(self, path):
+  def __init__(self, path):  # pylint: disable=super-init-not-called
```
oliverchang: can we just call the super init method instead?

src/python/bot/minimizer/minimizer.py
oliverchang: is this return needed at all?

src/python/fuzzing/corpus_manager.py
```diff
@@ -25,7 +25,7 @@
 
 try:
   from google_cloud_utils import gsutil
-  DEFAULT_GSUTIL_RUNNER = gsutil.GSUtilRunner
+  DEFAULT_GSUTIL_RUNNER = gsutil.GSUtilRunner  # pylint: disable=invalid-name
```
oliverchang: i'm a little surprised this gave an error. what was it complaining about?

jonathanmetzman:
> LGTM overall. A few of the lint errors look like they should actually be resolved instead of ignored, but it seems fine for us to clean those up later as we get to them.

Which ones? Maybe I shouldn't ignore them so theyre more likely to be fixed now that we know about them.

src/python/tests/appengine/handlers/cron/service_accounts_test.py
```diff
@@ -22,15 +22,15 @@ class ServiceAccountIdTest(unittest.TestCase):
...
 
   def test_max_length(self):
     self.assertEqual(
-        service_accounts._service_account_id('a' * 26), 'bot-%s' % ('a' * 26))
+        service_accounts._service_account_id('a' * 26), 'bot-%s' % ('a' * 26))  # pylint: disable=protected-access
 
   def test_more_than_max_length(self):
     with self.assertRaises(AssertionError):
-      service_accounts._service_account_id('a' * 27)
+      service_accounts._service_account_id('a' * 27)  # pylint: disable=protected-access
```
mbarbella-chromium: In this case it might be better to do this at the file-level.

src/python/bot/minimizer/minimizer.py
jonathanmetzman: Nope

src/python/fuzzing/corpus_manager.py
jonathanmetzman: It was complaining all caps names aren't valid class names. I'm frankly surprised it figured this out too.

src/python/google_cloud_utils/storage.py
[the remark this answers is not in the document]
jonathanmetzman: Good point. Stopped naming the exception.

src/python/bot/minimizer/minimizer.py
jonathanmetzman: Done

src/python/fuzzing/corpus_manager.py
oliverchang: wow nice. Let's just fix this to DefaultGSUtilRunner then

src/python/tests/core/bot/webserver/http_server_test.py
jonathanmetzman: I don't think so. Calling super fails I think because the grandparent is not a child of `object` (ie: it's an old-style class).

src/python/tests/appengine/handlers/cron/service_accounts_test.py
jonathanmetzman: I think we generally do it multiple times per file. But I can change it here if you'd still like me to.

oliverchang:
btw, I believe you'll need to merge in new changes from master for /gcbrun to work again on this PR.

src/python/fuzzing/corpus_manager.py
jonathanmetzman: Err....If I do that then pylint will complain that the assignment to `None` is badly named. I could refactor so that the assignment happens outside of try except, but I think it makes most sense to disable the check here.

oliverchang: sgtm

jonathanmetzman:
I'm not sure what exactly happened but I think I borked this PR using my workflow from gerrit. I'm going to close this and create a new PR.

PR #165, opened 2019-02-11
Don't Use Minijail With AFL

jonathanmetzman:
Remove use of minijail when USE_MINIJAIL is True for AFL fuzzing.
Also, refactor code that was made ugly because of minijail, particularly wrt merging the corpus
and AFL argument generation.

PR #195, opened 2019-02-13
Revert "Don't Use Minijail With AFL"

inferno-chromium:
Reverts google/clusterfuzz#165


inferno-chromium:
Causing issues on OSS-Fuzz. TBR.

PR #199, opened 2019-02-13
Add a flag to let oss-fuzz jobs set UNPACK_ALL_FUZZ_TARGETS_AND_FILES.

oliverchang:


oliverchang:
PTAL. /gcbrun

oliverchang:
> LGTM
> I don't think this should be documented, since it is something we will probably need to add, not the project maintainers.

+1

PR #202, opened 2019-02-14
Fix start_host.py.

oliverchang:
We need to set BOT_TMPDIR instead of TMPDIR after
https://github.com/google/clusterfuzz/pull/116.


oliverchang:
TBR to fix things.

jonathanmetzman:
LGTM

PR #205, opened 2019-02-14
[tests] Add a helper script for exporting test bucket names.

Dor1s:
Would anything like this be helpful as per your recent PSA? I can improve it to create the buckets if needed, plus make sure it works on macOS too.


oliverchang:
Hmm.. this is pretty internal team specific. Probably put it in an internal repo somewhere?

Dor1s:
SG!

PR #212, opened 2019-02-15
Load yaml safely

alex:
In all currently released versions of PyYAML, yaml.load can execute arbitrary code.

In this case data all comes from github, where changes require authn/authz, so it's not a critical security issue.


jonathanmetzman:
Thanks for the patch!

PR #213, opened 2019-02-15
Add extra logging when remounting NFS volume on Windows.

Dor1s:


src/python/bot/init_scripts/windows.py
```diff
@@ -65,6 +66,7 @@ def remount_if_needed():
   subprocess.call(['umount', '-f', nfs_root])
 
   # Mount the nfs drive.
+  logs.log_warn('Trying to remount the NFS volume.')
```
mbarbella-chromium: I think it's probably a bit safer to move this to 85. I'm not sure if the startup script mounts and creates the file, but if not we expect this to happen once.

Dor1s: I'm afraid that `subprocess.check_call` on line 72 may raise an exception, plus I think the startup script does mount the volume too.

mbarbella-chromium: Seems fine to keep it as-is then.

Issue #125, opened 2019-02-07
README clarification around language support


rudolfolah:
I know fuzzing is for testing inputs/outputs and is essentially language-agnostic, but it'd be nice if there's some mention of that in the README, or examples of running Clusterfuzz against code that's written in all sorts of languages (Chrome and C and C++ getting name-dropped is cool, but seeing Python or Ruby explicitly mentioned is cooler 😉 )

jonathanmetzman:
You are right that fuzzing/ClusterFuzz is language agnostic.

But I think we did this intentionally (CC @oliverchang).

We've only used ClusterFuzz for fuzzing C/C++ (and a small amount of rust), so we're only confident that with our docs it's easy to setup fuzzing for these kinds of programs.

While we would love it if people started using ClusterFuzz for fuzzing safe programming languages (some possible use cases [here](https://news.ycombinator.com/item?id=19107825)), we don't want to give people the impression that this is something they can easily use to improve their JS/Python etc. code.

inferno-chromium:
If you can find good memory instrumentation for those languages, it is pretty easy to extend ClusterFuzz to detect those crashes (just some parsing signatures in `stack_analyzer.py`). Infrastructure wise, you can run any program on Linux, Windows and Mac.

co16353sidak:
Hey everyone, is this an issue I could work on possibly ? would be a good start as my first contribution here. 
/cc: @inferno-chromium @jonathanmetzman @rudolfolah

jonathanmetzman:
I don't think so since IMO we did this intentionally.
Closing this issue, but if @inferno-chromium disagrees he can reopen.

jonathanmetzman:
Actually, it may be worth mentioning language agnostic part to appeal to go users or users of other languages that already have fuzzer support. Frankly, I don't even know if the tooling for Ruby/Python exists for them to even be integrated into ClusterFuzz. I think for clarity's sake we should say ClusterFuzz supports C/C++ and Rust and supporting go or LLVM-based languages should be doable.
@co16353sidak I don't think this is a good first issue unfortunately.

co16353sidak:
Okay 😞

rudolfolah:
@jonathanmetzman  thanks for updating the FAQ, think this resolves this issue! :smile:

Issue #226, opened 2019-02-20
Add a question on other language support to an FAQ


jonathanmetzman:
We've gotten enough questions on this that it makes sense to have a ready answer on whether ClusterFuzz supports languages other than C/C++.
It should mention:
1. How rust (and possibly other LLVM langs) works and go will probably be added.
2. Why fuzzing other languages probably won't serve the same function as fuzzing C/C++
3. What needs to be done to add support for another language.

PR #227, opened 2019-02-20
Add FAQ with section on language support

jonathanmetzman:
Add FAQ with section on language support detailing:
* What is supported and what we plan on supporting.
* Why LLVM is required
* Why language X isn't supported
* What is needed to add support for a new language

### DIFF
--- a/docs/reference/FAQ.md
+++ b/docs/reference/FAQ.md
@@ -29,35 +29,22 @@ reasons:
 ## What if I really don't want to use an LLVM-based toolchain?
 
 We strongly recommend using an LLVM-based toolchain. That said, you may be able
-to use some parts of ClusterFuzz with [gcc](https://gcc.gnu.org/) or extend
-ClusterFuzz to support your use of gcc. However, if you do this, you are in
-unexplored territory and we can not support this. Gcc can probably be used for
+to use some parts of ClusterFuzz with [GCC](https://gcc.gnu.org/) or extend
+ClusterFuzz to support your use of GCC. However, if you do this, you are in
+unexplored territory and we can not support this. GCC can probably be used for
 [blackbox fuzzing]({{ site.baseurl }}/setting-up-fuzzing/blackbox-fuzzing/)
-without modifying ClusterFuzz since gcc supports ASan. Using gcc with AFL or
+without modifying ClusterFuzz since GCC supports ASan. Using GCC with AFL or
 libFuzzer would probably require considerable changes to ClusterFuzz to get
 working. In the end, it is probably easier to make whatever changes are needed
 to build with clang.
 
-## Why isn't $MY_LANGUAGE Supported?
+## Why isn't $MY_LANGUAGE supported?
 
 In theory, ClusterFuzz is language agnostic and can be extended to support
 fuzzing code written in any language. In practice though, it is less easy to
 find interesting bugs in
 [memory safe](https://en.wikipedia.org/wiki/Memory_safety) languages so
 ClusterFuzz doesn't have support for fuzzing most of them.
-
-## How can I extend ClusterFuzz to support fuzzing my language
-
-Only experts should consider doing this. The first step is getting ClusterFuzz
-to recognize errors thrown by programs written in your language. This should
-make blackbox fuzzing possible. If you would like to add coverage-guided fuzzing
-support for your language, you need to a launcher like there is for AFL and
-libFuzzer that is capable of running fuzzers written in your language.
-
-## Are there plans to support fuzzing in other languages?
-
-The only language we are considering adding support for at this time is
-[Go](https://golang.org/).
 
 [SanitizerCoverage]: https://clang.llvm.org/docs/SanitizerCoverage.html
 [libFuzzer and AFL]: {{ site.baseurl }}/setting-up-fuzzing/libfuzzer-and-afl/
--- a/docs/reference/FAQ.md
+++ b/docs/reference/FAQ.md
@@ -12,8 +12,9 @@ permalink: /reference/faq/
 # What languages does ClusterFuzz support?
 
 ClusterFuzz definitely supports C, C++ when compiled with
-[clang](https://clang.llvm.org/). It also supports Rust and it probably supports
-other languages that can be compiled with an LLVM-based toolchain (e.g. Swift).
+[clang](https://clang.llvm.org/). It has also been tested with Rust and may work
+with other languages that can be compiled with an LLVM-based toolchain (e.g.
+Swift).
 
 ## Why is an LLVM-based toolchain needed for full support?
 
--- a/docs/reference/FAQ.md
+++ b/docs/reference/FAQ.md
@@ -30,7 +30,7 @@ An LLVM-based toolchain is needed for full ClusterFuzz support for two reasons:
 We strongly recommend using an LLVM-based toolchain. That said, you may be able
 to use some parts of ClusterFuzz with [GCC](https://gcc.gnu.org/) or extend
 ClusterFuzz to support your use of GCC. However, if you do this, you are in
-unexplored territory and we can not support this. GCC can probably be used for
+unexplored territory and we can not support this. GCC may work for
 [blackbox fuzzing]({{ site.baseurl }}/setting-up-fuzzing/blackbox-fuzzing/)
 without modifying ClusterFuzz since GCC supports ASan. Using GCC with AFL or
 libFuzzer would probably require considerable changes to ClusterFuzz to get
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
Closes #125 and #226

docs/reference/FAQ.md
```diff
@@ -0,0 +1,63 @@
...
+
+* Because use of [libFuzzer and AFL] in ClusterFuzz requires coverage
+  instrumentation only LLVM can provide.
+
+## What if I really don't want to use an LLVM-based toolchain?
+
+We strongly recommend using an LLVM-based toolchain. That said, you may be able
+to use some parts of ClusterFuzz with [gcc](https://gcc.gnu.org/) or extend
+ClusterFuzz to support your use of gcc. However, if you do this, you are in
+unexplored territory and we can not support this. Gcc can probably be used for
```
Dor1s: nit: "Gcc" -> "GCC"

Dor1s: I'd also change all "gcc" to "GCC". The official page seems to use all caps version: https://gcc.gnu.org/

docs/reference/FAQ.md
```diff
@@ -0,0 +1,63 @@
...
+find interesting bugs in
+[memory safe](https://en.wikipedia.org/wiki/Memory_safety) languages so
+ClusterFuzz doesn't have support for fuzzing most of them.
+
+## How can I extend ClusterFuzz to support fuzzing my language
+
+Only experts should consider doing this. The first step is getting ClusterFuzz
+to recognize errors thrown by programs written in your language. This should
+make blackbox fuzzing possible. If you would like to add coverage-guided fuzzing
+support for your language, you need to a launcher like there is for AFL and
```
Dor1s: "to add a launcher" or "to write a launcher" ?

docs/reference/FAQ.md
```diff
@@ -0,0 +1,63 @@
...
+---
+
+- TOC
+{:toc}
+
+# What languages does ClusterFuzz support?
+
+ClusterFuzz definitely supports C, C++ when compiled with
+[clang](https://clang.llvm.org/). It should also support Rust and it probably
+supports other languages that can be compiled with an LLVM-based toolchain (e.g.
```
Dor1s: I think that some statements on this page might be confusing. E.g. here we mention LLVM-based toolchain is a requirement (or a recommendation? "should also support" and "probably supports" sound weird?), but later on it becomes clear that a more important piece is the sanitizers. So, apparently you can do e.g. blackbox fuzzing with a target compiled with GCC and ASan, so LLVM toolchain is not a strong requirement.

I feel like we should think more on the structure of this page.

docs/reference/FAQ.md
```diff
@@ -0,0 +1,63 @@
...
+
+In theory, ClusterFuzz is language agnostic and can be extended to support
+fuzzing code written in any language. In practice though, it is less easy to
+find interesting bugs in
+[memory safe](https://en.wikipedia.org/wiki/Memory_safety) languages so
+ClusterFuzz doesn't have support for fuzzing most of them.
+
+## How can I extend ClusterFuzz to support fuzzing my language
+
+Only experts should consider doing this. The first step is getting ClusterFuzz
```
Dor1s: The first sentence might be scaring people off. Are you sure about it?

docs/reference/FAQ.md
```diff
@@ -0,0 +1,63 @@
...
+to recognize errors thrown by programs written in your language. This should
```
Dor1s: Would be good to mention stack_analyzer.py and the corresponding tests, as that's where new support must be added?

docs/reference/FAQ.md
Dor1s: Also, we might want to change this part in the foreseeable future (https://github.com/google/oss-fuzz/issues/611#issuecomment-464525025). I can hardly imagine anyone adding support for a new coverage-guided fuzzing thing before that. May just skip this part.

docs/reference/FAQ.md
```diff
@@ -0,0 +1,63 @@
...
+make blackbox fuzzing possible. If you would like to add coverage-guided fuzzing
+support for your language, you need to a launcher like there is for AFL and
+libFuzzer that is capable of running fuzzers written in your language.
+
+## Are there plans to support fuzzing in other languages?
+
+The only language we are considering adding support for at this time is
```
Dor1s: Are we ready to commit to this here? I think we can remove this question.

docs/reference/FAQ.md
```diff
@@ -0,0 +1,63 @@
...
+to use some parts of ClusterFuzz with [gcc](https://gcc.gnu.org/) or extend
+ClusterFuzz to support your use of gcc. However, if you do this, you are in
+unexplored territory and we can not support this. Gcc can probably be used for
+[blackbox fuzzing]({{ site.baseurl }}/setting-up-fuzzing/blackbox-fuzzing/)
+without modifying ClusterFuzz since gcc supports ASan. Using gcc with AFL or
+libFuzzer would probably require considerable changes to ClusterFuzz to get
+working. In the end, it is probably easier to make whatever changes are needed
+to build with clang.
+
+## Why isn't $MY_LANGUAGE Supported?
```
Dor1s: nit: "Supported" -> "supported"

docs/reference/FAQ.md
```diff
@@ -0,0 +1,63 @@
...
+
+In theory, ClusterFuzz is language agnostic and can be extended to support
+fuzzing code written in any language. In practice though, it is less easy to
+find interesting bugs in
+[memory safe](https://en.wikipedia.org/wiki/Memory_safety) languages so
+ClusterFuzz doesn't have support for fuzzing most of them.
+
+## How can I extend ClusterFuzz to support fuzzing my language
```
Dor1s: nit: add trailing "?"

docs/reference/FAQ.md
jonathanmetzman: I think so. It seems pretty difficult.

docs/reference/FAQ.md
jonathanmetzman: Do you think we should link to code? What if the code changes? Maybe just link to stack_analyzer.py?

docs/reference/FAQ.md
jonathanmetzman: > I think that some statements on this page might be confusing. E.g. here we mention LLVM-based toolchain is a requirement (or a recommendation? "should also support" and "probably supports" sound weird?),

The statements are intentionally a bit vague because we don't really know =)
I think CF supports rust because we have some rust fuzzers, but I'm not actually sure the support is fully there.

>but later on it becomes clear that a more important piece is the sanitizers. So, apparently you can do e.g. blackbox fuzzing with a target compiled with GCC and ASan, so LLVM toolchain is not a strong requirement.
> 
> I feel like we should think more on the structure of this page.

I've made the page less ambiguous. I initially wrote it so that it would scare people away from using GCC while letting brave souls know they could actually use it, I think the new version basically does the same while being less unclear.

jonathanmetzman:
Marty what do you think?

docs/reference/FAQ.md
```diff
@@ -0,0 +1,48 @@
...
+permalink: /reference/faq/
+---
+
+- TOC
+{:toc}
+
+# What languages does ClusterFuzz support?
+
+ClusterFuzz definitely supports C, C++ when compiled with
+[clang](https://clang.llvm.org/). It also supports Rust and it probably supports
```
mbarbella-chromium: Nit: Instead of explicitly mentioning support in the second sentence, could you say something like "It has also been tested with rust and is expected to work with..."

docs/reference/FAQ.md
```diff
@@ -0,0 +1,48 @@
...
+
+* Because use of [libFuzzer and AFL] in ClusterFuzz requires coverage
+  instrumentation only LLVM can provide.
+
+## What if I really don't want to use an LLVM-based toolchain?
+
+We strongly recommend using an LLVM-based toolchain. That said, you may be able
+to use some parts of ClusterFuzz with [GCC](https://gcc.gnu.org/) or extend
+ClusterFuzz to support your use of GCC. However, if you do this, you are in
+unexplored territory and we can not support this. GCC can probably be used for
```
mbarbella-chromium: It's fine to leave this as-is, but it may be worth explicitly calling out some of the things we'd expect to break (deduplication, anything the relies on having reliable crashes) here to make it clearer to anyone who wants this support that they should try to use clang/sanitizers. I'm a bit torn since it seems to be explained above already, but I could see people being put off by this.

jonathanmetzman: I've never used asan with gcc, but do we expect the stack traces to be wrong?

jonathanmetzman: Landing this. I can change in new CL if you'd like.

docs/reference/FAQ.md
jonathanmetzman: I made the change WRT rust, but I now use weaker language for swift et al.

PR #228, opened 2019-02-20
[docs] Mention that trunk clang should be used

jonathanmetzman:


docs/setting-up-fuzzing/libfuzzer_and_afl.md
```diff
@@ -20,11 +20,12 @@ features such as dictionaries and seed corpus.
...
 ### Compiler
 LibFuzzer and AFL need to use instrumentation from the Clang compiler. In our
-documentation, we use features provided by Clang **6.0** or greater. To get a
-Clang that can be used to follow the examples, download one from the [Clang
-releases page] or install one using your package manager. We will refer to these
-compilers as `$CC` and `$CXX`. Set these in the environment so that you can copy
-and paste the example commands:
+documentation, we use features provided by Clang **6.0** or greater. However, we
+recommend using as close to ToT clang as possible. To get a Clang that can be
+used to follow the examples, download one from the [Clang releases page] or
```
Dor1s: Even 8.0 is not on the releases page, but we're on 9.0 now. Let's add a small section on how to either build clang or use Chromium's bundle?

jonathanmetzman: :\ Chrome's clang doesn't ship libFuzzer's runtime with main.

Dor1s: :| let me think what else we can do

Dor1s: Maybe the snapshots page? https://llvm.org/builds/, it points to http://apt.llvm.org/ where you can learn how to get the 9 version?

mhlakhani: Can we use something other than `ToT`, or spell out the abbreviation (I assume it means Tip of Trunk)? It took me a few minutes to understand it and searches were not helpful.

jonathanmetzman: @mhlakhani
Is trunk an understandable term?

mhlakhani: @jonathanmetzman  I think "trunk" / "latest" / "master" all make sense. Trunk shows up if you google "llvm trunk" so that may be useful.

PR #235, opened 2019-02-22
Fix stack overflow caused by rewrapping sys.stdout

### LINKED ISSUE #234: Infinite recursion/stack-overflow in stack_symbolizer.py:72 when starting new bot "thread"
A couple times when I've left the local development server and a bot running for a while I see this issue happening repeatedly. See
[bot-log.txt](https://github.com/google/clusterfuzz/files/2894457/bot-log.txt)
([server-log.txt](https://github.com/google/clusterfuzz/files/2894469/server-log.txt) posted for completeness). This happens over and over again. The bot instance can't make any progress.

This is currently happening on my machine. It might be hard to debug because starting a new bot doesn't seem to repro the issue right away. I think I will need to attach to the currently running process if I want to debug this issue without waiting days possibly to repro this again.

I don't really understand what is wrong with the line where this issue occurs. It seems correct to me.
But I can't say I'm surprised that something like stack_symbolizer.py:disable_buffering() goes wrong with multiprocessing involved :-\

jonathanmetzman:
Fixes #234
Don't wrap `sys.stdout` if it is already wrapped. 
Previous assumptions that it would only be wrapped if env var `PYTHONUNBUFFERED` was set, this is no longer true, so explicitly check if it is already wrapped before wrapping.

### DIFF
--- a/src/python/crash_analysis/stack_parsing/stack_symbolizer.py
+++ b/src/python/crash_analysis/stack_parsing/stack_symbolizer.py
@@ -138,11 +138,14 @@ def chrome_dsym_hints(binary):
 
 def disable_buffering():
   """Make this process and child processes stdout unbuffered."""
-  if not os.environ.get('PYTHONUNBUFFERED'):
-    # Since sys.stdout is a C++ object, it's impossible to do
-    # sys.stdout.write = lambda...
+  os.environ['PYTHONUNBUFFERED'] = 'x'
+
+  if not isinstance(sys.stdout, LineBuffered):
+    # Don't wrap sys.stdout if it is already wrapped.
+    # See https://github.com/google/clusterfuzz/issues/234 for why.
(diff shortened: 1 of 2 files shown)


src/python/crash_analysis/stack_parsing/stack_symbolizer.py
```diff
@@ -138,11 +138,14 @@ def chrome_dsym_hints(binary):
 
 def disable_buffering():
   """Make this process and child processes stdout unbuffered."""
-  if not os.environ.get('PYTHONUNBUFFERED'):
-    # Since sys.stdout is a C++ object, it's impossible to do
-    # sys.stdout.write = lambda...
+  os.environ['PYTHONUNBUFFERED'] = 'x'
```
mbarbella-chromium: Is this still needed?

jonathanmetzman: I think we decided offline to keep it.

PR #236, opened 2019-02-22
Add a log message when deleting test cases in grouper.

mbarbella-chromium:
Trivial change, PTAL when you have time. This is the only obvious case I see where we aren't logging after deleting a test case, but there may be bulk cases I missed.


src/appengine/handlers/cron/grouper.py
```diff
@@ -99,6 +99,7 @@ def group_testcases():
     # Remove duplicates early on to avoid large groups.
     if (not testcase.bug_information and not testcase.uploader_email and
         has_testcase_with_same_params(testcase, testcase_map)):
+      logs.log('Deleting duplicate testcase %s.' % testcase_id)
```
Dor1s: Is `testcase_id` a string, not an integer?

mbarbella-chromium: Looks like get_open_testcase_id_iterator yields integers. I'll switch this to %d.

PR #238, opened 2019-02-28
Add fork mode as a strategy

jonathanmetzman:
Use libFuzzer's fork mode on 10% of runs.
Leave use of `-fork=2` on Windows (which have two cores in our CF instance) as a TODO.
Also, remove assertions from integration tests that assert libFuzzer behaves a way that is not correct for ToT.


jonathanmetzman:
Will up the strategy to 50 or more if it doesn't break things.
I can undo the assertion removal as it doesn't break anything. I didn't update the binaries and add an integration test because I really don't know about this binary files thing. Updating all the libFuzzer targets changes the size of the .git dir from 39M on my machine to 47.

src/python/bot/fuzzers/libFuzzer/launcher.py
```diff
@@ -787,6 +796,11 @@ def main(argv):
     arguments.append(constants.VALUE_PROFILE_ARGUMENT)
     fuzzing_strategies.append(strategy.VALUE_PROFILE_STRATEGY)
 
+  if do_fork():
+    # TODO(metzman): Use `-fork=2` on Windows.
+    arguments.append('%s%d' % (constants.FORK_FLAG, 1))
```
Dor1s: What does the value of this argument mean? Would it be better to move it into another constant and use on both lines 801 and 802. Would be a bit more convenient when you enable `2`.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_integration_test.py
```diff
@@ -135,9 +137,6 @@ def assert_has_stats(self, output, testcase_path):
 
     self.assertTrue(os.path.exists(testcase_path + '.stats2'))
 
-  def assert_corpus_loaded(self, output, directory):
```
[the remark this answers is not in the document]
jonathanmetzman: They fail on targets compiled with a new clang. I didn't exactly determine why though. @mmoroz may have some ideas because the "corpus loaded" string seems to only be used with dictionary analysis.
I'm searching for the strings in git and the only commits that seem to have this string are from before the compiler-rt move, in other words ancient.

jonathanmetzman:
> What about crash detection? Kostya has mentioned that we may have several crash inputs. Would CF handle that well?

I don't think we will have to deal with multiple crashes at this point since we only have `-fork=1` in use. There will be multiple OOMs and timeouts, for now we will just ignore all of them.

src/python/bot/fuzzers/libFuzzer/launcher.py
jonathanmetzman: Done

PR #239, opened 2019-03-01
Fix CSP for various upload forms.

mbarbella-chromium:
TBR for a small fix. Missed a case on some upload forms (e.g. test case, fuzzer).


inferno-chromium:
Really worried about the CSP and what stuff it can silently break. Can you please try all pages and dive deeper into flow, e.g. coverage viewer in fuzzer stats, google cloud storage links in fuzzers and data bundle page, etc to make these all work. also, maybe looking at exceptions for last x days ?

PR #240, opened 2019-03-01
Add a CSP builder class and use it to generate policy.

mbarbella-chromium:
This serves a few purposes:
* It is now easier to add a specific policy to a page (though we don't
do this yet).
* When we do use page-specific (or project-specific) policies, they
can build off of the default instead of duplicating the entire policy
with small modifications.
* The policy is explained better through comments in the
get_default_builder function.


mbarbella-chromium:
PTAL when you have time. This is just cleanup on top of the existing CSP changes, shouldn't change the current behavior.

jonathanmetzman:
Will try to take a look at this early monday

PR #241, opened 2019-03-05
Add missing handler decorators.

mbarbella-chromium:
Small fix to get CSP on the test case details page and clean up a few other cases.

### DIFF
--- a/src/appengine/handlers/corpora.py
+++ b/src/appengine/handlers/corpora.py
@@ -48,6 +48,7 @@ class CreateHandler(base_handler.Handler):
   """Create a corpus."""
 
   @handler.check_user_access(need_privileged_access=True)
+  @handler.post(handler.FORM, handler.HTML)
   @handler.require_csrf_token
   def post(self):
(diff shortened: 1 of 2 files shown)


src/appengine/handlers/corpora.py
```diff
@@ -48,6 +48,7 @@ class CreateHandler(base_handler.Handler):
   """Create a corpus."""
 
   @handler.check_user_access(need_privileged_access=True)
+  @handler.post(handler.FORM, handler.HTML)
```
jonathanmetzman: Not directly relevant to this PR:
Does this decorator need to be added to every `post` handler? 
Couldn't this be done in a less repetitive way?

mbarbella-chromium: Probably, but I can't think of anything obvious that would completely remove the issue. We'd still need to provide the content type information somehow, so even if we did it without the decorator we'd sill need to set it somewhere. It's definitely worth considering though since this seems fairly error-prone.

PR #243, opened 2019-03-07
 Add profiling support using Google Cloud Profiler

inferno-chromium:


src/python/tests/core/metrics/profiler_test.py
```diff
@@ -0,0 +1,65 @@
+# Copyright 2018 Google LLC
```
Dor1s: nit: Should we make it 2019 as well?

inferno-chromium: Done

Issue #244, opened 2019-03-07
create_config command should sanity check if prereqs (gcloud auth) is done


mykter:
I almost followed the instructions to setup a production instance, but put my client secrets in the config dir as that seemed logical.

`python butler.py create_config --oauth-client-secrets-path=$CLIENT_SECRETS_PATH --project-id=$CLOUD_PROJECT_ID $CONFIG_DIR`
runs for a while, and creates the `default` and `cron-service` services, displaying their info on the console.

It then moves on to:
`Running: python butler.py run setup --config-dir /home/mmacnair/clusterfuzz-config/ --non-dry-run`

which after several minutes of no output eventually fails with this backtrace:
```
| Traceback (most recent call last):
|   File "butler.py", line 282, in <module>
|     main()
|   File "butler.py", line 256, in main
|     command.execute(args)
|   File "src/local/butler/run.py", line 41, in execute
|     script.execute(args)
|   File "src/local/butler/scripts/setup.py", line 268, in execute
|     setup_config(args.non_dry_run)
|   File "src/local/butler/scripts/setup.py", line 200, in setup_config
|     config = data_types.Config.query().get()
|   File "/home/mmacnair/clusterfuzz/src/python/datastore/ndb_patcher.py", line 144, in get
|     return _retry_wrap(result_func)()
|   File "/home/mmacnair/clusterfuzz/src/third_party/google/api_core/retry.py", line 270, in retry_wrapped_func
|     on_error=on_error,
|   File "/home/mmacnair/clusterfuzz/src/third_party/google/api_core/retry.py", line 179, in retry_target
|     return target()
|   File "/home/mmacnair/clusterfuzz/src/python/datastore/ndb_patcher.py", line 143, in <lambda>
|     result_func = lambda: next(self.iter(limit=1), None)
|   File "/home/mmacnair/clusterfuzz/src/python/datastore/ndb_patcher.py", line 162, in iter
|     self, projection=projection, keys_only=keys_only)
|   File "/home/mmacnair/clusterfuzz/src/python/datastore/ndb_patcher.py", line 399, in _ndb_query_to_cloud_queries
|     subqueries.append(_client().query(
|   File "/home/mmacnair/clusterfuzz/src/python/datastore/ndb_patcher.py", line 201, in _client
|     init()
|   File "/home/mmacnair/clusterfuzz/src/python/datastore/ndb_patcher.py", line 418, in init
|     creds = credentials.get_default()[0]
[...]

PR #246, opened 2019-03-08
Use empty directory for config (#244)

inferno-chromium:
To avoid confusion with putting creds initially in a config directory, that get deleted by create_config command


inferno-chromium:
TBR simple comment change.

oliverchang:
LGTM

PR #247, opened 2019-03-08
 Add reporting for CSP violations.

mbarbella-chromium:
This still needs some testing to ensure the policy changes don't break anything, but the reporting aspect of it should be ready for review. A few of the policy changes have some pretty nice benefits.

PTAL when you have time. I'll ensure that pages don't break on staging before landing.

### DIFF
--- a/src/appengine/handlers/csp_reports.py
+++ b/src/appengine/handlers/csp_reports.py
@@ -26,11 +26,13 @@ class CspReportHandler(base_handler.Handler):
     logs.log_error('CSP violation: {}'.format(self.request.get('csp-report')))
 
   @handler.get(handler.JSON)
+  @handler.check_user_access(need_privileged_access=False)
   def get(self):
     """Handle a GET request."""
     self.log_csp_violation()
 
   @handler.post(handler.JSON, handler.JSON)
+  @handler.check_user_access(need_privileged_access=False)
   def post(self):
     """Handle a POST request."""
     self.log_csp_violation()
--- a/src/appengine/handlers/csp_reports.py
+++ b/src/appengine/handlers/csp_reports.py
@@ -1,0 +1,36 @@
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Log incoming reports of CSP violations.."""
+
+from handlers import base_handler
+from libs import handler
+from metrics import logs
+
+
+class CspReportHandler(base_handler.Handler):
+  """Redirect to documentation."""
+
+  def log_csp_violation(self):
+    """Create an error log for a CSP violation."""
+    logs.log_error('CSP violation: {}'.format(self.request.get('csp-report')))
+
+  @handler.get(handler.JSON)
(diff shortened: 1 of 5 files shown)


src/appengine/handlers/csp_reports.py
```diff
@@ -0,0 +1,36 @@
...
+
+class CspReportHandler(base_handler.Handler):
+  """Redirect to documentation."""
+
+  def log_csp_violation(self):
+    """Create an error log for a CSP violation."""
+    logs.log_error('CSP violation: {}'.format(self.request.get('csp-report')))
+
+  @handler.get(handler.JSON)
+  def get(self):
```
inferno-chromium: We can't do this without auth, this would just mean arbitrary log injection.

src/python/tests/appengine/libs/csp_test.py
```diff
@@ -29,10 +29,11 @@ def test_simple_policy(self):
     builder.add('connect-src', 'self', quote=True)
     builder.add('script-src', 'self', quote=True)
     builder.add('script-src', 'scripts.test.tld')
+    builder.add_sourceless('block-all-mixed-content')
```
inferno-chromium: Shouldn't we test with upgrade-insecure-requests that is actually used in code ?

mbarbella-chromium: Good catch. I initially had block-all-mixed-content in both places but didn't think to update it here.

src/appengine/handlers/csp_reports.py
mbarbella-chromium: That's fair, though we do lose some coverage from this (e.g. if there was ever a failure on the login page we'd never see it). That should be a pretty rare or minor case though, so seems better to require auth.

src/appengine/handlers/csp_reports.py
```diff
@@ -0,0 +1,38 @@
...
+  """Redirect to documentation."""
+
+  def log_csp_violation(self):
+    """Create an error log for a CSP violation."""
+    logs.log_error('CSP violation: {}'.format(self.request.get('csp-report')))
+
+  @handler.get(handler.JSON)
+  @handler.check_user_access(need_privileged_access=False)
+  def get(self):
+    """Handle a GET request."""
```
inferno-chromium: Why do we need the GET handler ? it makes stuff confusing as people can call on it.

src/appengine/handlers/csp_reports.py
```diff
@@ -0,0 +1,38 @@
...
+from libs import handler
+from metrics import logs
+
+
+class CspReportHandler(base_handler.Handler):
+  """Redirect to documentation."""
+
+  def log_csp_violation(self):
+    """Create an error log for a CSP violation."""
+    logs.log_error('CSP violation: {}'.format(self.request.get('csp-report')))
```
inferno-chromium: can we put self.request.get('csp-report') into a local and check if it is non-null. if null, raise that ExitException

src/appengine/server.py
```diff
@@ -162,6 +163,7 @@ def get(self, _):
     ('/corpora', corpora.Handler),
     ('/corpora/create', corpora.CreateHandler),
     ('/corpora/delete', corpora.DeleteHandler),
+    ('/csp-reports', csp_reports.CspReportHandler),
```
inferno-chromium: should this be named report-csp-failure ? since this is not a page to view csp-reports but to report a particular failure ?

mbarbella-chromium:
Additional comments addressed.

src/appengine/handlers/report_csp_failure.py
```diff
@@ -0,0 +1,33 @@
...
+"""Log incoming reports of CSP violations.."""
+
+from handlers import base_handler
+from libs import handler
+from libs import helpers
+from metrics import logs
+
+
+class ReportCspFailureHandler(base_handler.Handler):
+  """Redirect to documentation."""
```
inferno-chromium: Please update this docstring.

PR #248, opened 2019-03-08
Fix IOError when symbolizing on macOS.

inferno-chromium:
Uptake https://llvm.org/svn/llvm-project/compiler-rt/trunk@348659


inferno-chromium:
This should fix
IOError: [Errno 32] Broken pipe Traceback (most recent call last): File "/b/clusterfuzz/src/python/crash_analysis/stack_parsing/stack_symbolizer.py", line 274, in symbolize print >> self.pipe.stdin, symbolizer_input IOError: [Errno 32] Broken pipe
at symbolize (/b/clusterfuzz/src/python/crash_analysis/stack_parsing/stack_symbolizer.py:274)

jonathanmetzman:
LGTM

PR #249, opened 2019-03-09
More CSP fixes.

mbarbella-chromium:
* Analytics depends on img-src
* We load a code prettify script from cdn.rawgit.com

### DIFF
--- a/src/appengine/libs/csp.py
+++ b/src/appengine/libs/csp.py
@@ -85,9 +85,10 @@ def get_default_builder():
   # Google Analytics also uses img-src.
   builder.add('img-src', 'www.google-analytics.com')
 
-  # External style. Used for fonts and charting libraries.
+  # External style. Used for fonts, charting libraries, and code formatter.
   builder.add('style-src', 'fonts.googleapis.com')
   builder.add('style-src', 'www.gstatic.com')
+  builder.add('style-src', 'cdn.rawgit.com')
 
   # External fonts.
   builder.add('font-src', 'fonts.gstatic.com')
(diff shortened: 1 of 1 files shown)


src/appengine/libs/csp.py
```diff
@@ -77,9 +77,13 @@ def get_default_builder():
   builder.add('img-src', 'self', quote=True)
   builder.add('manifest-src', 'self', quote=True)
 
-  # External scripts. Google analytics and charting libraries we depend on.
+  # External scripts. Google analytics, charting libraries, and code formatter.
   builder.add('script-src', 'www.google-analytics.com')
   builder.add('script-src', 'www.gstatic.com')
+  builder.add('script-src', 'cdn.rawgit.com')
```
alex: This probably ought to have a TODO comment. AFAICT, it's a complete CSP bypass, since anyone can host any code they want on there.

mbarbella-chromium: Good catch! I'll move this to the group of unsafe things we need to fix.

mbarbella-chromium:
TBR for some small fixes. Abhishek, PTAL when you have time.

Do you remember why we load the code prettify script from rawgit? It looks like they're shutting down relatively soon, and we shouldn't allow this in the policy anyway for the reasons Alex mentioned.

inferno-chromium:
We should revert this last fix, cdn.rawgit.com is insecure. We can just host the script ourselves.

inferno-chromium:
Hosting google-code-prettify needs to be more thought, their code hardcodes the cdn urls (https://github.com/google/code-prettify/blob/master/src/run_prettify.js#L231). We can create a modified version, but i hate that for maintainability. Till then, lets just disable prettyprint (which i just like a lot, but security supercedes everything). https://github.com/google/clusterfuzz/pull/250

Issue #260, opened 2019-03-13
Uploads are broken in local development due to CSP


inferno-chromium:

inferno-chromium:
Fixed in https://github.com/google/clusterfuzz/commit/1c54212198744fe4ed81c095fb6f5798879943e8

PR #261, opened 2019-03-13
Don't enforce content security policies in local development mode (#260).

inferno-chromium:


jonathanmetzman:
Shouldn't the development mode model the production one? This might make more of a mess in the long run.

inferno-chromium:
Development mode relies on emulators, so can't always model the production mode exactly. Emulators rely on different local ports (see https://github.com/google/clusterfuzz/blob/master/src/local/butler/constants.py#L65), so we gets different hostnames localhost:9000, localhost:9008 for gcs, etc which fails in CSP. I dont see value in enabling CSP locally since it never talks to external hosts, but only emulators on local ports.

jonathanmetzman:
> Development mode relies on emulators, so can't always model the production mode exactly. Emulators rely on different local ports (see https://github.com/google/clusterfuzz/blob/master/src/local/butler/constants.py#L65), so we gets different hostnames localhost:9000, localhost:9008 for gcs, etc which fails in CSP. I dont see value in enabling CSP locally since it never talks to external hosts, but only emulators on local ports.

ACK

PR #262, opened 2019-03-13
Remove fork mode flag before merge step.

inferno-chromium:


inferno-chromium:
Basically noticed breakage when testing heartbleeed example locally. If fork mode strategy was picked, then merge step would hang since it would have -fork=1 and that will cause it to run fuzzing for that time duration :(

inferno-chromium:
Question for you guys: should max_len be removed before merge step ? that would be the only big change after this and need test updates.

Dor1s:
> Question for you guys: should max_len be removed before merge step ?

Yes, absolutely

PR #264, opened 2019-03-13
Use available cpus for libFuzzer fork mode.

inferno-chromium:


src/python/bot/fuzzers/libFuzzer/launcher.py
```diff
@@ -803,8 +809,8 @@ def main(argv):
     fuzzing_strategies.append(strategy.VALUE_PROFILE_STRATEGY)
 
   if do_fork():
-    # TODO(metzman): Use `-fork=2` on Windows.
-    num_fuzz_processes = 1
+    max_fuzz_threads = environment.get_value('MAX_FUZZ_THREADS', 1)
+    num_fuzz_processes = max(1, multiprocessing.cpu_count() / max_fuzz_threads)
```
jonathanmetzman: I don't like the cpu count stuff. Maybe do it in another CL? 
The rest of the CL LGTM, but I think this leads to some bad problems.
eg: On Windows (assuming cmd line issue is solved): 
max_fuzz_threads = 2
num_fuzz_processes = max(1, multiprocessing.cpu_count() / max_fuzz_threads) -> num_fuzz_processes = max(1, 2/2) ->  num_fuzz_processes = 1.

And then there's the issue of multiple bots, not to mention the fact that load from running one bot will be higher than what people expect.

src/python/bot/fuzzers/libFuzzer/launcher.py
```diff
@@ -163,6 +164,11 @@ def do_value_profile():
 
 def do_fork():
   """Return whether or not to do fork mode."""
+  # TODO(metzman): Find a workaround for Windows command line limit before
```
jonathanmetzman: I have a theory that using subset allows fork mode to work. Might be worth looking into.

src/python/bot/fuzzers/libFuzzer/launcher.py
inferno-chromium: On Windows, max_fuzz_threads is still 1. See windows_libfuzzer_chrome_asan job just uses libfuzzer template which has MAX_FUZZ_THREADS = 1
num_fuzz_processes = max(1, multiprocessing.cpu_count() / max_fuzz_threads)
= max(1,2/1) = 2
We never run multiple bots on same machine in production (other than Android which might need more thought and libfuzzer is not running in that config yet). Multiple bots people might try locally, but we also don't recommend in docs.
The only thing i worry about is non-optimal config when people try libFuzzer/AFL locally on multi-core machines OR on >n1-standard-1 bots. We might think of solving that via MAX_FUZZ_THREADS, so lets think more on this.

src/python/bot/fuzzers/libFuzzer/launcher.py
inferno-chromium: filed https://github.com/google/clusterfuzz/issues/267

PR #266, opened 2019-03-14
Recommend using last release than master

inferno-chromium:


mhlakhani:
thanks for putting this up so quickly!

PR #269, opened 2019-03-14
Fix possible bug in list2cmdline overriding.

jonathanmetzman:
Make sure to restore list2cmdline even if process_handle.run excepts.


jonathanmetzman:
PTAL.

When trying to fix [this issue](https://bugs.chromium.org/p/chromium/issues/detail?id=925988). I noticed a possible bug in this code. I'm not sure if this hack is still needed or what exactly it is used for, so I would be happier if we could get rid of it.

PR #270, opened 2019-03-14
Fix launcher handling

jonathanmetzman:
Fix handling of executable and Java launchers.

Rename get_launch_path_for_script to get_executable_command, and make it reusable.
Add tests for get_launch_path_for_script.
Make tests in process_handler_test.py obey conventions.

Bug: crbug.com/925988

### DIFF
--- a/src/python/base/utils.py
+++ b/src/python/base/utils.py
@@ -33,7 +33,6 @@ from base import retry
 from config import local_config
 from metrics import logs
 from system import environment
-from system import shell
 
 try:
   import psutil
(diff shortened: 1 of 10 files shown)


src/python/base/utils.py
```diff
@@ -308,15 +308,19 @@ def get_file_contents_with_fatal_error_on_failure(path):
...
-def get_launch_path_for_script(script_directory, script_filename):
-  """Return launch path for a script."""
-  # Hack for Java Scripts.
-  script_filename = script_filename.replace('.class', '')
-
-  interpreter_path = shell.get_interpreter_for_command(script_filename)
-  script_absolute_path = os.path.join(script_directory, script_filename)
-  script_launch_path = '%s %s' % (interpreter_path, script_absolute_path)
-  return script_launch_path
+def get_execute_command(file_to_execute):
```
inferno-chromium: i think this could be moved to shell.py as well just like get_interpreter_for_command

src/python/tests/core/base/utils_test.py
```diff
@@ -453,3 +453,33 @@ def test_longer_than_one_chunk(self):
       file_handle.write('A' * 60000)
     self.assertEqual('8360c01cef8aa7001d1dd8964b9921d4c187da29',
                      utils.file_hash(self.test_file))
+
+
+class GetExecuteCommand(unittest.TestCase):
```
inferno-chromium: Should be moved to shell_test.py based on comment above.

inferno-chromium:
LGTM

PR #271, opened 2019-03-15
Add initial basic Python bazel rules.

oliverchang:


oliverchang:
Yep will add those in future PRs!

PR #276, opened 2019-03-15
Use full path to fuzzer executable.

inferno-chromium:


inferno-chromium:
minor regression from https://github.com/google/clusterfuzz/commit/69e4108d9dd7871ebabbbe3df3212549fc051df4#diff-6bbe1ef1dddc4f3eedadcee180e3e41f

PR #277, opened 2019-03-15
Store daily fuzzer weights in bigquery.

mbarbella-chromium:
PTAL when you have time.

I'm having trouble finding docs on how exactly _PARTITIONTIME gets populated if you happen to know. It looks like it's null for newly added rows, and I'm guessing it updates those in a batch once per interval (daily in our cases)? If so, I think this seems to be working, but if not it may need a fix.

### DIFF
--- a/src/appengine/handlers/cron/fuzzer_weights.py
+++ b/src/appengine/handlers/cron/fuzzer_weights.py
@@ -325,7 +325,7 @@ def store_current_weights_in_bigquery():
         'job': target_job.job,
         'weight': target_job.weight
     }
-    rows.append(big_query.Insert(row, None))
+    rows.append(big_query.Insert(row=row, insert_id=None))
 
   client = big_query.Client(dataset_id='main', table_id='fuzzer_weights')
   client.insert(rows)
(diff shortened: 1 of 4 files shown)


src/appengine/handlers/cron/fuzzer_weights.py
```diff
@@ -317,6 +315,22 @@ def update_target_weights_for_engine(client, engine, specifications):
...
+  """Update a bigquery table containing the daily stats."""
+  rows = []
+  target_jobs = ndb_utils.get_all_from_model(data_types.FuzzTargetJob)
+  for target_job in target_jobs:
+    row = {
+        'fuzzer': target_job.fuzz_target_name,
+        'job': target_job.job,
+        'weight': target_job.weight
+    }
+    rows.append(big_query.Insert(row, None))
```
inferno-chromium: row=row, insert_id=None

Also, is it ok to ignore insert_id ?

mbarbella-chromium: Done. insert_id is an optional field, so this should be fine. In testing it seemed like all of the rows were inserted properly.

PR #285, opened 2019-03-20
Fix incorrect directory traversal error with ./

inferno-chromium:


inferno-chromium:
Why do tar file allow './' prefixes :(

src/python/tests/core/system/archive_test.py
```diff
@@ -23,18 +23,16 @@
...
-    if not zip_path:
-      raise Exception('Please set ZIP_PATH.')
-
-    output_directory = tempfile.mkdtemp(prefix='archive-py')
-    archive.unpack(zip_path, output_directory, trusted=True)
+  def test_unpack_file_with_cwd_prefix(self):
+    """Test unpack with trusted=False passes with file having './' prefix."""
+    tgz_path = os.path.join(TESTDATA_PATH, 'cwd-prefix.tgz')
+    output_directory = tempfile.mkdtemp(prefix='cwd-prefix')
+    archive.unpack(tgz_path, output_directory, trusted=False)
```
mbarbella-chromium: Do we do anything to clean up this state between runs? It's unlikely, but this could lead to some false positives if the file was left in place.

inferno-chromium: Done!

PR #286, opened 2019-03-20
Add Mutator Plugins as a Strategy

jonathanmetzman:


mhlakhani:
Will we add usage documentation for this after we've verified the feature works well? Really excited to see this idea land and be usable.

jonathanmetzman:
> Will we add usage documentation for this after we've verified the feature works well? Really excited to see this idea land and be usable.

Sure. Though it's a "big if" if this works well. I plan on seeing if it is helpful for us first before deciding if this feature will stick around.

configs/test/project.yaml
```diff
@@ -70,6 +70,9 @@ env:
   # the testcases against the target application.
   FUZZ_LOGS_BUCKET: test-fuzz-logs-bucket
 
+  # Default bucket to store mutator plugins.
+  MUTATOR_PLUGINS_BUCKET: test-mutator-plugins-bucket
```
jonathanmetzman: I suppose this CL is the last chance to decide if we prefer `mutator-plugins` or `mutator-plugin` as the standard bucket name. Wer're not really consistent about this, ie: corpus (singular) vs logs (plural).

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_run_test.py
```diff
@@ -61,6 +61,11 @@ class FuzzerTest(builtin_test.BaseEngineFuzzerTest):
   def test_run(self):
     """Test running libFuzzer fuzzer."""
     libfuzzer = fuzzer.LibFuzzer()
+
+    # Don't test mutator_plugins here.
```
jonathanmetzman: I test it elsewhere so it isn't needed here. But when the test ran before it would spend a really long time and then error out when talking to GCS. Maybe we should get this working?

jonathanmetzman:
Sorry for the spam folks, my local tests weren't quite working so I was using the upload/gcbrun workflow a bit too aggressively.

oliverchang:
for OSS-Fuzz, you'll need to add a line here: https://github.com/google/clusterfuzz/blob/master/src/appengine/handlers/cron/oss_fuzz_setup.py#L745 

to give workers permissions to access the relevant buckets.

jonathanmetzman:
cc @kcc @afd @paulthomson

jonathanmetzman:
@inferno-chromium @mbarbella-chromium PTAL.

src/python/bot/fuzzers/libFuzzer/fuzzer.py
```diff
@@ -60,3 +61,6 @@ def generate_arguments(self, fuzzer_path):
             '%s%d' % (constants.RSS_LIMIT_FLAG, constants.DEFAULT_RSS_LIMIT_MB))
 
     return ' '.join(arguments)
+
+  def get_mutator_plugin(self, fuzzer_binary_name):
+    mutator_plugin.get_mutator_plugin(fuzzer_binary_name)
```
oliverchang: we should move away from the model we have of methods setting some global state (in this case a env var). we should return the path to this plugin instead.

this will also make it easier if we decide to support this as an argument in libFuzzer itself.

src/python/bot/fuzzers/builtin.py
```diff
@@ -86,6 +91,8 @@ def run(self, input_directory, output_directory, no_of_files):
       from bot.untrusted_runner import file_host
       corpus_directory = file_host.rebase_to_worker_root(corpus_directory)
 
+    self.get_mutator_plugin(fuzzer_binary_name)
```
oliverchang: This should go into the launcher. Otherwise this may cause complications with supporting OSS-Fuzz since this code runs on the host, while the launcher runs on the worker.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -142,12 +143,16 @@ def fuzz(self,
...
+    try:
+      return self.run_and_wait(
+          additional_args=additional_args,
+          timeout=fuzz_timeout - self.SIGTERM_WAIT_TIME,
+          terminate_before_kill=True,
+          terminate_wait_time=self.SIGTERM_WAIT_TIME,
+          max_stdout_len=MAX_OUTPUT_LEN)
+    finally:
+      # Make sure to unset LD_PRELOAD if we set it to use a plugin.
+      mutator_plugin.unset_mutator_plugin()
```
oliverchang: the place where we set this should be closer to where we do this unsetting. As I suggested in a previous comment, can we move the setting (and the unsetting) into the launcher? 

a better solution may not need this unset at all if we e.g. add an argument to the fuzz interface to accept a path to a mutator. then we can just pass the env to the target via the process launching without messing up the parent env. i.e.

```python
def fuzz(self, ...., mutator_path=None):
  if mutator_path:
    env['LD_PRELOAD'] = ...
    ...
    self.run_and_wait(..., env=env, ....)
```

src/python/bot/fuzzers/mutator_plugin.py
```diff
@@ -0,0 +1,156 @@
...
+  def create_directories(self):
+    """Creates directories needed to use mutator plugins."""
+    plugins_root_dir = environment.get_value('MUTATOR_PLUGINS_DIR')
+    assert plugins_root_dir
+    shell.remove_directory(plugins_root_dir, recreate=True)
+    os.mkdir(_get_mutator_plugins_archives_dir())
+    os.mkdir(_get_mutator_plugins_unpacked_dir())
+
+  @staticmethod
+  def _extract_name_from_archive(plugin_archive_filename):
```
oliverchang: nit: style guide says not to use staticmethod: 

http://google.github.io/styleguide/pyguide.html#2174-decision

src/python/bot/fuzzers/mutator_plugin.py
```diff
@@ -0,0 +1,156 @@
...
+import os
+
+from base import utils
+from google_cloud_utils import gsutil
+from google_cloud_utils import storage
+from system import archive
+from system import environment
+from system import shell
+
+MUTATOR_PLUGINS_BUCKET_ENV_VAR = 'MUTATOR_PLUGINS_BUCKET'
```
oliverchang: this seems like an unnecessary indirection. just use 'MUTATOR_PLUGINS_BUCKET'  ?

src/python/bot/fuzzers/mutator_plugin.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+def _download_mutator_plugin_archive(mutator_plugin_archive):
+  """Downloads the |mutator_plugin_archive| from the mutator plugin storage
+  bucket to the plugin archives directory. Returns the path archive was
+  downloaded to."""
+  file_path = os.path.join(_get_mutator_plugins_archives_dir(),
+                           mutator_plugin_archive)
+  url = '%s/%s' % (_get_mutator_plugins_bucket_url(), mutator_plugin_archive)
+  assert gsutil.GSUtilRunner().download_file(url, file_path)
```
oliverchang: let's use storage.copy_file_from instead. no need to invoke gsutil here.

src/python/bot/fuzzers/mutator_plugin.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+def set_mutator_plugin():
+  """Sets LD_PRELOAD to the path of a usable mutator plugin shared object.
+  This should only be called after a call to get_mutator_plugin."""
+  paths = shell.get_files_list(_get_mutator_plugins_unpacked_dir())
+  # This function should not be called unless there is an unpacked plugin.
+  for path in paths:
+    if os.path.basename(path) == MUTATOR_SHARED_OBJECT_FILENAME:
+      environment.set_value('LD_PRELOAD', path)
```
oliverchang: as above, I don't htink it's a good idea to set this globally here. This would also mean any subprocess we create loads this, which may lead to unexpected issues down the track.

src/python/bot/fuzzers/libFuzzer/launcher.py
```diff
@@ -815,11 +832,21 @@ def main(argv):
     fuzzing_strategies.append(
         '%s_%d' % (strategy.FORK_STRATEGY, num_fuzz_processes))
 
+  extra_env = {}
+  if do_mutator_plugin():
+    mutator_plugin_path = mutator_plugin.get_mutator_plugin(target_name)
+    if mutator_plugin_path:
+      # TODO(metzman): Change the strategy to log which plugin was used, and not
```
inferno-chromium: just add the logs.log for mutator_plugin_path than leaving a TODO.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -383,7 +393,8 @@ def fuzz(self,
            corpus_directories,
            fuzz_timeout,
            artifact_prefix=None,
-           additional_args=None):
+           additional_args=None,
+           extra_env=None):
```
inferno-chromium: looks like extra_env is not passed to LibFuzzerCommon.fuzz

inferno-chromium: also need a test that shows LD_PRELOAD set to that path.

src/python/system/environment.py
```diff
@@ -836,6 +836,8 @@ def set_bot_environment():
   os.environ['FUZZ_INPUTS_MEMORY'] = os.environ['FUZZ_INPUTS']
   os.environ['FUZZ_INPUTS_DISK'] = os.path.join(inputs_dir,
                                                 'fuzzer-testcases-disk')
+  os.environ['MUTATOR_PLUGINS_DIR'] = os.path.join(inputs_dir,
+                                                   'mutator-plugins')
```
inferno-chromium: This directory needs to be in source code with a README.md file.

src/python/system/new_process.py
```diff
@@ -260,14 +264,25 @@ def run(self,
     Returns:
       A subprocess.Popen object for the process.
     """
+    # TODO: Rename popen_args to popen_kwargs.
     command = self.get_command(additional_args)
 
     if stdout == subprocess.PIPE and max_stdout_len:
       stdout = tempfile.TemporaryFile()
 
+    assert 'env' not in popen_args or extra_env is None
```
inferno-chromium: Any particular reason to block this usecase ?

src/python/bot/fuzzers/mutator_plugin.py
```diff
@@ -0,0 +1,155 @@
...
+from system import archive
+from system import environment
+from system import shell
+
+MUTATOR_SHARED_OBJECT_FILENAME = 'mutator-plugin.so'
+ARCHIVES_SUBDIR_NAME = 'archives'
+PLUGINS_SUBDIR_NAME = 'plugins'
+
+
+def clear_plugins_directory():
```
inferno-chromium: function is unused.

src/python/bot/fuzzers/mutator_plugin.py
```diff
@@ -0,0 +1,155 @@
...
+    self.fuzzer_binary_name = fuzzer_binary_name
+    self.job_name = environment.get_value('JOB_NAME')
+    self.create_directories()
+
+  def create_directories(self):
+    """Creates directories needed to use mutator plugins."""
+    plugins_root_dir = environment.get_value('MUTATOR_PLUGINS_DIR')
+    assert plugins_root_dir
+    shell.remove_directory(plugins_root_dir, recreate=True)
+    os.mkdir(_get_mutator_plugins_archives_dir())
```
inferno-chromium: shell.create_directory_if_needed to avoid cases when it already exists.

src/python/bot/fuzzers/mutator_plugin.py
```diff
@@ -0,0 +1,155 @@
...
+      os.path.splitext(mutator_plugin_archive_path)[0])
+  unpacked_plugin_dir = os.path.join(_get_mutator_plugins_unpacked_dir(),
+                                     mutator_plugin_name)
+  archive.unpack(mutator_plugin_archive_path, unpacked_plugin_dir)
+  return unpacked_plugin_dir
+
+
+def _extract_name_from_archive(plugin_archive_filename):
+  """Parses |plugin_archive_filename| which should be named using the schema:
+  '$NAME-$JOB-$FUZZ_TARGET.zip'. Returns tuple containing
```
inferno-chromium: Did you see my comment in design doc. i think we need an accompanying yaml or .options file that can have this info. name feels too hacky.

src/python/bot/fuzzers/mutator_plugin.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def _download_mutator_plugin_archive(mutator_plugin_archive):
+  """Downloads the |mutator_plugin_archive| from the mutator plugin storage
+  bucket to the plugin archives directory. Returns the path archive was
+  downloaded to."""
+  file_path = os.path.join(_get_mutator_plugins_archives_dir(),
+                           mutator_plugin_archive)
+  url = '%s/%s' % (_get_mutator_plugins_bucket_url(), mutator_plugin_archive)
+  assert storage.copy_file_from(url, file_path)
```
inferno-chromium: We don't usually assert for errors like these, but instead log_error and continue.

src/python/tests/core/bot/fuzzers/mutator_plugin_test.py
```diff
@@ -0,0 +1,168 @@
...
+            self.plugin_archive_filename))
+
+
+class ExtractNameFromArchiveTest(unittest.TestCase):
+  """Tests for _extract_name_from_archive."""
+
+  def test_extract_name_from_archive(self):
+    """Tests that _extract_name_from_archive extracts the name from the
+    archive."""
+    name = 'myplugin'
```
inferno-chromium: nit: add expected_ prefix to these vars.

src/python/bot/fuzzers/mutator_plugin.py
```diff
@@ -0,0 +1,155 @@
...
+
+def _unpack_mutator_plugin(mutator_plugin_archive_path):
+  """Unpacks |mutator_plugin_archive_path| in the unpacked plugins directory and
+  returns the path it was unpacked into."""
+  mutator_plugin_name = os.path.basename(
+      os.path.splitext(mutator_plugin_archive_path)[0])
+  unpacked_plugin_dir = os.path.join(_get_mutator_plugins_unpacked_dir(),
+                                     mutator_plugin_name)
+  archive.unpack(mutator_plugin_archive_path, unpacked_plugin_dir)
+  return unpacked_plugin_dir
```
inferno-chromium: Should we verify that unpack succeeded and return None on error ?

src/python/bot/fuzzers/mutator_plugin.py
```diff
@@ -0,0 +1,155 @@
...
+
+    # Quit if no usable plugins are available.
+    if not usable_mutator_plugins:
+      return None
+
+    plugin_archive_name = utils.random_element_from_list(usable_mutator_plugins)
+    plugin_archive_path = _download_mutator_plugin_archive(plugin_archive_name)
+    _unpack_mutator_plugin(plugin_archive_path)
+    mutator_plugin_path = find_mutator_plugin()
+    assert mutator_plugin_path is not None
```
inferno-chromium: We should gracefully exit with log_error, and return bool result rather than assert.

src/python/bot/fuzzers/libfuzzer.py
jonathanmetzman: >looks like extra_env is not passed to LibFuzzerCommon.fuzz

I think this was fixed in a more recent upload.

>also need a test that shows LD_PRELOAD set to that path.

I'll see how easy this is but I already test for this implicitly in the integration test I added.

src/python/bot/fuzzers/libFuzzer/fuzzer.py
jonathanmetzman: Done.

src/python/bot/fuzzers/builtin.py
jonathanmetzman: Done.

src/python/bot/fuzzers/libfuzzer.py
jonathanmetzman: I added an extra_env variable to our `new_process` helper functions.

inferno-chromium:
This CL is getting huge, best to split into 2-3 parts for faster review and easier understanding.
1. just plugin mutator strategy, buckets
2. mutator strategy
3. minijail support
4. whatver, bug fixes, etc.

src/python/bot/fuzzers/libFuzzer/launcher.py
```diff
@@ -100,6 +101,8 @@
 
 FORK_PROBABILITY = 0.1
 
+MUTATOR_PLUGIN_PROBABILITY = .5
```
inferno-chromium: nit, please use 0.5 instead of .5

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -383,11 +393,17 @@ def fuzz(self,
...
     corpus_directories = self._get_chroot_corpus_paths(corpus_directories)
-    return LibFuzzerCommon.fuzz(self, corpus_directories, fuzz_timeout,
-                                artifact_prefix, additional_args)
+    return LibFuzzerCommon.fuzz(
+        self,
+        corpus_directories,
+        fuzz_timeout,
+        artifact_prefix,
+        additional_args,
+        extra_env=extra_env)
```
inferno-chromium: nit: can just pass extra_env, similar to other default args like arifact_prefix, additional_args.

src/python/bot/fuzzers/mutator_plugin.py
```diff
@@ -0,0 +1,167 @@
...
+  def __init__(self, fuzzer_binary_name):
+    self.fuzzer_binary_name = fuzzer_binary_name
+    self.job_name = environment.get_value('JOB_NAME')
+    self.create_directories()
+
+  def create_directories(self):
+    """Creates directories needed to use mutator plugins."""
+
+    def recreate_directory(directory_path):
+      shell.create_directory_if_needed(
```
inferno-chromium: This looks very hacky, to do create_directory_if_needed first and then remove_directory. Why do we need to remove and re-download plugin everytime ? Is this for every launcher instance ? We need to make sure we dont even unpack if it was unpacked once.

src/python/bot/fuzzers/mutator_plugin.py
```diff
@@ -0,0 +1,167 @@
...
+      os.path.splitext(mutator_plugin_archive_path)[0])
+  unpacked_plugin_dir = os.path.join(_get_mutator_plugins_unpacked_dir(),
+                                     mutator_plugin_name)
+  archive.unpack(mutator_plugin_archive_path, unpacked_plugin_dir)
+  return unpacked_plugin_dir
+
+
+def _extract_name_from_archive(plugin_archive_filename):
+  """Parses |plugin_archive_filename| which should be named using the schema:
+  '$NAME-$JOB-$FUZZ_TARGET.zip'. Returns tuple containing
```
inferno-chromium: Just fyi, all this will completely break if we have '-' in either custom mutator name, job name or fuzz target name. What is our plan for that ?

src/python/bot/fuzzers/libFuzzer/launcher.py
jonathanmetzman: Done

src/python/bot/fuzzers/libfuzzer.py
jonathanmetzman: Done

src/python/bot/fuzzers/mutator_plugin.py
jonathanmetzman: Actually I think it only breaks if `$NAME` has `-` in it.
I think this is an acceptable issue since this naming schema is only temporary and `$NAME` is arbitrary so it isn't actually limiting use of mutator plugins on particular targets or jobs.

src/python/bot/fuzzers/mutator_plugin.py
jonathanmetzman: > This looks very hacky, to do create_directory_if_needed first and then remove_directory. Why do we need to remove and re-download plugin everytime ? 

What if the plugin gets updated?

>Is this for every launcher instance ? 

Yes.

>We need to make sure we dont even unpack if it was unpacked once.

I intentionally do this. I agree it would be less wasteful to unpack once for every 5 runs of launcher.py. I believe the code did this initially, but it was suggested that I change this so that mutator plugins would more easily work on OSS-Fuzz.

PR #287, opened 2019-03-21
 Log the error earlier for archive unpack.

inferno-chromium:


inferno-chromium:
TBR minor logging fix.
Currently gets logged at later point

Failed to extract everything from archive v8_builtins_generator.tgz. Traceback (most recent call last): File "c:\clusterfuzz\src\python\system\archive.py", line 326, in unpack tar_archive.extract(filename, output_directory) File "c:\python27\lib\tarfile.py", line 2109, in extract tarinfo = self.getmember(member) File "c:\python27\lib\tarfile.py", line 1829, in getmember raise KeyError("filename %r not found" % name) KeyError: "filename 'generate_builtins_tests.py' not found"

PR #288, opened 2019-03-21
 Don't filter filename at archive iteration time

inferno-chromium:
since we pass file list to extract function, need to have exact filename as-is with './'. instead filter './' at traversal check time.


inferno-chromium:
TBR simple fix to avoid exceptions.

PR #289, opened 2019-03-21
Minor fuzzer weight adjustments based on recent data.

mbarbella-chromium:
The intention of this change is to have fewer cases where weights
fluctuate wildly from day to day for fuzzers near a threshold. New
thresholds are meant to be cutoffs where something is very obviously
wrong in nearly all cases. It also reduces the penalty slightly for the
coverage growth rule, as a very large number of fuzzer/job pairs match
it.

Of ~3500 internal fuzzer/job pairs with data, these changes would cause
us to make the following adjustments for pairs matched:
startup_crash_count: 101 -> 101 (no change, this one is nearly binary)
slow_unit_count: 55 -> 14
timeout_count: 15 -> 5
crash_count: 83 -> 49
oom_count: 96 -> 76

### DIFF
--- a/src/appengine/handlers/cron/fuzzer_weights.py
+++ b/src/appengine/handlers/cron/fuzzer_weights.py
@@ -198,7 +198,7 @@ WHERE
 """
 
 COVERAGE_UNCHANGED_SPECIFICATION = QuerySpecification(
-    adjusted_weight=0.65,
+    adjusted_weight=0.70,
     threshold=1.0,
     query_format=COVERAGE_UNCHANGED_FORMAT,
     formatter=_coverage_formatter,
--- a/src/appengine/handlers/cron/fuzzer_weights.py
+++ b/src/appengine/handlers/cron/fuzzer_weights.py
@@ -81,7 +81,7 @@ GROUP BY
 # we'll find anything during fuzzing.
 STARTUP_CRASH_SPECIFICATION = QuerySpecification(
     adjusted_weight=0.10,
-    threshold=0.80,
+    threshold=0.90,
     query_format=GENERIC_QUERY_FORMAT.format(field_name='startup_crash_count'),
     formatter=_past_day_formatter,
     reason='frequent startup crashes')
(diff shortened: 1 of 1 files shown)


mbarbella-chromium:
PTAL. See first comment for more info.

src/appengine/handlers/cron/fuzzer_weights.py
```diff
@@ -198,7 +198,7 @@ def _coverage_formatter(query_format, dataset):
 """
 
 COVERAGE_UNCHANGED_SPECIFICATION = QuerySpecification(
-    adjusted_weight=0.5,
+    adjusted_weight=0.65,
```
inferno-chromium: Can you explain this more.

src/appengine/handlers/cron/fuzzer_weights.py
```diff
@@ -100,17 +100,17 @@ def _coverage_formatter(query_format, dataset):
...
-# Fuzzers which are crashing frequently may not be making full use of their
-# allotted time for fuzzing, and may end up being more effective once the known
-# issues are fixed.
+# Fuzzers which are crashing on almost every run are not making full use of
+# their allotted time for fuzzing, and may end up being more effective once the
+# known issues are fixed.
 CRASH_SPECIFICATION = QuerySpecification(
     adjusted_weight=0.50,
-    threshold=0.90,
+    threshold=0.99,
```
inferno-chromium: Why not 0.95 for everything - STARTUP_CRASH_SPECIFICATION, SLOW_UNIT_SPECIFICATION, TIMEOUT_SPECIFICATION, OOM_SPECIFICATION, CRASH_SPECIFICATION

mbarbella-chromium: I'm not opposed to that but I don't think it would do as good a job of addressing the problem this change was meant to solve. I was hoping to address the cases we discussed yesterday where weights were being sporadically adjusted from day to day for pairings on the border of one of the thresholds. This .99 looks a bit odd in particular but this one in particular has a lot of cases that really do crash every time, and this does a better job of filtering out some of the ones that fall close to the border (this has quite a few > .9, where most of the rest tend not to).

src/appengine/handlers/cron/fuzzer_weights.py
mbarbella-chromium: This seemed less severe than some of the other issues, especially now that the thresholds have been raised a bit. I was surprised by how many pairings were actually matching this rule, and how heavily it was skewing the weights down.

mbarbella-chromium:
Discussed offline. We plan to adjust the weights dynamically instead of relying on specific thresholds, which should actually solve some of the issues with fuzzers with flaky stats right around the cutoffs.

PR #291, opened 2019-03-21
Cleanup directory traversal detection code in archives.

inferno-chromium:


inferno-chromium:
Note that `output_directory` is already clean, since it is trusted and is set under os.path.realpath above. so just need to take care of `filename` coming from archive.

  # If the output directory is a symlink, get its actual path since we will be
  # doing directory traversal checks later when unpacking the archive.
  output_directory = os.path.realpath(output_directory)

src/python/system/archive.py
```diff
@@ -190,19 +190,22 @@ def is_archive(filename):
   return get_archive_type(filename) != ArchiveType.UNKNOWN
 
 
+def _normalize_filename(filename):
+  """Normalize file name from archive for directory traversal detection."""
+  if filename.startswith('./'):
+    filename = filename[2:]
+  if os.path.altsep:
+    filename = filename.replace(os.path.altsep, os.path.sep)
```
oliverchang: can we just use os.path.normpath ?

```
os.path.normpath(path)
Normalize a pathname by collapsing redundant separators and up-level references so that A//B, A/B/, A/./B and A/foo/../B all become A/B. This string manipulation may change the meaning of a path that contains symbolic links. On Windows, it converts forward slashes to backward slashes. To normalize case, use normcase().
```

inferno-chromium: We cannot use normpath as it would normalize output_dir/A/../../B to B and we do a realpath later on this (using os.path.realpath(os.path.join(output_dir, normalized_filename))) and compare. Basically, we only normalize things that wont hamper directory traversal detection.

inferno-chromium: Done, thanks for discussing offline, i forgot this was only on filename (which later gets joined to putput_directory) so normpath can be done.

PR #292, opened 2019-03-21
Add an Android startup script for open source users.

mbarbella-chromium:
PTAL. I'm still seeing some issues related to the test project credentials that I'd like to address before landing, but this seems to at least run the bot correctly. I don't believe it does anything differently for auth from the Mac script, so it's probably fine to review it now.

### DIFF
--- a/configs/test/bot/setup/android.bash
+++ b/configs/test/bot/setup/android.bash
@@ -70,11 +70,17 @@ if [ ! -d "$INSTALL_DIRECTORY/$APPENGINE" ]; then
   rm $APPENGINE_FILE
 fi
 
-echo "Ensure device is connected."
-if ! $ADB_PATH/adb devices -l | grep -q "$ANDROID_SERIAL"; then
+echo "Installing ClusterFuzz package dependencies."
+pip install crcmod==1.7 psutil==5.4.7 pyOpenSSL==19.0.0
+
+echo "Ensuring device is connected."
+if ! $ADB_PATH/adb -s $ANDROID_SERIAL get-state | grep -q "device"; then
   echo "Device $ANDROID_SERIAL is not connected."
   exit 1
 fi
+
+echo "Activating credentials with the Google Cloud SDK."
+$GSUTIL_PATH/gcloud auth activate-service-account --key-file=$GOOGLE_APPLICATION_CREDENTIALS
 
 # Otherwise, gsutil will error out due to multiple types of configured
 # credentials. For more information about this, see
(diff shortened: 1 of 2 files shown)


configs/test/bot/setup/android.bash
```diff
@@ -0,0 +1,97 @@
...
+
+echo "Fetching Google App Engine SDK."
+if [ ! -d "$INSTALL_DIRECTORY/$APPENGINE" ]; then
+  curl -O "https://commondatastorage.googleapis.com/clusterfuzz-data/$APPENGINE_FILE"
+  unzip -q $APPENGINE_FILE
+  rm $APPENGINE_FILE
+fi
+
+echo "Ensure device is connected."
+if ! $ADB_PATH/adb devices -l | grep -q "$ANDROID_SERIAL"; then
```
inferno-chromium: I think we should check 
$ADB_PATH/adb -s $ANDROID_SERIAL get-state 
should be 'device'

configs/test/bot/setup/android.bash
```diff
@@ -0,0 +1,97 @@
...
+  echo "Device $ANDROID_SERIAL is not connected."
+  exit 1
+fi
+
+# Otherwise, gsutil will error out due to multiple types of configured
+# credentials. For more information about this, see
+# https://cloud.google.com/storage/docs/gsutil/commands/config#configuration-file-selection-procedure
+echo "Specifying the proper Boto configuration file."
+BOTO_CONFIG_PATH=$($GSUTIL_PATH/gsutil -D 2>&1 | grep "config_file_list" | egrep -o "/[^']+gserviceaccount\.com/\.boto") || true
```
inferno-chromium: Why || true ?

configs/test/bot/setup/android.bash
```diff
@@ -0,0 +1,97 @@
...
+  unzip -q $APPENGINE_FILE
+  rm $APPENGINE_FILE
+fi
+
+echo "Ensure device is connected."
+if ! $ADB_PATH/adb devices -l | grep -q "$ANDROID_SERIAL"; then
+  echo "Device $ANDROID_SERIAL is not connected."
+  exit 1
+fi
+
```
inferno-chromium: Missing step.

echo "Activating credentials with the Google Cloud SDK."
$GSUTIL_PATH/gcloud auth activate-service-account --key-file=$GOOGLE_APPLICATION_CREDENTIALS

configs/test/bot/setup/android.bash
```diff
@@ -0,0 +1,97 @@
...
+echo "Specifying the proper Boto configuration file."
+BOTO_CONFIG_PATH=$($GSUTIL_PATH/gsutil -D 2>&1 | grep "config_file_list" | egrep -o "/[^']+gserviceaccount\.com/\.boto") || true
+if [ -f $BOTO_CONFIG_PATH ]; then
+  export BOTO_CONFIG="$BOTO_CONFIG_PATH"
+else
+  echo "WARNING: failed to identify the Boto configuration file and specify BOTO_CONFIG env."
+fi
+
+echo "Downloading ClusterFuzz source code."
+$GSUTIL_PATH/gsutil cp gs://$DEPLOYMENT_BUCKET/linux.zip clusterfuzz-source.zip
```
inferno-chromium: `rm -rf clusterfuzz` to remove old artifacts.

configs/test/bot/setup/android.bash
```diff
@@ -0,0 +1,97 @@
...
+unzip -f -q clusterfuzz-source.zip
+
+echo "Running ClusterFuzz."
+OS_OVERRIDE="ANDROID" PATH="$PATH" NFS_ROOT="$NFS_ROOT" GOOGLE_APPLICATION_CREDENTIALS="$GOOGLE_APPLICATION_CREDENTIALS" ROOT_DIR="$ROOT_DIR" PYTHONPATH="$PYTHONPATH" GSUTIL_PATH="$GSUTIL_PATH" python $ROOT_DIR/src/python/bot/startup/run.py &
```
inferno-chromium: Needs to pass ANDROID_SERIAL as well.

mbarbella-chromium: We assume this is already exported. It should be fine to include it since that matches some of the other invocations, though.

configs/test/bot/setup/android.bash
```diff
@@ -0,0 +1,104 @@
...
+  curl -O "https://commondatastorage.googleapis.com/clusterfuzz-data/$APPENGINE_FILE"
+  unzip -q $APPENGINE_FILE
+  rm $APPENGINE_FILE
+fi
+
+echo "Installing ClusterFuzz package dependencies."
+pip install crcmod==1.7 psutil==5.4.7 pyOpenSSL==19.0.0
+
+echo "Ensuring device is connected."
+if ! $ADB_PATH/adb -s $ANDROID_SERIAL get-state | grep -q "device"; then
```
inferno-chromium: Most places i have seen do an equality check for device, so would prefer that.

if [[ $("$ADB_PATH/adb" -s "$ANDROID_SERIAL" get-state) == "device" ]]; then

mbarbella-chromium: Done

PR #293, opened 2019-03-22
Use nodesource repo for latest nodejs

inferno-chromium:


inferno-chromium:
TBR minor fix. Otherwise this causes very old nodejs to be installed (which has only nodejs binary and not node binary)

PR #300, opened 2019-03-22
Fix license header year in non-python files.

inferno-chromium:


inferno-chromium:
Minor fix, can skip review.

PR #301, opened 2019-03-22
Scale fuzzer weights linearly instead of relying on thresholds.

mbarbella-chromium:
PTAL when you have time.

### DIFF
--- a/src/appengine/handlers/cron/fuzzer_weights.py
+++ b/src/appengine/handlers/cron/fuzzer_weights.py
@@ -104,15 +104,6 @@ TIMEOUT_SPECIFICATION = QuerySpecification(
     formatter=_past_day_formatter,
     reason='frequent timeouts')
 
-# Fuzzers which are crashing frequently may not be making full use of their
-# allotted time for fuzzing, and may end up being more effective once the known
-# issues are fixed.
-CRASH_SPECIFICATION = QuerySpecification(
-    query_format=GENERIC_QUERY_FORMAT.format(
-        field_name='crash_count', min_weight=0.70),
-    formatter=_past_day_formatter,
-    reason='frequent crashes')
-
 # Fuzzers with extremely frequent OOMs may contain leaks or other issues that
 # signal that they need some improvement. Run with a slightly reduced weight
 # until the issues are fixed.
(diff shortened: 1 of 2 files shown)


inferno-chromium:
will let @oliverchang take a first review pass on this (also i wont be able to do this until monday).

mbarbella-chromium:
No rush, I wasn't planning on landing it before Monday anyway.

src/appengine/handlers/cron/fuzzer_weights.py
```diff
@@ -221,37 +214,31 @@ def _coverage_formatter(query_format, dataset):
...
-  old_weight = old_match.adjusted_weight
+  new_weight = match.new_weight
+  old_weight = old_match.new_weight
 
   # Always update the weight if the previous value is the default. This is
   # required to deal with specifications that are meant to set the weight above
   # 1.0. Otherwise, prioritize only the most penalizing match for this pairing.
-  if old_match == RESTORE_DEFAULT_SPECIFICATION or new_weight < old_weight:
-    matched_specifications[key] = specification
+  if old_match == RESTORE_DEFAULT_MATCH or new_weight < old_weight:
```
inferno-chromium: How does this work for new fuzzers where new weight is 5.0 and is > default.

src/appengine/handlers/cron/fuzzer_weights.py
```diff
@@ -80,48 +82,43 @@ def _coverage_formatter(query_format, dataset):
...
-    query_format=GENERIC_QUERY_FORMAT.format(field_name='crash_count'),
+    query_format=GENERIC_QUERY_FORMAT.format(
+        field_name='crash_count', min_weight=0.70),
     formatter=_past_day_formatter,
     reason='frequent crashes')
 
 # Fuzzers with extremely frequent OOMs may contain leaks or other issues that
 # signal that they need some improvement. Run with a slightly reduced weight
 # until the issues are fixed.
 OOM_SPECIFICATION = QuerySpecification(
```
inferno-chromium: nit: move OOM_SPECIFICATION block before CRASH_SPECIFICATION to show min_weight ascending order.

src/appengine/handlers/cron/fuzzer_weights.py
mbarbella-chromium: If we don't have an entry yet we'll match the first part of this condition thanks to the default on 229 above (and we have another case where we might have assigned RESTORE_DEFAULT_MATCH on 289). That said, there are probably significantly more cases where we'll override it now. The crash rule in particular needs some kind of minimum threshold to avoid this.

mbarbella-chromium: Added minimums.

mbarbella-chromium:
Went back to something closer to the initial commit based on offline discussion with a different fix for the issues with weight increasing rules.

PR #302, opened 2019-03-23
Unset CHILD_PROCESS_TERMINATION_PATTERN before reproducibility call

inferno-chromium:


src/python/system/process_handler.py
```diff
@@ -297,6 +297,9 @@ def run_process(cmdline,
       # There is no special termination behavior. Simply stop the application.
       android.adb.stop_application()
 
+    # Clear log for a next clean testcase execution.
+    android.logger.clear_log()
```
inferno-chromium: Unbelievable how things worked uptil now without this.

inferno-chromium: My bad, there is a clear_log call in start of run_process.

inferno-chromium:
TBR simple fix, just removes CHILD_PROCESS_TERMINATION_PATTERN before test_for_reproducibility is run.

PR #304, opened 2019-03-23
Don't reimage gce bot for hard_reset, only reboot.

inferno-chromium:
This was an incorrect change done for gce. hard_reset can be called for minor reasons like ooms, so it is ok to just reboot the bot, and not needs reimage.


inferno-chromium:
TBR.

PR #308, opened 2019-03-25
Initial support for Fuchsia on ClusterFuzz.

flowerhack:
Feedback/changes very welcome.

This CL teaches ClusterFuzz to spin up a Fuchsia QEMU VM and communicate
with it in a basic way.

Right now, all the "fuzzing" command does is run "ls" on the target
system, as a sort of "hello world"; a follow-up CL will teach it to
select & use proper fuzzing commands.

We chose this approach because Fuchsia fully supports libfuzzer
on-target; running, e.g., "fuzz [fuzzer_name]" instead of "ls" will
simply fuzz using the named fuzzer, and return when it encounters some
result.  See https://fuchsia.googlesource.com/fuchsia/+/refs/heads/master/docs/development/workflows/libfuzzer.md

Follow-on CLs will include: using "fuzz" instead of "ls", syncing
corpora, perhaps utilizing GCE support in lieu of QEMU, etc

To run this locally, you must:
* Use the admin panel to set up a Fuchsia fuzzing "job".
-> The job must have Platform = FUCHSIA
-> The job must use the libfuzzer and engine_asan templates
-> You must include a libfuzzer_asan_test_fuzzer.zip file in "Custom
Build"
--> (I noticed that jobs tended to hang with 'Upto #' statements if a
custom build was not included; would love to find some workaround for
this, since the Fuchsia fuzzing job doesn't really *need* a custom
build.)
-> Finally, set `FUCHSIA-RESOURCES_PATH` to a proper GCS bucket path.
You can use mine for testing if you like:
gs://fuchsia_on_clusterfuzz_resources_v1/*
* Go to "Fuzzers" in the admin panel and add your job to the libFuzzer
fuzzer.
* Run `QUEUE_OVERRIDE=FUCHSIA OS_OVERRIDE=FUCHSIA python butler.py
run_bot --name someplace ~/someplace`.

For debugging, it's helpful to tail -f
~/someplace/clusterfuzz/bot/logs*.


flowerhack:
@oliverchang if you wouldn't mind reviewing? qemu actually works now :)

cc @mvanotti

flowerhack:
@nopsledder

inferno-chromium:
@oliverchang is OOO this week. @Dor1s @jonathanmetzman - can you help to review.

@flowerhack - can you fix test / lint failures and reupload. you can run "python butler.py format" to fix most of those. verify using "python butler.py lint"

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -15,15 +15,21 @@
 
 import copy
 import os
+import errno
 import shutil
 
+import time
+import subprocess
+
```
jonathanmetzman: [Imports need to follow the style guide](http://google.github.io/styleguide/pyguide.html#22-imports)
(ie: remove line 20 and sort imports).

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -318,6 +324,79 @@ def fuzz(self,
...
+  """libFuzzer runner (when Fuchsia is the target platform)."""
+  def __init__(self, executable_path, default_args=None):
+    # This code assumes the following layout for files in the storage bucket:
+    # * multiboot.bin
+    # * fvm.blk
+    # * fuchsia-ssh.zbi
+    # * qemu/*
+    # * .ssh/*
+    local_path = os.getcwd() + "/"
+    resources_path = local_path + "fuchsia_on_clusterfuzz_resources_v1"
```
jonathanmetzman: Replace these two lines with ` resources_path = os.path.join(os.getcwd(), "fuchsia_on_clusterfuzz_resources_v1"`. `os.path.join` is a cross platform way to join directories to files/sub directories.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -318,6 +324,79 @@ def fuzz(self,
...
+
+    try:
+      os.mkdir(resources_path)
+    except OSError as e:
+      if e.errno == errno.EEXIST:
+        pass
+    subbed_fuchsia_gsutil_command = [param.replace("{fuchsia_resources_path}", environment.get_value('FUCHSIA_RESOURCES_PATH', ''))
+                                     .replace("{local_resources_path}", resources_path) for param in constants.FUCHSIA_GSUTIL_COMMAND]
+    subprocess.call(subbed_fuchsia_gsutil_command)
+    qemu_path = resources_path + "/qemu-for-fuchsia/bin/qemu-system-x86_64"
```
jonathanmetzman: Please use `os.path.join` (ie: `os.path.join(resources_path, 'qemu-for-fuchsia', 'bin', 'qemu-system-x86_64'`)

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -318,6 +324,79 @@ def fuzz(self,
...
+    os.chmod(qemu_path, 0o550)
+    kernel_path = resources_path + "/multiboot.bin"
```
jonathanmetzman: Please use `os.path.join`

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -318,6 +324,79 @@ def fuzz(self,
...
+    os.chmod(kernel_path, 0o644)
+    pkey_path = resources_path + "/.ssh/pkey"
```
jonathanmetzman: Please use `os.path.join`

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -318,6 +324,79 @@ def fuzz(self,
...
+    os.chmod(pkey_path, 0o400)
+    sharefiles_path = resources_path + "/qemu-for-fuchsia/share/qemu"
```
jonathanmetzman: Please use `os.path.join`

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -318,6 +324,79 @@ def fuzz(self,
...
+    initrd_path = resources_path + "/fuchsia-ssh.zbi"
```
jonathanmetzman: Please use `os.path.join`

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -318,6 +324,79 @@ def fuzz(self,
...
+    os.chmod(initrd_path, 0o644)
+    drive_path = resources_path + "/fuchsia.qcow2"
```
jonathanmetzman: Please use `os.path.join`

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -318,6 +324,79 @@ def fuzz(self,
...
+    # * fuchsia-ssh.zbi
+    # * qemu/*
+    # * .ssh/*
+    local_path = os.getcwd() + "/"
+    resources_path = local_path + "fuchsia_on_clusterfuzz_resources_v1"
+
+    try:
+      os.mkdir(resources_path)
+    except OSError as e:
+      if e.errno == errno.EEXIST:
```
jonathanmetzman: Why check for this if the body is `pass`?

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -318,6 +324,79 @@ def fuzz(self,
...
+        pass
+    subbed_fuchsia_gsutil_command = [param.replace("{fuchsia_resources_path}", environment.get_value('FUCHSIA_RESOURCES_PATH', ''))
```
jonathanmetzman: Please use `copy_file_to` from `src/python/google_cloud_utils/storage.py` instead of calling gsutil directly.

src/python/bot/fuzzers/libFuzzer/constants.py
```diff
@@ -66,3 +66,41 @@
...
+	'-cpu',
+	'host,migratable=no',
+	'-netdev',
+	'user,id=net0,net=192.168.3.0/24,dhcpstart=192.168.3.9,host=192.168.3.2,hostfwd=tcp::{portnum}-:22',
+	'-device',
+	'e1000,netdev=net0,mac=52:54:00:63:5e:7b',
+	'-L',
+	'{sharefiles}']
+
+FUCHSIA_SSH_COMMAND_TEMPLATE = ["ssh", "-i", "{identity_file}", "-o", "StrictHostKeyChecking no", "localhost", "-p", "{portnum}"]
```
jonathanmetzman: In case our linters don't complain, we use single quotes instead of double for strings.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -318,6 +324,79 @@ def fuzz(self,
...
+    # This code assumes the following layout for files in the storage bucket:
+    # * multiboot.bin
+    # * fvm.blk
+    # * fuchsia-ssh.zbi
+    # * qemu/*
+    # * .ssh/*
+    local_path = os.getcwd() + "/"
+    resources_path = local_path + "fuchsia_on_clusterfuzz_resources_v1"
+
+    try:
```
jonathanmetzman: You should probably use `create_directory_if_needed` (from `src/python/system/shell.py`) for this.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -318,6 +324,79 @@ def fuzz(self,
...
+    sharefiles_path = resources_path + "/qemu-for-fuchsia/share/qemu"
+    initrd_path = resources_path + "/fuchsia-ssh.zbi"
+    os.chmod(initrd_path, 0o644)
+    drive_path = resources_path + "/fuchsia.qcow2"
+    os.chmod(drive_path, 0o644)
+
+    # TODO: Add a mechanism for choosing portnum dynamically.
+    portnum = "56339"
+
+    self.subbed_qemu_base_command = [param.replace("{qemu}", qemu_path)
```
jonathanmetzman: I have a suggested improvement for this pattern you use.
Make `FUCHSIA_QEMU_COMMAND_TEMPLATE` one big string and then call `FUCHSIA_QEMU_COMMAND_TEMPLATE.format(kernel=kernel_path, drive=drive_path, ...)`

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -318,6 +324,79 @@ def fuzz(self,
...
+                                .replace("{kernel}", kernel_path)
+                                .replace("{drive}", drive_path)
+                                .replace("{initrd}", initrd_path)
+                                .replace("{sharefiles}", sharefiles_path)
+                                .replace("{portnum}", portnum) for param in constants.FUCHSIA_QEMU_COMMAND_TEMPLATE]
+    self.subbed_ssh_base_command = [param.replace("{identity_file}", pkey_path)
```
jonathanmetzman: See my comment on line 362 about using `.format` instead of this technique.
If you need a list you can call `s.split(' ')` to separate the string into a list by each space.

src/python/bot/fuzzers/libfuzzer.py
jonathanmetzman: Downloading things in `libfuzzer.py` is a bit weird, I will need to think about where else this could go.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -318,6 +324,79 @@ def fuzz(self,
...
+    return command
+
+  def fuzz(self,
+           corpus_directories,
+           fuzz_timeout,
+           artifact_prefix=None,
+           additional_args=None):
+    """LibFuzzerCommon.fuzz override."""
+
+    subprocess.Popen(self.subbed_qemu_base_command)
```
jonathanmetzman: I think you should save the result of this and then after `LibFuzzerCommon.fuzz` returns handle errors in the process. Maybe fine without this for now as the TODO sort of addresses this.

src/python/bot/fuzzers/libfuzzer.py
jonathanmetzman: Why return an empty string from get_value if `FUCHSIA_RESOURCES_PATH` doesn't exist? I think this will cause the exception to manifest itself further from the root cause. You should probably log_error if `FUCHSIA_RESOURCES_PATH` doesn't exist instead.

src/python/bot/fuzzers/libFuzzer/constants.py
```diff
@@ -66,3 +66,41 @@
 
 # libFuzzer's exit code if a bug was found in the target code.
 TARGET_ERROR_EXITCODE = 77
+
+FUCHSIA_QEMU_COMMAND_TEMPLATE = ['{qemu}',
```
jonathanmetzman: I don't think these command templates should go in constants.py, I think they should go in libfuzzer.py (part of the FuchsiaQemuLibFuzzerRunner class IMO). I personally don't like the idea of a file called constants.py since it is confusing (it seems like a bad idea to put all of one's constants in one file, all functions in another etc.).
However, this file is only supposed to have constants that are meaningful to libFuzzer. These commands are not that IMO.

jonathanmetzman:
Really cool. Thanks!
I left some comments, mostly on the low level code details :-(
I'll take another look and try to run this after they are addressed.

src/python/bot/fuzzers/libfuzzer.py
flowerhack: I wound up using gsutil directly because I couldn't find anything in those APIs for the concept of "please download this *entire* folder"--there's quite a lot in the Fuchsia SDK, and it'd be odd to have to enumerate each file.  Is there a "download this whole subdirectory" command I didn't see?  Or, is there a preference for e.g. zipping those files up and unzipping them once they're downloaded to Clusterfuzz?

src/python/bot/fuzzers/libfuzzer.py
flowerhack: unfortunately, the subprocess API wants to receive commands in the form of a list, rather than one big string.

i could *try* to make it one big string, and call .split(" ") after, but that's harder than it looks-- in the SSH command, "StrictHostKeyChecking no" needs to be one clause (space included), and in the QEMU command "kernel.serial=legacy TERM=dumb" needs to be all one clause (space included).  I could add "dummy" characters for the command to be split along but I worry that's even more confusing.  What do you think?

src/python/bot/fuzzers/libFuzzer/constants.py
flowerhack: moved!

src/python/bot/fuzzers/libfuzzer.py
flowerhack: qemu needs to be left running in the background, so that subsequent commands can ssh in and do interesting things, so i can't really get a return code from this process unfortunately :(

(unless what you're referring to is the ssh call, in which case, it's now retrying until success)

src/python/bot/fuzzers/libfuzzer.py
flowerhack: yeah, this is a more high-level conversation i was interested in having with y'all :)

in the initial concept, i think we had the idea of downloading + launching QEMU being handled by launcher.py, but when i got into the details of that, my two main questions were:

* how should IPC be handled?  if i launch QEMU in launcher.py and run SSH in libfuzzer.py, then i need a way to communicate "use this portnumber and this directory location" to libfuzzer.py, and neither environment variables nor "hardcode those in as global constants" felt entirely right to me, and

* i think i must be confused about some of the functionality of launcher.py itself, because though I see "App launch command is [path_to_launcher.py] ." in my log messages, adding log lines to launcher.py:main() seemed to have no effect, which means either I was goofing where I was looking for logs, OR, launcher.py isn't being called the way i expect

at that point I realized it may be more efficient to throw something up, and then see where you all think the best place for such matters is :)

src/python/bot/fuzzers/libfuzzer.py
flowerhack: see discussion on L362 comments

src/python/bot/fuzzers/libfuzzer.py
flowerhack: good point. added a log_error for that!

src/python/bot/fuzzers/libfuzzer.py
flowerhack: the idea was "if this directory's already been created, we must've downloaded resources there already"--which is handily resolved by the create_directory_if_needed call you suggested, so this is out now :)

src/python/bot/fuzzers/libFuzzer/launcher.py
```diff
@@ -949,6 +949,7 @@ def main(argv):
   parsed_stats.update(stat_overrides)
 
   # Dump stats data for further uploading to BigQuery.
+if environment.platform() != "FUCHSIA":
```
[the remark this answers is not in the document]
flowerhack: I think this is leftover from one of the days some of the gcloud dependencies were acting wonky--removing this line doesn't break anything, so let's remove it

flowerhack:
I think I've addressed everything so far.

(1) Not sure what y'all's preferred review workflow is, but for now I assume I'll add a new commit for each response, and squash them all in the end.

(2) Is there a way to re-trigger Travis?  It *seems* like the current info is stale (it's e.g. complaining about very short lines being over the line length limit)

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -319,6 +325,141 @@ def fuzz(self,
...
+    ]
+
+    self.fuchsia_bucket_name = 'fuchsia_on_clusterfuzz_resources_v1'
+
+    self.fuchsia_gsutil_command_template = [
+        'gsutil', 'cp', '-r', '{fuchsia_resources_path}',
+        '{local_resources_path}'
+    ]
+
+    # This initialization routime assumes that the GCS bucket contains the standard Fuchsia SDK, as well as:
```
Dor1s: nit: typo in "routime" -> "routine"

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -319,6 +325,141 @@ def fuzz(self,
...
+    # * /qemu-for-fuchsia/*
+    # * /.ssh/*
+    resources_path = os.path.join(os.getcwd(), self.fuchsia_bucket_name)
+
+    shell.create_directory_if_needed(resources_path)
+
+    if 'FUCHSIA_RESOURCES_PATH' not in os.environ:
+      logs.log_error("Could not find path for remote Fuchsia resources bucket (FUCHSIA_RESOURCES_PATH)!")
```
Dor1s: Should we bail out here?

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -319,6 +325,141 @@ def fuzz(self,
...
+
+    subbed_fuchsia_gsutil_command = [
```
Dor1s: `fuchsia_gsutil_command_template` doesn't seem to be used anywhere else. Can we get rid of it and just build the command here like the following:

```python
    subbed_fuchsia_gsutil_command = [
        'gsutil', 'cp', '-r', environment.get_value('FUCHSIA_RESOURCES_PATH', '')),
        resources_path
    ]
```

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -319,6 +325,141 @@ def fuzz(self,
...
+        param.replace("{fuchsia_resources_path}",
+                      environment.get_value('FUCHSIA_RESOURCES_PATH', ''))
+        .replace("{local_resources_path}", resources_path)
+        for param in self.fuchsia_gsutil_command_template
+    ]
+    subprocess.call(subbed_fuchsia_gsutil_command)
```
Dor1s: `subprocess.check_call` would be better, to make sure we fail in case of an error and report an exception to StackDriver?

Dor1s: Actually, I think the preferred way is to use `process = new_process.ProcessRunner(executable_path, default_args=None)` followed by `result = process.run_and_wait(additional_args=None)` and checking the `result.return_code` afterwards

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -319,6 +325,141 @@ def fuzz(self,
...
+    pkey_path = os.path.join(resources_path, '.ssh', 'pkey')
+    os.chmod(pkey_path, 0o400)
+    sharefiles_path = os.path.join(resources_path, 'qemu-for-fuchsia', 'share',
+                                   'qemu')
+
+    # The FVM is minimally sized to begin with; extend it to make room for ephemeral packages etc.
+    drive_path = os.path.join(resources_path, "extended_fvm.blk")
+    raw_fvm_path = os.path.join(resources_path, "target", "x64", "fvm.blk")
+    copy_fvm = ["cp", raw_fvm_path, drive_path]
+    subprocess.call(copy_fvm)
```
Dor1s: Same here and for other `subprocess.call` calls

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -319,6 +325,141 @@ def fuzz(self,
...
+    fuchsia_zbi = os.path.join(resources_path, "target", "x64", "fuchsia.zbi")
+    initrd_path = os.path.join(resources_path, "fuchsia-ssh.zbi")
+    make_fuchsia_ssh_zbi_command = [
+        zbi_tool, "-o", initrd_path, fuchsia_zbi, "-e",
+        "data/ssh/authorized_keys=" + authorized_keys_path
+    ]
+    subprocess.call(make_fuchsia_ssh_zbi_command)
+    os.chmod(initrd_path, 0o644)
+
+    # TODO: Add a mechanism for choosing portnum dynamically.
```
Dor1s: nit: `TODO(flowerhack): `

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -319,6 +325,141 @@ def fuzz(self,
...
+    portnum = "56339"
+
+    self.subbed_qemu_base_command = [
+        param.replace("{qemu}", qemu_path).replace("{kernel}", kernel_path)
+        .replace("{drive}", drive_path).replace("{initrd}", initrd_path)
+        .replace("{sharefiles}", sharefiles_path).replace("{portnum}", portnum)
+        for param in self.fuchsia_qemu_command_template
```
Dor1s: Similar comment as the one above, can you just build the command here? IMO that would be more straightforward than current version

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -319,6 +325,141 @@ def fuzz(self,
...
+        param.replace("{identity_file}", pkey_path).replace(
+            "{portnum}", portnum) for param in self.fuchsia_ssh_command_template
+    ]
+
+    super(FuchsiaQemuLibFuzzerRunner, self).__init__(
+        executable_path=executable_path, default_args=default_args)
+
+  def get_command(self, additional_args=None):
+    command = self.subbed_ssh_base_command[:]
+    # TODO: Update this to dynamically pick a result from "fuzz list" and then run that fuzzer.
```
Dor1s: nit: `TODO(flowerhack): `

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -319,6 +325,141 @@ def fuzz(self,
...
+
+    # Start fuzzing.
+    return LibFuzzerCommon.fuzz(self, corpus_directories, fuzz_timeout,
+                                artifact_prefix, additional_args)
+
+  def run_single_testcase(self,
+                          testcase_path,
+                          timeout=None,
+                          additional_args=None):
+    pass
```
Dor1s: Could you please add a `TODO` here as well?

src/python/bot/fuzzers/libfuzzer.py
jonathanmetzman: Dummy chars definitely seems like the incorrect approach.
I like the latest approach.

src/python/bot/fuzzers/libfuzzer.py
jonathanmetzman: >it's now retrying until success)

Indefinitely?

I think I understand the workflow here, 
1. QEMU is started
2. We ssh into the VM in fuzz 
3. ?
But I wonder if the output from step 1 is useful (in case it fails). 
Also, I wonder if we should be shutting down the VM after fuzzing?

src/python/bot/fuzzers/libfuzzer.py
jonathanmetzman: >and neither environment variables nor "hardcode those in as global constants" felt entirely right to me, and

We do use environment variables for this sort of thing.

>" in my log messages, adding log lines to launcher.py:main() seemed to have no effect,

I think these logs show up in run_fuzzer.log rather than bot.log. launcher.py gets called in a new process.

>at that point I realized it may be more efficient to throw something up, and then see where you all think the best place for such matters is :)

Not a bad idea. I'm still thinking about whether this should land as is and be fixed later or fixed first.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -319,6 +321,145 @@ def fuzz(self,
...
+
+    if 'FUCHSIA_RESOURCES_PATH' not in os.environ:
+      logs.log_error(
+          ('Could not find path for remote Fuchsia resources bucket'
+           '(FUCHSIA_RESOURCES_PATH)!')
+      )
+
+    gsutil_command_arguments = [
+        'cp', '-r',
+        environment.get_value('FUCHSIA_RESOURCES_PATH', ''), resources_path
```
jonathanmetzman: Why call `environment.get_value('FUCHSIA_RESOURCES_PATH', '')` instead of `environment.get_value('FUCHSIA_RESOURCES_PATH')` since we know it will always be in environ.
I would do this btw:
```
fuchsia_resources_url = environment.get_value('FUCHSIA_RESOURCES_URL')
if fuchsia_resources_url is None:
  logs.log_error(
          ('Could not find path for remote Fuchsia resources bucket'
           '(FUCHSIA_RESOURCES_URL)!')
      )
      return  # Bail
gsutil_command_arguments = ['cp', '-r', gcs_resources_url, resources_path]
```
(I think "url" is more descriptive than "path")

jonathanmetzman:
Julia, it looks like there's still some linting issues.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -319,6 +321,143 @@ def fuzz(self,
                                 artifact_prefix, additional_args)
 
 
+class FuchsiaQemuLibFuzzerRunner(new_process.ProcessRunner, LibFuzzerCommon):
+  """libFuzzer runner (when Fuchsia is the target platform)."""
+
+  def __init__(self, executable_path, default_args=None):
```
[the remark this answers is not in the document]
jonathanmetzman: +1

src/python/bot/fuzzers/libfuzzer.py
flowerhack: done

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -319,6 +321,145 @@ def fuzz(self,
...
+        qemu_path, '-D', '/tmp/qemustderr', '-m', '2048', '-nographic',
+        '-kernel', kernel_path, '-initrd', initrd_path, '-smp', '4', '-drive',
+        'file= ' + drive_path + ',format=raw,if=none,id=blobstore', '-device',
+        'virtio-blk-pci,drive=blobstore', '-monitor', 'none', '-append',
+        'kernel.serial=legacy TERM=dumb', '-machine', 'q35', '-enable-kvm',
+        '-display', 'none', '-cpu', 'host,migratable=no', '-netdev',
+        ('user,id=net0,net=192.168.3.0/24,dhcpstart=192.168.3.9,'
+         'host=192.168.3.2,hostfwd=tcp::') + portnum + '-:22', '-device',
+        'e1000,netdev=net0,mac=52:54:00:63:5e:7b', '-L', sharefiles_path
+    ]
```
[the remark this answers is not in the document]
flowerhack: Hey @jonathanmetzman : is there a way to signal to the linter "please ignore these lines", or suppress a particular thing?

The linter is upset at the idea of formatting this more reasonably, but I really think we *do* want the clearer formatting here, rather than suggestions 1 &2.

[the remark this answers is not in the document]
flowerhack: (I've got it formatted the "readable" way for now, but the litner will whine :) )

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -319,6 +321,143 @@ def fuzz(self,
...
+    # * /.ssh/*
+    resources_path = os.path.join(os.getcwd(), self.fuchsia_bucket_name)
+
+    shell.create_directory_if_needed(resources_path)
+
+    fuchsia_resources_url = environment.get_value('FUCHSIA_RESOURCES_URL')
+    if fuchsia_resources_url is None:
+      logs.log_error(('Could not find path for remote Fuchsia resources bucket'
+                      '(FUCHSIA_RESOURCES_URL)!'))
+      return
```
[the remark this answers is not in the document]
flowerhack: I was avoiding explicit exceptions mostly out of the principle of "trying to mostly do what CF does in other instances", and it *seems* like CF's early-bailout-mechanism is log_error + return.

That being said!  @jonathanmetzman and I chatted a bit, and now we're asking @mbarbella-chromium :

What do you think is the right "bail out early" approach here?  If any of the following is true.
* the URL to the Fuchsia bucket isn't set
* we can't download Fuchsia resources
* we can't modify the Fuchsia image appropriately to make it usable

we can't successfully proceed to actual fuzzing.  Will log_error + return trigger the proper alerting "oh no something's gone badly", or would an exception work better?

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -319,6 +321,143 @@ def fuzz(self,
...
+    gsutil_command_arguments = [
+        'cp', '-r', fuchsia_resources_url, resources_path
+    ]
+    process = new_process.ProcessRunner('gsutil', gsutil_command_arguments)
+    result = process.run_and_wait()
+    if result.return_code or result.timed_out:
+      logs.log_error(
+          'Failed to download Fuchsia resources.', output=result.output)
+
```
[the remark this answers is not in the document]
flowerhack: see comment on L342

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -319,6 +321,143 @@ def fuzz(self,
...
+                                   'qemu')
+
+    # The FVM is minimally sized to begin with; extend it to make room for
+    # ephemeral packages etc.
+    drive_path = os.path.join(resources_path, 'extended_fvm.blk')
+    raw_fvm_path = os.path.join(resources_path, 'target', 'x64', 'fvm.blk')
+    process = new_process.ProcessRunner('cp', [raw_fvm_path, drive_path])
+    result = process.run_and_wait()
+    if result.return_code or result.timed_out:
+      logs.log_error('Failed to copy FVM.', output=result.output)
```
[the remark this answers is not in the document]
flowerhack: see comment on L342

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -319,6 +321,143 @@ def fuzz(self,
...
+    os.chmod(drive_path, 0o644)
+    fvm_tool_path = os.path.join(resources_path, 'tools', 'fvm')
+    os.chmod(fvm_tool_path, 0o500)
+    process = new_process.ProcessRunner(
+        fvm_tool_path, [drive_path, 'extend', '--length', '1G'])
+    result = process.run_and_wait()
+    if result.return_code or result.timed_out:
+      logs.log_error('Failed to extend FVM.', output=result.output)
```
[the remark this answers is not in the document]
flowerhack: see comment on L342

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -319,6 +321,143 @@ def fuzz(self,
...
+    os.chmod(zbi_tool, 0o500)
+    fuchsia_zbi = os.path.join(resources_path, 'target', 'x64', 'fuchsia.zbi')
+    initrd_path = os.path.join(resources_path, 'fuchsia-ssh.zbi')
+    process = new_process.ProcessRunner(zbi_tool, [
+        '-o', initrd_path, fuchsia_zbi, '-e',
+        'data/ssh/authorized_keys=' + authorized_keys_path
+    ])
+    result = process.run_and_wait()
+    if result.return_code or result.timed_out:
+      logs.log_error('Failed to add keys to Fuchsia ZBI.', output=result.output)
```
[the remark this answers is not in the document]
flowerhack: see comment on L342

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -319,6 +321,143 @@ def fuzz(self,
...
+        'kernel.serial=legacy TERM=dumb', '-machine', 'q35', '-enable-kvm',
+        '-display', 'none', '-cpu', 'host,migratable=no', '-netdev',
+        ('user,id=net0,net=192.168.3.0/24,dhcpstart=192.168.3.9,'
+         'host=192.168.3.2,hostfwd=tcp::') + portnum + '-:22', '-device',
+        'e1000,netdev=net0,mac=52:54:00:63:5e:7b', '-L', sharefiles_path
+    ]
+
+    self.ssh_command = [
+        'ssh', '-i', pkey_path, '-o', 'StrictHostKeyChecking no', 'localhost',
+        '-p', portnum
```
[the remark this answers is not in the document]
flowerhack: i like your idea! done

src/python/bot/fuzzers/libfuzzer.py
flowerhack: changed accordingly

src/python/bot/fuzzers/libfuzzer.py
flowerhack: add a bail

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -319,6 +321,143 @@ def fuzz(self,
...
+    pkey_path = os.path.join(resources_path, '.ssh', 'pkey')
+    os.chmod(pkey_path, 0o400)
+    sharefiles_path = os.path.join(resources_path, 'qemu-for-fuchsia', 'share',
+                                   'qemu')
+
+    # The FVM is minimally sized to begin with; extend it to make room for
+    # ephemeral packages etc.
+    drive_path = os.path.join(resources_path, 'extended_fvm.blk')
+    raw_fvm_path = os.path.join(resources_path, 'target', 'x64', 'fvm.blk')
+    process = new_process.ProcessRunner('cp', [raw_fvm_path, drive_path])
```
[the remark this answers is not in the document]
flowerhack: oh good point; this was a case of my hewing too closely to how someone else was consuming the SDK.  removed

src/python/bot/fuzzers/libfuzzer.py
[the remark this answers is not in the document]
flowerhack: very much agree. chatted with @jonathanmetzman to brainstorm the "best" place to put this functionality; we've moved it to fuzz_task.py for now but i'm open to other ideas / can pontificate on the justificaiton, etc

new version of init should *just* be getting environment vars and ssh arguments.

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1536,3 +1539,106 @@ def execute_task(fuzzer_name, job_type):
   del testcase_file_paths
   del testcases_metadata
   utils.python_gc()
+
+
+def fuchsia_qemu_setup():
```
flowerhack: i know this is a somewhat awkward place to put a platform-specific function, but platforms/fuchsia/device.py didn't seem quite right either; is there someplace where this would maek the most sense?

flowerhack:
i *think!* i've responded to all the comments i can for now.  looking forward to another round of feedback, and, once we're pretty sure of this overall design / approach at least for now, would be happy to add integration tests etc so long as i can talk to one of y'all for a bit of guidance!

src/python/bot/fuzzers/libfuzzer.py
flowerhack: changed it to current workflow where work has been moved to fuzz_task.py, see if that makes more sense

src/python/bot/fuzzers/libfuzzer.py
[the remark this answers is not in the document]
oliverchang: I think exception here makes sense. log_error is equally loud, but in this case it seems like continuing does not make sense.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -319,6 +321,61 @@ def fuzz(self,
...
+  """libFuzzer runner (when Fuchsia is the target platform)."""
+
+  def __init__(self, executable_path, default_args=None):
+    if (not environment.get_value('FUCHSIA_PKEY_PATH')) or (
+        not environment.get_value('FUCHSIA_PORTNUM')):
+      logs.log_error('FUCHSIA_PKEY_PATH and/or FUCHSIA_PORTNUM was not set')
+      return
+    self.ssh_root = [
+        'ssh', '-i',
+        str(environment.get_value('FUCHSIA_PKEY_PATH')), '-o',
```
oliverchang: why str() ? 

as above if we exception out in the case this isn't set we should be able to assume that this is a valid string.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -319,6 +321,61 @@ def fuzz(self,
...
+        'StrictHostKeyChecking no', 'localhost', '-o',
+        'UserKnownHostsFile=/dev/null', '-p',
+        str(environment.get_value('FUCHSIA_PORTNUM'))
```
oliverchang: same here.

flowerhack: python interprets a portnumber as an int, rather than a str, so it makes sense to coerce it here.  unless you see another way?

src/python/bot/fuzzers/libfuzzer.py
flowerhack: environment.get_value('FUCHSIA_PORTNUM') returned an integer, rather than a string, and python (reasonably) complained, so i wound up throwing str() in front of the other one too, for safety.

i can't imagine a way that pkey_path would get int'ified, though, so backing that one out.

src/python/bot/fuzzers/libfuzzer.py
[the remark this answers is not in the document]
flowerhack: wfm! switched these all to exceptions

src/python/bot/fuzzers/libfuzzer.py
oliverchang: ah, I thought you were doing it to make sure it's not None. ignore!

jonathanmetzman:
This error still needs to be fixed:
```
Using config file /workspace/.pylintrc
| ************* Module workspace.src.python.bot.fuzzers.libfuzzer
| W:353, 2: Parameters differ from overridden 'fuzz' method (arguments-differ)
```

src/python/bot/fuzzers/libfuzzer.py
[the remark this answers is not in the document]
jonathanmetzman: I thought I had submitted this comment earlier. Sorry for the delay
;\

src/python/bot/tasks/fuzz_task.py
jonathanmetzman: I think device.py makes sense? Why do you disagree?

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1536,3 +1540,103 @@ def execute_task(fuzzer_name, job_type):
...
+
+  shell.create_directory_if_needed(resources_path)
+
+  fuchsia_resources_url = environment.get_value('FUCHSIA_RESOURCES_URL')
+  if fuchsia_resources_url is None:
+    raise fuchsia.common.FuchsiaConfigError(
+        'Could not find path for remote'
+        'Fuchsia resources bucket (FUCHSIA_RESOURCES_URL')
+
+  gsutil_command_arguments = ['cp', '-r', fuchsia_resources_url, resources_path]
```
jonathanmetzman: @oliverchang What do you think of invoking `gsutil` directly instead of using `gsutil.py`? I'm inclined to accept it.

jonathanmetzman: Let's add `'-m'` before the `'cp'` to get parallelism.

src/python/platforms/fuchsia/common.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2019 Google LLC
```
oliverchang: let's avoid naming things "common" and be more specific. probably call this file `errors` or something similar?

src/python/fuzzing/tests.py
```diff
@@ -823,10 +822,6 @@ def get_command_line_for_application(file_to_run='',
     return android.adb.get_application_launch_command(
         all_app_args, testcase_path, testcase_file_url)
 
-  elif plt == 'FUCHSIA' and not launcher:
```
oliverchang: note: this stuff may be needed if you ever want to support black box fuzzing on fuchsia.

src/python/bot/tasks/fuzz_task.py
oliverchang: We should be consistent with our use of GSUtilRunner here. It's also needed for local GCS mocking, so please use that instead :)

src/python/fuzzing/tests.py
oliverchang: so may be worth replacing these with a TODO instead :)

src/python/bot/tasks/fuzz_task.py
flowerhack: eh.  not a super-strong rationale.  moved to device.py!

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -324,6 +326,70 @@ def fuzz(self,
...
+    if (not environment.get_value('FUCHSIA_PKEY_PATH')) or (
+        not environment.get_value('FUCHSIA_PORTNUM')):
+      raise fuchsia.errors.FuchsiaConfigError(
+          'FUCHSIA_PKEY_PATH and/or FUCHSIA_PORTNUM was not set')
+    self.ssh_root = [
+        'ssh', '-i',
+        environment.get_value('FUCHSIA_PKEY_PATH'), '-o',
+        'StrictHostKeyChecking no', 'localhost', '-o',
+        'UserKnownHostsFile=/dev/null', '-p',
+        str(environment.get_value('FUCHSIA_PORTNUM'))
```
nopsledder: Can we `#yapf: disable` here, too?  Ideally, we'd have something like:
```
  'ssh', 
  '-i', environment.get_value('FUCHSIA_PKEY_PATH'), 
  '-o', 'StrictHostKeyChecking no',
  '-o', 'UserKnownHostsFile=/dev/null', 
  '-p', str(environment.get_value('FUCHSIA_PORTNUM'))
  'localhost', 
```

flowerhack: done

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -321,6 +323,73 @@ def fuzz(self,
...
+
+    # Don't try to run the fuzzer until we know we can SSH successfully.
+    ssh_test_process = new_process.ProcessRunner('ssh',
+                                                 self.ssh_command('ls')[1:])
+    # Number of retries chosen somewhat arbitrarily; if we get too many
+    # spurious exceptions, we can tweak this.
+    final_result_return_code = ''
+    for _ in range(30):
+      result = ssh_test_process.run_and_wait()
+      if (not result.return_code) and (not result.timed_out):
```
oliverchang: nit: no need for () around the not conditions.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -321,6 +323,73 @@ def fuzz(self,
...
+           extra_env=None):
+    """LibFuzzerCommon.fuzz override."""
+
+    # Don't try to run the fuzzer until we know we can SSH successfully.
+    ssh_test_process = new_process.ProcessRunner('ssh',
+                                                 self.ssh_command('ls')[1:])
+    # Number of retries chosen somewhat arbitrarily; if we get too many
+    # spurious exceptions, we can tweak this.
+    final_result_return_code = ''
+    for _ in range(30):
```
oliverchang: put this number (30) into a class global. also, 30 seems a bit high? is this that flaky? 

i.e.

```python
class FuchsiaQemuLibFuzzerRunner(new_process.ProcessRunner, LibFuzzerCommon):
  ...
  SSH_RETRIES = ...
```

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -321,6 +323,73 @@ def fuzz(self,
...
+      result = ssh_test_process.run_and_wait()
+      if (not result.return_code) and (not result.timed_out):
+        # SSH works; start fuzzing.
```
oliverchang: nit: please put this ssh testing functionality into another private method,

this also make the retrying logic easier, see comment below.

e.g. 

```python
@retry.wrap(...)
def _test_qemu_ssh(self):
  ...


def fuzz(...):
  self.test_qemu_ssh()
  ...
  LibFuzzerCommon.fuzz(...)
```

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -321,6 +323,73 @@ def fuzz(self,
...
+        return LibFuzzerCommon.fuzz(self, corpus_directories, fuzz_timeout,
+                                    artifact_prefix, additional_args)
+      else:
+        final_result_return_code = result.return_code
+        time.sleep(2)
```
oliverchang: use @retry.wrap() for the retry logic. 

example: https://github.com/google/clusterfuzz/blob/master/src/python/google_cloud_utils/credentials.py

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -500,6 +569,8 @@ def get_runner(fuzzer_path, temp_dir=None):
       shutil.copy(os.path.realpath('/bin/sh'), os.path.join(minijail_bin, 'sh'))
 
     runner = MinijailLibFuzzerRunner(fuzzer_path, minijail_chroot)
+  elif environment.platform() == "FUCHSIA":
```
oliverchang: nit: use single quotes for strings.

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1326,6 +1327,8 @@ def execute_task(fuzzer_name, job_type):
 
   # Run the fuzzer to generate testcases. If error occurred while trying
   # to run the fuzzer, bail out.
+  if platform == "FUCHSIA":
```
oliverchang: single quotes here also.

src/python/platforms/fuchsia/device.py
```diff
@@ -16,6 +16,119 @@
...
+       'host=192.168.3.2,hostfwd=tcp::') + portnum + '-:22',
+      '-device', 'e1000,netdev=net0,mac=52:54:00:63:5e:7b',
+      '-L', sharefiles_path
+  ]
+  # yapf: enable
+  # Fuzzing jobs that SSH into the QEMU VM need access to these env vars.
+  environment.set_value('FUCHSIA_PKEY_PATH', pkey_path)
+  environment.set_value('FUCHSIA_PORTNUM', portnum)
+  qemu_process = new_process.ProcessRunner(qemu_command[0], qemu_command[1:])
+  qemu_process.run()
```
oliverchang: can you please add a TODO here for cleanup logic? 

as is, does this create a new process every run?

src/python/platforms/fuchsia/device.py
```diff
@@ -16,6 +16,119 @@
...
+def qemu_setup():
+  """ Sets up and runs a QEMU VM in the background.
+  Does not block the calling process.
+  Fuchsia fuzzers assume a QEMU VM is running; call this routine prior to
+  beginning Fuchsia fuzzing tasks.
+  This initialization routine assumes that the GCS bucket contains the
+  standard Fuchsia SDK, as well as:
+  * /qemu-for-fuchsia/*
+  * /.ssh/* """
+  fuchsia_bucket_name = 'fuchsia_on_clusterfuzz_resources_v1'
```
oliverchang: this seems to be only used for the local path, and isn't actually being used as a bucket name. just use the string directly later in the path.join, and maybe rename to somethng like 'fuchsia-resources'.

src/python/platforms/fuchsia/device.py
```diff
@@ -16,6 +16,119 @@
...
+
+  resources_path = os.path.join(os.getcwd(), fuchsia_bucket_name)
```
oliverchang: please don't use os.getcwd() as a place for storing things. maybe some generic dir under $ROOT_DIR/bot/resources/fuchsia... @inferno-chromium @mbarbella-chromium  any thoughts here?

src/python/platforms/fuchsia/device.py
```diff
@@ -16,6 +16,119 @@
...
+
+  shell.create_directory(resources_path)
+
+  fuchsia_resources_url = environment.get_value('FUCHSIA_RESOURCES_URL')
+  if fuchsia_resources_url is None:
```
oliverchang: nit: just `not fuchsia_resources_url` here.

src/python/platforms/fuchsia/device.py
```diff
@@ -16,6 +16,119 @@
...
+  os.chmod(fvm_tool_path, 0o500)
+  process = new_process.ProcessRunner(fvm_tool_path,
+                                      [drive_path, 'extend', '--length', '1G'])
+  result = process.run_and_wait()
+  if result.return_code or result.timed_out:
+    raise errors.FuchsiaSdkError('Failed to extend FVM: ' + result.output)
+
+  # Need to bake keys into ZBI so we can SSH into it.  See:
+  # fuchsia.googlesource.com/fuchsia/+/refs/heads/master/sdk/docs/ssh.md
+  zbi_tool = os.path.join(resources_path, 'tools', 'zbi')
```
oliverchang: could you break these different setup steps into separate functions? as is this function is getting a little too crowded and making things difficult to comprehend :)

oliverchang:
Also, don't forget to `python butler.py lint` locally and commenting `/gcbrun` on this PR to run all our tests.

src/python/platforms/fuchsia/device.py
flowerhack: tried out bot/resources/fuchsia; lmk if that works for you?

src/python/bot/fuzzers/libfuzzer.py
flowerhack: 30 was kind of arbitrary; switched it to 10 which should be more than fine

src/python/platforms/fuchsia/device.py
flowerhack: todo added!

bot/resources/README.md
```diff
@@ -0,0 +1 @@
+Placeholder for external resources for fuzzing on particular platforms.
```
jonathanmetzman: I think a Readme should go in `resources/fuchsia`. Not sure if the one here is necessary.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -321,6 +324,71 @@ def fuzz(self,
...
+
+    self._test_qemu_ssh()
+    LibFuzzerCommon.fuzz(self, corpus_directories, fuzz_timeout,
+                         artifact_prefix, additional_args)
+
+  def run_single_testcase(self,
+                          testcase_path,
+                          timeout=None,
+                          additional_args=None):
+    #TODO(flowerhack): Fill out this command.
```
jonathanmetzman: nit: please add a space between "#" and "T"

bot/resources/README.md
flowerhack: moved!

src/python/platforms/fuchsia/device.py
flowerhack: moved to bot/resources/fuchsia at @jonathanmetzman 's suggestion

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -321,6 +324,71 @@ def fuzz(self,
...
+    pass
+
+  def ssh_command(self, *args):
+    return self.ssh_root + list(args)
+
+  @retry.wrap(retries=SSH_RETRIES, delay=SSH_WAIT, function='_test_qemu_ssh')
+  def _test_qemu_ssh(self):
+    """ Tests that a VM is up and can be successfully SSH'd into.
```
jonathanmetzman: Why do you have this blank space at the beginning and end of every docstring? Can you get rid of it?

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -321,6 +324,71 @@ def fuzz(self,
...
+    Raises an exception if no success after MAX_SSH_RETRIES. """
+    ssh_test_process = new_process.ProcessRunner('ssh',
+                                                 self.ssh_command('ls')[1:])
```
jonathanmetzman: I don't love what's being done with ssh_command here.
Can we make another method that ssh_command adds to self.ssh_root?

src/python/platforms/fuchsia/device.py
```diff
@@ -16,6 +16,142 @@
...
+
+  # Perform some more initiailization steps.
+  extend_fvm(fuchsia_resources_dir, drive_path)
+  add_keys_to_zbi(fuchsia_resources_dir, initrd_path, fuchsia_zbi)
+
+  # TODO(flowerhack): Add a mechanism for choosing portnum dynamically.
+  portnum = '56339'
+
+  # yapf: disable
+  qemu_command = [
```
jonathanmetzman: Maybe just call this `qemu_args` and remove `qemu_path`?
That way we don't need to slice the list.

src/python/platforms/fuchsia/device.py
```diff
@@ -16,6 +16,142 @@
...
+  os.chmod(fvm_tool_path, 0o500)
+  process = new_process.ProcessRunner(fvm_tool_path,
+                                      [drive_path, 'extend', '--length', '1G'])
+  result = process.run_and_wait()
+  if result.return_code or result.timed_out:
+    raise errors.FuchsiaSdkError('Failed to extend FVM: ' + result.output)
+
+
+def add_keys_to_zbi(fuchsia_resources_dir, initrd_path, fuchsia_zbi):
+  """ Adds keys to the ZBI so we can SSH into it.  See:
```
jonathanmetzman: super nit: I think we generally do one space after a period.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -321,6 +324,71 @@ def fuzz(self,
...
+    return self.ssh_command('ls')
+
+  def fuzz(self,
+           corpus_directories,
+           fuzz_timeout,
+           artifact_prefix=None,
+           additional_args=None,
+           extra_env=None):
+    """LibFuzzerCommon.fuzz override."""
+
```
oliverchang: nit: no need for newline here.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -321,6 +324,71 @@ def fuzz(self,
...
+    """ Tests that a VM is up and can be successfully SSH'd into.
+    Raises an exception if no success after MAX_SSH_RETRIES. """
+    ssh_test_process = new_process.ProcessRunner('ssh',
+                                                 self.ssh_command('ls')[1:])
+    result = ssh_test_process.run_and_wait()
+    if result.return_code or result.timed_out:
+      raise fuchsia.errors.FuchsiaConnectionError(
+          'Failed to establish initial SSH connection: ' +
+          str(result.return_code))
+    return
```
oliverchang: nit: useless return.

src/python/platforms/fuchsia/device.py
```diff
@@ -16,6 +16,142 @@
...
+
+
+def initialize_resources_dir():
+  """ Download Fuchsia QEMU resources from GCS bucket. """
+  resources_dir = environment.get_value('RESOURCES_DIR')
+  if not resources_dir:
+    raise errors.FuchsiaConfigError('Could not find RESOURCES_DIR')
+  fuchsia_resources_dir = os.path.join(resources_dir, 'fuchsia')
+
+  shell.create_directory(fuchsia_resources_dir)
```
oliverchang: should this recreate every time it runs? Since we're redownloading things every time anyway, may be better to start from a clean slate?

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -17,7 +17,10 @@
 import os
 import shutil
 
-from libFuzzer import constants
+from . import engine_common
```
inferno-chromium: Please use python 2 style imports for now, rather than relative imports. We will be transitioning to Py3, and currently this might cause an issue (see also revert https://github.com/google/clusterfuzz/commit/559c76b4e04c55e940309c2fb7383037ff8c3e60). basically we call launcher.py in a new shell and it can cause `ValueError: Attempted relative import in non-package`

just do import engine_common
from libFuzzer import constants.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -321,6 +324,71 @@ def fuzz(self,
...
+
+  SSH_RETRIES = 10
+  SSH_WAIT = 2
+
+  def __init__(self, executable_path, default_args=None):
+    if (not environment.get_value('FUCHSIA_PKEY_PATH')) or (
+        not environment.get_value('FUCHSIA_PORTNUM')):
+      raise fuchsia.errors.FuchsiaConfigError(
+          'FUCHSIA_PKEY_PATH and/or FUCHSIA_PORTNUM was not set')
+    # yapf: disable
```
inferno-chromium: use "#pylint: disable=" with the particular rule name than yapf.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -321,6 +324,71 @@ def fuzz(self,
...
 
 
+class FuchsiaQemuLibFuzzerRunner(new_process.ProcessRunner, LibFuzzerCommon):
+  """libFuzzer runner (when Fuchsia is the target platform)."""
+
+  SSH_RETRIES = 10
+  SSH_WAIT = 2
+
+  def __init__(self, executable_path, default_args=None):
+    if (not environment.get_value('FUCHSIA_PKEY_PATH')) or (
```
inferno-chromium: environment.get_value('FUCHSIA_PKEY_PATH') and FUCHSIA_PORTNUM in a local var and then use in this if and usage below.

src/python/platforms/fuchsia/__init__.py
```diff
@@ -13,4 +13,5 @@
 # limitations under the License.
 """"Fuchsia initialization."""
 
-import device
+from . import device
```
inferno-chromium: Just python 2 style import for now. import device and import errors.

src/python/platforms/fuchsia/device.py
```diff
@@ -16,6 +16,142 @@
 # TODO(mbarbella): Re-enable this check once functions below are implemented.
```
inferno-chromium: can you remove this todo.

src/python/bot/fuzzers/libfuzzer.py
flowerhack: unfortunately this isn't a pylint rule as far as i can tell--the formatting is perfectly *fine* according to PEP, but yapf is overzealous and tries to fit as many list elements as possible onto a single line.  so really we want yapf to be less aggressive, rather than disabling any particular violation

src/python/bot/fuzzers/libfuzzer.py
flowerhack: ah, sorry, must've rebased at the wrong time. fixed

src/python/bot/fuzzers/libfuzzer.py
flowerhack: I'm not sure I quite understood the alternative you were suggesting -- but does what I've done here instead ('ssh', self.ssh_Args + ['ls']) make sense / work for you?

src/python/platforms/fuchsia/device.py
flowerhack: set recreate=True

flowerhack:
"Will you still be adding tests in this PR, or will do it in another?"

Yep, the plan's to do it in my next changelist, which I'm working on now.  I was planning to do it all-in-one initially, but when I realized how complicated the scaffolding's going to be for our tests, I thought it may be better to do it across two changelists, since (1) smaller changes and (2) this code won't be running in production yet anyway.

flowerhack:
and: i think i've answered everyone's comments for now!

flowerhack:
meta: is there some way to view *all* conversations on a Github PR?  i can't find a conversation that i resolved, that i'd like to unresolve.

workaround for now: @inferno-chromium ; i changed the TODO in device.py to be assigned to me, rather than resolving it outright--we can't remove the "dummy" functions there altogether, because CF expects them to exist (and errors out if they don't), and we may want those functions to do something in the future anyway--so, leaving it as a TODO for me for now

src/python/bot/fuzzers/libfuzzer.py
jonathanmetzman: Yep that's better.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -321,6 +323,68 @@ def fuzz(self,
                                 artifact_prefix, additional_args, extra_env)
 
 
+class FuchsiaQemuLibFuzzerRunner(new_process.ProcessRunner, LibFuzzerCommon):
+  """libFuzzer runner (when Fuchsia is the target platform)."""
+
+  SSH_RETRIES = 10
+  SSH_WAIT = 2
```
jonathanmetzman: nit: can you add the suffix '_SECONDS' to this attribute please?
Can come in another CL.

PR #309, opened 2019-03-25
 Fix factory reset for android cuttlefish bots.

inferno-chromium:


src/python/platforms/android/adb.py
```diff
@@ -84,9 +84,9 @@
 
 def bad_state_reached():
   """Wait when device is in a bad state and exit."""
-  time.sleep(BAD_STATE_WAIT)
```
inferno-chromium: This makes sure we show the fatal message first and then time.sleep (correct order).

PR #312, opened 2019-03-26
Android code cleanup.

inferno-chromium:


src/python/platforms/android/gestures.py
```diff
@@ -22,28 +22,26 @@
...
   package_name = os.getenv('PKG_NAME')
   if not package_name:
     # No package to send gestures to, bail out.
     return
 
-  if len(gestures) != 1 or not gestures[0].startswith('monkey'):
-    # Bad gesture string, bail out.
-    return
+  assert len(gestures) == 1
+  assert gestures[0].startswith('monkey,')
```
mbarbella-chromium: Since we accept user-supplied strings this could lead to exceptions rather than errors if someone ever types something random in for Android. The old behavior seems slightly better to me.

src/python/platforms/android/gestures.py
```diff
@@ -22,28 +22,26 @@
 MONKEY_THROTTLE_DELAY = 100
 
 # Maximum number of monkey events per testcase.
-NUM_MONKEY_EVENTS = 50
+NUM_MONKEY_EVENTS = 25
```
mbarbella-chromium: Won't this cause issues with reproducibility for existing tests? Or do we not expect to have any relevant ones open at the moment?

src/python/platforms/android/gestures.py
inferno-chromium: Makes sense. Reverting.

src/python/platforms/android/gestures.py
inferno-chromium: Nothing interesting is open atm, so this is fine.

PR #314, opened 2019-03-26
Fix log parsing in Android.

inferno-chromium:


src/python/platforms/android/logger.py
```diff
@@ -49,17 +49,19 @@ def filter_log_output(output):
...
   filtered_output = ''
-  last_process_id = 0
+  last_process_tuple = (None, None)
   for line in output.splitlines():
     line = line.strip()
     if not is_line_valid(line):
       continue
 
-    m_line = re.match('[^D]/([^:]+)[:] (.*)', line)
+    m_line = re.match(r'^[VDIWEFS]/([^(]+)\((\d+)\)[:] (.*)', line)
```
mbarbella-chromium: Can you add a comment here? This is still very hard to follow.

PR #315, opened 2019-03-26
List output from all processes when invoking ps on Android devices.

mbarbella-chromium:
PTAL. I'm not totally sure if this is correct in all cases, but seemed small enough to draft the CL so we could discuss it here.

I am sure that for the dex2oat case in wait_until_package_optimization_complete that this wasn't working correctly. Running ps only seems to show a very limited number of processes. I'm not sure if this would have caused any issues in the other 2 places it was used, and it seemed like it could be very noisy in the memory statistics printing case.

### DIFF
--- a/src/python/platforms/android/adb.py
+++ b/src/python/platforms/android/adb.py
@@ -398,8 +398,7 @@ def get_package_name(apk_path=None):
 def get_process_and_child_pids(process_name):
   """Return process and child pids matching a process name."""
   pids = []
-  ps_output = get_ps_output()
-  ps_output_lines = ps_output.splitlines()
+  ps_output_lines = get_ps_output().splitlines()
 
   while True:
     old_pids_length = len(pids)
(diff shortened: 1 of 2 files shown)


src/python/platforms/android/adb.py
```diff
@@ -398,7 +398,7 @@ def get_package_name(apk_path=None):
 def get_process_and_child_pids(process_name):
   """Return process and child pids matching a process name."""
   pids = []
-  ps_output = get_memory_usage_info()
+  ps_output = get_ps_output()
```
inferno-chromium: can directly use it on next line.

src/python/platforms/android/adb.py
```diff
@@ -908,8 +908,8 @@ def wait_until_package_optimization_complete():
   start_time = time.time()
 
   while time.time() - start_time < REBOOT_TIMEOUT:
-    memory_output = get_memory_usage_info()
-    package_optimization_finished = 'dex2oat' not in memory_output
+    ps_output = get_ps_output()
```
inferno-chromium: can directly use it on next line.

src/python/system/process_handler.py
```diff
@@ -336,10 +336,10 @@ def run_process(cmdline,
 
   # If a crash is found, then we add the memory state as well.
   if return_code and plt == 'ANDROID':
-    memory_usage_info = android.adb.get_memory_usage_info()
+    process_info = android.adb.get_ps_output()
     if memory_usage_info:
```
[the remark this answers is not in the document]
mbarbella-chromium: Went with ps_output like in the other cases instead and fixed all of the usages here.

PR #317, opened 2019-03-27
Fix another logcat parsing issue

inferno-chromium:


inferno-chromium:
This is yet another case, see error
"Failed to parse logcat line: F/chromium(11384):"
there is no trailing space at end.

inferno-chromium:
Another TBR for minor fix.

PR #326, opened 2019-03-28
Add more ignore signatures.

inferno-chromium:


inferno-chromium:
Based on seeing https://clusterfuzz.com/testcase-detail/5674452384874496

PR #330, opened 2019-03-28
Don't recover when running adb reboot.

inferno-chromium:


inferno-chromium:
This causes unneccessary stop_cvd->launch_cvd device reset when reboot is an expected thing to do - "Unable to query device, resetting device connection."

PR #332, opened 2019-03-29
Revert "Don't recover when running adb reboot."

inferno-chromium:
Reverts google/clusterfuzz#330


inferno-chromium:
This is leading to more failures, keeping recovery calls as-in in reboot.

Issue #297, opened 2019-03-22
Add other presubmit tests to Travis, CI


inferno-chromium:
Right now, we do 
* lint
* run_all_tests

but missing several others from PRESUBMIT.py. We should move those and then get rid of PRESUBMIT.py.

inferno-chromium:
@oliverchang as fyi. noticed this while making changes to data_types.py

PR #335, opened 2019-03-29
Add lint presubmit rules (#297).

inferno-chromium:
* License header check
* Python import order check
* Python test directory `__init__.py` check
* YAML validation check


src/local/butler/lint.py
```diff
@@ -15,10 +15,140 @@
...
+  source_filename = os.path.basename(file_path)
+  for check_filename in _LICENSE_CHECK_IGNORE_FILENAMES:
+    if check_filename == source_filename:
+      return
+
+  if _LICENSE_CHECK_STRING in open(file_path).read():
+    return
+
+  print 'Missing license header for %s.' % file_path
+  sys.exit(1)
```
jonathanmetzman: Why stop linting if one file doesn't have a license? See my comment on `lint.py:177`

src/local/butler/lint.py
```diff
@@ -15,10 +15,140 @@
...
+
+
+def py_import_order(file_path):
+  """Validate that python imports are alphabetized."""
+
+  def _validate_block(import_block):
```
jonathanmetzman: If `_validate_block` doesn't use closures, why define it within `py_import_order`?
I think a separate module or a class is a better way to organize functions than defining them within each other.

src/local/butler/lint.py
```diff
@@ -15,10 +15,140 @@
...
+    if not import_block:
+      return []
+
+    sorted_import_block = sorted(import_block, key=lambda i: i.lower())
+    if sorted_import_block == import_block:
+      return []
+
+    return ['\n'.join(sorted_import_block)]
+
+  def _validate_imports(data):
```
jonathanmetzman: Same comment as above.

src/local/butler/lint.py
```diff
@@ -15,10 +15,140 @@
...
+    from_imports = []
+    corrected_import_blocks = []
+    for line in data.splitlines():
+      if line.startswith('import '):
+        imports.append(line)
+      else:
+        corrected_import_blocks += _validate_block(imports)
+        imports = []
+
+      if line.startswith('from '):
```
jonathanmetzman: I'm not sure separating `from` blocks and `import` blocks is correct. 
Currently this allows:
```
import b
from a import x
```
Which contradicts the example in the style guide: http://google.github.io/styleguide/pyguide.html#313-imports-formatting.

I think a block should end on a newline.

Is there an open source implementation of this check that we could use instead?

src/local/butler/lint.py
```diff
@@ -15,10 +15,140 @@
...
+    corrected_import_blocks += _validate_block(from_imports)
+
+    if not corrected_import_blocks:
+      return
+
+    suggestions = '\n\n--------\n\n'.join(corrected_import_blocks)
+    print('File {filename} has non-alphabetized import blocks. '
+          'Suggested order:\n\n{suggestions}').format(
+              filename=file_path, suggestions=suggestions)
+    sys.exit(1)
```
jonathanmetzman: Why stop linting if one file has a bad import order? See my comment on lint.py:177

src/local/butler/lint.py
```diff
@@ -15,10 +15,140 @@
...
+  """Check test directory has a __init__.py file. Otherwise, the test does not
+  execute at all."""
+  if not file_path.endswith(_PY_TEST_SUFFIX):
+    return
+
+  test_directory = os.path.dirname(file_path)
+  if _PY_INIT_FILENAME not in os.listdir(test_directory):
+    print 'Missing {filename} file in test directory {dir}.'.format(
+        filename=_PY_INIT_FILENAME, dir=test_directory)
+    sys.exit(1)
```
jonathanmetzman: Why stop linting if one directory is missing an init file? See my comment on lint.py:177

src/local/butler/lint.py
```diff
@@ -15,10 +15,140 @@
...
+  """Run yaml validation for the given source path."""
+  if os.path.basename(file_path) in _YAML_EXCEPTIONS:
+    return
+
+  try:
+    with open(file_path) as handle:
+      yaml.safe_load(handle.read())
+  except Exception as e:
+    print 'Failed yaml validation for %s.\n\n%s' % (file_path, e)
+    sys.exit(1)
```
jonathanmetzman: Why stop linting if one yaml file is invalid? See my comment on lint.py:177

src/local/butler/lint.py
```diff
@@ -15,10 +15,140 @@
...
 
+_GOLINT_EXCEPTIONS = [
+    'types.go'  # Not all model names conform to Go naming conventions.
+]
+_LICENSE_CHECK_FILENAMES = ['Dockerfile']
+_LICENSE_CHECK_EXTENSIONS = [
+    '.bash',
+    '.c',
+    '.cc',
+    '.cpp',
```
jonathanmetzman: I think we are missing `.proto` and `.ps1`

src/local/butler/lint.py
```diff
@@ -34,23 +164,32 @@ def execute(_):
   else:
     _, output = common.execute('git diff --name-only FETCH_HEAD')
 
+  file_paths = [f for f in output.splitlines() if os.path.exists(f)]
```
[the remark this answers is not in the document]
inferno-chromium: If file is deleted, then we don't want lint to run on it and cause exception.

src/local/butler/lint.py
```diff
@@ -34,23 +164,32 @@ def execute(_):
...
   ]
-  go_changed_file_paths = [f for f in output.splitlines() if f.endswith('.go')]
+  go_changed_file_paths = [f for f in file_paths if f.endswith('.go')]
+  yaml_changed_file_paths = [f for f in file_paths if f.endswith('.yaml')]
 
   for file_path in py_changed_file_paths:
-    if os.path.exists(file_path):
-      common.execute('pylint ' + file_path)
-      common.execute('yapf -d ' + file_path)
+    common.execute('pylint ' + file_path)
```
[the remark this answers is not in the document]
inferno-chromium: yes probably another cl to account for this.

src/local/butler/lint.py
inferno-chromium: done

src/local/butler/lint.py
```diff
@@ -15,10 +15,140 @@
...
+    '.yaml',
+]
+_LICENSE_CHECK_IGNORE_FILENAMES = ['technology.css']
+_LICENSE_CHECK_IGNORE_DIRECTORIES = [
+    'third_party',
+    'templates',  # Generated code.
+]
+_LICENSE_CHECK_STRING = 'http://www.apache.org/licenses/LICENSE-2.0'
+_PY_TEST_SUFFIX = '_test.py'
+_PY_INIT_FILENAME = '__init__.py'
```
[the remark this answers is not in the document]
inferno-chromium: Globals are sometimes better since 
1. easier to modify - e.g. _LICENSE_CHECK_EXTENSIONS
2. if you have do local, you have to do it lowercase (otherwise lint complains). lowercase does not make look like constant.
_PY_INIT_FILENAME was used twice, so just thought to keep it in constant.

src/local/butler/lint.py
inferno-chromium: can be just removed, done.

src/local/butler/lint.py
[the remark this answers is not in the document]
jonathanmetzman: What changed though?

src/local/butler/lint.py
inferno-chromium: see comment above.

src/local/butler/lint.py
inferno-chromium: see comment above.

src/local/butler/lint.py
inferno-chromium: see comment above.

src/local/butler/lint.py
inferno-chromium: see comment above.

src/local/butler/lint.py
inferno-chromium: can't find any.

src/local/butler/lint.py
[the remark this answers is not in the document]
inferno-chromium: We had the "if os.path.exists(file_path):" repeated inside every loop. Now we just do it once at start instead of repeating it for every file type check (like before it was repeated for python, go files, etc).

src/local/butler/lint.py
jonathanmetzman: Should we file a bug then to fix this?

src/local/butler/lint.py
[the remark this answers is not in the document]
jonathanmetzman: >problem is output becomes more messed up to read if you combine everything 

Why? They aren't printing in parallel, just print one error after another, like a compiler does.

>once you have a failure, person would anyway have to run "python butler.py lint" and then fix failure one at a time and only once it finishes without error, then reupload.

The same is true for compilers, but they go to great effort to continue printing more error messages after the first error, because it is easier to fix all issues at once than to rerun the command and fix them one at a time.

When doing the mutator plugin CL it was very annoying to run the command for each error (particularly since the lint command is slow when there are many files to run on).

src/local/butler/lint.py
inferno-chromium: filed https://github.com/google/clusterfuzz/issues/340

src/local/butler/lint.py
```diff
@@ -15,10 +15,138 @@
...
+def license_validate(file_path):
+  """Run license header validation."""
+  filename = os.path.basename(file_path)
+  extension = os.path.splitext(file_path)[1]
+  if (filename not in _LICENSE_CHECK_FILENAMES and
+      extension not in _LICENSE_CHECK_EXTENSIONS):
+    return
+
+  path_directories = file_path.split(os.sep)
+  for directory in _LICENSE_CHECK_IGNORE_DIRECTORIES:
```
oliverchang: nit: consider 

```python
if any(d in _LICENSE_CHECK_IGNORE_DIRECTORIES for d in path_directories):
  return
```

src/local/butler/lint.py
```diff
@@ -15,10 +15,138 @@
...
+    if directory in path_directories:
+      return
+
+  source_filename = os.path.basename(file_path)
+  for check_filename in _LICENSE_CHECK_IGNORE_FILENAMES:
```
oliverchang: same here:

```python
if any(source_filename == ignore for ignore in _LICENSE_CHECK_IGNORE_FILENAMES):
  return
```

src/local/butler/lint.py
```diff
@@ -15,10 +15,138 @@
...
+    if check_filename == source_filename:
+      return
+
+  if _LICENSE_CHECK_STRING in open(file_path).read():
```
oliverchang: nit:

```python
with open(file_path) as f:
  if _LICENSE_CHECK_STRING in  f.read():
    return
```

to make the closing explicit.

src/local/butler/lint.py
```diff
@@ -15,10 +15,138 @@
...
+    return
+
+  print 'Missing license header for %s.' % file_path
```
oliverchang: nit: for future python 3 compatibility, let's use `from __future__ import print_function` to enforce parentheses here.

PR #337, opened 2019-03-29
Cluster_size.sh

ameet20:
#!/bin/bash

# N is the node number of hadoop cluster
N=$1

if [ $# = 0 ]
then
	echo "Please specify the node number of hadoop cluster!"
	exit 1
fi

# change slaves file
i=1
rm config/slaves
while [ $i -lt $N ]
do
	echo "hadoop-slave$i" >> config/slaves
	((i++))
done 

echo ""

echo -e "\nbuild docker hadoop image\n"

# rebuild image
sudo docker build -t :1.0 .

echo ""

### DIFF
--- a/Cluster_size.sh
+++ b/Cluster_size.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+# N is the node number of hadoop cluster
+N=$1
+
+if [ $# = 0 ]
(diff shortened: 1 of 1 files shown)


inferno-chromium:
spam! closing

PR #339, opened 2019-03-29
Minor stack_analyzer cleanup.

mbarbella-chromium:
See https://bugs.chromium.org/p/chromium/issues/detail?id=945249#c5 for context.


src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -267,6 +267,7 @@
     'asan_check_access',
     'asan_osx_dynamic.dylib',
     'assert',
+    'do_malloc',
```
mbarbella-chromium: This is for tcmalloc frames like: (anonymous namespace)::do_malloc(unsigned long)

Because of the prefix it has to be in IGNORE_CONTAINS.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -632,6 +635,10 @@ def ignore_stack_frame(stack_frame, symbolized):
   # properly cross-platform.
   normalized_stack_frame = stack_frame.replace('\\', '/')
 
+  # Ignore frames that exactly match the app name.
```
mbarbella-chromium: This seems a little bit risky but I think it makes sense in most cases. I'm happy to land this with just the tcmalloc fixes if you think that seems better.

mbarbella-chromium:
PTAL when you have time.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -401,6 +402,8 @@
     'strcmp',
     'strcpy',
     'strlen',
+    'tcmalloc',
```
inferno-chromium: instead of these 2, better to use '/tcmalloc/', similar to jemalloc above.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -632,6 +635,10 @@ def ignore_stack_frame(stack_frame, symbolized):
   # properly cross-platform.
   normalized_stack_frame = stack_frame.replace('\\', '/')
 
+  # Ignore frames that exactly match the app name.
+  if environment.get_value('APP_NAME') == normalized_stack_frame:
```
inferno-chromium: can you move "environment.get_value('APP_NAME')" into a local and then also add a null check

if app_name and app_name == normalized_stack_frame ?

mbarbella-chromium: Any reason for this? It should work fine with None.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
inferno-chromium: Discussed offline, keeping this.

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -562,8 +562,6 @@ def test_assert(self):
                                   expected_state, expected_stacktrace,
                                   expected_security_flag)
 
-    environment.remove_key('ASSERTS_HAVE_SECURITY_IMPLICATION')
```
inferno-chromium: can just remove all of these environment.remove_key from this file ?

inferno-chromium: Also, where did the test go ? is it intentional to remove it ? can keep it with chrome frame, does not hurt.

mbarbella-chromium: I removed it intentionally. We already know that IGNORE_CONTAINS works, so seems redundant to include another test for it.

mbarbella-chromium: And yeah, all of those should be removed. I thought I had checked but I'll do another pass and clean those up.

PR #342, opened 2019-03-29
Fix mutator plugins for OSS-Fuzz

jonathanmetzman:
Bind plugin directory in chroot.
Also fix test so that the directory isn't bound implicitly.

### DIFF
--- a/src/python/bot/fuzzers/libFuzzer/launcher.py
+++ b/src/python/bot/fuzzers/libFuzzer/launcher.py
@@ -677,6 +677,28 @@ def get_printable_command(command, fuzzer_path, use_minijail):
   return engine_common.get_command_quoted(command)
 
 
+def maybe_use_mutator_plugin(target_name, extra_env, fuzzing_strategies, chroot):
+  """Decide whether to use a mutator plugin. If yes and there is a usable plugin
+  available for |target_name|, then add it to LD_PRELOAD in |extra_env|, add a
+  strategy to fuzzing_strategies, and add chroot bindings if |chroot| is not
+  None."""
+  if not do_mutator_plugin():
+    return
+  mutator_plugin_path = mutator_plugin.get_mutator_plugin(target_name)
+  if not mutator_plugin_path:
+    return
+
+  logs.log('Using mutator plugin: %s' % mutator_plugin_path)
+  # TODO(metzman): Change the strategy to record which plugin was used, and
+  # not simply that a plugin was used.
+  fuzzing_strategies.append(strategy.MUTATOR_PLUGIN_STRATEGY)
+  extra_env['LD_PRELOAD'] = mutator_plugin_path
+  if chroot:
+    mutator_plugin_dir = os.path.dirname(mutator_plugin_path)
+    chroot.add_binding(
+        minijail.ChrootBinding(mutator_plugin_dir, mutator_plugin_dir, False))
+
+
 def main(argv):
   """Run libFuzzer as specified by argv."""
   atexit.register(fuzzer_utils.cleanup)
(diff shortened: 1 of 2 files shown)


src/python/bot/fuzzers/libFuzzer/launcher.py
```diff
@@ -830,14 +855,11 @@ def main(argv):
...
   extra_env = {}
-  if do_mutator_plugin():
-    mutator_plugin_path = mutator_plugin.get_mutator_plugin(target_name)
-    if mutator_plugin_path:
-      logs.log('Using mutator plugin: %s' % mutator_plugin_path)
-      # TODO(metzman): Change the strategy to record which plugin was used, and
-      # not simply that a plugin was used.
-      fuzzing_strategies.append(strategy.MUTATOR_PLUGIN_STRATEGY)
-      extra_env['LD_PRELOAD'] = mutator_plugin_path
+  if use_minijail:
```
inferno-chromium: Why need this when minijail_chroot already has the right value:

  if use_minijail:
    minijail_chroot = runner.chroot
  else:
    minijail_chroot = None

src/python/bot/fuzzers/libFuzzer/launcher.py
```diff
@@ -677,6 +677,31 @@ def get_printable_command(command, fuzzer_path, use_minijail):
   return engine_common.get_command_quoted(command)
 
 
+def maybe_use_mutator_plugin(target_name, extra_env, fuzzing_strategies,
+                             chroot):
+  """Decide whether to use a mutator plugin. If yes and there is a usable plugin
+  available for |target_name|, then add it to LD_PRELOAD in |extra_env|, add a
+  strategy to |fuzzing_strategies|, and add chroot bindings if |chroot| is not
+  None."""
+  if not do_mutator_plugin():
```
inferno-chromium: nit: i prefer "if do_mutator_plugin" to be in main, along with other do_ functions. that way, this function is just use_mutator_plugin instead of maybe_.

src/python/bot/fuzzers/libFuzzer/launcher.py
```diff
@@ -677,6 +677,31 @@ def get_printable_command(command, fuzzer_path, use_minijail):
...
+
+  mutator_plugin_path = mutator_plugin.get_mutator_plugin(target_name)
+  if not mutator_plugin_path:
+    return
+
+  logs.log('Using mutator plugin: %s' % mutator_plugin_path)
+  # TODO(metzman): Change the strategy to record which plugin was used, and
+  # not simply that a plugin was used.
+  extra_env['LD_PRELOAD'] = mutator_plugin_path
+  fuzzing_strategies.append(strategy.MUTATOR_PLUGIN_STRATEGY)
```
inferno-chromium: move this line 697 in caller, similar to comment above. that way, also you dont need to pass fuzzing_strategies

src/python/bot/fuzzers/libFuzzer/launcher.py
jonathanmetzman: Done.

src/python/bot/fuzzers/libFuzzer/launcher.py
jonathanmetzman: Changed it to do this but I don't like this pattern and I think we should refactor this code so that all strategy handling isn't done in main.

src/python/bot/fuzzers/libFuzzer/launcher.py
jonathanmetzman: Done.

src/python/bot/fuzzers/libFuzzer/launcher.py
inferno-chromium: +1

jonathanmetzman:
Do the integration tests run in parallel? The minijail test and the non-minijail might overwrite each others' directories.
I can use uuid to get a reasonable guarantee that the directories used by different tests don't overlap.

inferno-chromium:
Regular integration tests (including libfuzzer integration) run in parallel, untrusted runner tests run one by one.

See /build/clusterfuzz$ cat local/tests/run_all_tests
....
INTEGRATION=1 $SCRIPT_DIR/run_tests

and

/build/clusterfuzz$ cat local/tests/run_tests 
#!/bin/bash -ex

echo 'Running unit tests...'
python butler.py py_unittest -t appengine -m "$@"
python butler.py py_unittest -t core -m "$@"
python butler.py go_unittest "$@"

Yes any pid or even different tempdir ?

PR #344, opened 2019-03-30
Prevent data corruption in /data/data for android cuttlefish bots.

inferno-chromium:


inferno-chromium:
tested this on a couple of iteration on test cuttlefish bots. tbr simple fix. this is intended to remove "device in bad state" errors completely on cuttlefish bots.

PR #356, opened 2019-04-02
Add logging to find if launcher finished for no new units case.

inferno-chromium:


inferno-chromium:
TBR simple logging change to debug AFL failures in "Failed to upload fuzzer stats: empty stats."

PR #357, opened 2019-04-02
Don't mark builds bad when they should be ignored.

inferno-chromium:
Also, remove unneeded crash_state argument from
ignore_stacktrace since it is now always set and
defaults to 'NULL'.


inferno-chromium:
This is a fix for "Bad build windows_asan_chrome." error, even though there is an ignore signature for that.

PR #358, opened 2019-04-02
Don't change environment in check_for_bad_build.

inferno-chromium:
check_for_bad_build is used in progression and regression task
and this messes up the profile (e.g. redzone) that is set by
testcase itself.

### DIFF
--- a/bot/env.yaml
+++ b/bot/env.yaml
@@ -37,7 +37,8 @@ APP_PATH_DEBUG: ''
 # Flag to indicate if the crashing ASSERTs are security issues.
 ASSERTS_HAVE_SECURITY_IMPLICATION: false
 
-# Flag to indicate if we want to do bad build check on the application build.
+# Flag to indicate if we want to do bad build check on the application build. This launches an
+# application without a testcase and verifies that it does not crash on startup.
 BAD_BUILD_CHECK: true
 
 # Store build artifacts in cache to save network bandwidth.
@@ -80,6 +81,9 @@ FUZZER_TIMEOUT: 5400
 # Increase GCE metadata server timeout to mitigate transient errors.
 GCE_METADATA_TIMEOUT: 60
 
+# Enable leak detection for ASan builds.
+LSAN: false
+
 # Maximum number of simultaneous fuzzing / testcase execution threads.
 MAX_FUZZ_THREADS: 4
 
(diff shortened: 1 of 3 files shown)


bot/env.yaml
```diff
@@ -108,9 +112,6 @@ REINSTALL_APP_BEFORE_EACH_TASK: false
 # necessary to reproduce most crashes.
 REQUIRED_APP_ARGS: ''
 
-# Default quarantine size for ASan.
```
inferno-chromium: We already support this via ADDITIONAL_*SAN_OPTIONS, so this is bad and creates mess due to different quarantine default on Android(16) vs non-Android (256)

src/python/system/environment.py
```diff
@@ -151,23 +151,13 @@ def get_asan_options(redzone_size, malloc_context_size, quarantine_size_mb,
...
-  else:
-    # Broken on multiple platforms. E.g. Android tracked in b/25228125.
+  # FIXME: Support container overflow on Android.
+  if bot_platform == 'ANDROID':
     asan_options['detect_container_overflow'] = 0
 
-  # Stack use-after-return.
-  # FIXME: Support Stack UAR on all platforms.
-  if (bot_platform in ['LINUX', 'MAC'] and
-      'detect_stack_use_after_return=' not in asan_options):
```
inferno-chromium: detect_stack_use_after_return= does not work, this is now a dict and we later override it 

  # Additional options. These override the defaults.
  additional_tool_options = get_value('ADDITIONAL_%s_OPTIONS' % tool_name)
  if additional_tool_options:
    tool_options.update(_parse_memory_tool_options(additional_tool_options))

src/python/system/environment.py
```diff
@@ -151,23 +151,13 @@ def get_asan_options(redzone_size, malloc_context_size, quarantine_size_mb,
...
-    max_uar_stack_size_log = 16
-    asan_options['detect_stack_use_after_return'] = 1
-    asan_options['max_uar_stack_size_log'] = max_uar_stack_size_log
-  elif bot_platform == 'WINDOWS':
-    # FIXME: See crbug.com/915245. Disable this explicitly on Windows.
-    asan_options['detect_stack_use_after_return'] = 0
```
inferno-chromium: This shouldn't be needed - https://github.com/google/sanitizers/issues/1032

src/python/fuzzing/tests.py
```diff
@@ -903,9 +903,6 @@ def check_for_bad_build(job_type, crash_revision):
   if not environment.get_value('BAD_BUILD_CHECK'):
     return False
 
-  # Do not detect leaks while checking for bad builds.
-  environment.reset_current_memory_tool_options(leaks=False)
```
inferno-chromium: I think disabling leak detection does not make sense, LSAN=False by default. And in jobs with LSAN=true, i think we care about leaks being found during bad build check.

PR #359, opened 2019-04-03
Fix failure to account for null crash state crashes in bad builds.

inferno-chromium:


inferno-chromium:
This was a regression in https://github.com/google/clusterfuzz/commit/46643f0954cfff32a3685e6cbd6c8bb6b3157002

@oliverchang - regression caused legit "Bad build linux_asan_chrome_v8_arm" crashes to disappear. so this should fix this. rest of places are ok, since they deal with testcase crashes.

PR #366, opened 2019-04-04
Do remaining futurize Py2->3 Stage 1, fix imports.

inferno-chromium:


docker/ci/requirements.txt
```diff
@@ -4,6 +4,7 @@ ecdsa==0.13
 extras==1.0.0
 fixtures==3.0.0
 funcsigs==1.0.2
+future==0.17.1
```
inferno-chromium: This is for stage 2 needs.

src/protos/heartbeat_pb2_grpc.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import absolute_import
```
inferno-chromium: This one does not matter since it is auto-generated, but fixing for now.

src/python/bot/fuzzers/ml/rnn/constants.py
```diff
@@ -53,10 +53,10 @@
 DISPLAY_LEN = 138
 
 # Training script name.
-TRAINING_SCRIPT_NAME = 'train.py'
+TRAINING_SCRIPT_NAME = 'ml.rnn.train'
```
inferno-chromium: See why - https://stackoverflow.com/questions/16981921/relative-imports-in-python-3

src/python/bot/untrusted_runner/untrusted.py
```diff
@@ -137,8 +138,8 @@ def ListFiles(self, request, context):
     return file_impl.list_files(request, context)
 
   @wrap_servicer
-  def CopyFileTo(self, request, context):
-    return file_impl.copy_file_to_worker(request, context)
+  def CopyFileTo(self, request_iterator, context):
```
inferno-chromium: I am not sure if this change is right way here. Can you double check. Does the protos need to be regenerated. This was done to fix this error:

| W:141, 2: Parameters differ from overridden 'CopyFileTo' method (arguments-differ)

inferno-chromium: Also unsure why it didn't show up in local pylint run, but showed up on CI (both travis and googlecloud).

PR #367, opened 2019-04-05
Revert "Do remaining futurize Py2->3 Stage 1, fix imports."

inferno-chromium:
Broke libFuzzer launcher

  File "/mnt/scratch0/clusterfuzz/src/python/bot/fuzzers/libFuzzer/launcher.py", line 38, in <module>
    from . import constants
ValueError: Attempted relative import in non-package


inferno-chromium:
TBR.

PR #369, opened 2019-04-05
Don't overload NFS with storing file if not available.

inferno-chromium:


src/python/bot/startup/heartbeat.py
```diff
@@ -84,8 +84,6 @@ def beat(previous_state, log_filename):
   except Exception:
     current_state = None
 
-  logs.log('Old state %s, current state %s.' % (previous_state, current_state))
```
inferno-chromium: This is just to remove noisy unneeded logging.

PR #375, opened 2019-04-09
Improve stacktrace environment output for easier copying.

mbarbella-chromium:
PTAL when you have time. This is a reland of an old CL to make it easier to copy environment variables.

Note that this should not be merged until https://github.com/google/clusterfuzz-tools/pull/549 has been deployed.


mbarbella-chromium:
Just realized this needs a few more changes on the tools side. I'll prepare another change there before landing this.

src/python/system/environment.py
```diff
@@ -319,8 +324,12 @@ def get_environment_settings_as_string():
       if not environment_variable_value:
         continue
 
-      environment_string += '[Environment] %s = %s\n' % (
-          environment_variable, environment_variable_value)
+      if platform() == 'WINDOWS':
+        environment_string += '[Environment] set "%s=%s"\n' % (
+            quote(environment_variable), quote(environment_variable_value))
+      else:
+        environment_string += '[Environment] export %s="%s"\n' % (
```
inferno-chromium: Do we really need export and set as example here ? We can skip those i think?

src/python/system/environment.py
```diff
@@ -319,8 +324,12 @@ def get_environment_settings_as_string():
       if not environment_variable_value:
         continue
 
-      environment_string += '[Environment] %s = %s\n' % (
-          environment_variable, environment_variable_value)
+      if platform() == 'WINDOWS':
+        environment_string += '[Environment] set "%s=%s"\n' % (
```
inferno-chromium: i don't think quotes around %s=%s work, it should %s="%s"

src/python/system/environment.py
```diff
@@ -308,7 +313,7 @@ def get_environment_settings_as_string():
       sanitizer_options_file_path = (
           android.device.get_sanitizer_options_file_path('ASAN'))
       environment_string += (
-          '[Environment] ASAN Options file = %s with contents %s \n' %
+          '[Environment] ASAN options file "%s" with contents %s \n' %
```
inferno-chromium: change "contents %s \n" to "contents:\n%s\n" to match others.

inferno-chromium:
clusterfuzz-tools side might have to strip quotes ?

src/python/system/environment.py
mbarbella-chromium: Leaving them out SGTM.

src/python/system/environment.py
mbarbella-chromium: Shouldn't matter after the change to remove export/set.

mbarbella-chromium:
> clusterfuzz-tools side might have to strip quotes ?

Right. I'll prepare that change as well before landing this and releasing the newer version.

src/python/system/environment.py
mbarbella-chromium: Done

src/python/system/environment.py
mbarbella-chromium: Done

PR #376, opened 2019-04-10
Fix android tests.

inferno-chromium:


src/python/tests/core/platforms/android/device_test.py
```diff
@@ -0,0 +1,45 @@
...
+
+from platforms.android import adb
+from platforms.android import device
+from system import environment
+from tests.test_libs import helpers as test_helpers
+from tests.test_libs import test_utils
+
+
+@test_utils.android_device_required
+class BaseDeviceTest(unittest.TestCase):
```
mbarbella-chromium: Could we move this class to some place in test_libs?

inferno-chromium: created android_helpers.py similar to untrusted_runner_helpers.py in test_libs.

PR #380, opened 2019-04-11
 Fixes to allow setting ANDROID_SERIAL with format usb:<id>.

mbarbella-chromium:
PTAL


src/python/tests/core/platforms/android/adb_test.py
```diff
@@ -246,3 +248,15 @@ def test_apk_path_in_arg(self):
...
     self.assertEqual(
         adb.get_package_name(self.test_apk_path), self.test_apk_pkg_name)
+
+
+@test_utils.slow
+class ResetUsbTest(android_helpers.AndroidTest):
+  """Tests for reset_usb."""
+
+  def setUp(self):
+    test_helpers.patch_environ(self)
```
inferno-chromium: this is already done by android_helpers.AndroidTest, so setUp can be skipped.

src/python/tests/core/platforms/android/adb_test.py
```diff
@@ -246,3 +248,15 @@ def test_apk_path_in_arg(self):
...
+
+  def test_with_device(self):
+    """Tests reset_usb with a connected device if available."""
+    self.assertTrue(adb.reset_usb())
```
inferno-chromium: nit: newline after this line.

src/python/platforms/android/adb.py
```diff
@@ -577,32 +538,81 @@ def reset_device_connection():
 
 def reset_usb():
   """Reset USB bus for a device serial."""
+
+  def _get_device_path():
```
inferno-chromium: There is
environment.set_value('DEVICE_PATH', adb.get_device_path()) call, so can't move it inside.

Maybe a get_device_path() and that relies on get_device_path_using_usb and get_device_path_using_serial ?
then reset_usb just calls get_device_path ?

src/python/tests/core/platforms/android/adb_test.py
inferno-chromium: We probably want to mock time.sleep as otherwise this test makes things too slow.

src/python/tests/core/platforms/android/adb_test.py
```diff
@@ -246,3 +248,15 @@ def test_apk_path_in_arg(self):
     """Test apk path passed as argument."""
     self.assertEqual(
         adb.get_package_name(self.test_apk_path), self.test_apk_pkg_name)
+
+
+@test_utils.slow
```
inferno-chromium: this slow can be removed after mocking time.sleep.

src/python/tests/core/platforms/android/adb_test.py
mbarbella-chromium: I considered that, but couldn't that cause issues with other tests if the device isn't yet ready?

src/python/tests/core/platforms/android/adb_test.py
mbarbella-chromium: Done

src/python/tests/core/platforms/android/adb_test.py
mbarbella-chromium: Done

src/python/tests/core/platforms/android/adb_test.py
mbarbella-chromium: Will remove this depending on what we decide to do here.

src/python/platforms/android/adb.py
mbarbella-chromium: Do we need that other call? It looks like the variable is only used here. It seems like it would be easy enough to factor this out so that it works safely, but it would be nice to remove this if we can.

Either way I think this function could use a bit more cleanup which I'll do once the question above is resolved.

mbarbella-chromium: Discussed off-review. Needed to keep the other call but cleaned this up a bit more.

PR #382, opened 2019-04-12
 Fix check failure comparison regex in stack parsing.

inferno-chromium:
This is causing flakiness in linux_debug_chrome job testcases.

### DIFF
--- a/src/python/crash_analysis/stack_parsing/stack_analyzer.py
+++ b/src/python/crash_analysis/stack_parsing/stack_analyzer.py
@@ -868,8 +868,8 @@ def fix_check_failure_string(failure_string):
   # CHECK_FAILURE_PATTERN, so we looked for "failed:" as preceding string.
   failure_string = re.sub(r'(?<=failed): .*vs\..*$', r'', failure_string)
 
-  # Cover example like len > 0 (-1 vs. 0).".
-  failure_string = re.sub(r' \(.*vs\..*\)\.?$', r'', failure_string)
+  # Cover example like len > 0 (-1 vs. 0)".
+  failure_string = re.sub(r' \(.*vs\..*\).*', r'', failure_string)
 
   # Strip unneeded chars at end.
   return failure_string.strip(' .\'"[]')
--- a/src/python/crash_analysis/stack_parsing/stack_analyzer.py
+++ b/src/python/crash_analysis/stack_parsing/stack_analyzer.py
@@ -866,10 +866,10 @@ def fix_check_failure_string(failure_string):
   # Cover example like "CHECK_EQ( (unsigned)ptr[0],1u) failed: 25 vs. 1".
   # This only happens on Android, where we cannot strip the
   # CHECK_FAILURE_PATTERN, so we looked for "failed:" as preceding string.
-  failure_string = re.sub(r'(?<=failed): .* vs\. .*$', r'', failure_string)
+  failure_string = re.sub(r'(?<=failed): .*\svs\.\s.*$', r'', failure_string)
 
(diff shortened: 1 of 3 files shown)


src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -1968,6 +1968,20 @@ def test_check_failure_with_comparison(self):
                                   expected_state, expected_stacktrace,
                                   expected_security_flag)
 
+  def test_check_failure_with_comparison2(self):
+    """Test for special CHECK failure formats (CHECK_EQ, CHECK_LE, etc.)."""
+    data = self._read_test_data('check_failure_with_comparison2.txt')
+    expected_type = 'CHECK failure'
+    expected_address = ''
+    expected_state = ('layout_snapped_paint_offset == snapped_paint_offset '
+                      'in compositing_layer_propert\n')
```
inferno-chromium: Note that this truncation is due to our 80 char line limit. i can think of something better later but need to think of way to not break existing test, so that fix can come later.

inferno-chromium:
A bunch of testcases should get fixed with this - see https://clusterfuzz.com/testcases?open=yes&q=vs

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -866,10 +866,10 @@ def fix_check_failure_string(failure_string):
   # Cover example like "CHECK_EQ( (unsigned)ptr[0],1u) failed: 25 vs. 1".
   # This only happens on Android, where we cannot strip the
   # CHECK_FAILURE_PATTERN, so we looked for "failed:" as preceding string.
-  failure_string = re.sub(r'(?<=failed): .*vs\..*$', r'', failure_string)
+  failure_string = re.sub(r'(?<=failed): .* vs\. .*$', r'', failure_string)
```
Dor1s: Are those spaces? Should we use `\s` instead?

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -866,10 +866,10 @@ def fix_check_failure_string(failure_string):
...
 
-  # Cover example like len > 0 (-1 vs. 0).".
-  failure_string = re.sub(r' \(.*vs\..*\)\.?$', r'', failure_string)
+  # Cover example like len > 0 (-1 vs. 0)".
+  failure_string = re.sub(r' \(.* vs\. .*\).*', r'', failure_string)
```
Dor1s: same question

src/python/crash_analysis/stack_parsing/stack_analyzer.py
Dor1s: May be even `\s+`, just in case?

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
Dor1s: Thanks for explaining this!

src/python/crash_analysis/stack_parsing/stack_analyzer.py
inferno-chromium: \s seems good, if there are more space, they just get ignored. our condition needs to do a minimal check.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
inferno-chromium: done

PR #383, opened 2019-04-12
Add logging to debug windows libfuzzer backup failures.

inferno-chromium:


inferno-chromium:
gs://clusterfuzz-windows-corpus/libfuzzer/*/* is all 22 bytes and empty.

jonathanmetzman:
Thanks!

src/python/fuzzing/corpus_manager.py
```diff
@@ -102,6 +102,11 @@ def backup_corpus(backup_bucket_name, corpus, directory):
   try:
     backup_archive_path = shutil.make_archive(backup_archive_path,
                                               BACKUP_ARCHIVE_FORMAT, directory)
+    logs.log('Created corpus backup file.',
```
jonathanmetzman: I'm actually not sure if this will provide any more info. I think the backups are being created they just don't contain anything.

inferno-chromium: From stackdriver logs, it does seem like directory has corpus units, so this is just to see what arguments are passed to shutil.make_archive. something very weird, so this is just to log and confirm the arguments and size of archive locally.

PR #384, opened 2019-04-14
Fix incorrect path patching on non-existent paths.

inferno-chromium:
This fixes an issue with empty backup archives on windows libfuzzer jobs.

### DIFF
--- a/src/python/system/path_patcher.py
+++ b/src/python/system/path_patcher.py
@@ -48,9 +48,6 @@ def _is_windows():
 
 def _short_name_modifier(original_path):
   """Get the short path of `path` on windows."""
-  if not _is_windows():
-    return original_path
-
   # These paths don't work with prefix, skip.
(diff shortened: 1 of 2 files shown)


inferno-chromium:
So, the bug was returned result |path| was modified and has \\?\ in front on a non-existent file path in  archive. It causes the path to not work when writing file in a new archive (verified with shutil.make_archive on a win machine with/without patch patcher). So, best to not touch path in those cases.

src/python/system/path_patcher.py
```diff
@@ -48,9 +48,6 @@ def _is_windows():
 
 def _short_name_modifier(original_path):
   """Get the short path of `path` on windows."""
-  if not _is_windows():
```
inferno-chromium: This is redundant for sure, and unneccessary before every operation. we already don't patch()/unpatch() if we are not on windows.

Issue #183, opened 2019-02-13
Python 3 migration


oliverchang:
Adding a tracking bug for Python 3 migration.

The main blocker we have so far is:
- [x] NDB support for Python 3. This appears to be ongoing: https://github.com/googleapis/google-cloud-python/blob/master/ndb/MIGRATION_NOTES.md

inferno-chromium:
Need to move to latest mozprocess version sometime - https://pypi.org/project/mozprocess/#files. As you said, it supports Python 3 now.

gpshead:
FYI the URL for NDB on Python 3 seems to have changed to https://github.com/googleapis/python-ndb/blob/master/MIGRATION_NOTES.md

inferno-chromium:
```
git ls-files | grep requirements.txt | xargs -n1 -t caniusepython3 -r
caniusepython3 -r docker/base/requirements.txt 
Finding and checking dependencies ...

🎉  You have 0 projects blocking you from using Python 3!

caniusepython3 -r docker/ci/requirements.txt 
Finding and checking dependencies ...

You need 1 project to transition to Python 3.
Of that 1 project, 1 has no direct dependencies blocking its transition:

  googledatastore
caniusepython3 -r src/local/requirements.txt 
Finding and checking dependencies ...

🎉  You have 0 projects blocking you from using Python 3!

caniusepython3 -r src/platform_requirements.txt 
Finding and checking dependencies ...

🎉  You have 0 projects blocking you from using Python 3!

caniusepython3 -r src/requirements.txt 
Finding and checking dependencies ...

🎉  You have 0 projects blocking you from using Python 3!
```

inferno-chromium:
I think googledatastore should go away after moving to the new ndb package, right @oliverchang ?

oliverchang:
right, googledatastore will no longer be required afterwards.

inferno-chromium:
Step 1a completed:
```
git ls-files "*.py" | xargs futurize -w -n --stage1 -x libfuturize.fixes.fix_absolute_import
```

inferno-chromium:
Need to enable --py3k in pylint here, but first need help on fixing absolute import things.
```
butler/lint.py:    _execute_command_and_track_error('pylint ' + file_path)
```

inferno-chromium:
Some notes for tracking
```
ml/rnn/generator.py:   rnn_runner = new_process.ProcessRunner('python')
[...]

PR #386, opened 2019-04-15
Futurize Py2->3 Stage 1, fix imports except src/bot/fuzzers (#183)

inferno-chromium:


inferno-chromium:
This is just a reland of https://github.com/google/clusterfuzz/commit/559c76b4e04c55e940309c2fb7383037ff8c3e60 without src/bot/fuzzers pieces. basically anything that we launch via subprocess.call, like python2 launcher.py or ml/rnn/generate or train.py, they break on "ValueError: Attempted relative import in non-package". we need to either make those a package or think of something else.
TBR.

oliverchang:
LGTM!

PR #390, opened 2019-04-17
Don't mark build as bad when leaks are detected.

inferno-chromium:
Leaks can be frequent, however they don't impact reproducing
of regular crashes (as they are reported during shutdown). So,
instead of marking build as bad and blocking fuzzing, we instead
let the build go through which helps with regular fuzzing and
then report the leak with a proper testcase report.


inferno-chromium:
Should help with cases like https://bugs.chromium.org/p/chromium/issues/detail?id=953722. This should help with reporting these automatically.

PR #392, opened 2019-04-18
Remove deprecated backup uri, remove appengine sdk dependencies (#183).

inferno-chromium:


inferno-chromium:
Done, thanks @oliverchang for the idea, now all appengine sdk dependencies are removed.

inferno-chromium:
ready for re-review.

inferno-chromium:
@oliverchang - ping for re-review.

src/appengine/handlers/cron/backup.py
```diff
@@ -44,25 +49,33 @@ def get(self):
...
+        'output_url_prefix': output_url_prefix,
+        'entity_filter': {
+            'kinds': kinds
+        }
+    }
+
+    try:
+      request = _datastore_client().projects().export(
+          projectId=app_id, body=body)
+      response = request.execute()
```
oliverchang: nit: consider just folding this into line 67 as:

```python
_datastore_client()...export(
   ....).execute()
```

since request is never used again.

src/appengine/handlers/cron/backup.py
```diff
@@ -44,25 +49,33 @@ def get(self):
       return
 
     kinds = [
-        kind.kind_name
-        for kind in metadata.Kind.all()
-        if (not kind.kind_name.startswith('_') and
-            kind.kind_name not in EXCLUDED_MODELS)
+        kind for kind in ndb.Model._kind_map  # pylint: disable=protected-access
```
oliverchang: what was wrong with Kind.all ?

we should avoid using internal methods/fields since they may not exist in the python 3 rewrite.

src/appengine/handlers/cron/backup.py
inferno-chromium: Done!

src/appengine/handlers/cron/backup.py
inferno-chromium: Discussed over chat, metadata is coming from appengine sdk, so this dependency needs to be removed. Later when we fix ndb we need to verify this _kind_map works.

PR #393, opened 2019-04-18
Fix incorrect merge_new_feature metric.

inferno-chromium:
This metric is incorrect after https://github.com/google/clusterfuzz/commit/1b51e2a9e1c96894bd745bd698aeb5619db9a37b
Instead rely on best effort calculation from intial_feature_coverage and feature_coverage.

### DIFF
--- a/src/local/butler/scripts/setup.py
+++ b/src/local/butler/scripts/setup.py
@@ -142,7 +142,7 @@ _CORPUS_SIZE as corpus_size,
 avg(t.average_exec_per_sec) as avg_exec_per_sec,
 avg(t.fuzzing_time_percent) as fuzzing_time_percent,
 sum(t.new_units_added) as new_tests_added,
-sum(t.merge_new_features) as new_cov_features,
+sum(t.new_cov_features) as new_cov_features,
 avg(t.crash_count*100) as regular_crash_percent,
 avg(t.oom_count*100) as oom_percent,
 avg(t.leak_count*100) as leak_percent,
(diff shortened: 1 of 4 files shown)


src/local/butler/scripts/setup.py
```diff
@@ -142,7 +142,7 @@ def __init__(self):
 avg(t.average_exec_per_sec) as avg_exec_per_sec,
 avg(t.fuzzing_time_percent) as fuzzing_time_percent,
 sum(t.new_units_added) as new_tests_added,
-sum(t.merge_new_features) as new_cov_features,
+sum(t.new_cov_features) as new_cov_features,
```
jonathanmetzman: maybe just "features" cov is a bit redundant

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
```diff
@@ -377,9 +377,9 @@ def test_basic_fuzz(self, mock_stdout):
                   80,
               'merge_edge_coverage':
                   1769,
```
jonathanmetzman: I say this nearly every time someone changes a stat but these tests really need to be refactored so that adding/changing a stat doesn't require fixing 10 tests.

src/python/bot/fuzzers/libFuzzer/stats.py
```diff
@@ -178,8 +176,8 @@ def parse_performance_features(log_lines, strategies, arguments):
       'max_len': 0,
       'manual_dict_size': 0,
       'merge_edge_coverage': 0,
-      'merge_new_files': 0,
-      'merge_new_features': 0,
+      'new_cov_edges': 0,
```
jonathanmetzman: Actually, defaulting to zero is problematic right?

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
inferno-chromium: yes need to think more.

src/local/butler/scripts/setup.py
inferno-chromium: sure, will do.

src/python/bot/fuzzers/libFuzzer/stats.py
inferno-chromium: thought more on -1. we can't do -1 since fuzzer-stats shows stuff by "sum(t.new_cov_features) as new_cov_features", so it need some support for exclusions format. And once we support exclusions format, we can just exclude by corpus_subset strategy name.

inferno-chromium:
@mbarbella-chromium as fyi to use new metric names for graphs

PR #395, opened 2019-04-22
Calculate new_features and new_edges metrics only after INITED.

inferno-chromium:
This happened after
https://github.com/google/clusterfuzz/commit/2f0b0ab5c6efa5c0614ccea8ce63f0f9264ff89f
when merge_new_features metric was removed and we started calculating
new_features and new_edges metrics from start and end of fuzzing run.
The bug was that we were setting edge_coverage and feature_coverage
metrics before INITED or while seed corpus is processed. So, if we
crashed right after INITED, we will have a negative value for
new_features and new_edges.


src/python/bot/fuzzers/libFuzzer/stats.py
```diff
@@ -291,10 +296,10 @@ def parse_performance_features(log_lines, strategies, arguments):
 
   # new_cov_* is a reliable metric when corpus subset strategy is not used.
   if not stats['strategy_corpus_subset']:
-    if 'initial_edge_coverage' in stats and 'edge_coverage' in stats:
```
inferno-chromium: This was incorrect as these metrics were inited to 0. so, this was always true, instead we should check if end stats metrics is non-zero.

src/python/tests/core/bot/fuzzers/libFuzzer/launcher_test_data/oom_in_seed_corpus.txt
```diff
@@ -0,0 +1,127 @@
...
+INFO: seed corpus: files: 19330 min: 1b max: 1048575b total: 26242064b rss: 1150Mb
+#128	pulse  cov: 788 ft: 1541 corp: 92/157b exec/s: 64 rss: 1153Mb
+#256	pulse  cov: 948 ft: 2285 corp: 182/404b exec/s: 64 rss: 1156Mb
+#512	pulse  cov: 1204 ft: 3767 corp: 348/998b exec/s: 73 rss: 1161Mb
+#1024	pulse  cov: 1386 ft: 5453 corp: 697/2646b exec/s: 73 rss: 1179Mb
+#2048	pulse  cov: 1632 ft: 8051 corp: 1342/7769b exec/s: 73 rss: 1231Mb
+#4096	pulse  cov: 3183 ft: 14143 corp: 2253/21Kb exec/s: 80 rss: 1326Mb
+#8192	pulse  cov: 3637 ft: 20165 corp: 4125/63Kb exec/s: 80 rss: 1594Mb
+#16384	pulse  cov: 3894 ft: 28749 corp: 7462/307Kb exec/s: 79 rss: 1881Mb
+#19333	INITED cov: 3912 ft: 31745 corp: 8755/8252Kb exec/s: 74 rss: 2053Mb
```
inferno-chromium: Before fix, we were doing

edge_coverage=3894
initial_edge_coverage=3912
new_edges = 3894-3912= negative.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
```diff
@@ -1023,9 +1075,9 @@ def test_parse_log_and_stats_timeout(self):
         'corpus_size': 0,
         'crash_count': 0,
         'dict_used': 0,
-        'edge_coverage': 1321,
+        'edge_coverage': 0,
```
inferno-chromium: they look valid changes as they crashed during seed corpus processing. i dont think we should have coverage metrics if fuzzing didn't start. wdyt @Dor1s @jonathanmetzman

src/python/bot/fuzzers/libFuzzer/stats.py
```diff
@@ -261,15 +262,19 @@ def parse_performance_features(log_lines, strategies, arguments):
 
     match = LIBFUZZER_LOG_START_INITED_REGEX.match(line)
     if match:
-      stats['initial_edge_coverage'] = int(match.group(1))
-      stats['initial_feature_coverage'] = int(match.group(2))
+      stats['initial_edge_coverage'] = stats['edge_coverage'] = int(
+          match.group(1))
+      stats['initial_feature_coverage'] = stats['feature_coverage'] = int(
+          match.group(2))
+      start = True
```
jonathanmetzman: nit: `inited` or `libfuzzer_inited` seems more descriptive to me

src/python/bot/fuzzers/libFuzzer/stats.py
jonathanmetzman: Why do we need these `if`s at all? If the above code works shouldn't feature_coverage always be 0 if initial_feature_coverage is zero? Same for edge. Maybe change to an assert.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
jonathanmetzman: Makes sense

src/python/bot/fuzzers/libFuzzer/stats.py
inferno-chromium: Done

src/python/bot/fuzzers/libFuzzer/stats.py
inferno-chromium: Done

src/python/bot/fuzzers/libFuzzer/stats.py
```diff
@@ -291,12 +294,13 @@ def parse_performance_features(log_lines, strategies, arguments):
 
   # new_cov_* is a reliable metric when corpus subset strategy is not used.
```
Dor1s: Should we record these for all the strategies? Even without the subset it's not super reliable, as it doesn't account for corpus merging, i.e. it's affected by the unstable coverage. Plus, we can always filter out irrelevant runs using BigQuery, so we don't need to dismiss the stats based on the strategy, we should upload everything.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
Dor1s: Either way is fine, but this means we would see zero on the charts for e.g. super OOMy targets. Maybe we still want to see the coverage it is able to reach before crashing?

src/python/bot/fuzzers/libFuzzer/stats.py
```diff
@@ -176,8 +176,6 @@ def parse_performance_features(log_lines, strategies, arguments):
       'max_len': 0,
       'manual_dict_size': 0,
       'merge_edge_coverage': 0,
-      'new_edges': 0,
```
Dor1s: Why remove the default values? I suspect that we may end up with `null` values in BigQuery, do we want that?

inferno-chromium: Yes it helps to calculate averages correctly for graphs. we used to use similar strategy for corpus_rss_mb and i think it used to work.

Dor1s: Got it, thanks!

src/python/bot/fuzzers/libFuzzer/stats.py
inferno-chromium: just makes data studio stuff more complicated. see also reason below.

Dor1s: I'm afraid it's inevitable that we need to filter out certain values for different columns. It's just a price we have to pay for having so many stats and configurations :)

Dor1s: Chatted offline, the final argument was that we surface these values in Fuzzer Stats, and we don't filter by strategy there, which creates very confusing picture if we record the values from runs with corpus subset strategy.

PR #397, opened 2019-04-22
Handle IOErrors from get_device_path in reset_usb.

mbarbella-chromium:
PTAL. This will still cause an error if we run into the exception when trying to store the path as DEVICE_PATH (which I think is desired), but will fall back to using it if we hit the exception later.

### DIFF
--- a/src/python/platforms/android/adb.py
+++ b/src/python/platforms/android/adb.py
@@ -604,7 +604,12 @@ def reset_usb():
 
   # We need to get latest device path since it could be changed in reboots or
   # adb root restarts.
-  device_path = get_device_path()
+  try:
+    device_path = get_device_path()
+  except IOError:
+    # We may reach this state if the device is no longer available.
+    device_path = None
+
   if not device_path:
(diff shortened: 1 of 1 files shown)


src/python/platforms/android/adb.py
```diff
@@ -604,7 +604,12 @@ def reset_usb():
 
   # We need to get latest device path since it could be changed in reboots or
   # adb root restarts.
-  device_path = get_device_path()
+  try:
+    device_path = get_device_path()
+  except IOError:
```
inferno-chromium: This feels a little too generic, why not do the try-catch around line 580-581 and return None there ?

mbarbella-chromium: I'm fine with that if you'd prefer it. The main reason to do it here was to ensure that we see errors if we fail when we set the value for DEVICE_PATH, which could cause more trouble later.

inferno-chromium: This is fine.

PR #398, opened 2019-04-23
Fuchsia-on-Clusterfuzz integration test.

flowerhack:
Verifies that the Fuchsia launcher can run Fuchsia on QEMU and SSH into
it accordingly.

Also does some minor fixups throughout the Fuchsia code (does QEMU
teardown after fuzzing is finished, ensures portnum is a str and not an
int, etc)

A few specific questions for reviewers are asked inline.


src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_integration_test.py
```diff
@@ -662,3 +664,119 @@ def test_merge_reductions(self, mock_get_timeout):
     """Tests merging. Wrapper around _test_merge_reductions."""
     mock_get_timeout.return_value = get_fuzz_timeout(1.0)
     self._test_merge_reductions('minijail-merge')
+
+
+@test_utils.integration
+@test_utils.with_cloud_emulators('datastore')
+class TestLauncherZFuchsia(BaseLauncherTest):
```
flowerhack: This typo-looking nomenclature is actually a poor hack that I'd like guidance on fixing.

It turns out, due to (I believe) the way we're using environment variables, order implicitly matters for these integration tests--if Fuchsia tests run before Minijail tests, the Minijail tests will fail.

I *believe* this is because of OS_OVERRIDE and QUEUE_OVERRIDE getting set (a test that contains only those lines reproduces the bug; removing the lines at least allows Minijail to succeed).

The specific *manner* in which the failure manifests is that get_llvm_symbolizer_path() returns NoneType around libfuzzer.py:L553ish when you hit the Minijail tests.

I tried everything I could think of to "unset" these two env vars (del os.environ['variable'], os.environ.pop, environment.set_value('variable', '')), but nothing seemed to work (one example is still included in tearDown for this class)?

If anyone has any ideas for what might be happening here, let me know; I'd like to fix this a proper non-brittle way!

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_integration_test.py
```diff
@@ -662,3 +664,119 @@ def test_merge_reductions(self, mock_get_timeout):
...
+    """Tests running a single round of fuzzing on a Fuchsia target, using
+    'echo' in place of a fuzzing command."""
+
+    # Cannot call setUp from our class's setUp method, because we're using the
+    # cloud emulation decorator, which assumes the base class is
+    # unittest.TestCase, and recurses infinitely if this is not true (as in our
+    # case, with BaseLauncherTest as the base)
+    # TODO(flowerhack): Does it make more sense to factor out all the setUp
+    # into its own method, which will have to be specifically called by every
+    # class in this file?
```
flowerhack: This is a question for y'all; I'm neutral either way, since neither route is super-aesthetically-pleasing for me

flowerhack:
requesting review from @oliverchang and/or @jonathanmetzman (and add others if you feel that's more appropriate!)

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_integration_test.py
```diff
@@ -78,7 +81,6 @@ def setup_testcase_and_corpus(testcase, corpus, fuzz=False):
 def run_launcher(*args):
   """Run launcher.py."""
   string_io = StringIO.StringIO()
-
```
jonathanmetzman: Try not to make unnecessary changes please.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_integration_test.py
jonathanmetzman: Add `helpers.patch_environ()` (import `helpers` from `tests.test_libs`) to `setUp()`.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_integration_test.py
```diff
@@ -662,3 +664,119 @@ def test_merge_reductions(self, mock_get_timeout):
...
+  def setUp(self):
+    # Set up a Fuzzer.
+    data_types.Fuzzer(
+        revision=1,
+        additional_environment_string=
+        'FUCHSIA_RESOURCES_URL = gs://fuchsia-on-clusterfuzz-v2/*',
+        builtin=True,
+        differential=False,
+        file_size='builtin',
+        jobs=[u'libfuzzer_asan_test_fuzzer'],
```
jonathanmetzman: Do you need to make this string a unicode string? If not please remove the u (same for other strings in this CL)

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_integration_test.py
```diff
@@ -662,3 +664,119 @@ def test_merge_reductions(self, mock_get_timeout):
...
+        name='libFuzzer',
+        source='builtin',  # change to "test@example.com" if it acts up
```
jonathanmetzman: Maybe remove this comment?

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_integration_test.py
```diff
@@ -662,3 +664,119 @@ def test_merge_reductions(self, mock_get_timeout):
...
+        name='engine_asan',
+        environment_string=(
+            'LSAN = True\n'
+            'ADDITIONAL_ASAN_OPTIONS = quarantine_size_mb=64:strict_memcmp=1'
+            ':symbolize=0:fast_unwind_on_fatal=0'
+            ':allocator_release_to_os_interval_ms=500\n')).put()
+
+    environment.set_value('QUEUE_OVERRIDE', 'FUCHSIA')
+    environment.set_value('OS_OVERRIDE', 'FUCHSIA')
+    os.environ['FUCHSIA_RESOURCES_URL'] = 'gs://fuchsia-on-clusterfuzz-v2/*'
```
jonathanmetzman: What happened to v1?

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_integration_test.py
jonathanmetzman: I'm confused, why is there a need to call setUp manually?
If you want to share it across multiple classes maybe use a mixin?

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_integration_test.py
```diff
@@ -662,3 +664,119 @@ def test_merge_reductions(self, mock_get_timeout):
...
+    resources_dir = environment.get_value('RESOURCES_DIR')
+    if not resources_dir:
+      raise Exception('Could not find RESOURCES_DIR')
+    fuchsia_resources_dir = os.path.join(resources_dir, 'fuchsia')
+    pkey_path = os.path.join(fuchsia_resources_dir, '.ssh', 'pkey')
+    portnum = '56339'
+    environment.set_value('FUCHSIA_PKEY_PATH', pkey_path)
+    environment.set_value('FUCHSIA_PORTNUM', portnum)
+
+  def tearDown(self):
```
jonathanmetzman: This won't be necessary once we start patching the environment, delete this method.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_integration_test.py
flowerhack: perfect! thank you!

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_integration_test.py
flowerhack: this just happened to be the name of the cloud bucket i created for this project; v1 was more of a "testing ground" in practice.  i could rename the bucket to something else, if desired, but it's really just a UUID for where we're storing a copy of the SDK & friends for now.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_integration_test.py
flowerhack: Right, maybe my phrasing wasn't super clear:

* the with_cloud_emulators wrapper wraps a class, in this case TestLauncherFuchsia, and calls its setUp method

* but, it does not call any further parent's setUp methods! (e.g. BaseLauncherTest, which is what we want)

* and, if you try to call super(self).setUp within TestLauncherFuchsia, it winds up recursing infinitely, since you're just calling the decorator's setUp method

Do you prefer a mixin to simply being explicit when calling setUp while using this decorator?

(I moved this line into setUp but it's still the same line fyi)

flowerhack:
@jonathanmetzman think i've responded to everything for now...

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -336,10 +336,10 @@ def __init__(self, executable_path, default_args=None):
           'FUCHSIA_PKEY_PATH and/or FUCHSIA_PORTNUM was not set')
     # yapf: disable
     self.ssh_args = [
-        '-i', fuchsia_pkey_path,
+        '-i', str(fuchsia_pkey_path),
```
oliverchang: why is str() necessary here?

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1325,10 +1325,11 @@ def execute_task(fuzzer_name, job_type):
   if timeout_multiplier > 1:
     testcase_count /= timeout_multiplier
 
+  qemu_process = None
```
oliverchang: nit: this line is unnecessary. variables set in the if have function scope.

src/python/platforms/fuchsia/device.py
```diff
@@ -91,8 +91,7 @@ def qemu_setup():
 
   # Finally, launch QEMU.
   qemu_process = new_process.ProcessRunner(qemu_path, qemu_args)
-  # TODO(flowerhack): Implement teardown logic.
-  qemu_process.run()
```
oliverchang: why was this removed?

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_integration_test.py
```diff
@@ -662,3 +665,110 @@ def test_merge_reductions(self, mock_get_timeout):
...
+    environment.set_value('OS_OVERRIDE', 'FUCHSIA')
+    os.environ['FUCHSIA_RESOURCES_URL'] = 'gs://fuchsia-on-clusterfuzz-v2/*'
+    # set_bot_environment gives us access to RESOURCES_DIR
+    environment.set_bot_environment()
+    resources_dir = environment.get_value('RESOURCES_DIR')
+    if not resources_dir:
+      raise Exception('Could not find RESOURCES_DIR')
+    fuchsia_resources_dir = os.path.join(resources_dir, 'fuchsia')
+    pkey_path = os.path.join(fuchsia_resources_dir, '.ssh', 'pkey')
+    portnum = '56339'
```
oliverchang: nit: no reason for this variable. just use it directly in line 758

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_integration_test.py
```diff
@@ -662,3 +665,110 @@ def test_merge_reductions(self, mock_get_timeout):
...
+        name='engine_asan',
+        environment_string=(
+            'LSAN = True\n'
+            'ADDITIONAL_ASAN_OPTIONS = quarantine_size_mb=64:strict_memcmp=1'
+            ':symbolize=0:fast_unwind_on_fatal=0'
+            ':allocator_release_to_os_interval_ms=500\n')).put()
+
+    environment.set_value('QUEUE_OVERRIDE', 'FUCHSIA')
+    environment.set_value('OS_OVERRIDE', 'FUCHSIA')
+    os.environ['FUCHSIA_RESOURCES_URL'] = 'gs://fuchsia-on-clusterfuzz-v2/*'
```
oliverchang: nit: use environment.set_value here to be consistent.

src/python/bot/tasks/fuzz_task.py
flowerhack: but we need access to qemu_process again on L1482, right?

src/python/platforms/fuchsia/device.py
flowerhack: Because I implemented it :)  Added a clarifying comment to qemu_process--tl;dr, it's the responsibility of the caller to kill the process, and added the appropriate kills for each time this function's called

src/python/bot/fuzzers/libfuzzer.py
flowerhack: it's not oop

src/python/bot/tasks/fuzz_task.py
oliverchang: yeah, but it should still work, because it's only ever accessed when platform == 'FUCHSIA'

PR #400, opened 2019-04-23
Support AFL driver's close fd mask feature

jonathanmetzman:
Instruct the driver to close fd mask when specified in the libFuzzer options.


jonathanmetzman:
I've landed this even though we haven't uptaked the libFuzzer revision that can use this feature on CF.
I did this because 1. We use trunk on oss-fuzz so oss-fuzz benefits. 2. We will roll Chrome soon (hopefully) 3. There's no harm in having this feature if Chrome can't use it.

jonathanmetzman:
I've landed this even though we haven't uptaked the libFuzzer revision that can use this feature on CF.
I did this because 1. We use trunk on oss-fuzz so oss-fuzz benefits. 2. We will roll Chrome soon (hopefully) 3. There's no harm in having this feature if Chrome can't use it.

PR #401, opened 2019-04-23
Bump up fork mode frequency to 50% of runs

jonathanmetzman:
Fork mode seems to be stable and provide benefits, bump up frequency to 50%.


jonathanmetzman:
cc @kcc

jonathanmetzman:
The main problem with bumping up frequency is it will cause timeouts and OOMs to be reported even less frequency since we aren't reporting those or even surfacing their count in fork mode.

jonathanmetzman:
I think Marty had something.
Since I've enabled this, libFuzzer has found an average of 37 ASAN security bugs per month (including this month which isn't finished) which is up from an average of 31 while AFL has found 16, which is down from an average of 24.

I don't really know if this change is the cause or the number of testcases added is the cause.

jonathanmetzman:
@mbarbella-chromium do you have data on this?

jonathanmetzman:
> LGTM! Do you have any spreadsheet / chart showing performance difference for the fork strategy?

I did some basic queries in internal and found that fork runs are 16% more likely to report a crash (unique and not) than a non fork run. (crashes don't include corpus, startup, timeout or OOM)

Dor1s:
Nice, thanks for sharing!

PR #408, opened 2019-04-24
Android sample startup script improvements.

mbarbella-chromium:
* Run ClusterFuzz for multiple devices at once.
* Ensure that the device is connected and restart if ClusterFuzz quits unexpectedly.

### DIFF
--- a/configs/test/bot/setup/android.bash
+++ b/configs/test/bot/setup/android.bash
@@ -13,6 +13,25 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+function run_bot () {
+  bot_directory=$INSTALL_DIRECTORY/bots/$(echo $1 | sed s/:/-/)
+  mkdir -p $bot_directory
+  cp -r $INSTALL_DIRECTORY/clusterfuzz $bot_directory/clusterfuzz
+  echo "Created bot directory $bot_directory."
+
+  while true; do
+    if ! $("$ADB_PATH/adb" -s "$1" wait-for-device); then
+      echo "Device $1 is not connected."
+      exit 1
+    fi
+
+    echo "Running ClusterFuzz instance for bot $1."
+    OS_OVERRIDE="ANDROID" ANDROID_SERIAL="$1" PATH="$PATH" NFS_ROOT="$NFS_ROOT" GOOGLE_APPLICATION_CREDENTIALS="$GOOGLE_APPLICATION_CREDENTIALS" ROOT_DIR="$bot_directory/clusterfuzz" PYTHONPATH="$PYTHONPATH" GSUTIL_PATH="$GSUTIL_PATH" python $bot_directory/clusterfuzz/src/python/bot/startup/run.py
+
+    echo "ClusterFuzz instance for bot $1 quit unexpectedly. Waiting for device."
+  done
+}
+
 if [ -z "$CLOUD_PROJECT_ID" ]; then
   echo "\$CLOUD_PROJECT_ID is not set."
   exit 1
@@ -24,8 +43,7 @@ if [ -z "$GOOGLE_APPLICATION_CREDENTIALS" ]; then
 fi
 
 if [ -z "$ANDROID_SERIAL" ]; then
-  echo "\$ANDROID_SERIAL is not set."
-  exit 1
+  echo "\$No ANDROID_SERIAL set. Will automatically detect devices and start ClusterFuzz for each."
 fi
 
 NFS_ROOT=  # Fill in NFS information if available.
@@ -71,13 +89,7 @@ if [ ! -d "$INSTALL_DIRECTORY/$APPENGINE" ]; then
 fi
 
 echo "Installing ClusterFuzz package dependencies."
-pip install crcmod==1.7 psutil==5.4.7 pyOpenSSL==19.0.0
-
-echo "Ensuring device is connected."
-if [[ $("$ADB_PATH/adb" -s "$ANDROID_SERIAL" get-state) != "device" ]]; then
-  echo "Device $ANDROID_SERIAL is not connected."
-  exit 1
-fi
+python -m pip install crcmod==1.7 psutil==5.4.7 pyOpenSSL==19.0.0
 
 echo "Activating credentials with the Google Cloud SDK."
 $GSUTIL_PATH/gcloud auth activate-service-account --key-file=$GOOGLE_APPLICATION_CREDENTIALS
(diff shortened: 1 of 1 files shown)


mbarbella-chromium:
PTAL. This is a slight behavior change, but if we don't necessarily expect all devices to be connected at startup time I can try to better mimic the old behavior.

configs/test/bot/setup/android.bash
```diff
@@ -24,8 +43,7 @@ if [ -z "$GOOGLE_APPLICATION_CREDENTIALS" ]; then
 fi
 
 if [ -z "$ANDROID_SERIAL" ]; then
-  echo "\$ANDROID_SERIAL is not set."
-  exit 1
+  echo "\$No ANDROID_SERIAL set. Will automatically detect devices and start ClusterFuzz for each."
```
inferno-chromium: nit: remove $ from start

inferno-chromium: Why not move this block at line 114, does not need to be here.

configs/test/bot/setup/android.bash
```diff
@@ -98,7 +110,10 @@ rm -rf clusterfuzz
 $GSUTIL_PATH/gsutil cp gs://$DEPLOYMENT_BUCKET/linux.zip clusterfuzz-source.zip
 unzip -q clusterfuzz-source.zip
 
-echo "Running ClusterFuzz."
-OS_OVERRIDE="ANDROID" ANDROID_SERIAL="$ANDROID_SERIAL" PATH="$PATH" NFS_ROOT="$NFS_ROOT" GOOGLE_APPLICATION_CREDENTIALS="$GOOGLE_APPLICATION_CREDENTIALS" ROOT_DIR="$ROOT_DIR" PYTHONPATH="$PYTHONPATH" GSUTIL_PATH="$GSUTIL_PATH" python $ROOT_DIR/src/python/bot/startup/run.py &
-
-echo "Success!"
+if [ -z "$ANDROID_SERIAL" ]; then
+  for serial in `adb devices -l | awk -F' ' '{ print $3 }' | grep usb:`; do
```
inferno-chromium: since this is open-source version, i think it should rely on actual serial than the usb version.

configs/test/bot/setup/android.bash
```diff
@@ -98,7 +110,10 @@ rm -rf clusterfuzz
...
+    run_bot $serial &
+  done
+else
+  run_bot $ANDROID_SERIAL
```
inferno-chromium: missing & at end ?

configs/test/bot/setup/android.bash
```diff
@@ -13,6 +13,25 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+function run_bot () {
+  bot_directory=$INSTALL_DIRECTORY/bots/$(echo $1 | sed s/:/-/)
```
inferno-chromium: move $(echo $1 | sed s/:/-/) into a local or atleast put $1 into local and use throughout.

configs/test/bot/setup/android.bash
```diff
@@ -13,6 +13,25 @@
...
+  mkdir -p $bot_directory
+  cp -r $INSTALL_DIRECTORY/clusterfuzz $bot_directory/clusterfuzz
```
inferno-chromium: what if dir already existed, can we delete and recreate

configs/test/bot/setup/android.bash
```diff
@@ -13,6 +13,25 @@
...
+  echo "Created bot directory $bot_directory."
+
+  while true; do
+    if ! $("$ADB_PATH/adb" -s "$1" wait-for-device); then
```
inferno-chromium: ADB_PATH looks wrong initialized, it should be resources/platform/android dir of clusterfuzz.

configs/test/bot/setup/android.bash
```diff
@@ -13,6 +13,25 @@
...
+      echo "Device $1 is not connected."
+      exit 1
```
inferno-chromium: should we exit or wait?

mbarbella-chromium: Since the command above is wait-for-device I'm not actually sure if we can get into this state. I'm actually just going to remove it, since this only made sense in the older code it was based on.

configs/test/bot/setup/android.bash
mbarbella-chromium: Moved $1 into a local.

configs/test/bot/setup/android.bash
mbarbella-chromium: Done.

configs/test/bot/setup/android.bash
mbarbella-chromium: Whoops, moved that back to ANDROID_SERIAL.

mbarbella-chromium: Hmm, good point. Moved the echos to the existing blocks there.

configs/test/bot/setup/android.bash
mbarbella-chromium: I suppose that would be more resilient to device reconnects, but I don't like the idea of maintaining two different versions of this. It's probably not going to be much of a burden so making the change for now, and we can revisit it later if it causes any trouble down the road.

configs/test/bot/setup/android.bash
mbarbella-chromium: Good point. The script behavior should be consistent in both cases, even though it doesn't need to run in parallel here.

configs/test/bot/setup/android.bash
mbarbella-chromium: Good point. There's no reason to download platform tools in this script because of that, so that simplifies a few things. Done.

configs/test/bot/setup/android.bash
```diff
@@ -98,7 +101,11 @@ rm -rf clusterfuzz
 $GSUTIL_PATH/gsutil cp gs://$DEPLOYMENT_BUCKET/linux.zip clusterfuzz-source.zip
 unzip -q clusterfuzz-source.zip
 
-echo "Running ClusterFuzz."
-OS_OVERRIDE="ANDROID" ANDROID_SERIAL="$ANDROID_SERIAL" PATH="$PATH" NFS_ROOT="$NFS_ROOT" GOOGLE_APPLICATION_CREDENTIALS="$GOOGLE_APPLICATION_CREDENTIALS" ROOT_DIR="$ROOT_DIR" PYTHONPATH="$PYTHONPATH" GSUTIL_PATH="$GSUTIL_PATH" python $ROOT_DIR/src/python/bot/startup/run.py &
-
-echo "Success!"
+if [ -z "$ANDROID_SERIAL" ]; then
+  echo "No \$ANDROID_SERIAL set. Will automatically detect devices and start ClusterFuzz for each."
+  for serial in `$ADB_PATH/adb devices | awk -F' ' '{ print $1 }' | egrep -v '^(|List)$'`; do
```
inferno-chromium: $1 looks wrong, it is getting serial and not usb: version ?

mbarbella-chromium: I don't understand the issue here. This was meant to get the true serial instead of usb:<id> to address the previous comment:

> since this is open-source version, i think it should rely on actual serial than the usb version.

inferno-chromium: ah right, reread your comment, you mean to keep this.

mbarbella-chromium: Right, I'll leave it as-is for now since this is probably better for the general case. We can check in a modified version in our config repo for our startup script if needed, and if they diverge too much we can get rid of this change later.

Issue #406, opened 2019-04-23
Py2to3: Futurize stage 2


inferno-chromium:
As part of Python 3 migration (#183)

git ls-files "*.py" | xargs futurize -w -n --stage2 

obviously not as simple. need to be done in steps, and str and unicode fixes are the more tricky ones.

inferno-chromium:
Need to think about these

```
src/local/butler/generate_datastore_models.py:  if py_type == long:
src/python/platforms/android/device.py:  if isinstance(value, long):
```

inferno-chromium:
Remaining ones, probably for Q3 and need to do str-unicode ones more carefully.

libfuturize.fixes.fix_unicode_keep_u
libfuturize.fixes.fix_future_standard_library_urllib (maybe start using requests)
libfuturize.fixes.fix_future_standard_library
libfuturize.fixes.fix_object
libfuturize.fixes.fix_basestring
libfuturize.fixes.fix_future_builtins

inferno-chromium:
Remaining (libfuturize.fixes.fix_object does not seem to be required, except ndb_patcher which is going away):

libfuturize.fixes.fix_unicode_keep_u
libfuturize.fixes.fix_future_standard_library
libfuturize.fixes.fix_basestring
libfuturize.fixes.fix_future_builtins

oliverchang:
i'll tackle the unicode/str conversion

inferno-chromium:
For future, need to careful of any migration change in docker/ dir, example breakage due to https://github.com/google/clusterfuzz/commit/6d9f9e76a1bb37a8c2a5cdd451b14dec4e58dd4a#diff-f1db2c4f1d1b92aef6e74abff3119154 (my bad :() since we don't have future lib during start_host.py time. also got noticed when we updated host image.

inferno-chromium:
Just for tracking, we include everything now except libfuturize.fixes.fix_unicode_keep_u and libfuturize.fixes.fix_future_builtins (these need more work, will come later). minor exclusion is ndb_patcher which will go away soon.

To replicate / reapply fixes, do
git ls-files "*.py" | xargs futurize -w -n --stage2 -x libfuturize.fixes.fix_unicode_keep_u -x libfuturize.fixes.fix_future_builtins

oliverchang:
All the futurize fixes were completed.

PR #411, opened 2019-04-26
Part 1: Py2->3 Futurize stage 2 without str-unicode, urllib changes (#406)

inferno-chromium:


inferno-chromium:
This patch does

git ls-files "*.py" | xargs futurize -w -n --stage2 -x libfuturize.fixes.fix_unicode_keep_u -x libfuturize.fixes.fix_future_standard_library_urllib -x libfuturize.fixes.fix_future_standard_library -x libfuturize.fixes.fix_object -x libfuturize.fixes.fix_basestring -x libfuturize.fixes.fix_future_builtins

excludes some things that will take more time and will go in Q3 i think.

inferno-chromium:
had to fix up a lot of old_div after this futurize step, but i think good for cleanup.

src/appengine/handlers/cron/grouper.py
```diff
@@ -219,8 +219,8 @@ def group_testcases():
 
 def group_testcases_with_same_issues(testcase_map):
   """Group testcases that are associated with same underlying issue."""
-  for testcase_1_id, testcase_1 in testcase_map.iteritems():
-    for testcase_2_id, testcase_2 in testcase_map.iteritems():
+  for testcase_1_id, testcase_1 in list(testcase_map.items()):
```
oliverchang: I'm a little surprised by these changes, since the previous version was using generators and this change makes it always convert the whole thing to a list. I would've expected the change to omit the list() call so that under Python 3 the behaviour is the same as before.

inferno-chromium: https://stackoverflow.com/questions/17695456/why-does-python-3-need-dict-items-to-be-wrapped-with-list :( :( i can remove all the list() since it was just overcautionary but we probably have to rerun futurize sometime and need to find way so that it does not reconvert it again.

inferno-chromium: Probably will convert to option 2 in https://python-future.org/compatible_idioms.html#iterating-through-dict-keys-values-items

```
# Python 2 and 3: option 3
from future.utils import iteritems
# or
from six import iteritems

for (key, value) in iteritems(heights):
    ...
```

oliverchang: Using six sounds good! Converting everything to a list could potentially have issues because of extra memory etc.

PR #413, opened 2019-04-26
Automatically redirect to login page.

oliverchang:
Instead of showing the access denied page first.


oliverchang:
PTAL. not sure what to do about the misalignment of the login button with the page title, because the button itself is already centred. The title looks off because the centering on that does not take into account the hamburger menu icon to the very left.

oliverchang:
actually hold off on this one for a bit.. have a better fix for this.

oliverchang:
now ready for review!

PR #417, opened 2019-04-26
Fuchsia-on-Clusterfuzz integration test.

flowerhack:
Verifies that the Fuchsia launcher can run Fuchsia on QEMU and SSH into
it accordingly.

Also does some minor fixups throughout the Fuchsia code (does QEMU
teardown after fuzzing is finished, ensures portnum is a str and not an
int, etc)


flowerhack:
I have made this new PR for #398 as a workaround for issue #418 .

It is identical-ish to the previous PR, modulo some changes to actually fix the previous CI failure:

* we now dynamically pick a port after checking it's available, instead of just using a "likely" free port,
* we don't run via KVM because it doesn't seem to be available on GCB,
* we set a correct x86 CPU for the QEMU VM

I also added some louder logging to make debugging QEMU/SSH issues easier in the future.

@oliverchang @jonathanmetzman , and sorry for losing the comment history from the previous CL :(

src/python/platforms/fuchsia/device.py
```diff
@@ -74,25 +82,24 @@ def qemu_setup():
...
   ]
   # yapf: enable
 
-  # Fuzzing jobs that SSH into the QEMU VM need access to these env vars.
+  # Fuzzing jobs that SSH into the QEMU VM need access to this env var.
   environment.set_value('FUCHSIA_PKEY_PATH', pkey_path)
-  environment.set_value('FUCHSIA_PORTNUM', portnum)
 
   # Finally, launch QEMU.
+  print 'Running QEMU. Command: ' + qemu_path + ' ' + str(qemu_args)
```
oliverchang: nit: please use print() with `from __future__ import print_function`

oliverchang: (same for all other prints in your PR).

src/python/platforms/fuchsia/device.py
```diff
@@ -59,8 +63,12 @@ def qemu_setup():
   extend_fvm(fuchsia_resources_dir, drive_path)
   add_keys_to_zbi(fuchsia_resources_dir, initrd_path, fuchsia_zbi)
 
-  # TODO(flowerhack): Add a mechanism for choosing portnum dynamically.
-  portnum = '56339'
+  tcp = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
oliverchang: nit: add a short comment here explaining that this is for getting a free port?

src/python/platforms/fuchsia/device.py
```diff
@@ -74,25 +82,24 @@ def qemu_setup():
       '-monitor', 'none',
       '-append', '"kernel.serial=legacy TERM=dumb"',
       '-machine', 'q35',
-      '-enable-kvm',
       '-display', 'none',
-      '-cpu', 'host,migratable=no',
+      '-cpu', 'Haswell,+smap,-check,-fsgsbase',
```
oliverchang: Could you explain what this cpu option is doing? Is this what's turning off KVM if it's an unsupported CPU ?

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1327,10 +1327,11 @@ def execute_task(fuzzer_name, job_type):
   if timeout_multiplier > 1:
     testcase_count /= timeout_multiplier
 
+  qemu_process = None
```
oliverchang: nit: per previous comment, this is not needed.

oliverchang:
Also, in the future, I believe you can just do a rebase and force push to your PR branch

flowerhack:
think i've addressed all code review at this point.

Issue #399, opened 2019-04-23
Py2to3: Replace App Engine module with firebase auth


oliverchang:
As part of Python 3 migration (#183)

oliverchang:
@inferno-chromium just a FYI that I've started on this, so that we don't duplicate work :)

inferno-chromium:
This will be awesome, thanks @oliverchang for working on this. I think this will also help to get rid of remaining testbed instances (other than these below, there is memcache but that is already in plans to change to something else).

```
src/python/tests/appengine/handlers/base_handler_test.py:from google.appengine.ext import testbed
src/python/tests/appengine/handlers/base_handler_test.py:    self.testbed = testbed.Testbed()
src/python/tests/appengine/handlers/base_handler_test.py:    self.testbed.activate()
src/python/tests/appengine/handlers/base_handler_test.py:    self.testbed.init_user_stub()
src/python/tests/appengine/handlers/base_handler_test.py:    self.testbed.deactivate()
src/python/tests/appengine/handlers/fuzzer_stats_data/testFuzzer_1_fuzzer_None_by-job_2016-10-20_2016-10-21.txt:    "projectId": "testbed-test",
src/python/tests/appengine/handlers/fuzzer_stats_data/testFuzzer_None_by-fuzzer_2016-10-20_2016-10-22.txt:    "projectId": "testbed-test", 
src/python/tests/appengine/handlers/fuzzer_stats_data/testFuzzer_job_by-fuzzer_2016-10-20_2016-10-21.txt:    "projectId": "testbed-test", 
src/python/tests/appengine/handlers/issue_redirector_test.py:from google.appengine.ext import testbed
src/python/tests/appengine/handlers/issue_redirector_test.py:    self.testbed = testbed.Testbed()
src/python/tests/appengine/handlers/issue_redirector_test.py:    self.testbed.activate()
src/python/tests/appengine/handlers/issue_redirector_test.py:    self.testbed.init_user_stub()
src/python/tests/appengine/handlers/issue_redirector_test.py:    self.testbed.deactivate()
src/python/tests/appengine/libs/handler_test.py:from google.appengine.ext import testbed
src/python/tests/appengine/libs/handler_test.py:    self.testbed = testbed.Testbed()
src/python/tests/appengine/libs/handler_test.py:    self.testbed.activate()
src/python/tests/appengine/libs/handler_test.py:    self.testbed.init_user_stub()
[...]

PR #419, opened 2019-04-27
Fix firebase exception in butler create_config (#399)

inferno-chromium:


inferno-chromium:
Currently we exception on trying to get a bucket with firebase-api-key, when its purpose was just to replace it in project.yaml.

Issue #403, opened 2019-04-23
Py2to3: Remove dependency on appengine blobstore and app_identity sign apis


inferno-chromium:
As part of Python 3 migration (#183)

inferno-chromium:
@oliverchang - just putting this in your queue, no rush, but more for tracking. I don't know this code but can take this if you provide some pointers.

```
src/appengine/handlers/upload_testcase.py:from google.appengine.ext import blobstore
src/appengine/handlers/upload_testcase.py:from google.appengine.ext.webapp import blobstore_handlers
src/appengine/libs/gcs.py:from google.appengine.api import app_identity
```

PR #423, opened 2019-04-29
Replace app_identity.sign_blob with Cloud API call (#403).

oliverchang:


src/appengine/libs/gcs.py
```diff
@@ -35,6 +36,26 @@
...
+  service_account = 'projects/-/serviceAccounts/' + service_account_email()
+
+  response = iam.projects().serviceAccounts().signBlob(
+      name=service_account,
+      body={
+          'delegates': [],
+          'payload': base64.b64encode(data),
+      }).execute()
+
+  return base64.b64decode(response['signedBlob'])
```
inferno-chromium: Do we want to verify signedBlob in response Or response is not None Or better error checking.

src/appengine/libs/gcs.py
```diff
@@ -35,6 +36,26 @@
     ['url', 'bucket', 'key', 'google_access_id', 'policy', 'signature'])
 
 
+def service_account_email():
```
inferno-chromium: This could be moved to utils as a helper.

src/appengine/libs/gcs.py
inferno-chromium: Isn't the response have signature field - https://cloud.google.com/iam/reference/rest/v1/projects.serviceAccounts/signBlob ?

oliverchang: That one is the deprecated API :) (See top of the page).

https://cloud.google.com/iam/credentials/reference/rest/v1/projects.serviceAccounts/signBlob is the correct documentation.

inferno-chromium: My bad, ignore this then.

src/appengine/libs/gcs.py
oliverchang: Done.

src/appengine/libs/gcs.py
oliverchang: Wrapped this in a try

PR #425, opened 2019-04-29
Disable login and logout for local development (#399).

inferno-chromium:


inferno-chromium:
Basically disables login, logout and switch account since firebase api key does not work locally ?

src/appengine/handlers/corpora.py
```diff
@@ -25,9 +25,9 @@
 class Handler(base_handler.Handler):
   """Manage data bundles."""
 
+  @handler.unsupported_on_local_server
```
oliverchang: any reason for the order changes in this PR?

inferno-chromium: It only mattered for

```
class LogoutHandler(base_handler.Handler):
  """Log out handler."""	  """Log out handler."""


   @handler.unsupported_on_local_server
  @handler.require_csrf_token	  @handler.require_csrf_token
  @handler.get(handler.HTML)	  @handler.get(handler.HTML)
```
where wrong order would mean we show the csrf error first. so i thought to fix the others as well. it is just unnecessary work, when instead we should be bailing out early.

Issue #245, opened 2019-03-08
Make stack ignore lists completely customizable and all regex based


inferno-chromium:
https://github.com/google/clusterfuzz/blob/master/src/python/crash_analysis/stack_parsing/stack_analyzer.py#L246

inferno-chromium:
Also, https://github.com/google/clusterfuzz/commit/cdfe140489e85896a3e5f5747acc9ab3d0df7fdb for stack blacklist regexes.

inferno-chromium:
@oliverchang as fyi. if you don't like any var names, etc, feel free to rename (would need config change as well). i hate stack_clean_regex, so uploading another PR to completely remove it.

PR #428, opened 2019-04-30
Make stack frame ignore signatures customizable (#245)

inferno-chromium:


configs/test/project.yaml
```diff
@@ -46,6 +46,11 @@ firebase:
   # API key for Firebase (public).
   api_key: firebase-api-key
 
+stacktrace:
+  # Stack frames to ignore when determining the crash signature.
+  stack_frame_ignore_regexes:
+
```
mbarbella-chromium: Nit: should this just be one blank line?

PR #432, opened 2019-05-01
Remove dependency on blobstore (#403).

oliverchang:
Handle the upload ourselves instead of using the blobstore upload
handler.

### DIFF
--- a/src/appengine/handlers/upload_testcase.py
+++ b/src/appengine/handlers/upload_testcase.py
@@ -14,12 +14,10 @@
 """Handler that uploads a testcase"""
 
 import ast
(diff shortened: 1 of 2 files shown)


oliverchang:
Thanks! Note that this introduces a 32MB upload limit (App Engine limit) for uploads via this API endpoint. This keeps the implementation simple for us, without requiring changes in the clients. Will confirm with the relevant users of this to ensure that this seems enough before merging.

inferno-chromium:
Yes that sounds good for now. Once we move to Appengine Flex, we won't run into urlfetch and other limits i think.

PR #336, opened 2019-03-29
Add pyyaml to CI.

inferno-chromium:

PR #433, opened 2019-05-01
Use latest selenium, use chromedriver in headless mode.

inferno-chromium:


src/requirements.txt
```diff
@@ -17,4 +17,5 @@ pytz==2018.5
 PyYAML==5.1
 requests==2.21.0
 requests-toolbelt==0.9.1
+selenium==3.141.0
```
inferno-chromium: This move allows use to use webdriver for fuzzing as well, have something in plan.

docker/ci/Dockerfile
```diff
@@ -38,9 +38,10 @@ RUN echo "deb [arch=amd64] http://storage.googleapis.com/bazel-apt stable jdk1.8
 
 RUN npm install -g bower polymer-bundler
 
-# RUN curl -s https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - && \
-#     echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list && \
-#     apt-get update && apt-get install -y google-chrome-stable
+# Install latest Chrome stable, needed for chromedriver testing.
```
inferno-chromium: This move is to try reenabling js_unittest next in CI. we need to have google chrome stable there first.

docker/ci/requirements.txt
```diff
@@ -15,7 +15,6 @@ pyfakefs==3.4.3
 pylint==1.9.3
 python-mimeparse==1.5.2
 PyYAML==5.1
```
Dor1s: nit: do we need PyYAML in both places, here and in src/requirements.txt ?

inferno-chromium: I think this is needed for lint run
src/local/butler/lint.py:import yaml
It was breaking things, that is why explicitly got added in https://github.com/google/clusterfuzz/pull/336

inferno-chromium: Found it, yes butler lint runs before setup(setup runs butler bootstrap). and lint needs pyyaml
https://github.com/google/clusterfuzz/blob/master/cloudbuild.yaml#L31

Dor1s: Got it, thanks!

PR #435, opened 2019-05-02
Update Go bazel rules.

oliverchang:


oliverchang:
For some reason my local bazel stopped working with the old version that we had. Upgrading this fixed it.

oliverchang:
(TBR for simple upgrade).

PR #436, opened 2019-05-02
Use full description with crash state when updating bug.

inferno-chromium:


inferno-chromium:
e.g. see unclear comment in https://bugs.chromium.org/p/chromium/issues/detail?id=958689#c2

PR #437, opened 2019-05-02
Add response to bigquery message.

inferno-chromium:


inferno-chromium:
Somehow logging from Appengine is not capturing vars, so changing this to atleast debug the current bigger missing afl stat problem

inferno-chromium:
TBR to debug afl issue.

PR #439, opened 2019-05-03
Fix AFL stat upload to Bigquery.

inferno-chromium:
This is a regression from
https://github.com/google/clusterfuzz/commit/6d9f9e76a1bb37a8c2a5cdd451b14dec4e58dd4a
a // b is float when b is a float. Causes bigquery error - "Invalid schema update. Field average_exec_per_sec has changed type from INTEGER to FLOAT"


inferno-chromium:
TBR simple fix.

PR #443, opened 2019-05-03
Py2to3: Futurize stage 2: libfuturize.fixes.fix_object (#406)

inferno-chromium:


inferno-chromium:
TBR simple migration step. ndb_patcher is skipped, need to chat on that and also later going away.

oliverchang:
LGTM

PR #446, opened 2019-05-05
Make messages in regression and progression task more informative.

inferno-chromium:
Also, make tooltip for 0:X revisions more informative.

See bug:
https://bugs.chromium.org/p/chromium/issues/detail?id=959285


inferno-chromium:
Minor logging fix, TBR.

oliverchang:
LGTM

PR #448, opened 2019-05-06
Py2to3: Futurize stage 2: Use six.integer_types instead of int, long (#406)

inferno-chromium:


inferno-chromium:
The regenerate was empty, already tried. i think it is because there is no int array field.

oliverchang:
Great, thanks for checking!

PR #452, opened 2019-05-09
Use ADB instead of ADB_PATH, matches upstream asan_device_setup

inferno-chromium:
This configuration is unsupported and is currently
causing confusion when people set it incorrectly as
a directory.


inferno-chromium:
TBR to fix a breakage.

PR #454, opened 2019-05-10
Make decision of new units by relying on new tests dir size.

inferno-chromium:
new_units_added metric is not always available, esp in fork mode.

### DIFF
--- a/src/python/bot/fuzzers/libFuzzer/launcher.py
+++ b/src/python/bot/fuzzers/libFuzzer/launcher.py
@@ -441,7 +441,7 @@ def get_corpus_directories(main_corpus_directory,
   if minijail_chroot:
     bind_corpus_dirs(minijail_chroot, corpus_directories)
 
-  return corpus_directories
+  return corpus_directories, new_testcases_directory
 
 
 def generate_new_testcase_mutations_using_radamsa(
@@ -859,7 +859,7 @@ def main(argv):
   fuzz_timeout = get_fuzz_timeout(is_mutations_run)
 
   # Get list of corpus directories.
-  corpus_directories = get_corpus_directories(
+  corpus_directories, new_testcases_directory = get_corpus_directories(
       corpus_directory, fuzzer_path, fuzzing_strategies, minijail_chroot)
 
   # Bind corpus directories in minijail.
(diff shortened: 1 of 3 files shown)


src/python/bot/fuzzers/libFuzzer/launcher.py
```diff
@@ -980,7 +980,9 @@ def main(argv):
 
   # Make a decision on whether merge step is needed at all. If there are no
   # new units added by libFuzzer run, then no need to do merge at all.
-  new_units_added = parsed_stats.get('new_units_added', 0)
+  new_units_added = (
+      shell.get_directory_file_count(new_testcases_directory) or
+      parsed_stats.get('new_units_added'))  # redundant, but needed for tests.
```
jonathanmetzman: Can you change the comment to a TODO to mock `shell.get_directory_file_count`. I think this is fine as a temporary solution but seems weird to have long term unless absolutely necessary.

jonathanmetzman:
LGTM

src/python/bot/fuzzers/libFuzzer/launcher.py
oliverchang: +1. mocking is definitely better here.

inferno-chromium: fixed.
minijail test was broken since self.command[0] is not _fuzzer there but minijail bin path.
also test_oom test ends up showing case where merge units=0, so we have a way to emulate that using None in mock_popen, so used that.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
```diff
@@ -1190,8 +1190,11 @@ def test_oom_crash(self, mock_stdout):
     self.fs.CreateDirectory('/fake/corpus_oom')
     os.environ['FUZZ_CORPUS_DIR'] = '/fake/corpus_oom'
 
-    with mock.patch('subprocess.Popen',
-                    create_mock_popen(self.oom_output)) as mock_popen:
+    new_units_added = 11
+    with mock.patch(
+        'subprocess.Popen',
+        create_mock_popen(self.oom_output, '/fake/inputs-disk/temp-1337/new',
+                          None, new_units_added)) as mock_popen:
```
inferno-chromium: None is used to show

              'new_units_added':
                  0,	                  0,

which just means we tried merge (since new_testcase_directory has units), but then running with -merge=1 failed, so units=0.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
```diff
@@ -100,7 +100,7 @@ def _write_fake_units(self):
 
     def communicate(self, input_data=None):
       """Mock subprocess.Popen.communicate."""
-      if self.command[0].endswith('_fuzzer'):
+      if '/fake/build_dir/fake_fuzzer' in self.command:
```
inferno-chromium: This functionality was already broken for 2 minijail tests. since command[0] is minijail bin.

PR #455, opened 2019-05-13
Use ':' for fuzz target project qualified name separator if project name is a path

inferno-chromium:
path.


src/python/datastore/data_types.py
```diff
@@ -1063,7 +1063,8 @@ def fuzz_target_project_qualified_name(project, binary):
     # Don't prefix with project name if it's the default project.
     return binary
 
-  project_prefix = project + '_'
+  seperator = '/' if project.startswith('/') else '_'
```
oliverchang: nit: correct spelling is "separator"

inferno-chromium: Done!

PR #456, opened 2019-05-13
Revert "Use ':' for fuzz target project qualified name separator if project name is a path"

inferno-chromium:
Reverts google/clusterfuzz#455


inferno-chromium:
Wont work, due to gcs paths.

inferno-chromium:
need to think more here.

PR #457, opened 2019-05-14
 Normalize project name for special chars

inferno-chromium:


inferno-chromium:
This keeps the project name as-is in UI (like search box on homepage, project name on testcase detail), but normalizes it for fuzz target name since it is used in local and gcs paths (':' breaks windows, '/' breaks everything). this will show up in fuzzer stats.

PR #459, opened 2019-05-15
Cron: fix schedule_corpus_pruning to work for CUSTOM_BINARY jobs too.

Dor1s:
Bug: https://bugs.chromium.org/p/chromium/issues/detail?id=963221

### DIFF
--- a/src/appengine/handlers/cron/schedule_corpus_pruning.py
+++ b/src/appengine/handlers/cron/schedule_corpus_pruning.py
@@ -43,8 +43,8 @@ def _get_latest_job_revision(job):
 
 def get_tasks_to_schedule():
   """Return a list of (task_target, job_name, queue_name) tuples."""
-  tasks = []
- for job in data_types.Job.query():
+  tasks_to_schedule = []
+  for job in data_types.Job.query():
     if not utils.string_is_true(job.get_environment().get('CORPUS_PRUNE')):
       continue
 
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
Maybe add a test to ensure this too when/if you get a chance.

Dor1s:
Please review the tests then :)

src/appengine/handlers/cron/schedule_corpus_pruning.py
```diff
@@ -41,24 +41,37 @@ def _get_latest_job_revision(job):
...
+
+    if utils.string_is_true(job.get_environment().get('CUSTOM_BINARY')):
+      # Custom binary jobs do not have revisions.
+      latest_revision = None
+    else:
+      latest_revision = _get_latest_job_revision(job)
+      if not latest_revision:
+        continue
+
+    queue_name = tasks.queue_for_job(job.name)
```
jonathanmetzman: Why change the variable name from queue to queue_name?

Dor1s: Found myself confused by `queue`, what type that variable was. Don't have strong opinion here though. Also unsure about `tasks_to_schedule`, maybe `tasks_arguments` or `parameters` or just `result` :)

src/appengine/handlers/cron/schedule_corpus_pruning.py
```diff
@@ -41,24 +41,37 @@ def _get_latest_job_revision(job):
...
+    for target_job in fuzz_target_utils.get_fuzz_target_jobs(job=job.name):
+      task_target = target_job.fuzz_target_name
+      if latest_revision:
+        task_target += '@%s' % latest_revision
+
+      tasks_to_schedule.append((task_target, job.name, queue_name))
```
oliverchang: nit: no need for the list and appends. just use `yield`

Issue #464, opened 2019-05-16
Flaky testcase might not get ever closed due to unrelated stack


inferno-chromium:
See https://clusterfuzz.com/testcase-detail/6294814012866560. this is now crashing on an unrelated lsan stack and might never get closed.

inferno-chromium:
Also, if they are marked as duplicate of another issue, they keep updating that issue saying that it still reproduces on tip-of-tree trunk, creating even more confusion.

PR #466, opened 2019-05-17
Account for security flag in flaky stack, similar to minimize task.

inferno-chromium:
This fixes https://github.com/google/clusterfuzz/issues/464


inferno-chromium:
See https://github.com/google/clusterfuzz/blob/master/src/python/bot/tasks/minimize_task.py#L828 for minimize task.

PR #468, opened 2019-05-17
[oss-fuzz][i386] Enable i386 jobs where appropriate in OSS-Fuzz

jonathanmetzman:


src/python/tests/appengine/handlers/cron/oss_fuzz_setup_test.py
```diff
@@ -316,6 +317,26 @@ def test_execute(self):
...
         'ISSUE_VIEW_RESTRICTIONS = none\n')
 
+    job = data_types.Job.query(
+        data_types.Job.name == 'libfuzzer_asan_i386_lib3').get()
+    self.assertIsNotNone(job)
+    self.assertEqual(job.platform, 'LIB3_LINUX')
+    self.assertItemsEqual(job.templates, ['asan', 'libfuzzer'])
+    self.assertEqual(
+        job.environment_string, 'RELEASE_BUILD_BUCKET_PATH = '
+        'gs://clusterfuzz-builds-i386/lib3/lib3-address-([0-9]+).zip\n'
```
inferno-chromium: Should lib3-address-([0-9]+).zip have i386 in there. this might be needed since if someone downloads file it shouldn't conflict with existing one and makes it easier to see which build type is it. this kind of change will be hard to do later. so @oliverchang @jonathanmetzman wdyt ?

src/python/tests/appengine/handlers/cron/oss_fuzz_setup_test.py
```diff
@@ -316,6 +317,26 @@ def test_execute(self):
...
+        'FUZZ_LOGS_BUCKET = lib3-logs.clusterfuzz-external.appspot.com\n'
+        'CORPUS_BUCKET = lib3-corpus.clusterfuzz-external.appspot.com\n'
```
inferno-chromium: we are sharing the corpus with x86_64, that is fine ?

jonathanmetzman: Yep.

jonathanmetzman: I probably need to ensure pruning is off for i386 though

src/python/tests/appengine/handlers/cron/oss_fuzz_setup_test.py
jonathanmetzman: I'm not sure but adding this might add a lot of ugly special casing. We don't do it for AFL builds btw.

docker/oss-fuzz/worker/Dockerfile
```diff
@@ -13,4 +13,5 @@
 # limitations under the License.
 FROM gcr.io/clusterfuzz-images/oss-fuzz/base
 
+RUN apt-get install -y libc6-i386 lib32gcc1
```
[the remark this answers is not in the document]
jonathanmetzman: Agreed offline not to do this.

src/python/tests/appengine/handlers/cron/oss_fuzz_setup_test.py
oliverchang: sgtm

src/python/tests/appengine/handlers/cron/oss_fuzz_setup_test.py
jonathanmetzman: Did this.

PR #471, opened 2019-05-21
Add randomized fuzzer selection to Fuchsia runs.

flowerhack:
Apologies for the length of the CL; I promise it's shorter than it looks
(though I can chop it up further).

Also apologies for the delay; there was a Fuchsia hack-it week, and some
instability/thrashing on the Fuchsia side of things.

This PR does the following:

* Adds the fuchsia_utils directory, which contains a (hopefully) stable
interface for working with Fuchsia. This includes logic for extracting
Fuzzers from a Fuchsia build.

* Moves the Fuchsia build logic into build_manager.  While it's a very
tiny class right now, and doesn't take much advantage of build_manager
features, this should allow it to e.g. use the weighted random selection
in future CLs, handle revision # logic more gracefully, etc.

* Adds the ability to *choose* a fuzzer at random, and run it.  (Left
for a future CL is extracting the crash log from the device and placing
it someplace useful, though *writing* the log should be handled
automagically by our wrapper class.)

My only question is the license headers: I added the standard Google LLC
license header to the files we're yoinking from Fuchsia, but *also* left
the Fuchsia license header in place.  I do not know whether we can just
merge these, if there's an elegant way to handle it, etc--I would've
just left the files totally intact, but the linter complained about the
header not looking right :)  So yeah, let me know.


flowerhack:
@oliverchang @nopsledder

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -362,7 +366,15 @@ def fuzz(self,
            additional_args=None,
            extra_env=None):
     """LibFuzzerCommon.fuzz override."""
-    return self._test_qemu_ssh()
+    self._test_qemu_ssh()
+    # Doing a timeout, to work around process-exit issue?
+    self.fuzzer.run([])
```
oliverchang: could you clarify what the issue is here?

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -334,19 +337,20 @@ class FuchsiaQemuLibFuzzerRunner(new_process.ProcessRunner, LibFuzzerCommon):
...
-        '-p', str(fuchsia_portnum),
-        'localhost'
-    ]
-    # yapf: enable
+          ('FUCHSIA_PKEY_PATH, FUCHSIA_PORTNUM, or FUCHSIA_RESOURCES_DIR was '
+           'not set'))
+    self.host = Host.from_dir(
+        os.path.join(fuchsia_resources_dir, 'build', 'out', 'default'))
+    self.device = Device(self.host, 'localhost', fuchsia_portnum)
+    pkg, tgt = environment.get_value('FUZZ_TARGET').split('/')
```
oliverchang: s/pkg/package/ and s/tgt/target/

oliverchang: Also, what's the format of "FUZZ_TARGET" ? I'm assuming `package_name/binary_name` ?

@inferno-chromium Seems like another case where we need to support `/` in a fuzz target name. Seems like we need to normalize the binary name too (and not just project).

src/python/fuchsia_util/__init__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2019 Google LLC
```
oliverchang: For discussion:

It seems better to include these scripts with the build, rather than copying them to CF as-is since the Fuchsia SDK is unstable. That way, for each build we should be guaranteed to have a set of scripts that work. This is important for regression task etc where we do bisects.

We're also in the midst of doing PY2->PY3, and copying things here may complicate things. Do these scripts work with PY3 already?

If we go with that, we also can't import the scripts as-is, because importing is fairly permanent and we shouldn't try to reimport if we set up a build for another revision. We may need something like a stable CLI interface here for the necessary functionality.

src/python/bot/fuzzers/libfuzzer.py
flowerhack: ah! that issue's obsolete; was only an issue while debugging. removed comment

src/python/bot/fuzzers/libfuzzer.py
flowerhack: Yep, that's the expected format of FUZZ_TARGET.

Completed s/pkg/package/ and s/tgt/target/.

(Leaving comment open--I don't have context on the question of "supporting slashes in target names" but I assume @inferno-chromium may need to peek :) )

src/python/fuchsia_util/__init__.py
flowerhack: Discussed during meeting today.

Consensus seems to be:

* we're hard-forking these scripts from Fuchsia, feel free to request changes s.t. they meet Fuchsia's style guide

* we agreed that the Fuchsia team will be responsible for validating that "whatever is placed in the build bucket" is run-able via these scripts

* we accept the risk that, should we need to change/modify these scripts (which we're hoping is rather unlikely!), performing bisects for older builds will become either impossible or require a cumbersome hack

Finally, I *think* these scripts work with Python 3, but I'm not sure how to test such a thing.  Guidance would be very appreciated!  :D

oliverchang: >we accept the risk that, should we need to change/modify these scripts (which we're hoping is rather unlikely!), performing bisects for older builds will become either impossible or require a cumbersome hack

We also brought up the idea of having a minimal "input pusher" utility with every build for this purpose, since that's the part that is most likely to change. Did we reach a decision on whether or not to do this? @nopsledder

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -334,19 +337,21 @@ class FuchsiaQemuLibFuzzerRunner(new_process.ProcessRunner, LibFuzzerCommon):
...
-        '-o', 'StrictHostKeyChecking no',
-        '-o', 'UserKnownHostsFile=/dev/null',
-        '-p', str(fuchsia_portnum),
-        'localhost'
-    ]
-    # yapf: enable
+          ('FUCHSIA_PKEY_PATH, FUCHSIA_PORTNUM, or FUCHSIA_RESOURCES_DIR was '
+           'not set'))
+    self.host = Host.from_dir(
+        os.path.join(fuchsia_resources_dir, 'build', 'out', 'default'))
```
oliverchang: nit: put os.path.join('build', 'out', 'default') into a class CONSTANT.

src/python/build_management/build_manager.py
```diff
@@ -711,6 +713,57 @@ def setup(self):
...
+    host = Host.from_dir(
+        os.path.join(fuchsia_resources_dir, 'build', 'out', 'default'))
+    fuzz_targets = Fuzzer.filter(host.fuzzers, '')
+    # TODO(flowerhack): Get list of fuzz targets and pass to the actual
+    # randomizer, instead of using random.choice().
+    fuzz_target = random.choice(fuzz_targets)
+
+    environment.set_value('FUZZ_TARGET', str(fuzz_target))
+    logs.log(
+        str('Extracted fuzz target ' + fuzz_target[0] + '/' + fuzz_target[1]))
```
oliverchang: I'm a bit confused by this line. Is this just taking the first and second characters of `fuzz_target`?

src/python/build_management/build_manager.py
```diff
@@ -711,6 +713,57 @@ def setup(self):
...
+    fuchsia_resources_dir = fuchsia.device.initialize_resources_dir()
+    environment.set_value('FUCHSIA_RESOURCES_DIR', fuchsia_resources_dir)
+
+    logs.log('Retrieved build r%d.' % self.revision)
+    logs.log('Extracting fuzz targets.' + fuchsia_resources_dir)
+    logs.log("Extracting fuzz targets." + fuchsia_resources_dir)
+    environment.set_value('FUCHSIA_DIR',
+                          os.path.join(fuchsia_resources_dir, 'build'))
+
+    symbolize_path = os.path.join(fuchsia_resources_dir, 'build', 'zircon',
```
oliverchang: put all these path components as a separate class constant like *_PATH = os.path.join(..., ...) and join them here instead.

src/python/build_management/build_manager.py
```diff
@@ -1118,7 +1171,18 @@ def setup_regular_build(revision):
       target_weights=target_weights)
   if build.setup():
     return build
+  return None
 
+
+def setup_fuchsia_build():
+  """Sets up Fuchsia build."""
+  base_build_dir = ''
```
oliverchang: nit: add TODO here to use real values.

src/python/build_management/build_manager.py
```diff
@@ -711,6 +713,57 @@ def setup(self):
...
+  @property
+  def build_dir(self):
+    return self._build_dir
+
+  def setup(self):
+    logs.log('Retrieving build %d.' % self.revision)
+
+    # Bucket for QEMU resources.
+    fuchsia_resources_dir = fuchsia.device.initialize_resources_dir()
+    environment.set_value('FUCHSIA_RESOURCES_DIR', fuchsia_resources_dir)
```
oliverchang: I think now that we're trying to get this to fit the concept of a build, we should be storing these in the build directory instead.

In fact, making this fit as a "RegularBuild" (e.g. by subclassing that) will be even  better, because it comes with all the benefits of a proper build on CF, and should just work when we have revision numbers for dealing with different builds. We'll have to do some minor refactoring to factor out the target selection functionality though, so that you can override it.

Regular builds are builds which expect the following format. 

```
gs://bucket/path/fuchsia-resources-([0-9.]+).zip
```

When you call setup_build these are all unpacked into a subdir of `BUILDS_DIR` with a consistent path.

Fine to leave for another CL, but please do leave a TODO.

src/python/build_management/build_manager.py
```diff
@@ -711,6 +713,57 @@ def setup(self):
...
+
+    logs.log('Retrieved build r%d.' % self.revision)
+    logs.log('Extracting fuzz targets.' + fuchsia_resources_dir)
+    logs.log("Extracting fuzz targets." + fuchsia_resources_dir)
```
oliverchang: this line is duplicated.

src/python/tests/core/build_management/build_manager_test.py
```diff
@@ -155,6 +156,29 @@ def test_setup_fail(self):
...
+
+  def setUp(self):
+    test_helpers.patch_environ(self)
+
+  def _assert_env_vars(self):
+    self.assertTrue(os.environ['FUZZ_TARGET'])
+
+  def test_setup(self):
+    """Tests setting up a build."""
+    self.tmp_resources_dir = tempfile.mkdtemp()
```
oliverchang: please do this in setUp/tearDown, and  make sure to delete the dir in tearDown.

src/python/build_management/build_manager.py
flowerhack: oh, whups.  it's retrieved as a tuple, but it makes more sense to coerce it to a string and *then* print things... fixed now

src/python/tests/core/build_management/build_manager_test.py
```diff
@@ -155,6 +157,36 @@ def test_setup_fail(self):
...
+
+  def setUp(self):
+    test_helpers.patch_environ(self)
+    self.tmp_resources_dir = tempfile.mkdtemp()
+
+  def tearDown(self):
+    try:
+      shutil.rmtree(self.tmp_resources_dir)
+    except OSError as exception:
+      if exception.errno != errno.ENOENT:  # file doesn't exist
```
oliverchang: what's this needed for?

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -328,25 +331,29 @@ def fuzz(self,
...
-        'localhost'
-    ]
-    # yapf: enable
+          ('FUCHSIA_PKEY_PATH, FUCHSIA_PORTNUM, or FUCHSIA_RESOURCES_DIR was '
+           'not set'))
+    self.host = Host.from_dir(
+        os.path.join(fuchsia_resources_dir, self.FUCHSIA_BUILD_REL_PATH))
+    self.device = Device(self.host, 'localhost', fuchsia_portnum)
+    # Fuchsia fuzzer names have the format {package_name}/{binary_name}.
+    package, target = environment.get_value('FUZZ_TARGET').split('/')
```
oliverchang: can you add a TODO here, and assign it to me to figure out how to handle the '/' in the binary name? as is, this won't work with a few things.

src/python/platforms/fuchsia/device.py
```diff
@@ -24,6 +24,9 @@
 from google_cloud_utils import gsutil
```
jonathanmetzman: If this file gets imported (directly or indirectly) by appengine code there will be issues.
Maybe you can stop them by doing something like [this](https://github.com/google/clusterfuzz/pull/471/files#diff-8703e56d42b3f1a98a675c86907e8b90L24)

jonathanmetzman: ie: import it in code that won't get run on appengine.

src/python/platforms/fuchsia/util/host.py
```diff
@@ -0,0 +1,194 @@
...
+    This class abstracts the details of various repository, tool, and build
+    paths, as well as details about the host architecture and platform.
+
+    Attributes:
+      fuzzers:   The fuzzer binaries available in the current Fuchsia image
+      build_dir: The build output directory, if present.
+  """
+
+  # Convenience file descriptor for silencing subprocess output
+  DEVNULL = open(os.devnull, 'w')
```
jonathanmetzman: This will be run on appengine if this file gets imported o appengine.

src/python/tests/core/build_management/build_manager_test.py
flowerhack: removed

oliverchang:
Everything seems to work on App Engine, merging.

PR #472, opened 2019-05-21
py2-3: more stage2 fixes (#406).

oliverchang:
Apply libfuturize.fixes.fix_future_standard_library and
libfuturize.fixes.fix_basestring.

### DIFF
--- a/src/python/bot/fuzzers/options.py
+++ b/src/python/bot/fuzzers/options.py
@@ -13,8 +13,10 @@
 # limitations under the License.
 """Fuzzer options."""
 
+from future import standard_library
+standard_library.install_aliases()
 from builtins import object
-import ConfigParser
+import configparser
 import os
 import random
 import re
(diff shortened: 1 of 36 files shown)


src/python/system/path_patcher.py
```diff
@@ -14,11 +14,15 @@
...
 
-import __builtin__
+from future import standard_library
+standard_library.install_aliases()
+import builtins
 import functools
 import os
 import sys
 
+import six
```
inferno-chromium: nit: move alongwith with other system imports. before "import sys"

oliverchang: Done.

src/python/tests/test_libs/test_utils.py
```diff
@@ -384,3 +387,14 @@ def decorator(func):  # pylint: disable=unused-argument
                                func)
 
   return decorator
+
+
+class MockStdout(io.BufferedWriter):
```
oliverchang: this was needed because the new io.StringIO and io.BytesIO are now strict about str vs unicode

src/python/system/path_patcher.py
```diff
@@ -34,10 +37,18 @@ def _patch_single(obj, attr, fn):
   setattr(obj, attr, fn)
   setattr(getattr(obj, attr), '__path_patcher__', True)
 
+  if obj == builtins and six.PY2:
```
oliverchang: this was needed because setattr(__builtins__, ...) no longer works after these fixes. On Python 3, setting builtins.<attr> does the same thing, so isn't needed there.

src/python/bot/fuzzers/options.py
```diff
@@ -88,11 +90,12 @@ def __init__(self, options_file_path, cwd=None):
     else:
       self._cwd = os.path.dirname(options_file_path)
 
-    self._config = ConfigParser.ConfigParser()
-    try:
-      self._config.read(options_file_path)
-    except ConfigParser.Error:
-      raise FuzzerOptionsException('Failed to parse fuzzer options file.')
+    self._config = configparser.ConfigParser()
+    with open(options_file_path, 'r') as f:
```
oliverchang: This was changed to an explicit file open to make pyFakeFS work, since python-future seems to patch something in a way such that it doesn't work with open() here anymore.

PR #481, opened 2019-05-22
Set default values from new_edges and new_features metrics.

inferno-chromium:


src/python/bot/fuzzers/libFuzzer/stats.py
```diff
@@ -176,6 +176,8 @@ def parse_performance_features(log_lines, strategies, arguments):
       'max_len': 0,
       'manual_dict_size': 0,
       'merge_edge_coverage': 0,
+      'new_edges': None,
```
oliverchang: let's actually set these to 0. Having these as "None" could break new projects.

inferno-chromium:
Discussed offline, schema autodetect does not work with None value. So, this will break new projects.

PR #484, opened 2019-05-23
Add delete job button to jobs page.

inferno-chromium:
This deletes the job and removes associations in Fuzzer and
FuzzerJob entities.

### DIFF
--- a/src/appengine/handlers/jobs.py
+++ b/src/appengine/handlers/jobs.py
@@ -227,3 +227,6 @@ class DeleteJobHandler(base_handler.Handler):
 
     # Delete job.
     job.key.delete()
+
+    helpers.log('Deleted job %s' % job.name, helpers.MODIFY_OPERATION)
+    self.redirect('/jobs')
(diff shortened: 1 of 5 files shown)


jonathanmetzman:
Didn't we not have this in the past because deleting a job can mess many things up?

inferno-chromium:
No, we didn't do the cleanup properly, now it does by removing fuzzer and fuzzerjob mappings. i skipped template deletion, because that is even more confusing as it can uses by jobs and i have no idea on how it impacts. deleting a job on the other hand should be ok with this cleanup.

src/appengine/handlers/jobs.py
```diff
@@ -41,7 +43,7 @@ def get_queues():
 
 
 class Handler(base_handler.Handler):
-  """Manage sets of environment variables for bots."""
+  """Return list of jobs and templates."""
```
jonathanmetzman: The new language seems a bit like it is for a comment instead of a class.

inferno-chromium: Done

PR #486, opened 2019-05-23
Remove README from Fuchsia resources dir.

flowerhack:
Downloading the Fuchsia resources, which is done as part of integration
testing, clobbers the placeholder README file, causing annoying diffs.

Since we attempt to create the resources/fuchsia directory anyway, let's
just rely on that to ensure our directory exists, rather than leaving a
placeholder file there.

PR #487, opened 2019-05-24
Fix exception due to resources/fuchsia removal

inferno-chromium:
Git does not allow empty dir, so need to create intermediates
bot/resources/fuchsia after
https://github.com/google/clusterfuzz/pull/486


inferno-chromium:
i confused CI failure, which was legit. basically removing bot/resources/fuchsia/README.md removes resources/fuchsia subdir completely. so needs to create intermediates.

inferno-chromium:
TBR, passes on gcb

flowerhack:
oh gosh, i had thought `recreate=True` would handle this case. my bad! thanks for the fix

PR #488, opened 2019-05-24
Fix OSS-Fuzz host image breakage.

inferno-chromium:
future package is not installed during docker setup. This
is not needed since range call is not expensive here, skip.


inferno-chromium:
TBR to fix image breakage asap for https://github.com/google/oss-fuzz/issues/2459

jonathanmetzman:
LGTM

PR #492, opened 2019-05-27
Also install future in Windows startup script.

oliverchang:


oliverchang:
Also, what's the proper way to install this on our macOS machines these days?

inferno-chromium:
actually why do we need in startup script ? we have this src/requirements.txt for bots. do we have other scripts like docker ones that need future (e.g. https://github.com/google/clusterfuzz/commit/4bf86226113c2e862edee42574b2cf39264d661f)

oliverchang:
This is just to make it safer to run the remainder of the fixers. For instance, I'm not even sure how https://github.com/google/clusterfuzz/blob/master/src/python/bot/startup/run_bot.py#L20 works right now, because it happens before the `modules.fix_module_search_paths()`, which is what adds `third_party` to `sys.path`.

inferno-chromium:
ah right, i think all these builtins import should be moved after fix_module_search_paths for safety and to work across mac, android bots.

Issue #496, opened 2019-05-28
Do run_server in integration tests


oliverchang:
title.

oliverchang:
This was done a long while back.

PR #500, opened 2019-05-29
Add an integration test for run_server (#496).

oliverchang:
Add `butler.py integration_tests` for this and future end-to-end
integration tests.


src/local/butler/run_server.py
```diff
@@ -200,7 +200,12 @@ def execute(args):
             datastore_emulator_port=constants.DATASTORE_EMULATOR_PORT,
             log_level=args.log_level,
             pubsub_emulator_host=constants.PUBSUB_EMULATOR_HOST,
-            local_gcs_server_host=constants.LOCAL_GCS_SERVER_HOST))
+            local_gcs_server_host=constants.LOCAL_GCS_SERVER_HOST),
+        extra_environments={
+            # PYTHONPATH on CI includes an outdated App Engine SDK, which has
+            # bugs in dev_appserver.py.
+            'PYTHONPATH': '',
```
inferno-chromium: Do we not need to have path set to ROOT_DIR/src atleast ?

src/local/butler/integration_tests.py
```diff
@@ -0,0 +1,65 @@
...
+    common.execute(
+        ('python butler.py run setup '
+         '--non-dry-run --local --config-dir={config_dir}'
+        ).format(config_dir=constants.TEST_CONFIG_DIR),
+        exit_on_error=False)
+
+    request = urllib.request.urlopen('http://' + constants.DEV_APPSERVER_HOST)
+    request.read()  # Raises exception on error
+  except Exception:
+    print('Logs:')
```
inferno-chromium: print('Error occurred:') in its own line ? or custom exception message in raise ?

src/local/butler/integration_tests.py
```diff
@@ -0,0 +1,65 @@
...
+        output_lines=lines)
+
+    # Sleep a small amount of time to ensure the server is definitely ready.
+    time.sleep(1)
+
+    # Call setup ourselves instead of passing --bootstrap since we have no idea
+    # when that finishes.
+    # TODO(ochang): Make bootstrap a separate butler command and just call that.
+    common.execute(
+        ('python butler.py run setup '
```
inferno-chromium: What happens if we hit around the same time run setup is called by run_server, we can have race condition corruption ?

oliverchang: setup is synchronous, so it will finish before we hit the server.

src/local/butler/run_server.py
oliverchang: nope, not necessary here (and this isn't set in dev environment either).

src/local/butler/integration_tests.py
oliverchang: Done!

Issue #462, opened 2019-05-16
Investigate the issues with corpus merge timing out


Dor1s:
#458 might help, but might not

The issue is that if at some point corpora for a target becomes too large, the merge / pruning can be always timing out since then. That means we'll be losing some efficiency and may not never resolve the issue.

inferno-chromium:
I think we have a mitigation for this in corpus pruning task. Maybe the gsutil rsync itself fails due to large corpus size, need to check after 2 runs.

```
  # If our last execution failed, shrink to a randomized corpus of usable size
  # to prevent corpus from growing unbounded and recurring failures when trying
  # to minimize it.
  if last_execution_failed:
    _limit_corpus_size(context.initial_corpus_path,
                       CORPUS_SIZE_LIMIT_FOR_FAILURES)
```

Dor1s:
Yeah, there might be different situations, e.g. `gsutil` timing out, and/or the target might be too slow and fails to merge even 10K inputs (as per the limit)?

I was thinking about another corpus pruning task (e.g. maybe weekly?) that would run on those problematic fuzz targets and allocate enough time to finish pruning for them.

Can't say that I like this idea, but that's what I'm doing locally from time to time when I have to fix such target.

inferno-chromium:
corpus pruning task gets like a 24 hour duration every day (so i dont see advantage with weekly run). maybe just need to increase the gsutil duration for sync on those ? i am worried about local disk filing up too.

Dor1s:
Oh, 24 hours is a lot! I forgot about that. Good point regarding `gsutil`, thanks! I should check stats + logs and see how many such targets we have and where they are failing.

Dor1s:
Have been chasing one fuzz target affected, but turned out I hit https://bugs.chromium.org/p/chromium/issues/detail?id=952995 again, i.e. corpus pruning succeeded and we have all the stats, but https://clusterfuzz.com/fuzzer-stats/by-fuzzer/fuzzer/libFuzzer/job/libfuzzer_chrome_asan doesn't show those.

I'll check more fuzz targets.

Dor1s:
Out of all Chromium fuzz targets, only 2 have problems with corpus pruning right now:

1. `grpc_uri_fuzzer_test` seems to be timing out, I'm still running it locally, will see how (if) it ends
2. `grpc_hpack_parser_fuzzer_test` quickly OOMs and then eventually ends up with a corrupted merge control file:

```
[...]

PR #508, opened 2019-05-30
Corpus pruning: do not fail the task if quarantine or shared corpus failed to sync (#462).

Dor1s:


src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -218,10 +218,11 @@ def sync_to_disk(self):
       raise CorpusPruningException('Failed to sync corpus to disk.')
 
     if not self.quarantine_corpus.rsync_to_disk(self.quarantine_corpus_path):
-      raise CorpusPruningException('Failed to sync quarantine corpus to disk.')
+      logs.log_error('Failed to sync quarantine corpus to disk.')
```
inferno-chromium: can you add ", fuzz_target=self.fuzz_target" to both places, makes it easier to debug :)

Dor1s: Hm, I feel like last time I did that, @oliverchang told me to remove, as it's auto added. Am I mistaken?

inferno-chromium: This is not needed in launchers since we add it in logs.configure,

see
  logs.configure(
      'run_fuzzer', {
          'fuzzer': fuzzer_name,
          'engine': 'libFuzzer',
          'job_name': environment.get_value('JOB_NAME')
      })

but this is corpus pruning task, so it wont have this fuzzer var set.

Dor1s: Ah, I see, TIL! Thank you!

Issue #504, opened 2019-05-30
Refactor issue management


oliverchang:
To make it easier to add new issue trackers.

inferno-chromium:
I think this is done, closing.

PR #510, opened 2019-05-31
Extend issue tracker interface to support actions (#504)

oliverchang:
Actions are modifications on an issue, e.g. a comment and/or modifying
labels.

### DIFF
--- a/src/python/issue_management/monorail/__init__.py
+++ b/src/python/issue_management/monorail/__init__.py
@@ -103,13 +103,60 @@ class Issue(issue_tracker.Issue):
     return self._monorail_issue.components
 
   @property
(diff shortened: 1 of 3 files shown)


src/python/issue_management/monorail/__init__.py
```diff
@@ -103,15 +103,62 @@ def components(self):
     return self._monorail_issue.components
 
   @property
-  def comments(self):
+  def actions(self):
     """Get the issue comments."""
```
inferno-chromium: nit: fix description

oliverchang: done.

PR #512, opened 2019-05-31
Clear build urls directory on OSS-Fuzz worker.

inferno-chromium:
When a task starts on OSS-Fuzz, only host's build urls dir
is cleared. However, we download the urls on worker and if
a stale list is present, we reuse it. So, new revisions result
in a "Error getting build url for job" error.


oliverchang:
thanks! LGTM

PR #513, opened 2019-05-31
Refactor android code.

inferno-chromium:


src/python/platforms/android/wifi.py
```diff
@@ -0,0 +1,35 @@
+# Copyright 2019 Google LLC
```
mbarbella-chromium: I don't see an issue with leaving this as its own file, but I think this could fit fairly cleanly in settings as well. It doesn't seem likely that we'd ever add more here.

inferno-chromium: might be moving configure_wifi_and_airplane_mode there, so keeping as-is.

PR #515, opened 2019-05-31
Refactor android code, fix exceptions.

inferno-chromium:


inferno-chromium:
TBR to fix exceptions
AttributeError: 'module' object has no attribute 'battery'

PR #516, opened 2019-06-01
Remove old unneeded chrome lib hacks.

inferno-chromium:
These files were deleted in cleanup, causing
error while loading shared libraries: /usr/lib/x86_64-linux-gnu/libgcrypt.so.11: invalid ELF header

these shouldn't be needed.


inferno-chromium:
TBR to fix docker images.

PR #521, opened 2019-06-04
Move cleanup and update_issue to new issue tracker iterface (#504).

oliverchang:
This required some changes to the interface, namely:

- Add a `LabelStore` class to encapsulate logic for adding/removing
 labels in a case insensitive way.
- Add a `new_comment` argument to save().
- Add a readonly `open` property to Issue interface.
- Add a `get_original_issue()` method to the IssueTracker interface.

### DIFF
--- a/src/appengine/handlers/cron/cleanup.py
+++ b/src/appengine/handlers/cron/cleanup.py
@@ -338,14 +338,14 @@ def get_issue_for_testcase(testcase):
   if not testcase.bug_information:
     return None
 
-  issue_tracker_manager = issue_tracker_utils.get_issue_tracker_manager(
-      testcase=testcase, use_cache=True)
-  if not issue_tracker_manager:
+  issue_tracker = issue_tracker_utils.get_issue_tracker_for_testcase(
+      testcase, use_cache=True)
+  if not issue_tracker:
     return None
 
   try:
-    issue_id = int(testcase.bug_information)
-    issue = issue_tracker_manager.get_original_issue(issue_id)
+    issue_id = testcase.bug_information
+    issue = issue_tracker.get_original_issue(issue_id)
   except:
     logs.log_error(
         'Error occurred when fetching issue %s.' % testcase.bug_information)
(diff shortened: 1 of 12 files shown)


src/appengine/handlers/cron/cleanup.py
```diff
@@ -433,10 +435,8 @@ def mark_issue_as_closed_if_testcase_is_fixed(testcase, issue):
     comment += INTERNAL_INCORRECT_COMMENT
     comment += ' and re-open the issue.'
 
-  issue.comment = comment
   issue.status = 'Verified'
-  issue.open = False
```
oliverchang: this is not actually needed, because it's not sent in the actual API request to monorail. the openness of an issue is based on its status value.

src/python/tests/core/issue_management/issue_tracker_test.py
```diff
@@ -0,0 +1,95 @@
...
+    store.remove('LaBel2')
+    self.assertItemsEqual(['Label3'], store)
+    self.assertItemsEqual([], store.added)
+    self.assertItemsEqual(['Label1', 'Label2'], store.removed)
+
+    store.remove('Label4')
+    self.assertItemsEqual(['Label3'], store)
+    self.assertItemsEqual([], store.added)
+    self.assertItemsEqual(['Label1', 'Label2'], store.removed)
+
```
inferno-chromium: nit: add a test where we do add and remove at same time.

src/appengine/handlers/cron/cleanup.py
```diff
@@ -338,14 +338,14 @@ def get_issue_for_testcase(testcase):
...
-  if not issue_tracker_manager:
+  issue_tracker = issue_tracker_utils.get_issue_tracker_for_testcase(
+      testcase, use_cache=True)
+  if not issue_tracker:
     return None
 
   try:
-    issue_id = int(testcase.bug_information)
-    issue = issue_tracker_manager.get_original_issue(issue_id)
+    issue_id = testcase.bug_information
```
inferno-chromium: Isn't testcase.bug_information a string, should get_issue and get_original_issue do the int conversion automatically ?

inferno-chromium:
@jonathanmetzman - this is nice refactoring, can you please do a review pass as well.

inferno-chromium:
i reviewed this as well, so not blocked on @jonathanmetzman. but can address his comments in a future cl.

src/python/issue_management/issue_tracker.py
```diff
@@ -50,6 +111,11 @@ def merged_into(self):
   def merged_into(self, new_merged_into):
     raise NotImplementedError
 
+  @property
+  def open(self):
```
jonathanmetzman: I think `is_open` would be less confusing.

oliverchang: Done.

src/appengine/handlers/cron/cleanup.py
oliverchang: Yes get_issue and get_original_issue should handle the int conversion automatically. No need for callers to convert to int everywhere.

PR #522, opened 2019-06-04
Use new msan archives for chromium docker images.

oliverchang:
Rebuilt archives without libGL.

### DIFF
--- a/docker/chromium/base/Dockerfile
+++ b/docker/chromium/base/Dockerfile
@@ -56,11 +56,11 @@ RUN mkdir $NFS_DIR && \
 
 # Get pre-built msan libraries (with and without origin tracking).
 RUN mkdir /msan-chained-origins && \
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
Have you run the fuzzers without the libraries? How do we know they will still run? I think they won't because they will be missing dependencies.

oliverchang:
Yes, I also tested this. Given that Chrome MSan does not seem to have this issue (it should load it anyway because we pass --use-gl=swiftshader, and the fuzz targets runs fine with the uninstrumented version this seems safe enough. In any case we should not be making anything worse for fuzz targets.

oliverchang:
> I'm just curious, why does CF download its own version of msan libraries when MSAN chrome builds come with their own?

The MSan chrome builds are still 14.04, while our images are 16.04. Neither of these work as-is on our newer debian versions for the whole of Chrome.

PR #524, opened 2019-06-04
Add an initial basic test for the reproduce tool.

mbarbella-chromium:
Uploading this while it's still incomplete as the CL is starting to get pretty big. Let me know if this seems too hacky to land without polishing a bit more.

Running butler.py reproduce doesn't do much now, but it does actually work when jobs are manually coded in. I've only tested it with simple Linux binaries. Next it's going to need to need to get the real definitions from ClusterFuzz reports.


inferno-chromium:
@oliverchang - can you please take a first review pass.

src/local/butler/reproduce.py
```diff
@@ -0,0 +1,99 @@
...
+
+  # Directories not needed by the reproduce tool but which require values.
+  environment.set_value('BOT_TMPDIR', '/tmp/unused')
+  environment.set_value('CRASH_STACKTRACES_DIR', '/tmp/unused')
+  environment.set_value('FUZZER_DIR', '/tmp/unused')
+  environment.set_value('USER_PROFILE_ROOT_DIR', '/tmp/unused')
+
+  environment_values = (
+      environment.parse_environment_definition(testcase.job_definition))
+  for key, value in six.iteritems(environment_values):
```
oliverchang: should we just use `commands.update_environment_for_job` here?

src/local/butler/reproduce.py
```diff
@@ -0,0 +1,99 @@
...
+  return _SimplifiedTestcase(testcase_json)
+
+
+def _download_testcase(_):
+  """Download the test case and return its path."""
+  # TODO(mbarbella): Implement this.
+  return '/tmp/blah'
+
+
+def _prepare_environment(testcase, build_directory):
```
oliverchang: is there a general plan to use more parts of ClusterFuzz code to do setup like this, now that this code is in the same tree?

Would have to take a closer look to figure out how much work needs to be done, but I think ideally we'd be able to do some simple refactorings such that the reproduce tool can just call some functions to do all of these set up, and then these will always be in sync with production. 

If we plan to e.g. call the libFuzzer launcher from this, this might become important since there is a lot of prerequisite setup that has to happen before we can run it properly.

butler.py
```diff
@@ -258,6 +258,13 @@ def main():
       default='us-central',
       help='Region for App Engine.json.')
 
+  parser_reproduce = subparsers.add_parser(
+      'reproduce', help='Reproduce a crash or error from a test case.')
+  parser_reproduce.add_argument(
+      '--testcase', type=int, required=True, help='Testcase ID.')
+  parser_reproduce.add_argument(
+      '--build-dir', type=str, required=True, help='Build directory.')
```
inferno-chromium: Build directory containing the target app and dependencies.

src/local/butler/reproduce.py
```diff
@@ -0,0 +1,99 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""reproduce.py reproduces test cases locally."""
+
+from __future__ import print_function
+
```
inferno-chromium: from builtins import object

src/local/butler/reproduce.py
```diff
@@ -0,0 +1,99 @@
...
+
+def _prepare_environment(testcase, build_directory):
+  """Prepare environment variables based on the test case and build path."""
+  environment.set_default_vars()
+
+  # Environment variables used by the tool.
+  environment.set_value('APP_DIR', build_directory)
+  environment.set_value('BUILDS_DIR', build_directory)
+  environment.set_value('INPUT_DIR', '/tmp/fixme')
```
inferno-chromium: We don't need all these vars like BOT_TMPDIR, INPUT_DIR, etc to be set. Please just call CF func
environment.set_bot_environment() and remove environment.set_default_vars() [that is called as well by set_bot_environment].

src/local/butler/reproduce.py
inferno-chromium: +1

src/local/butler/reproduce.py
```diff
@@ -0,0 +1,99 @@
...
+
+from src.python.fuzzing import tests
+from src.python.system import environment
+
+
+class _SimplifiedTestcase(object):
+  """Minimal representation of a test case."""
+
+  def __init__(self, testcase_json):
+    self.crash_state = testcase_json['crash_state']
```
inferno-chromium: This is faulty logic, that came from old reproduce tool. Now we want to serialize/deserialize the whole testcase and job object returned from server (checkout to_dict to serialize and some way to deserialize). there can be numerous fields in testcase that might be get silently used (e.g. metadata or whatever property). Please create a appengine handler that return full entities in some encoded form.

Also _get_testcase() should use that rather than creating testcase itself with few fields. for test, use test_utils.create_generic_testcase.

src/local/butler/reproduce.py

src/local/butler/reproduce.py
mbarbella-chromium: I agree that we should try to handle the serialization/deserialization in a generic way so that we don't have to explicitly list out fields here. I'd like to leave this as-is for this change, but I'll clean this up next.

Right now we're making use of the added job definition field, and I think going forward we'll still want this to be its own class. This was the bare minimum from testcase needed to get this example working, but it wasn't meant to be how we do things long-term. As for using a separate class, since we won't be able to access datastore, it may be useful to be able to override specific fields or functions (e.g. put).

Let me know if this sounds good to you. I think this is the last unresolved point.

src/local/butler/reproduce.py
```diff
@@ -0,0 +1,106 @@
...
+  return '/tmp/blah'
+
+
+def _prepare_environment(testcase, build_directory):
+  """Prepare environment variables based on the test case and build path."""
+  # Create a temporary directory to use as ROOT_DIR with a copy of the default
+  # bot and configuration directories nested under it.
+  old_root_dir = environment.get_value('ROOT_DIR')
+  root_dir = tempfile.mkdtemp()
+  shutil.copytree(
```
inferno-chromium: I don't think this is enough, there is resources dir needed as well. I think we can remove the mkdtemp logic for now. Right now it is in gitignore, so those dirs wont conflict. When we later package reproduce tool, it won't matter either since that dir will be temp in itself. then can remove line 96 as well.

mbarbella-chromium: It seems to delete the READMEs without it, which is inconvenient. I'll see if I can resolve this now.

mbarbella-chromium: Added the resources directory.

src/local/butler/reproduce.py
```diff
@@ -0,0 +1,110 @@
...
+  root_dir = tempfile.mkdtemp()
+  environment.set_value('ROOT_DIR', root_dir)
+  shutil.copytree(
+      os.path.join(old_root_dir, 'bot'), os.path.join(root_dir, 'bot'))
+  shutil.copytree(
+      os.path.join(old_root_dir, 'configs'), os.path.join(root_dir, 'configs'))
+  shutil.copytree(
+      os.path.join(old_root_dir, 'resources'),
+      os.path.join(root_dir, 'resources'))
+  environment.set_value('ROOT_DIR', root_dir)
```
inferno-chromium: Isn't this set here as well ?

src/local/butler/reproduce.py
```diff
@@ -0,0 +1,110 @@
...
+  # TODO(mbarbella): Implement this.
+  return '/tmp/blah'
+
+
+def _prepare_environment(testcase, build_directory):
+  """Prepare environment variables based on the test case and build path."""
+  # Create a temporary directory to use as ROOT_DIR with a copy of the default
+  # bot and configuration directories nested under it.
+  old_root_dir = environment.get_value('ROOT_DIR')
+  root_dir = tempfile.mkdtemp()
```
inferno-chromium: can you please move line 66-76 into some helper ? and maybe shutil.copytree in a loop.

src/python/base/errors.py
```diff
@@ -110,8 +110,12 @@ def __str__(self):
 class BadConfigError(Error):
   """Error thrown when configuration is bad."""
 
-  def __init__(self):
-    super(BadConfigError, self).__init__('Bad configuration.')
+  def __init__(self, config_dir=None):
```
inferno-chromium: why is config_dir optional, lets make it there always, i see those are two call sites.

mbarbella-chromium: Good catch. I added this for debugging and meant to revert it but might as well keep.

src/local/butler/reproduce.py
mbarbella-chromium: Oh good catch. Yeah, it looks like there was some other issue. Still debugging as it's not reproducing for me locally. Seems to be a subtle side-effect after the test runs.

src/local/butler/reproduce.py
mbarbella-chromium: SGTM, but going to wait until the test failures are resolved to modify this.

PR #525, opened 2019-06-04
Rename tests.py

flowerhack:
It's confusing to have tests.py, which is not actually a set of tests,
    and which handles a pretty big variety of functions.
    
    Rename it to testcase_manager, which (I think) more accurately captures
    what it's responsible for, though I'm open to other ideas!


flowerhack:
@oliverchang should be ready for review!

PR #526, opened 2019-06-05
Add closed_time property to issues (#504).

oliverchang:


oliverchang:
TBR to fix issues.

src/python/tests/appengine/handlers/cron/cleanup_test.py
```diff
@@ -783,8 +790,9 @@ def test_notify_closed_issue_if_testcase_is_open_9(self):
     testcase.bug_information = str(self.issue.id)
     testcase.put()
     self.issue._monorail_issue.open = False
-    self.issue.closed = test_utils.CURRENT_TIME - datetime.timedelta(
-        days=data_types.NOTIFY_CLOSED_BUG_WITH_OPEN_TESTCASE_DEADLINE + 1)
+    self.issue._monorail_issue.closed = (
```
inferno-chromium: Shouldn't all these be self.issue.closed_time ? I think we should convert the tests to invoke correct wrapper instead of self.issue._monorail_issue.* ?

oliverchang: closed_time is readonly.

Once everything is moved over I will clean up our tests to use a mock implementation of the interface than relying on the underlying monorail Issue.

inferno-chromium: sg, no rush, thanks for explaining.

PR #530, opened 2019-06-05
Add support for custom bucket suffixes when creating corpus buckets.

oliverchang:


configs/test/project.yaml
```diff
@@ -54,6 +54,9 @@ stacktrace:
   # Stacks to ignore from being considered as a crash altogether.
   stack_blacklist_regexes:
 
+# Suffix to append to automatically created buckets.
+# bucket_domain_suffix: .bucket.suffix
```
inferno-chromium: maybe remove the .(dot) from front. We automatically add the dot, and someone can specify it incorrectly leading to two dots. That bring another question, which is do we need to document this, if yes, please can you explain more in comment,

oliverchang: Done!

don't think we need to document this, since the default create_config process will work for pretty much everyone.

PR #531, opened 2019-06-06
Convert one more issue.add_label to issue.labels.add (#504).

oliverchang:
This is called by issue_filer and cleanup, which were both converted to use the new interface.

Also fix another use of issue.summary.


oliverchang:
TBR. /gcbrun

PR #533, opened 2019-06-06
Fix App Engine import exception.

oliverchang:


oliverchang:
TBR. one more case that #500 would have caught :)

inferno-chromium:
Thanks a lot, LGTM! Sorry for breakage.

Issue #541, opened 2019-06-07
Move issue updates to appengine cron.


inferno-chromium:

inferno-chromium:
This is needed for Google Issue Tracker integration (https://developers.google.com/issue-tracker/), so that all updates go through appengine and we don't need to do these from GCE.

inferno-chromium:
@oliverchang - i think for progression task, we can just remove the issue comment code completely. See cleanup.mark_issue_as_closed_if_testcase_is_fixed, i think we do double update which might be not needed. we can just add the range in the message in cleanup function and then progression task one can go away. wdyt ? 

Rest, should be handled by my cl above.

oliverchang:
having a single update SGTM.

PR #542, opened 2019-06-07
Move flakiness label update to appengine cleanup cron.

inferno-chromium:
Also, convert user uploaded testcase issue update code to
use apis using new_comment. This fix is part of
https://github.com/google/clusterfuzz/issues/541

### DIFF
--- a/src/python/datastore/data_handler.py
+++ b/src/python/datastore/data_handler.py
@@ -1062,23 +1062,6 @@ def update_issue_impact_labels(testcase, issue):
                    label_utils.impact_to_string(new_impact))
 
 
-def get_issue_for_testcase(testcase):
-  """Return issue associated with the testcase (if any)."""
-  if not testcase.bug_information:
-    return None
-
-  issue_id = int(testcase.bug_information)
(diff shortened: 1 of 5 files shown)


src/python/datastore/data_handler.py
```diff
@@ -1181,15 +1164,13 @@ def create_user_uploaded_testcase(key,
...
-      issue.comment = ('ClusterFuzz is analyzing your testcase. '
-                       'Developers can follow the progress at %s.' % report_url)
-      issue.save()
+  issue = issue_tracker_utils.get_issue_for_testcase(testcase)
+  if issue:
+    report_url = TESTCASE_REPORT_URL.format(
+        domain=get_domain(), testcase_id=testcase_id)
+    comment = ('ClusterFuzz is analyzing your testcase. '
+               'Developers can follow the progress at %s.' % report_url)
+    issue.save(new_comment=comment)
```
inferno-chromium: Note that this function is called from two places
- directly from appengine for one testcase upload
- from unpack task with multiple testcases. i am not expecting someone to upload multiple testcases for a single bug. since this usecase is not common, we can just leave code as-is and not need to move to cleanup cron.

PR #546, opened 2019-06-10
Verify core_pattern setting for AFL in launcher.

inferno-chromium:
Fuzzer.py code runs on host bot, so it does not work in OSS-Fuzz.
A project code might lead to core_pattern file change, so try
verifying and change core_pattern to correct value in launcher
itself (since we run in docker and have sudoer).


src/python/bot/fuzzers/afl/fuzzer.py
```diff
@@ -46,23 +45,7 @@ def generate_arguments(self, fuzzer_path):  # pylint: disable=unused-argument
...
     return ''
 
-  def _check_system_settings(self):
-    """Check system settings required for AFL."""
-    kernel_core_pattern_file_path = '/proc/sys/kernel/core_pattern'
-    if (os.path.exists(kernel_core_pattern_file_path) and
-        open(kernel_core_pattern_file_path).read().strip() != 'core'):
-      logs.log_fatal_and_exit('AFL needs core_pattern to be set to core.')
-
-    cpu_scaling_file_path = (
```
inferno-chromium: I need to port this part later. i think i need to check cpu*, not just cpu0.

src/python/bot/fuzzers/afl/launcher.py
```diff
@@ -1251,6 +1252,29 @@ def associate_features_with_file(self, features, path):
...
+    if not os.path.exists(constants.CORE_PATTERN_FILE_PATH):
+      return False
+
+    return open(constants.CORE_PATTERN_FILE_PATH).read().strip() == 'core'
+
+  if _check_core_pattern_file():
+    return
+
+  return_code = subprocess.call(
+      'timeout 2s sudo bash -c "echo core > {path}"'.format(
```
oliverchang: why do we need the timeout?

inferno-chromium: If we don't have sudoer access, this allows to bailout from the subprocess command (subprocess.call timeout arg only in py3) and then properly exit on log fatal, rather than leading to infinite hang (since we would be stuck to wait on user to enter sudo pass).

oliverchang: how about using -n ?

     -n, --non-interactive
                 Avoid prompting the user for input of any kind.  If a password is required for the command to run, sudo will display an
                 error message and exit.

inferno-chromium: Done!

PR #547, opened 2019-06-10
Fuzz strategy selection

mukundv-chrome:
Updated cron.yaml files and cleaned up BigQuery query and updated filter on unstable targets.
Corresponding tests and test data were also updated accordingly.


mbarbella-chromium:
Closing a duplicate.

Issue #503, opened 2019-05-29
Tracking bug: enabling DFT-based fuzzing


Dor1s:
There are three main components:

1) Build / test support in OSS-Fuzz: this is tracked in https://github.com/google/oss-fuzz/issues/1632

2) Job support on ClusterFuzz side: we don't need a new job for DFT-based fuzzing, but we do need a separate build. This creates a problem: how do I enable a build on OSS-Fuzz side (`project.yaml` is the obvious answer), but do not create any additional jobs on CF side?

3) Strategy support on ClusterFuzz: might get too complicated, given that we need to download a separate build for that.

Dor1s:
While writing this, I came up to another solution:

1) Set up DataFlow build in a way that it would have both ASan and DFSan binaries in the same archive.

2) Create a new Job for DataFlow config. Use it as an ASan job, except two minor differences:
 A) always use `collect_data_flow` option
 B) always use `fork` mode (LF currently crashes when `collect_data_flow` is used without `fork` mode)

Besides cleaner set up and less confusing strategy logic, this will also guarantee that we're using exactly the same source code for both ASan and DFSan binaries. I don't know whether this is a blocker, but it's definitely not beneficial to have mismatch in the corresponding ASan and DFSan binaries.

I'm really favoring this approach now. Will think more and pitch it to you in the meeting today :)

Dor1s:
Thanks team for the discussion today. Now I'm leaning towards a new strategy again. I'll take a closer look at the code and try to predict / draft the changes necessary for enabling that.

Dor1s:
The new functionality is broken into three CLs:

1) https://github.com/google/clusterfuzz/pull/580: Add support for dataflow builds to the OSS-Fuzz setup handler.
2) https://github.com/google/clusterfuzz/pull/582: Add AuxiliaryBuild class for extra builds like dataflow.

3) https://github.com/google/clusterfuzz/pull/583: Add dataflow tracing strategy to libFuzzer launcher.

Dor1s:
The strategy is enabled and seems to be running fine. I'm closing this. The documentation is being tracked as #558.

PR #548, opened 2019-06-10
[WORK IN PROGRESS] Draft changes for adding DataFlow strategy (#503).

Dor1s:


Dor1s:
Oliver and Abhishek, please take a look at this draft and let me know whether you think I'm moving in the right direction :)

src/python/bot/fuzzers/libFuzzer/launcher.py
```diff
@@ -197,6 +200,14 @@ def do_mutator_plugin():
           strategy.MUTATOR_PLUGIN_STRATEGY, default=MUTATOR_PLUGIN_PROBABILITY))
 
 
+def do_dataflow_tracing():
+  """Return whether or now to use dataflow tracing."""
```
Dor1s: typo

Dor1s:
> Oliver and Abhishek, please take a look at this draft and let me know whether you think I'm moving in the right direction :)

ping @oliverchang @inferno-chromium

src/python/build_management/build_manager.py
```diff
@@ -1153,7 +1153,20 @@ def setup_trunk_build():
...
 
-  return setup_regular_build(revision)
+  build = setup_regular_build(revision)
+  if build:
+    setup_auxiliary_builds(revision)
+
+  return build
+
+
+def setup_auxiliary_builds(revision):
```
oliverchang: since this is specific to fuzzing, it seems like this particular function belongs in fuzz_task.

src/appengine/handlers/cron/oss_fuzz_setup.py
```diff
@@ -521,6 +524,12 @@ def sync_cf_job(project, info, corpus_bucket, quarantine_bucket, logs_bucket,
     if selective_unpack:
       job.environment_string += 'UNPACK_ALL_FUZZ_TARGETS_AND_FILES = False\n'
 
+    if 'dataflow' in info.get('fuzzing_engines'):
+      dataflow_build_bucket_path = get_build_bucket_path(
+          project, 'dataflow', 'dataflow', template.architecture),
```
inferno-chromium: 'dataflow' repeated twice ? dataflow as fuzzing engine makes sense, but then why is it called a memory tool too, shouldn't template.memory_tool be used ?

src/appengine/handlers/cron/oss_fuzz_setup.py
```diff
@@ -521,6 +524,12 @@ def sync_cf_job(project, info, corpus_bucket, quarantine_bucket, logs_bucket,
...
+      job.environment_string += (
+          'DATAFLOW_BUILD_BUCKET_PATH = %s\n' % dataflow_build_bucket_path)
```
inferno-chromium: Please do remember to update docs for this env var here (can be done later, but fyi to not forget).

docs/production-setup/setting_up_a_fuzzing_job.md
docs/reference/job_definition.md

src/python/bot/fuzzers/libFuzzer/launcher.py
```diff
@@ -895,7 +906,20 @@ def main(argv):
     arguments.append(constants.VALUE_PROFILE_ARGUMENT)
     fuzzing_strategies.append(strategy.VALUE_PROFILE_STRATEGY)
 
-  if do_fork():
+  use_dataflow_tracing = False
+
+  # Depends on the presense of DFSan instrumented build.
+  dataflow_build_path = environment.get_value('DATAFLOW_BUILD_DIR_???')
```
inferno-chromium: ???

src/python/bot/fuzzers/libFuzzer/launcher.py
```diff
@@ -895,7 +906,20 @@ def main(argv):
     arguments.append(constants.VALUE_PROFILE_ARGUMENT)
     fuzzing_strategies.append(strategy.VALUE_PROFILE_STRATEGY)
 
-  if do_fork():
+  use_dataflow_tracing = False
```
inferno-chromium: Just initialize it to "dataflow_build_path and do_dataflow_tracing()", use in if, rather than setting to true later and initing to false ?

src/python/build_management/build_manager.py
```diff
@@ -1153,7 +1153,20 @@ def setup_trunk_build():
     logs.log_error('Unable to find a matching revision.')
     return None
 
-  return setup_regular_build(revision)
+  build = setup_regular_build(revision)
+  if build:
+    setup_auxiliary_builds(revision)
```
inferno-chromium: Add a comment on what auxiliary_builds mean.

src/python/build_management/build_manager.py
```diff
@@ -1153,7 +1153,20 @@ def setup_trunk_build():
...
+
+  return build
+
+
+def setup_auxiliary_builds(revision):
+  """Sets up auxiliary builds when necessary (e.g. DFSan build)."""
+  dataflow_build_bucket_path = environment.get_value(
+      'DATAFLOW_BUILD_BUCKET_PATH')
+  # Download the latest revision, as we cannot guarantee excat matching and do
```
inferno-chromium: typo exact

src/python/build_management/build_manager.py
inferno-chromium: Yes can go there.

src/appengine/handlers/cron/oss_fuzz_setup.py
Dor1s: `dataflow` is both a new engine and a new sanitizer. I agree it's kinda annoying, but don't see a better way than specifying it twice.

`template.memory_tool` cannot be used, as we don't have a template for DFsan (we don't create a new job for it), and here `template.memory_tool` points to the sanitizer which fuzzing binaries are built with (e.g. ASan). DataFlow binaries are built with dataflow sanitizer.

"use named arguments" -- done, thanks! Also added a comment to clarify that.

src/appengine/handlers/cron/oss_fuzz_setup.py
Dor1s: Thanks, filed https://github.com/google/clusterfuzz/issues/558 to remember that.

src/python/bot/fuzzers/libFuzzer/launcher.py
Dor1s: Just a sign that it's a sketch :) Will use a normal name or some other way to pass the build dir.

src/python/build_management/build_manager.py
Dor1s: Done.

src/python/build_management/build_manager.py
Dor1s: Done.

src/python/bot/fuzzers/libFuzzer/launcher.py
Dor1s: Good idea, thanks, done!

src/python/build_management/build_manager.py
Dor1s: Not sure I understand what you mean. I agree that it's better to call this function from the fuzz_task, as auxiliary builds aren't needed for any other task. But I think the function itself should stay in the build_manager, as it would need to use some of the classes / functions defined in the build_manager.

Dor1s:
Still needs some work (add new tests + make sure nothing breaks), but has much more implementation details now. Please take a look, if you get a change. Otherwise, can wait until more production ready version :)

@oliverchang what do I need to run to re-generate proto-dependent code in Go? https://github.com/google/clusterfuzz/blob/master/src/protos/README.md has only Python instructions.

src/python/build_management/build_manager.py
oliverchang: setup_auxiliary_builds() should belong in fuzz_task, since it will only ever be called in fuzzing. We can still call classes / functions defined in build_manager from there.

oliverchang: so, inside fuzz_task:

```
def set_up_dataflow_build(...):
  build_manager.RegularBuild()

build_manager.setup_build(...)
set_up_data_flow_build(...)
```

You may need to do some minor refactoring to avoid the APP_PATH/APP_DIR etc from being overwritten.

Dor1s: Alright, done! Thanks for mentioning the env variables, I think I've made sure those won't be overwritten, but I'll double check it in tests.

Dor1s:
Woohoo, existing tests pass. Now it's time to add some new ones :)

Dor1s:
Ok, everything looks good, but this is way too large, so I'm going to split this into 4 CLs.

Dor1s:
Alright, done in
* https://github.com/google/clusterfuzz/pull/580
* https://github.com/google/clusterfuzz/pull/582
* https://github.com/google/clusterfuzz/pull/583
* https://github.com/google/clusterfuzz/pull/581 (unrelated to DataFlow)

PR #549, opened 2019-06-11
Convert more code to use new issue tracker interface (#504).

oliverchang:
Convert grouper, download, and access.


inferno-chromium:
Landing it so that we can get some data on stability before releasing next version of CF.

PR #551, opened 2019-06-11
Remove duplicate issue updates in progression task.

inferno-chromium:
This is also needed for
https://github.com/google/clusterfuzz/issues/541


oliverchang:
thanks!! LGTM

PR #555, opened 2019-06-12
Revert "Speculative fix for android issue with unusable gpu."

inferno-chromium:
Reverts google/clusterfuzz#554

This did not fix it. Also, it could be related to:
https://bugs.chromium.org/p/chromium/issues/detail?id=927130


inferno-chromium:
TBR

PR #560, opened 2019-06-12
Fuzz strategy selection bigquery

mukundv-chrome:
Added code to create bigquery table schema for strategy probabilities and update table daily with cron job.
More consistent naming


configs/test/bigquery/datasets.yaml
```diff
@@ -209,3 +209,19 @@ resources:
...
+      timePartitioning:
+        type: DAY
+      schema:
+        fields:
+          - name: strategy_name
+            type: STRING
+          - name: probability
+            type: FLOAT
+          - name: run_count
+            type: INTEGER
```
mbarbella-chromium: Nit: newline at end of file.

src/appengine/handlers/cron/fuzz_strategy_selection.py
```diff
@@ -102,7 +102,25 @@ def _query_multi_armed_bandit_probs(client):
   return client.query(query=BANDIT_PROBABILITY_QUERY).rows
 
 
-def _query_and_upload_strategy_weights(client):
```
mbarbella-chromium: Were these weights->probs changes intentional? It seems better to leave them as-is to me if not.

src/appengine/handlers/cron/fuzz_strategy_selection.py
```diff
@@ -102,7 +102,25 @@ def _query_multi_armed_bandit_probs(client):
...
+
+  for row in data:
+    bigquery_row = {
+        'strategy_name': row['strategy'],
+        'probability': row['bandit_weight'],
+        'run_count': row['run_count']
+    }
+    bigquery_data.append(big_query.Insert(row=bigquery_row, insert_id=None))
+
+  client = big_query.Client(
```
mbarbella-chromium: This is making me realize that there's no reason to pass client as a variable from the handler to _query_multi_armed_bandit_probs (should that be weights as well). It was useful for the fuzzer weights code since we needed to query in multiple places, but you only use the client once so might as well instantiate it there.

src/appengine/handlers/cron/fuzz_strategy_selection.py
mukundv-chrome: They were intentional, I wanted to change every mention of 'weights' in the file to 'probabilities' for consistency sake because before it was a mix of the two

configs/test/bigquery/datasets.yaml
```diff
@@ -209,3 +209,20 @@ resources:
...
+
```
mbarbella-chromium: It seems like there some trailing whitespace here that should be deleted.

PR #562, opened 2019-06-13
Add a new test case info handler for the reproduce tool.

mbarbella-chromium:
This handler includes the full serialized test case object as well as
the job definition associated with it.


mbarbella-chromium:
PTAL when you have time.

PR #564, opened 2019-06-14
 [Reproduce tool] Get serialized test case objects from the new handler.

mbarbella-chromium:
PTAL when you have time.

There's still a fair amount that needs to be cleaned up here (see TODOs), and I want to break the file up. This change is large enough that I thought it made sense to upload it for review now, with the cleanup in follow-up changes.


src/local/butler/reproduce.py
```diff
@@ -16,41 +16,114 @@
...
+import urllib
+import webbrowser
 
+from src.python.base import utils
 from src.python.bot.tasks import commands
 from src.python.fuzzing import testcase_manager
 from src.python.system import environment
 from src.python.system import shell
 
+# TODO(mbarbella): This should work cross platform.
```
inferno-chromium: Yes this TODO can be removed.

src/local/butler/reproduce.py
```diff
@@ -16,41 +16,114 @@
...
+AUTHORIZATION_CACHE_FILE = os.path.join(
+    os.path.expanduser('~'), '.config', 'clusterfuzz', 'authorization-cache')
 
-class _SimplifiedTestcase(object):
+# TODO(mbarbella): Don't use the old clusterfuzz-tools client id.
+OAUTH_URL = 'https://accounts.google.com/o/oauth2/v2/auth?%s' % (
+    urllib.urlencode({
+        'scope':
+            'email profile',
+        'client_id': ('981641712411-sj50drhontt4m3gjc3hordjmp'
```
inferno-chromium: Yes, we really don't want this. As you said above, this should work without it, please find a way to get current user auth email.

src/local/butler/reproduce.py
```diff
@@ -16,41 +16,114 @@
...
+                      'c7bn50f.apps.googleusercontent.com'),
+        'response_type':
+            'code',
+        'redirect_uri':
+            'urn:ietf:wg:oauth:2.0:oob'
+    }))
+
+# TODO(mbarbella): This value should come from the configuration.
+TESTCASE_URL = 'https://clusterfuzz.com/reproduce-tool/testcase-info'
```
inferno-chromium: put something in data_handler ? It defined several _URL globals and uses get_domain()

src/local/butler/reproduce.py
```diff
@@ -16,41 +16,114 @@
...
+def _get_authorization():
+  """Get the value for an oauth authorization header."""
+  # Try to read from cache.
+  authorization = utils.read_data_from_file(
+      AUTHORIZATION_CACHE_FILE, eval_data=False)
+  if authorization:
+    return authorization
+
+  # Prompt the user for a code if we don't have one.
+  with SuppressOutput():
```
inferno-chromium: SupressOutput feels very hacky, can we just exclude it, it might be ok to have some stdout, stderr ?

src/local/butler/reproduce.py
```diff
@@ -16,41 +16,114 @@
...
+    return True
+
+
+def _get_authorization():
+  """Get the value for an oauth authorization header."""
+  # Try to read from cache.
+  authorization = utils.read_data_from_file(
+      AUTHORIZATION_CACHE_FILE, eval_data=False)
+  if authorization:
+    return authorization
```
inferno-chromium: When to know it expires ? or maybe keep some date ?

src/local/butler/reproduce.py
```diff
@@ -82,20 +155,26 @@ def _prepare_environment(testcase, build_directory):
...
   environment.set_bot_environment()
-  commands.update_environment_for_job(testcase.job_definition)
 
   # Overrides that should not be set to the default values.
   environment.set_value('APP_DIR', build_directory)
   environment.set_value('BUILDS_DIR', build_directory)
+
+
+def _update_environment_for_job(testcase, build_directory):
+  commands.update_environment_for_job(testcase.job_definition)
```
inferno-chromium: testcase.job_definition is string or list, since update_environment_for_job

src/local/butler/reproduce.py
mbarbella-chromium: This is essentially just copied from the old tool, but not having it is problematic. It ends up outputting around the same time as the prompt for the verification code, which looks very confusing (essentially drowns out the request for the verification code).

src/local/butler/reproduce.py
mbarbella-chromium: Good catch, we'll want to handle that in this change. We won't know it's expired until we make a request and it fails, as far as I can tell. The old tool has some logic to handle that, so I'll add it here as well.

src/local/butler/reproduce.py

src/local/butler/reproduce.py

src/local/butler/reproduce.py
mbarbella-chromium: I'd like to address this in a later change. We do have functions like get_domain but this wouldn't work with only the test config (which is all someone would have in a fresh checkout). We're going to need to come up with a solution here, but I don't think there's a quick fix for this one.

src/local/butler/reproduce.py
mbarbella-chromium: It's a string here, which I believe should work correctly. Is there any issue with this?

src/local/butler/reproduce.py

src/local/butler/reproduce.py
```diff
@@ -39,8 +38,8 @@
     urllib.urlencode({
         'scope':
             'email profile',
-        'client_id': ('981641712411-sj50drhontt4m3gjc3hordjmp'
-                      'c7bn50f.apps.googleusercontent.com'),
+        'client_id': ('602540103821-lccrsee9e5hbe3lpdsghin0'
```
inferno-chromium: We need to remove the hardcode, can you add a TODO here. this should come from config or somewhere. can remove "# TODO(mbarbella): Don't use the old clusterfuzz-tools client id."

src/local/butler/reproduce.py
```diff
@@ -79,24 +78,25 @@ def __exit__(self, *_):
     return True
 
 
-def _get_authorization():
+def _get_authorization(force_reauthorization):
   """Get the value for an oauth authorization header."""
-  # Try to read from cache.
-  authorization = utils.read_data_from_file(
+  # Try to read from cache unless we need to reauthorize.
+  if not force_reauthorization:
```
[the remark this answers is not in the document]
mbarbella-chromium: It shouldn't, no. The if below is still nested in this block.

src/local/butler/reproduce.py
mbarbella-chromium: In the most recent commit it's been changed to "# TODO(mbarbella): Client ID and domain should be configurable."

PR #569, opened 2019-06-17
Add support for uploading testcases from data bundle for coverage.

inferno-chromium:
This fixes an important TODO which is causing low coverage from
blackbox fuzzers.

### DIFF
--- a/src/python/fuzzing/coverage_uploader.py
+++ b/src/python/fuzzing/coverage_uploader.py
@@ -18,18 +18,18 @@ import os
 from base import utils
 from bot.fuzzers import builtin_fuzzers
 from config import local_config
+from fuzzing import testcase_manager
 from google_cloud_utils import gsutil
 from google_cloud_utils import storage
 from metrics import logs
 from system import environment
-from system import shell
 
 LIST_FILE_BASENAME = 'file_list.txt'
-MAX_TESTCASE_DIRECTORY_SIZE = 1 * 1000 * 1000 * 1000  # 1 GB
-TESTCASES_PER_DAY = 1000
+TESTCASES_PER_DAY = 5000
 
 
-def upload_testcases_if_needed(fuzzer_name, testcase_list, testcase_directory):
+def upload_testcases_if_needed(fuzzer_name, testcase_list, testcase_directory,
+                               data_directory):
   """Upload test cases from the list to a cloud storage bucket."""
   # Since builtin fuzzers have a coverage minimized corpus, no need to upload
   # test case samples for them.
@@ -41,20 +41,17 @@ def upload_testcases_if_needed(fuzzer_name, testcase_list, testcase_directory):
   if not bucket_name:
     return
 
-  # Only consider test cases in the output directory. We might upload too much
-  # if we search the data directory as well, or have missing resources.
-  # TODO(mbarbella): Support resources in data bundles.
-  testcase_list = [
-      os.path.relpath(testcase, testcase_directory)
-      for testcase in testcase_list
-      if testcase.startswith(testcase_directory)
-  ]
-  if not testcase_list:
-    return
-
-  # Bail out if this batch of test cases is too large.
-  directory_size = shell.get_directory_size(testcase_directory)
-  if directory_size >= MAX_TESTCASE_DIRECTORY_SIZE:
+  files_list = []
+  has_testcases_in_testcase_directory = False
+  has_testcases_in_data_directory = False
+  for testcase_path in testcase_list:
+    if testcase_path.startswith(testcase_directory):
+      files_list.append(os.path.relpath(testcase_path, testcase_directory))
+      has_testcases_in_testcase_directory = True
+    elif testcase_path.startswith(data_directory):
+      files_list.append(os.path.relpath(testcase_path, data_directory))
+      has_testcases_in_data_directory = True
+  if not files_list:
     return
 
   formatted_date = str(utils.utcnow().date())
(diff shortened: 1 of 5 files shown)


src/python/fuzzing/coverage_uploader.py
```diff
@@ -86,9 +83,24 @@ def upload_testcases_if_needed(fuzzer_name, testcase_list, testcase_directory):
...
+  if has_testcases_in_testcase_directory:
+    # Sync everything in |testcase_directory| since it is fuzzer-generated.
+    runner.rsync(testcase_directory, gcs_base_url)
+
+  if has_testcases_in_data_directory:
+    # Sync all fuzzer generated testcase in data bundle directory.
+    runner.rsync(
+        data_directory,
+        gcs_base_url,
+        exclusion_pattern=('"(^|.*/)(?!{fuzz_prefix})[^/]+$"'.format(
```
Dor1s: Would this work on Windows bots (if we ever upload anything from them)?

src/python/fuzzing/coverage_uploader.py
```diff
@@ -41,20 +41,17 @@ def upload_testcases_if_needed(fuzzer_name, testcase_list, testcase_directory):
...
-  if not testcase_list:
-    return
-
-  # Bail out if this batch of test cases is too large.
-  directory_size = shell.get_directory_size(testcase_directory)
-  if directory_size >= MAX_TESTCASE_DIRECTORY_SIZE:
+  files_list = []
+  has_testcases_in_testcase_directory = False
+  has_testcases_in_data_directory = False
+  for testcase_path in testcase_list:
```
Dor1s: There is an issue with `TESTCASES_PER_DAY` limit enforcement. basically, we have some fuzzers that manage to upload many thousands of testcases even though the limit if `1000` -- I can show you the number on my desktop. I think we should either:

1) break out of the loop, if `len(files_list) >= TESTCASES_PER_DAY`
2) write only up to `TESTCASES_PER_DAY` elements from `files_list` after the loop

We can also consider shuffling either `testcase_list` or `files_list` (not sure if it's needed, maybe it's already random order?)

Otherwise, the corpus is skewed towards those fuzzers that generate a lot of inputs and upload 10-20K instead of 1K. If we account for the limit in this function, we'll still have a race, but should have much better numbers anyway.

src/python/fuzzing/coverage_uploader.py
```diff
@@ -18,18 +18,18 @@
...
 from google_cloud_utils import gsutil
 from google_cloud_utils import storage
 from metrics import logs
 from system import environment
-from system import shell
 
 LIST_FILE_BASENAME = 'file_list.txt'
-MAX_TESTCASE_DIRECTORY_SIZE = 1 * 1000 * 1000 * 1000  # 1 GB
-TESTCASES_PER_DAY = 1000
+TESTCASES_PER_DAY = 5000
```
Dor1s: Let's change it to `2000` for now, or even leave `1000`? `gs://clusterfuzz-fuzzer-testcases/2019-06-17/` for example takes 35GB. I'm checking the yesterday's directory right now, as it may be even bigger.

inferno-chromium: Ok putting back to 1000.

src/python/fuzzing/coverage_uploader.py
inferno-chromium: Nope, thanks for bringing this up. Simplifying it to work on both.

src/python/fuzzing/coverage_uploader.py
inferno-chromium: Fixed

src/python/fuzzing/coverage_uploader.py
```diff
@@ -81,14 +78,33 @@ def upload_testcases_if_needed(fuzzer_name, testcase_list, testcase_directory):
...
+    # Sync all fuzzer generated testcase in data bundle directory.
+    runner.rsync(
+        data_directory,
+        gcs_base_url,
+        exclusion_pattern=('"(?!.*{fuzz_prefix})"'.format(
+            fuzz_prefix=testcase_manager.FUZZ_PREFIX)))
+
+    # Sync all possible resource dependencies as a best effort.
+    runner.rsync(
+        data_directory, gcs_base_url, exclusion_pattern='"(?!.*resource)"')
```
mbarbella-chromium: Which fuzzers is this pattern expected to work for? I'm having trouble thinking of specific fuzzers that write directly to the data directory other than the layout test ones, and I'm not sure that adding this clause helps there.

inferno-chromium: Added explanation.

Dor1s:
FTR, this seems to have a positive impact!

```
>>> import json
>>> o = json.loads(open('668784.json').read())
>>> n = json.loads(open('670198.json').read())
>>> 
>>> for i in [o, n]: print i['data'][0]['totals']
... 
{u'instantiations': {u'count': 1483026, u'covered': 425603, u'percent': 28.698283104948935}, u'regions': {u'count': 5021356, u'covered': 1118738, u'percent': 22.279599375148866, u'notcovered': 3902618}, u'functions': {u'count': 610117, u'covered': 183536, u'percent': 30.08209900723959}, u'lines': {u'count': 6929883, u'covered': 1810249, u'percent': 26.122360218780027}}
{u'instantiations': {u'count': 1488595, u'covered': 458372, u'percent': 30.792257128366007}, u'regions': {u'count': 5028031, u'covered': 1191125, u'percent': 23.68969085512798, u'notcovered': 3836906}, u'functions': {u'count': 611126, u'covered': 194989, u'percent': 31.906513550397136}, u'lines': {u'count': 6940684, u'covered': 1940285, u'percent': 27.955241875296437}}
```

PR #574, opened 2019-06-18
Fix extra period in testcase comment metadata.

inferno-chromium:
Fixes https://bugs.chromium.org/p/chromium/issues/detail?id=914004


inferno-chromium:
TBR minor fix.

PR #578, opened 2019-06-18
Strategy selection code refactor

mukundv-chrome:
Refactored strategy selection code to be separate from libFuzzer launcher, tests included.


src/python/bot/fuzzers/libFuzzer/launcher.py
```diff
@@ -880,32 +802,42 @@ def main(argv):
...
+      arguments.append('%s%d' % (constants.FORK_FLAG, num_fuzz_processes))
+      fuzzing_strategies.append(
+          '%s_%d' % (strategy.FORK_STRATEGY, num_fuzz_processes))
 
   extra_env = {}
-  if do_mutator_plugin():
-    if use_mutator_plugin(target_name, extra_env, minijail_chroot):
-      fuzzing_strategies.append(strategy.MUTATOR_PLUGIN_STRATEGY)
+  # TODO(metzman): Support Windows.
+  if strategy_pool[strategy.MUTATOR_PLUGIN_STRATEGY]:
```
mbarbella-chromium: Nit: "and" instead of deep nesting.

src/python/bot/fuzzers/libFuzzer/strategy_selection.py
```diff
@@ -0,0 +1,139 @@
...
+  ml_rnn_p = engine_common.get_strategy_probability(
+      strategy.CORPUS_MUTATION_ML_RNN_STRATEGY,
+      default=CORPUS_MUTATION_ML_RNN_PROBABILITY)
+  return engine_common.decide_with_probability(radamsa_p + ml_rnn_p)
+
+
+def do_radamsa_or_ml_rnn_generator():
+  """Return whether to use radamsa or ml rnn generator.
+
+  True for radamsa, false for ml rnn."""
```
mbarbella-chromium: This function should be differentiated from the others somehow without following the usual do_* format. I also think this return value is confusing, could you instead return a dict here and call strategy_pool.update(return_value_of_this_function) below?

src/python/bot/fuzzers/libFuzzer/strategy_selection.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def do_fork():
+  """Return whether or not to do fork mode."""
+  return engine_common.decide_with_probability(
+      engine_common.get_strategy_probability(
+          strategy.FORK_STRATEGY, default=FORK_PROBABILITY))
+
+
+def do_mutator_plugin():
```
mbarbella-chromium: These helper functions actually don't seem to useful to me now that they've been normalized. Could you just have a single function that you use for all of these simple cases that takes the strategy and probability as arguments?

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_strategy_selection_test.py
```diff
@@ -0,0 +1,53 @@
...
+"""Tests for strategy selection file."""
+
+import unittest
+
+from bot.fuzzers import strategy
+from bot.fuzzers.libFuzzer import strategy_selection
+from tests.test_libs import helpers as test_helpers
+
+
+class TestStrategySelectionGeneratorPatch(unittest.TestCase):
```
mbarbella-chromium: Nit: Patched rather than Patch

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_strategy_selection_test.py
```diff
@@ -0,0 +1,53 @@
...
+    test_helpers.patch_environ(self)
+    test_helpers.patch(self,
+                       ['bot.fuzzers.engine_common.decide_with_probability'])
+    self.mock.decide_with_probability.return_value = True
+
+  def test_random_pool_generator(self):
+    """Deterministically tests the random strategy pool generator."""
+    strategy_pool = strategy_selection.generate_strategy_pool()
+    self.assertTrue(strategy_pool[strategy.CORPUS_MUTATION_RADAMSA_STRATEGY])
+    self.assertFalse(strategy_pool[strategy.CORPUS_MUTATION_ML_RNN_STRATEGY])
```
mbarbella-chromium: Could you explain these first two with a comment? Since this depends on the implementation, it would be good to document why we expect ml rnn to be False instead of radamsa in case someone updates the code later and wants to know if changing the order actually broke something or if they just need to update the test.

mbarbella-chromium:
Jonathan, would you mind taking a look at this as well? It would be good to get a fresh set of eyes on it.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_strategy_selection_test.py
mukundv-chrome: Sure, but not sure I understand. Changing the order of what?

mbarbella-chromium: Discussed offline, the current wording LGTM.

inferno-chromium:
adding @oliverchang for review since he will be planning fuzzing engine refactoring in some time. want to make sure that this wont cause any issues.

src/python/bot/fuzzers/libFuzzer/launcher.py
```diff
@@ -880,32 +804,42 @@ def main(argv):
...
-    arguments.append('%s%d' % (constants.FORK_FLAG, num_fuzz_processes))
-    fuzzing_strategies.append(
-        '%s_%d' % (strategy.FORK_STRATEGY, num_fuzz_processes))
+  # TODO(crbug.com/920355): Reenable this when fork mode works with ChromeOS's
+  # MSAN.
+  if strategy_pool[strategy.FORK_STRATEGY]:
+    job_name = environment.get_value('JOB_NAME')
+    memory_tool = environment.get_memory_tool_name(job_name)
+
+    if memory_tool != 'MSAN' or (not environment.is_chromeos_system_job()):
```
jonathanmetzman: Fork mode should work with MSAN now, can you remove this comparison.

src/python/bot/fuzzers/libFuzzer/launcher.py
```diff
@@ -880,32 +804,42 @@ def main(argv):
...
     arguments.append(constants.VALUE_PROFILE_ARGUMENT)
     fuzzing_strategies.append(strategy.VALUE_PROFILE_STRATEGY)
 
-  if do_fork():
-    max_fuzz_threads = environment.get_value('MAX_FUZZ_THREADS', 1)
-    num_fuzz_processes = max(1, multiprocessing.cpu_count() // max_fuzz_threads)
-    arguments.append('%s%d' % (constants.FORK_FLAG, num_fuzz_processes))
-    fuzzing_strategies.append(
-        '%s_%d' % (strategy.FORK_STRATEGY, num_fuzz_processes))
+  # TODO(crbug.com/920355): Reenable this when fork mode works with ChromeOS's
```
jonathanmetzman: This comment can be removed per my comment below, but FYI this should have probably been placed right before the if memory_tool != 'MSAN'

src/python/bot/fuzzers/libFuzzer/launcher.py
```diff
@@ -847,10 +765,16 @@ def main(argv):
     if os.path.exists(default_dict_path):
       arguments.append(constants.DICT_FLAG + default_dict_path)
 
+  # Strategy pool is the list of strategies that we attempt to
```
jonathanmetzman: I can't tell for sure on github but I think the formatting of these comments is odd. Can you try to fit as many words per line as possible (up to 80 chars)? Thanks!

src/python/bot/fuzzers/libFuzzer/launcher.py
jonathanmetzman: No need for paren around second clause.

src/python/bot/fuzzers/libFuzzer/strategy_selection.py
```diff
@@ -0,0 +1,105 @@
...
+
+FORK_PROBABILITY = 0.50
+
+MUTATOR_PLUGIN_PROBABILITY = 0.50
+
+
+def choose_generator():
+  """Return whether to use radamsa, ml rnn, or no generator."""
+
+  radamsa_p = engine_common.get_strategy_probability(
```
jonathanmetzman: maybe `_prob` or `_probability` instead of instead of `_p`?
We tend not to abbreviate things in naming in CF.

src/python/bot/fuzzers/libFuzzer/strategy_selection.py
```diff
@@ -0,0 +1,105 @@
...
+generated here may be modified in the launcher before being launched."""
+
+from bot.fuzzers import engine_common
+from bot.fuzzers import strategy
+
+# Probability of using `-max_len` option. Not applicable if already declared in
+# .options file.
+RANDOM_MAX_LENGTH_PROBABILITY = 0.15
+
+# Probability of doing ML RNN mutations on the corpus in this run.
```
jonathanmetzman: @mbarbella-chromium 
Do you think these comments are worth keeping? They don't seem useful to me (except in cases where they do more than explain what's clear, like on line 23.

src/python/bot/fuzzers/libFuzzer/strategy_selection.py
```diff
@@ -0,0 +1,105 @@
...
+  """Return dictionary representing a random selection of strategies
+  for launcher to consider."""
+  strategy_pool = {}
+
+  # Decide whether or not to include radamsa, ml rnn, or no generator
+  strategy_pool.update(choose_generator())
+
+  # Decide whether or not to include remaining strategies
+  strategy_pool[
+      strategy.CORPUS_SUBSET_STRATEGY] = engine_common.do_corpus_subset()
```
jonathanmetzman: Since we're refactoring this anyway can a strategy be an object with a name, probability and key as attributes. I don't like having three different constants like we do now.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_strategy_selection_test.py
```diff
@@ -0,0 +1,57 @@
...
+    self.assertTrue(strategy_pool[strategy.MUTATOR_PLUGIN_STRATEGY])
+
+
+class TestStrategySelectionPatchless(unittest.TestCase):
+  """Tests to see whether a strategy pool is properly generated by the file."""
+
+  def test_strategy_pool_generator(self):
+    """Ensures that a call to generate_strategy_pool does not yield an
+    exception."""
+    strategy_selection.generate_strategy_pool()
```
jonathanmetzman: I think this should test some behavior other than just "doesn't except"

src/python/bot/fuzzers/libFuzzer/strategy_selection.py
mbarbella-chromium: Nope. Definitely seems like it would be better to remove them and put all of the _PROBABILITY constants in one block.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_strategy_selection_test.py
mukundv-chrome: It was before, but we wanted the tests to be more deterministic because the pools generated in this case would be totally random. The other test is patched and tests some deterministic behavior

src/python/bot/fuzzers/libFuzzer/strategy_selection.py
mukundv-chrome: What did you mean by this? Right now I thought we are only keeping track of a name and probability for each strategy. Did you mean to move the probability constants into the strategy file?

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_strategy_selection_test.py
jonathanmetzman: Can you explain this in a comment then? thanks!

src/python/bot/fuzzers/libFuzzer/launcher.py
```diff
@@ -880,32 +803,35 @@ def main(argv):
...
-        '%s_%d' % (strategy.FORK_STRATEGY, num_fuzz_processes))
+        '%s_%d' % (strategy.FORK_STRATEGY.name, num_fuzz_processes))
 
   extra_env = {}
-  if do_mutator_plugin():
-    if use_mutator_plugin(target_name, extra_env, minijail_chroot):
-      fuzzing_strategies.append(strategy.MUTATOR_PLUGIN_STRATEGY)
+  # TODO(metzman): Support Windows.
+  if (strategy_pool[strategy.MUTATOR_PLUGIN_STRATEGY.name] and
+      environment.platform() != 'WINDOWS' and
```
oliverchang: nit: move this Windows check (and corresponding comment) into use_mutator_plugin.

src/python/bot/fuzzers/libFuzzer/launcher.py
```diff
@@ -880,32 +803,35 @@ def main(argv):
...
+  if strategy_pool[strategy.RANDOM_MAX_LENGTH_STRATEGY.name]:
+    max_len_argument = fuzzer_utils.extract_argument(
+        arguments, constants.MAX_LEN_FLAG, remove=False)
+    if not max_len_argument:
+      max_length = random.SystemRandom().randint(1, MAX_VALUE_FOR_MAX_LENGTH)
+      arguments.append('%s%d' % (constants.MAX_LEN_FLAG, max_length))
+      fuzzing_strategies.append(strategy.RANDOM_MAX_LENGTH_STRATEGY.name)
 
-  if do_recommended_dictionary():
+  if strategy_pool[strategy.RECOMMENDED_DICTIONARY_STRATEGY.name]:
```
oliverchang: nit:
```
if (strategy_pool... and
     add_recommended_dictionary...):
```

src/python/bot/fuzzers/libFuzzer/launcher.py
```diff
@@ -847,10 +765,15 @@ def main(argv):
     if os.path.exists(default_dict_path):
       arguments.append(constants.DICT_FLAG + default_dict_path)
 
+  # Strategy pool is the list of strategies that we attempt to enable, whereas
```
oliverchang: nit: Using "list" here to describe "Strategy pool" is a little misleading.

src/python/bot/fuzzers/libFuzzer/strategy_selection.py
```diff
@@ -0,0 +1,82 @@
...
+  """Return dictionary representing a random selection of strategies
+  for launcher to consider."""
+  strategy_pool = {}
+
+  # Decide whether or not to include radamsa, ml rnn, or no generator
+  strategy_pool.update(choose_generator())
+
+  # Decide whether or not to include remaining strategies
+  strategy_pool[
+      strategy.CORPUS_SUBSET_STRATEGY.name] = engine_common.do_corpus_subset()
```
oliverchang: Having to use .name everywhere is a little clunky, How about just using the entire strategy as a key? 
Also, using a dict to store booleans doesn't seem like the best fit when you can just use a set. 

A even better way might be to have something like:

```python
class StrategyPool(object):

  def __init__(self):
    self.strategy_names = set()
  
  def do(strategy):
    return strategy.name in self.strategy_names
```

mukundv-chrome: Sounds good, I think this looks cleaner too I'll look at it tomorrow

mukundv-chrome: looks a lot cleaner now, thanks!

src/python/bot/fuzzers/libFuzzer/launcher.py
mukundv-chrome: Still think that we can come up with a better word for this than list because even list might be misleading. Going to keep strategy pool until we have a better way to distinguish chosen and implemented strategies

PR #581, opened 2019-06-18
Add libFuzzer launcher test for the fork mode strategy.

Dor1s:


src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
```diff
@@ -1804,6 +1804,177 @@ def test_fuzz_with_mutations_using_ml_rnn(self, mock_execute, *_):
           '/fake/inputs-disk/temp-1337/mutations',
       ]])
 
+  @mock.patch('bot.fuzzers.libFuzzer.launcher.add_recommended_dictionary',
+              lambda x, y, z: False)
+  @mock.patch('multiprocessing.cpu_count', return_value=2)
+  @mock.patch('sys.stdout', new_callable=test_utils.MockStdout)
+  def test_fuzz_with_fork_mode(self, mock_stdout, *_):
```
jonathanmetzman: Thanks!

What is this testing btw? that -fork=2 is used on the command line and that a strategy is included in stats?

This isn't a problem so much with this test specifically, it is a problem with the tests in this file in general: Many of them test the same behavior: such as general stats parsing in addition to what they are intended to test. This makes the file a nightmare to deal with when making changes to stats as each test here needs to be changed.

Instead of assertingEqual on all of stats, can you please assertIn for strategy_fork? And could you leave a TODO saying this should be done for all other tests here? Not sure if we should do this for the assertion on popen.commands as well.

Thanks!

Dor1s: Great point! Filed https://github.com/google/clusterfuzz/issues/594 to track this, maybe let's discuss in the team meeting more.

Dor1s: > What is this testing btw? that -fork=2 is used on the command line and that a strategy is included in stats?

Right, it tests that the certain flags are added when the strategy is selected, ensures that those flags are removed when doing the merge step, and verifies that the stats are dumped correctly.

I share your opinion regarding the current structure of the tests (see my previous comment), but it's better to follow the currently used style (even if it's not perfect) and keep the file consistent until we refactor it as a whole.

jonathanmetzman: >but it's better to follow the currently used style (even if it's not perfect) 

It's not a style, it's a bad practice (like using strcpy) that probably was not picked intentionally (like a style) and ended up becoming more common through copy pasting. I don't really see value in being consistent here, fixing the behavior here would have been better IMO than consistency.

PR #582, opened 2019-06-18
Add support for auxiliary builds and use it for dataflow builds (#503).

Dor1s:
This CL implements `AuxiliaryBuild` class that should be helpful for doing fuzzing (or other tasks) with an additional build. Currently, the only use of the auxiliary build functionality is for DFSan instrumented builds which can be specified for libFuzzer jobs. This CL enables it as well, plus adds various tests for the new functionality.


src/python/build_management/build_manager.py
```diff
@@ -219,7 +223,7 @@ def _unpack_build(base_build_dir, build_dir, build_url, target_weights=None):
     return False
 
   unpack_everything = environment.get_value('UNPACK_ALL_FUZZ_TARGETS_AND_FILES')
-  if not unpack_everything:
+  if not unpack_everything and not is_auxiliary_build:
```
oliverchang: If we always unpack everything, there will be disk issues. Let's change the logic of _set_random_fuzz_target_for_fuzzing_if_needed to use the existing FUZZ_TARGET, if it exists. No need for this is_auxiliary_build argument.

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1242,6 +1243,55 @@ def key_fn(crash):
   return new_crash_count, known_crash_count, processed_groups
 
 
+def setup_auxiliary_build():
+  """Sets up auxiliary builds when necessary (e.g. DFSan build)."""
```
oliverchang: please see comment in build_manager.py re refactoring and using RegularBuild instead.

src/python/build_management/build_manager.py
```diff
@@ -984,6 +993,44 @@ def delete(self):
...
 
 
+class AuxiliaryBuild(Build):
+  """An additional build that accompanies some other fuzzinng build."""
+
+  def __init__(self,
+               base_build_dir,
+               revision,
+               build_url,
+               build_dir_name='revisions'):
```
oliverchang: is it OK to use the same build dir name? will there be any clashes?

Dor1s: I was thinking that `_get_build_directory` would take care of this and generate a different hash, but now I realize that it might become a problem if fuzzing and auxiliary builds would have the same filename. Thanks, will change.

src/python/build_management/build_manager.py
Dor1s: My intention was to never unpack everything for auxiliary builds, as the fuzz target must be chosen before that. I even had an explicit assertion for that, but ended up removing it when was writing the tests. So, right now the code would not exit and would unpack everything instead, but I don't think that might happen in reality.

src/python/build_management/build_manager.py
```diff
@@ -984,6 +993,44 @@ def delete(self):
     raise BuildManagerException('Cannot delete system build.')
 
 
+class AuxiliaryBuild(Build):
```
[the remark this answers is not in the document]
Dor1s: I thought about that initially, but then it looked too hacky to me, as it seemed like we would have to add some branching / special handling in several places (e.g. `_pre_setup`, `build_update` logic, `_setup_application_path`). Thinking long term, we may have other use cases for auxiliary builds (e.g. a coverage-instrumented build, or some new fuzzing engine that needs different builds), so I thought that those use cases would complicate a `RegularBuild` even further, while with `AuxiliaryBuild` we have more flexibility and lesser risk of landing a breaking CL, so I pivoted to a separate build type.

RPC is kinda annoying to add, but it looks like it's done just once and shouldn't require frequent modifications. Maybe I misunderstood this part?

src/python/bot/tasks/fuzz_task.py
Dor1s: Ack, replied!

src/python/build_management/build_manager.py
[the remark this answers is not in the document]
oliverchang: I don't think this will look too hacky if we plumb things properly. In this case, the only change required should be to plumb APP_PATH/APP_DIR/APP_REVISION etc by adding them as arguments?

If we have the ability to pass them as argument, we don't need any hacks in `_pre_setup` etc, because we're not overwriting anything that we care about in the previously set up build. Rather than making things more hacky, this is making RegularBuild more general and useful.

This will enable more auxiliary builds in the future also, and removing this hardcoding on the global BUCKET_PATH/APP_PATH/APP_DIR will make build_manager much more flexible in the future. Right now there is a lot of duplicated logic.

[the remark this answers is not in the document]
Dor1s: Sounds good, thanks for clarifications! Will give it a shot.

src/python/build_management/build_manager.py
oliverchang: Ok I see, _get_file_match_callback() will still return the right matcher.

In this case I think we can remove this auxiliary argument, and simply change the logic of _set_random_fuzz_target_for_fuzzing_if_needed to do nothing if FUZZ_TARGET is already set.

src/python/build_management/build_manager.py
```diff
@@ -487,32 +493,37 @@ def delete(self):
...
+  def _delete_partial_build_file(self):
+    """Deletes partial build file (if present). This is needed to make sure we
+    clean up build directory if the previous build was partial."""
+    partial_build_file_path = os.path.join(self.build_dir, PARTIAL_BUILD_FILE)
+    if os.path.exists(partial_build_file_path):
+      self.delete()
+
   def _pre_setup(self):
     """Common pre-setup."""
     self._reset_cwd()
```
Dor1s: These two calls, as well as `_setup_build_directories` below aren't needed for an auxiliary build, but seem harmless. I'd rather avoid calling those anyway, but suspect that you might be against adding `is_auxiliary_build` flag to the `Build` class as well. Let me know what you think :)

oliverchang: Yes, let's just keep them as is. We should avoid creating special cases.

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1242,6 +1243,44 @@ def key_fn(crash):
...
+      continue
+
+    revision = revisions.convert_revision_to_integer(match.group(1))
+    break
+
+  if revision is None:
+    logs.log_error('Unable to find a matching revision.')
+    return None
+
+  return build_manager.setup_regular_build(
```
oliverchang: check for failure and log error here?

src/python/build_management/build_manager.py
```diff
@@ -487,32 +493,37 @@ def delete(self):
 class Build(BaseBuild):
   """Repesents a build type at a particular revision."""
 
-  def __init__(self, base_build_dir, revision):
+  def __init__(self, base_build_dir, revision, build_prefix='APP'):
```
oliverchang: why not make build_prefix='' by default? and make prefix prepend to existing APP_* env vars?

src/python/build_management/build_manager.py
```diff
@@ -692,7 +715,11 @@ def build_dir(self):
   def setup(self):
     """Sets up build with a particular revision."""
     self._pre_setup()
-    environment.set_value('BUILD_URL', self.build_url)
+    if self.is_auxiliary_build:
```
oliverchang: why not key this off self.build_prefix? then we can avoid is_auxiliary_build. 

Also this can be made simpler if we make self.build_prefix = '' by default, and keep the 'APP_' parts while prepending.

src/python/bot/tasks/fuzz_task.py
Dor1s: Ok, done! Didn't get a chance to do a thorough self-review today, so the PR might not be production ready, but PTAL if you get a chance :)

Dor1s:
Looks like Travis having troubles downloading zlib, but the tests have passed so we're good. I've resolved all the comments (as they are addressed now) and rebased onto fresh master. PTAL @oliverchang :)

src/python/build_management/build_manager.py
```diff
@@ -709,7 +743,13 @@ def setup(self):
       # setting application path environment variables.
       logs.log('Build already exists.')
 
-    self._setup_application_path(build_update=build_update)
+    if self.build_prefix:
+      # A build prefix typically means an auxiliary build. We should not call
+      # _setup_application_path and overwrite multiple env vars in such case.
+      environment.set_value('%s_BUILD_DIR' % self.build_prefix, self.build_dir)
+    else:
+      self._setup_application_path(build_update=build_update)
```
oliverchang: why not call this? would be nice to have PREFIX_APP_DIR etc set.

src/python/build_management/build_manager.py
```diff
@@ -1111,84 +1151,88 @@ def get_revisions_list(bucket_path, testcase=None):
   return revision_list
 
 
-def setup_trunk_build():
+def setup_trunk_build(bucket_paths_env_vars=DEFAULT_BUILD_BUCKET_PATH_ENV_VARS,
```
oliverchang: side node: would be even better if we remove the dependency on global (env) vars being set in this function, and get callers to supply bucket paths, but can leave for future cleanup :)

src/python/build_management/build_manager.py
```diff
@@ -1111,84 +1151,88 @@ def get_revisions_list(bucket_path, testcase=None):
...
-  release_build_urls = get_build_urls_list(release_build_bucket_path)
-  if not release_build_urls:
-    logs.log_error('Error getting list of release build urls from %s.' %
-                   release_build_bucket_path)
+    urls_list = get_build_urls_list(bucket_path)
+    if not urls_list:
+      logs.log_error('Error getting list of build urls from %s.' % bucket_path)
+      return None
+
+    build_urls.append({'bucket_path': bucket_path, 'urls_list': urls_list})
```
oliverchang: nit: use a NamedTuple for these instead of a dict for each entry.

src/python/build_management/build_manager.py
```diff
@@ -1111,84 +1151,88 @@ def get_revisions_list(bucket_path, testcase=None):
...
 
-  release_build_url = revisions.find_build_url(release_build_bucket_path,
-                                               release_build_urls, revision)
-  if not release_build_url:
+  build_url = revisions.find_build_url(bucket_path, build_urls, revision)
+  if not build_url:
     logs.log_error(
         'Error getting build url for job %s (r%d).' % (job_type, revision))
 
+    if build_prefix:
```
oliverchang: why is this condition needed here?

src/python/build_management/build_manager.py
```diff
@@ -692,7 +722,11 @@ def build_dir(self):
   def setup(self):
     """Sets up build with a particular revision."""
     self._pre_setup()
-    environment.set_value('BUILD_URL', self.build_url)
+    if self.build_prefix:
+      environment.set_value('%s_BUILD_URL' % self.build_prefix, self.build_url)
+    else:
+      environment.set_value('BUILD_URL', self.build_url)
```
oliverchang: nit: consider having self.env_prefix and use here and everywhere else that does similar things.

src/python/build_management/build_manager.py
Dor1s: Let me try this (and something else) in a separate CL.

src/python/build_management/build_manager.py
Dor1s: Good idea, will try in a follow up CL with a few other changes that are a bit risky.

src/python/build_management/build_manager.py
Dor1s: Done!

PR #583, opened 2019-06-18
Add dataflow tracing strategy and refactor strategy parsing code (#503).

Dor1s:


src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
```diff
@@ -1804,6 +1817,348 @@ def test_fuzz_with_mutations_using_ml_rnn(self, mock_execute, *_):
...
+    with mock.patch(
+        'subprocess.Popen',
+        create_mock_popen(
+            self.no_crash_output, '/fake/inputs-disk/temp-1337/new',
+            '/fake/main_corpus_dir', new_units_added)) as mock_popen:
+      launcher.main([
+          'launcher.py',
+          '/fake/testcase_basic',
+          'fake_fuzzer',
+          '-max_len=80',
```
jonathanmetzman: Any reason to use -max_len=80 here?

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
```diff
@@ -1804,6 +1817,348 @@ def test_fuzz_with_mutations_using_ml_rnn(self, mock_execute, *_):
...
+          '-b',
+          '/lib,/lib,0',
+          '-b',
+          '/lib64,/lib64,0',
+          '-b',
+          '/usr/lib,/usr/lib,0',
+          '-b',
+          '/fake/build_dir,/fake/build_dir,0',
+          '-b',
+          '/fake/dfsan_build,/fake/dfsan_build,0',
```
jonathanmetzman: Again maybe test for inclusion rather than equality.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
```diff
@@ -1804,6 +1817,348 @@ def test_fuzz_with_mutations_using_ml_rnn(self, mock_execute, *_):
...
+
+    os.environ['FUZZ_CORPUS_DIR'] = '/fake/corpus_dir'
+    os.environ['DATAFLOW_BUILD_DIR'] = '/fake/dfsan_build'
+
+    new_units_added = 11
+    with mock.patch(
+        'subprocess.Popen',
+        create_mock_popen(
+            self.no_crash_output, '/fake/inputs-disk/temp-1337/new',
+            '/fake/main_corpus_dir', new_units_added)) as mock_popen:
```
jonathanmetzman: Why call this main_corpus_dir here when it is just corpus_dir above?

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
```diff
@@ -1804,6 +1817,348 @@ def test_fuzz_with_mutations_using_ml_rnn(self, mock_execute, *_):
...
+
+    self.fs.CreateDirectory('/fake/corpus_dir')
+    self.fs.CreateFile('/fake/testcase_basic')
+    self.fs.CreateDirectory('/fake/dfsan_build')
+    self.fs.CreateFile('/fake/dfsan_build/fake_fuzzer')
+
+    os.environ['FUZZ_CORPUS_DIR'] = '/fake/corpus_dir'
+    os.environ['DATAFLOW_BUILD_DIR'] = '/fake/dfsan_build'
+
+    new_units_added = 11
```
jonathanmetzman: Lines 1837-1848 are basically duplicated on lines (in original file, not patched): 1363, 1436, 1493, 1614, 1722, 1769, and once more in this patch on line 2026. Can the code be shared instead?

I realize this patch isn't doing anything wrong given the state of this file, but I think the file could be improved quite a bit.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
```diff
@@ -1804,6 +1817,348 @@ def test_fuzz_with_mutations_using_ml_rnn(self, mock_execute, *_):
...
+          'Command: /fake/build_dir/fake_fuzzer '
+          '-max_len=80 -rss_limit_mb=2048 -timeout=25 '
+          '-collect_data_flow=/fake/dfsan_build/fake_fuzzer -fork=1 '
+          '-artifact_prefix=/fake/ -max_total_time=2650 -print_final_stats=1 '
+          '/fake/inputs-disk/temp-1337/new /fake/corpus_dir')
+      self.assertIn(expected_command, mock_stdout.getvalue())
+      self.assertIn('Bot: test-bot', mock_stdout.getvalue())
+      self.assertIn('Time ran:', mock_stdout.getvalue())
+
+  @mock.patch('bot.fuzzers.libFuzzer.launcher.add_recommended_dictionary',
```
jonathanmetzman: I think we should use test_helpers.patch in setUp, instead of repeating the same patch decorators for add_recommended_dictionary, cpu_count and stdout.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
```diff
@@ -1804,6 +1817,348 @@ def test_fuzz_with_mutations_using_ml_rnn(self, mock_execute, *_):
...
+            self.no_crash_output, '/fake/inputs-disk/temp-1337/new',
+            '/fake/main_corpus_dir', new_units_added)) as mock_popen:
+      launcher.main([
+          'launcher.py',
+          '/fake/testcase_basic',
+          'fake_fuzzer',
+          '-max_len=80',
+      ])
+
+      self.assertEqual(mock_popen.commands, [[
```
jonathanmetzman: 2039-2069 seems to be duplicated across minijail tests, can you please find a way to abstract this assertion to avoid this?

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
```diff
@@ -1804,6 +1817,348 @@ def test_fuzz_with_mutations_using_ml_rnn(self, mock_execute, *_):
...
+          '-S',
+          'mknod',
+          '-m',
+          '666',
+          '/fake/inputs-disk/temp-1337/CHROOT/dev/urandom',
+          'c',
+          '1',
+          '9',
+      ], [
+          '/fake_root/resources/platform/{}/minijail0'.format(
```
jonathanmetzman: `os.path.join('/fake_root/resources/platform', environment.platform().lower(), `minijail0')`seems preferable to calling format here.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
```diff
@@ -1804,6 +1817,348 @@ def test_fuzz_with_mutations_using_ml_rnn(self, mock_execute, *_):
...
+          '-timeout=25',
+          '-collect_data_flow=/fake/dfsan_build/fake_fuzzer',
+          '-fork=1',
+          '-artifact_prefix=/',
+          '-max_total_time=2650',
+          '-print_final_stats=1',
+          '/new',
+          '/corpus_dir',
+      ], [
+          '/fake_root/resources/platform/{}/minijail0'.format(
```
jonathanmetzman: Same comment about format.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
```diff
@@ -1804,6 +1817,348 @@ def test_fuzz_with_mutations_using_ml_rnn(self, mock_execute, *_):
...
+      launcher.main([
+          'launcher.py',
+          '/fake/testcase_basic',
+          'fake_fuzzer',
+          '-max_len=80',
+      ])
+
+      self.assertEqual(mock_popen.commands, [[
+          '/fake/build_dir/fake_fuzzer', '-max_len=80', '-rss_limit_mb=2048',
+          '-timeout=25', '-collect_data_flow=/fake/dfsan_build/fake_fuzzer',
```
jonathanmetzman: Can we check that `'-collect_data_flow=/fake/dfsan_build/fake_fuzzer'` is in the commands rather than compare against all commands?

src/python/bot/fuzzers/libFuzzer/constants.py
```diff
@@ -22,6 +22,8 @@
 # libFuzzer flags.
```
jonathanmetzman: One day we should make strategies an object that have a flag, strategy_name and libfuzzer argument.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
```diff
@@ -1804,6 +1817,348 @@ def test_fuzz_with_mutations_using_ml_rnn(self, mock_execute, *_):
...
+              u'timeout_count':
+                  0,
+              u'timeout_limit':
+                  25,
+              'timestamp':
+                  1337.0
+          })
+
+      # Output printed.
+      self.assertIn(self.no_crash_output, mock_stdout.getvalue())
```
jonathanmetzman: I'm not sure what this assert and those on 2002 and 2003 are testing that isn't tested elsewhere, why do we need them?

src/python/bot/fuzzers/libFuzzer/launcher.py
```diff
@@ -197,6 +200,14 @@ def do_mutator_plugin():
           strategy.MUTATOR_PLUGIN_STRATEGY, default=MUTATOR_PLUGIN_PROBABILITY))
 
 
+def do_dataflow_tracing():
```
mbarbella-chromium: Just a heads up, this will likely conflict with https://github.com/google/clusterfuzz/pull/578 (I wouldn't block on that change since it should be easy enough to resolve, just an FYI).

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
Dor1s: Sure, but better refactor all the tests at once. Or maybe even re-write them if they won't really work after the fuzzing engine pluggability work.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
Dor1s: Agree, it's redundant, but these lines likely will go away as we re-invent the launcher interface. Not sure it's worth investing in refactoring at this point.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
```diff
@@ -1804,6 +1817,348 @@ def test_fuzz_with_mutations_using_ml_rnn(self, mock_execute, *_):
...
+          '-rss_limit_mb=2048',
+          '-timeout=25',
+          '-merge=1',
+          '/fake/inputs-disk/temp-1337/merge-corpus',
+          '/fake/inputs-disk/temp-1337/new',
+          '/fake/corpus_dir',
+      ]])
+
+      # Check new testcases added.
+      self.assertEqual(
```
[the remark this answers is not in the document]
Dor1s: That we did communicate with popen and it was the fuzzing command (not merge), therefore the expected number of testcases was written to the fakefs.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
```diff
@@ -1804,6 +1817,348 @@ def test_fuzz_with_mutations_using_ml_rnn(self, mock_execute, *_):
...
+          sorted(os.listdir('/fake/corpus_dir')),
+          sorted(mock_popen.testcases_written))
+
+      for corpus_file in os.listdir('/fake/corpus_dir'):
```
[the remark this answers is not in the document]
Dor1s: That every fake testcase we've written has the content equal to its name.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
```diff
@@ -1804,6 +1817,348 @@ def test_fuzz_with_mutations_using_ml_rnn(self, mock_execute, *_):
...
+              u'slow_unit_count':
+                  0,
+              u'slow_units_count':
+                  0,
+              u'slowest_unit_time_sec':
+                  0,
+              u'startup_crash_count':
+                  0,
+              u'strategy_dataflow_tracing':
+                  0,
```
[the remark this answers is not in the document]
Dor1s: Hmmm, that's a wonderful catch. Thanks a lot! I've refactored how we parse the strategies and also added an assertion that should reduce the risk a newly added strategy not being parsed correctly in stats.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
Dor1s: Just testing end-to-end, I suppose :)

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
Dor1s: Not every test in this class patches those, so not sure about moving everything to the `setUp`. Added a note to #594.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
Dor1s: Renamed.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
Dor1s: Can use any other number, just important to verify that it didn't change.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
Dor1s: Added to #594 for potential test refactoring.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
Dor1s: Other tests in this file are using `'{}'.format`.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
```diff
@@ -1804,6 +1817,348 @@ def test_fuzz_with_mutations_using_ml_rnn(self, mock_execute, *_):
...
+          '/fake/build_dir,/out,0',
+          '-b',
+          '/fake/inputs-disk/temp-1337/new,/new,1',
+          '-b',
+          '/fake/corpus_dir,/corpus_dir,1',
+          '/fake/build_dir/fake_fuzzer',
+          '-max_len=80',
+          '-rss_limit_mb=2048',
+          '-timeout=25',
+          '-collect_data_flow=/fake/dfsan_build/fake_fuzzer',
```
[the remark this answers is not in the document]
Dor1s: mentioned in #594 for a potential refactoring

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
Dor1s: We should not mix different styles in a single file / project. Refactoring of all the tests is beyond of scope of this CL and might not be worth doing given the upcoming fuzzing engine pluggability changes.

src/python/bot/fuzzers/libFuzzer/constants.py
jonathanmetzman: I think @mukundv-chrome did something like this in a recent CL.

Dor1s: The one that was landed had only names and probabilities, but didn't have flags / arguments.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
jonathanmetzman: Is there something about dataflow in particular that could change this?

Dor1s: There isn't anything that is expected to change this, and that is being verified by the test :)

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
jonathanmetzman: ` '{}'.format` is inconsistent with the rest of the codebase which uses `os.path.join`.
Please change this (and the others if you feel the inconsistency withing the file is important).

Dor1s: It's a bikeshedding at this point. My opinion regarding consistency is based on my previous software engineering experience. Looks like Google guidelines (http://google.github.io/styleguide/pyguide.html#4-parting-words) are mentioning something along this lines, but of course that's not a discussion I'd really like to spend time on.

If we want to refactor something, it's potentially a good task and should be done thoroughly in a separate CL. I've filed https://github.com/google/clusterfuzz/issues/594 to make sure your comments aren't lost and we'll eventually address those or close as WontFix.

Please let me know if you think there are any issues that are crucial for enabling the dataflow tracing strategy support.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
mbarbella-chromium: Chiming in after our offline discussion. I don't feel too strongly about how this is handled here. It seems reasonable to me to land this change with the same format. I don't think it would make a cleanup significantly more difficult to have one more occurrence, and I don't think there's much value in having only a single use with a different convention.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_test.py
jonathanmetzman: I don't think there's any reason to test that this hasn't been removed, so please take this out of the test.

oliverchang:
These tests will be rewritten anyway as part of the upcoming refactoring efforts, so trying to fix up things here is probably out of scope and not worth the effort. Seems OK to me to land for now!

PR #584, opened 2019-06-18
 Add OAuth configuration options for the new reproduce tool.

mbarbella-chromium:
PTAL.

### DIFF
--- a/src/appengine/libs/handler.py
+++ b/src/appengine/libs/handler.py
@@ -250,6 +250,7 @@ def oauth(func):
 
     header and getting user email.
   """
+
   @functools.wraps(func)
   def wrapper(self):
     """Wrapper."""
--- a/src/appengine/libs/handler.py
+++ b/src/appengine/libs/handler.py
@@ -153,29 +153,16 @@ def unsupported_on_local_server(func):
   return wrapper
 
 
-# TODO(mbarbella): Remove the |use_reproduce_tool_client| argument and related
-# plumbing once clusterfuzz-tools is removed. We need to support both while the
-# new tool is in development.
-def get_access_token(verification_code, use_reproduce_tool_client=False):
+def get_access_token(verification_code):
   """Get the access token from verification code.
 
     See: https://developers.google.com/identity/protocols/OAuth2InstalledApp
   """
-  if use_reproduce_tool_client:
-    client_id_config_key = 'reproduce_tool_client_id'
-  else:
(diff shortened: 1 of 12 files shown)


src/appengine/libs/handler.py
```diff
@@ -153,16 +153,29 @@ def wrapper(self):
   return wrapper
 
 
-def get_access_token(verification_code):
+# TODO(mbarbella): Remove the |use_reproduce_tool_client| argument and related
+# plumbing once clusterfuzz-tools is removed. We need to support both while the
```
inferno-chromium: We can just use same client id and secret for the new version and rename the attributes names and set their values when deployed. Since it is only used on internal, once deployed, you can instantly change the values.
s/clusterfuzz_tools_oauth_client_id/reproduce_tool_client_id
s/clusterfuzz_tools_client_secret/reproduce_tool_client_secret
remove use_reproduce_tool_client arg completely from all functions
deploy on staging to set reproduce_tool_client_id and reproduce_tool_client_secret manually. i am even thinking to put both client id and secret in configuration as it feels weird to put one in a yaml and other in admin config, so lets move both to admin config.

mbarbella-chromium: SGTM. For now I'll just reuse the clusterfuzz-tools client id/secret, and at some point in the future I'll switch it over.

src/go/cloud/db/types/types.go
```diff
@@ -105,6 +105,7 @@ type Config struct {
 	RelaxSecurityBugRestrictions        bool           `datastore:"relax_security_bug_restrictions"`
 	CoverageReportsBucket               string         `datastore:"coverage_reports_bucket"`
 	GithubCredentials                   string         `datastore:"github_credentials"`
+	ReproduceToolClientSecret           string         `datastore:"reproduce_tool_client_secret"`
```
inferno-chromium: Where is client id here ?

src/appengine/libs/handler.py
```diff
@@ -158,11 +158,11 @@ def get_access_token(verification_code):
 
     See: https://developers.google.com/identity/protocols/OAuth2InstalledApp
   """
```
inferno-chromium: Can you also see that reproduce_tool_client_id gets auto-added to whitelisted_oauth_client_ids, so that it is not needed to be listed there.

mbarbella-chromium: Done

PR #587, opened 2019-06-19
Print exceptions in run_one_test_parallel.

oliverchang:
Usually, an exception happens in a unit test, which is surfaced
properly. However, if we hit an exception in the underlying test runner
code, we get an exception without a stacktrace. This catches and prints
the stack before it is lost.

### DIFF
--- a/src/local/butler/py_unittest.py
+++ b/src/local/butler/py_unittest.py
@@ -32,6 +32,7 @@ import platform
 import signal
 import sys
 import time
+import traceback
 import unittest
 
 from local.butler import appengine
@@ -95,7 +96,7 @@ class MeasureCoverage(object):
   def __enter__(self):
     pass
 
-  def __exit__(self, exc_type, value, traceback):
+  def __exit__(self, exc_type, value, _):
     COV.stop()
 
     if not self.enabled:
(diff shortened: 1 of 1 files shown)


src/local/butler/py_unittest.py
```diff
@@ -128,20 +129,26 @@ def test_worker_init():
...
+
+    # Verbosity=0 since we cannot see real-time test execution order when tests
+    # are executed in parallel.
+    result = unittest.TextTestRunner(
+        stream=stream, verbosity=0, buffer=suppress_output).run(suite)
+
+    return TestResult(stream.getvalue(),
+                      len(result.errors), len(result.failures),
+                      len(result.skipped), result.testsRun)
+  except:
```
inferno-chromium: except Exception:

oliverchang: This was intentional, since not all exceptions are necessarily derived from Exception.

oliverchang: actually, BaseException probably covers all that we care about here (including system exiting ones). will use that

oliverchang:
Also added a fix for a unicode issue that @flowerhack was running into. /gcbrun

Issue #579, opened 2019-06-18
Incorrect issue title text with NULL, no crash params, fix tests


inferno-chromium:
Link: https://bugs.chromium.org/p/chromium/issues/detail?id=943254

Should we report it like "Indirect leak in $fuzz_target_name"?

inferno-chromium:
Will hold on this until issue tracker refactoring is done by @oliverchang

PR #593, opened 2019-06-20
Fix confusing issue title with NULL crash state.

inferno-chromium:
Fixes
https://github.com/google/clusterfuzz/issues/579

### DIFF
--- a/src/appengine/handlers/cron/cleanup.py
+++ b/src/appengine/handlers/cron/cleanup.py
@@ -744,8 +744,11 @@ def _get_severity_from_labels(security_severity_label, labels):
   return data_types.SecuritySeverity.MISSING
 
 
-def _get_security_severity_issue_comment(policy, testcase, issue):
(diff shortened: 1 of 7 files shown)


src/appengine/handlers/cron/cleanup.py
```diff
@@ -744,8 +744,8 @@ def _get_severity_from_labels(security_severity_label, labels):
   return data_types.SecuritySeverity.MISSING
 
 
-def _update_issue_severity_labels(policy, testcase, issue):
-  """Update severity labels on issue."""
+def _get_security_severity_issue_comment(policy, testcase, issue):
```
oliverchang: nit: maybe something like _update_issue_severity_and_get_comment() to make it clear that it's updating the severity on the issue too?

inferno-chromium: My bad, i read the function wrong and then did this rename. Fixed it now, thanks for catching.

PR #597, opened 2019-06-21
Add appengine.libs.request_cache.

oliverchang:
This is a cache that's scoped to a request. Use this to replace the
issue tracker cache.

### DIFF
--- a/src/python/base/memoize.py
+++ b/src/python/base/memoize.py
@@ -39,11 +39,22 @@ class FifoInMemory(object):
 
   def __init__(self, capacity):
     self.capacity = capacity
-    self.cache = collections.OrderedDict()
     self.lock = threading.Lock()
+    self._cache = None
+
+  @property
+  def cache(self):
+    """Get the cache backing. None may be returned."""
+    if self._cache is None:
+      self._cache = collections.OrderedDict()
+
+    return self._cache
 
   def put(self, key, value):
     """Put (key, value) into cache."""
+    if self.cache is None:
+      return
+
     # Lock to avoid race condition in popitem.
     self.lock.acquire()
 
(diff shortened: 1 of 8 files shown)


src/python/base/memoize.py
```diff
@@ -39,11 +39,22 @@ class FifoInMemory(object):
...
+  def cache(self):
+    """Get the cache backing. None may be returned."""
+    if self._cache is None:
+      self._cache = collections.OrderedDict()
+
+    return self._cache
 
   def put(self, key, value):
     """Put (key, value) into cache."""
+    if self.cache is None:
```
inferno-chromium: How can it be None if it is initialized to collections.OrderedDict() in cache property function.

src/python/tests/appengine/libs/request_cache_test.py
```diff
@@ -0,0 +1,106 @@
...
+        (2,),
+    ], c.called)
+
+  def test_no_request(self):
+    """Test no request available."""
+    self.mock.get_request.return_value = None
+    c = CacheClass()
+    self.assertEqual(2, c.foo(1))
+    self.assertEqual(2, c.foo(1))
+
```
inferno-chromium: do we want to self.assertEqual(3, c.foo(2)) with different arg as well

src/python/base/memoize.py
oliverchang: It can be None if it's overridden by a child class.

src/python/tests/appengine/libs/request_cache_test.py
oliverchang: Done!

src/python/base/memoize.py
inferno-chromium: Ok.

PR #598, opened 2019-06-21
Call an init function from config modules.

oliverchang:
Modules can define a "module_init" module containing a "appengine"
function which is called on App Engine startup.

Also expose a way to register issue tracker implementations.

### DIFF
--- a/src/appengine/libs/issue_management/issue_tracker_utils.py
+++ b/src/appengine/libs/issue_management/issue_tracker_utils.py
@@ -29,6 +29,9 @@ _ISSUE_TRACKER_CONSTRUCTORS = {
 
 def register_issue_tracker(tracker_type, constructor):
   """Register an issue tracker implementation."""
+  if tracker_type in _ISSUE_TRACKER_CONSTRUCTORS:
+    raise ValueError(
+        'Tracker type {type} is already registered.'.format(type=tracker_type))
   _ISSUE_TRACKER_CONSTRUCTORS[tracker_type] = constructor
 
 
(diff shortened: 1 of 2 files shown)


src/appengine/libs/issue_management/issue_tracker_utils.py
```diff
@@ -22,6 +22,14 @@
...
 
 _ISSUE_TRACKER_CACHE_CAPACITY = 8
+_ISSUE_TRACKER_CONSTRUCTORS = {
+    'monorail': monorail.get_issue_tracker,
+}
+
+
+def register_issue_tracker(tracker_type, constructor):
+  """Register an issue tracker implementation."""
+  _ISSUE_TRACKER_CONSTRUCTORS[tracker_type] = constructor
```
inferno-chromium: nit: Should we add something to skip registration if exists already ?

src/appengine/libs/issue_management/issue_tracker_utils.py
```diff
@@ -43,10 +51,12 @@ def get_issue_tracker(project_name=None):
   if not issue_project_config:
     raise ValueError('Issue tracker for {} does not exist'.format(project_name))
 
-  if issue_project_config['type'] == 'monorail':
-    return monorail.get_issue_tracker(project_name)
+  constructor = _ISSUE_TRACKER_CONSTRUCTORS.get(issue_project_config['type'])
+  if not constructor:
+    raise ValueError('Invalid issue tracker type ' +
```
inferno-chromium: nit: colon after type, type:

oliverchang: done.

src/appengine/libs/issue_management/issue_tracker_utils.py
oliverchang: made this raise an exception. should only be registered once.

PR #599, opened 2019-06-21
Minor changes to better support custom issue trackers.

oliverchang:
- Pass config to issue tracker constructor.
- Add a clear() method to LabelStore.

### DIFF
--- a/src/appengine/libs/issue_management/issue_tracker.py
+++ b/src/appengine/libs/issue_management/issue_tracker.py
@@ -74,6 +74,11 @@ class LabelStore(object):
     self._added.clear()
     self._removed.clear()
 
+  def clear(self):
+    """Clear the store, and reset tracking."""
+    self.reset()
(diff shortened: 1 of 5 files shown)


src/appengine/libs/issue_management/issue_tracker.py
```diff
@@ -74,6 +74,11 @@ def reset(self):
     self._added.clear()
     self._removed.clear()
 
+  def clear(self):
+    """Clear the store, and reset tracking."""
+    self.reset()
```
inferno-chromium: Do we need both reset and clear function. Can we clear backing in all cases. Just the two function names are similar and confusing and i wonder if people will incorrectly use one vs other.

oliverchang: Renamed reset to reset_tracking to make the difference clearer. We do need both.

Issue #600, opened 2019-06-21
Impact label updated on unreproducible crash


inferno-chromium:
See
https://bugs.chromium.org/p/chromium/issues/detail?id=977101#c5

PR #601, opened 2019-06-21
Fix incorrect impact update when testcase is unreproducible.

inferno-chromium:
When a testcase is unreproducible and does not impact stable
and beta branches, then impact label should not be added.
This fixes https://github.com/google/clusterfuzz/issues/600

### DIFF
--- a/src/appengine/libs/issue_management/issue_filer.py
+++ b/src/appengine/libs/issue_management/issue_filer.py
@@ -170,8 +170,12 @@ def update_issue_impact_labels(testcase, issue):
       new_impact = data_types.SecurityImpact.STABLE
     elif testcase.impact_beta_version:
       new_impact = data_types.SecurityImpact.BETA
-    else:
+    elif testcase.is_crash():
       new_impact = data_types.SecurityImpact.HEAD
+    else:
+      # Testcase is unreproducible and does not impact stable and beta branches.
+      # In this case, there is no impact information.
+      return
   else:
     # No impact information.
     return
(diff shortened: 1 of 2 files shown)


src/appengine/libs/issue_management/issue_filer.py
```diff
@@ -170,8 +170,12 @@ def update_issue_impact_labels(testcase, issue):
       new_impact = data_types.SecurityImpact.STABLE
     elif testcase.impact_beta_version:
       new_impact = data_types.SecurityImpact.BETA
-    else:
+    elif testcase.is_crash():
       new_impact = data_types.SecurityImpact.HEAD
+    else:
+      # Testcase is unreproducible and does not impact stable and beta branches.
+      # In this case, there is no impact information.
```
Dor1s: Should we apply `Security_Impact-None` in such case?

inferno-chromium: No that is when it is legit crash, that does not impact production (e.g. experimental feature). This None impact has always been confusing.

inferno-chromium: Even though we define data_types.SecurityImpact.NONE, we never use/apply it in code. i am even thinking to remove it to avoid confusion ?

inferno-chromium: I take the last comment back, which is we need to define it in code to detect current labels and change as needed.

Dor1s: Got it, thanks for clarifying.

PR #603, opened 2019-06-22
Enable impact label setting for component builds.

inferno-chromium:
We enabled impact calculation in impact task, but
forgot to remove this condition in issue_filer.


inferno-chromium:
TBR, minor followup to https://github.com/google/clusterfuzz/pull/450

PR #604, opened 2019-06-24
Use common get_original_issue implementation for all issue trackers (#504).

oliverchang:
Also, remove the merged_into setter, since it's not used by anything.

### DIFF
--- a/src/appengine/libs/issue_management/monorail/__init__.py
+++ b/src/appengine/libs/issue_management/monorail/__init__.py
@@ -75,11 +75,11 @@ class Issue(issue_tracker.Issue):
   @property
   def merged_into(self):
     """The issue that this is merged into."""
-    return self._monorail_issue.merged_into
+    if self._monorail_issue.merged_into_project != self.issue_tracker.project:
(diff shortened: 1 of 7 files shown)


src/appengine/libs/issue_management/monorail/__init__.py
```diff
@@ -234,10 +234,6 @@ def get_issue(self, issue_id):
 
     return Issue(monorail_issue)
 
-  def get_original_issue(self, issue_id):
-    """Retrieve the original issue object traversing the list of duplicates."""
-    return Issue(self._itm.get_original_issue(int(issue_id)))
```
inferno-chromium: Probably i can remove some of these dead/duplicate functions in monorail part of code.
E.g. https://github.com/google/clusterfuzz/blob/master/src/appengine/libs/issue_management/monorail/issue_tracker_manager.py#L305. wdyt ? i can do this when i am back.

oliverchang: Yep, sgtm!

PR #607, opened 2019-06-25
Disable Radamsa + ML RNN for LPM targets

jonathanmetzman:
Radamsa and ML RNN mutate testcases in a corpus.
More often than not, this makes items in the corpus invalid protos.
Since LF/LPM isn't smart enough not to merge these back into the corpus (I think because they reveal new coverage in the parser), the corpus gets (permanently?) gunked up with tescases that LPM can't mutate as a result. This takes time away from real testcases.

### DIFF
--- a/src/python/bot/fuzzers/engine_common.py
+++ b/src/python/bot/fuzzers/engine_common.py
@@ -157,8 +157,8 @@ def get_merge_timeout(default_merge_timeout):
 def is_lpm_fuzz_target(fuzzer_path):
   """Returns True if |fuzzer_path| is a libprotobuf-mutator based fuzz
   target."""
-  with open(fuzzer_path) as fuzzer_handle:
-    return utils.search_string_in_file('TestOneProtoInput', fuzzer_handle)
+  with open(fuzzer_path) as file_handle:
+    return utils.search_string_in_file('TestOneProtoInput', file_handle)
 
 
 def get_issue_owners(fuzz_target_path):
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
PTAL.

I suspect it's probably worth doing this for AFL as well since LPM targets are getting built for AFLin OSS-Fuzz.

src/python/bot/fuzzers/engine_common.py
```diff
@@ -154,6 +154,13 @@ def get_merge_timeout(default_merge_timeout):
                                  'MERGE_TIMEOUT_OVERRIDE')
 
 
+def is_lpm_fuzz_target(fuzzer_path):
+  """Returns True if |fuzzer_path| is a libprotobuf-mutator based fuzz
+  target."""
+  with open(fuzzer_path) as fuzzer_handle:
```
Dor1s: nit: maybe `file_handle` instead of `fuzzer_handle`?

mhlakhani:
Nice catch indeed!

> I suspect it's probably worth doing this for AFL as well since LPM targets are getting built for AFLin OSS-Fuzz.

To add to this - I don't think AFL has support for custom mutators so this problem should definitely exist there unless all the LPM targets are doing something special (and invoking LPM within `LLVMFuzzerTestOneInput`)? We also ran into this issue and had to disable AFL for our equivalent of LPM targets.

jonathanmetzman:
> Nice catch indeed!
> 
> > I suspect it's probably worth doing this for AFL as well since LPM targets are getting built for AFLin OSS-Fuzz.
> 
> To add to this - I don't think AFL has support for custom mutators so this problem should definitely exist there unless all the LPM targets are doing something special (and invoking LPM within `LLVMFuzzerTestOneInput`)? We also ran into this issue and had to disable AFL for our equivalent of LPM targets.

We disabled these on the build side in Chromium, but right now AFL and LPM are running on CF. Actually they've even found bugs. But I'll probably disable this in CF as well.

PR #615, opened 2019-06-28
Add support for running tests from config modules

oliverchang:


src/local/butler/py_unittest.py
```diff
@@ -281,8 +278,19 @@ def execute(args):
       (appengine_config)  # pylint: disable=pointless-statement
     except ImportError:
       print('Note: unable to import appengine_config.')
-  else:
+  elif args.target == 'core':
     test_directory = CORE_TEST_DIRECTORY
+  else:
+    os.environ['CONFIG_DIR_OVERRIDE'] = args.config_dir
```
inferno-chromium: Please add a comment that this is for the modules branch.

oliverchang: Done!

PR #624, opened 2019-06-28
 [Reproduce tool] Support libFuzzer.

mbarbella-chromium:


src/python/base/modules.py
```diff
@@ -90,15 +94,13 @@ def fix_module_search_paths():
     sys.path.insert(0, source_directory)
     python_path.insert(0, source_directory)
 
-  # Since paths have now been fixed, import the custom logs module.
-  from metrics import logs
```
mbarbella-chromium: This was causing some problems for me since it leads to some imports that need the fixes below to work properly. The new behavior seems a bit safer to me, but I'm not completely sure what was different about the paths when running from the tool compared to production. I can investigate more if needed.

mbarbella-chromium:
No rush on this, PTAL when you have time.

src/local/butler/reproduce.py
```diff
@@ -214,33 +218,50 @@ def _prepare_initial_environment(build_directory):
...
+  _copy_root_subdirectory(root_dir, temp_root_dir, 'src')
 
   environment.set_value('CONFIG_DIR_OVERRIDE',
                         os.path.join(temp_root_dir, 'configs', 'test'))
+  # TODO(mbarbella): Don't force the user to set APPENGINE_DIRECTORY explicitly.
+  environment.set_value(
+      'PYTHONPATH',
+      modules.get_pythonpath_separator().join([
+          os.path.join(temp_root_dir, 'src'),
+          environment.get_value('APPENGINE_DIRECTORY')
```
oliverchang: why is this needed?

src/local/butler/reproduce.py
```diff
@@ -214,33 +218,50 @@ def _prepare_initial_environment(build_directory):
   _copy_root_subdirectory(root_dir, temp_root_dir, 'bot')
```
oliverchang: would it make sense to share this functionality with the one in butler's run_bot ? Seems better to reduce the duplication here to make things more maintainable.

src/python/base/modules.py
```diff
@@ -90,15 +94,13 @@ def fix_module_search_paths():
...
-
   # Work around google package issues with App Engine SDK.
+  errors = []
```
oliverchang: Add a comment on why errors is needed.

src/python/base/modules.py
```diff
@@ -47,6 +47,14 @@ def _patch_appengine_modules_for_bots():
     pass
 
 
+def get_pythonpath_separator():
```
oliverchang: can we just use os.pathsep here?

src/local/butler/reproduce.py
mbarbella-chromium: Found a better way to handle this. There were issues with the import in the libFuzzer launcher's subprocess, but not in the main process. Adding the sdk path to PYTHONPATH instead of updating sys.path was a workaround for it.

PR #625, opened 2019-06-29
Ensure that fuzz target exists in DFSan build, log_warn otherwise (#503).

Dor1s:
Haven't seen this being a problem, but I'm sure it'll happen some day.

### DIFF
--- a/src/python/bot/fuzzers/libFuzzer/launcher.py
+++ b/src/python/bot/fuzzers/libFuzzer/launcher.py
@@ -845,7 +845,8 @@ def main(argv):
           '%s%s' % (constants.COLLECT_DATA_FLOW_FLAG, dataflow_binary_path))
       fuzzing_strategies.append(strategy.DATAFLOW_TRACING_STRATEGY.name)
     else:
-      logs.log_warn('Fuzz target is not found in dataflow build.')
+      logs.log_error(
+          'Fuzz target is not found in dataflow build, skiping strategy.')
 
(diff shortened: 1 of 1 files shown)


src/python/bot/fuzzers/libFuzzer/launcher.py
```diff
@@ -840,9 +840,12 @@ def main(argv):
...
         dataflow_build_dir, os.path.relpath(fuzzer_path, build_directory))
-    arguments.append(
-        '%s%s' % (constants.COLLECT_DATA_FLOW_FLAG, dataflow_binary_path))
-    fuzzing_strategies.append(strategy.DATAFLOW_TRACING_STRATEGY.name)
+    if os.path.exists(dataflow_binary_path):
+      arguments.append(
+          '%s%s' % (constants.COLLECT_DATA_FLOW_FLAG, dataflow_binary_path))
+      fuzzing_strategies.append(strategy.DATAFLOW_TRACING_STRATEGY.name)
+    else:
+      logs.log_warn('Fuzz target is not found in dataflow build.')
```
inferno-chromium: log_error as otherwise we will never see it on our dashboards.

also, can add ", skip strategy" in message.

Dor1s: Done and done. However, I'm not sure whether we want to see it on the error dashboard. Since we don't synchronize build revisions, we'll likely be hitting this from time to time, and we'll be just ignoring that error then. IMO such errors make the dashboard hard to navigate / triage :(

PR #626, opened 2019-07-01
Support LOAS auth.

oliverchang:
PTAL. This bypasses the need for firebase auth if there are LOAS creds available.

### DIFF
--- a/src/appengine/libs/auth.py
+++ b/src/appengine/libs/auth.py
@@ -59,6 +59,10 @@ def get_current_user():
   if environment.is_local_development():
     return User('user@localhost')
 
+  loas_user = environment.get_value('LOAS_PEER_USERNAME')
+  if loas_user:
+    return User(loas_user + '@google.com')
+
   current_request = get_current_request()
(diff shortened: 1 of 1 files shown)


src/appengine/libs/auth.py
```diff
@@ -59,6 +59,10 @@ def get_current_user():
   if environment.is_local_development():
     return User('user@localhost')
 
+  loas_user = environment.get_value('LOAS_PEER_USERNAME')
+  if loas_user:
+    return User(loas_user + '@google.com')
```
Dor1s: Should this always be `@google.com`? I suspect that we should move it to the config (or maybe there is some way to obtain the domain dynamically), but LGTM anyway as it shouldn't cause any issues at this point!

oliverchang: I believe this is specific to Google, so it does not make sense to make this configurable.

Dor1s: Ack, I though LOAS is yet another SSO protocol like SAML and others :)

PR #628, opened 2019-07-02
Fix broken issue link when a group bug is used + add tests.

Dor1s:
Noticed on https://clusterfuzz.com/testcase-detail/5103503034417152, the issue link points to `https://clusterfuzz.com/testcase-detail/null`.

### DIFF
--- a/src/appengine/libs/issue_management/issue_tracker_utils.py
+++ b/src/appengine/libs/issue_management/issue_tracker_utils.py
@@ -154,10 +154,17 @@ def get_similar_issues_url(issue_tracker, testcase, only_open=True):
 def get_issue_url(testcase):
   """Return issue url for a testcase."""
   issue_tracker = get_issue_tracker_for_testcase(testcase)
-  if not issue_tracker or not testcase.bug_information:
+  if not issue_tracker:
     return None
 
-  return issue_tracker.issue_url(testcase.bug_information)
+  issue_id = (
+      testcase.bug_information
+      if testcase.bug_information else testcase.group_bug_information)
+  if not issue_id:
+    return None
+
+  # Use str(issue_id) as |group_bug_information| is an integer.
+  return issue_tracker.issue_url(str(issue_id))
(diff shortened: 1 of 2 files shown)


Dor1s:
Wondering whether `get_issue_for_testcase` also should check `group_bug_information`. If yes, I'll create a helper for getting issue id and use it in both places.

Dor1s:
some network glitch, trying again

Dor1s:
> Wondering whether get_issue_for_testcase also should check group_bug_information. If yes, I'll create a helper for getting issue id and use it in both places.

After taking a closer look at the call sites for `get_issue_for_testcase`, it seems to be that it shouldn't hurt to update it as well. I'll upload another patchset.

inferno-chromium:
@oliverchang should review this. i think i did a wrong review here, since get_issue_for_url seems like the wrong place. Either get_issue_for_testcase as you mentioned or somewhere else. but i just worry that group bug information might cause wrong result elsewhere. since @oliverchang refactored this, he will know this inside out.

Dor1s:
Ack, waiting for Oliver to check what's going on first.

Dor1s:
Sorry I couldn't resist and made that change, as it actually made the whole CL smaller and likely better.

Dor1s:
Thanks for clarifying! I've reverted that change and added a comment to `get_issue_for_testcase`.

src/appengine/libs/issue_management/issue_tracker_utils.py
```diff
@@ -154,10 +156,17 @@ def get_similar_issues_url(issue_tracker, testcase, only_open=True):
 def get_issue_url(testcase):
   """Return issue url for a testcase."""
```
inferno-chromium: Maybe expand on description here that this is for display and account for group bug.

Dor1s: Done!

PR #630, opened 2019-07-02
[Reproduce tool] Bug fixes and TODO cleanup.

mbarbella-chromium:
PTAL. Just some minor cleanup.

### DIFF
--- a/src/local/butler/reproduce.py
+++ b/src/local/butler/reproduce.py
@@ -61,6 +61,11 @@ _GET_METHOD = 'GET'
 _POST_METHOD = 'POST'
 
 
+class ReproduceToolException(Exception):
+  """Base class for reproduce tool exceptions."""
+  pass
+
+
 class SerializedTestcase(object):
   """Minimal representation of a test case."""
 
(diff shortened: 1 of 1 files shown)


src/local/butler/reproduce.py
```diff
@@ -272,7 +279,15 @@ def _reproduce_crash(testcase_id, build_directory):
...
   """Attempt to reproduce a crash then report on the result."""
-  result = _reproduce_crash(args.testcase, args.build_dir)
+  try:
+    result = _reproduce_crash(args.testcase, args.build_dir)
+  except ReproduceToolException as exception:
+    print(exception)
+    return
+
+  if result.is_crash():
+    print('Test case reproduced successfully:\n')
```
inferno-chromium: Remove :\n from this and Output:\n from next

just do a common
print('Output:\n' + result.get_stacktrace())

src/local/butler/reproduce.py
```diff
@@ -272,7 +279,15 @@ def _reproduce_crash(testcase_id, build_directory):
 
 def execute(args):
   """Attempt to reproduce a crash then report on the result."""
-  result = _reproduce_crash(args.testcase, args.build_dir)
+  try:
+    result = _reproduce_crash(args.testcase, args.build_dir)
+  except ReproduceToolException as exception:
+    print(exception)
```
inferno-chromium: Why not just raise ?

mbarbella-chromium: It just seems like cleaner output. The cases this catches are supposed to be unrecoverable errors that we've handled explicitly.

src/local/butler/reproduce.py

PR #634, opened 2019-07-02
Use current project instead of default when choosing name for /src.

inferno-chromium:


inferno-chromium:
I think this has always been wrong, i noticed this in android since we were showing Google: instead of Android:.

inferno-chromium:
> hmm, how does this work on App Engine, where PROJECT_NAME from the job won't be set?

Fixed, good catch.

Issue #635, opened 2019-07-03
Permissions not working for /testcases?


jonathanmetzman:
One user reported to me that they couldn't access the list of testcases for vtest_fuzzer even though I gave them permission on the configuration page. Here is a screenshot of what they saw instead:
![image](https://user-images.githubusercontent.com/31354670/60556244-39135680-9cf5-11e9-9662-4c6834c5f597.png) When I go to that page I see a list of testcases.

Note that this screenshot shows an error on my part, I accidentally granted permission for vtest_fuzzer instead of libFuzzer_chromeos_vtest_fuzzer. I corrected this mistake, but the user still says they can't access any of the testcases.

inferno-chromium:
Verified with a gmail account, this happen when user is trying to access security bugs added on config page. Two bugs
- Testcases list is not working
- Testcase detail page is not working for overridden fuzzers (e.g. libfuzzer, afl)

inferno-chromium:
Should be fixed.

inferno-chromium:
And now the fuzzer filter is broken since it needs lower names.

inferno-chromium:
Phew fixed!

jonathanmetzman:
Thanks everyone!

PR #638, opened 2019-07-03
Don't lowercase string in String filters.

inferno-chromium:
This fixes yet another bug from
https://github.com/google/clusterfuzz/issues/635
We shouldn't lowercase string in String filters as otherwise
queries won't work. This didnt get noticed uptil now since
we used to store fuzzer_name_indices in lowercase (before https://github.com/google/clusterfuzz/commit/7e1d45708959d7748af782fa1fc31d8880a1744f) and rest
of fields like job, project, etc were always in lowercase.
This functionality was always broken. This fixes the filters,
whereas search functionality with keywords will still allow
the case insensitive search.

After this, we will have:
1. For exact match, use {whatever field name}:{case sensitive value}, e.g. fuzzer:AbC
2. for case insensitive, just type a string in search field, e.g. abc

### DIFF
--- a/src/appengine/libs/filters.py
+++ b/src/appengine/libs/filters.py
@@ -66,7 +66,7 @@ def get_boolean(value):
 
 def get_string(value):
   """Get sanitized string."""
-  return value.strip().lower()
+  return value.strip()
 
 
 class Filter(object):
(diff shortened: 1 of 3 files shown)


oliverchang:
LGTM, thanks!

PR #563, opened 2019-06-14
[fuchsia] Output Fuchsia OS crash data to dashboard.

flowerhack:
Initially I promised this would include both crash data & a symbolized
stacktrace; however, some hurdles wrt symbolized output on the Fuchsia
end of things has made me think it may be cleaner/faster to go ahead and
get this "hello world" out there.

When Clusterfuzz finds a crash, a testcase will appear on the
Clusterfuzz dashboard, e.g. like so:

```	[Environment] ASAN_OPTIONS="redzone=32:print_summary=1:handle_sigill=1:allocator_release_to_os_interval_ms=500:print_suppressions=0:strict_memcmp=1:allow_user_segv_handler=0:allocator_may_return_null=1:use_sigaltstack=1:handle_sigfpe=1:handle_sigbus=1:detect_stack_use_after_return=1:alloc_dealloc_mismatch=0:detect_leaks=1:print_scariness=1:max_uar_stack_size_log=16:handle_abort=1:check_malloc_usable_size=0:quarantine_size_mb=64:detect_odr_violation=0:symbolize=0:handle_segv=1:fast_unwind_on_fatal=0"
[Command line] python /usr/local/google/home/flowerhack/june13-2/clusterfuzz/src/python/bot/fuzzers/libFuzzer/launcher.py libFuzzer dummy_fuzzer_name
+----------------------------------------Release Build Stacktrace----------------------------------------+
Warning: Permanently added '[localhost.corp.google.com]:41105' (ED25519) to the list of known hosts.
INFO: Seed: 3708787003
INFO: Loaded 1 modules   (9 inline 8-bit counters): 9 [0x482cdb589060, 0x482cdb589069),
INFO: Loaded 1 PC tables (9 PCs): 9 [0x482cdb589070,0x482cdb589100),
INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 4096 bytes
INFO: A corpus is not provided, starting from an empty corpus
==33853== ERROR: libFuzzer: deadly signal
NOTE: libFuzzer has rudimentary signal handlers.
      Combine libFuzzer with AddressSanitizer or similar for better crash reports.
SUMMARY: libFuzzer: deadly signal
MS: 1 ChangeByte-; base unit: 1e66b44810d66b9f096d6a5b1344d65f45401873
0x48,0x49,0x21,
HI!
artifact_prefix='data'; Test unit written to datacrash-7a8dc3985d2a90fb6e62e94910fc11d31949c348
Base64: SEkh```

[...]

PR #644, opened 2019-07-03
[fuchsia] Temporarily fix crash detection.

flowerhack:
Pull request #563 was meant to put crash data on the dashboard; however,
some changes I made during code review changed how the "is_crash"
detection logic work, making no crashes appear on the dashboard.

Given that we fully intend to *properly* implement crash detection as
soon as we fix "getting symbolized crashes out of Fuchsia", for now,
let's just assume all Fuchsia crashes are real crashes, so we can start
stress-testing our dashboard.


flowerhack:
@oliverchang

oliverchang:
Discussed this on chat with @flowerhack. The plan is to fix this properly instead of relying on this hack.

Issue #616, opened 2019-06-28
Remove dead and duplicate functions from monorail tracker


inferno-chromium:
https://github.com/google/clusterfuzz/blob/master/src/appengine/libs/issue_management/monorail/issue_tracker_manager.py#L305

e.g. get_original_issue.

inferno-chromium:
Rest of stuff is helper functions that might be useful in future. So, only removing this duplicate function.

PR #648, opened 2019-07-05
Remove unused duplicate function in monorail (#616).

inferno-chromium:


inferno-chromium:
Rest of stuff is helper functions that might be useful in future. So, only removing this custom function.

inferno-chromium:
TBR minor dup function remove.

oliverchang:
LGTM

Issue #483, opened 2019-05-23
Make fuzzing engines easily pluggable (for easy eval)


inferno-chromium:
Also, remove minijail hacks (as part of refactoring work).

oliverchang:
This is done, just need to migrate AFL and add honggfuzz (will create separate issues for that).

PR #654, opened 2019-07-08
Start refactoring fuzz_task (#483).

oliverchang:
Should have no functional change.

Create a FuzzingSession class, and factor out blackbox fuzzing into its own method "do_blackbox_fuzzing". 

Note that there are still many improvements to be done. This CL attempts to keep the diffs small
(although github diffs don't do a great job of recognising what changed).

### DIFF
--- a/src/python/bot/fuzzers/builtin.py
+++ b/src/python/bot/fuzzers/builtin.py
@@ -25,6 +25,7 @@ from bot.fuzzers import utils as fuzzers_utils
 from datastore import data_types
 from fuzzing import testcase_manager
 from system import environment
+from system import shell
 
 
 class BuiltinFuzzerResult(object):
@@ -65,9 +66,7 @@ def get_corpus_directory(input_directory, project_qualified_name):
     from bot.untrusted_runner import file_host
     file_host.create_directory(corpus_directory, create_intermediates=True)
   else:
-    if not os.path.exists(corpus_directory):
-      os.mkdir(corpus_directory)
-
+    shell.create_directory(corpus_directory)
   return corpus_directory
 
 
--- a/src/python/bot/fuzzers/builtin.py
+++ b/src/python/bot/fuzzers/builtin.py
@@ -53,6 +53,24 @@ class BuiltinFuzzer(object):
         os.path.dirname(sys.modules[self.__module__].__file__))
 
 
+def get_corpus_directory(input_directory, project_qualified_name):
+  """Get the corpus directory given a project qualified fuzz target name."""
+  corpus_directory = os.path.join(input_directory, project_qualified_name)
+  if environment.is_trusted_host():
+    from bot.untrusted_runner import file_host
+    corpus_directory = file_host.rebase_to_worker_root(corpus_directory)
+
+  # Create corpus directory if it does not exist already.
+  if environment.is_trusted_host():
+    from bot.untrusted_runner import file_host
+    file_host.create_directory(corpus_directory, create_intermediates=True)
+  else:
+    if not os.path.exists(corpus_directory):
+      os.mkdir(corpus_directory)
+
+  return corpus_directory
+
+
 class EngineFuzzer(BuiltinFuzzer):
   """Builtin fuzzer for fuzzing engines such as libFuzzer."""
 
@@ -94,20 +112,9 @@ class EngineFuzzer(BuiltinFuzzer):
     project_qualified_name = data_types.fuzz_target_project_qualified_name(
         utils.current_project(), fuzzer_binary_name)
 
-    corpus_directory = os.path.join(input_directory, project_qualified_name)
-    if environment.is_trusted_host():
-      from bot.untrusted_runner import file_host
-      corpus_directory = file_host.rebase_to_worker_root(corpus_directory)
-
     arguments = self.generate_arguments(fuzzer_path)
-
-    # Create corpus directory if it does not exist already.
-    if environment.is_trusted_host():
-      from bot.untrusted_runner import file_host
-      file_host.create_directory(corpus_directory, create_intermediates=True)
-    else:
-      if not os.path.exists(corpus_directory):
-        os.mkdir(corpus_directory)
+    corpus_directory = get_corpus_directory(input_directory,
+                                            project_qualified_name)
 
     # Create fuzz testcases.
     for i in range(no_of_files):
--- a/src/python/bot/fuzzers/engine.py
+++ b/src/python/bot/fuzzers/engine.py
@@ -15,6 +15,8 @@
 
 from builtins import object
 
+_ENGINES = {}
+
 
 class FuzzOptions(object):
   """Represents options passed to the engine. Can be overridden to provide more
@@ -47,6 +49,10 @@ class Result(object):
 class Engine(object):
   """Base interface for a grey box fuzzing engine."""
 
+  @property
+  def name(self):
+    return self.__class__.__name__
+
   def prepare(self, corpus_dir, target_path, build_dir):
     """Prepare for a fuzzing session, by generating options. Returns a
     FuzzOptions object."""
@@ -72,3 +78,20 @@ class Engine(object):
     """Optional: scrub a testcase of potentially sensitive bytes. Returns a
     bool."""
     raise NotImplementedError
+
+
+def register_engine(name, impl):
+  """Register a fuzzing engine."""
+  if impl in _ENGINES:
+    raise ValueError('Engine {name} is already registered'.format(name=name))
+
+  _ENGINES[name] = impl
+
+
+def get(name):
+  """Get an implemntation of a fuzzing engine, or None if one does not exist."""
+  engine_class = _ENGINES.get(name)
+  if engine_class:
+    return engine_class()
+
+  return None
--- a/src/python/bot/fuzzers/engine.py
+++ b/src/python/bot/fuzzers/engine.py
@@ -51,7 +51,8 @@ class Engine(object):
 
   @property
   def name(self):
-    return self.__class__.__name__
+    """Get the name of the engine."""
+    raise NotImplementedError
 
   def prepare(self, corpus_dir, target_path, build_dir):
     """Prepare for a fuzzing session, by generating options. Returns a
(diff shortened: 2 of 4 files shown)


oliverchang:
PTAL. currently staging this on a number of bots to ensure nothing is broken.

src/python/bot/fuzzers/builtin.py
```diff
@@ -53,6 +53,24 @@ def fuzzer_directory(self):
...
+  if environment.is_trusted_host():
+    from bot.untrusted_runner import file_host
+    corpus_directory = file_host.rebase_to_worker_root(corpus_directory)
+
+  # Create corpus directory if it does not exist already.
+  if environment.is_trusted_host():
+    from bot.untrusted_runner import file_host
+    file_host.create_directory(corpus_directory, create_intermediates=True)
+  else:
+    if not os.path.exists(corpus_directory):
```
inferno-chromium: nit: shell.create_directory

src/python/bot/tasks/fuzz_task.py
```diff
@@ -726,6 +728,8 @@ def run_fuzzer(fuzzer, fuzzer_directory, testcase_directory, data_directory,
   testcase_manager.remove_testcases_from_directories(testcase_directories)
 
   # Set an environment variable for fuzzer name.
+  # TODO(ochang): Investigate removing this. Only users appear to be chromebot
```
inferno-chromium: chromebot is no longer running, so it can be removed. only webbot. but yes fuzzer_logs is an active user.

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1254,319 +1258,391 @@ def get_strategy_distribution_from_ndb():
...
-  # Pull testcase directory to host. The testcase file contents could have been
-  # changed (by e.g. libFuzzer) and stats files could have been written.
-  if environment.is_trusted_host():
-    from bot.untrusted_runner import file_host
-    file_host.pull_testcases_from_worker()
+    engine_impl = engine.get(fuzzer.name)
+    if engine_impl:
+      crashes = self.do_engine_fuzzing(engine_impl)
+      # TODO(ochang): fill these.
+      testcase_file_paths = []
```
inferno-chromium: Dont understand this, how can we proceed without these ? So this "if engine_impl" branch is dead currently, probably a FIXME or TODO to indicate that.

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1254,319 +1258,391 @@ def get_strategy_distribution_from_ndb():
...
+          self.do_blackbox_fuzzing(fuzzer, fuzzer_directory, self.job_type))
 
-  # Synchronize corpus files with GCS after fuzzing
-  if sync_corpus_directory:
-    new_files = gcs_corpus.get_new_files()
-    new_files_count = len(new_files)
-    logs.log('%d new corpus files generated by fuzzer %s (job %s).' %
-             (new_files_count, fuzz_target.project_qualified_name(), job_type))
+    if crashes is None:
+      # Error occurred.
```
inferno-chromium: log_error  or no crashes occurred.

src/python/bot/fuzzers/engine.py
```diff
@@ -72,3 +78,20 @@ def cleanse(self, target_path, input_path, output_path, max_time):
     """Optional: scrub a testcase of potentially sensitive bytes. Returns a
     bool."""
     raise NotImplementedError
+
+
+def register_engine(name, impl):
+  """Register a fuzzing engine."""
+  if impl in _ENGINES:
```
jonathanmetzman: Shoul this be if name in `_ENGINES`?

src/python/bot/fuzzers/engine.py
```diff
@@ -72,3 +78,20 @@ def cleanse(self, target_path, input_path, output_path, max_time):
     """Optional: scrub a testcase of potentially sensitive bytes. Returns a
     bool."""
     raise NotImplementedError
+
+
+def register_engine(name, impl):
```
jonathanmetzman: Can we use `impl.name` instead of having name as a param?

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1254,319 +1258,391 @@ def get_strategy_distribution_from_ndb():
...
+     sync_corpus_directory, fuzzer_metadata) = run_fuzzer(
+         fuzzer, fuzzer_directory, self.testcase_directory, self.data_directory,
+         testcase_count)
+
+    if error_occurred:
+      return None, None, None
+
+    fuzzer_binary_name = fuzzer_metadata.get('fuzzer_binary_name')
+    if fuzzer_binary_name:
+      self.fuzz_target = record_fuzz_target(fuzzer.name, fuzzer_binary_name,
```
inferno-chromium: Why is this branch in blackbox fuzzing  and line 1371 ?

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1254,319 +1258,391 @@ def get_strategy_distribution_from_ndb():
...
+    max_threads = utils.maximum_parallel_processes_allowed()
+    needs_stale_process_cleanup = False
+    test_number = 0
+    testcases_before_stale_process_cleanup = environment.get_value(
+        'TESTCASES_BEFORE_STALE_PROCESS_CLEANUP', 1)
+    thread_delay = environment.get_value('THREAD_DELAY')
+    thread_error_occurred = False
+
+    # TODO: Remove environment variable once fuzzing engine refactor is
+    # complete.  Set multi-armed bandit strategy selection distribution as an
```
jonathanmetzman: nit: I think  we generally leave one space after a period.

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1254,319 +1258,391 @@ def get_strategy_distribution_from_ndb():
...
+    # environment variable so we can access it in launcher.
+    if environment.get_value('USE_BANDIT_STRATEGY_SELECTION'):
```
jonathanmetzman: Why are we doing strategy selection with blackbox fuzzers? Doesn't strategy selection only work for LF?
Is it because we are doing LF fuzzing in this function for now? If that's the case I think comments should make it clearer.

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1254,319 +1258,391 @@ def get_strategy_distribution_from_ndb():
...
+
+    # For timeout multipler greater than 1, we need to decrease testcase count
+    # to prevent exceeding task lease time.
+    if self.timeout_multiplier > 1:
+      testcase_count /= self.timeout_multiplier
+
+    # Run the fuzzer to generate testcases. If error occurred while trying
+    # to run the fuzzer, bail out.
+    (error_occurred, testcase_file_paths,
+     sync_corpus_directory, fuzzer_metadata) = run_fuzzer(
```
jonathanmetzman: I wonder if run_fuzzer would be better as a method.

src/python/bot/fuzzers/engine.py
```diff
@@ -47,6 +49,10 @@ def __init__(self, crashes, stats):
 class Engine(object):
   """Base interface for a grey box fuzzing engine."""
 
+  @property
+  def name(self):
+    return self.__class__.__name__
```
jonathanmetzman: This will need to be overridden for AFL which we call "afl" everywhere but in class names.
I also think we may want to do the same for libFuzzer which we generally call "libFuzzer" and not "LibFuzzer". In general I don't like the idea of python class names (which obey certain conventions) determining the name we use elsewhere.
I'm not sure this property is a good idea due to this problem, it seems like it can introduce subtle problems.

src/python/bot/fuzzers/builtin.py
oliverchang: Done.

src/python/bot/tasks/fuzz_task.py
oliverchang: Ack. will be removing this in a future CL.

src/python/bot/tasks/fuzz_task.py
oliverchang: This is just a placeholder for future implementation. Added a TODO.

src/python/bot/fuzzers/engine.py
oliverchang: Fixed!

src/python/bot/fuzzers/engine.py
oliverchang: Done.

src/python/bot/tasks/fuzz_task.py

src/python/bot/tasks/fuzz_task.py
oliverchang: There is no functional change in this CL, it just sets up the pipes for future refactoring and implementation.

src/python/bot/tasks/fuzz_task.py
oliverchang: Ack. There are many things which can be moved, will be done in future CLs to avoid making the diffs too large.

src/python/bot/fuzzers/engine.py
oliverchang: Replaced this with `raise NotImplementedError` so that we explicitly specify the name.

src/python/bot/tasks/fuzz_task.py
oliverchang: Line 1371 is in do_blackbox_fuzzing() :) Were you referring to something else?

If you're referring to the FUZZER_NAME setting duplication, that was always there. We set FUZZER_NAME before run_fuzzer, and set it again if we have a fuzzer_name override.

src/python/bot/tasks/fuzz_task.py
oliverchang: This can only happen when run_fuzzer failed, and we already log_error in some of those cases. 

This also matches previous behaviour, but I think we can pipe the error a bit better here, since it's not obvious what failed here looking at the code. Added a TODO.

src/python/bot/tasks/fuzz_task.py
jonathanmetzman: Right, but the name of the function do_blackbox_fuzzing makes some of the body of this function confusing for the time being.

oliverchang: I added a comment to the start of do_blackbox_fuzzing to hopefully clear things up.

PR #655, opened 2019-07-08
[DONT MERGE] [AFL][LPM] Don't pick LPM targets in AFL jobs

jonathanmetzman:


jonathanmetzman:
This solution is a bit ugly because I'm trying to avoid searching through the file unless necessary, while at the same time not picking any LPM targets for AFL jobs. I'm trying to think of other solutions.

oliverchang:
Is this PR still relevant?

jonathanmetzman:
It's still relevant as OSS-Fuzz users are probably still uploading AFL with LPM fuzz targets. But I don't plan on landing this any time soon.

PR #659, opened 2019-07-08
Disable corpus subset when DFT is used, also extend strategy_pool API (#503).

Dor1s:
A bit speculative change, but feels intuitive that DFT is less useful when used with corpus subset.

Will be good to deploy it soon as it'll be midnight by UTC in ~1.5 hours and we can look at the daily stats to evaluate the impact of this change.

### DIFF
--- a/src/python/bot/fuzzers/libFuzzer/strategy_selection.py
+++ b/src/python/bot/fuzzers/libFuzzer/strategy_selection.py
@@ -40,11 +40,6 @@ class StrategyPool(object):
     """Add a strategy into our existing strategy pool."""
     self.strategy_names.add(strategy_tuple.name)
 
-  def remove_strategy(self, strategy_tuple):
-    """Remove a strategy from our existing strategy pool."""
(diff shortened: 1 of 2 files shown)


src/python/bot/fuzzers/libFuzzer/strategy_selection.py
```diff
@@ -40,6 +40,11 @@ def add_strategy(self, strategy_tuple):
     """Add a strategy into our existing strategy pool."""
     self.strategy_names.add(strategy_tuple.name)
 
+  def remove_strategy(self, strategy_tuple):
```
mbarbella-chromium: Discussed offline. This could end up getting a bit brittle depending on the order of how things are modified, and we don't necessarily expect everything enabled in the pool to be added to fuzzing_strategies if it causes issues.

Dor1s: Done!

PR #661, opened 2019-07-08
Strategy selection traffic split

mukundv-chrome:
When 'USE_BANDIT_STRATEGY_SELECTION' variable is enabled, splits fuzzing traffic 80% default strategy selection method, 20% multi armed bandit strategy selection method.

Default strategy selection method used otherwise.


src/python/bot/tasks/fuzz_task.py
```diff
@@ -1400,9 +1401,12 @@ def execute_task(fuzzer_name, job_type):
   # TODO: Remove environment variable once fuzzing engine refactor is complete.
   # Set multi-armed bandit strategy selection distribution as an environment
   # variable so we can access it in launcher.
+  environment.set_value('STRATEGY_SELECTION_METHOD', 'DEFAULT')
```
mbarbella-chromium: Minor nit: lowercase names for the methods themselves.

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1400,9 +1401,12 @@ def execute_task(fuzzer_name, job_type):
...
   if environment.get_value('USE_BANDIT_STRATEGY_SELECTION'):
-    distribution = get_strategy_distribution_from_ndb()
-    environment.set_value('STRATEGY_SELECTION_DISTRIBUTION', distribution)
+    if engine_common.decide_with_probability(.2):
```
mbarbella-chromium: Let's define this as a constant in this file.

src/python/bot/fuzzers/libFuzzer/stats.py
```diff
@@ -183,6 +184,10 @@ def parse_performance_features(log_lines, strategies, arguments):
       'timeout_count': 0,
   }
 
+  # Extract strategy selection method.
+  stats['strategy_selection_method'] = environment.get_value(
```
mbarbella-chromium: Could you use a default value of 'default' here rather than worrying about setting it in the default case? It feels a bit safer to me than what we're currently doing.

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1400,9 +1402,12 @@ def execute_task(fuzzer_name, job_type):
   # TODO: Remove environment variable once fuzzing engine refactor is complete.
   # Set multi-armed bandit strategy selection distribution as an environment
   # variable so we can access it in launcher.
+  environment.set_value('STRATEGY_SELECTION_METHOD', 'default')
   if environment.get_value('USE_BANDIT_STRATEGY_SELECTION'):
-    distribution = get_strategy_distribution_from_ndb()
-    environment.set_value('STRATEGY_SELECTION_DISTRIBUTION', distribution)
+    if engine_common.decide_with_probability(MULTI_ARMED_BANDIT_TRAFFIC_SPLIT):
```
mbarbella-chromium: Late nit: use "and" here rather than doubly-nesting this.

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1400,9 +1402,12 @@ def execute_task(fuzzer_name, job_type):
   # TODO: Remove environment variable once fuzzing engine refactor is complete.
   # Set multi-armed bandit strategy selection distribution as an environment
   # variable so we can access it in launcher.
+  environment.set_value('STRATEGY_SELECTION_METHOD', 'default')
```
mbarbella-chromium: This can be deleted if the previous comment is addressed.

PR #666, opened 2019-07-09
Fix an exception with 0-size test cases in the viewer handler.

mbarbella-chromium:
PTAL. We could do a more generic fix in storage.read_data but it would require us to check the size in all cases, which seems like it could hurt performance in cases where we don't expect issues. Since we already know the size in the handler this seems like the best quick fix to me, but I'm open to either option.

### DIFF
--- a/src/appengine/handlers/viewer.py
+++ b/src/appengine/handlers/viewer.py
@@ -45,13 +45,20 @@ class Handler(base_handler.Handler):
       if not access.has_access():
         raise helpers.AccessDeniedException()
 
-    if blobs.get_blob_size(key) > MAX_ALLOWED_CONTENT_SIZE:
+    blob_size = blobs.get_blob_size(key)
+    if blob_size > MAX_ALLOWED_CONTENT_SIZE:
       raise helpers.EarlyExitException('Content exceeds max allowed size.', 400)
 
-    try:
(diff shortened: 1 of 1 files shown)


src/appengine/handlers/viewer.py
```diff
@@ -45,13 +45,20 @@ def get(self):
...
-    try:
-      content = unicode(blobs.read_key(key), errors='replace')
-    except Exception:
-      raise helpers.EarlyExitException('Failed to read content.', 400)
+    # TODO(mbarbella): Workaround for an issue in the Cloud Storage API. Remove
+    # once it is fixed properly upstream:
+    # https://github.com/googleapis/google-cloud-python/issues/6572
+    if blob_size:
+      try:
+        content = unicode(blobs.read_key(key), errors='replace')
```
inferno-chromium: maybe remove this unicode thing along the way and fix https://bugs.chromium.org/p/chromium/issues/detail?id=920160 :) stage to make sure it works.

mbarbella-chromium: SGTM.

mbarbella-chromium: Seems to cause issues later in template rendering, which I guess explains why we had the hack initially. I'm going to leave it out of this change.

PR #668, opened 2019-07-10
Fix an exception when run_fuzzer fails.

oliverchang:


oliverchang:
tbr for a trivial fix.

PR #680, opened 2019-07-12
Pass schema as part of bigquery load call

inferno-chromium:
This allows for changing field types, which the original schema update call does not and fails with e.g. "Provided Schema does not match Table cluster-fuzz:afl_stats.TestcaseRun. Field seconds_executed has changed type from FLOAT to INTEGER". Also, allows removing of old unused types.

### DIFF
--- a/src/appengine/handlers/cron/load_bigquery_stats.py
+++ b/src/appengine/handlers/cron/load_bigquery_stats.py
@@ -96,27 +96,6 @@ class Handler(base_handler.Handler):
         projectId=project_id, datasetId=dataset_id, body=table_body)
     return self._execute_insert_request(table_insert)
 
-  def _update_schema_if_needed(self, bigquery, dataset_id, table_id, schema):
-    """Update the table's schema if needed."""
-    if not schema:
-      return
-
-    project_id = utils.get_application_id()
-    table = bigquery.tables().get(
-        datasetId=dataset_id, tableId=table_id, projectId=project_id).execute()
-
-    if 'schema' in table and table['schema'] == schema:
-      return
-
-    body = {
-        'schema': schema,
-    }
-
-    logs.log('Updating schema for %s:%s' % (dataset_id, table_id))
-    bigquery.tables().patch(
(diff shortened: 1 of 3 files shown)


src/appengine/handlers/cron/load_bigquery_stats.py
```diff
@@ -140,21 +119,24 @@ def _load_data(self, bigquery, fuzzer):
...
-      self._update_schema_if_needed(bigquery, dataset_id, table_id, schema)
-
       gcs_path = fuzzer_stats.get_gcs_stats_path(kind_name, fuzzer, timestamp)
+      load = {
+          'destinationTable': {
+              'projectId': project_id,
+              'tableId': table_id + '$' + date_string,
+              'datasetId': dataset_id,
+          },
+          'schemaUpdateOptions': ['ALLOW_FIELD_ADDITION',],
```
inferno-chromium: Note that this is intentionally added, otherwise, new fields don't get accepted.

src/python/metrics/fuzzer_stats_schema.py
```diff
@@ -161,10 +161,6 @@
     'mode': 'NULLABLE',
     'name': 'leak_count',
     'type': 'INTEGER'
-}, {
```
inferno-chromium: This makes it awesome to delete old unused attributes.

src/appengine/handlers/cron/load_bigquery_stats.py
```diff
@@ -140,21 +119,24 @@ def _load_data(self, bigquery, fuzzer):
...
+          'sourceFormat': 'NEWLINE_DELIMITED_JSON',
+          'sourceUris': ['gs:/' + gcs_path + '*.json'],
+          'writeDisposition': 'WRITE_TRUNCATE',
```
inferno-chromium: WRITE_TRUNCATE is the one which allows passed schema to overwrite field type over what is currently in table.

oliverchang:
LGTM, thanks for fixing this! I should've tested this with internal as well.

Issue #565, opened 2019-06-16
Reproduce found crashes on other job types / platforms / sanitizers


inferno-chromium:
See also https://github.com/google/oss-fuzz/issues/775

inferno-chromium:
See also cross job deduplication and other improvements - https://bugs.chromium.org/p/chromium/issues/detail?id=787078

inferno-chromium:
See also
https://bugs.chromium.org/p/chromium/issues/detail?id=950353

jonathanmetzman:
A possible example of a bug that ASAN reports incorrectly: crbug.com/976429

Dor1s:
Another potential benefit here is that for crashes found e.g. on Android, we might prefer reproducing and reporting on Linux, if possible.

inferno-chromium:
Started - https://github.com/google/clusterfuzz/commit/08cbad1c5b655356003327c166098124bd982c27

inferno-chromium:
Will be evaluating results first, fixing exceptions. And then do a CL to show in UI. Also, need to exclude cases to avoid cpu wastage.

inferno-chromium:
Exception fix list
- https://github.com/google/clusterfuzz/commit/0ad2cb40bafe3a849dd989f3cddd68dc2d6bef64
- https://github.com/google/clusterfuzz/commit/00224d652c03b686cdf2fa7524d30089a4322b4b
- https://github.com/google/clusterfuzz/commit/db17ff45b081b398c09bd7ee12f978c67586eed8
- https://github.com/google/clusterfuzz/commit/db4f12d0a62bdead8960438f86d70655f1c84e82

inferno-chromium:
Should be done. Many more fixes went in, forgot to associate them all to this.

PR #681, opened 2019-07-15
Reproduce crashes in other jobs in a project, store result.

inferno-chromium:


src/python/bot/tasks/task_creation.py
```diff
@@ -142,13 +143,23 @@ def create_regression_task_if_needed(testcase):
   tasks.add_task('regression', testcase.key.id(), testcase.job_type)
 
 
-def create_stack_task_if_needed(testcase):
-  """Creates a stack task if needed."""
-  second_stack_job_type = environment.get_value('SECOND_STACK_JOB_TYPE')
-  if not second_stack_job_type:
-    return
+def create_variant_tasks_if_needed(testcase):
```
mbarbella-chromium: Should we do any additional filtering on something like APP_NAME?

inferno-chromium: Yes i have to think more and add things. APP_NAME becomes different even for Chrome, it is chrome.exe for win, chrome for linux, and some Chromium.app for mac.

src/python/bot/tasks/task_creation.py
```diff
@@ -142,13 +142,23 @@ def create_regression_task_if_needed(testcase):
...
+  """Creates a variant task if needed."""
+  testcase_id = testcase.key.id()
+  jobs = ndb_utils.get_all_from_model(data_types.Job)
```
Dor1s: nit: I'm thinking if we can do some extra filtering at this step, like `name != testcase.job_type`, but that feels insufficient. If we could check the project, that might be a good optimization I think (as OSS-Fuzz has a ton of jobs / projects), but we don't have project name in the Job entity as is.

Dor1s:
Mentioning issue #565 here to get a nice reference link in the issue :)

src/python/bot/tasks/task_creation.py
inferno-chromium: every oss-fuzz job has project name, e.g. "PROJECT_NAME = bignum-fuzzer"
also, i filter by project name inside the loop.

Dor1s: Sorry, my point was that with current approach we're querying all the jobs from the datastore and then we're skipping like 99.5% of them. If we could avoid doing that and do the filtering inside the datastore query, it might be more efficient. I don't know whether it's a big gain or not, that's why it's just a nit :)

inferno-chromium: It cannot be done in query, since the things we are looking for are in the job environment string. like PROJECT_NAME, EXPERIMENTAL, etc.

PR #684, opened 2019-07-16
For variant testcase, create queue based on job platform.

inferno-chromium:
This should fix exceptions - 
"Wrong platform for job linux_ubsan_vptr_chrome: job queue [-linux], bot queue [-mac]."


inferno-chromium:
TBR exception fix.

PR #685, opened 2019-07-16
Fix exception when failed to publish to pubsub, log error.

inferno-chromium:
Fix for exception:
https://pantheon.corp.google.com/errors/CPi7-pPYxpiDSQ?time=PT1H&refresh=auto&project=google.com:clusterfuzz&organizationId=433637338589


inferno-chromium:
Maybe the wrong fix, should i do like pull_from_subscription, raise a runtime error. but i want it to recover.

  def pull_from_subscription(self,
                             subscription,
                             max_messages=_DEFAULT_MAX_MESSAGES,
                             acknowledge=False):
....
    if response is None:
      raise RuntimeError('Invalid subscription: ' + subscription)

PR #686, opened 2019-07-16
More fuzz_task refactoring (#483).

oliverchang:
- Move run_fuzzer into FuzzingSession.generate_blackbox_testcases.
- Sketch out do_engine_fuzzing.
- Add crash_time to engine.Crash, and add constructors for
  fuzz_task.Crash from testcase_manager.Crash and engine.Crash.

### DIFF
--- a/src/python/bot/tasks/fuzz_task.py
+++ b/src/python/bot/tasks/fuzz_task.py
@@ -97,7 +97,7 @@ Redzone = collections.namedtuple('Redzone', ['size', 'weight'])
 def get_unsymbolized_crash_stacktrace(stack_file_path):
   """Read unsymbolized crash stacktrace."""
   with open(stack_file_path, 'rb') as f:
-    return (utils.decode_to_unicode(f.read()), None)
+    return utils.decode_to_unicode(f.read())
 
 
 class Crash(object):
@@ -1207,6 +1207,7 @@ class FuzzingSession(object):
       fuzzer_output = builtin_result.output
       sync_corpus_directory = builtin_result.corpus_directory
 
+      # Return code is always 0 as builtin fuzzers log errors directly. 
       fuzzer_return_code = 0
     else:
       # Make sure we have a file to execute for the fuzzer.
--- a/src/python/bot/tasks/fuzz_task.py
+++ b/src/python/bot/tasks/fuzz_task.py
@@ -114,15 +114,24 @@ class Crash(object):
           'Unable to read stacktrace from file %s.' % crash.stack_file_path)
       return None
 
-    return Crash(crash.file_path, crash.crash_time, crash.return_code,
-                 crash.resource_list, crash.gestures,
-                 orig_unsymbolized_crash_stacktrace)
+    return Crash(
+        file_path=crash.file_path,
+        crash_time=crash.crash_time,
+        return_code=crash.return_code,
+        resource_list=crash.resource_list,
+        gestures=crash.gestures,
+        unsymbolized_crash_stacktrace=orig_unsymbolized_crash_stacktrace)
 
   @classmethod
   def from_engine_crash(cls, crash):
     """Create a Crash from a engine.Crash."""
-    return Crash(crash.input_path, crash.crash_time, 1, [], [],
-                 crash.stacktrace)
+    return Crash(
+        file_path=crash.input_path,
+        crash_time=crash.crash_time,
+        return_code=1,
+        resource_list=[],
+        gestures=[],
+        unsymbolized_crash_stacktrace=crash.stacktrace)
 
   def __init__(self, file_path, crash_time, return_code, resource_list,
                gestures, unsymbolized_crash_stacktrace):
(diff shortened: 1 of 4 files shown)


oliverchang:
PTAL. staging this on a few bots before merging.

src/python/bot/tasks/fuzz_task.py
```diff
@@ -96,46 +95,52 @@ class FuzzErrorCode(object):
...
-    self.stack_file_path = crash.stack_file_path
+  @classmethod
+  def from_testcase_manager_crash(cls, crash):
+    """Create a Crash from a testcase_manager.Crash."""
+    try:
+      orig_unsymbolized_crash_stacktrace = (
+          get_unsymbolized_crash_stacktrace(crash.stack_file_path))
+    except Exception:
+      logs.log_error(
+          'Unable to read stacktrace from file %s.' % crash.stack_file_path)
```
jonathanmetzman: Does the actual exception get included in the log?

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1319,6 +1173,155 @@ def sync_new_corpus_files(self):
...
+      fuzzer_command = 'builtin'
+      builtin_fuzzer = builtin_fuzzers.get(fuzzer.name)
+
+      builtin_result = builtin_fuzzer.run(
+          self.data_directory, self.testcase_directory, testcase_count)
+
+      fuzzer_output = builtin_result.output
+      sync_corpus_directory = builtin_result.corpus_directory
+
+      fuzzer_return_code = 0
```
jonathanmetzman: Realize this code isn't new, but probably worth explaining why return code is always 0 for builtin fuzzers (because they can log exceptions directly).

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1601,9 +1613,9 @@ def run(self):
       # testcase is analyzed.
       android.device.initialize_device()
 
-    # Transform tests.Crash into fuzz_task.Crash.
+    # Transform crashes into fuzz_task.Crash.
     # And filter the crashes (e.g. removing errorneous crashes).
-    crashes = [Crash(crash) for crash in crashes]
+    crashes = filter(None, [crash_constructor(crash) for crash in crashes])
```
jonathanmetzman: I don't understand what's the purpose of `filter(None, ...)`. Could you please explain?
It's not trying to get rid of Nones is it? (I don't think it works for that purpose).
Is this a common idiom?

src/python/bot/tasks/fuzz_task.py
oliverchang: Yes, log_error should include the exception in the log.

src/python/bot/tasks/fuzz_task.py
oliverchang: It is indeed to get rid of Nones, and I believe this is a common idiom. Passing None as the first argument means we use the identity function.

src/python/bot/tasks/fuzz_task.py
oliverchang: Done!

src/python/bot/tasks/fuzz_task.py
jonathanmetzman: ACK

src/python/bot/tasks/fuzz_task.py
```diff
@@ -96,46 +95,52 @@ class FuzzErrorCode(object):
...
+      return None
+
+    return Crash(crash.file_path, crash.crash_time, crash.return_code,
+                 crash.resource_list, crash.gestures,
+                 orig_unsymbolized_crash_stacktrace)
+
+  @classmethod
+  def from_engine_crash(cls, crash):
+    """Create a Crash from a engine.Crash."""
+    return Crash(crash.input_path, crash.crash_time, 1, [], [],
```
inferno-chromium: can we use named arguments for clarity, like return_code=1

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1319,6 +1173,156 @@ def sync_new_corpus_files(self):
...
+    # Set minimum redzone size, do not detect leaks and zero out the
+    # quarantine size before running the fuzzer.
+    environment.reset_current_memory_tool_options(
+        redzone_size=16, leaks=False, quarantine_size_mb=0)
+
+    if fuzzer.builtin:
+      fuzzer_command = 'builtin'
+      builtin_fuzzer = builtin_fuzzers.get(fuzzer.name)
+
+      builtin_result = builtin_fuzzer.run(
```
inferno-chromium: Why is the engine/builtin code inside generate_blackbox_testcases, was this expected or no other choice.

oliverchang: This will be removed once the refactor is done and we've migrated to new engine fuzzing code.

src/python/bot/tasks/fuzz_task.py

PR #691, opened 2019-07-16
For testcase variant, pick the associated engine fuzzer for that job.

inferno-chromium:
When we run a libFuzzer testcase with an afl job, it needs to
setup afl libfuzzer.

### DIFF
--- a/src/python/bot/fuzzers/builtin_fuzzers.py
+++ b/src/python/bot/fuzzers/builtin_fuzzers.py
@@ -35,3 +35,12 @@ def get(fuzzer_name):
     return None
 
   return BUILTIN_FUZZERS[fuzzer_name]
+
+
+def get_fuzzer_for_job(job_name):
+  """Return a fuzzer override for engine jobs."""
+  for fuzzer_name in BUILTIN_FUZZERS:
+    if fuzzer_name.lower() in job_name.lower():
+      return fuzzer_name
+
+  return None
--- a/src/python/bot/tasks/variant_task.py
+++ b/src/python/bot/tasks/variant_task.py
@@ -14,6 +14,7 @@
 """Variant task for analyzing testcase variants with a different job."""
 
 from base import utils
+from bot.fuzzers import builtin_fuzzers
 from bot.tasks import setup
 from build_management import build_manager
 from datastore import data_handler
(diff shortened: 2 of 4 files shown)


src/python/bot/fuzzers/builtin_fuzzers.py
```diff
@@ -35,3 +35,12 @@ def get(fuzzer_name):
     return None
 
   return BUILTIN_FUZZERS[fuzzer_name]
+
+
+def get_fuzzer_for_job(job_name):
+  """Return a fuzzer override for engine jobs."""
+  for fuzzer_name in BUILTIN_FUZZERS:
+    if fuzzer_name.lower() in job_name.lower():
```
jonathanmetzman: This problem was present before but I don't like that the way to determine a job uses an engine is if the engine is in the name. What if libFuzzer ends up in a project name?

src/python/bot/tasks/variant_task.py
```diff
@@ -39,7 +40,9 @@ def execute_task(testcase_id, job_type):
   testcase = data_handler.get_testcase_by_id(testcase_id)
 
   # Setup testcase and its dependencies.
-  file_list, _, testcase_file_path = setup.setup_testcase(testcase)
+  fuzzer_override = builtin_fuzzers.get_fuzzer_for_job(job_type)
```
jonathanmetzman: I'm confused why this is called `fuzzer_override`. Shouldn't it just be `fuzzer`

src/python/bot/fuzzers/builtin_fuzzers.py
inferno-chromium: libFuzzer can't be in project name, a project can have multiple engines. we say to put engine name in job - https://google.github.io/clusterfuzz/setting-up-fuzzing/libfuzzer-and-afl/

src/python/bot/tasks/variant_task.py
inferno-chromium: Just tries to match fuzzer_override arg on next line
  fuzzer_override = builtin_fuzzers.get_fuzzer_for_job(job_type)
  file_list, _, testcase_file_path = setup.setup_testcase(
      testcase, fuzzer_override=fuzzer_override)

also fuzzer_override makes it clear we are trying to override fuzzer_name based on job_type

Issue #690, opened 2019-07-16
Fix proto exception in triage cron.


inferno-chromium:
```
| ERROR    2019-07-16 18:42:03,137 wsgi.py:263]
| Traceback (most recent call last):
|   File "/build/google-cloud-sdk/platform/google_appengine/google/appengine/runtime/wsgi.py", line 240, in Handle
|     handler = _config_handle.add_wsgi_middleware(self._LoadHandler())
|   File "/build/google-cloud-sdk/platform/google_appengine/google/appengine/runtime/wsgi.py", line 299, in _LoadHandler
|     handler, path, err = LoadObject(self._handler)
|   File "/build/google-cloud-sdk/platform/google_appengine/google/appengine/runtime/wsgi.py", line 85, in LoadObject
|     obj = __import__(path[0])
|   File "/build/clusterfuzz/src/appengine/server.py", line 22, in <module>
|     from handlers import base_handler
|   File "/build/clusterfuzz/src/appengine/handlers/base_handler.py", line 39, in <module>
|     from libs import form
|   File "/build/clusterfuzz/src/appengine/libs/form.py", line 23, in <module>
|     from libs import helpers
|   File "/build/clusterfuzz/src/appengine/libs/helpers.py", line 22, in <module>
|     from datastore import data_handler
|   File "/build/clusterfuzz/src/appengine/python/datastore/data_handler.py", line 27, in <module>
|     from base import tasks
|   File "/build/clusterfuzz/src/appengine/python/base/tasks.py", line 27, in <module>
|     from datastore import ndb_utils
|   File "/build/clusterfuzz/src/appengine/python/datastore/ndb_utils.py", line 17, in <module>
|     from datastore import ndb_patcher
|   File "/build/clusterfuzz/src/appengine/python/datastore/ndb_patcher.py", line 37, in <module>
|     from google.cloud import datastore
|   File "/build/clusterfuzz/src/appengine/third_party/google/cloud/datastore/__init__.py", line 61, in <module>
|     from google.cloud.datastore.batch import Batch
|   File "/build/clusterfuzz/src/appengine/third_party/google/cloud/datastore/batch.py", line 24, in <module>
|     from google.cloud.datastore import helpers
|   File "/build/clusterfuzz/src/appengine/third_party/google/cloud/datastore/helpers.py", line 29, in <module>
|     from google.cloud.datastore_v1.proto import datastore_pb2
[...]

PR #692, opened 2019-07-16
Fix protobuf in ENV to 3.6.1 (#690).

oliverchang:
This is blocked on moving to cloud datastore NDB.


oliverchang:
should fix the local run_server issue

PR #696, opened 2019-07-17
Add deployment bucket need for metadata mock server, fix docs.

inferno-chromium:


inferno-chromium:
Ideally i think we can figure out project-num and deployment-bucket using project name, but will leave it for future ?

PR #697, opened 2019-07-18
Don't add reward-topanel label on non-security bugs.

oliverchang:


oliverchang:
PTAL. this was a regression from my issue tracker refactoring.

PR #704, opened 2019-07-19
Revert "Fix formatting in render-either and impact-button HTML components."

inferno-chromium:
Reverts google/clusterfuzz#689

This breaks rendering and causes spaces between anchor tags on impact button. This was done as shown by example 2 in https://css-tricks.com/fighting-the-space-between-inline-block-elements/#article-header-id-0


inferno-chromium:
See rendering on https://clusterfuzz.com/?impact=stable&reproducible=yes

inferno-chromium:
@Dor1s - you will now love html/css even more :)

Dor1s:
//cc @jonathanmetzman apparently those line endings inside HTML tags are intentional.

PR #705, opened 2019-07-19
Fix dangling process issue in mac init.

oliverchang:
If we break out of the iteration by returning, the `proc.kill` after the
`yield` is never run.

### DIFF
--- a/src/python/bot/init_scripts/mac.py
+++ b/src/python/bot/init_scripts/mac.py
@@ -32,9 +32,11 @@ def _execute(cmd):
   """Execute command and return output as an iterator."""
   proc = subprocess.Popen(
       cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
-  for line in iter(proc.stdout.readline, b''):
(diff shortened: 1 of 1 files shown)


oliverchang:
TBR

oliverchang:
@inferno-chromium is there a way to restart all our mac bots without having to SSH into them? Some of them can't be accessed because we hit the process limit.

Issue is here:

```
def _execute(cmd):
  """Execute command and return output as an iterator."""
  proc = subprocess.Popen(
      cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)

  for line in iter(proc.stdout.readline, b''):
    yield line

  proc.kill()
```

Whatever iterates on _execute can exit break out of the loop. In this case, `proc.kill` is never done.

inferno-chromium:
LGTM, thanks for discovering. I am asking on ticket to restart. there is an internal link to restart that i have but it also fails when it cannot ssh. but infra folks have something else to just force reboot machine which i think you need in this case.

PR #711, opened 2019-07-19
BigQuery table separation - MAB experiments

mukundv-chrome:
Separated the bigquery table into separate tables for production and experimentation.

### DIFF
--- a/configs/test/bigquery/datasets.yaml
+++ b/configs/test/bigquery/datasets.yaml
@@ -223,6 +223,20 @@ resources:
             type: STRING
           - name: probability
             type: FLOAT
+          - name: run_count
+            type: INTEGER
+  - name: fuzz-strategy-experiments-table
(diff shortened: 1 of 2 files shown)


configs/test/bigquery/datasets.yaml
```diff
@@ -223,6 +223,20 @@ resources:
             type: STRING
           - name: probability
             type: FLOAT
+          - name: run_count
+            type: INTEGER
+  - name: fuzz-strategy-experiments-table
```
mbarbella-chromium: I'm unsure if we'll be able to deploy the modifications to the other table since we've already updated the schema. I'm willing to try it but it may be best to leave it out for now, and we can re-add it later.

mbarbella-chromium: Actually, let's keep this as-is. I'm willing to modify the tables manually now if needed so this doesn't cause a headache later.

PR #718, opened 2019-07-22
Fix handler rules for OSS-Fuzz.

inferno-chromium:


src/appengine/handlers/bots.py
```diff
@@ -90,6 +90,7 @@ def _get_host_workers_heartbeats():
 class Handler(base_handler.Handler):
   """Handler that gets the bot list."""
 
+  @handler.check_admin_access_if_oss_fuzz
   @handler.check_user_access(need_privileged_access=False)
```
Dor1s: Is this line still needed in all of these?

inferno-chromium: Yes above condition only exclude oss-fuzz.

PR #719, opened 2019-07-22
[Reproduce tool] Fix an exception and don't terminate "chrome".

mbarbella-chromium:
PTAL. I'm planning on enabling the tool on OSS-Fuzz once this lands and testing for any issues there.

### DIFF
--- a/src/local/butler/reproduce.py
+++ b/src/local/butler/reproduce.py
@@ -227,6 +227,9 @@ def _reproduce_crash(testcase_url, build_directory):
   # Get the return code and symbolized stacktrace before cleaning up.
   return_value = (result.is_crash(), result.get_stacktrace())
 
+  # Clean up the temporary root directory created in prepare environment.
+  shell.remove_directory(environment.get_value('ROOT_DIR'))
+
   return return_value
 
 
@@ -245,6 +248,3 @@ def execute(args):
 
   print('{status_message} Output:\n\n{output}'.format(
       status_message=status_message, output=stacktrace))
-
-  # Clean up the temporary root directory created in prepare environment.
-  shell.remove_directory(environment.get_value('ROOT_DIR'))
--- a/src/local/butler/reproduce.py
+++ b/src/local/butler/reproduce.py
@@ -162,6 +162,10 @@ def _prepare_initial_environment(build_directory):
   environment.set_value('BUILD_DIR', build_directory)
   environment.set_value('BUILDS_DIR', build_directory)
 
+  # Avoid kililng the application we're testing as developers may have it
+  # running on the side.
+  environment.set_value('KILL_PROCESSES_MATCHING_APP_NAME', 'False')
+
 
 def _update_environment_for_testcase(testcase, build_directory):
   """Update environment variables that depend on the test case."""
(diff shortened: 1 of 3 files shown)


src/python/system/process_handler.py
```diff
@@ -500,14 +500,17 @@ def terminate_stale_application_instances():
...
-  app_name = environment.get_value('APP_NAME')
   builds_directory = environment.get_value('BUILDS_DIR')
   llvm_symbolizer_filename = environment.get_executable_filename(
       'llvm-symbolizer')
   platform = environment.platform()
   start_time = time.time()
 
-  processes_to_kill = [app_name]
+  processes_to_kill = []
+  if environment.get_value('KILL_PROCESS_MATCHING_APP_NAME', True):
```
mbarbella-chromium: This will likely need a better long-term fix (e.g. running the tool as another user or isolated in some other way), but if we assume users are running the tool directly it's fairly annoying to have it constantly terminate "chrome".

src/local/butler/reproduce.py
```diff
@@ -220,24 +226,27 @@ def _reproduce_crash(testcase_url, build_directory):
...
     return
 
-  if result.is_crash():
+  if is_crash:
     status_message = 'Test case reproduced successfully.'
   else:
     status_message = 'Unable to reproduce desired crash.'
 
   print('{status_message} Output:\n\n{output}'.format(
-      status_message=status_message, output=result.get_stacktrace()))
```
mbarbella-chromium: This call requires that we still have access to the config directory, deleted when we clean up root dir.

src/local/butler/reproduce.py
```diff
@@ -180,6 +180,12 @@ def _update_environment_for_testcase(testcase, build_directory):
   build_manager.set_environment_vars(
       [environment.get_value('FUZZER_DIR'), build_directory])
 
+  # Avoid killing the "chrome" application for the user running the tool, as it
+  # will commonly be open on the side. We expect the tool to properly terminate
+  # instances it launches regardless.
+  if environment.get_value('APP_NAME') == 'chrome':
```
inferno-chromium: Why only chrome, i think we shouldn't be killing for anything else, only killing children when running the reproduce tool locally. lets say we are fuzzing firefox, ie, etc. can remove the if and fixup comment.

src/local/butler/reproduce.py
```diff
@@ -220,24 +226,27 @@ def _reproduce_crash(testcase_url, build_directory):
   result = testcase_manager.test_for_crash_with_retries(testcase, testcase_path,
                                                         timeout)
 
+  # Get the return code and symbolized stacktrace before cleaning up.
+  return_value = (result.is_crash(), result.get_stacktrace())
```
inferno-chromium: Instead of this, why not move the remove_directory in a cleanup function at very end. dont like tuples being used here.

src/python/system/process_handler.py
inferno-chromium: KILL_PROCESSES_MATCHING_APP_NAME with PROCESSES ?

src/local/butler/reproduce.py
mbarbella-chromium: Good point. I'm pretty sure everything should just work as-is without terminating the app we're fuzzing. There might be some annoying corner case but I'm going to try doing this unconditionally for now.

src/local/butler/reproduce.py
mbarbella-chromium: The main issue is that _reproduce_crash is the entry point for tests, and I don't like the idea of having to do additional cleanup after calling this.

inferno-chromium: Fine as-is then.

mbarbella-chromium: I think I might see a way around it. I'll try to address it if possible but if I can't find a good solution I'll skip and land it as-is.

mbarbella-chromium: Discussed some of this offline.

Looks like the conversion to pyfakefs is going to be a bit more involved than I thought. I'm going to move this to a TODO and change it back later, but leave it as-is for now.

src/local/butler/reproduce.py

src/python/system/process_handler.py
mbarbella-chromium: Done

PR #723, opened 2019-07-23
BigQuery fix for Fuzz Strategy Selection Cron

mukundv-chrome:


src/appengine/handlers/cron/fuzz_strategy_selection.py
```diff
@@ -175,10 +175,10 @@ def _store_probabilities_in_bigquery(data):
             row['run_count']
     }
     bigquery_data.append(big_query.Insert(row=bigquery_row, insert_id=None))
-
-  client = big_query.Client(
-      dataset_id='main', table_id='fuzz_strategy_experiments')
-  client.insert(bigquery_data)
+  if bigquery_data:
```
Dor1s: Do you want to log an error or a warning in `else` branch?

mukundv-chrome: Might be noisy during this experimentation, but would be useful in the long term. ill add it

mbarbella-chromium: Seems fine to me so long as it's a regular log rather than error or warning. We expect this to happen if we aren't running with multi-armed bandit enabled.

src/appengine/handlers/cron/fuzz_strategy_selection.py
```diff
@@ -175,10 +176,14 @@ def _store_probabilities_in_bigquery(data):
             row['run_count']
     }
     bigquery_data.append(big_query.Insert(row=bigquery_row, insert_id=None))
+
   if bigquery_data:
     client = big_query.Client(
         dataset_id='main', table_id='fuzz_strategy_experiments')
     client.insert(bigquery_data)
+  else:
+  	logs.log("No fuzz strategy distribution data was found to upload to "
```
mbarbella-chromium: Nit: whitespace

PR #726, opened 2019-07-24
[fuchsia] Update log capture.

flowerhack:
Previously Fuchsia relied on an internal "loglistener" tool to collect
logs.  This tool no longer works and thus the scripts in
platforms/fuchsia/util/* had to be updated accordingly.

Bonus: we can now get actual symbolized logs!

This also gets us slightly further along in detecting crashes: the
"fuzz" function on the runner now checks that the symbolized stacktrace
indicates a crash, and returns the appropriate output and return code
accordingly.

Unfortunately, since Fuchsia's stackframe format (for now) is different
than conventional ASAN, not everything "just works" yet :(  Next items
of business: see if we can make our format match ASAN's, and, failing
that, add some custom stack analysis logic to stack_analyzer.py.

Other updates in this fix:

* We now only need to download from one bucket, FUCHSIA_BUILD_URL,
instead of two buckets (FUCHSIA_RESOURCES_URL is gone now), to update
our Fuchsia builds.

* Fixed default permissions on the symbolizer binaries, so that they can
actually be used by Clusterfuzz.

* Suppressed output from Fuchsia fuzzers by default, so that overeager
functions in the stack analyzer don't kill fuzzers that are in the
process of running and collecting log data.

* Change that should be backported to the fuzzer scripts in Fuchsia
itself: turns out Fuchsia's version of `scp` doesn't correctly parse the
-p flag in all cases (though it handles -P just fine, go figure).

* Change that should be backported to the fuzzer scripts in Fuchsia
itself: added a `results_output` function, used to explicitly call crash
directories by their true name, instead of the symlinked "latest"
directory, to avoid any issues with naming, raciness, etc.

* Change that should be backported to the fuzzer scripts in Fuchsia
itself: we need last_fuzz_cmd; keep it in that interface.


flowerhack:
@oliverchang for review

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -363,6 +371,41 @@ def get_command(self, additional_args=None):
...
+
+    # Clusterfuzz assumes that the Libfuzzer output points to an absolute path,
+    # where it can find the crash file.
+    # This doesn't work in our case due to how Fuchsia is run.
+    # So, we make a new file, change the appropriate line with a regex to point
+    # to the true location. Apologies for the hackery.
+    crash_testcase_file_path = None
+    for outfile in os.listdir(self.fuzzer.results_output()):
+      if os.path.isfile(os.path.join(self.fuzzer.results_output(),
+                                     outfile)) and 'crash-' in outfile:
```
oliverchang: there's also at least 'timeout-' and 'oom-'. a better way may be to parse this from the logs (i.e. do lines 395... first) and then use that to copy the right file.

flowerhack:
FYI @eepeep

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -345,16 +348,21 @@ def __init__(self, executable_path, default_args=None):
...
-    package, target = environment.get_value('FUZZ_TARGET').split('/')
-    self.fuzzer = Fuzzer(self.device, package, target)
     self.device.set_ssh_option('StrictHostKeyChecking no')
     self.device.set_ssh_option('UserKnownHostsFile=/dev/null')
     self.device.set_ssh_identity(fuchsia_pkey_path)
+    # Fuchsia fuzzer names have the format {package_name}/{binary_name}.
+    # TODO(ochang): Properly handle fuzzers with '/' in the binary name.
+    package, target = environment.get_value('FUZZ_TARGET').split('/')
+    test_data_dir = os.path.join(fuchsia_resources_dir_plus_build, "test_data",
+                                 "fuzzing", package, target)
```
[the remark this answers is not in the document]
flowerhack: I put `os.path.join('test_data', 'fuzzing')` into a constant, but i can't include that full path because we don't parse environment variables until the __init__ function.  Let me know if you want more moved into the class constants or if this looks good to you

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -363,6 +371,41 @@ def get_command(self, additional_args=None):
...
+    for outfile in os.listdir(self.fuzzer.results_output()):
+      if os.path.isfile(os.path.join(self.fuzzer.results_output(),
+                                     outfile)) and 'crash-' in outfile:
+        crash_testcase_file_path = os.path.join(self.fuzzer.results_output(),
+                                                outfile)
+    _, new_file_handle_path = tempfile.mkstemp()
+    with open(new_file_handle_path, 'w') as new_file:
+      with open(self.fuzzer.results_output('fuzz-0.log')) as old_file:
+        for line in old_file:
+          new_text, num_replacements = re.subn(
```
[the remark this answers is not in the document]
flowerhack: good point! changed

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -371,14 +414,26 @@ def fuzz(self,
...
-    fuzzer_process_result.return_code = 0
-    fuzzer_process_result.output = ''
+    if is_crash:
+      fuzzer_process_result.return_code = 1
+    else:
+      fuzzer_process_result.return_code = 0
+    fuzzer_process_result.output = symbolized_output
     fuzzer_process_result.time_executed = 0
-    fuzzer_process_result.command = self.fuzzer.last_fuzz_cmd
+    fuzzer_process_result.command = environment.get_value('FUZZ_TARGET')
```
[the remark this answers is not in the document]
flowerhack: mm, good point. the scripts shouldn't have removed last_fuzz_cmd; just added that back in

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -371,14 +414,23 @@ def fuzz(self,
...
     self._test_qemu_ssh()
-    self.fuzzer.run([])
-    # TODO(flowerhack): Modify fuzzer.run() to return a ProcessResult, rather
-    # than artisinally handcrafting one here.
+    self.fuzzer.start([])
+    self.fetch_and_process_logs_and_crash()
+
+    # Checking for is_crash here, instead of a more conventional location,
+    # since we're having to manually construct a ProcessResult.
+    with open(self.fuzzer.results_output('fuzz-0.log')) as logfile:
```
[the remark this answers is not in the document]
flowerhack: good point. moved this into interface

src/python/bot/fuzzers/libfuzzer.py
flowerhack: rejiggered it around in the way suggested, lmk what you think?

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -364,6 +373,38 @@ def get_command(self, additional_args=None):
...
     return self.ssh_command('ls')
 
+  def fetch_and_process_logs_and_crash(self):
+    """Fetch symbolized logs and crashes."""
+    # Get the crash from the device.
+    self.device.fetch(
+        self.fuzzer.data_path('crash*'), self.fuzzer.results_output())
+
+    for logname in os.listdir(self.fuzzer.results_output()):
+      if logname in self.fuzzer.logfile:
```
oliverchang: nit: use os.path.basename(self.fuzzer.logfile) == logname ?

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -364,6 +373,38 @@ def get_command(self, additional_args=None):
...
+    _, new_file_handle_path = tempfile.mkstemp()
+    with open(new_file_handle_path, 'w') as new_file:
+      with open(self.fuzzer.logfile) as old_file:
+        for line in old_file:
+          line_match = re.match(crash_location_regex, line)
+          if line_match:
+            crash_testcase_file_path = os.path.join(
+                self.fuzzer.results_output(),
+                line_match.group(3).replace('data/', ''))
+          new_text = re.sub(crash_location_regex,
```
oliverchang: now that the code has been re-structured, we can indent this bit and just do:

`line = re.sub(crash_location_regex, ..., line`

This way we can get rid of line 391, and the `new_text` variable.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -363,6 +372,37 @@ def get_command(self, additional_args=None):
     # list" and then run that fuzzer.
     return self.ssh_command('ls')
 
+  def fetch_and_process_logs_and_crash(self):
+    """Fetch symbolized logs and crashes."""
+    # Get the crash from the device.
+    self.device.fetch(
+        self.fuzzer.data_path('crash*'), self.fuzzer.results_output())
```
oliverchang: this needs to take into account the other types of files such as oom- etc as well. Should we also do this in the parsing bits from line 390 onwards?

Issue #663, opened 2019-07-09
Support multiple fuzz target archives for a job.


oliverchang:
Rather than a single archive containing all targets.

This may be needed to support some larger internal targets.

inferno-chromium:
Is this done, please close if yes.

oliverchang:
Still testing this. Will close once I confirm things work.

oliverchang:
This is done.

PR #727, opened 2019-07-24
Don't pick targets which aren't built yet. (#663).

oliverchang:


src/python/google_cloud_utils/storage.py
```diff
@@ -325,19 +337,23 @@ def _get_object_properties(self, remote_path):
...
-        'size':
-            os.path.getsize(fs_path),
-        'metadata':
-            self._get_metadata(bucket, path),
+    data = {
+        'bucket': bucket,
+        'name': path,
     }
 
+    if not os.path.isdir(fs_path):
```
inferno-chromium: Can you add a comment on why we don't send these for directories. Is it ok to generalize ?

src/python/google_cloud_utils/storage.py
```diff
@@ -188,14 +188,26 @@ def list_blobs(self, remote_path):
...
+    iterator = bucket.list_blobs(prefix=path, delimiter=delimiter)
+    for blob in iterator:
       properties['bucket'] = bucket_name
       properties['name'] = blob.name
       properties['updated'] = blob.updated
       properties['size'] = blob.size
 
       yield properties
 
+    if not recursive:
```
inferno-chromium: Can you explain this block more in a comment, is it for dirs ?

src/python/build_management/build_manager.py
```diff
@@ -1154,13 +1154,18 @@ def _get_targets_list(bucket_path):
...
+  targets_list_path = os.path.join(bucket_dir_path, TARGETS_LIST_FILENAME)
   data = storage.read_data(targets_list_path)
   if not data:
     return None
 
-  return data.splitlines()
+  # Filter out targets which are not yet built.
+  targets = data.splitlines()
+  listed_targets = set(
+      os.path.basename(path.rstrip('/'))
```
inferno-chromium: shouldn't list_blobs strip the / or is too problematic there ?

src/python/google_cloud_utils/storage.py
oliverchang: added a comment.

src/python/google_cloud_utils/storage.py
oliverchang: added a comment.

src/python/build_management/build_manager.py
oliverchang: I think it's good to leave the '/'. It's what `gsutil ls` returns.

src/python/google_cloud_utils/storage.py
```diff
@@ -188,14 +188,27 @@ def list_blobs(self, remote_path):
...
     properties = {}
 
-    for blob in bucket.list_blobs(prefix=path):
+    if recursive:
+      delimiter = None
+    else:
+      delimiter = '/'
+
+    iterator = bucket.list_blobs(prefix=path, delimiter=delimiter)
+    for blob in iterator:
```
Dor1s: Is it safe to rely on this behavior regarding delimiter presence? Maybe adding an extra check for `if recursive` wouldn't hurt?

PR #729, opened 2019-07-24
[Reproduce tool] Add basic support for Android.

mbarbella-chromium:


src/local/butler/reproduce.py
```diff
@@ -208,6 +209,28 @@ def _get_testcase_id_from_url(testcase_url):
...
+def _prepare_environment_for_android(testcase_path):
+  """Additional environment overrides needed to run on an Android device."""
+  environment.set_value('OS_OVERRIDE', 'ANDROID')
+
+  # Bail out if we don't have an Android device connected.
+  if not environment.get_value('ANDROID_SERIAL'):
+    # TODO(mbarbella): Handle the one-device case gracefully.
+    raise errors.ReproduceToolUnrecoverableError('Please set ANDROID_SERIAL.')
+
+  # TODO(mbarbella): Should we run device.initialize_device?
```
mbarbella-chromium: This will be trivial to add, but some of the settings changes and other modifications seem invasive. Should we do it anyway?

src/python/bot/tasks/setup.py
```diff
@@ -63,33 +63,35 @@ def _copy_testcase_to_device_and_setup_environment(testcase,
 
   # The following steps need privileged job access.
   job_type_has_privileged_access = environment.get_value('PRIVILEGED_ACCESS')
-  if job_type_has_privileged_access:
```
mbarbella-chromium: There should be no functional changes here, just noticed that it was formatted in a way that would have been better served by an early bailout.

mbarbella-chromium:
PTAL when you have time. I still want to test this a bit more, so no rush.

inferno-chromium:
@oliverchang - can you help to review.

src/local/butler/reproduce.py
oliverchang: Maybe add a warning/prompt and then do this?

src/local/butler/reproduce.py
```diff
@@ -220,6 +243,16 @@ def _reproduce_crash(testcase_url, build_directory):
   testcase_path = _download_testcase(testcase_id, testcase, configuration)
   _update_environment_for_testcase(testcase, build_directory)
 
+  # Validate that we're running on the right platform for this test case.
+  platform = environment.platform().lower()
+  if testcase.platform == 'android' and platform == 'linux':
+    _prepare_environment_for_android(testcase_path)
```
oliverchang: maybe re-structure the conditions a little here such that we give a better error condition if testcase platform is 'android' but environment platform != 'linux'?

src/local/butler/reproduce.py
inferno-chromium: +1 to prompt. And then we need to call initialize_device, otherwise things wont work, like setup_host_and_device_forwarder_if_needed, setup_asan_if_needed are all called from there. either can create another lightweight version of the function for just reproduce itself, like which calls e.g. setup_adb, etc.

src/local/butler/reproduce.py
```diff
@@ -208,6 +209,28 @@ def _get_testcase_id_from_url(testcase_url):
...
+
+  # Bail out if we don't have an Android device connected.
+  if not environment.get_value('ANDROID_SERIAL'):
+    # TODO(mbarbella): Handle the one-device case gracefully.
+    raise errors.ReproduceToolUnrecoverableError('Please set ANDROID_SERIAL.')
+
+  # TODO(mbarbella): Should we run device.initialize_device?
+
+  # Push the test case and build APK to the device.
+  apk_path = environment.get_value('APP_PATH')
```
inferno-chromium: add assert or something to check if APK_PATH is set correctly.

src/local/butler/reproduce.py
```diff
@@ -208,6 +209,28 @@ def _get_testcase_id_from_url(testcase_url):
...
+  device.install_application_if_needed(apk_path, True)
```
inferno-chromium: use named argument force_update=True

src/local/butler/reproduce.py
```diff
@@ -208,6 +209,28 @@ def _get_testcase_id_from_url(testcase_url):
...
+
+  testcase_manager.get_command_line_for_application(
+      write_command_line_file=True)
+
```
inferno-chromium: PKG_NAME is no set, which will cause gestures to not work. 

https://github.com/google/clusterfuzz/blob/master/src/python/platforms/android/gestures.py#L39

src/local/butler/reproduce.py
```diff
@@ -208,6 +209,28 @@ def _get_testcase_id_from_url(testcase_url):
...
+  environment.set_value('FUZZ_INPUTS', os.path.dirname(testcase_path))
```
inferno-chromium: Why is this android specific, FUZZ_INPUTS should be set for all.

src/local/butler/reproduce.py
```diff
@@ -208,6 +209,28 @@ def _get_testcase_id_from_url(testcase_url):
...
+  environment.set_value('OS_OVERRIDE', 'ANDROID')
+
+  # Bail out if we don't have an Android device connected.
+  if not environment.get_value('ANDROID_SERIAL'):
+    # TODO(mbarbella): Handle the one-device case gracefully.
+    raise errors.ReproduceToolUnrecoverableError('Please set ANDROID_SERIAL.')
+
+  # TODO(mbarbella): Should we run device.initialize_device?
+
+  # Push the test case and build APK to the device.
```
inferno-chromium: Why not move all this code from
https://github.com/google/clusterfuzz/blob/442a074fefbe5fdcbbe0df046e985c850c64a73e/src/python/build_management/build_manager.py#L665 into android device.py and then use the helper here. we need to avoid code duplication in reproduce tool as much as possible as otherwise we will have more maintenance later.

src/local/butler/reproduce.py
mbarbella-chromium: Added a warning. Based on Abhishek's point I think it's fine to always require that this happen to continue, but the prompt is still a great idea to make sure users know what they're getting into.

mbarbella-chromium:
There's still one open comment thread regarding PKG_NAME, but otherwise I think I've addressed all comments. This still needs more local testing, but should be close to ready.

inferno-chromium:
> There's still one open comment thread regarding PKG_NAME, but otherwise I think I've addressed all comments. This still needs more local testing, but should be close to ready.

ah right, can ignore this one. PKG_NAME is always in job and other case https://github.com/google/clusterfuzz/blob/299f2e2ba20c4c6eb1596fe350f58c9f1bd52c14/src/python/bot/tasks/setup.py#L71 is custom apk and uncommon, can skip for now.

src/local/butler/reproduce.py
```diff
@@ -176,6 +178,10 @@ def _update_environment_for_testcase(testcase, build_directory):
   app_path = os.path.join(build_directory, environment.get_value('APP_NAME'))
   environment.set_value('APP_PATH', app_path)
 
+  if not os.path.exists(app_path):
```
inferno-chromium: if not app_path or not os.path.exists

src/local/butler/reproduce_tool/prompts.py
```diff
@@ -0,0 +1,28 @@
...
+def get_string(prompt, add_colon=True):
+  """Prompt the user for a string from console input."""
+  if add_colon:
+    prompt = prompt + ': '
+  return input(prompt)
+
+
+def get_boolean(prompt):
+  """Return a boolean representing a yes/no answer to a prompt."""
+  result = get_string(prompt + ' (Y/n): ', add_colon=False)
```
inferno-chromium: Why not just do ' (Y/n)' so that colon is added always ?

src/python/platforms/android/device.py
```diff
@@ -438,3 +438,20 @@ def setup_host_and_device_forwarder_if_needed():
...
+
+def update_build(apk_path, force_update=True):
+  """Prepares the device and updates the build if necessary."""
+  # Prepare device for app install.
+  initialize_device()
+
+  # On Android, we may need to write a command line file. We do this in
+  # advance so that we do not have to write this to the device multiple
+  # times.
+  # TODO(mbarbella): Build code should not depend on fuzzing.
```
inferno-chromium: Please remove this TODO, this is now done. also, can move "from fuzzing " into the module imports at top.

src/local/butler/reproduce.py
mbarbella-chromium: Factored this out into a new function (for mocking) and used a default value instead.

src/local/butler/reproduce_tool/prompts.py
mbarbella-chromium: Done

src/python/platforms/android/device.py
mbarbella-chromium: Unfortunately we have the same problem in this file. testcase_manager depends on a lot of the rest of the codebase, so importing it from the files with these sorts of helper functions will always be problematic. Not sure what the best way to handle it is, but updated the TODO so it makes sense in this file.

PR #732, opened 2019-07-24
[Python 2->3] Re-run futurize stage 2.

mbarbella-chromium:
It seems like the tool isn't meant to run over the same code multiple times, so this contains manual edits to revert some changes that seem unnecessary (such as assuming we wanted Python 2's items behavior rather than Python 3's).

Part of #406

### DIFF
--- a/src/appengine/handlers/reproduce_tool/get_config.py
+++ b/src/appengine/handlers/reproduce_tool/get_config.py
@@ -16,7 +16,7 @@
 from future import standard_library
 standard_library.install_aliases()
 
-import urllib
+from urllib import parse
 
 from config import db_config
 from datastore import data_handler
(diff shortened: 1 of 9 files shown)


mbarbella-chromium:
PTAL when you have time. Just some minor issues caught by futurize on some recent changes.

src/appengine/handlers/reproduce_tool/get_config.py
```diff
@@ -16,7 +16,7 @@
 from future import standard_library
 standard_library.install_aliases()
 
-import urllib
+from urllib import parse
```
inferno-chromium: we usually do "import urllib.parse" in other places so that a generic parse function can't conflict.

PR #728, opened 2019-07-24
Move fuzz target chmod to somewhere more general.

oliverchang:

PR #733, opened 2019-07-25
Trying to fix travis

oliverchang:


oliverchang:
no idea what issue this was, but it seems to have fixed travis. TBRing

oliverchang:
Yeah, but not sure why this isn't an issue on container builder.

Dor1s:
Hm, and actually that combination of flags was just moved from another place in #728. Doesn't make any sense indeed :|

PR #734, opened 2019-07-25
[MAB] Cron job refactor and added AFL to cron job

mukundv-chrome:
Refactored the multi-armed bandit cron job to make it easier to add other engines
Added AFL to cron job


src/appengine/handlers/cron/fuzz_strategy_selection.py
```diff
@@ -21,6 +21,7 @@
 metric to be for edges, crash, features, or units. Currently based on new
 edges."""
 
+from bot.fuzzers import strategy
```
mbarbella-chromium: src/python/bot isn't accessible from app engine. We can get rid of that restriction but it makes our App Engine deployments take a bit longer. Maybe we will have to move this to somewhere like src/python/fuzzing after all.

mbarbella-chromium: (Note: see src/appengine/skip_files.yaml for context on the previous comment)

src/appengine/handlers/cron/fuzz_strategy_selection.py
```diff
@@ -124,6 +100,14 @@
       GROUP BY
         strategy)))
 """
+
+STRATEGY_SUBQUERY = """
```
mbarbella-chromium: Nit: could you rename all of these queries that still require a .format with _FORMAT or _TEMPLATE at the end of their names?

src/appengine/handlers/cron/fuzz_strategy_selection.py
```diff
@@ -33,6 +34,11 @@
 MEDIUM_TEMPERATURE_PARAMETER = .25
 LOW_TEMPERATURE_PARAMETER = .15
 
+# A list of tuples of engine names and associated strategies to include in
+# query.
+ENGINE_LIST = [('libFuzzer', strategy.libfuzzer_query_strategy_list),
```
Dor1s: nit: would be better to use `collections.namedtuple`

src/python/datastore/data_types.py
```diff
@@ -1131,6 +1131,7 @@ class FuzzStrategyProbability(Model):
   probability_medium_temperature = ndb.FloatProperty()
   probability_high_temperature = ndb.FloatProperty()
   probability_low_temperature = ndb.FloatProperty()
+  engine = ndb.StringProperty()
```
Dor1s: I think you need to run `python butler.py generate_datastore_models` to re-generate `src/go/cloud/db/types/types.go` file.

src/python/fuzzing/strategy.py
```diff
@@ -57,6 +57,16 @@
     VALUE_PROFILE_STRATEGY,
 ]
 
+# List of strategies to include in multi-armed bandit query for libfuzzer
+libfuzzer_query_strategy_list = [
+    CORPUS_MUTATION_RADAMSA_STRATEGY, RANDOM_MAX_LENGTH_STRATEGY,
```
Dor1s: nit: can we use alphabetical order here?

src/python/fuzzing/strategy.py
```diff
@@ -57,6 +57,16 @@
     VALUE_PROFILE_STRATEGY,
 ]
 
+# List of strategies to include in multi-armed bandit query for libfuzzer
+libfuzzer_query_strategy_list = [
```
Dor1s: Why does this need to be in this file? I think you can include it in `src/appengine/handlers/cron/fuzz_strategy_selection.py` as that seems to be the only place where this is used.

src/appengine/handlers/cron/fuzz_strategy_selection.py
```diff
@@ -225,4 +239,5 @@ class Handler(base_handler.Handler):
   @handler.check_cron()
   def get(self):
     """Process all fuzz targets and update FuzzStrategy weights."""
-    _query_and_upload_strategy_probabilities()
+    for engine_name, strategy_list in ENGINE_LIST:
```
Dor1s: nit: would need to update this once `ENGINE_LIST` is changed to a namedtuple

src/appengine/handlers/cron/fuzz_strategy_selection.py
```diff
@@ -124,38 +113,60 @@
...
-      high_temperature_query=BANDIT_PROBABILITY_SUBQUERY.format(
+  formatted_query = BANDIT_PROBABILITY_QUERY_FORMAT.format(
+      high_temperature_query=BANDIT_PROBABILITY_SUBQUERY_FORMAT.format(
           temperature_type='high',
-          temperature_value=HIGH_TEMPERATURE_PARAMETER),
-      low_temperature_query=BANDIT_PROBABILITY_SUBQUERY.format(
-          temperature_type='low', temperature_value=LOW_TEMPERATURE_PARAMETER),
-      medium_temperature_query=BANDIT_PROBABILITY_SUBQUERY.format(
+          temperature_value=HIGH_TEMPERATURE_PARAMETER,
+          strategies=','.join([
```
Dor1s: lines 142 - 145 are repeated three times, can you please generate the strategies string just once and then use it for all the queries?

src/python/fuzzing/strategy.py
mukundv-chrome: had these here just because we define the strategy named tuple here, but yes makes more sense to keep in fuzz_strategy_selection

PR #735, opened 2019-07-26
[stack_analyzer] Add "libfuzzer: deadly signal" to crash detection.

flowerhack:
This was brought up mostly by a desire to have a convenient test case
for bringing Fuchsia onto Clusterfuzz; simply throwing a __builtin_trap
is a pretty reliable way to get a libfuzzer crash.

However, Clusterfuzz doesn't currently recognize this form of crash, so
we're adding it now!

### DIFF
--- a/src/python/crash_analysis/stack_parsing/stack_analyzer.py
+++ b/src/python/crash_analysis/stack_parsing/stack_analyzer.py
@@ -103,7 +103,8 @@ KASAN_CRASH_TYPE_ADDRESS_REGEX = re.compile(
     r'BUG: KASAN: (.*) (in|on).*(addr|address) ([a-f0-9]+)')
 KASAN_GPF_REGEX = re.compile(r'general protection fault:.*KASAN')
 LIBFUZZER_TIMEOUT_REGEX = re.compile(r'.*ERROR:\s*libFuzzer:\s*timeout')
-LIBFUZZER_DEADLY_SIGNAL_REGEX = re.compile(r'.*ERROR:\s*libFuzzer:\s*deadly signal')
+LIBFUZZER_DEADLY_SIGNAL_REGEX = re.compile(
+    r'.*ERROR:\s*libFuzzer:\s*deadly signal')
 LINUX_GDB_CRASH_TYPE_REGEX = re.compile(r'Program received signal ([a-zA-Z]+),')
 LINUX_GDB_CRASH_ADDRESS_REGEX = re.compile(r'rip[ ]+([xX0-9a-fA-F]+)')
 LSAN_DIRECT_LEAK_REGEX = re.compile(r'Direct leak of ')
@@ -1493,11 +1494,11 @@ def get_crash_data(crash_data, symbolize_flag=True):
 
       # Libfuzzer deadly signal errors.
       update_state_on_match(
-        LIBFUZZER_DEADLY_SIGNAL_REGEX,
-        line,
-        state,
-        new_type='deadly signal',
-        reset=True)
+          LIBFUZZER_DEADLY_SIGNAL_REGEX,
+          line,
+          state,
+          new_type='deadly signal',
+          reset=True)
 
     if state.fatal_error_occurred:
       error_line_match = update_state_on_match(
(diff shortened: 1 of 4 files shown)


flowerhack:
@oliverchang for review

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -1321,6 +1321,17 @@ def test_asan_double_free(self):
                                   expected_state, expected_stacktrace,
                                   expected_security_flag)
 
+  def test_libfuzzer_deadly_signal(self):
+    data = self._read_test_data('libfuzzer_deadly_signal.txt')
```
oliverchang: you need to add this file :)

flowerhack: oh whups, good point. fixed!

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -1490,6 +1492,14 @@ def get_crash_data(crash_data, symbolize_flag=True):
           address_filter=lambda s: '0x' + s,
           reset=True)
 
+      # Libfuzzer deadly signal errors.
+      update_state_on_match(
+          LIBFUZZER_DEADLY_SIGNAL_REGEX,
+          line,
+          state,
+          new_type='deadly signal',
```
oliverchang: Nit: capitalise 'deadly' to 'Deadly'.

inferno-chromium: We should prefer to avoid spaces, use hyphen. Also, don't like the use of word deadly. maybe "Fatal-signal".

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -1321,6 +1321,17 @@ def test_asan_double_free(self):
...
+    expected_type = 'deadly signal'
+    expected_state = 'NULL'
+    expected_address = ''
+    expected_stacktrace = data
+    expected_security_flag = True
```
inferno-chromium: Is this really security implication, probably not, change crash_analyzer to skip this crash type.

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -1321,6 +1321,17 @@ def test_asan_double_free(self):
                                   expected_state, expected_stacktrace,
                                   expected_security_flag)
 
+  def test_libfuzzer_deadly_signal(self):
+    data = self._read_test_data('libfuzzer_deadly_signal.txt')
+    expected_type = 'deadly signal'
+    expected_state = 'NULL'
```
inferno-chromium: This is wrong, it shouldn't be null. We should get a fuzzer name there. @oliverchang wdyt ?

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
oliverchang: Good point, +1 to not security by default.

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
oliverchang: We should be parsing the stack here. Looks like the format in the test file is not something we recognise.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
inferno-chromium: qq: Are you testing this with a libFuzzer + ASan binary. Did you set ASAN_OPTIONS as handle_{sig*}=1. I was expecting this to be covered by the sanitizer instrumentation which mentions the details on deadly signal type, like Abrt, etc.

@kcc - wdyt ?

oliverchang: I think this is an issue with the fuchsia asan integration. There's no real concept of signals, so asan isn't able to dump the stack automatically. I think the current implementation falls back to the libFuzzer stack dumper when e.g. __builtin_trap() is called. Maybe longer term we can make ASan handle the fake "signal", but for now this seems OK to unblock things. @nopsledder what do you think?

inferno-chromium: Thanks for explaining. Just change to "Fatal-signal" and fine to keep this.

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
inferno-chromium: Fyi, you need to add this in https://github.com/google/clusterfuzz/blob/master/src/python/crash_analysis/crash_analyzer.py#L233 to return False, then this test should be "expected_security_flag = False"

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
inferno-chromium: This is fine as-is, see reason above.

flowerhack:
replied to all cr

Issue #523, opened 2019-06-04
some old reproducible bugs are not auto-filed


inferno-chromium:
e.g.
https://clusterfuzz.com/testcases?issue=no&open=yes&page=4&project=chromium&reproducible=yes
https://oss-fuzz.com/testcases?issue=no&open=yes&page=4&reproducible=yes
https://bugs.chromium.org/p/chromium/issues/detail?id=969499

inferno-chromium:
actually the issue is similar bug with that crash are still open. we lose the group_bug_information mapping, that seems like a issue.

Dor1s:
Other examples are:
- https://oss-fuzz.com/testcase-detail/5642897908563968
- https://oss-fuzz.com/testcase-detail/5082243548053504

from https://oss-fuzz.com/testcases?project=rapidjson&open=yes

inferno-chromium:
@Dor1s for rapidjson, i found the culprit - https://github.com/google/oss-fuzz/commit/a3c935fe9ca7f82bafa520731525e1cc38acf650 :)

Dor1s:
Thanks @inferno-chromium, good to know! :)

inferno-chromium:
Should be fixed with the two fixes above.

PR #736, opened 2019-07-26
Associate testcase with existing issue if similar testcase is uninteresting.

inferno-chromium:
This fixes remaining part of https://github.com/google/clusterfuzz/issues/523

### DIFF
--- a/src/appengine/handlers/cron/triage.py
+++ b/src/appengine/handlers/cron/triage.py
@@ -209,7 +209,7 @@ def _is_crash_important(testcase):
           data_types.FILE_UNREPRODUCIBLE_TESTCASE_MIN_CRASH_THRESHOLD)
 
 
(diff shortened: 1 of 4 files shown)


src/appengine/handlers/cron/triage.py
```diff
@@ -168,7 +209,7 @@ def is_crash_important(testcase):
           data_types.FILE_UNREPRODUCIBLE_TESTCASE_MIN_CRASH_THRESHOLD)
 
 
-def is_similar_bug_open_or_recently_closed(testcase, issue_tracker):
+def _is_similar_bug_open_or_recently_closed(testcase, issue_tracker):
```
oliverchang: nit: name doesn't make it obvious it could update an existing issue.

how about something like:  "_check_and_update_similar_bugs()" ?

inferno-chromium: Done!

PR #742, opened 2019-07-29
Don't count ignored stacks in variant task.

inferno-chromium:


inferno-chromium:
This should remove the wrong_usage crash from v8_foozie on every job.

PR #744, opened 2019-07-30
[MAB] AFL strategy selection refactor 

mukundv-chrome:
Refactored strategy selection process to work for both libFuzzer and AFL (and will let us add MAB support for other engines easier).


mbarbella-chromium:
LGTM once tests are passing.

PR #743, opened 2019-07-30
Use full arguments from job for variant task.

inferno-chromium:

PR #745, opened 2019-07-30
Revert "Use full arguments from job for variant task."

inferno-chromium:
Reverts google/clusterfuzz#743

This will break testcases that use flag files.


inferno-chromium:
we need to find a way to add required arguments from job definition in addition to minimized arguments. @mbarbella-chromium - thoughts ?

mbarbella-chromium:
I can't think of a way to do it that isn't a bit hacky, but you could find individual flags not in the original job definition but in minimized arguments, and add them to the new job's args.

inferno-chromium:
Closing for now, as i think using the job args is more important than testcase flags file args. Filed bug to track using the flags file somehow - https://github.com/google/clusterfuzz/issues/746

PR #751, opened 2019-07-30
[MAB] Added logs to MAB cron job

mukundv-chrome:


src/appengine/handlers/cron/fuzz_strategy_selection.py
```diff
@@ -228,6 +228,7 @@ def _query_and_upload_strategy_probabilities(engine):
   are based on new_edges feature."""
   strategy_data = []
   data = _query_multi_armed_bandit_probabilities(engine)
+  logs.log("Queried distribution for {}.".format(engine.name))
```
oliverchang: nit: these should all be single quotes.

PR #752, opened 2019-07-31
[fuchsia] Test Fuchsia ASAN crash output.

flowerhack:
Fuchsia's symbolizer outputs some additional information (as compared to
e.g. Linux ASAN).  However, it still matches on the appropriate regexes
and is understood by Clusterfuzz--let's add a test to make sure it stays
that way!

### DIFF
--- a/src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
+++ b/src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
@@ -1347,6 +1347,8 @@ class StackAnalyzerTestcase(unittest.TestCase):
 
   def test_fuchsia_asan(self):
     """Test for Fuchsia ASan crashes."""
+    # TODO(flowerhack): Once the duplicated frames issue is fixed for Fuchsia,
+    # update this test to recognize proper frames.
     data = self._read_test_data('fuchsia_asan.txt')
     expected_type = 'Heap-buffer-overflow\nWRITE 1'
     expected_state = 'foo_function\nfoo_function\nbar_function\n'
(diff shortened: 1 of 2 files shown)


flowerhack:
@oliverchang

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -1321,6 +1321,18 @@ def test_asan_double_free(self):
                                   expected_state, expected_stacktrace,
                                   expected_security_flag)
 
+  def test_fuchsia_asan(self):
+    """Test for Fuchsia ASan crashes."""
+    data = self._read_test_data('fuchsia_asan.txt')
+    expected_type = 'Heap-buffer-overflow'
+    expected_state = 'NULL'
```
oliverchang: to test this properly, can you get a crash with more frames above LLVMFuzzerTestOneInput? The reason this is NULL is because we ignore all the frames in the stack.

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -1345,6 +1345,18 @@ def test_libfuzzer_fuzz_target_exited(self):
                                   expected_state, expected_stacktrace,
                                   expected_security_flag)
 
+  def test_fuchsia_asan(self):
+    """Test for Fuchsia ASan crashes."""
+    data = self._read_test_data('fuchsia_asan.txt')
+    expected_type = 'Heap-buffer-overflow\nWRITE 1'
+    expected_state = 'foo_function\nfoo_function\nbar_function\n'
```
oliverchang: as discussed, we need to fix this duplicated frame appearing in the stack. Could you add a TODO? 

This CL seems OK to land for now though to get the ball rolling and since there's no functionality change.

flowerhack: added a TODO

PR #754, opened 2019-07-31
Implement stats, testcase, logs uploading for engine fuzzing.

oliverchang:
- Put log header logic into engine_common.get_log_header() function for getting log headers.
- Remove fuzzer name setting in dump_big_query_data. It sets an incorrect name anyway, and is later overwritten in fuzz_task.
- Refactor testcase_manager log/testcase uploading to make it easier to reduce its functionality.

### DIFF
--- a/src/python/bot/fuzzers/engine_common.py
+++ b/src/python/bot/fuzzers/engine_common.py
@@ -87,7 +87,7 @@ def get_testcase_run(stats, fuzzer_command):
   build_revision = fuzzer_utils.get_build_revision()
   job = environment.get_value('JOB_NAME')
   # fuzzer name is filled by fuzz_task.
-  testcase_run = fuzzer_stats.TestcaseRun('', job, build_revision,
+  testcase_run = fuzzer_stats.TestcaseRun(None, job, build_revision,
                                           current_timestamp())
 
   testcase_run['command'] = fuzzer_command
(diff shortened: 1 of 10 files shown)


src/python/fuzzing/testcase_manager.py
```diff
@@ -471,12 +468,19 @@ def run_testcase_and_return_result_in_queue(crash_queue,
...
+        upload_testcase(file_path, log_time)
 
     if upload_output:
       # Include full output for uploaded logs (crash output, merge output, etc).
       crash_result_full = CrashResult(return_code, crash_time, output)
-      upload_testcase_output(crash_result_full, file_path)
+
+      # To provide consistency between stats and logs, we use timestamp taken
+      # from stats.
+      log_time = _get_testcase_time(file_path)
```
inferno-chromium: Can we save a _get_testcase_time call since it is called twice for upload_output branch

src/python/bot/fuzzers/engine_common.py
```diff
@@ -79,16 +82,22 @@ def decide_with_probability(probability):
...
-                        fuzzer_name, fuzzer_command):
-  """Dump BigQuery stats."""
+def get_testcase_run(stats, fuzzer_command):
+  """Get testcase run for stats."""
   build_revision = fuzzer_utils.get_build_revision()
   job = environment.get_value('JOB_NAME')
-  testcase_run = fuzzer_stats.TestcaseRun(fuzzer_name_prefix + fuzzer_name, job,
-                                          build_revision, current_timestamp())
+  # fuzzer name is filled by fuzz_task.
+  testcase_run = fuzzer_stats.TestcaseRun('', job, build_revision,
```
inferno-chromium: Can we use None instead of ''

oliverchang: Done!

src/python/fuzzing/testcase_manager.py
oliverchang: Done!

PR #757, opened 2019-07-31
 [Reproduce tool] Run Xvfb and blackbox before launching test binaries.

mbarbella-chromium:
PTAL when you have time. Once this lands test cases with gestures should be working correctly, which was the last major blocker on Linux.

### DIFF
--- a/src/local/butler/reproduce.py
+++ b/src/local/butler/reproduce.py
@@ -46,6 +46,8 @@ from system import environment
 from system import new_process
 from system import shell
 
+DISPLAY = ':99'
+
 
 class SerializedTestcase(object):
   """Minimal representation of a test case."""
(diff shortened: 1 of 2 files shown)


src/local/butler/reproduce.py
```diff
@@ -132,6 +134,24 @@ def _download_testcase(testcase_id, testcase, configuration):
   return testcase_path
 
 
+def _setup_x():
+  """Start Xvfb and blackbox before running the test application."""
+  environment.set_value('DISPLAY', ':99')
```
inferno-chromium: Can you move "':99'" on a constant.

inferno-chromium: It should check platform and bail out if not linux.

src/local/butler/reproduce.py
```diff
@@ -132,6 +134,24 @@ def _download_testcase(testcase_id, testcase, configuration):
...
+
+  xvfb_runner = new_process.ProcessRunner('/usr/bin/Xvfb')
+  xvfb_process = xvfb_runner.run(additional_args=[
+      ':99', '-screen', '0', '1280x1024x24', '-ac', '-nolisten', 'tcp'
+  ])
+  time.sleep(5)  # Allow some time for Xvfb to start.
+
+  blackbox_runner = new_process.ProcessRunner('/usr/bin/blackbox')
```
inferno-chromium: Maybe output that starting Blackbox and same for starting Xvfb

src/local/butler/reproduce.py

src/local/butler/reproduce.py

Issue #507, opened 2019-05-30
How to handle a large corpus that we cannot rsync to disk


Dor1s:
When discussing issue #462 with @inferno-chromium, another problem came up. When a target's corpus becomes too huge, we may be unable to rsync it to the disk ever. How should we handle that?

One idea is to delete the corpus and use the latest backup, but that sounds kinda disruptive and we'll also have to track and count the failures somehow prior to doing that hard reset.

inferno-chromium:
Just fyi, this isn't a major problem atm, i only saw this issue with compositor_frame_fuzzer on internal.

inferno-chromium:
This worked to fix compositor_frame_fuzzer. 

@oliverchang fyi - 30K files take like an hour to delete. Since this is not a recurring occurence i am not spending time to optimize this. If you have immediate ideas to implement, i can try, but closing for now.

PR #759, opened 2019-08-01
Limit corpus and quarantine at GCS

inferno-chromium:
We used to limit corpus after it is synced locally. This is problematic
if the sync itself fails. We limit corpus size before sync to avoid this
issue.

This should fix
https://github.com/google/clusterfuzz/issues/507

### DIFF
--- a/src/python/bot/tasks/corpus_pruning_task.py
+++ b/src/python/bot/tasks/corpus_pruning_task.py
@@ -113,8 +113,8 @@ def _get_corpus_file_paths(corpus_path):
 
 
 def _limit_corpus_size(corpus_url, size_limit):
-  """Limit number of files in a corpus directory."""
-  files_list = list(storage.list_blobs_full(corpus_url))
+  """Limit number of files in a corpus url."""
+  files_list = list(storage.list_blobs(corpus_url))
   corpus_size = len(files_list)
 
   if corpus_size <= size_limit:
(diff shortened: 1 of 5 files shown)


src/python/google_cloud_utils/storage.py
```diff
@@ -854,6 +854,17 @@ def list_blobs(cloud_storage_path, recursive=True):
     yield blob['name']
 
 
+@retry.wrap(
+    retries=DEFAULT_FAIL_RETRIES,
+    delay=DEFAULT_FAIL_WAIT,
+    function='google_cloud_utils.storage.list_blobs')
+def list_blobs_full(cloud_storage_path, recursive=True):
```
oliverchang: I think it might be cleaner to get callers to do the concatenation to get the full bucket path rather than add another function with mostly duplicated code?

Maybe add something like storage.join_path(), which helps concatenate GCS paths, and can be used in more places?

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -113,19 +112,22 @@ def _get_corpus_file_paths(corpus_path):
...
 
+  logs.log(
+      'Limit corpus at {corpus_url} from {corpus_size} to {size_limit}.'.format(
+          corpus_url=corpus_url, corpus_size=corpus_size,
+          size_limit=size_limit))
   files_to_delete = random.sample(files_list, corpus_size - size_limit)
   for file_to_delete in files_to_delete:
-    file_to_delete_full_path = os.path.join(corpus_directory, file_to_delete)
-    shell.remove_file(file_to_delete_full_path)
+    storage.delete(file_to_delete)
```
oliverchang: I wonder how slow this is for large corpora. Another thought: is there some easy way of downloading just the selection of files we choose in random.sample? We'll delete the extra ones when we sync back anyway.

src/python/google_cloud_utils/storage.py
inferno-chromium: Done!

src/python/bot/tasks/corpus_pruning_task.py
inferno-chromium: Will be measured in gce env after this lands. i have some ideas on optimization, but can invest if we really need it. Right now, it seems to only happen on one fuzzer on internal. To download a selection of files, i can't of a clean way to hack into rsync_to_disk.

PR #760, opened 2019-08-01
[Python 2->3] Run futurize as part of butler.py lint

mbarbella-chromium:
PTAL when you have a chance.


src/local/butler/lint.py
```diff
@@ -185,6 +185,14 @@ def execute(_):
   for file_path in py_changed_file_paths:
     _execute_command_and_track_error('pylint ' + file_path)
     _execute_command_and_track_error('yapf -d ' + file_path)
+    futurize_output = _execute_command_and_track_error(
+        'futurize -0 -x libfuturize.fixes.fix_unicode_keep_u '
```
inferno-chromium: please put the fixes in a list var global / butler constants.py and then just add them here with -x

src/python/platforms/fuchsia/util/device.py
```diff
@@ -12,8 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Utilities for managing Fuchsia devices."""
+from __future__ import absolute_import
+from builtins import object
+from builtins import range
 
-from host import Host
+from .host import Host
```
inferno-chromium: can you move it after the system imports.

src/local/butler/lint.py
```diff
@@ -185,6 +185,14 @@ def execute(_):
...
+        '-x libfuturize.fixes.fix_future_builtins -x lib2to3.fixes.fix_dict '
+        '-x lib2to3.fixes.fix_ws_comma -x lib2to3.fixes.fix_idioms ' +
```
inferno-chromium: Why fix_ws_comma ?

mbarbella-chromium: idioms and ws_comma are disabled by default and not explicitly disabling them makes the tool print an extra log line for each. Just did it to save some spam here. I'll leave a comment explaining that when addressing the first comment.

src/local/butler/lint.py
mbarbella-chromium: Done

src/python/platforms/fuchsia/util/device.py
mbarbella-chromium: Done

PR #762, opened 2019-08-02
Radamsa and ML RNN support in AFL

mukundv-chrome:


src/python/bot/fuzzers/afl/launcher.py
```diff
@@ -257,9 +258,23 @@ class FuzzingStrategies(object):
   FAST_CAL_RANDOM_STRATEGY = 'strategy_fast_cal_random'
   FAST_CAL_MANUAL_STRATEGY = 'strategy_fast_cal_manual'
 
-  def __init__(self):
+  def __init__(self, target_path):
     strategy_pool = strategy_selection.generate_default_strategy_pool(
         strategy_list=strategy.AFL_STRATEGY_LIST, use_generator=False)
+
+    if environment.get_value('USE_RADAMSA'):
```
oliverchang: can this go into generate_default_strategy_pool instead?

src/python/bot/fuzzers/afl/launcher.py
```diff
@@ -539,7 +560,17 @@ def __init__(self,
     self._afl_input = None
     self._afl_output = None
 
-    self.strategies = FuzzingStrategies()
+    logs.log(
```
oliverchang: let's make this logging more general and move this to generate_new_testcase_mutations. Also, right now this will log even if we aren't doing a mutations run, which doesn't seem too useful.

src/python/bot/fuzzers/afl/launcher.py
```diff
@@ -1120,6 +1152,40 @@ def libfuzzerize_corpus(self):
...
+    generation_timeout = engine_common.get_new_testcase_mutations_timeout()
+    old_input_directory_file_count = shell.get_directory_file_count(
+        self.afl_input.input_directory)
+
+    # Generate new testcase mutations using Radamsa.
+    if self.strategies.generator == engine_common.Generator.RADAMSA:
+      expected_completion_time = time.time() + generation_timeout
+      engine_common.generate_new_testcase_mutations_using_radamsa(
+          self.afl_input.input_directory, self.afl_input.input_directory,
+          expected_completion_time)
```
oliverchang: since you are here, can you things more consistent by making generate_new_testcase_mutations_using_radamsa take in a timeout rather than an expected completion time?

src/python/bot/fuzzers/afl/launcher.py
```diff
@@ -1120,6 +1152,40 @@ def libfuzzerize_corpus(self):
 
     return new_units_generated, new_units_added, corpus_size
 
+  def generate_new_testcase_mutations(self):
+    """Generate new testcase mutations using radamsa, etc."""
+    target_name = os.path.basename(self.target_path)
+    project_qualified_fuzzer_name = (
```
oliverchang: nit: s/fuzzer/target/

src/python/bot/fuzzers/afl/launcher.py
```diff
@@ -257,9 +258,23 @@ class FuzzingStrategies(object):
...
+      strategy_pool.add_strategy(strategy.CORPUS_MUTATION_RADAMSA_STRATEGY)
+      logs.log('Added radamsa to strategy pool.')
+
+    # Select a generator to use for existing testcase mutations.
+    self.generator = engine_common.select_generator(strategy_pool, target_path)
+    self.is_mutations_run = self.generator != engine_common.Generator.NONE
+
+    # Initialize to false. This is set to true if testcases are properly
+    # generated.
+    self.radamsa_generator = False
```
oliverchang: are these actually getting piped into stats anywhere? I think you need to modify afl.stats.StatsGetter.set_strategy_stats  ?

Also, do we need to have these 2 variables? We could just have
self.generator_strategy = None # engine_common.Generator.VAL

src/python/bot/fuzzers/afl/launcher.py
```diff
@@ -1120,6 +1152,40 @@ def libfuzzerize_corpus(self):
...
+
+    # Generate new testcase mutations using Radamsa.
+    if self.strategies.generator == engine_common.Generator.RADAMSA:
+      expected_completion_time = time.time() + generation_timeout
+      engine_common.generate_new_testcase_mutations_using_radamsa(
+          self.afl_input.input_directory, self.afl_input.input_directory,
+          expected_completion_time)
+
+      # If new mutations are successfully generated, add radamsa stragegy.
+      if shell.get_directory_file_count(
```
oliverchang: Can we remove the duplication of code here with lines 1185 onwards ?

e.g. something like:

```python
if self.strategies.generator = ...
  engine_common.generate...
elif self.strategies.generator = ...
  engine_common.generate...

if shell.get_directory_file_count(
    self.afl_input.input_directory) > old_input_directory_file_count:
  self.generator_strategy = self.generator_candidate

File: src/python/bot/fuzzers/afl/launcher.py [hunk 7]
```diff
@@ -257,9 +258,23 @@ class FuzzingStrategies(object):
...
+  def __init__(self, target_path):
     strategy_pool = strategy_selection.generate_default_strategy_pool(
         strategy_list=strategy.AFL_STRATEGY_LIST, use_generator=False)
+
+    if environment.get_value('USE_RADAMSA'):
+      strategy_pool.add_strategy(strategy.CORPUS_MUTATION_RADAMSA_STRATEGY)
+      logs.log('Added radamsa to strategy pool.')
+
+    # Select a generator to use for existing testcase mutations.
+    self.generator = engine_common.select_generator(strategy_pool, target_path)
```
oliverchang: let's rename this to self.candidate_generator or something similar to avoid confusion.

src/python/bot/fuzzers/afl/launcher.py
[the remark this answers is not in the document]
mukundv-chrome: this was just hardcoded for testing purposes. taking it out

src/python/bot/fuzzers/afl/stats.py
```diff
@@ -233,10 +234,20 @@ def set_strategy_stats(self, fuzzing_strategies):
...
+              fuzzing_strategies.corpus_subset_size)
 
     if fuzzing_strategies.fast_cal == strategies.FastCal.MANUAL:
       self.stats[fuzzing_strategies.FAST_CAL_MANUAL_STRATEGY] = 1
     elif fuzzing_strategies.fast_cal == strategies.FastCal.RANDOM:
       self.stats[fuzzing_strategies.FAST_CAL_RANDOM_STRATEGY] = 1
+
+    if (fuzzing_strategies.generator_strategy == engine_common.Generator.RADAMSA
+       ):
+      self.stats['strategy_{}'.format(
```
oliverchang: nit: the format seems unnecessary here (and everywhere else where we're doing the equivalent of concatenating a single value). just use '+'.

src/python/bot/fuzzers/afl/launcher.py
```diff
@@ -317,14 +326,21 @@ def decide_fast_cal_random(self, num_files):
   def print_strategies(self):
     """Print the strategies used for logging purposes."""
     fuzzing_strategies = []
+
+    if self.generator_strategy == engine_common.Generator.RADAMSA:
+      fuzzing_strategies.append('strategy_{}'.format(
```
oliverchang: nit: the format() call seems unnecessary here. just use '+'.

src/python/bot/fuzzers/afl/launcher.py
```diff
@@ -1120,6 +1139,40 @@ def libfuzzerize_corpus(self):
 
     return new_units_generated, new_units_added, corpus_size
 
+  def generate_new_testcase_mutations(self):
```
oliverchang: any way we could refactor this so that libFuzzer and AFL can share the same code rather than duplicating logic?

mukundv-chrome: Was going to, but they handle it a little differently. AFL adds to the existing input corpus whereas LibFuzzer makes and returns a new one, I thought this would be the cleanest way to deal with that difference. Any ideas on how to handle?

oliverchang: how about a general function which takes in an output and and input dir? When calling from libFuzzer, we create the new dir and pass it to the function as the output. for AFL, we just pass it with the same arguments for input and output?

mukundv-chrome: Sure. I'll get started on that. The other thing was that I thought it might get messy because for libFuzzer we return the new directory, but we don't need to for AFL. Also, the way that they handle keeping track of used strategies is different so I have to work around that to set the generator strategy variables for each of them.

mukundv-chrome: I'll do the refactor and you can let me know how it looks

mukundv-chrome: Done. I think it's better like this.

src/python/bot/fuzzers/engine_common.py
```diff
@@ -62,6 +66,139 @@
...
+  # Generate new testcase mutations using ML RNN model.
+  elif candidate_generator == Generator.ML_RNN:
+    generate_new_testcase_mutations_using_ml_rnn(
+        corpus_directory, new_testcase_mutations_directory, fuzzer_name,
+        generation_timeout)
+
+  # If new mutations are successfully generated, set generator strategy.
+  if shell.get_directory_file_count(
+      new_testcase_mutations_directory) > pre_mutations_filecount:
+    generator_strategy = candidate_generator
```
oliverchang: It seems a little weird for this function to return a Generator value based on if it succeeded or not. How about just return True/False, since the caller knows what it passed?

mukundv-chrome: Sure

mukundv-chrome:
> Looking good! Just one more small comment.
> 
> Please work with @mbarbella-chromium on testing this (staging this on a couple of actual bots), since we don't have great tests for testcase mutations and it may break things.

Yup! That's how I've been testing this week. Radamsa seems to be working properly in local bots but we've had some difficulty locally running the ML RNN models. That part of the code is a lot simpler, though.

PR #763, opened 2019-08-02
Modify project setup to read bucket information from config.

oliverchang:


configs/test/project.yaml
```diff
@@ -96,3 +96,11 @@ env:
   # input to fuzzer as part of --input-dir argument (disabled by default).
   # WEB_TESTS_URL: 'gs://bucket/path-to-tests-archive.zip'
   # UPDATE_WEB_TESTS: true
+
+project_setup:
```
inferno-chromium: Do we want to comment this block since it wont apply to any other user unless they need project setup and they need to create these buckets manually (outside of create_config). So, can add a comment that this is for XYZ usecase where you want automatic project setup and need to create custom buckets ?

inferno-chromium: Also, do we need to put a commented cron entry for this, since this project-setup cron is not listed in test config.

src/appengine/handlers/cron/project_setup.py
```diff
@@ -881,17 +876,19 @@ def get(self):
       logs.log_error('Failed to get AFL Fuzzer entity.')
       return
 
-    # TODO(ochang): Move hardcodes to config.
+    bucket_config = local_config.ProjectConfig().sub_config(
+        'project_setup.build_buckets')
+
```
inferno-chromium: We should probably bail out if this is None. or leave a log_error with bailout

inferno-chromium:
Sorry, left some late comments.

configs/test/project.yaml
oliverchang: Hmm, I'm not sure if this is a use case many others will have, so I don't think documenting this/commented cron is necessary.

I added these here for easier testing, but I can comment this out in another PR.

src/appengine/handlers/cron/project_setup.py
oliverchang: Will address in another PR.

configs/test/project.yaml
inferno-chromium: Nice, in that case, just remove this in a future cl.

PR #765, opened 2019-08-02
Implement more of new engine fuzzing pipeline.

oliverchang:
- Add owner/component metadata.
- Implement reproduce, minimize, cleanse.
- Write tests for fuzz_task.FuzzingSession.do_engine_fuzzing.

### DIFF
--- a/src/python/bot/fuzzers/libFuzzer/engine.py
+++ b/src/python/bot/fuzzers/libFuzzer/engine.py
@@ -258,7 +258,8 @@ class LibFuzzerEngine(engine.Engine):
     # TODO(ochang): Custom crash state.
     # TODO(ochang): Recommended dictionary.
 
-    return engine.Result(fuzz_logs, fuzz_result.command, crashes, parsed_stats)
+    return engine.Result(fuzz_logs, fuzz_result.command, crashes, parsed_stats,
+                         fuzz_result.time_executed)
 
   def reproduce(self, target_path, input_path, arguments, max_time):
     """Reproduce a crash given an input.
@@ -270,9 +271,14 @@ class LibFuzzerEngine(engine.Engine):
       max_time: Maximum allowed time for the reproduction.
 
     Returns:
-      A Result object.
+      A tuple of (return_code, output).
     """
-    raise NotImplementedError
+    runner = libfuzzer.get_runner(target_path)
+    launcher.set_sanitizer_options(target_path)
+
+    result = runner.run_single_testcase(
+        input_path, timeout=max_time, additional_args=arguments)
+    return result.return_code, result.output
 
   def minimize_corpus(self, target_path, arguments, output_dir, input_dirs,
                       max_time):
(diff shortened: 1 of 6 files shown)


inferno-chromium:
@Dor1s - can you please help to review as well.

src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -270,9 +271,14 @@ def reproduce(self, target_path, input_path, arguments, max_time):
       max_time: Maximum allowed time for the reproduction.
 
     Returns:
-      A Result object.
+      A tuple of (return_code, output).
     """
-    raise NotImplementedError
+    runner = libfuzzer.get_runner(target_path)
+    launcher.set_sanitizer_options(target_path)
```
inferno-chromium: it is weird to have set_sanitizer_options in launcher ? should it be in engine_common or something.

src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -337,4 +355,15 @@ def cleanse(self, target_path, arguments, input_path, output_path, max_time):
     Returns:
       A boolean indicating success.
     """
-    raise NotImplementedError
+    runner = libfuzzer.get_runner(target_path)
+    launcher.set_sanitizer_options(target_path)
+
+    cleanse_tmp_dir = self._create_temp_corpus_dir('cleanse-workdir')
+    artifact_prefix = constants.ARTIFACT_PREFIX_FLAG + cleanse_tmp_dir + '/'
```
inferno-chromium: nit: can create a helper for _get_artifact_prefix(output_dir)

src/python/bot/fuzzers/libFuzzer/engine.py
Dor1s: I guess the assumption is that different fuzzing engines may need to overwrite some of the default values? What would happen if e.g. AFL launcher wouldn't call this method? Will the default values be set up by any other code?

src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -270,9 +271,14 @@ def reproduce(self, target_path, input_path, arguments, max_time):
...
     Returns:
-      A Result object.
+      A tuple of (return_code, output).
     """
-    raise NotImplementedError
+    runner = libfuzzer.get_runner(target_path)
+    launcher.set_sanitizer_options(target_path)
+
+    result = runner.run_single_testcase(
+        input_path, timeout=max_time, additional_args=arguments)
```
Dor1s: Should we take care of the arguments in this function? I mean calling `remove_fuzzing_arguments` and adding retries like in https://github.com/google/clusterfuzz/blob/4b3e8bdc430c29768ff631ff02379b821bf4f899/src/python/bot/fuzzers/libFuzzer/launcher.py#L466

src/python/bot/fuzzers/libFuzzer/engine.py
oliverchang: I can move this later in another PR. I think it makes sense to share this for all engines, and they can overwrite things if need be.

src/python/bot/fuzzers/libFuzzer/engine.py
oliverchang: This will be done in fuzz() instead, as we will return a crash object with the filtered args. Coming in another PR :)

src/python/bot/fuzzers/libFuzzer/engine.py
oliverchang: Done!

Issue #761, opened 2019-08-01
Support HELP_TEMPLATE or something similar 


oliverchang:
To be used as a message for displaying reproduction instructions.

With support for substituting in testcase ID, fuzz target names etc.

oliverchang:
@mbarbella-chromium assigning to you as discussed, thanks :)

inferno-chromium:
i can think of
* testcase id
* fuzz target name (without libfuzzer_)
* engine name

PR #772, opened 2019-08-05
Support HELP_TEMPLATE for better reproduction instructions.

mbarbella-chromium:
PTAL. This supports everything mentioned in #761 and it should be easy enough to add more.

### DIFF
--- a/src/appengine/handlers/testcase_detail/show.py
+++ b/src/appengine/handlers/testcase_detail/show.py
@@ -372,6 +372,9 @@ def _format_reproduction_help(reproduction_help):
 
     return '<a href="{url}">{url}</a>'.format(url=plain_text_input)
 
+  if not reproduction_help:
+    return ''
+
   result = ''
   for line in reproduction_help.splitlines():
     # Add a break before each line except the first.
--- a/src/appengine/handlers/testcase_detail/show.py
+++ b/src/appengine/handlers/testcase_detail/show.py
@@ -363,6 +363,7 @@ def _get_blob_size_string(blob_key):
 
 def _format_reproduction_help(reproduction_help):
   """Format a reproduction help string as HTML."""
+
   def _linkify(plain_text_input):
     """Convert links in a plain text string to HTML."""
     if (not plain_text_input.startswith('http://') and
--- a/src/appengine/handlers/testcase_detail/show.py
+++ b/src/appengine/handlers/testcase_detail/show.py
@@ -362,29 +362,11 @@ def _get_blob_size_string(blob_key):
 
 
 def _format_reproduction_help(reproduction_help):
-  """Format a reproduction help string as HTML."""
-
-  def _linkify(plain_text_input):
-    """Convert links in a plain text string to HTML."""
-    if (not plain_text_input.startswith('http://') and
-        not plain_text_input.startswith('https://')):
-      return plain_text_input
-
-    return '<a href="{url}">{url}</a>'.format(url=plain_text_input)
-
+  """Format a reproduction help string as HTML (linkified with break tags)."""
   if not reproduction_help:
     return ''
 
-  result = ''
-  for line in reproduction_help.splitlines():
-    # Add a break before each line except the first.
-    if result:
-      result += '<br>'
-
-    input_parts = [_linkify(part) for part in line.split(' ')]
-    result += ' '.join(input_parts)
-
-  return result
+  return jinja2.utils.urlize(reproduction_help).replace('\n', '<br>')
 
 
 def get_testcase_detail(testcase):
--- a/src/appengine/handlers/testcase_detail/show.py
+++ b/src/appengine/handlers/testcase_detail/show.py
@@ -361,6 +361,28 @@ def _get_blob_size_string(blob_key):
   return utils.get_size_string(blob_size)
 
 
+def _format_reproduction_help(reproduction_help):
+  """Format a reproduction help string as HTML."""
+  def _linkify(plain_text_input):
+    """Convert links in a plain text string to HTML."""
+    if (not plain_text_input.startswith('http://') and
+        not plain_text_input.startswith('https://')):
+      return plain_text_input
+
+    return '<a href="{url}">{url}</a>'.format(url=plain_text_input)
+
+  result = ''
+  for line in reproduction_help.splitlines():
+    # Add a break before each line except the first.
+    if result:
+      result += '<br>'
+
+    input_parts = [_linkify(part) for part in line.split(' ')]
+    result += ' '.join(input_parts)
+
+  return result
+
+
 def get_testcase_detail(testcase):
   """Get testcase detail for rendering the testcase detail page."""
   config = db_config.get()
@@ -368,9 +390,9 @@ def get_testcase_detail(testcase):
   crash_state = testcase.crash_state
   crash_state_lines = crash_state.strip().splitlines()
   crash_type = data_handler.get_crash_type_string(testcase)
-  formatted_reproduction_help = data_handler.get_formatted_reproduction_help(
-      testcase)
-  # TODO(mbarbella): Remove this once everything has been moved to HELP_FORMAT.
+  formatted_reproduction_help = _format_reproduction_help(
+      data_handler.get_formatted_reproduction_help(testcase))
+  # When we have a HELP_TEMPLATE, ignore any default values set for HELP_URL.
   if not formatted_reproduction_help:
     reproduction_help_url = data_handler.get_reproduction_help_url(
         testcase, config)
--- a/src/appengine/handlers/testcase_detail/show.py
+++ b/src/appengine/handlers/testcase_detail/show.py
@@ -86,7 +86,7 @@ def highlight_common_stack_frames(crash_stacktrace):
   stack_trace_line_format = '^ *#([0-9]+) *0x[0-9a-f]+ (.*)'
 
   for line in crash_stacktrace.splitlines():
-    # Stacktrace seperator prefix.
+    # Stacktrace separator prefix.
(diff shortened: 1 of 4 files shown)


src/python/datastore/data_handler.py
```diff
@@ -269,6 +269,26 @@ def get_reproduction_help_url(testcase, config):
...
+  help_format = get_value_from_job_definition(testcase.job_type, 'HELP_FORMAT')
+  if not help_format:
+    return None
+
+  result = help_format.replace('%TESTCASE%', str(testcase.key.id()))
+  result = result.replace('%FUZZER_NAME%', testcase.fuzzer_name)
+
+  if testcase.overridden_fuzzer_name:
+    # Remove the libfuzzer_ or afl_ prefixes from the target name.
+    target_name = testcase.overridden_fuzzer_name.split('_', 1)[-1]
```
oliverchang: this won't give the correct results for a lot of cases. could you please do a datastore read here, by reading:

```python
target = ndb.Key(data_types.FuzzTarget, testcase.overriden_fuzzer_name).get()
```
Then get target.binary_name etc

oliverchang:
Also, please support %PROJECT%, from data_handler.get_project_name.

src/python/datastore/data_handler.py
```diff
@@ -269,6 +269,26 @@ def get_reproduction_help_url(testcase, config):
...
+  else:
```
inferno-chromium: no need of else branch, move result = result.replace('%FUZZ_TARGET%', target_name) inside the if

src/python/tests/core/datastore/data_handler_test.py
```diff
@@ -543,3 +543,38 @@ def test_update_comment_truncate(self):
...
+
+@test_utils.with_cloud_emulators('datastore')
+class GetFormattedReproductionHelpTest(unittest.TestCase):
+  """Test get_formatted_reproduction_help."""
+
+  def setUp(self):
+    job = data_types.Job()
+    job.name = 'my_job'
+    job.environment_string = (
+        'HELP_FORMAT = prefix %TESTCASE% %FUZZER_NAME% %FUZZ_TARGET% suffix')
```
inferno-chromium: maybe \n to seperate fields ?

src/appengine/handlers/testcase_detail/show.py
```diff
@@ -368,8 +368,14 @@ def get_testcase_detail(testcase):
   crash_state = testcase.crash_state
   crash_state_lines = crash_state.strip().splitlines()
   crash_type = data_handler.get_crash_type_string(testcase)
-  reproduction_help_url = data_handler.get_reproduction_help_url(
-      testcase, config)
+  formatted_reproduction_help = data_handler.get_formatted_reproduction_help(
+      testcase)
+  # TODO(mbarbella): Remove this once everything has been moved to HELP_FORMAT.
```
inferno-chromium: can just remove reproduction_help_url completely too and add the field on staging internal, external, also update job types to have both HELP_URL and REPRO_HELP_TEMPLATE

src/appengine/handlers/testcase_detail/show.py
```diff
@@ -368,8 +368,14 @@ def get_testcase_detail(testcase):
   crash_state = testcase.crash_state
   crash_state_lines = crash_state.strip().splitlines()
   crash_type = data_handler.get_crash_type_string(testcase)
-  reproduction_help_url = data_handler.get_reproduction_help_url(
-      testcase, config)
+  formatted_reproduction_help = data_handler.get_formatted_reproduction_help(
```
inferno-chromium: This field needs to be set in configuration.py and also needs to be added in template of configuration page as a multiline text field.

src/appengine/handlers/testcase_detail/show.py
```diff
@@ -86,7 +86,7 @@ def highlight_common_stack_frames(crash_stacktrace):
   stack_trace_line_format = '^ *#([0-9]+) *0x[0-9a-f]+ (.*)'
```
inferno-chromium: This field is not added in data_types.py as a generic configuration attribute, not also updated in go datastore model.

src/python/datastore/data_handler.py
```diff
@@ -269,6 +269,26 @@ def get_reproduction_help_url(testcase, config):
       testcase.job_type, 'HELP_URL', default=config.reproduction_help_url)
 
 
+def get_formatted_reproduction_help(testcase):
+  """Return url to reproduce the bug."""
+  help_format = get_value_from_job_definition(testcase.job_type, 'HELP_FORMAT')
```
inferno-chromium: instead of HELP_FORMAT, we should use HELP_TEMPLATE (need to add some 1-2 line documentation of the feature in jobs documentation).
also, we should default to some repro_help_template config table attribute ? 

also, i think data_handler has some use of HELP_URL, can you fix it to use this.

inferno-chromium:
How will a multiline template be defined in job definition, please try this in staging to see how usable it is. a multiline text field in configuration page works for generic case, but i dont understand how it will work for job definition (also, before rendering in inner-html, newlines need to be changed in br tags, we have a filter for that).
This cl needs more work, also fyi, some oss-fuzz jobs use HELP_URL, so we need to add the new field ?

oliverchang:
Do we need this duplicated field in both configuration and job definition? We can set a default env var in our yaml for the case where we want to set a global default. 

But good point re newlines in job definition. Maybe we can support do a replacement of '\n' with actual newlines for this field.

inferno-chromium:
> Do we need this duplicated field in both configuration.py and job definition? We can set a default env var in our yaml for the case where we want to set a global default.
> 
> But good point re newlines in job definition. Maybe we can support do a replacement of '\n' with actual newlines for this field.

ah right, we can just have this in project.yaml. it can be even in env: section, so that then no change in functionality, gets set as environment variable. lets go with that.

src/appengine/handlers/testcase_detail/show.py
mbarbella-chromium: I don't think the case of having both makes too much sense. +1 to removing reproduction_help_url though. I think I'll just replace it all at once in this change.

src/python/tests/core/datastore/data_handler_test.py
mbarbella-chromium: Done

src/appengine/handlers/testcase_detail/show.py
mbarbella-chromium: Based on our offline discussion, keeping both. Updated the comment to clarify that we prioritize the template.

src/appengine/handlers/testcase_detail/show.py
mbarbella-chromium: Leaving this as-is for now. I think the easy access to help URL is nice since we've decided to keep both. For most use cases this could still work fine.

src/appengine/handlers/testcase_detail/show.py
mbarbella-chromium: Marking resolved since we plan to use the default environment to update these.

src/python/datastore/data_handler.py
mbarbella-chromium: Renamed and fixed other usage.

mbarbella-chromium:
PTAL, I think most of the major comments are addressed.

I think we should remove HELP_URL at some point since this code feels more complicated than it needs to be, but I'd rather handle that separately once we're sure this is in good shape and have everything converted to it.

src/python/tests/core/datastore/data_handler_test.py
```diff
@@ -543,3 +543,48 @@ def test_update_comment_truncate(self):
...
+  def setUp(self):
+    job = data_types.Job()
+    job.name = 'my_job'
+    job.environment_string = (
+        'HELP_FORMAT = -%TESTCASE%\\n-%FUZZER_NAME%\\n-%FUZZ_TARGET%\\n'
+        '-%PROJECT%\\n\n'
+        'PROJECT_NAME = test_project')
+    job.put()
+
+    fuzz_target = data_types.FuzzTarget()
```
oliverchang: i think you need to set id= here to set the key?

mbarbella-chromium: Any examples of this? I don't fully understand the potential issue.

I think this works as is since we set the key in the _pre_put_hook.

oliverchang: oops, I forgot about that. should work as is then!

src/appengine/handlers/testcase_detail/show.py
```diff
@@ -361,15 +361,47 @@ def _get_blob_size_string(blob_key):
...
+  result = ''
+  for line in reproduction_help.splitlines():
+    # Add a break before each line except the first.
+    if result:
+      result += '<br>'
+
+    input_parts = [_linkify(part) for part in line.split(' ')]
+    result += ' '.join(input_parts)
+
+  return result
```
oliverchang: I think django and jinja2 provides the urlize and linebreaksbr filters?

mbarbella-chromium: Since these are getting supplied to our polymer template we don't have easy access to filters. This doesn't feel great here, though. It might be better to implement a similar function in javascript.

oliverchang: Can we call it from python from here: https://github.com/pallets/jinja/blob/master/jinja2/utils.py#L190 ?

mbarbella-chromium: Good call. Done!

mbarbella-chromium:
Referencing #761

PR #776, opened 2019-08-07
issue tracker: Support explicit label mappings for substituted labels.

oliverchang:


oliverchang:
TBR

src/appengine/libs/issue_management/issue_tracker_policy.py
```diff
@@ -93,21 +113,27 @@ def get_new_issue_properties(self, is_security, is_crash):
...
+    labels = issue_type.get('labels')
+    if labels:
+      policy.labels.extend(_to_str_list(labels))
 
     if is_crash:
-      if 'crash_labels' in issue_type:
-        policy.labels.extend(issue_type['crash_labels'])
+      crash_labels = issue_type.get('crash_labels')
+      if crash_labels:
+        policy.labels.extend(_to_str_list(crash_labels))
```
inferno-chromium: is this only an issue when you do labels.extend, how about labels.add ? we could also overwrite the add and extend in LabelStore itself if we are forcing labels to be str ?

oliverchang: Nice suggestion! I'll made add/remove in LabelStore force the arguments to be str, but in this case |labels| isn't a LabelStore, but a regular list, so these are still needed.

PR #786, opened 2019-08-07
Make HELP_URL and HELP_FORMAT work through env in project.yaml

inferno-chromium:
Part of fixing #761


inferno-chromium:
Verified on staging, both HELP_URL and HELP_FORMAT work as expected.

PR #789, opened 2019-08-08
Apply label substitution mappings to issue policy labels.

oliverchang:


src/appengine/libs/issue_management/issue_tracker_policy.py
```diff
@@ -88,6 +83,10 @@ def substitution_mapping(self, label):
 
     return str(mapped)
```
inferno-chromium: is this str needed ? or should other branches also return str(label).

oliverchang:
Actually, this is wrong. Closing.

PR #794, opened 2019-08-08
Make REVISION_VARS_URL work, remove unused config attribute.

### LINKED ISSUE #793: REVISION_VARS_URL is unused.
The only reference is revisions.get_src_map, which is dead code. It's also mentioned in our docs in build_pipeline.md

We should keep REVISION_VARS_URL though, and remove revision_vars_url from data_types.Config. We can set a default value in the project.yaml env section and get the value via data_handler.get_value_from_job_definition_or_environment

inferno-chromium:
Fixes #793.

### DIFF
--- a/src/python/build_management/revisions.py
+++ b/src/python/build_management/revisions.py
@@ -96,9 +96,7 @@ def _clank_revision_file_to_revisions_dict(content):
   clank_revision = component_revision_mappings['clank_revision']
 
   # Initialize revisions dictionary with chromium repo.
-  revisions_dict = get_component_revisions_dict(
-      chromium_revision,
-      'linux_asan_chrome_mp')  # FIXME: Find a way to remove this hardcode.
+  revisions_dict = get_component_revisions_dict(chromium_revision, None)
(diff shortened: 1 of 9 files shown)


src/python/build_management/revisions.py
```diff
@@ -97,7 +96,9 @@ def _clank_revision_file_to_revisions_dict(content):
   clank_revision = component_revision_mappings['clank_revision']
 
   # Initialize revisions dictionary with chromium repo.
-  revisions_dict = get_component_revisions_dict(chromium_revision, 'default')
+  revisions_dict = get_component_revisions_dict(
+      chromium_revision,
+      'linux_asan_chrome_mp')  # FIXME: Find a way to remove this hardcode.
```
oliverchang: do we need this particular value hardcoded at all? since we can just specify this in the local config, and get_value_from_job_definition_or_environment will get the right value. We can even just pass None here, and I think things will still work?

inferno-chromium: Yes None works.

PR #797, opened 2019-08-08
Support SUMMARY_PREFIX_SEPARATOR.

### LINKED ISSUE #792: Support custom fuzz target separators in issue titles
Right now, in our issue titles, we have the following format:

```
prefix/target_name: Crash in ....
```

The separator '/' should be customisable.

oliverchang:
Fixes #792.

### DIFF
--- a/src/python/datastore/data_handler.py
+++ b/src/python/datastore/data_handler.py
@@ -205,20 +205,23 @@ def filter_stacktrace(stacktrace):
 
 def get_issue_summary(testcase):
   """Gets an issue description string for a testcase."""
-  # Get summary prefix. Note that values for fuzzers take priority over those
-  # from job definitions.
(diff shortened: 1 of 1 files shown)


inferno-chromium:
@oliverchang - ideally if we can use a generic seperator instead of / that works for all, i would prefer to change to that, than reading another env var / iterating through job entity (can become expensive).

oliverchang:
> @oliverchang - ideally if we can use a generic seperator instead of / that works for all, i would prefer to change to that, than reading another env var / iterating through job entity (can become expensive).

Yeah, I'm not too thrilled about adding yet another var either. What do you think about using ':' for everything?

inferno-chromium:
Yes colon : sounds good everywhere

PR #800, opened 2019-08-08
Fix exception with clank revision parsing.

inferno-chromium:
None job type does not work since REVISION_VARS_URL is already
overwritten by job environment on bot. So, unable to fetch default
vars url to get chromium revision. Hardcode for now.


inferno-chromium:
TBR

oliverchang:
LGTM

Issue #795, opened 2019-08-08
Support %REVISION% in HELP_FORMAT


oliverchang:
It can be useful to get the latest crashing revision in the reproduce instructions in HELP_FORMAT. 

%REVISION% should map to:

```python
testcase.get_metadata('last_tested_crash_revision') or testcase.crash_revision
```

inferno-chromium:
I will do this since i am already in this function.

PR #804, opened 2019-08-08
Show fuzz target and engine on testcase details, issue filing.

### LINKED ISSUE #781: Show fuzz target + engine on testcase details instead of fully qualified fuzzer name.
This seems especially important when the fully qualified name is long and after normalizing. 

Right now, on the testcase details page, we will show:

```
Fuzzer: libFuzzer_a-b-c-d-e-fuzzer
```
( where a-b-c-d-e is the normalized project name)

Proposed change:
```
Engine: libFuzzer
Target: fuzzer
```

We can still link to the testcase list page with the full libFuzzer_a-b-c-d-e-fuzzer filter on the "fuzzer" text.

inferno-chromium:
Fixes #781, #795

### DIFF
--- a/src/appengine/handlers/testcase_detail/show.py
+++ b/src/appengine/handlers/testcase_detail/show.py
@@ -603,12 +603,7 @@ def get_testcase_detail(testcase):
           reproduction_help_url,
       'is_local_development':
           environment.is_running_on_app_engine_development(),
-      'fuzzer_display': {
-          'engine': fuzzer_display.engine,
-          'target': fuzzer_display.target,
-          'name': fuzzer_display.name,
-          'fully_qualified_name': fuzzer_display.fully_qualified_name,
-      }
+      'fuzzer_display': vars(fuzzer_display),
   }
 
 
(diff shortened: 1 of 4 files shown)


src/python/datastore/data_handler.py
```diff
@@ -74,6 +75,9 @@
     '</b>' % (data_types.FILE_CONSISTENT_UNREPRODUCIBLE_TESTCASE_DEADLINE,
               data_types.UNREPRODUCIBLE_TESTCASE_WITH_BUG_DEADLINE))
 
+FuzzerDisplay = collections.namedtuple(
```
mbarbella-chromium: Display feels a little bit ambiguous and it doesn't look like we're really using it only for display purposes. It might be fine to keep it as-is since I'm having trouble thinking of a better name (something like FuzzerInfo comes to mind but that's still ambiguous, so not really an improvement).

src/appengine/handlers/testcase_detail/show.py
```diff
@@ -601,6 +603,12 @@ def get_testcase_detail(testcase):
           reproduction_help_url,
       'is_local_development':
           environment.is_running_on_app_engine_development(),
+      'fuzzer_display': {
```
oliverchang: maybe add a as_dict() ? so this can be used in more places in the future?

inferno-chromium: Done, let me use vars()

src/python/datastore/data_handler.py
inferno-chromium: Yes cant think of anything better.

Issue #802, opened 2019-08-08
new_edge stat may be incorrect when -max_len is used


Dor1s:
1) If we fuzz with some `-max_len` value, the `initial_edge_coverage` stat will not represent the actual edge coverage of the full corpus, because libFuzzer will ignore the contents past the first `max_len` bytes in every input.

2) During the merge, we reasonably remove `-max_len` argument, in order to avoid truncating the corpus and losing the coverage we have.

3) The resulting `edge_coverage` is extracted from the merge log and might be greater than the `initial_edge_coverage` even if we didn't generate any inputs, just because `-max_len` caused the initial coverage to be incomplete.

4) The resulting `new_edge` metric is calculated as the difference between these two (excluding the corpus subset strategy, as in that case we were able to realize the very same problem and fixed it some time ago):

https://github.com/google/clusterfuzz/blob/649d43cbd729acfd8a6342ad078b23fd191b9869/src/python/bot/fuzzers/libFuzzer/stats.py#L291

I don't know a good solution in mind. We can skip `new_edge` for random max length strategy, but that would mean we lose a signal for the strategy.

My best idea so far is to make libFuzzer's merge to give us all the numbers (i.e. initial coverage + new coverage), but that'll complicate libFuzzer's interface, so I don't really like this either.

Dor1s:
@mbarbella-chromium @mukundv-chrome for the short term, should we exclude random max length strategy? What'll be the best option for your ongoing experiment(s)?

Another bad idea is to do an extra run with `-runs=0` on the full corpus to get the real `initial_edge_coverage`. We can do it only for the runs where random max length is used, but that will be quite expensive and unwise anyway :(

Dor1s:
> @mbarbella-chromium @mukundv-chrome for the short term, should we exclude random max length strategy? What'll be the best option for your ongoing experiment(s)?

Discussed offline. The experimental data should be good for now, so I'll upload a CL to record `new_edges = 0` when random max length strategy is used.

Dor1s:
@inferno-chromium suggested a nice solution on libFuzzer side (to print stats after reading the first corpus dir), I've uploaded a CL as https://reviews.llvm.org/D66020, but we probably need to chose some other keyword. `INITED` is supposed to indicate that ALL inputs files are read, so in our case it's kinda `PARTIALLY INITED` :)

Dor1s:
[...]

PR #805, opened 2019-08-08
libFuzzer stats: skip new_edges and new_features when random max length is used (#802).

Dor1s:


inferno-chromium:
Sad, but LGTM. We need to find some way to get this working, can brainstorm later.

Dor1s:
Merging for now, will deploy in ~1hr to get it in around the UTC midnight.

PR #806, opened 2019-08-08
Fix get_component_revisions_dict for chromium, remove hardcode.

inferno-chromium:


inferno-chromium:
This should fix a lot of exceptions in impact task currently.

PR #812, opened 2019-08-09
Give fork mode more exit time, otherwise runs into SIGTERM.

inferno-chromium:


src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -46,6 +46,9 @@ class LibFuzzerCommon(object):
   # Window of time for libFuzzer to exit gracefully before we KILL it.
   LIBFUZZER_CLEAN_EXIT_TIME = 10.0
 
+  # Additional window of time for libFuzzer fork mode to exit gracefully.
+  LIBFUZZER_FORK_MODE_CLEAN_EXIT_TIME = 50.0
```
Dor1s: Are you sure 50 is enough? Maybe 60-100? I think we should fix it in libFuzzer (if possible) and then get rid of this, so using 100 for extra safety sounds good to me.

Dor1s: for instance, I had a 60 seconds overhead locally

inferno-chromium: Done!

oliverchang:
LGTM

PR #813, opened 2019-08-09
[Reproduce tool] Performance improvements and minor code cleanup.

mbarbella-chromium:
PTAL. This leads to some significant speedups (4GB->108MB for the directory copying, and reduction from 10s->4s for X-related sleeps). Also changes the test case directory to a fixed path in ~/.config/clusterfuzz.

### DIFF
--- a/src/local/butler/reproduce.py
+++ b/src/local/butler/reproduce.py
@@ -186,7 +186,7 @@ def _prepare_initial_environment(build_directory, iterations):
         os.path.join(root_dir, directory_name),
         os.path.join(temp_root_dir, directory_name),
         ignore=lambda directory, contents:
-            contents if directory in ignore_paths else [])
+        contents if directory in ignore_paths else [])
 
   _update_directory('bot')
   _update_directory('configs')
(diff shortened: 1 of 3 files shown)


src/local/butler/reproduce.py
```diff
@@ -100,11 +103,10 @@ def _download_testcase(testcase_id, testcase, configuration):
...
         'Unable to download test case.')
 
-  # Create a temporary directory where we can store the test case.
   bot_absolute_filename = response['x-goog-meta-filename']
-  testcase_directory = os.path.join(
-      environment.get_value('ROOT_DIR'), 'current-testcase')
-  shell.create_directory(testcase_directory)
+  # Store the test case in the config directory for debuggability.
+  testcase_directory = os.path.join(CONFIG_DIRECTORY, 'current-testcase')
+  shell.remove_directory(testcase_directory, recreate=True)
```
inferno-chromium: shell.remove_directory just does os.mkdir, probably need to change to os.makedirs so that intermediates are created.

mbarbella-chromium: Good catch. Changed in shell.remove_directory.

PR #351, opened 2019-04-01
Fix corpus merging on libFuzzer and refactor

jonathanmetzman:
Merge units into an empty directory and then copy new units into the corpus. Do this so that reduced files for features can be added to the corpus, even if the initial version isn't removed until merging.

Refactor:
Rename `create_directory_if_needed` to `create_directory` and add a `recreate` option.
Fix some import orders.
Use `extra_env` instead of copying `os.environ`.

PR #815, opened 2019-08-09
libFuzzer: use two step merge after fuzzing for accurate stats (#802).

Dor1s:
1) Use the two step merge process implemented and documented in https://reviews.llvm.org/D66107

2) Parse edge and feature coverage from the merge log. The first merge step is done using the initial corpus only, i.e. it gives the "initial" stats. The second merge step is executed on the newly generated files and provides the overall stats (coverage data is appended to the "initial" corpus, that's why we need to specify the same `-merge_control_file` for two steps_. The diff between two stats is supposed to be the most accurate data we can obtain.

3) This implementation runs libFuzzer's merge in a way that prioritizes reduced inputs over the longer equivalents from the original corpus. We might be able to get rid of `move_mergeable_units`, I'll look into that later, once this change settles in.

4) libFuzzer on all platforms / projects should be recent enough at this point, therefore I've recompiled the test fuzzers as well to include that new merge functionality.

5) I'll try to stage this on a bot tomorrow as it feels a little scary, even though tests were extremely helpful and caught a few mistakes I made.


Dor1s:
> think whether it's safe to land this until libFuzzer gets rolled in everywhere.

the latest patchset should take care of this, updating the tests is the only blocker then

inferno-chromium:
I think you meant to modify parse_stats_from_merge_log ? since this one is the main fuzzing log.

Dor1s:
> I think you meant to modify parse_stats_from_merge_log ? since this one is the main fuzzing log.

Good catch, apparently we didn't get `new_edges` from the merge log previously, as I had thought. Revamping now

Dor1s:
@inferno-chromium could you please do an initial review? I'll update the tests if we agree on the implementation:

1) feature coverage is still extracted as it used to be. I think it's better than nothing, because merge log doesn't have feature coverage at all.

2) edge coverage is calculated in the merge log only. If merge log looks bad, we don't record any edge coverage.

Dor1s:
> LGTM, but would love if you can move feature_coverage too ? otherwise, we forget about that bug.

Looks like I can do that too! Uploaded: https://reviews.llvm.org/D66030

Dor1s:
//cc @jonathanmetzman @oliverchang 

We had an idea, but it's incomplete and needs more thought:

1) For fuzzing, don't create a new empty directory. Use either the main corpus or the corpus subset as the first corpus dir. That way libFuzzer will remove reduced corpus items (when REDUCE occurs).

2) For merge, don't create a new empty directory either. Merge everything into the main corpus dir. That way we'll get correct stats (edge coverage, feature coverage, new units).

That's where the flow is. We either get `REDUCE` to work correctly (point 1) or we get correct stats from the merge (point 2). Looks like we need some solution for REDUCE problem.

One idea I can think of: if libFuzzer failed to delete the unit after reduce, it should print the name of that input -- that'll be easy to implement. In such case, we will continue fuzzing with a new empty dir. Then we'll delete the original reduced inputs ourselves, and then we will merge into the main corpus dir, but then the stats will be unfair, e.g. the initial edge coverage might be lower if we have deleted some of the inputs :(

3) Don't count the number of corpus files ourselves, and rely on the stats provided by libFuzzer to decide whether merge is needed at all.

Dor1s:
Thinking more about this. Feels like the real solution here would be to make the merge prefer REDUCED inputs. That way we can continue fuzzing with a new empty dir, but during merge we'll be merging back to the main corpus directory. That'll give us the most correct statistics for `new_edges`, `new_features`, `new_units_added`.

oliverchang:
> Thinking more about this. Feels like the real solution here would be to make the merge prefer REDUCED inputs. That way we can continue fuzzing with a new empty dir, but during merge we'll be merging back to the main corpus directory. That'll give us the most correct statistics for new_edges, new_features, new_units_added.

Missing some context here -- is the problem that we're adding a bunch of reduced files, which skews our stats? This solution also sounds best to me in that case, and simplifies how we do things. 

Although now I'm a little confused about why we merging into an empty dir in the first place -- if this was to get reduced inputs, should we just rely on corpus pruning to do this?

Dor1s:
> Missing some context here -- is the problem that we're adding a bunch of reduced files, which skews our stats? This solution also sounds best to me in that case, and simplifies how we do things.

The current stats should be fine, but the way we get them is very complicated, plus we skip e.g. `new_edges` when either corpus subset or random max len strategy is used. We can always get those stats from the merge log now, but we need to merge into the main corpus dir. Currently we merge into an empty dir and the stats printed by merge are pointless (as it assumes the initial coverage was 0 and everything is the new coverage / new files).

> Although now I'm a little confused about why we merging into an empty dir in the first place -- if this was to get reduced inputs, should we just rely on corpus pruning to do this?

Need to check with Jonathan, but yeah looks like libFuzzer does not prefer reduced inputs during merge, and that's why https://github.com/google/clusterfuzz/pull/351 changed our merge process to use an empty dir.

Dor1s:
Uploaded a draft implementation of the merge process that would account for reduced inputs: https://reviews.llvm.org/D66107

Dor1s:
Discussed with @jonathanmetzman offline, he's also supportive of this idea to rely on libFuzzer's output and do less hacky things on CF side.

Dor1s:
Reminder to myself: consider removing some regexes from `src/python/bot/fuzzers/libFuzzer/stats.py` after implementing this new merge logic

Dor1s:
@jonathanmetzman @mbarbella-chromium PTAL

@inferno-chromium @oliverchang more of a FYI, but if you get a change, PTAL too

src/python/bot/fuzzers/engine.py
```diff
@@ -111,15 +111,17 @@ def reproduce(self, target_path, input_path, arguments, max_time):
...
+                      max_time):
     """Optional (but recommended): run corpus minimization.
 
     Args:
       target_path: Path to the target.
       arguments: Additional arguments needed for corpus minimization.
-      input_dirs: Input corpora.
-      output_dir: Output directory to place minimized corpus.
+      existing_corpus_dirs: Input corpora that existed before the fuzzing run.
+      new_corpus_dir: Input corpus that was generated during the fuzzing run.
```
oliverchang: this makes the interface harder to understand and implement, so I think it's the wrong place to change this. 

since this is specific to libFuzzer, we should place this in the implementation for that instead, rather than forcing all engines to adopt this convention.

we could e.g. do this in libFuzzer's engine.py by having our own _minimize_corpus() function there which has this extra argument that the more general interface (minimiz_corpus) piggy backs on.

inferno-chromium: +1. yes this shouldn't be in main engine fuzzer interface. also not in libfuzzer's main minimize_corpus interface, it should be in a helper function _minimize_corpus with extra arg that libfuzzer's main minimize_corpus function can call. i prefer to keep existing name and interfaces, esp the input and output ones.

Dor1s: Thank you, guys! The problem is that libFuzzer's specific `_merge_new_units` function is the last place where we know which directory has the new inputs. From this point on, we have three options how to pass that knowledge down to the actual minimization code (i.e. `minimize_corpus`):

1) Pass it as an explicit argument -- this is the current patchset.
2) Pass it as an optional argument -- similar approach, though seemed a bit more confusing to me.
3) Pass it as a certain element of the `input_dirs` (e.g. the first or the last) -- seemed fragile to me, as it'd be hard to enforce the order besides leaving a comment in the code.

Since `minimize_corpus` is an optional interface yet to be implemented for fuzzing engines other than libFuzzer, it seemed best to me to extend it by this. An additional argument to support this exception would be the prediction that if we ever add minimization for other fuzzing engines, we'll want to capture accurate stats from it as well, which means we will likely need to know which inputs are "new".

Please let me know if either 2) or 3) seems better than 1) to you. I think I like 2) as it appears to be more flexible, but that's still an extension of the interface. Option 3) doesn't change the interface but feels fragile.

@jonathanmetzman @mbarbella-chromium PTAL as well please.

Dor1s: We can discuss it next week, I don't want to bother you while you're out/travelling :)

oliverchang: My suggestion was to have something like the following:

```python
def _minimize_corpus_two_step(self, target_path, arguments, existing_corpus_dirs,
                      new_corpus_dir, output_corpus_dir, reproducers_dir,
                      max_time):
   ...

def minimize_corpus(self): # without the additional arg
  # plain minimize without two step.
```

Then call _minimize_corpus_two_step from _merge_new_units instead.

There will likely be some way to structure the code here such that code is shared as much as possible, maybe with _minimize_corpus_two_step calling minimize_corpus, or by creating another general helper which is used by both.

Would this work?

I imagine we won't need this two step minimization all the time, especially once we do the refactoring to make corpus pruning task use this interface instead, where incremental stats aren't needed.

Note that the intention of `minimize_corpus` was always to be general and used for e.g. corpus pruning -- it just happened that we could reuse it in `fuzz()`. Making this interface specific to fuzzing (and libFuzzer) isn't ideal.

src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -171,8 +172,7 @@ def _merge_new_units(self, target_path, corpus_dir, new_corpus_dir,
     # anyway.
     merge_corpus = self._create_merge_corpus_dir()
 
-    merge_dirs = [new_corpus_dir]
-    merge_dirs.extend(fuzz_corpus_dirs)
+    merge_dirs = copy.copy(fuzz_corpus_dirs)
```
jonathanmetzman: nit: `fuzz_corpus_dirs` is a list right?
`copy.copy` isn't needed to copy it. Just do `merge_dirs = fuzz_corpus_dirs[:]`

src/python/bot/fuzzers/engine.py
Dor1s: > Would this work?

Sounds reasonable, I'll try and see how it looks then. Thanks for the suggestion!

> I imagine we won't need this two step minimization all the time, especially once we do the refactoring to make corpus pruning task use this interface instead, where incremental stats aren't needed.

Even though we don't have incremental stats from the pruning, we still have overall corpus stats -- we should re-use the new stats parsing to get the stats from the final line in the merge log, when we switch to calling `minimize_corpus` instead of `runner.merge`.

I also think we may want incremental stats for the pollination cases (will also need to figure out where to store those stats), but it's outside of the scope of this change. //cc @mbarbella-chromium who might be interested in tracking cross-pollination.

> Note that the intention of `minimize_corpus` was always to be general and used for e.g. corpus pruning -- it just happened that we could reuse it in `fuzz()`.

Yeah, I see... even its name is rather confusing in a context of merging new units.

src/python/bot/fuzzers/libFuzzer/stats.py
```diff
@@ -322,11 +325,17 @@ def parse_performance_features(log_lines,
...
   """Extract stats from a log produced by libFuzzer run with -merge=1."""
-  stats = {}
-  for line in log_lines:
-    match = LIBFUZZER_MERGE_LOG_EDGE_COVERAGE_REGEX.match(line)
+  stats = {
+      'edge_coverage': 0,
+      'feature_coverage': 0,
+  }
+
+  # Revese the list as an optimization. The line of our interest is the last.
```
jonathanmetzman: nit: "Reverse"

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -634,12 +634,27 @@ def merge(self,
     # TODO(flowerhack): Integrate some notion of a merge timeout.
     self._push_corpora_from_host_to_target(corpus_directories)
 
+    target_tmp_dir = None
+    if tmp_dir:
+      # Temp dir may have artifacts, e.g. merge control file for two step merge.
```
jonathanmetzman: nit: "tmp" for consistency

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -634,12 +634,27 @@ def merge(self,
...
+        additional_args[i] = argument.replace(tmp_dir, target_tmp_dir)
+
     # Run merge.
     _, _ = self.fuzzer.merge(
         self._corpus_directories_libfuzzer(corpus_directories) +
         additional_args)
 
     self._pull_new_corpus_from_target_to_host(corpus_directories)
+    if tmp_dir:
+      # Fetch artifacts from the temp dir (e.g. merge control file).
```
jonathanmetzman: nit: "tmp"

src/python/bot/fuzzers/libFuzzer/engine.py
Dor1s: yes, done, thanks!

src/python/bot/fuzzers/libFuzzer/stats.py
Dor1s: done

src/python/bot/fuzzers/libfuzzer.py
Dor1s: done

src/python/bot/fuzzers/libfuzzer.py
Dor1s: done

src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -372,23 +370,62 @@ def minimize_corpus(self, target_path, arguments, input_dirs, output_dir,
...
+      """Execute the fuzz target, check the result, record the stats."""
+      result = runner.merge(
+          corpus_directories,
+          merge_timeout=max_time,
+          tmp_dir=merge_tmp_dir,
+          additional_args=additional_args,
+          artifact_prefix=reproducers_dir)
+
+      if result.timed_out:
+        raise MergeError('Merging new testcases (step %d) timed out' % step)
```
jonathanmetzman: nit: I think we generally end the error messages with a period (though we're not entirely consistent and I don't really think it's that helpful to have). Here and 398.

src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -372,23 +370,62 @@ def minimize_corpus(self, target_path, arguments, input_dirs, output_dir,
...
+        merge_stats['edge_coverage'] - merge_stats['initial_edge_coverage'])
+    merge_stats['new_features'] = (
+        merge_stats['feature_coverage'] -
+        merge_stats['initial_feature_coverage'])
 
     # TODO(ochang): Get crashes found during merge.
-    return engine.FuzzResult(merge_result.output, merge_result.command, [], {},
-                             merge_result.time_executed)
+    return engine.FuzzResult(result_1.output + result_2.output,
+                             result_2.command, [], merge_stats,
```
jonathanmetzman: result_2.command feels a bit wrong here since we really are doing two. But I don't have anything better to suggest.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -634,12 +634,27 @@ def merge(self,
...
     self._push_corpora_from_host_to_target(corpus_directories)
 
+    target_tmp_dir = None
+    if tmp_dir:
+      # Tmp dir may have artifacts, e.g. merge control file for two step merge.
+      target_tmp_dir = self._corpus_target_subdir(os.path.basename(tmp_dir))
+      self.fuzzer.device.rm(target_tmp_dir, recursive=True)
+      self.fuzzer.device.store(tmp_dir, target_tmp_dir)
+      additional_args = copy.copy(additional_args)
+      for i, argument in enumerate(additional_args):
```
jonathanmetzman: It's weird to me that `additional_args=None` in this function when it will exception if it is actually None.
Is this just to support the API?

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -634,12 +634,27 @@ def merge(self,
     # TODO(flowerhack): Integrate some notion of a merge timeout.
     self._push_corpora_from_host_to_target(corpus_directories)
 
+    target_tmp_dir = None
+    if tmp_dir:
+      # Tmp dir may have artifacts, e.g. merge control file for two step merge.
+      target_tmp_dir = self._corpus_target_subdir(os.path.basename(tmp_dir))
+      self.fuzzer.device.rm(target_tmp_dir, recursive=True)
+      self.fuzzer.device.store(tmp_dir, target_tmp_dir)
+      additional_args = copy.copy(additional_args)
```
jonathanmetzman: nit: How about `additional_args = [arg if arg != tmp_dir else target_tmp_dir for arg in additional args]`
Maybe not since the rest of the file copies it (probably a flaw in the API, `additional_args` should probably be safe to mutate).

src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -372,23 +370,62 @@ def minimize_corpus(self, target_path, arguments, input_dirs, output_dir,
...
-
-    if merge_result.return_code != 0:
-      raise MergeError('Merging new testcases failed')
+    merge_control_file = os.path.join(merge_tmp_dir, 'MCF')
+
+    # Make sure the merge control file does not exist at this point.
+    shell.remove_file(merge_control_file)
+
+    additional_args = arguments + [
+        '%s%s' % (constants.MERGE_CONTROL_FILE_ARGUMENT, merge_control_file)
```
oliverchang: one more comment: we should just use '+' in such cases. 

See http://google.github.io/styleguide/pyguide.html#310-strings (first example of "No").

src/python/bot/fuzzers/libFuzzer/stats.py
```diff
@@ -322,11 +325,17 @@ def parse_performance_features(log_lines,
...
-  stats = {}
-  for line in log_lines:
-    match = LIBFUZZER_MERGE_LOG_EDGE_COVERAGE_REGEX.match(line)
+  stats = {
+      'edge_coverage': 0,
+      'feature_coverage': 0,
+  }
+
+  # Reverse the list as an optimization. The line of our interest is the last.
+  for line in log_lines[::-1]:
```
oliverchang: nit: I believe [::-1] actually creates a new list under the hood (O(n)) so this isn't much of an optimisation :)

Use `reversed(log_lines)` which returns an iterator.

src/python/bot/fuzzers/libFuzzer/engine.py
Dor1s: Oh, nice, done!

src/python/bot/fuzzers/libFuzzer/engine.py
Dor1s: +1, but I've reverted to the original messages here, i.e. leaving those without trailing periods

src/python/bot/fuzzers/libFuzzer/engine.py
Dor1s: yeah... the only difference is the last (new) corpus dir, so it seems fine to log the final command only

src/python/bot/fuzzers/libFuzzer/stats.py
Dor1s: I was more worried about applying the regexp to all unnecessary lines, but `reversed` makes a lot of sense here -- thank you, done!

src/python/bot/fuzzers/libfuzzer.py
Dor1s: the catch here is that arguments might contain more that paths, e.g. `-merge-control-file=<path>

src/python/bot/fuzzers/libfuzzer.py
Dor1s: That's weird indeed. All other methods are doing:

```
    additional_args = copy.copy(additional_args)
    if additional_args is None:
      additional_args = []
```

let me do it here for consistency and in a few other places where it's missing

Dor1s:
Travis has succeeded (though says Pending here -- weird) and most importantly the GCB tests succeeded too.

I'm currently staging this on a Chromium bot.

Please take another look :)

src/python/platforms/fuchsia/util/fuzzer.py
```diff
@@ -351,11 +351,15 @@ def merge(self, fuzzer_args):
       fuzzer_args.append('data/corpus/')
       fuzzer_args.append('data/corpus.prev/')
 
-    # Save mergefile in case we are interrupted
-    fuzzer_args = ['-merge=1', '-merge_control_file=data/.mergefile'
-                  ] + fuzzer_args
+    merge_args = ['-merge=1']
```
oliverchang: @flowerhack @eepeep  FYI for changes here. Was this completely broken before? 

@Dor1s can you please make sure that existing tests pass with:

```
FUCHSIA_TESTS=1 python butler.py py_unittest -p '*engine*' 
```

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -634,12 +634,30 @@ def merge(self,
...
+      additional_args = []
+
+    target_tmp_dir = None
+    if tmp_dir:
+      # Tmp dir may have artifacts, e.g. merge control file for two step merge.
+      target_tmp_dir = self._corpus_target_subdir(os.path.basename(tmp_dir))
+      self.fuzzer.device.rm(target_tmp_dir, recursive=True)
+      self.fuzzer.device.store(tmp_dir, target_tmp_dir)
+      for i, argument in enumerate(additional_args):
+        additional_args[i] = argument.replace(tmp_dir, target_tmp_dir)
```
oliverchang: is this to replace the merge_control_file arg? since this is in a libfuzzer specific code, we can just change the interface to add an argument for it to make it easier to deal with in all the subclasses which need to map it into a different path. 

i.e.

```python
def merge(self, ..., merge_control_file=None):
  if merge_control_file:
     # in subclasses map path as necessary.
     additional_arguments.append(constants.MERGE_CONTROL_FILE_ARGUMENT + merge_control_file)
    ...
```

src/python/platforms/fuchsia/util/fuzzer.py
```diff
@@ -351,11 +351,15 @@ def merge(self, fuzzer_args):
...
+
+    # Save mergefile in case we are interrupted, but do not overwrite the
+    # argument if there is already a merge control file specified.
+    if not any('-merge_control_file=' in arg for arg in fuzzer_args):
```
oliverchang: is this needed if we pass it via the libfuzzer.py interface?

src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -371,24 +438,27 @@ def minimize_corpus(self, target_path, arguments, input_dirs, output_dir,
...
 
-    if merge_result.timed_out:
+    if result.timed_out:
       raise MergeError('Merging new testcases timed out')
 
-    if merge_result.return_code != 0:
+    if result.return_code != 0:
       raise MergeError('Merging new testcases failed')
 
+    merge_stats = stats.parse_stats_from_merge_log(result.output.splitlines())
```
Dor1s: One more thing I've noticed while staging: when merge times out, we shouldn't have any incremental stats. I think it's coming from the old code, maybe I should delete it right in this CL -- will think about it.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -634,12 +634,30 @@ def merge(self,
...
+
     # Run merge.
     _, _ = self.fuzzer.merge(
         self._corpus_directories_libfuzzer(corpus_directories) +
         additional_args)
 
     self._pull_new_corpus_from_target_to_host(corpus_directories)
+    if tmp_dir:
```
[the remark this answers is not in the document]
Dor1s: Checking the argument provided to this call. Changed this to work with `merge_control_file` argument directly.

src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -353,6 +349,77 @@ def reproduce(self, target_path, input_path, arguments, max_time):
...
+          found.
+      max_time: Maximum allowed time for the minimization.
+
+    Returns:
+      A Result object.
+    """
+    # Create the merge tmp dir here to make sure it persists for both steps of
+    # the merge being performed below. Otherwise, `minimize_corpus` would
+    # re-create it for every step and delete the merge control file therefore.
+    self.merge_tmp_dir = self._create_temp_corpus_dir('merge-workdir')
```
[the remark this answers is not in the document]
Dor1s: Created `__init__` to manage `self._merge_control_file` attribute. Reverted the `tmp_dir` changes I had, the MCF is now in a separate directory (thanks @oliverchang for the suggestion)

src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -371,24 +438,27 @@ def minimize_corpus(self, target_path, arguments, input_dirs, output_dir,
     """
     runner = libfuzzer.get_runner(target_path)
     libfuzzer.set_sanitizer_options(target_path)
-    merge_tmp_dir = self._create_temp_corpus_dir('merge-workdir')
+    if not self.merge_tmp_dir:
+      self.merge_tmp_dir = self._create_temp_corpus_dir('merge-workdir')
```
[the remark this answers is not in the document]
Dor1s: Using a separate directory for MCF, works great and looks better. Thanks, Oliver!

src/python/bot/fuzzers/libFuzzer/engine.py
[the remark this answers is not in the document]
Dor1s: Actually, I don't think we need `__init__` just for that purpose, removed

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -634,12 +634,30 @@ def merge(self,
     # TODO(flowerhack): Integrate some notion of a merge timeout.
     self._push_corpora_from_host_to_target(corpus_directories)
 
+    additional_args = copy.copy(additional_args)
```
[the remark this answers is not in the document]
Dor1s: https://github.com/google/clusterfuzz/issues/1251

src/python/bot/fuzzers/libFuzzer/engine.py
Dor1s: I've deleted edge and feature coverage stats parsing from the fuzzing log. Now all coverage stats we have will be coming from the merge step. If numbers are zero, that means merge for a fuzz target fails to finish.

Filed https://github.com/google/clusterfuzz/issues/1254 to see how often fuzz targets fail to merge. Maybe there are issues we should fix, maybe things are alright, will see.

src/python/platforms/fuchsia/util/fuzzer.py
Dor1s: Believe it or not, all passed. Thanks for providing me with these commands!

Dor1s:
Works on staging, fingers crossed.

PR #816, opened 2019-08-09
Speculative fix for AFL exception.

inferno-chromium:
Input directory should be available at fuzz time, and not in init.


mukundv-chrome:
LGTM. I don't see any side effects at initial glance. In the init it should choose the candidate generator when FuzzingStrategies() is created, then when we hit the logic that was moved it will actually set the generator that was used by the fuzzing run. As long as stats for strategies are passed after this, we should be good.

PR #822, opened 2019-08-12
[MAB] MAB Production Behavior

mukundv-chrome:


configs/test/bigquery/datasets.yaml
```diff
@@ -209,25 +209,19 @@ resources:
...
           - name: weight
             type: FLOAT
-  - name: fuzz-strategy-experiments-table
+  - name: fuzz-strategy-probability-table
     type: gcp-types/bigquery-v2:tables
     properties:
       datasetId: $(ref.main-dataset.datasetReference.datasetId)
       tableReference:
-        tableId: fuzz_strategy_experiments
+        tableId: fuzz_strategy_probability
```
mbarbella-chromium: Happen to know if this table still exists in the internal project from before? We'll likely need to delete it if so.

src/appengine/handlers/cron/fuzz_strategy_selection.py
```diff
@@ -31,9 +31,7 @@
 from libs import handler
 from metrics import logs
 
-HIGH_TEMPERATURE_PARAMETER = .5
-MEDIUM_TEMPERATURE_PARAMETER = .25
-LOW_TEMPERATURE_PARAMETER = .15
+TEMPERATURE_PARAMETER = .15
```
mbarbella-chromium: Could you leave a comment explaining why we went with this temperature?

configs/test/bigquery/datasets.yaml
mukundv-chrome: It doesn't, I'm pretty sure we deleted it earlier

src/appengine/handlers/cron/fuzz_strategy_selection.py
mukundv-chrome: Sure.

PR #824, opened 2019-08-12
Add %SANITIZER% and %ENGINE% in help format.

inferno-chromium:


src/python/datastore/data_handler.py
```diff
@@ -311,16 +311,24 @@ def get_formatted_reproduction_help(testcase):
...
-
-  result = help_format.replace('%TESTCASE%', str(testcase.key.id()))
-  result = result.replace('%PROJECT%', get_project_name(testcase.job_type))
-  result = result.replace('%REVISION%', str(last_tested_crash_revision))
-  result = result.replace('%FUZZER_NAME%', fuzzer_display.name)
-  result = result.replace('%FUZZ_TARGET%', fuzzer_display.target or '')
+  fuzzer_display = get_fuzzer_display(testcase)
+  fuzzer_name = fuzzer_display.name or ''
+  fuzz_target = fuzzer_display.target or ''
+  engine = (fuzzer_display.engine or '').lower()
```
oliverchang: one small concern here: why should these be lowered by default? I know we need this to make things work, but we shouldn't be doing this here, since it's an implementation detail tied to another system.

inferno-chromium: Yes, lets keep original

Issue #265, opened 2019-03-13
Add support / tests for golang stacktraces


Dor1s:
This is (very likely) needed for https://github.com/google/oss-fuzz/pull/2188

Crash examples are in these two comments: https://github.com/google/oss-fuzz/pull/2188#issuecomment-468002483

Dor1s:
The golang oss-fuzz PR is almost ready (https://github.com/google/oss-fuzz/pull/2188), so this is a blocker now. I won't be able to add new stacktraces this week. If anyone's willing to help, that'll be appreciated.

Dor1s:
Calling this done, the rest of work will be tracked in #860.

PR #827, opened 2019-08-13
Stack Analyzer: add parsing for Golang stacktraces (#265).

Dor1s:
This should unblock https://github.com/google/oss-fuzz/pull/2188.

The security flag distinction can be improved, but at this point it's hard to judge how. Looks like the same panic (i.e. assert) may come from a non-security issue (e.g. a standard package errors out on something) as well as from an assertion added by the fuzzer author intentionally, which may be a legit security issue then.

I propose to mark all of these as security for now (just being cautious) and iterate / improve based on the feedback from the users later on.

Also, stacktraces are pretty inconvenient in a sense that function names and file names are located on different lines. This implementation does not parse file names, it only handles function names.

### DIFF
--- a/src/python/crash_analysis/crash_analyzer.py
+++ b/src/python/crash_analysis/crash_analyzer.py
@@ -91,6 +91,13 @@ UBSAN_CRASH_TYPES_SECURITY = [
     'Non-positive-vla-bound-value',
     'Object-size',
 ]
+GOLANG_CRASH_TYPES_NON_SECURITY = [
+    'index out of range',
+    'integer divide by zero',
+    'makeslice: len out of range',
+    'slice bounds out of range',
+    'stack overflow',
+]
 
 # Default page size of 4KB.
 NULL_DEREFERENCE_BOUNDARY = 0x1000
@@ -296,6 +303,9 @@ def is_security_issue(crash_stacktrace, crash_type, crash_address):
     return True
 
   if crash_type in UBSAN_CRASH_TYPES_NON_SECURITY:
+    return False
+
+  if crash_type in GOLANG_CRASH_TYPES_NON_SECURITY:
     return False
 
   # Floating point exceptions.
--- a/src/python/crash_analysis/crash_analyzer.py
+++ b/src/python/crash_analysis/crash_analyzer.py
@@ -92,11 +92,11 @@ UBSAN_CRASH_TYPES_SECURITY = [
     'Object-size',
 ]
 GOLANG_CRASH_TYPES_NON_SECURITY = [
-    'index out of range',
-    'integer divide by zero',
-    'makeslice: len out of range',
-    'slice bounds out of range',
-    'stack overflow',
+    'Index out of range',
+    'Integer divide by zero',
+    'Makeslice: len out of range',
+    'Slice bounds out of range',
+    'Stack overflow',
 ]
 
 # Default page size of 4KB.
--- a/src/python/crash_analysis/stack_parsing/stack_analyzer.py
+++ b/src/python/crash_analysis/stack_parsing/stack_analyzer.py
@@ -241,6 +241,12 @@ WINDOWS_CDB_CRASH_TYPE_REGEX = re.compile(
 WINDOWS_CDB_STACK_OVERFLOW_REGEX = re.compile(
     r'.*ExceptionCode: .*\(Stack overflow\).*')
 
+# Golang specific stack signatures.
+GOLANG_PANIC_REGEX = re.compile(r'^panic: (.*)')
+GOLANG_PANIC_MAX_LEN = 80
+GOLANG_STACK_FRAME_FUNCTION_REGEX = re.compile(
+    r'^([0-9a-zA-Z\.\-\_\\\/]+)\([0-9a-zA-Z\s,\_]*\)$')
+
 # Mappings of Android kernel error status codes to strings.
 ANDROID_KERNEL_STATUS_TO_STRING = {
     0b0001: 'Alignment Fault',
@@ -1005,6 +1011,21 @@ def llvm_test_one_input_override(frame, frame_struct):
   return frame
 
 
+def _reduce_string(text, length):
+  """Remove the middle part of the text to make it fit the given lenth."""
+  replacement_token = '<...>'
+  if len(text) <= length or len(text) < 2 * len(replacement_token):
+    # Do nothing if the text fits the given length or is too short otherwise.
+    return text
+
+  symbols_to_erase = len(text) + len(replacement_token) - length
+  index_to_erase = length - len(replacement_token)
+  index_to_erase = index_to_erase // 2 - index_to_erase % 1
+  result = text[:index_to_erase] + replacement_token
+  result += text[index_to_erase + symbols_to_erase:]
+  return result
+
+
 def get_crash_data(crash_data, symbolize_flag=True):
   """Get crash parameters from crash data.
   Crash parameters include crash type, address, state and stacktrace.
@@ -1227,6 +1248,13 @@ def get_crash_data(crash_data, symbolize_flag=True):
         new_type='Memcpy-param-overlap',
         reset=True,
         address_from_group=2)
+
+    # Golang stacktraces.
+    golang_panic_match = GOLANG_PANIC_REGEX.match(line)
+    if golang_panic_match:
+      reason = golang_panic_match.group(1)
+      state.crash_type = _reduce_string(reason, GOLANG_PANIC_MAX_LEN)
+      continue
 
     # Sanitizer SEGV crashes.
     segv_match = SAN_SEGV_REGEX.match(line)
@@ -1682,6 +1710,15 @@ def get_crash_data(crash_data, symbolize_flag=True):
         state,
         group=1,
         frame_filter=lambda s: s):
+      continue
+
+    # Golang stack frames.
+    if add_frame_on_match(
+        GOLANG_STACK_FRAME_FUNCTION_REGEX,
+        line,
+        state,
+        group=1,
+        frame_filter=lambda s: s.split('/')[-1]):
       continue
 
   # Detect cycles in stack overflow bugs and update crash state.
--- a/src/python/crash_analysis/stack_parsing/stack_analyzer.py
+++ b/src/python/crash_analysis/stack_parsing/stack_analyzer.py
@@ -426,7 +426,7 @@ STACK_FRAME_IGNORE_REGEXES = [
     r'^\<unknown\>$',
     r'^\[vdso\]$',
 
-    # Golang specific regexes to ignore.
+    # Golang specific frames to ignore.
     r'^runtime.throw$',
     r'^runtime.newstack$',
     r'^runtime.morestack$',
--- a/src/python/crash_analysis/stack_parsing/stack_analyzer.py
+++ b/src/python/crash_analysis/stack_parsing/stack_analyzer.py
@@ -264,7 +264,7 @@ GOLANG_CRASH_TYPES_MAP = [
 ]
 
 GOLANG_STACK_FRAME_FUNCTION_REGEX = re.compile(
-    r'^([0-9a-zA-Z\.\-\_\\\/\(\)\*]+)\([0-9a-zA-Z\s,\_]*\)$')
+    r'^([0-9a-zA-Z\.\-\_\\\/\(\)\*]+)\([x0-9a-f\s,\.]*\)$')
 
 # Mappings of Android kernel error status codes to strings.
 ANDROID_KERNEL_STATUS_TO_STRING = {
--- a/src/python/crash_analysis/stack_parsing/stack_analyzer.py
+++ b/src/python/crash_analysis/stack_parsing/stack_analyzer.py
@@ -241,11 +241,14 @@ WINDOWS_CDB_CRASH_TYPE_REGEX = re.compile(
 WINDOWS_CDB_STACK_OVERFLOW_REGEX = re.compile(
     r'.*ExceptionCode: .*\(Stack overflow\).*')
 
-# Golang specific stack signatures.
-GOLANG_PANIC_REGEX = re.compile(r'^panic: (.*)')
-GOLANG_PANIC_MAX_LEN = 80
+# Golang specific regular expressions.
+GOLANG_PANIC_REGEX = re.compile(
+    r'^(panic: runtime error|fatal error): (index out of range|'
+    r'integer divide by zero|invalid memory address|'
+    r'makeslice: len out of range|slice bounds out of range|stack overflow|'
+    r').*')
 GOLANG_STACK_FRAME_FUNCTION_REGEX = re.compile(
-    r'^([0-9a-zA-Z\.\-\_\\\/]+)\([0-9a-zA-Z\s,\_]*\)$')
+    r'^([0-9a-zA-Z\.\-\_\\\/\(\)\*]+)\([0-9a-zA-Z\s,\_]*\)$')
 
 # Mappings of Android kernel error status codes to strings.
 ANDROID_KERNEL_STATUS_TO_STRING = {
@@ -422,6 +425,11 @@ STACK_FRAME_IGNORE_REGEXES = [
     r'.*Inline\ Function\ \@',
     r'^\<unknown\>$',
     r'^\[vdso\]$',
+
+    # Golang specific regexes to ignore.
+    r'^runtime.throw$',
+    r'^runtime.newstack$',
+    r'^runtime.morestack$',
 ]
 
 STACK_FRAME_IGNORE_REGEXES_IF_SYMBOLIZED = [
@@ -1011,21 +1019,6 @@ def llvm_test_one_input_override(frame, frame_struct):
   return frame
 
 
-def _reduce_string(text, length):
-  """Remove the middle part of the text to make it fit the given lenth."""
-  replacement_token = '<...>'
-  if len(text) <= length or len(text) < 2 * len(replacement_token):
-    # Do nothing if the text fits the given length or is too short otherwise.
-    return text
-
-  symbols_to_erase = len(text) + len(replacement_token) - length
-  index_to_erase = length - len(replacement_token)
-  index_to_erase = index_to_erase // 2 - index_to_erase % 1
-  result = text[:index_to_erase] + replacement_token
-  result += text[index_to_erase + symbols_to_erase:]
-  return result
-
-
 def get_crash_data(crash_data, symbolize_flag=True):
   """Get crash parameters from crash data.
   Crash parameters include crash type, address, state and stacktrace.
@@ -1252,8 +1245,7 @@ def get_crash_data(crash_data, symbolize_flag=True):
     # Golang stacktraces.
     golang_panic_match = GOLANG_PANIC_REGEX.match(line)
     if golang_panic_match:
-      reason = golang_panic_match.group(1)
-      state.crash_type = _reduce_string(reason, GOLANG_PANIC_MAX_LEN)
+      state.crash_type = golang_panic_match.group(2)
       continue
 
     # Sanitizer SEGV crashes.
--- a/src/python/crash_analysis/stack_parsing/stack_analyzer.py
+++ b/src/python/crash_analysis/stack_parsing/stack_analyzer.py
@@ -242,11 +242,27 @@ WINDOWS_CDB_STACK_OVERFLOW_REGEX = re.compile(
     r'.*ExceptionCode: .*\(Stack overflow\).*')
 
 # Golang specific regular expressions.
-GOLANG_PANIC_REGEX = re.compile(
-    r'^(panic: runtime error|fatal error): (index out of range|'
-    r'integer divide by zero|invalid memory address|'
-    r'makeslice: len out of range|slice bounds out of range|stack overflow|'
-    r').*')
+GOLANG_DIVISION_BY_ZERO_REGEX = re.compile(
+    r'^panic: runtime error: integer divide by zero.*')
+GOLANG_INDEX_OUT_OF_RANGE_REGEX = re.compile(
+    r'^panic: runtime error: index out of range.*')
+GOLANG_INVALID_MEMORY_ADDRESS_REGEX = re.compile(
+    r'^panic: runtime error: invalid memory address.*')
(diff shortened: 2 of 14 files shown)


src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -241,6 +241,12 @@
 WINDOWS_CDB_STACK_OVERFLOW_REGEX = re.compile(
     r'.*ExceptionCode: .*\(Stack overflow\).*')
 
+# Golang specific stack signatures.
+GOLANG_PANIC_REGEX = re.compile(r'^panic: (.*)')
```
jonathanmetzman: I suppose there isn't much we can do, but looking for the word "panic" might have some False positives.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -1005,6 +1011,21 @@ def llvm_test_one_input_override(frame, frame_struct):
   return frame
 
 
+def _reduce_string(text, length):
+  """Remove the middle part of the text to make it fit the given lenth."""
+  replacement_token = '<...>'
+  if len(text) <= length or len(text) < 2 * len(replacement_token):
```
jonathanmetzman: Is the second condition handling cases where "length" is really small?
Maybe we could just assert that it is greater than `2 * len(replacement_token)`

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -1005,6 +1011,21 @@ def llvm_test_one_input_override(frame, frame_struct):
   return frame
 
 
+def _reduce_string(text, length):
+  """Remove the middle part of the text to make it fit the given lenth."""
```
jonathanmetzman: nit: "|text|" and "|length|"

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -1005,6 +1011,21 @@ def llvm_test_one_input_override(frame, frame_struct):
   return frame
 
 
+def _reduce_string(text, length):
```
jonathanmetzman: Maybe we could just do [this](https://github.com/google/clusterfuzz/blob/c2962b66096574bb514cb4646906d45904538b5e/src/python/crash_analysis/stack_parsing/stack_analyzer.py#L593) instead of this function for simplicity.

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_data/golang_sigsegv_panic.txt
```diff
@@ -0,0 +1,10 @@
+root@943ca8071e8b:/out# ./fuzzer-bzip2 
+panic: runtime error: invalid memory address or nil pointer dereference
+[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x5da137]
+
+goroutine 1 [running]:
+math.glob..func1(0x5da117)
+  /src/go/src/math/exp_asm.go:11 +0x7
+math.init.ializers()
```
jonathanmetzman: hmm... is this function "math.init.ializers()" weird on purpose?

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -241,6 +241,12 @@
 WINDOWS_CDB_STACK_OVERFLOW_REGEX = re.compile(
     r'.*ExceptionCode: .*\(Stack overflow\).*')
 
+# Golang specific stack signatures.
+GOLANG_PANIC_REGEX = re.compile(r'^panic: (.*)')
+GOLANG_PANIC_MAX_LEN = 80
```
jonathanmetzman: nit: does it make sense `LINE_LENGTH_CAP` instead of this?

jonathanmetzman:
> 
> I propose to mark all of these as security for now (just being cautious) and iterate / improve based on the feedback from the users later on.

I don't think this is a wrong answer, I'm happy with it for now.

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -2435,3 +2435,50 @@ def test_log_fatal_google(self):
...
+    expected_stacktrace = data
+    expected_security_flag = True
+    self._validate_get_crash_data(data, expected_type, expected_address,
+                                  expected_state, expected_stacktrace,
+                                  expected_security_flag)
+
+  def test_golang_libfuzzer_panic(self):
+    """Test golang stacktrace with panic and libFuzzer's deadly signal."""
+    data = self._read_test_data('golang_libfuzzer_panic.txt')
+    expected_type = ('parse //%B9%B9%B9%B9%B9%01%00%00%00%0<...>%B9%B9%B9%B9: '
```
inferno-chromium: This is not a type, type should be Panic, with security flag false, and this needs to be in state as-is.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
Dor1s: sounds better, but removed this anyway!

src/python/crash_analysis/stack_parsing/stack_analyzer.py
Dor1s: yeah, this is an overkill, plus we won't put long custom string into the crash type, removing this

src/python/crash_analysis/stack_parsing/stack_analyzer.py
Dor1s: remove

src/python/crash_analysis/stack_parsing/stack_analyzer.py
Dor1s: yes, and removed :)

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -1228,6 +1249,13 @@ def get_crash_data(crash_data, symbolize_flag=True):
         reset=True,
         address_from_group=2)
 
+    # Golang stacktraces.
+    golang_panic_match = GOLANG_PANIC_REGEX.match(line)
+    if golang_panic_match:
+      reason = golang_panic_match.group(1)
+      state.crash_type = _reduce_string(reason, GOLANG_PANIC_MAX_LEN)
```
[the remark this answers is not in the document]
Dor1s: Ack, removed this

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_data/golang_sigsegv_panic.txt
Dor1s: I think so :D from https://golang.org/doc/go1.12#runtime:

<...> Doing a traceback during the initialization of a global variable will now show a function named `PKG.init.ializers`.

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -2435,3 +2435,50 @@ def test_log_fatal_google(self):
     self._validate_get_crash_data(data, expected_type, expected_address,
                                   expected_state, expected_stacktrace,
                                   expected_security_flag)
+
+  def test_golang_asan_panic(self):
+    """Test golang stacktrace with panic and ASan."""
+    data = self._read_test_data('golang_asan_panic.txt')
+    expected_type = 'asn1: string not valid UTF-8'
```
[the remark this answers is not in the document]
Dor1s: Going with `ASSERT` by default, but adding a bunch of error that seem to be the most popular ones (like `index out of range`).

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -2435,3 +2435,50 @@ def test_log_fatal_google(self):
...
+    expected_security_flag = True
+    self._validate_get_crash_data(data, expected_type, expected_address,
+                                  expected_state, expected_stacktrace,
+                                  expected_security_flag)
+
+  def test_golang_sigsegv_panic(self):
+    """Test golang stacktrace with panic and SIGSEGV."""
+    data = self._read_test_data('golang_sigsegv_panic.txt')
+    expected_type = (
+        'runtime error: invalid memory address or nil pointer dereference')
```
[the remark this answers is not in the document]
Dor1s: Done

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
Dor1s: Done

src/python/crash_analysis/crash_analyzer.py
```diff
@@ -91,6 +91,13 @@
     'Non-positive-vla-bound-value',
     'Object-size',
 ]
+GOLANG_CRASH_TYPES_NON_SECURITY = [
+    'index out of range',
```
inferno-chromium: Our style is capitalize with hyphen seperation, also we try to use existing types if possible.
Index-out-of-range (matches ubsan ?)
Divide-by-zero (matches ubsan ?)
Makeslice-length-out-of-range
Slice-bounds-out-of-range
Stack-overflow (matches regular stuff)

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -2435,3 +2435,174 @@ def test_log_fatal_google(self):
...
                                   expected_state, expected_stacktrace,
                                   expected_security_flag)
+
+  def test_golang_asan_panic(self):
+    """Test golang stacktrace with panic and ASan."""
+    data = self._read_test_data('golang_asan_panic.txt')
+    expected_type = 'ASSERT'
+    expected_address = ''
+    expected_state = (
+        'asn1: string not valid UTF-8\nasn1.Fuzz\nruntime.raise\n')
```
inferno-chromium: need to add runtime.raise in ignore list, so that it does not appear here.

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -2435,3 +2435,174 @@ def test_log_fatal_google(self):
...
+                                  expected_state, expected_stacktrace,
+                                  expected_security_flag)
+
+  def test_golang_sigsegv_panic(self):
+    """Test golang stacktrace with panic and SIGSEGV."""
+    data = self._read_test_data('golang_sigsegv_panic.txt')
+    expected_type = 'invalid memory address'
+    expected_address = ''
+    expected_state = (
+        'runtime error: invalid memory address or nil pointer dereference\n'
```
inferno-chromium: We should not add this frame at all, this info is already incorporated in crash type and this is bad for deduplication.

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -2435,3 +2435,174 @@ def test_log_fatal_google(self):
...
+                                  expected_security_flag)
+
+  def test_golang_panic_runtime_error_invalid_memory_address(self):
+    """Test golang stacktrace with panic caused by invalid memory address."""
+    data = self._read_test_data(
+        'golang_panic_runtime_error_invalid_memory_address.txt')
+    expected_type = 'invalid memory address'
+    expected_address = ''
+    expected_state = (
+        'runtime error: invalid memory address or nil pointer dereference\n'
```
inferno-chromium: yes this frame shouldnt be in crash state, see comment above.

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -2435,3 +2435,174 @@ def test_log_fatal_google(self):
...
+                                  expected_state, expected_stacktrace,
+                                  expected_security_flag)
+
+  def test_golang_panic_runtime_error_index_out_of_range(self):
+    """Test golang stacktrace with panic caused by index out of range."""
+    data = self._read_test_data(
+        'golang_panic_runtime_error_index_out_of_range.txt')
+    expected_type = 'index out of range'
+    expected_address = ''
+    expected_state = ('runtime error: index out of range\n'
```
inferno-chromium: yes this frame shouldnt be in crash state, see comment above.

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -2435,3 +2435,174 @@ def test_log_fatal_google(self):
...
+                                  expected_state, expected_stacktrace,
+                                  expected_security_flag)
+
+  def test_golang_panic_runtime_error_integer_divide_by_zero(self):
+    """Test golang stacktrace with panic caused by integer divide by zero."""
+    data = self._read_test_data(
+        'golang_panic_runtime_error_integer_divide_by_zero.txt')
+    expected_type = 'integer divide by zero'
+    expected_address = ''
+    expected_state = ('runtime error: integer divide by zero\n'
```
inferno-chromium: yes this frame shouldnt be in crash state, see comment above.

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -2435,3 +2435,174 @@ def test_log_fatal_google(self):
...
+
+  def test_golang_panic_runtime_error_makeslice_len_out_of_range(self):
+    """Test golang stacktrace with panic caused by makeslice len out of range.
+    """
+    data = self._read_test_data(
+        'golang_panic_runtime_error_makeslice_len_out_of_range.txt')
+    expected_type = 'makeslice: len out of range'
+    expected_address = ''
+    expected_state = (
+        'runtime error: makeslice: len out of range\npanic\ngc.newliveness\n')
```
inferno-chromium: yes this runetime error frame shouldnt be in crash state, see comment above.

src/python/crash_analysis/crash_analyzer.py
inferno-chromium: Use like UBSAN_CRASH_TYPES_MAP in stack_analyzer to map these.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -241,6 +241,15 @@
...
     r'.*ExceptionCode: .*\(Stack overflow\).*')
 
+# Golang specific regular expressions.
+GOLANG_PANIC_REGEX = re.compile(
+    r'^(panic: runtime error|fatal error): (index out of range|'
+    r'integer divide by zero|invalid memory address|'
+    r'makeslice: len out of range|slice bounds out of range|stack overflow|'
+    r').*')
+GOLANG_STACK_FRAME_FUNCTION_REGEX = re.compile(
+    r'^([0-9a-zA-Z\.\-\_\\\/\(\)\*]+)\([0-9a-zA-Z\s,\_]*\)$')
```
inferno-chromium: This regex can impact other non-go things and can add weird things to their crash state. we need to think if this regex can be made more specific in style and other thing we can do to limit damage is

```
    if golang_panic_match:
      state.crash_type = golang_panic_match.group(2)
      is_go_crash = True
      continue

and then in lower one, do like
if is_go_crash and add_frame_on_match

File: src/python/crash_analysis/crash_analyzer.py (hunk 11 above)
Comment (Dor1s) [reply src/python/crash_analysis/crash_analyzer.py]: I think the existing naming style was derived from the sanitizer error messages, while Golang seems to have its own style and a different set of errors? Is there any value in trying to map Golang errors into sanitizer errors?

I kinda like that these look differently plus the code is less complicated this way (i.e. we don't need the string-to-string mapping).

File: src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py (hunk 12 above)
Comment (Dor1s) [reply src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py]: Ah, this actually comes from the libFuzzer stacktrace, we may want to do this instead: https://github.com/google/oss-fuzz/pull/2188#issuecomment-521815547

But I'm fine with suppressing in too, will add.

File: src/python/crash_analysis/crash_analyzer.py (hunk 11 above)
Comment (inferno-chromium) [reply src/python/crash_analysis/crash_analyzer.py]: So this was about reducing parity of crash types too.

like 'stack overflow' and 'Stack-overflow' become two different crash types. Also, makes it look weird in UI where go and c/c++ are listed in same place. This is outside of sanitizers too, where we try to capitalize (always) and converting from spacing to hyphen is recommended.
If you dont like map, we can do processing through a generic function, just capitalize and convert spacing to hyphen in a helper. On the other hand, another advantage of list is you dont have to hardcode it in the regex, you can do like '|.join(whatever_map.keys()). it is easier to add new stuff.

File: src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py (hunk 12 above)
Comment (inferno-chromium) [reply src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py]: yes need to suppress for better deduplication.

File: src/python/crash_analysis/crash_analyzer.py (hunk 11 above)
Comment (Dor1s) [reply src/python/crash_analysis/crash_analyzer.py]: Two more points to clarify before addressing this:

> makes it look weird in UI where go and c/c++ are listed in same place

1. That's exactly what bothers me: a Golang developer might find it more convenient to receive reports with the error types they used to observe in their day-to-day work, rather than adopting to some other convention?

   Also, I doubt we'll have both C/C++ and Golang crashes in the same OSS-Fuzz project, so technically these will be listed at the same place only for the admins who see everything, not for the regular users.

2. If you still insist on unifying the crash types, what do you think about separating stack and crash analysis for different languages?

   I'm concerned that a naive mapping would not scale well once we bring in e.g. Rust or another language. Even with this change alone, the security flag distinction would need to be changed, as `Index-out-of-bounds` would not be a UBSan crash type anymore (assuming we map Golang's `index out of range` into it), and in case of UBSan stack trace it'd have security flag `True`, wihle for Golang it'd be `False` (as it's a standard runtime error, like a release assert).

   In such case, I'm thinking about an entry level dispatcher, that would check for some tokens and then call the corresponding `golang_stack_analyzer`, `cpp_stack_analyzer`, etc. But will need to think a bit more, if we decide to go this way. //cc @oliverchang as well

Comment (oliverchang) [reply src/python/crash_analysis/crash_analyzer.py]: Also need to think more about it, but my hunch is we should unify crash types to make things look more consistent in our UI.

Re different semantics for different languages, I believe the language considerations should go into crash_analyzer, rather than relying on slightly different strings to imply this.

Comment (Dor1s) [reply src/python/crash_analysis/crash_analyzer.py]: Ok, thanks! I filed https://github.com/google/clusterfuzz/issues/850 to brainstorm this more, and for now I'll try to address your comments and get this landed today, as that OSS-Fuzz PR is waiting for too long already :)

Comment (Dor1s) [reply src/python/crash_analysis/crash_analyzer.py]: Chatted with Abhishek, decided to go with the Go specific crashes types for now due to severity analysis complications (but capitalize the first letter), and also add a `is_golang` flag to prevent messing up the existing testcases.

File: src/python/crash_analysis/stack_parsing/stack_analyzer.py (hunk 18 above)
Comment (Dor1s) [reply src/python/crash_analysis/stack_parsing/stack_analyzer.py]: Good call! The panic signature is pretty generic and might be similar to non-Go, but I'll check the presence of the go sources in the stacktrace -- should be safe.

File: src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py (hunk 12 above)
Comment (Dor1s) [reply src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py]: Done

File: src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py (hunk 13 above)
Comment (Dor1s) [reply src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py]: Done

File: src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py (hunk 14 above)
Comment (Dor1s) [reply src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py]: Done

File: src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py (hunk 15 above)
Comment (Dor1s) [reply src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py]: Done

File: src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py [hunk 19]
```diff
@@ -2435,3 +2435,174 @@ def test_log_fatal_google(self):
...
+
+  def test_golang_panic_runtime_error_index_out_of_range(self):
+    """Test golang stacktrace with panic caused by index out of range."""
+    data = self._read_test_data(
+        'golang_panic_runtime_error_index_out_of_range.txt')
+    expected_type = 'index out of range'
+    expected_address = ''
+    expected_state = ('runtime error: index out of range\n'
+                      'http.(*conn).serve.func1\n'
+                      'panic\n')
```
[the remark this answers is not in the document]
Dor1s: Done

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
Dor1s: Done

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
Dor1s: Done

src/python/crash_analysis/stack_parsing/stack_analyzer.py
Dor1s: I think I can improve the regex itself a bit more -- the arguments are either empty, or hex numbers, or `...`. Updating now.

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_data/golang_sigsegv_panic.txt
jonathanmetzman: ACK

Dor1s:
Travis having some networking issues, but I've updated one testcase, so re-running both travis and gcbrun.

PR #829, opened 2019-08-13
Make dropdowns for project, fuzzer, job searchable.

inferno-chromium:
This adds a dedicated search box and allows searching for
sub-strings.

### DIFF
--- a/src/appengine/private/components/crash-stats/search-control-panel.html
+++ b/src/appengine/private/components/crash-stats/search-control-panel.html
@@ -546,6 +546,9 @@
 
         // Update the search params in url.
         this.submit();
+
+        // Don't restore focus to the keyword box.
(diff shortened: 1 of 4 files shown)


inferno-chromium:
> Looks good to me overall.
> 
> Is there a way to set a minimum width on the dropdown? Or maybe force the selection to be cleared if the user types something without a match? I noticed that if you type something with no match (consider a typo like tesst in one of the search boxes), then re-select the dropdown and correct it without fully clearing (backspace x2 in the previous example), you'll be left with a very narrow box that doesn't show the results well.

Cannot fix this yet, will do in another cl.

inferno-chromium:
@oliverchang fyi!

oliverchang:
awesome!!!

PR #834, opened 2019-08-14
Add crash revision when regression range is not available.

inferno-chromium:


inferno-chromium:
unsure if we should add crash revision everytime. if you feel that, i can add that always.

src/appengine/handlers/revisions_info.py
```diff
@@ -25,13 +25,25 @@ class Handler(base_handler.Handler):
...
+    revision = self.request.get('revision')
+    if revision:
+      if not revision.isdigit():
+        raise helpers.EarlyExitException('Revision is not an integer.', 400)
+      start_revision = end_revision = revision
+    else:
+      revision_range = self.request.get('range')
+      if revision_range:
+        try:
+          [start_revision, end_revision] = revision_range.split(':')
```
oliverchang: nit: [] should be unnecessary here.

PR #835, opened 2019-08-14
More fuzzing engine refactoring (#483).

oliverchang:
- Add fuzzer_binary_name to fuzzer metadata.
- Add custom crash state, dictionary analysis.
- Move strategy stats setting to more general location.
- Start integration tests for libFuzzer.

### DIFF
--- a/src/python/bot/fuzzers/engine.py
+++ b/src/python/bot/fuzzers/engine.py
@@ -107,7 +107,7 @@ class Engine(object):
     raise NotImplementedError
 
   def minimize_corpus(self, target_path, arguments, output_dir, input_dirs,
-                      max_time):
+                      reproducers_dir, max_time):
     """Optional (but recommended): run corpus minimization.
 
     Args:
@@ -115,6 +115,8 @@ class Engine(object):
       arguments: Additional arguments needed for corpus minimization.
       output_dir: Output directory to place minimized corpus.
       input_dirs: Input corpora.
+      reproducers_dir: The directory to put reproducers in when crashes are
+          found.
       max_time: Maximum allowed time for the minimization.
 
     Returns:
--- a/src/python/bot/fuzzers/engine.py
+++ b/src/python/bot/fuzzers/engine.py
@@ -106,15 +106,15 @@ class Engine(object):
     """
     raise NotImplementedError
 
-  def minimize_corpus(self, target_path, arguments, output_dir, input_dirs,
+  def minimize_corpus(self, target_path, arguments, input_dirs, output_dir,
                       reproducers_dir, max_time):
     """Optional (but recommended): run corpus minimization.
 
     Args:
       target_path: Path to the target.
       arguments: Additional arguments needed for corpus minimization.
-      output_dir: Output directory to place minimized corpus.
       input_dirs: Input corpora.
+      output_dir: Output directory to place minimized corpus.
       reproducers_dir: The directory to put reproducers in when crashes are
           found.
       max_time: Maximum allowed time for the minimization.
--- a/src/python/bot/fuzzers/libFuzzer/engine.py
+++ b/src/python/bot/fuzzers/libFuzzer/engine.py
@@ -163,8 +163,8 @@ class LibFuzzerEngine(engine.Engine):
       result = self.minimize_corpus(
           target_path=target_path,
           arguments=arguments,
-          output_dir=merge_corpus,
           input_dirs=merge_dirs,
+          output_dir=merge_corpus,
           reproducers_dir=None,
           max_time=engine_common.get_merge_timeout(
               launcher.DEFAULT_MERGE_TIMEOUT))
@@ -181,13 +181,13 @@ class LibFuzzerEngine(engine.Engine):
 
     stat_overrides['new_units_added'] = new_units_added
 
+    logs.log('Stats calculated', stats=stat_overrides)
+
     # Record the stats to make them easily searchable in stackdriver.
     if new_units_added:
-      logs.log(
-          'New units added to corpus: %d.' % new_units_added,
-          stats=stat_overrides)
+      logs.log('New units added to corpus: %d.' % new_units_added)
     else:
-      logs.log('No new units found.', stats=stat_overrides)
+      logs.log('No new units found.')
 
   def fuzz(self, target_path, options, reproducers_dir, max_time):
     """Run a fuzz session.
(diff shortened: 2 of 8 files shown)


src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -178,6 +181,14 @@ def _merge_new_units(self, target_path, corpus_dir, new_corpus_dir,
 
     stat_overrides['new_units_added'] = new_units_added
 
+    # Record the stats to make them easily searchable in stackdriver.
+    if new_units_added:
+      logs.log(
+          'New units added to corpus: %d.' % new_units_added,
+          stats=stat_overrides)
```
jonathanmetzman: Not that important, but are new_units_added one of the `stat_overrides` kind of redundant. Also unclear why stats should be reported in this log and not elsewhere (I guess we need to do it somewhere arbitrary).

Dor1s: I think this logging was added for quick debugging, as we have the stats in the BigQuery otherwise.

src/python/bot/fuzzers/libFuzzer/stats.py
```diff
@@ -153,7 +153,10 @@ def parse_line_for_strategy_prefix(line, strategy_name):
   return stats
 
 
-def parse_performance_features(log_lines, strategies, arguments):
+def parse_performance_features(log_lines,
+                               strategies,
+                               arguments,
+                               include_strategies=True):
```
Dor1s: Can't understand why we need `include_strategies` switch. Can we just rely on the `strategies` being (or not being) `None`/`[]`?

Dor1s: Ah, I guess it's a temporary solution to support both old launcher and the new code. Maybe add another TODO to remove the strategies from this function completely? Should be doable once we migrate to the new interface + https://github.com/google/clusterfuzz/issues/802 gets resolved.

src/python/tests/core/bot/fuzzers/libFuzzer/engine_test.py
```diff
@@ -203,54 +211,217 @@ def mock_merge(*args, **kwargs):  # pylint: disable=unused-argument
...
-            'strategy_mutator_plugin': 0,
-            'strategy_random_max_len': 0,
-            'strategy_recommended_dict': 0,
-            'strategy_selection_method': 'default',
-            'strategy_value_profile': 0,
-            'timeout_count': 0,
-            'timeout_limit': 123,
-        },
-        result.stats)
+    self.assertDictEqual({
```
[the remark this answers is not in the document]
oliverchang: Why so?

src/python/bot/fuzzers/libFuzzer/stats.py
oliverchang: Yes, this is needed to support both at the moment. will add a TODO.

src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -178,6 +181,14 @@ def _merge_new_units(self, target_path, corpus_dir, new_corpus_dir,
...
+    else:
+      logs.log('No new units found.', stats=stat_overrides)
+
   def fuzz(self, target_path, options, reproducers_dir, max_time):
```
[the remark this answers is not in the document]
oliverchang: I don't think it should. We should minimise the statefulness of this class.

src/python/bot/fuzzers/engine.py
```diff
@@ -107,14 +107,16 @@ def reproduce(self, target_path, input_path, arguments, max_time):
     raise NotImplementedError
 
   def minimize_corpus(self, target_path, arguments, output_dir, input_dirs,
-                      max_time):
+                      reproducers_dir, max_time):
```
inferno-chromium: reproducers_dir feels like confusing name, how about just crashes_dir
also, can we switch order of args
like input_dirs, output_dir, crashes_dir ? feels like input should come first ? or output_dir, crashes_dir, input_dirs ?

src/python/bot/fuzzers/libFuzzer/engine.py
inferno-chromium: yes i think stats logging can be moved to its own logs.log call
logs.log('Calculated stats.', stats=stat_overrides).

src/python/bot/fuzzers/libFuzzer/stats.py
```diff
@@ -288,8 +294,7 @@ def parse_performance_features(log_lines, strategies, arguments):
   # new_edges and new_features may not be correct when either corpus subset or
   # random max length strategy is used. Skip recording these stats in such case.
   # See https://github.com/google/clusterfuzz/issues/802.
-  if (not stats['strategy_corpus_subset'] and
-      not stats['strategy_random_max_len']):
+  if ('corpus_subset' not in strategies and 'random_max_len' not in strategies):
```
inferno-chromium: aren't these always initialized to 0 ? did something change there ?

src/python/bot/fuzzers/libFuzzer/engine.py
oliverchang: Done!

src/python/bot/fuzzers/engine.py
oliverchang: Hmm, "crashes" seems ambiguous to me -- are we referring to crash stacktrace? or the input that causes it? Leaving as is for now, but can rename if you feel strongly. 

Reordering sounds good -- done!

src/python/bot/fuzzers/libFuzzer/stats.py
oliverchang: I think that's no longer necessary now that we're no longer using autodetection.

src/python/tests/core/bot/fuzzers/libFuzzer/engine_test.py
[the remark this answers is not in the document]
jonathanmetzman: Actually, the pattern is OK in this context because we are actually testing combined functionality.
But as I pointed out in [this PR](https://github.com/google/clusterfuzz/pull/583), when we test a new stat we assert that stats dictionary is equal to a hardcoded one containing the stat we expect rather than testing for the presence of the stat.
This technical debt requires copying tons of code and updating tons of tests that have nothing to do with the feature being added. Still hoping it gets removed where harmful in this refactor.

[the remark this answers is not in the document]
oliverchang: Yeah, definitely agree with this issue in the existing tests.

PR #836, opened 2019-08-14
Replace an os.rename with a shutil.move (fuchsia)

oliverchang:
os.rename does not work when things are on different filesystems.


oliverchang:
TBR for a trivial fix.

flowerhack:
TIL!

flowerhack:
Thanks for the fix :)

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -405,7 +405,7 @@ def fetch_and_process_logs_and_crash(self):
                           r'\1\2' + crash_testcase_file_path, line)
           new_file.write(line)
     os.remove(self.fuzzer.logfile)
```
inferno-chromium: Do we need this os.remove then ?

PR #837, opened 2019-08-14
Store the crash stacktrace earlier in analyze_task.

mbarbella-chromium:
PTAL when you have time. These changes should prevent some cases where the stack isn't displayed immediately in analyze task (if we're doing a second run for reproducibility, or in certain error cases). This will help me debug an issue with https://clusterfuzz.com/testcase-detail/5720416649084928. There's likely going to be a follow-up fix needed for whatever is causing the failure there, but these changes should be useful regardless.

### DIFF
--- a/src/python/bot/tasks/analyze_task.py
+++ b/src/python/bot/tasks/analyze_task.py
@@ -54,7 +54,7 @@ def store_testcase_dependencies_from_bundled_testcase_archive(
   # Nothing to do if this is not a bundled testcase archive with
   # multiple testcase.
   if not metadata.bundled:
-    return 1
+    return True
 
   # Cleanup, before re-running the app to get resource list.
   process_handler.terminate_stale_application_instances()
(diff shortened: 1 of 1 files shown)


src/python/bot/tasks/analyze_task.py
```diff
@@ -302,6 +303,11 @@ def execute_task(testcase_id, job_type):
   testcase.crash_type = state.crash_type
   testcase.crash_address = state.crash_address
   testcase.crash_state = state.crash_state
+  crash_stacktrace_output = utils.get_crash_stacktrace_output(
+      application_command_line, state.crash_stacktrace,
+      unsymbolized_crash_stacktrace)
+  testcase.crash_stacktrace = data_handler.filter_stacktrace(
+      crash_stacktrace_output)
```
inferno-chromium: So, we want to store stack for this return case ??

```
  # See if we have to ignore this crash.
  if crash_analyzer.ignore_stacktrace(state.crash_stacktrace):
    close_invalid_testcase_and_update_status(testcase, metadata, 'Irrelavant')
    return
```

mbarbella-chromium: It seems like we're hanging somewhere below that return, but I did want to handle this case as well. Either way, we can compute everything at the new spot so no reason not to store it as early as possible.

PR #840, opened 2019-08-14
Cleaned up strategy file.

mukundv-chrome:
Got rid of redundant list in file and renamed lists since this file has strategies for multiple engines.


src/python/fuzzing/strategy.py
```diff
@@ -87,6 +65,23 @@
     CORPUS_SUBSET_STRATEGY,
 ]
 
+# Lists of prefix and boolean strategies maintained for libFuzzer stats.
+libfuzzer_strategies_with_prefix_value = [
```
Dor1s: can you capitalize these two as well? this way all the names will be consistent inside this file

inferno-chromium: +1

PR #842, opened 2019-08-15
Fix potential issue with finding similar issues.

oliverchang:
IssueTracker.find_issues may return a generator.


oliverchang:
TBR. trivial change.

PR #844, opened 2019-08-15
[MAB] Update AFL stats schema

mukundv-chrome:


src/python/metrics/fuzzer_stats_schema.py
```diff
@@ -161,6 +161,18 @@
...
+    'name': 'strategy_corpus_mutations_radamsa',
+    'type': 'INTEGER'
+}, {
+    'mode': 'NULLABLE',
+    'name': 'strategy_corpus_mutations_ml_rnn',
+    'type': 'INTEGER'
+}, {
+    'mode': 'NULLABLE',
+    'name': 'strategy_selection_method',
+    'type': 'STRING'
```
mukundv-chrome: It didn't look like this schema list was in any particular order, so I added at the end

PR #845, opened 2019-08-15
Don't create cross platform variant tasks for test cases with gestures.

mbarbella-chromium:


mbarbella-chromium:
PTAL. Is there a test that should be updated for this? At a quick glance I didn't see anything obvious, but happy to add one if needed.

src/python/bot/tasks/task_creation.py
```diff
@@ -173,6 +176,10 @@ def create_variant_tasks_if_needed(testcase):
     if testcase.project_name != project_name:
       continue
 
+    # Don't create cross-platform variants for test cases with gestures.
+    if testcase.gestures and job.platform != testcase_platform:
```
inferno-chromium: This means we dont try the testcase at all. i would prefer if we did try the testcase on other platforms, but without the gestures set. can shift this change to just like https://github.com/google/clusterfuzz/blob/master/src/python/bot/tasks/variant_task.py#L51

where we compare platform
if testcase.platform != environment.platform().lower():
  testcase.gestures = None (or [])

src/python/bot/tasks/task_creation.py
```diff
@@ -150,6 +150,9 @@ def create_variant_tasks_if_needed(testcase):
     return
 
   testcase_id = testcase.key.id()
+  testcase_platform = data_types.Job.query(
```
inferno-chromium: testcase.platform has this info and avoids the db call.

mbarbella-chromium: It seems like testcase platform loses some of the platform id information, hence this call. Thinking about it, that might be fine in the context of gestures, though. Will remove this.

mbarbella-chromium: No longer relevant in the new implementaiton.

src/python/bot/tasks/task_creation.py
mbarbella-chromium: Good call, this is a much better idea. I don't like the idea of modifying the gesture string since that could lead to some subtle breakages if we ever decided to put the test case in variant task, but addressed it in a similar way.

src/python/bot/testcase_manager.py
```diff
@@ -670,12 +670,14 @@ def test_for_crash_with_retries(testcase,
                                 testcase_path,
                                 test_timeout,
                                 http_flag=False,
-                                compare_crash=True):
+                                compare_crash=True,
+                                use_gestures=True):
```
inferno-chromium: can we move use_gesture before compare_crash, feels like compare_crash should be last and important for visibility. same in caller.

mbarbella-chromium: Done!

PR #846, opened 2019-08-16
[fuchsia] Detect and enable KVM.

flowerhack:
Detecting KVM is tricky, so let's use an environment variable (to
generally be set by the docker image) to determine whether it's safe to
turn KVM on for Fuchsia VMs.

### DIFF
--- a/src/python/platforms/fuchsia/device.py
+++ b/src/python/platforms/fuchsia/device.py
@@ -113,6 +113,12 @@ def qemu_setup():
   ]
   # yapf: enable
 
+  # Detecing KVM is tricky, so let's use an environment variable, set from the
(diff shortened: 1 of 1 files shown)


flowerhack:
@oliverchang

src/python/platforms/fuchsia/device.py
```diff
@@ -113,6 +113,12 @@ def qemu_setup():
   ]
   # yapf: enable
 
+  # Detecing KVM is tricky, so let's use an environment variable, set from the
+  # docker image, to determine whether to turn it on or not.
+  kvm = environment.get_value('FUCHSIA_USE_KVM')
```
oliverchang: nit: no need for the variable. just use get_value in the if.

PR #849, opened 2019-08-16
Use new engine interface for minimization.

oliverchang:
- Stop using hacky launcher arguments.
- Add an RPC for untrusted workers.
- Add untrusted minimize test to run_all_tests.


src/python/bot/tasks/minimize_task.py
```diff
@@ -1085,14 +1087,35 @@ def _run_libfuzzer_testcase(testcase, testcase_file_path):
...
+                                       arguments, testcase_path, output_path,
+                                       timeout)
+
+  target_path = engine_common.find_fuzzer_path(
+      environment.get_value('BUILD_DIR'), target_name)
+
+  engine_impl = LibFuzzerEngine()
+  if tool_name == 'minimize':
+    func = engine_impl.minimize_testcase
+  else:
```
inferno-chromium: assert tool_name == 'cleanse' ?

src/python/tests/test_libs/untrusted_runner_helpers.py
```diff
@@ -196,6 +197,10 @@ def setUp(self):
 
     environment.set_value('GSUTIL_PATH', os.path.dirname(_which('gsutil')))
 
+    test_utils.setup_pubsub('test-clusterfuzz')
```
inferno-chromium: nit: can we use utils.default_project_name to avoid hardcoding 'test-clusterfuzz' in these two statements.

src/python/bot/tasks/minimize_task.py
```diff
@@ -1122,25 +1145,15 @@ def _unset_dedup_flags():
...
   if set_dedup_flags:
     _set_dedup_flags()
 
-  # A small buffer is added to the timeout to allow the current test to
-  # finish, and file to be written. Since we should terminate beforehand, a
-  # long delay only slows fuzzing in cases where it's necessary.
-  _, _, output = process_handler.run_process(command, timeout=timeout + 60)
+  result = run_libfuzzer_engine(tool_name, target_name, arguments,
+                                testcase_file_path, rebased_output_file_path,
+                                timeout)
```
inferno-chromium: Does the 60 sec buffer need to be added to timeout ?

src/python/bot/tasks/minimize_task.py
```diff
@@ -1122,25 +1145,15 @@ def _unset_dedup_flags():
...
-                '--cf-{tool_name}-to={output_file_path}').format(
-                    tool_name=tool_name,
-                    output_file_path=rebased_output_file_path,
-                    timeout=timeout)
-  command = testcase_manager.get_command_line_for_application(
-      file_to_run=testcase_file_path,
-      app_args=arguments,
-      needs_http=testcase.http_flag)
-  logs.log('Executing command: %s' % command)
+  arguments = data_handler.get_arguments(testcase, strip_target=False).split()
```
inferno-chromium: This strip_target feels a little hacky as it relies on location in arguments (maybe we stop using arguments altogether in future). Why not just call get_fuzzer_display().target since we have the testcase object here.

src/python/datastore/data_handler.py
```diff
@@ -301,7 +301,7 @@ def get_fuzzer_display(testcase):
       fully_qualified_name=fuzz_target.fully_qualified_name())
 
 
-def get_arguments(testcase):
+def get_arguments(testcase, strip_target=True):
```
inferno-chromium: we can just remove modifying this function, can just use get_fuzzer_display above.

src/python/bot/untrusted_runner/tasks_impl.py
```diff
@@ -75,3 +76,24 @@ def prune_corpus(request, _):
...
+def process_testcase(request, _):
+  """Process testcase."""
+  tool_name_map = {
+      untrusted_runner_pb2.ProcessTestcaseRequest.MINIMIZE: 'minimize',
+      untrusted_runner_pb2.ProcessTestcaseRequest.CLEANSE: 'cleanse',
+  }
+
+  # TODO(ochang): Support other engines.
+  assert request.engine == 'libFuzzer'
+
```
inferno-chromium: can add assert request.operation in tool_name_map

src/python/bot/tasks/minimize_task.py
```diff
@@ -1085,14 +1087,35 @@ def _run_libfuzzer_testcase(testcase, testcase_file_path):
...
+  if environment.is_trusted_host():
+    from bot.untrusted_runner import tasks_host
+    # TODO(ochang): Remove hardcode.
+    return tasks_host.process_testcase('libFuzzer', tool_name, target_name,
+                                       arguments, testcase_path, output_path,
+                                       timeout)
+
+  target_path = engine_common.find_fuzzer_path(
+      environment.get_value('BUILD_DIR'), target_name)
+
```
inferno-chromium: What if target_path is not found, should we account for that ?

oliverchang: added a check here, to return a dummy value.

src/python/datastore/data_handler.py
oliverchang: Done!

src/python/bot/tasks/minimize_task.py
oliverchang: I don't think this is necessary anymore, since we don't having a timeout which needs to encompass both the launcher process + the actual libFuzzer minimize.

PR #853, opened 2019-08-16
Clarify the regex for runtime.raise stack frame and fix a test.

Dor1s:
Basically the issue is that `runtime.raise` might be a stack frame as is, or a part of a more detailed frame parsed from the sanitizer stacktrace, e.g. from:

```
    #4 0x5c8a90 in runtime.raise runtime/sys_linux_amd64.s:149
```


Dor1s:
Also need to follow up on https://oss-fuzz.com/testcase-detail/5642462671929344

Dor1s:
FTW, i've fixed crash states for the few testcases that were created with this stack frame.

PR #854, opened 2019-08-17
Stack analyzer: better parsing for Golang stacktraces.

Dor1s:
1) skip sanitizer generic regex for known Go crash types.
2) add `fatail error` to the generic `ASSERT` regex.

The problem is that the generic sanitizer regex overwrites the crash type and state extracted by golang regexes, the only example so far is https://oss-fuzz.com/testcase-detail/5642462671929344

Some of these cases might be suppressed by passing `-handle_abrt=0` to libFuzzer, but I'm not sure if this one would be suppressed -- need to check if we prefer that way.


src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -1320,14 +1320,15 @@ def get_crash_data(crash_data, symbolize_flag=True):
...
     # Sanitizer regular crash (includes ills, abrt, etc).
-    update_state_on_match(
-        SAN_ADDR_REGEX,
-        line,
-        state,
-        type_from_group=2,
-        address_from_group=4,
-        reset=True,
-        type_filter=fix_sanitizer_crash_type)
+    if not is_golang:
```
inferno-chromium: I don't think we should disable sanitizer stacks completely. what if we have a go crash without go stack and only asan stack (we want to create a asan stack signature). i think your intent is to not update with sanitizer stack if a go stack is found. this is easy to do with initing a bool found_golang_crash=false and then set it when you detect golang crash type.

```
if update_state_on_match(
     golang_crash_regex, line, state, new_type=golang_crash_type):
         found_golang_crash = True
          state.crash_state = ''
```

and then doing if not found_golang_crash here.

Dor1s: Yes, that is much better, thanks! There is one more example now: https://oss-fuzz.com/testcase-detail/5664480284901376

Dor1s:
@inferno-chromium please take a look again :) I've addressed your comments, but also complicated things a bit more

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -61,7 +61,8 @@
 ASAN_REGEX = re.compile(r'.*(ERROR: AddressSanitizer)[: ]*' r'[ ]*([^(:]+)')
 ASSERT_REGEX = re.compile(
     r'(?:\[.*?\]|.*\.(?:%s):.*)?' % ('|'.join(C_CPP_EXTENSIONS)) +
-    r'\s*(?:ASSERT(?:ION)? FAIL(?:URE|ED)|panic): (.*)', re.IGNORECASE)
+    r'\s*(?:ASSERT(?:ION)? FAIL(?:URE|ED)|panic|fatal error): (.*)',
```
inferno-chromium: Fatal error is not an assertion failure. i looked at some examples like 
https://stackoverflow.com/questions/37551288/golang-fatal-error-runtime-out-of-memory, so this feels like it should be seperate.
Maybe just change change our current fatal error regex. It should then create the right crash type 'Fatal error'
FATAL_ERROR_REGEX = re.compile(r'#\s*Fatal error(:| in) (.*)')

Dor1s: Yeah, that felt weird, but I didn't notice the fatal error regex. Thanks for the suggestion, done.

Dor1s:
Re-running due to some flake:

```
Step #3: RuntimeError: datastore emulator did not get ready in time.
```

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -97,7 +97,8 @@
 FATAL_ERROR_CHECK_FAILURE = re.compile(
     r'#\s+(Check failed: |RepresentationChangerError: node #\d+:)?(.*)')
 FATAL_ERROR_DCHECK_FAILURE = re.compile(r'#\s+(Debug check failed: )(.*)')
-FATAL_ERROR_REGEX = re.compile(r'#\s*Fatal error in (.*)')
+FATAL_ERROR_REGEX = re.compile(r'(^|#\s*)Fatal error(:| in) (.*)',
```
oliverchang: I'm not sure i like conflating two conpletely different crash types here... I think it's better to be explicit about differences rather than trying generalise things here. Especially if we plan on refactoring this code in the future

inferno-chromium: Can do, better to do two different regexes then. Also, that branch does "state.fatal_error_occurred = True", separating will help to not set it in golang change, also simplifies the if that will not be needed if we have two regexes.
```
if is_golang:
          state.crash_state = fatal_error_match.group(3) + '\n'
        else:
          state.crash_state = filter_stack_frame(fatal_error_match.group(3))
```

Dor1s: Done, thanks!

PR #857, opened 2019-08-19
Fix upload filename for testcase re-uploads.

inferno-chromium:
Re-upload testcase uses filename from blob which is a full absolute
path and not just filename. Extract filename from it.


inferno-chromium:
This was causing filenames like 'mntscratch0clusterfuzzbotinputsfuzzer-testcasesb53212acd7bc877cb81c3e8531c4a804f2d1493a34c28a527643b19fdcfuzz-2' in uploads

PR #858, opened 2019-08-19
Rebase testcase path in minimize_task.

oliverchang:
Previously this was handled via get_command_line_for_application.


oliverchang:
TBR to fix issues.

PR #859, opened 2019-08-19
More preparation for testing refactored engine fuzzing.

oliverchang:
- Remove target name from stored arguments. We can just use the Testcase
  overridden_fuzzer_name attribute.
- Register 'libFuzzer_test' engine during startup in run_bot.
- Allow APP_PATH to not be set if APP_NAME isn't specifiedl.
- Support fuzz targets in custom builds.
- Various other fixes.

### DIFF
--- a/src/python/bot/fuzzers/init.py
+++ b/src/python/bot/fuzzers/init.py
@@ -1,0 +1,22 @@
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Fuzzing engine initialization."""
+
+from bot.fuzzers import engine
+from bot.fuzzers.libFuzzer import engine as libFuzzer_engine
+
+
+def run():
+  """Initialise builtin fuzzing engines."""
+  engine.register('libFuzzer_test', libFuzzer_engine.LibFuzzerEngine)
--- a/src/python/bot/fuzzers/libFuzzer/engine.py
+++ b/src/python/bot/fuzzers/libFuzzer/engine.py
@@ -59,7 +59,7 @@ class LibFuzzerEngine(engine.Engine):
 
   @property
   def name(self):
-    return 'libFuzzer'
+    return 'libFuzzer_test'
 
   def prepare(self, corpus_dir, target_path, _):
     """Prepare for a fuzzing session, by generating options. Returns a
--- a/src/python/bot/tasks/fuzz_task.py
+++ b/src/python/bot/tasks/fuzz_task.py
@@ -144,7 +144,7 @@ class Crash(object):
         http_flag=needs_http)
 
   @classmethod
-  def from_engine_crash(cls, target_name, crash):
+  def from_engine_crash(cls, crash):
     """Create a Crash from a engine.Crash."""
     return Crash(
         file_path=crash.input_path,
@@ -153,8 +153,7 @@ class Crash(object):
         resource_list=[],
         gestures=[],
         unsymbolized_crash_stacktrace=crash.stacktrace,
-        # We store the target name as the first argument.
-        arguments=[target_name] + crash.reproduce_args,
+        arguments=' '.join(crash.reproduce_args),
         application_command_line='')  # TODO(ochang): Write actual command line.
 
   def __init__(self,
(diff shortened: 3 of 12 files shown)


src/python/bot/fuzzers/init.py
```diff
@@ -0,0 +1,22 @@
...
+# limitations under the License.
+"""Fuzzing engine initialization."""
+
+from bot.fuzzers import engine
+from bot.fuzzers.libFuzzer import engine as libFuzzer_engine
+
+
+def run():
+  """Initialise builtin fuzzing engines."""
+  engine.register('libFuzzer_test', libFuzzer_engine.LibFuzzerEngine)
```
inferno-chromium: I am guessing this is temporary as you are testing things ? Will this start creating any testcase with libFuzzer_test engine ?

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1673,8 +1677,8 @@ def run(self):
 
     # Check if we have an application path. If not, our build failed
     # to setup correctly.
-    app_path = environment.get_value('APP_PATH')
-    if not app_path:
+    if (environment.get_value('APP_NAME') and
```
inferno-chromium: Why are we depending on environment.get_value('APP_NAME') here ? What is the case where we dont have APP_NAME, but dont have APP_PATH ?

src/python/bot/testcase_manager.py
```diff
@@ -512,21 +512,25 @@ def __init__(self,
     self._gestures = gestures
     self._needs_http = needs_http
 
-    engine_impl = engine.get(fuzzer_name)
-    if engine_impl:
+    fuzz_target = data_handler.get_fuzz_target(fuzzer_name)
+    if fuzz_target:
+      engine_impl = engine.get(fuzz_target.engine)
+      if not engine_impl:
+        raise RuntimeError(engine_impl)
```
inferno-chromium: Why engine_impl in argument as it is None ? Maybe a message is better 'Engine not found or something'

src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -59,7 +59,7 @@ class LibFuzzerEngine(engine.Engine):
 
   @property
   def name(self):
-    return 'libFuzzer'
+    return 'libFuzzer_test'
```
Dor1s: +1, is `_test` temporary?

src/python/bot/tasks/fuzz_task.py
```diff
@@ -278,8 +277,12 @@ def __init__(self, crashes, context):
       assert crashes[0].security_flag == c.security_flag
 
     self.crashes = crashes
+    fully_qualified_fuzzer_name = context.fuzzer_name
+    if context.fuzz_target:
```
Dor1s: nit: why not

```
else:
  fully_qualified_fuzzer_name = context.fuzzer_name
```

IMO that'd be a bit more readable, even though one line more :)

src/python/bot/testcase_manager.py
Dor1s: good catch

src/python/tests/core/bot/tasks/fuzz_task_test.py
```diff
@@ -1368,7 +1369,7 @@ def test_basic(self):
     self.assertEqual(1, crashes[0].return_code)
     self.assertEqual('stack', crashes[0].unsymbolized_crash_stacktrace)
     self.assertEqual(1.0, crashes[0].crash_time)
-    self.assertListEqual(['test_target', 'args'], crashes[0].arguments)
+    self.assertEqual('args', crashes[0].arguments)
```
Dor1s: shouldn't this be a list?

src/python/bot/tasks/fuzz_task.py
oliverchang: This is for the case of the fuzzing pipeline, where we don't have APP_NAME = launcher.py. otherwise we bail because APP_PATH isn't set.

src/python/bot/fuzzers/libFuzzer/engine.py
oliverchang: Yes this is for testing only.

src/python/bot/fuzzers/init.py
oliverchang: Yes this is a temporary thing. Will only be enabled with a small number of test jobs marked as experimental.

oliverchang:
Compatibility with older testcases will be addressed in a future PR. This is just to test the new implementation with a small number of jobs.

src/python/bot/testcase_manager.py
oliverchang: Oops, fixed.

src/python/bot/tasks/fuzz_task.py

src/python/tests/core/bot/tasks/fuzz_task_test.py
oliverchang: This was wrong before. This needs to be same type as Testcase.minimized_arguments, which is a string.

PR #863, opened 2019-08-19
Fix paper-spinner not showing in fuzzer, upload testcase forms, speedup fuzzer update

### LINKED ISSUE #862: Fuzzers page does not work well with large number of jobs
For libFuzzer, if we have a lot of existing jobs (>100) and want to add a new job and associate it. Submit button feels stuck, does not show updates, and take 1-2 min to finish. Can we show some update or something better to optimize this when only one job is updated.

@oliverchang as fyi.

inferno-chromium:
Fixes #862

### DIFF
--- a/src/python/fuzzing/fuzzer_selection.py
+++ b/src/python/fuzzing/fuzzer_selection.py
@@ -37,18 +37,18 @@ def update_mappings_for_fuzzer(fuzzer, mappings=None):
   entities = ndb_utils.get_all_from_query(query)
   old_mappings = {}
   for entity in entities:
-    old_mappings[(entity.job, entity.platform)] = entity
+    old_mappings[entity.job] = entity
 
   new_mappings = []
   for job_name in mappings:
+    mapping = old_mappings.pop(job_name, None)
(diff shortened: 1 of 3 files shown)


src/python/fuzzing/fuzzer_selection.py
```diff
@@ -37,20 +37,20 @@ def update_mappings_for_fuzzer(fuzzer, mappings=None):
...
-    old_mappings[(entity.job, entity.platform)] = entity
+    old_mappings[entity.job] = entity
 
   new_mappings = []
   for job_name in mappings:
+    mapping = old_mappings.pop(job_name, None)
+    if mapping:
+      continue
+
     job = data_types.Job.query(data_types.Job.name == job_name).get()
```
inferno-chromium: I don't think we should query everytime, performance benefits are significant here.
also, i dont think job can be associated with multiple platforms, so this optimization works.

inferno-chromium:
@mbarbella-chromium fyi. i see jobs update with platform change calls update_platform_for_job, so we should be good.

PR #866, opened 2019-08-20
Fuzzing engine fixes.

oliverchang:
- Add -runs=100 to reproduce runs.
- Run engine MAX_TESTCASES times in a given session.
- Add log header to reproduce output.
- Support reproducing with legacy testcases.


src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -303,10 +303,13 @@ def reproduce(self, target_path, input_path, arguments, max_time):
     runner = libfuzzer.get_runner(target_path)
     launcher.set_sanitizer_options(target_path)
 
+    runs_argument = constants.RUNS_FLAG + str(constants.RUNS_TO_REPRODUCE)
```
Dor1s: Sorry if I asked this before, but shouldn't we call `remove_fuzzing_arguments` here? I guess the assumption is that these arguments are coming from a testcase and therefore we shouldn't have any fuzzing arguments (like `-dict=`) in there?

oliverchang: Yes, we call this in fuzz(). This is why an engine.Crash object has a "reproduce_args" attribute.

However, this may not be the case for legacy testcases. Will think about how to best handle this.

PR #868, opened 2019-08-20
Create a /dev/shm directory since some fuzzers might use it.

jonathanmetzman:
Don't bother making it a tmpfs directory for the sake of simplicity.
At some point I want to switch to using a chroot instead of minijail. When I do this, I will actually mount the real /dev directory, and this "fix" and any alternatives will become obsolete.

### DIFF
--- a/src/python/system/minijail.py
+++ b/src/python/system/minijail.py
@@ -146,12 +146,13 @@ class MinijailChroot(object):
       logs.log_error('Failed to call mknod.', output=e.output)
 
   def _create_devices(self):
-    """Create /dev/null, /dev/random and /dev/urandom."""
+    """Create /dev/null, /dev/random, /dev/urandom, and /dev/shm."""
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
The right way to do this would be to do something like this: `mount -t tmpfs -o size=512m tmpfs /mnt/ramdisk`

But that will require some small refactoring of the method and its unittests. I don't really think it is worth doing that since this an edge case, and also I'm not sure what the performance implications are of creating a ramdisk/tmpfs that most fuzzers don't use (are the non-users negatively impacted?)

jonathanmetzman:
Fixes crbug.com/991259

oliverchang:
> Don't bother making it a tmpfs directory for the sake of simplicity.
At some point I want to switch to using a chroot instead of minijail. When I do this, I will actually mount the real /dev directory, and this "fix" and any alternatives will become obsolete.

Yes! Then we can completely get rid of minijail once the refactor is done.

PR #869, opened 2019-08-20
Add vendor_ccs field for OSS-Fuzz project.yaml

inferno-chromium:


src/appengine/handlers/cron/project_setup.py
```diff
@@ -422,22 +422,26 @@ def sync_user_permissions(project, info):
...
-      raise ProjectSetupError('Bad primary_contact %s.' % primary_contact)
-
-  if 'auto_ccs' in info:
-    auto_ccs = info.get('auto_ccs')
-    if isinstance(auto_ccs, list):
-      ccs.extend(auto_ccs)
-    elif isinstance(auto_ccs, basestring):
-      ccs.append(auto_ccs)
+
+  def _get_ccs(field_name, allow_list=True):
```
mbarbella-chromium: Is allow_list needed since we have the isinstance check anyway?

inferno-chromium: I mean we shouldn't allow primary_contact to be a list even if the maintainer gives it that way ? Can remove it too, just wanted to match old style.

mbarbella-chromium: Ah, that seems reasonable.

PR #870, opened 2019-08-20
Fix exceptions with testcase reproduction for non-engine testcases.

oliverchang:
overridden_fuzzer_name is not always set.


oliverchang:
TBR to fix exceptions. /gcbrun.

PR #871, opened 2019-08-21
Fix exception with getting fuzzer name of uploaded testcase.

oliverchang:
In such cases, there is no fuzzer or overridden fuzzer set.


oliverchang:
TBR for simple fix.

PR #875, opened 2019-08-22
Fix strategy column names for new libFuzzer fuzzing implementation.

oliverchang:
Some strategies embed a numeric value in the name. This is expected to be
parsed before being set in the stats.

In a future PR, once we complete the migration, we can get rid of the
encoding + decoding steps which append the '_value' and then parses it.

### DIFF
--- a/src/python/tests/core/bot/tasks/fuzz_task_test.py
+++ b/src/python/tests/core/bot/tasks/fuzz_task_test.py
@@ -1344,7 +1344,7 @@ class DoEngineFuzzingTest(fake_filesystem_unittest.TestCase):
     engine_impl.prepare.return_value = engine.FuzzOptions(
         '/corpus', ['arg'], {
             'strategy_1': 1,
-            'strategy_2': 2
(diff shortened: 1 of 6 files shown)


src/python/tests/core/bot/tasks/fuzz_task_test.py
```diff
@@ -1342,7 +1342,10 @@ def test_basic(self):
     engine_impl = mock.Mock()
     engine_impl.name = 'libFuzzer'
     engine_impl.prepare.return_value = engine.FuzzOptions(
-        '/corpus', ['arg'], ['strategy_1', 'strategy_2'])
+        '/corpus', ['arg'], {
+            'strategy_1': 1,
+            'strategy_2': 2
```
Dor1s: super nit: maybe change `2` to e.g. `50` to keep it super obvious that it might be any value, not necessarily 1, 2, 3, etc.

oliverchang: Done!

PR #878, opened 2019-08-23
Allow non-string additional vars in project setup.

inferno-chromium:
This is needed to set UBSan to true for builds that have both
ASan and UBSan combined.


inferno-chromium:
TBR simple string conversion fix.

PR #861, opened 2019-08-19
Account for testcase minimized flags when picking variant task argument.

inferno-chromium:
Add unique arguments from testcase minimized arguments that don't
exist in APP_ARGS. Fixes #746 and also fixes the regression that
cause engine fuzzer testcase to stop working.

PR #880, opened 2019-08-23
Fix wrong WINDOW_ARG used in variant task.

inferno-chromium:
This is a regression from
https://github.com/google/clusterfuzz/pull/861
Previous code always set it and was just doing the unneeded
hasattr check.


inferno-chromium:
Because of this regression, variant task shows weird crashes like

```
2 == sscanf(size_str.c_str(), "%dx%d", &width, &height) in shell.cc
content::Shell::GetShellDefaultSize
content::Shell::Initialize
```

PR #884, opened 2019-08-24
Use ints for sanitizer field values.

inferno-chromium:
symbolize=1 relies on int. Also, we convert int to string when
joining for final _OPTIONS string.


inferno-chromium:
TBR to make my last change actually work.

PR #885, opened 2019-08-25
CSS improvements

inferno-chromium:
- Make search box and dropdown filters consistent, with respect
to iron-icon margin, color, etc.
- Share css styles for dropdown filters.
- Add testcase project name to upload testcase page.


inferno-chromium:
TBR minor css fixes.

PR #895, opened 2019-08-27
Untrusted fixes for new fuzzing pipeline.

oliverchang:
- Rebase all paths.
- Pull testcases from worker after end of fuzzing.


oliverchang:
TBR to get unblock some testing

PR #896, opened 2019-08-27
Recreate FUZZ_DATA as well in shell.clear_data_directories.

oliverchang:


oliverchang:
ah, good point. I can add a call to persistent_cache.clear_values().

oliverchang:
> > ah, good point. I can add a call to persistent_cache.clear_values().
> 
> Yes exactly!

Done. I had to remove the dependency on shell in persistent_cache to avoid a circular dependency.

PR #897, opened 2019-08-27
Fix potential exception when deleting a persistent cache value.

oliverchang:
delete_value() may be called on a value that doesn't exist.


oliverchang:
TBR trivial fix.

Issue #907, opened 2019-08-28
/corpus-backup/make-public is failing on OSS-Fuzz


Dor1s:
Trying to re-run, below is the error message from yesterday:

```
'NoneType' object has no attribute 'project_qualified_name' (/base/data/home/apps/s~clusterfuzz-external/cron-service:.../handlers/base_handler.py:232)
Traceback (most recent call last):
  File "/base/alloc/tmpfs/dynamic_runtimes/python27g/941d77da994078b1/python27/python27_lib/versions/third_party/webapp2-2.3/webapp2.py", line 545, in dispatch
    return method(*args, **kwargs)
  File "/base/data/home/apps/s~clusterfuzz-external/cron-service:.../libs/handler.py", line 95, in wrapper
    return func(self)
  File "/base/data/home/apps/s~clusterfuzz-external/cron-service:.../handlers/cron/corpus_backup.py", line 103, in get
    corpus_backup_bucket_name)
  File "/base/data/home/apps/s~clusterfuzz-external/cron-service:.../handlers/cron/corpus_backup.py", line 37, in _make_corpus_backup_public
    target.project_qualified_name(), corpus_backup_date)
AttributeError: 'NoneType' object has no attribute 'project_qualified_name'
```

PR #908, opened 2019-08-28
Corpus backup task: silently skip outdated fuzz targets (#907).

Dor1s:
I think we can make it a warning, but I'd prefer to figure out why the issue is happening in the first place. Maybe there is some other issue which we can fix and avoid these confusing failures across multiple cron tasks. This should help us to track down the root cause.

### DIFF
--- a/src/appengine/handlers/cron/corpus_backup.py
+++ b/src/appengine/handlers/cron/corpus_backup.py
@@ -99,5 +99,9 @@ class MakePublicHandler(base_handler.Handler):
           target_jobs)
 
       for target in fuzz_targets:
+        if not target:
+          # This is expected if any fuzzer/job combinations become outdated.
+          continue
+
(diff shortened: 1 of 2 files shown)


mbarbella-chromium:
> If no one else is worried about this, I can add a `if not fuzz_target` check in the corpus backup task to skip such instances.

I'm not too worried about these cases in general so avoiding this change and checking it there sounds better to me. I don't feel too strongly if anyone else has an opinion.

oliverchang:
> > If no one else is worried about this, I can add a `if not fuzz_target` check in the corpus backup task to skip such instances.
> 
> I'm not too worried about these cases in general so avoiding this change and checking it there sounds better to me. I don't feel too strongly if anyone else has an opinion.

That also sounds better to me.

Dor1s:
Thanks! I changed the CL as discussed above. PTAL :)

Issue #353, opened 2019-04-01
Minimize *SAN_OPTIONS


mbarbella-chromium:
Sanitizer options should be minimized in a similar fashion to command line arguments during minimization. We currently specify a long list of sanitizer options, and most of them aren't needed. If we could remove all or most of these, it would make it more obvious to developers when they actually need to set them as well as which are important.

inferno-chromium:
See also https://bugs.chromium.org/p/chromium/issues/detail?id=787930

inferno-chromium:
Only the non-engine fuzzer usecase is left, putting that back in Marty's queue.

rajibmitra:
I would love to work on this if no one is working at this moment.

mbarbella-chromium:
I'm not planning on working on this in the short term (and will update the issue if I do start it), so if you'd like to take a look in the meantime I'd be happy to review any pull requests.

PR #909, opened 2019-08-28
Minimize *SAN_OPTIONS for engine fuzzers

inferno-chromium:
Part 1 of fixing https://github.com/google/clusterfuzz/issues/353

### DIFF
--- a/src/python/bot/tasks/minimize_task.py
+++ b/src/python/bot/tasks/minimize_task.py
@@ -1268,10 +1268,7 @@ def do_libfuzzer_minimization(testcase, testcase_file_path):
           timeout,
           expected_state.crash_state,
           set_dedup_flags=True)
-      if (not crash_result or crash_result.is_security_issue() !=
-          initial_crash_result.is_security_issue() or
-          not CrashComparer(crash_result.get_state(),
-                            initial_crash_result.get_state()).is_similar()):
+      if not crash_result:
         logs.log(
             'Skipped needed {options_env_var} option: {option_name}'.format(
                 options_env_var=options_env_var, option_name=option_name))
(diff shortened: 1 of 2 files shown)


inferno-chromium:
this might slow down minimize untrusted test, i can fix it later before landing, but review should be ready.

src/python/bot/tasks/minimize_task.py
```diff
@@ -1244,6 +1248,40 @@ def handle_unreproducible():
...
 
+  # Minimize *_OPTIONS env variable first.
+  env = {}
+  for tool in environment.SUPPORTED_MEMORY_TOOLS_FOR_OPTIONS:
+    options_env_var = tool + '_OPTIONS'
+    options = environment.get_memory_tool_options(options_env_var)
+    if not options:
+      continue
+
+    minimized_options = copy.deepcopy(options)
```
oliverchang: nit: I think deepcopy is overkill here. We can just use options.copy() which performs a shallow copy.

src/python/bot/tasks/minimize_task.py
```diff
@@ -1244,6 +1248,40 @@ def handle_unreproducible():
...
+    for option_name, option_var in six.iteritems(options):
+      minimized_options.pop(option_name)
+      environment.set_memory_tool_options(options_env_var, minimized_options)
+
+      _, crash_result = _run_libfuzzer_tool(
```
oliverchang: should we be calling libfuzzer with minimize here? can we just run the testcase without minimizing (i.e. by calling _run_libfuzzer_testcase)? seems weird to do it here and then again from lines 1287 onwards.

src/python/bot/tasks/minimize_task.py
oliverchang: I think shallow copies should be sufficient here -- we don't need to make copies of the individual strings in the options dict.

PR #911, opened 2019-08-29
Clear all persistent cache values when clearing data dirs.

### LINKED ISSUE #903: fuzzer-common-data-bundles is not accessible
```
fs.js:145
    throw err;
    ^

Error: ENOENT: no such file or directory, scandir '/mnt/scratch0/bots/oss-fuzz-linux-zone1-host-high-end-b1b2-3/clusterfuzz/bot/inputs/fuzzer-common-data-bundles/WebKit/JSTests'
```

same error for other data bundles, e.g. spidermonkey

oliverchang:
Fixes #903.

### DIFF
--- a/src/python/base/persistent_cache.py
+++ b/src/python/base/persistent_cache.py
@@ -36,7 +36,7 @@ def initialize():
     os.makedirs(cache_directory_path)
 
 
-def clear_values():
+def clear_values(clear_all=False):
   """Remove all values."""
   cache_directory_path = environment.get_value('CACHE_DIR')
   if not os.path.exists(cache_directory_path):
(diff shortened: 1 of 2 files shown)


oliverchang:
merging to fix exceptions.

inferno-chromium:
Can you please explain why we keep persist files ?

oliverchang:
> Can you please explain why we keep persist files ?

Because those indicate values that survive across reboots. clear_values() is called on startup.

inferno-chromium:
> > Can you please explain why we keep persist files ?
> 
> Because those indicate values that survive across reboots. clear_values() is called on startup.

ah right, remember now, ah, so clear_values wasn't clearing all, yikes!

PR #915, opened 2019-08-29
Add source mapping for custom GitLab servers + tests.

Dor1s:
People have told me that we have some projects hosting at e.g. https://gitlab.gnome.org/GNOME/libxml2 and their stacktraces are not linkified.

### DIFF
--- a/src/python/build_management/source_mapper.py
+++ b/src/python/build_management/source_mapper.py
@@ -103,7 +103,8 @@ class GitHubVCS(VCSViewer):
 
 
 class GitLabVCS(GitHubVCS):
-  VCS_URL_REGEX = re.compile(r'(https://gitlab\.com/(.*?))(\.git)?$')
+  VCS_URL_REGEX = re.compile(
+      r'(https://gitlab(\.[\w\.\-]+)?\.(com|org)/(.*?))(\.git)?$')
 
 
 class GoogleSourceVCS(VCSViewer):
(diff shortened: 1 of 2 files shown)


src/python/build_management/source_mapper.py
```diff
@@ -103,7 +103,8 @@ class GitHubVCS(VCSViewer):
 
 
 class GitLabVCS(GitHubVCS):
-  VCS_URL_REGEX = re.compile(r'(https://gitlab\.com/(.*?))(\.git)?$')
+  VCS_URL_REGEX = re.compile(
+      r'(https://gitlab(\.[\w\.\-]+)?\.(com|org)/(.*?))(\.git)?$')
```
inferno-chromium: This is fine, but the reason it is not working is the missing vars like VCS_REVISION_SUB, VCS_REVISION_PATH_LINE_SUB, VCS_REVISION_DIFF_SUB

see https://github.com/google/clusterfuzz/blob/master/src/python/build_management/source_mapper.py#L210 which is used for linkification.

Dor1s: I think it inherits those from `GitHubVCS`, and it seems to work for the projects hosted at `gitlab.com`, e.g. https://oss-fuzz.com/testcase-detail/5665832347107328

inferno-chromium: ah right, didnt notice that. awesome thanks!

PR #916, opened 2019-08-29
[MAB] Fix a bug causing us to delete valid strategy probabilities.

mbarbella-chromium:


mbarbella-chromium:
PTAL. Just a quick bug fix needed to re-enable multi-armed bandit.

PR #921, opened 2019-08-30
Sanitizer options minimization fixes

inferno-chromium:
- Store minimized options even for the bailout case when
  _skip_minimization or libfuzzer minimization failed.
- Use minimized options for post minimization tasks.

Final part of addressing #353

### DIFF
--- a/src/python/bot/tasks/minimize_task.py
+++ b/src/python/bot/tasks/minimize_task.py
@@ -451,6 +451,7 @@ def execute_task(testcase_id, job_type):
     return
 
   if flaky_stack:
+    testcase = data_handler.get_testcase_by_id(testcase_id)
     testcase.flaky_stack = flaky_stack
     testcase.put()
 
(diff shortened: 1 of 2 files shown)


src/python/bot/tasks/minimize_task.py
```diff
@@ -1287,6 +1286,9 @@ def handle_unreproducible():
 
     environment.set_memory_tool_options(options_env_var, minimized_options)
     env[options_env_var] = environment.get_memory_tool_options(options_env_var)
+  if env:
+    testcase = data_handler.get_testcase_by_id(testcase.key.id())
+    testcase.set_metadata('env', env)
```
oliverchang: maybe we should promote this to an actual entity property?

inferno-chromium: Can do this, but looks like JsonProperty is not supported both in ndb patcher and also in generate_datastore_models.

oliverchang: I think that's just a wrapper which stores the serialised json as a blob property, so we can easily implement a similar wrapper like `get_json` which just does json.loads() on a property.

inferno-chromium: Thought more on this. We won't be using |env| on unreproducible testcases (since they dont get minimized) and on non-engine fuzzer testcase, so feels like we are unnecessary inflating the entity there. Also, set_metadata and get_metadata exactly do the json dumps and loads, so doing get_json and set_json is kind of duplicating that. Skipping for now, will also talk to you on if your get_json/set_json suggestion was in testcase class or ndb_patcher itself.

PR #922, opened 2019-08-30
Remove environment variable when value is empty.

inferno-chromium:


inferno-chromium:
> Just looked at the rest of the function. Ignore the previous comment, this should be fine.

Basically, we were called reset environment even before this call here - https://github.com/google/clusterfuzz/blob/master/src/python/bot/tasks/commands.py#L83
This caused ASAN_OPTIONS to be set regardless. We want to override that.

PR #923, opened 2019-08-31
Always reset memory tool options when preparing the environment.

Dor1s:
This partially reverts #921. I suspect that this is why we have marked some testcases as Unreproducible this morning (see the chat for more context).

### DIFF
--- a/src/python/bot/tasks/setup.py
+++ b/src/python/bot/tasks/setup.py
@@ -126,9 +126,6 @@ def _get_application_arguments(testcase, task_name):
 def _setup_memory_tools_environment(testcase):
   """Set up environment for various memory tools used."""
   env = testcase.get_metadata('env')
-  if not env:
-    environment.reset_current_memory_tool_options(redzone_size=testcase.redzone)
-    return
 
   for options_name, options_value in six.iteritems(env):
     if not options_value:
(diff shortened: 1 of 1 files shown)


inferno-chromium:
Closing since root cause seem fixed now with https://github.com/google/clusterfuzz/pull/924
Basically, print_summary=1 is needed in UBSan to know it is a sanitizer crash and we have a complete stack (end of stack marker). if it is not there, we can't know if we have a full crash. 
libFuzzer *SAN_OPTIONS minimization missed the crash_result.is_crash() crash which would have returned false on removing the print_summary. 
I have deleted env metadata from all existing testcases, so they should use the full options. i have changed all those testcases to reproducible and updated all bugs for that mistake (thanks Max for query). we can force minimization on them next week to get minimized options. for now, test works with https://oss-fuzz.com/testcase-detail/5710383731441664 (where minimization tried and then progression still says reproducible). 

```
  # Check if we have a complete stacktrace by location stacktrace end marker.
  # If not, bail out.
  if not has_marker(stacktrace, STACKTRACE_END_MARKERS):
    return False
```

oliverchang:
Actually closing per https://github.com/google/clusterfuzz/pull/923#issuecomment-526882051

PR #924, opened 2019-08-31
Fix issue with ubsan crashes turning unreproducible.

inferno-chromium:
- Add missing is_crash() in minimize options. This makes sure
  that it is not minimizing any option that changes return code to
  zero.
- Explicit fix return code when a memory tool output or check failure
  output is found. This check was in process_handler.run_process, but
  not in engine.reproduce branch.

### DIFF
--- a/src/python/bot/tasks/minimize_task.py
+++ b/src/python/bot/tasks/minimize_task.py
@@ -1263,8 +1263,7 @@ def do_libfuzzer_minimization(testcase, testcase_file_path):
       reproduced = False
       for _ in range(MINIMIZE_SANITIZER_OPTIONS_RETRIES):
         crash_result = _run_libfuzzer_testcase(testcase, testcase_file_path)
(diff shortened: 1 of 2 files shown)


inferno-chromium:
@oliverchang - can you please change minimize task's _run_libfuzzer_testcase to use engine_impl.reproduce and not process_handler.run_process. this caused this bug to surface since one fixed return code whereas other one didn't :(

inferno-chromium:
TBR

oliverchang:
> @oliverchang - can you please change minimize task's _run_libfuzzer_testcase to use engine_impl.reproduce and not process_handler.run_process. this caused this bug to surface since one fixed return code whereas other one didn't :( also, see [#923 (comment)](https://github.com/google/clusterfuzz/pull/923#issuecomment-526882051)

LGTM, thanks for fixing this! Yes this was in the plans :)

Issue #910, opened 2019-08-28
Weight projects by number of targets


oliverchang:
Per title. As an automated way of weighting projects, we can set the project's weight equal to its number of targets.

Dor1s:
What about OSS-Fuzz weights specified manually? Would be nice to get rid of it, but I'm afraid we still need it to account for project importance somehow.

PR #925, opened 2019-09-02
Add automated job weighting based on number of targets (#910).

oliverchang:
- Set FUZZ_TARGET_COUNT in build_manager when picking a fuzz target.
- Store this count in Datastore, and use to automaticaly determine job
  weights. Manually set weights still take precedence.
- Rename fuzzer_weights.py to fuzzer_and_job_weights.py
- Remove symbolized and production setup RPCs as they are unused.

OSS-Fuzz project weights aren't affected by this PR, and will come in a later one.


oliverchang:
WIP. not ready for review. /gcbrun

src/appengine/handlers/cron/fuzzer_and_job_weights.py
```diff
@@ -20,19 +20,32 @@
...
 QuerySpecification = collections.namedtuple(
     'QuerySpecification', ['query_format', 'formatter', 'reason'])
 
 SpecificationMatch = collections.namedtuple('SpecificationMatch',
                                             ['new_weight', 'reason'])
 
+DEFAULT_MULTIPLIER = 10.0  # Used for blackbox and unrun jobs.
+DEFAULT_SANITIZER_WEIGHT = 0.1
+
+SANITIZER_WEIGHTS = {
```
oliverchang: these are up for discussion/debate.

src/python/build_management/build_manager.py
```diff
@@ -392,6 +392,9 @@ def _get_fuzz_targets_from_dir(build_dir):
 
 def _set_random_fuzz_target_for_fuzzing_if_needed(fuzz_targets, target_weights):
   """Sets a random fuzz target for fuzzing."""
+  fuzz_targets = list(fuzz_targets)
```
inferno-chromium: can you remove repeation of 
fuzz_targets = list(fuzz_targets) from line 406

src/python/datastore/data_types.py
```diff
@@ -1326,6 +1331,12 @@ class FuzzerJob(Model):
   job = ndb.StringProperty()
   platform = ndb.StringProperty()
   weight = ndb.FloatProperty(default=1.0)
+  automated_weight = ndb.FloatProperty(default=1.0)
+
+  @property
+  def actual_weight(self):
+    """Get the actual weight for this job."""
+    return self.weight or self.automated_weight
```
inferno-chromium: When will self.weight is false ? with default as 1.0, wont it always take precedence over automated_weight ?

src/appengine/handlers/cron/fuzzer_and_job_weights.py
```diff
@@ -326,6 +339,35 @@ def store_current_weights_in_bigquery():
...
+
+
+def update_job_weights():
+  """Update job weights."""
+  for job in data_types.Job.query():
+    multiplier = DEFAULT_MULTIPLIER
+    if environment.is_engine_fuzzer_job(job.name):
+      targets_count = ndb.Key(data_types.FuzzTargetsCount, job.name).get()
+      if targets_count:
+        multiplier = targets_count.count
```
inferno-chromium: I am a little bit worried about this. So, a project like ffmpeg with >400 targets should have 400 weight ? isnt that too high, should we have a cap so that cycles are not taken up by one project. (maybe a test for this ?)

src/appengine/handlers/cron/fuzzer_and_job_weights.py
Dor1s: looks reasonable, though I wouldn't mind equalizing MSan and UBSan

src/appengine/handlers/cron/fuzzer_and_job_weights.py
```diff
@@ -20,19 +20,32 @@
...
     'QuerySpecification', ['query_format', 'formatter', 'reason'])
 
 SpecificationMatch = collections.namedtuple('SpecificationMatch',
                                             ['new_weight', 'reason'])
 
+DEFAULT_MULTIPLIER = 10.0  # Used for blackbox and unrun jobs.
+DEFAULT_SANITIZER_WEIGHT = 0.1
+
+SANITIZER_WEIGHTS = {
+    'ASAN': 0.5,
```
Dor1s: How about we use a factor of `SANITIZER_WEIGHTS` for all of these, e.g. `5 * SANITIZER_WEIGHTS`, `1 * SANITIZER_WEIGHTS`, etc?

src/appengine/handlers/cron/fuzzer_and_job_weights.py
Dor1s: Why not? FFmpeg is a huge project with dozens of dependencies. If the goal of this code alone is to allocate the same number of cycles for every individual fuzz target, I'm against any capping or e.g. logarithmic weight calculation.

Same logic applies for the internal stuff. If a directory has a lot of fuzz targets, why should they get fewer cycles each?

Dor1s: And we still have manually assigned weights at our discretion. Any super critical stuff can get a higher weight in there and each fuzz target will get more CPU cycles than the default.

src/python/datastore/data_types.py
```diff
@@ -1105,6 +1105,11 @@ def fuzz_target_project_qualified_name(project, binary):
   return normalized_project_prefix + binary
 
 
+class FuzzTargetsCount(Model):
+  """Fuzz targets count."""
```
Dor1s: Maybe add a comment that entity's id is the corresponding job name?

src/python/datastore/data_types.py
Dor1s: Should we return the product of these two? That way we'll always have equal CPU allocation by default, but special preference might be assigned through the manual weights.

src/appengine/handlers/cron/fuzzer_and_job_weights.py
```diff
@@ -20,19 +20,32 @@
...
 from metrics import logs
+from system import environment
 
 QuerySpecification = collections.namedtuple(
     'QuerySpecification', ['query_format', 'formatter', 'reason'])
 
 SpecificationMatch = collections.namedtuple('SpecificationMatch',
                                             ['new_weight', 'reason'])
 
+DEFAULT_MULTIPLIER = 10.0  # Used for blackbox and unrun jobs.
```
jonathanmetzman: nit: "unrun" -> "not yet run"

src/appengine/handlers/cron/fuzzer_and_job_weights.py
jonathanmetzman: How are different architectures going to be handled?
In internal CF we don't give x86/arm* simulator equal weight to x64. If we continue assigning manual weights, will this still hold.

Is this also true for non-blackbox vs blackbox jobs?

src/python/datastore/data_types.py
jonathanmetzman: Should probably rename field if it is going to be a multiplier and not a simple weight.

src/appengine/handlers/cron/fuzzer_and_job_weights.py
mbarbella-chromium: +1 to Max's comments. I think cases where we don't want to treat targets as equal to one another are the uncommon ones, and we can just use manual weights if we ever feel like a particular project is using too many CPUs.

src/appengine/handlers/cron/fuzzer_and_job_weights.py
```diff
@@ -20,19 +20,32 @@
 from base import utils
 from datastore import data_handler
 from datastore import data_types
+from datastore import ndb
```
mbarbella-chromium: Is there any advantage to making these changes in this file rather than a new one? I don't feel too strongly if you think merging them makes more sense, but I didn't see anything obvious at a glance.

src/appengine/handlers/cron/fuzzer_and_job_weights.py
oliverchang: We currently don't give architectures special consideration. I can add a TODO here.

What do you mean re non-blackbox vs blackbox jobs? Currently we assign a default of 10 to those, but I expect that we need to set most of them manually.

jonathanmetzman: > We currently don't give architectures special consideration. I can add a TODO here.
> 

Another thought that is probably a topic for future discussion, but I suspect the "optimal" amount that libFuzzer/x64/ASAN should be run is lower when there are jobs like AFL/x64/ASAN and libFuzzer/x64/ASAN since they will find many of the same bugs. Same is true for other libFuzzer jobs (we get similar mutations, corpus elements from LF/UBSAN as we do from LF/ASAN so perhaps AFL/ASAN should be run more frequently than LF/ASAN).

> What do you mean re non-blackbox vs blackbox jobs? Currently we assign a default of 10 to those, but I expect that we need to set most of them manually.

Sorry the question didn't make much sense. I'm asking if this CL will affect the relative weighting of coverage-guided to blackbox fuzzing, ie: will libFuzzer/AFL get run a lot more/less compared than it is now.

oliverchang:
Changed to use multipliers instead. /gcbrun

PR #928, opened 2019-09-02
Move clearing of common data bundles directory to its own function.

inferno-chromium:
clear_testcase_directories is called before start of every task in cleanup_task_state,
so it clears the downloaded web tests bundle everytime. This is
a regression from #896.
Also, add a test for persistent_cache's clear_values function
with clear_all=True. This is not related to fix, but good to have.

### DIFF
--- a/src/python/bot/tasks/update_task.py
+++ b/src/python/bot/tasks/update_task.py
@@ -42,8 +42,8 @@ from system import environment
 from system import process_handler
 from system import shell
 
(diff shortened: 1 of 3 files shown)


inferno-chromium:
Basically fuzzer-common-data-bundles directory is empty on every run, and broken everywhere still :(

inferno-chromium:
TBR simple regression fix.

oliverchang:
Ah, thanks for fixing!

PR #930, opened 2019-09-03
Relax APP_PATH checks when APP_NAME isn't set.

oliverchang:
This can happen in the case of greybox fuzzing in the new refactored
engine fuzzing path.


oliverchang:
TBR to fix some exceptions

PR #931, opened 2019-09-03
Speculative fix for untrusted exception.

oliverchang:
Convert protobuf sequences to python lists when receiving responses from
worker. Otherwise things may not work as expected on these values (e.g.
JSON serialisation).


oliverchang:
TBR

PR #935, opened 2019-09-03
Fix exception to read /system/build.prop for Android.

inferno-chromium:
Reproduce tool calls configure_system_build_properties to
read /system/build.prop, but device is not running as root
when flashing reimaged. Make sure it is root, otherwise
file read fails.


inferno-chromium:
TBR minor fix.

Issue #621, opened 2019-06-28
Run fork mode more frequently (always?) on multi-core machines.


jonathanmetzman:
Our Windows bots have 2 cores each.
It looks like our Mac bots have 8.
I think fork mode probably provides a big performance boost on these platforms. We should look into doing it more frequently for them.

This theory could be checked by filtering by platform on the strategy performance graphs.

cc @mbarbella-chromium @mukundv-chrome

inferno-chromium:
Yes we should just enable fork mode by default on multi-cpu machines. And then also fix time_executed to be time_executed * no_of_hours. maybe then change stat from total_fuzzing_time_hrs to cpu_hours.

PR #939, opened 2019-09-04
Show total fuzzing time instead of runs count in stats.

### LINKED ISSUE #937: Show time executed / cpu hours instead of number of runs in fuzzer stats.

inferno-chromium:
Fixes #937.

### DIFF
--- a/src/local/butler/scripts/setup.py
+++ b/src/local/butler/scripts/setup.py
@@ -135,7 +135,7 @@ leak_percent: "Percent of fuzzing runs that crashed on memory leaks (should be 0
 timeout_percent: "Percent of fuzzing runs that had testcases timeout (should be 0)"
 startup_crash_percent: "Percent of fuzzing runs that crashed on startup (should be 0)"
 avg_unwanted_log_lines: "Average number of unwanted log lines in fuzzing runs (should be 0)"
-runs_count: "Number of libFuzzer fuzzing runs or sessions (default is ~1 hr)"
+total_fuzzing_time_hrs: "Total time in hours for which the fuzzer(s) ran. Will be lower if fuzzer hits a crash frequently."
 logs: "Link to fuzzing logs"
 corpus_backup: "Backup copy of the minimized corpus generated based on code coverage"'''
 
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
Ping to do this in OSS-Fuzz as well.

jonathanmetzman:
Why not name this cpu_hours? I think it conveys the same thing without taking up precious screen real estate

jonathanmetzman:
Actually, it seems done already in oss-fuzz. I think the old version is still showing up because of caching (compare https://oss-fuzz.com/fuzzer-stats/by-fuzzer/date-start/2019-08-29/date-end/2019-09-04/fuzzer/libFuzzer/job/libfuzzer_asan_aspell to https://oss-fuzz.com/fuzzer-stats/by-day/date-start/2019-08-29/date-end/2019-09-04/fuzzer/libFuzzer/job/libfuzzer_asan_aspell)

inferno-chromium:
> Why not name this cpu_hours? I think it conveys the same thing without taking up precious screen real estate

We can change stat name for OSS-Fuzz (in db), but for general clusterfuzz case, it seems wrong to say cpu hours, without actually accounting for number of cpus that it is run on. we should probably enable fork mode always - https://github.com/google/clusterfuzz/issues/621, then change timeexecuted=timeexecuted*no of cpus. Unsure, how can we do this across every fuzzing engine.

inferno-chromium:
> Actually, it seems done already in oss-fuzz. I think the old version is still showing up because of caching (compare https://oss-fuzz.com/fuzzer-stats/by-fuzzer/date-start/2019-08-29/date-end/2019-09-04/fuzzer/libFuzzer/job/libfuzzer_asan_aspell to https://oss-fuzz.com/fuzzer-stats/by-day/date-start/2019-08-29/date-end/2019-09-04/fuzzer/libFuzzer/job/libfuzzer_asan_aspell). Good change!

Yes if entity already exists, it needs to be manually updated everywhere. i did the change on all db across all project types. if we want to remove old results, we can clear memcache, but i feel it was ok.

jonathanmetzman:
> > Why not name this cpu_hours? I think it conveys the same thing without taking up precious screen real estate
> 
> We can change stat name for OSS-Fuzz (in db), but for general clusterfuzz case, it seems wrong to say cpu hours, without actually accounting for number of cpus that it is run on. we should probably enable fork mode always - #621, then change timeexecuted=timeexecuted*no of cpus. Unsure, how can we do this across every fuzzing engine.

Ah very good point!

PR #941, opened 2019-09-04
Trial new fuzzing engine pipeline for libFuzzer.

oliverchang:
20% chance of running during fuzzing.
0% chance of running on other tasks.

### DIFF
--- a/src/python/bot/fuzzers/engine.py
+++ b/src/python/bot/fuzzers/engine.py
@@ -21,7 +21,7 @@ from system import environment
 
 _ENGINES = {}
 _TRIAL_PROBABILITY_FUZZ = 0.2
-_TRIAL_PROBABILITY_OTHERS = 0.05
+_TRIAL_PROBABILITY_OTHERS = 0.0
 
 
 class FuzzOptions(object):
--- a/src/python/bot/fuzzers/engine.py
+++ b/src/python/bot/fuzzers/engine.py
@@ -15,8 +15,13 @@
 
 from builtins import object
 from collections import namedtuple
+import random
+
+from system import environment
 
 _ENGINES = {}
+_TRIAL_PROBABILITY_FUZZ = 0.2
+_TRIAL_PROBABILITY_OTHERS = 0.05
 
 
 class FuzzOptions(object):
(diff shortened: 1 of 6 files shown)


src/python/bot/fuzzers/engine.py
```diff
@@ -15,8 +15,13 @@
 
 from builtins import object
 from collections import namedtuple
+import random
+
+from system import environment
 
 _ENGINES = {}
+_TRIAL_PROBABILITY_FUZZ = 0.2
+_TRIAL_PROBABILITY_OTHERS = 0.05
```
inferno-chromium: Might be better to start with 0 first, fix the fuzzing failures first. breaking existing open testcases is very painful, maybe just use USE_NEW_ENGINE_IMPL on clusterfuzz.corp where changes are not visible atm and we will be refiling bugs later ?

src/python/bot/fuzzers/engine.py
```diff
@@ -172,3 +177,16 @@ def get(name):
     return engine_class()
 
   return None
+
+
+def do_trial():
+  """Returns or not we should trial using this new implementation."""
+  if environment.get_value('USE_NEW_ENGINE_IMPL'):
```
inferno-chromium: Shouldn't USE_NEW_ENGINE_IMPL be used together with trial ? Is there a place you are enabling 100% ? Can also try on clusterfuzz-aarya-test project?

src/python/bot/fuzzers/engine.py
oliverchang: Sure, I already tested this with some test jobs, but will set to 0 for now to be extra careful.

src/python/bot/fuzzers/engine.py
oliverchang: Yes, I will be enabling this for a few jobs manually to better test things.

oliverchang: (To better explain my comment).

For all jobs, we will have a 20% chance of fuzzing. For certain jobs, I want to enable this completely, including reproduction etc

inferno-chromium: SGTM.

PR #942, opened 2019-09-04
Encode/decode unicode output for EngineFuzz rpc.

oliverchang:
/gcbrun


oliverchang:
TBR

src/python/bot/untrusted_runner/tasks_host.py
```diff
@@ -138,7 +138,7 @@ def engine_fuzz(engine_impl, target_name, sync_corpus_directory,
   crashes = [
       engine.Crash(
           input_path=file_host.rebase_to_host_root(crash.input_path),
-          stacktrace=crash.stacktrace,
+          stacktrace=utils.decode_to_unicode(crash.stacktrace),
```
inferno-chromium: We have except clause when encoding fails, is that ok here ?

    except:
      obj = unicode(''.join(char for char in obj if ord(char) < 128), encoding)

PR #944, opened 2019-09-04
Add automatic assignee feature for issue filing.

inferno-chromium:


inferno-chromium:
fyi, this is needed by android.

PR #947, opened 2019-09-05
Add a temporary use_minijail override to libfuzzer.get_runner.

oliverchang:
The new engine implementation does not support minijail.


oliverchang:
TBR. /gcbrun

PR #949, opened 2019-09-05
[fuchsia] Add run_single_testcase to enable reproducibility.

flowerhack:
Reproduction works by running run_single_testcase via the runner.

run_single_testcase expects to get the testcase from some location on
host, so we add a bit of extra logic for "push testcase from host onto
the VM and run it there".

### DIFF
--- a/src/python/bot/fuzzers/libfuzzer.py
+++ b/src/python/bot/fuzzers/libfuzzer.py
@@ -436,8 +436,25 @@ class FuchsiaQemuLibFuzzerRunner(new_process.ProcessRunner, LibFuzzerCommon):
                           testcase_path,
                           timeout=None,
                           additional_args=None):
-    # TODO(flowerhack): Fill out this command.
-    pass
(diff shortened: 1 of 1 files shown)


src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -436,8 +436,25 @@ def run_single_testcase(self,
...
-    # TODO(flowerhack): Fill out this command.
-    pass
+    # TODO(flowerhack): a lot of this code is repetitive with `fuzz`; once we
+    # transition to the new fuzzing interface, a lot of this should be
+    # factored out.
+    # Copy the testcase from the host to the target.
+    testcase_path_name = os.path.basename(os.path.normpath(testcase_path))
+    self.device.store(testcase_path, self.fuzzer.data_path())
+    # Run a single round, using the testcase we just pushed to target.
+    self.fuzzer.start([self.fuzzer.data_path(testcase_path_name)])
```
oliverchang: as discussed, you need to pipe additional_args into the fuzzer CLI. You also need to fix this in fuzz().

PR #950, opened 2019-09-05
Exception fixes.

oliverchang:
- Handle merge errors during fuzzing gracefully.
- Sync corpora to the right place.
- Log when we're trialing the new engine fuzzing codepath.


oliverchang:
/gcbrun. TBR

PR #953, opened 2019-09-05
Fix issue status overwrite issue with automatic assignee.

inferno-chromium:


inferno-chromium:
Phew, finally it works! Can only set status once :(

PR #955, opened 2019-09-05
Fix merge error logging.

oliverchang:


src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -179,7 +179,7 @@ def _merge_new_units(self, target_path, corpus_dir, new_corpus_dir,
         stat_overrides.update(
             stats.parse_stats_from_merge_log(result.logs.splitlines()))
     except MergeError:
-      logs.log_error('Merge failed.')
+      logs.log_warn('Merge failed', target=os.path.basename(target_path))
```
inferno-chromium: Is target addition needed here ? Do we do the logs.configure with fuzzer_name, etc that we used to do in launcher.py before (maybe that can't be done anymore :() ?

oliverchang: We can't do that since it's no longer in a separate process.. will think more. Keeping this for now.

PR #956, opened 2019-09-06
Fix incorrect logging in testcase_manager.

oliverchang:


src/python/bot/testcase_manager.py
```diff
@@ -606,13 +606,14 @@ def _pre_run_cleanup(self):
...
     """Get crash state from a CrashResult."""
     state = crash_result.get_symbolized_data()
-    logs.log(
-        ('Crash occurred in {crash_time} seconds (round {round_number}). '
-         'State:\n{crash_state}').format(
-             crash_time=crash_result.crash_time,
-             round_number=round_number,
-             crash_state=state.crash_state),
-        output=state.crash_stacktrace)
+    if state.is_crash():
```
inferno-chromium: This should be crash_result.is_crash()
and state needs to be moved inside if

oliverchang: oops, fixed. we need to return state, so state will remain outside the if.

PR #957, opened 2019-09-06
Push minijail support down to libFuzzer runner.

oliverchang:
We still need to support minijail for ChromeOS. This CL pushes some of
the special cases for supporting it to the libfuzzer runner classes.

Some of the cleanup here will also make it easier for Fuchsia,

Summary of changes:
- Remove minimize/cleanse functionality from launchers.
- Automatically bind corpus dirs in MinijailLibFuzzerRunner
- Introduce artifact_prefix args for minimize/cleanse to make it easier
  to override.
- Add a common function for getting the testcase path from logs.

### DIFF
--- a/src/python/bot/fuzzers/libfuzzer.py
+++ b/src/python/bot/fuzzers/libfuzzer.py
@@ -35,6 +35,10 @@ from system import shell
 
 MAX_OUTPUT_LEN = 1 * 1024 * 1024  # 1 MB
 
+# Regex to find testcase path from a crash.
+CRASH_TESTCASE_REGEX = (r'.*Test unit written to\s*'
+                        r'(.*(crash|oom|timeout|leak)-.*)')
+
 
 class LibFuzzerException(Exception):
   """LibFuzzer exception."""
@@ -60,6 +64,15 @@ class LibFuzzerCommon(object):
       return artifact_prefix
 
     return artifact_prefix + sep
+
+  def get_testcase_path(self, log_lines):
+    """Get testcase path from log lines."""
+    for line in log_lines:
+      match = re.match(CRASH_TESTCASE_REGEX, line)
+      if match:
+        return match.group(1)
+
+    return None
 
   def analyze_dictionary(self,
                          dictionary_path,
@@ -238,6 +251,7 @@ class LibFuzzerCommon(object):
                      testcase_path,
                      output_path,
                      timeout,
+                     artifact_prefix=None,
                      additional_args=None):
     """Minimize crasher with libFuzzer.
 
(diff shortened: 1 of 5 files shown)


inferno-chromium:
@jonathanmetzman - can you please help to review, this is to make chromeos minijail jobs to work with new engine impl.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -262,8 +276,12 @@ def minimize_crash(self,
                                         max_total_time)
 
     additional_args.extend([
-        constants.MINIMIZE_CRASH_ARGUMENT, max_total_time_argument,
-        constants.EXACT_ARTIFACT_PATH_FLAG + output_path, testcase_path
+        constants.MINIMIZE_CRASH_ARGUMENT,
+        max_total_time_argument,
+        constants.EXACT_ARTIFACT_PATH_FLAG + output_path,
+        constants.ARTIFACT_PREFIX_FLAG +
+        self._normalize_artifact_prefix(artifact_prefix),
```
inferno-chromium: What if artifact_prefix is None, it will cause exception i think seeing the first line in _normalize_artifact_prefix (checks endswith).

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -476,6 +497,14 @@ def __init__(self, executable_path, chroot, default_args=None):
         chroot=chroot,
         default_args=default_args)
 
+  def get_testcase_path(self, log_lines):
+    """Get testcase path from log lines."""
+    path = LibFuzzerCommon.get_testcase_path(self, log_lines)
+    if not path:
+      return path
+
+    return os.path.join(self.chroot.directory, path[1:])
```
inferno-chromium: instead of path[1:], can we explicitly do lstrip(os.sep), seems more clear ?

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -502,6 +531,19 @@ def _get_chroot_directory(self, directory_path):
...
+  def _bind_corpus_dirs(self, corpus_directories):
+    """Bind corpus directories to the minijail chroot.
+
+    Also makes sure that the directories are world writeable.
+
+    Args:
+      corpus_directories: A list of corpus paths.
+    """
+    for corpus_directory in corpus_directories:
+      target_dir = '/' + os.path.basename(corpus_directory)
```
inferno-chromium: Is this linux only code, how about making it in cross-platform ?

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -502,6 +531,19 @@ def _get_chroot_directory(self, directory_path):
...
+      self.chroot.add_binding(
+          minijail.ChrootBinding(corpus_directory, target_dir, True))
```
inferno-chromium: nit: named arg for True

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -526,9 +568,24 @@ def fuzz(self,
...
+        self,
+        corpus_directories,
+        fuzz_timeout,
+        artifact_prefix='/',
+        additional_args=additional_args,
+        extra_env=extra_env)
+
+    log_lines = result.output.splitlines()
+    for line in log_lines:
+      crash_location_regex.match(line)
```
inferno-chromium: What is this line doing with the match ?

PR #963, opened 2019-09-10
LibFuzzerEngine: Pass artifact_prefix as argument to LibFuzzerRunner.

oliverchang:
Instead of constructing our own argument. LibFuzzerRunner will
transparently bind things for minijail to work.


oliverchang:
TBR

PR #965, opened 2019-09-10
[Reproduce tool] Varied cleanup.

mbarbella-chromium:
- Show warnings for flaky stacks/non-reproducible test cases.
- Improve authentication prompt.
- Add an argument to enable verbose logging and disable it by default.
- Add an option to skip Android device setup.
- Improve several prompts and log messages.
- Provide faster feedback by showing output after the initial run.

### DIFF
--- a/src/local/butler/reproduce.py
+++ b/src/local/butler/reproduce.py
@@ -232,11 +232,11 @@ def _prepare_initial_environment(build_directory, iterations, verbose):
   environment.set_value('BUILDS_DIR', build_directory)
 
   # Some functionality must be disabled when running the tool.
-  environment.set_value('REPRODUCE_TOOL', 'True')
+  environment.set_value('REPRODUCE_TOOL', True)
 
   # Force logging to console for this process and child processes.
   if verbose:
-    environment.set_value('LOG_TO_CONSOLE', 'True')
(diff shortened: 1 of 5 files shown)


src/local/butler/reproduce.py
```diff
@@ -236,7 +235,8 @@ def _update_directory(directory_name, ignore_paths=None):
   environment.set_value('REPRODUCE_TOOL', 'True')
 
   # Force logging to console for this process and child processes.
-  environment.set_value('LOG_TO_CONSOLE', 'True')
+  if verbose:
+    environment.set_value('LOG_TO_CONSOLE', 'True')
```
mbarbella-chromium: I'm not sure these log messages make sense to the average user. I tried to improve the output when running a bit to give the user faster feedback and set expectations that repeated runs will take a while.

mbarbella-chromium:
PTAL when you have time. This is a collection of small, unrelated changes intended to make the reproduce tool more user friendly.

src/local/butler/reproduce.py
inferno-chromium: s/'True'/True in all places.

mbarbella-chromium: Done!

Issue #882, opened 2019-08-23
Don't say "No variants found" before looking


jonathanmetzman:
In the variant analysis section of testcases, it says: "No variants found." before any variants are looked for.
Instead, it should say "Variant analysis not started" until variants are tested. Then if none are found it can say  "No variants found."

PR #966, opened 2019-09-10
Show correct status message for variants.

inferno-chromium:
Fixes https://github.com/google/clusterfuzz/issues/882

### DIFF
--- a/src/appengine/handlers/testcase_detail/testcase_variants.py
+++ b/src/appengine/handlers/testcase_detail/testcase_variants.py
@@ -54,12 +54,17 @@ class Handler(base_handler.Handler):
       }
       if not is_pending:
         item.update({
-            'revision': variant.revision,
-            'crashType': variant.crash_type,
+            'revision':
+                variant.revision,
+            'crashType':
+                variant.crash_type,
             'crashStateLines': (variant.crash_state or '').strip().splitlines(),
-            'securityFlag': variant.security_flag,
-            'isSimilar': variant.is_similar,
-            'reproducerKey': variant.reproducer_key,
+            'securityFlag':
+                variant.security_flag,
+            'isSimilar':
+                variant.is_similar,
+            'reproducerKey':
+                variant.reproducer_key,
         })
       items.append(item)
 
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
Thanks! Will take a look

inferno-chromium:
This is slightly different from what was proposed, we don't show the table at all when it is pending. I think it feels ok, since no need to send the ajax request.

src/appengine/private/components/testcase-detail/testcase-detail.html
```diff
@@ -727,6 +729,10 @@
         return this.isStatusReady(info) &&
             !this.isNA(info.testcase.minimized_keys);
       },
+      shouldShowVariants(info) {
+        return !info.testcase.one_time_crasher_flag &&
+            info.testcase.minimized_keys;
```
jonathanmetzman: Should this be `&& !info.testcase.minimized_keys`?
I assume this code should be returning false if no minimized testcase is available right?
What if minimization errors out?

inferno-chromium: if minimization fails, it calls mark_unreproducible_if_flaky, which sets one_time_crasher_flag to true.
and it should be "info.testcase.minimized_keys" since variant pending task are only created after minimization is finished.

jonathanmetzman: ACK

inferno-chromium:
> LGTM.
> 
> I sort of think it's better to say "not available" rather than show nothing but I guess this is fine.

Rethought on this. i agree with you, so redid cl, can you please take another look @jonathanmetzman

src/appengine/handlers/testcase_detail/testcase_variants.py
```diff
@@ -54,12 +54,17 @@ def _display_status(status):
       }
       if not is_pending:
         item.update({
-            'revision': variant.revision,
-            'crashType': variant.crash_type,
+            'revision':
+                variant.revision,
+            'crashType':
+                variant.crash_type,
             'crashStateLines': (variant.crash_state or '').strip().splitlines(),
```
jonathanmetzman: Line 59 now seems inconsistent with the other lines.
Why change them?

inferno-chromium: Weird pyformat thing, can look later why.

PR #968, opened 2019-09-11
Speculative fix for another unicode exception.

oliverchang:


oliverchang:
> Getting worried that every engine integration will have to do these decode_to_unicode hacks, sounds pain! can there can be common pipeline to process output that returns it decoded with decode_to_unicode

Good point. Let me merge this first and think more on these unicode issues..

PR #969, opened 2019-09-11
Add BLAZE_TEST_ARGS.

inferno-chromium:
--test_arg and --test_env require each option to be specified
one at a time. This also works on Bazel.


src/python/tests/core/datastore/data_handler_test.py
```diff
@@ -681,3 +681,41 @@ def test_blackbox_fuzzer_testcase_with_default_help_format(self):
          'ASAN_OPTIONS="handle_abort=1:redzone=512" '
          'UBSAN_OPTIONS="halt_on_error=1" ./binary'
         ).format(id=testcase.key.id()))
+
+  def test_blaze_test_args(self):
+    """Test the function with a blackbox fuzzer test case, with HELP_FORMAT
```
Dor1s: nit: the comment doesn't seem fully correct, it's not "blackbox" case

inferno-chromium: Done, good catch!

PR #972, opened 2019-09-13
[fuchsia] Enables reproducible crashes for Fuchsia testcases.

flowerhack:
To detect crashes from reproducer runs, we need to (1) properly return
the return code from the fuzz run, and (2) check for some text that only
occurs in reproducer runs.  This CL tweaks some of the code in
util/fuchsia to propertly return this information, and tweaks the way
that code is called, accordingly.

In addition, the way Fuchsia dumps logs by default means that the
stacktrace occurs *before* the `==ERROR: .*` line, which confuses the
stack analyzer, so we tweak the log processing subroutine to drop that
line at the beginning.

Finally, we re-add in some code that was accidentally removed, which
suppresses stdout during fuzzing runs (quiet=True), since Clusterfuzz
can prematurely halt jobs that appear to have crashed, before we've
finished processing all the Fuchsia logs.

And, finally-finally, there's an actual implementation of
run_single_testcase in the Fuchsia runner here :)

### DIFF
--- a/src/python/bot/fuzzers/libfuzzer.py
+++ b/src/python/bot/fuzzers/libfuzzer.py
@@ -443,6 +443,7 @@ class FuchsiaQemuLibFuzzerRunner(new_process.ProcessRunner, LibFuzzerCommon):
     """LibFuzzerCommon.fuzz override."""
     self._test_qemu_ssh()
 
+    #TODO(flowerhack): Pass libfuzzer args (additional_args) here
     return_code = self.fuzzer.start([])
     self.fuzzer.monitor(return_code)
     self.fetch_and_process_logs_and_crash()
@@ -468,6 +469,7 @@ class FuchsiaQemuLibFuzzerRunner(new_process.ProcessRunner, LibFuzzerCommon):
     testcase_path_name = os.path.basename(os.path.normpath(testcase_path))
     self.device.store(testcase_path, self.fuzzer.data_path())
 
+    # TODO(flowerhack): Pass libfuzzer args (additional_args) here
     return_code = self.fuzzer.start(['repro', 'data/' + testcase_path_name])
     self.fuzzer.monitor(return_code)
     self.fetch_and_process_logs_and_crash()
@@ -481,6 +483,14 @@ class FuchsiaQemuLibFuzzerRunner(new_process.ProcessRunner, LibFuzzerCommon):
     fuzzer_process_result.time_executed = 0
     fuzzer_process_result.command = self.fuzzer.last_fuzz_cmd
     return fuzzer_process_result
+
+  def minimize_crash(self,
+                     testcase_path,
+                     output_path,
+                     timeout,
+                     artifact_prefix=None,
+                     additional_args=None):
+    return new_process.ProcessResult()
 
   def ssh_command(self, *args):
     return ['ssh'] + self.ssh_root + list(args)
--- a/src/python/bot/fuzzers/libfuzzer.py
+++ b/src/python/bot/fuzzers/libfuzzer.py
@@ -442,8 +442,9 @@ class FuchsiaQemuLibFuzzerRunner(new_process.ProcessRunner, LibFuzzerCommon):
            extra_env=None):
     """LibFuzzerCommon.fuzz override."""
     self._test_qemu_ssh()
-    self.fuzzer.start([])
-    self.fuzzer.monitor()
+
+    ret = self.fuzzer.start([])
+    self.fuzzer.monitor(ret)
     self.fetch_and_process_logs_and_crash()
 
     with open(self.fuzzer.logfile) as logfile:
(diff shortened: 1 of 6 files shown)


src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -415,8 +415,9 @@ def fuzz(self,
            extra_env=None):
     """LibFuzzerCommon.fuzz override."""
     self._test_qemu_ssh()
-    self.fuzzer.start([])
-    self.fuzzer.monitor()
+
+    ret = self.fuzzer.start([])
```
oliverchang: nit: rename ret to something more meaningful.

also, can we pass libfuzzer arguments (i.e. additional_args) here?

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -436,8 +437,23 @@ def run_single_testcase(self,
                           testcase_path,
                           timeout=None,
                           additional_args=None):
-    # TODO(flowerhack): Fill out this command.
-    pass
+    # We need to push the testcase to the device and pass in the name.
+    testcase_path_name = os.path.basename(os.path.normpath(testcase_path))
+    self.device.store(testcase_path, self.fuzzer.data_path())
+
+    ret = self.fuzzer.start(['repro', 'data/' + testcase_path_name])
```
oliverchang: can we pass additional_args here?

src/python/platforms/fuchsia/util/device.py
```diff
@@ -253,17 +253,24 @@ def process_logs(self, logfile, guess_pid=False):
...
 
-          # Check for a unit being dumped, i.e. a finding.
+          # Check for one of two things:
+          # 1) a unit being dumped (e.g. a finding from a regular fuzz run)
+          # 2) a nonzero return code plus a "Running: [foo]" message (which
+          # indicates this is a *reproducer* run that has successfully crashed)
+          repro_match = repro_pattern.search(line)
           match = mutation_pattern.search(line)
-          if match:
+          if match or (repro_match and retcode > 0):
```
oliverchang: can retcode be negative? maybe just `and retcode):` ?

src/python/platforms/fuchsia/util/device.py
```diff
@@ -253,17 +253,24 @@ def process_logs(self, logfile, guess_pid=False):
...
     artifact_pattern = re.compile(r'Test unit written to data/(\S*)')
+    repro_pattern = re.compile(r'Running: .*')
+    line_of_actual_crash = None
     with open(logfile) as log:
       with open(logfile + '.tmp', 'w') as tmp:
         for line in log:
           # Check for a line that tells us the process ID
           match = pid_pattern.search(line)
           if match:
+            line_with_crash_message = line
```
oliverchang: this hack is fine for now, but can this ordering be fixed on the device side?

src/python/platforms/fuchsia/util/device.py
```diff
@@ -282,7 +289,18 @@ def process_logs(self, logfile, guess_pid=False):
...
-    os.rename(logfile + '.tmp', logfile)
+
+    # Clusterfuzz's stack analyzer expects the
+    # `==[num]== ERROR: [SanitizerName]: [failure type]` line
+    # to occur *before* the stacktrace, so make a new tempfile
+    # where we insert that line at the top.
+    with open(logfile + '.tmp', 'r+') as tmp:
+      with open(logfile + '.tmp2', 'w') as tmp2:
+        tmp2.write(line_with_crash_message)
+        for line in tmp:
```
oliverchang: rather than iterate line by line here, we can just use `shutil.copyfileobj`

src/python/platforms/fuchsia/util/device.py
```diff
@@ -282,7 +289,18 @@ def process_logs(self, logfile, guess_pid=False):
 
           # Echo the line
           tmp.write(line)
-    os.rename(logfile + '.tmp', logfile)
+
+    # Clusterfuzz's stack analyzer expects the
+    # `==[num]== ERROR: [SanitizerName]: [failure type]` line
+    # to occur *before* the stacktrace, so make a new tempfile
+    # where we insert that line at the top.
+    with open(logfile + '.tmp', 'r+') as tmp:
```
oliverchang: why r+ ?

src/python/platforms/fuchsia/util/device.py
```diff
@@ -282,7 +289,18 @@ def process_logs(self, logfile, guess_pid=False):
...
+      with open(logfile + '.tmp2', 'w') as tmp2:
```
oliverchang: nit: maybe rename .tmp2 to .final or something more meaningful.

src/python/platforms/fuchsia/util/device.py
flowerhack: i don't think this works, though?  shutil.copyfileobj clobbers the whole file with the source, and we need that first line to stay there... if there's a better way to append one file to another, though, while still keeping that first line there, i'm all ears

src/python/platforms/fuchsia/util/device.py
flowerhack: no reason :) fixed

src/python/platforms/fuchsia/util/device.py
flowerhack: yeah, the plan is to patch up compiler-rt so that Fuchsia emits things the proper way in the first place.  i added a TODO a few lines down so this isn't lost, but rest assured we're working on a real fix!

src/python/platforms/fuchsia/util/device.py
flowerhack: good point, fixed

src/python/platforms/fuchsia/util/device.py
oliverchang: I believe shutil will still work here, as it'll respect the current position (which has advanced because of your `write()` call).

flowerhack: good point, fixed

src/python/bot/fuzzers/libfuzzer.py
flowerhack: renamed to return_code

i added a todo for the libfuzzer arguments, if you don't mind?  it's just that the testing of this has been delicate / finnicky enough that i'm really nervous about adding more stuff that may-or-may-not cause issues, and i'd rather isolate those in a different PR

src/python/bot/fuzzers/libfuzzer.py
flowerhack: renamed

flowerhack:
also: added a test for the stack_analyzer in our repro case

PR #976, opened 2019-09-13
[Reproduce tool] Fix incorrect one-time crasher warning.

mbarbella-chromium:
Minor bug fix for a flipped condition (causing an incorrect warning prompt).


mbarbella-chromium:
TBRing since this should be noncontroversial.

PR #981, opened 2019-09-16
Read APP_ARGS in TestcaseRunner.

oliverchang:


oliverchang:
TBR

src/python/bot/testcase_manager.py
```diff
@@ -548,8 +548,14 @@ def __init__(self,
       self._is_black_box = False
       self._engine_impl = engine_impl
 
-      # Read target_name + args from flags file.
-      arguments = get_additional_command_line_flags(testcase_path)
+      # Read target_name + args.
+      arguments = environment.get_value('APP_ARGS')
```
inferno-chromium: LGTM

Can create a helper for this in data_handler and use it in this place and 
https://github.com/google/clusterfuzz/blob/master/src/python/bot/tasks/fuzz_task.py#L126
Feel free to TBR if you agree.

inferno-chromium: The else part is broken too i think

    else:
      self._is_black_box = True	      self._is_black_box = True
      self._command = get_command_line_for_application(	      self._command = get_command_line_for_application(
          testcase_path, needs_http=needs_http)

https://github.com/google/clusterfuzz/blob/master/src/python/bot/tasks/minimize_task.py#L291

PR #983, opened 2019-09-16
Properly pipe arguments through engine fuzzing to testcase_manager.

oliverchang:


oliverchang:
TBR to fix things. will think on a more general/better way to do this.

PR #990, opened 2019-09-16
[Python 2->3 migration] Run final futurize checks on src/appengine.

mbarbella-chromium:


src/appengine/handlers/cron/project_setup.py
```diff
@@ -819,7 +820,7 @@ def _sync_job(self, project, info, corpus_bucket_name, quarantine_bucket_name,
         for key, value in sorted(six.iteritems(additional_vars)):
           job.environment_string += ('{} = {}\n'.format(
               key,
-              str(value).encode('unicode-escape')))
+              str(value).encode('unicode-escape').decode('utf-8')))
```
mbarbella-chromium: In python3 the old version would return a bytes, so when formatting we'll always end up with b'<whatever literal we had before>'. I think this preserves the behaviors we want, and makes test pass again.

mbarbella-chromium:
Not quite ready to land this, I want to skim the files where builtins imports were added to make sure nothing looks too out of place before landing.

Oliver, could you take a look at the project_setup.py change? That's the only manual bit of cleanup.

mbarbella-chromium:
Apparently in production App Engine has an explicit type check for str on any values passed as headers, so this might create a bit of a minefield. So far I've seen issues with the places where we set headers directly and call Handler.redirect (though luckily the latter are all funneled through one call site).

Still trying to get these issues resolved, but each of these create larger chunks of the code that we can't test before we switch to Python 3 for real. I don't see any way around the explicit type check other than not using newstr. This is only affecting a small number of files so we should still get a good amount of testing for the new behavior, and luckily this shouldn't be a true positive issue when we switch to Python 3 since the type will be correct after we transition.

mbarbella-chromium:
Seems to be working correctly as best I can tell. I'm sure I missed some forms and the cron handlers, but hopefully test coverage was sufficient there. If anything breaks feel free to revert this, but I'll be keeping an eye out for errors and trying to fix any simple ones in follow-up changes.

PR #994, opened 2019-09-17
Fix exception in fuchsia log processing.

oliverchang:
We were trying to write `line_with_crash_message` even in cases where
there is no crash.

Also simplify logic a bit and get rid of an intermediate file.


oliverchang:
TBR. /gcbrun

PR #995, opened 2019-09-17
Properly take issue_body_footer into account.

### LINKED ISSUE #993: Initialize issue_body_footer when reading issue tracker config
It seems like we forgot to add initialization of `issue_body_footer` in https://github.com/google/clusterfuzz/blob/dfe686cde7429ed0ba482b0b0b4f27106b0a0d5f/src/appengine/libs/issue_management/issue_tracker_policy.py#L114, that's why the issues filed recently do not have the text explaining that OSS-Fuzz team doesn't read bug tracked comments and can be contacted via GitHub only.

Bug example: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=17216

oliverchang:
Fixes #993.


oliverchang:
TBRing

PR #996, opened 2019-09-17
In upload testcase, when quiet flag is set, don't send emails.

inferno-chromium:


inferno-chromium:
This helps Jose to avoid sending emails when a mass upload is done.

PR #997, opened 2019-09-17
Add debug logging for the argv passed to libFuzzer launcher.

Dor1s:
temporary measure just to understand what's wrong with the arguments order


Dor1s:
Filed #999 and closing this one as per chat with Abhishek.

PR #1002, opened 2019-09-17
Revert "[Python 2->3 migration] Run final futurize checks on src/appengine."

mbarbella-chromium:
Reverts google/clusterfuzz#990


mbarbella-chromium:
TBR, bypassing checks for a recent revert.

PR #1005, opened 2019-09-18
[fuchsia] Restart QEMU in event of connection failure.

flowerhack:
This is a (hopefully) temporary, (hopeful) fix for an issue we've
observed when running CF in production (though not when running
locally):

Locally, you can run one QEMU instance and run your fuzzing &
reproduction jobs on it in sequence.

In production, however, the QEMU instance seems to die sometime after
the first fuzzing run (and thus attempts at reproduction time out).

Update the logic s.t. _test_qemu_ssh failures lead to tearing down QEMU,
restarting it, and trying one last time to make a connection work.

### DIFF
--- a/src/python/bot/fuzzers/libfuzzer.py
+++ b/src/python/bot/fuzzers/libfuzzer.py
@@ -449,16 +449,9 @@ class FuchsiaQemuLibFuzzerRunner(new_process.ProcessRunner, LibFuzzerCommon):
     os.remove(self.fuzzer.logfile)
     shutil.move(new_file_handle_path, self.fuzzer.logfile)
 
-  def fuzz(self,
-           corpus_directories,
-           fuzz_timeout,
-           artifact_prefix=None,
-           additional_args=None,
-           extra_env=None):
-    """LibFuzzerCommon.fuzz override."""
-
-    # Since we're seeing Fuchsia connection errors:
-    # - Test the connection.  If this works, proceed.
+  def _test_ssh(self):
+    """Test the ssh connection."""
+    # Test the connection.  If this works, proceed.
     # - If we fail, restart QEMU and test the connection again.
     # - If that fails, throw the error; we can't seem to recover.
     try:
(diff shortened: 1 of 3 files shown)


src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -441,7 +456,16 @@ def fuzz(self,
            additional_args=None,
            extra_env=None):
     """LibFuzzerCommon.fuzz override."""
-    self._test_qemu_ssh()
+
+    # Since we're seeing Fuchsia connection errors:
+    # - Test the connection.  If this works, proceed.
+    # - If we fail, restart QEMU and test the connection again.
+    # - If that fails, throw the error; we can't seem to recover.
+    try:
```
oliverchang: nit: put this in a common method for reuse

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -461,10 +485,28 @@ def fuzz(self,
     fuzzer_process_result.command = self.fuzzer.last_fuzz_cmd
     return fuzzer_process_result
 
+  def _restart_qemu(self):
```
oliverchang: nit: add a docstring

src/python/platforms/fuchsia/device.py
```diff
@@ -31,7 +31,7 @@
 from system import shell
 
 
-def qemu_setup():
+def setup_qemu_values(initial_setup=True):
```
oliverchang: nit: it seems like the common case is initial_setup=False, so maybe use that as the default?

oliverchang: actually, looking at this a bit more, a better way would be to factor out the initial setup into its own function, which we call from build_setup.

src/python/bot/fuzzers/libfuzzer.py
oliverchang: Done.

src/python/bot/fuzzers/libfuzzer.py
oliverchang: Done.

src/python/platforms/fuchsia/device.py
oliverchang: (i didn't do this -- please do this refactor in another PR if you agree)

oliverchang:
(I fixed some of the comments I made -- but there is one remaining: https://github.com/google/clusterfuzz/pull/1005#discussion_r325464750)

PR #1006, opened 2019-09-18
Reenable KVM for fuchsia

oliverchang:
Fix a bug from before where we weren't passing the right CPU argument.


oliverchang:
Verified that this seems to work on our bots (SSH port responds)

oliverchang:
(somehow these changes didn't get pushed before I commited: https://github.com/google/clusterfuzz/pull/1007)

PR #1008, opened 2019-09-18
Add a delay after starting fuchsia.

oliverchang:
Previously we were relying solely on the SSH expontential backoff, which
is not enough time, as we would wait a maximum of 2 + 4 + 8 = 14 seconds
for QEMU to get ready.

Also increase the exponential backoff for retrying SSH a little to use
powers of 3.


oliverchang:
TBR

PR #1011, opened 2019-09-18
Remove custom-crash-state hack and set it properly in stack analyzer.

inferno-chromium:
This stopped working for fuzz target timeout and oom with the
new engine impl. Removing this hack and making it properly work.


inferno-chromium:
The reason it broke in new engine impl because it was only in the fuzz() function and not in reproduce. But anyway, this is a crazy hack and this seems like the way it should have been done.

PR #1014, opened 2019-09-18
[Python 2->3] Reland running final futurize checks on src/appengine.

mbarbella-chromium:


src/appengine/handlers/base_handler.py
```diff
@@ -262,6 +264,22 @@ def redirect(self, url, **kwargs):
     check_redirect_url(url)
     super(Handler, self).redirect(url, **kwargs)
 
+  # TODO(mbarbella): Delete this once the Python 3 migration is complete.
```
mbarbella-chromium: This is the main change from the previous version. It post-processes all headers to make sure we won't leave any as newstr.

PR #1017, opened 2019-09-19
Respect artifact_prefix when writing fuchsia crash artifacts.

oliverchang:


oliverchang:
TBR. /gcbrun

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -441,13 +443,15 @@ def fetch_and_process_logs_and_crash(self):
             self.device.fetch(
                 self.fuzzer.data_path(crash_name), self.fuzzer.results_output())
```
oliverchang: @flowerhack We can simplify this by fetching directly into artifact_prefix, but I'm not sure if there are restrictions here in what we can pass as the second argument, so I've done things this way to be safe.

Please change this if we can indeed fetch directly into artifact_prefix

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -503,14 +507,13 @@ def run_single_testcase(self,
...
 
     # We need to push the testcase to the device and pass in the name.
-    self._test_qemu_ssh()
     testcase_path_name = os.path.basename(os.path.normpath(testcase_path))
     self.device.store(testcase_path, self.fuzzer.data_path())
 
     # TODO(flowerhack): Pass libfuzzer args (additional_args) here
     return_code = self.fuzzer.start(['repro', 'data/' + testcase_path_name])
     self.fuzzer.monitor(return_code)
-    self.fetch_and_process_logs_and_crash()
```
inferno-chromium: Why call this self.process_logs_and_crash() at all if it will bail out instantly on the first line ?

inferno-chromium: If we remove this, then can even remove the artifact_prefix check since we expect it to be always there.

oliverchang: yep I realised this afterwards :) will clean this up

src/python/bot/fuzzers/libfuzzer.py
flowerhack: Fetching directly into artifact_prefix should work fine for this function.

I'll put together a follow-up CL shortly to reflect this!

PR #1018, opened 2019-09-19
[fuchsia] Add integration test for reproduction, and pass additional

flowerhack:
libfuzzer arguments.

This is some follow-up / housekeeping work to make sure reproduction &
libfuzzer arguments work well for Fuchsia!


src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_launcher_integration_test.py
```diff
@@ -710,14 +710,22 @@ def tearDown(self):
       'Temporarily disabling the Fuchsia test until build size reduced.')
```
oliverchang: I can do this, but just a FYI that this test will be moving to engine_test.py

flowerhack: good to know, thanks!

PR #1021, opened 2019-09-20
[fuchsia] Refactor QEMU lifecycle code.

flowerhack:
Suggestions welcome, if anyone's got ideas for making this bit of code
more clear/obvious, but I think this at least reads more clearly than
what we had before.

QemuProcess handles all the work of creating, running, and killing QEMU
instances.  To interface properly with the VM (e.g. via SSH), you
should:

.create() a VM,
call _setup_qemu_device_and_fuzzer, which will use the env vars set in
.create() to configure a proper SSH command,
and then run() the VM.

This looks a bit more clean than the stuff we were passing around before
via various setup functions.

### DIFF
--- a/src/python/bot/fuzzers/libfuzzer.py
+++ b/src/python/bot/fuzzers/libfuzzer.py
@@ -26,9 +26,7 @@ from bot.fuzzers import engine_common
 from bot.fuzzers import utils as fuzzer_utils
 from bot.fuzzers.libFuzzer import constants
 from platforms import fuchsia
-from platforms.fuchsia.device import run_qemu_instance
-from platforms.fuchsia.device import setup_qemu_instance
-from platforms.fuchsia.device import setup_qemu_values
+from platforms.fuchsia.device import QemuProcess
 from platforms.fuchsia.util.device import Device
 from platforms.fuchsia.util.fuzzer import Fuzzer
 from platforms.fuchsia.util.host import Host
@@ -376,9 +374,16 @@ class FuchsiaQemuLibFuzzerRunner(new_process.ProcessRunner, LibFuzzerCommon):
 
   FUZZER_TEST_DATA_REL_PATH = os.path.join('test_data', 'fuzzing')
 
-  def _setup_fuzzer_and_device(self):
-    """ Build a Fuzzer object based on the QEMU values.
-    Call this only after setup_qemu_values()"""
+  def _setup_qemu_device_and_fuzzer(self):
+    """Initialize QEMU, then build a Device and Fuzzer object
+    based on its settings."""
+    if not self.qemu:
+      self.qemu = QemuProcess()
+
+    self.qemu.create()
+
+    # These environment variables are set when a QemuProcess is `create`ed.
+    # We need them in order to ssh / otherwise communicate with the VM.
     fuchsia_pkey_path = environment.get_value('FUCHSIA_PKEY_PATH')
     fuchsia_portnum = environment.get_value('FUCHSIA_PORTNUM')
     fuchsia_resources_dir = environment.get_value('FUCHSIA_RESOURCES_DIR')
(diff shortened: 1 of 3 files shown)


flowerhack:
@oliverchang

src/python/platforms/fuchsia/device.py
```diff
@@ -54,98 +53,140 @@ def setup_qemu_values(initial_setup=True):
...
+  (Calling it more than once per download will cause errors.)
+  This function does not run a VM, merely performs setup for a VM.
+  """
+  qemu_vars = _fetch_qemu_vars()
+  extend_fvm(qemu_vars['fuchsia_resources_dir'], qemu_vars['drive_path'])
+  add_keys_to_zbi(qemu_vars['fuchsia_resources_dir'], qemu_vars['initrd_path'],
+                  qemu_vars['fuchsia_zbi'])
+
+
+class QemuException(Exception):
```
oliverchang: nit: s/Exception/Error/ per style guide (http://google.github.io/styleguide/pyguide.html#244-decision)

src/python/platforms/fuchsia/device.py
```diff
@@ -54,98 +53,140 @@ def setup_qemu_values(initial_setup=True):
...
+  def run(self):
+    """Actually runs a QEMU VM, assuming `create` has already been called."""
+    if not self.process_runner:
+      raise QemuException('Attempted to `run` QEMU VM before calling `create`')
+    self.popen = self.process_runner.run(
+        stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    time.sleep(_QEMU_WAIT_SECONDS)
+
+  def kill(self):
+    if not self.popen:
```
oliverchang: nit: docstring. also maybe raise an exception instead?

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -400,20 +408,20 @@ def _setup_fuzzer_and_device(self):
...
     self.fuzzer = Fuzzer(
         self.device, package, target, output=test_data_dir, foreground=True)
 
   def __init__(self, executable_path, default_args=None):
-    qemu_path, qemu_args = setup_qemu_values(initial_setup=False)
-    qemu_process = setup_qemu_instance(qemu_path, qemu_args)
-    self._setup_fuzzer_and_device()
-    self.qemu_instance = run_qemu_instance(qemu_process)
+    self.qemu = None
+    self._setup_qemu_device_and_fuzzer()
```
oliverchang: I don't think we should keep running a new qemu instance every time we initialise the LibfuzzerRunner, since we create several in a several run, leading to killing/restarting every time we try do a reproduction round for instance. It's probably better to revert to starting the initial process once the build is setup, but keep the restart logic.

src/python/platforms/fuchsia/device.py
flowerhack: renamed

src/python/platforms/fuchsia/device.py
flowerhack: added a docstring.

presumably if someone's calling "kill", they just want the process good and dead, and we shouldn't throw an error if it happens to already be dead?  (i notice terminate_processes_matching_names just ignores psutil.NoSuchProcess errors, for example)

src/python/bot/fuzzers/libfuzzer.py
flowerhack: chatted offline; refactored so that this logic works cleanly (via starting process in build_manager.py and leaving furhter manipulation in libfuzzer.py)

flowerhack:
this should be ready for another look!

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -376,9 +375,17 @@ class FuchsiaQemuLibFuzzerRunner(new_process.ProcessRunner, LibFuzzerCommon):
 
   FUZZER_TEST_DATA_REL_PATH = os.path.join('test_data', 'fuzzing')
 
-  def _setup_fuzzer_and_device(self):
-    """ Build a Fuzzer object based on the QEMU values.
-    Call this only after setup_qemu_values()"""
+  def _setup_device_and_fuzzer(self, qemu_is_running=True):
```
oliverchang: instead of adding this condition here, just create another separate method which only starts qemu and use it in the places where you'd normally pass qemu_is_running=False

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -376,9 +375,17 @@ class FuchsiaQemuLibFuzzerRunner(new_process.ProcessRunner, LibFuzzerCommon):
...
+    """Build a Device and Fuzzer object based on QEMU's settings."""
+
+    # If QEMU isn't currently running, set that up first.
+    if not qemu_is_running:
+      qemu = QemuProcess()
```
oliverchang: can we just create a device.start_qemu() method which does these three things? There seems to be no reason to have callers call these 3 methods here rather than a single one.

also related to above comment, this can be removed completely.

src/python/platforms/fuchsia/device.py
oliverchang: sure, that seems reasonable.

PR #1022, opened 2019-09-20
[fuchsia] Simplify log-processing code.

flowerhack:
We only need to call process_logs_and_crash when *fuzzing* (as opposed
to reproducing), since the processing is used to determine how to pull
down the testcase once fuzzing is complete.  Reproduction simply has a
testcase_path passed in.

Additionally, we modify the call to device.fetch() to simply store the
result in artifact_prefix directly, rather than results_output.


flowerhack:
@oliverchang

PR #1025, opened 2019-09-20
Start removing code from launcher.py.

oliverchang:
- Fix mutator plugins in minijail.
- Fix engine exit code checking.
- Move remaining tests into engine_test, and enable minijail tests.
- Remove libfuzzer_launcher_integration_test, and many tests from
  libfuzzer_launcher_test.


src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -216,6 +223,13 @@ def fuzz(self, target_path, options, reproducers_dir, max_time):
         artifact_prefix=reproducers_dir,
         extra_env=options.extra_env)
 
+    if (not environment.get_value('USE_MINIJAIL') and
```
inferno-chromium: This means we don't get results for OSS-Fuzz, should try finding some workaround ?

src/python/tests/core/bot/fuzzers/libFuzzer/engine_test.py
```diff
@@ -345,6 +390,19 @@ def setUp(self):
...
+
+@test_utils.integration
+class IntegrationTests(BaseIntegrationTest):
+  """Base libFuzzer launcher tests."""
+
+  def setUp(self):
+    BaseIntegrationTest.setUp(self)
+    self.crash_dir = TEMP_DIR
+
+  def compare_arguments(self, target_path, arguments, corpora, actual):
```
inferno-chromium: corpora_or_testcase ?

src/python/bot/fuzzers/libFuzzer/engine.py
oliverchang: I'm not sure what the best workaround here. Longer term we'll move from minijail to something more lightweight like chroot which should make this issue disappear.. but in any case this is a limitation that has always existed.

src/python/tests/core/bot/fuzzers/libFuzzer/engine_test.py
oliverchang: Done!

PR #1028, opened 2019-09-20
Use correct sendgrid types when sending email.

mbarbella-chromium:
I'm not sure if the explicit str() conversions are necessary, but this fixes the exceptions when sending mail. Sendgrid's code has a bunch of isinstance checks against str, and if it doesn't find a str it seems to expect to find one of these types it defines.

### DIFF
--- a/src/appengine/libs/mail.py
+++ b/src/appengine/libs/mail.py
@@ -14,7 +14,10 @@
 """Helpers for sending mail."""
 from builtins import str
 from sendgrid import SendGridAPIClient
+from sendgrid.helpers.mail import From
+from sendgrid.helpers.mail import HtmlContent
 from sendgrid.helpers.mail import Mail
+from sendgrid.helpers.mail import To
 
 from config import db_config
 from metrics import logs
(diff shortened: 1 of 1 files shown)


mbarbella-chromium:
PTAL whenever you have time. I'll land this as TBR once the test runs finish.

src/appengine/libs/mail.py
```diff
@@ -33,10 +36,10 @@ def send(to_email, subject, html_content):
     return
 
   message = Mail(
-      from_email=str(from_email),
-      to_emails=str(to_email),
+      from_email=From(str(from_email)),
+      to_emails=To(str(to_email)),
       subject=subject,
```
inferno-chromium: Do we not need to do anything with subject ?

mbarbella-chromium: Good catch. This worked without it because all of the instances of newstr we actually used came from the explicit str conversions here, but if we passed one in we'd need to use sendgrid's Subject. Will fix this in another change.

PR #1030, opened 2019-09-23
Changed file read syntax to use with statement

farziengineer:


farziengineer:
@googlebot I signed it!

PR #1033, opened 2019-09-23
Use sendgrid's Subject type when sending email.

mbarbella-chromium:


src/appengine/libs/mail.py
```diff
@@ -38,7 +39,7 @@ def send(to_email, subject, html_content):
   message = Mail(
       from_email=From(str(from_email)),
       to_emails=To(str(to_email)),
-      subject=subject,
+      subject=Subject(subject),
```
inferno-chromium: Does it need to be str(subject) like others.

mbarbella-chromium: I'm actually not sure why we have the explicit conversions. Ideally we shouldn't use them anywhere, but I was afraid to remove them in the other cases since I wasn't sure if callers depended on this behavior.

PR #1045, opened 2019-09-26
[fuchsia] Add corpus syncing capabilities.

flowerhack:
Modify FuchsiaQemuLibFuzzerRunner:fuzz to push & pull corpora.

Add an integration test to make sure this happens.

Tech fuchsia/util/* scripts to ignore "Warning: Permanently added [address] to the list of known hosts" when running `ls`.

Tech fuchsia/util/fuzzer.py:start to accept custom corpus directories.

Add some logging for clarity.

Updated build to a new version.


src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -478,14 +474,30 @@ def fuzz(self,
     """LibFuzzerCommon.fuzz override."""
     self._test_ssh()
 
-    #TODO(flowerhack): Pass libfuzzer args (additional_args) here
+    logs.log('SSH works; sync corpus directories')
+    # Push corpus directories to the device.
+    data_dst = self.fuzzer.data_path('corpus')
+    for corpus in corpus_directories:
+      for corpfile in os.listdir(corpus):
+        self.fuzzer.device.store(os.path.join(corpus, corpfile), data_dst)
```
flowerhack: TODO before merging: teach `store` to push multiple files at once (e.g. pass a `-r` argument)

src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -334,6 +334,7 @@ def minimize_corpus(self, target_path, arguments, input_dirs, output_dir,
       A Result object.
     """
     runner = libfuzzer.get_runner(target_path)
+    logs.log('Runner is ' + str(type(runner)))
```
oliverchang: remember to remove this.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -478,14 +474,30 @@ def fuzz(self,
     """LibFuzzerCommon.fuzz override."""
     self._test_ssh()
 
-    #TODO(flowerhack): Pass libfuzzer args (additional_args) here
+    logs.log('SSH works; sync corpus directories')
+    # Push corpus directories to the device.
+    data_dst = self.fuzzer.data_path('corpus')
+    for corpus in corpus_directories:
+      for corpfile in os.listdir(corpus):
```
oliverchang: s/corpfile/corpus_file/

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -478,14 +474,30 @@ def fuzz(self,
...
     self.fuzzer.monitor(return_code)
     self.process_logs_and_crash(artifact_prefix)
 
     with open(self.fuzzer.logfile) as logfile:
       symbolized_output = logfile.read()
 
+    logs.log('Fuzzer ran; pull down corpus')
+    # Pull corpus directories from the device.
+    data_src = self.fuzzer.data_path('corpus/*')
+    # By convention, the *first* corpus directory in the list is always where
```
oliverchang: to clarify: you should be pulling from the first corpus on the VM onto the first corpus on the host, rather than everything in the first corpus.

src/python/platforms/fuchsia/util/fuzzer.py
```diff
@@ -206,7 +206,12 @@ def start(self, fuzzer_args):
       else:
         fuzzer_args.append('-jobs=1')
     self.device.ssh(['mkdir', '-p', self.data_path('corpus')])
-    if [x for x in fuzzer_args if not x.startswith('-')]:
+    # If all the arguments are prepended with '-', then we do not need to add
+    # a corpus.
+    # This list comprehension returns a list of all arguments that do *not*
+    # start with '-'.
+    # Thus, if the list is empty, we append data/corpus/.
+    if not [x for x in fuzzer_args if not x.startswith('-')]:
```
oliverchang: why was this needed? we should always be passing a corpus dir, even if it's empty.

src/python/bot/fuzzers/libfuzzer.py
flowerhack: nvm, it actually already does this via glob

src/python/bot/fuzzers/libfuzzer.py
flowerhack: changed to corpus_dir since that seems more accurate (it's a list of directories, not files, right?)

src/python/platforms/fuchsia/util/fuzzer.py
flowerhack: there is an option to pass in a custom corpus directory. if no corpus directory is passed in we assume you're using the default

src/python/bot/fuzzers/libfuzzer.py
flowerhack: Ok, how do we define "first" corpus on the VM?  (Is it first lexicographically?)

And to make sure I understand: You're saying I should pull the *first corpus* on the VM into the *first corpus* on the host?

(Once I'm sure I understand I'll update accordingly!)

flowerhack:
Apologies for the delay on this.

This integration test passes for me locally now (had to update the build / add a fuzzer to specifically target corpus generation).

I'll run the test again in the morning to make extra-sure things work, but other than that I think this is ready for hopefully the last review round!

src/python/bot/fuzzers/libfuzzer.py
oliverchang: first one provided in the list of arguments. when libFuzzer fuzzes,

e.g. `./fuzzer corpus1 corpus2`

All new units are written to the first provided corpus (i.e. corpus1). 

For the Fuchsia Qemu runner, we want to make this appear as if everything was happening on the host, hence why we should copy back items from the first (in the list of arguments) back to the corresponding directory on the host.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -478,14 +474,30 @@ def fuzz(self,
     """LibFuzzerCommon.fuzz override."""
     self._test_ssh()
 
-    #TODO(flowerhack): Pass libfuzzer args (additional_args) here
+    logs.log('SSH works; sync corpus directories')
+    # Push corpus directories to the device.
+    data_dst = self.fuzzer.data_path('corpus')
+    for corpus in corpus_directories:
+      for corpus_dir in os.listdir(corpus):
```
oliverchang: why is this "corpus_dir"? each corpus should contain corpus items (files), rather than more directories?

flowerhack: because i got my variable names flipped around in my head, whups, fixed to be more clear

src/python/bot/fuzzers/libfuzzer.py
flowerhack: alright, done!  should be more clear what it's doing now (added comments too)

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -511,6 +544,32 @@ def fuzz(self,
...
 
+  def merge(self,
+            corpus_directories,
+            merge_timeout,
+            artifact_prefix=None,
+            tmp_dir=None,
+            additional_args=None):
+    # TODO(flowerhack): Integrate some notion of a merge timeout.
+    # TODO(flowerhack): device.py doesn't have a notion of custom
+    # corpus directories, teach it
```
flowerhack: This second TODO should get done & tested before actually merging this CL.

Merging "works" at the moment in that it doesn't *break* anything, but it seems like the names used for the temporary and merge directories on target and on the host differ, and thus CF fails to actually pull down the merged corpus D:

It shouldn't be hard to fix, just a bit fiddly, and it's late, so I'll wrap this bit up Monday morning.

The rest of the CL should be fine to look at / code review / etc, though

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -484,6 +480,45 @@ def _restart_qemu(self):
     start_qemu()
     self._setup_device_and_fuzzer()
 
+  def _corpus_target_subdir(self, relpath):
```
oliverchang: nit: docstring

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -484,6 +480,45 @@ def _restart_qemu(self):
...
+    return os.path.join(self.corpus_directories_target, relpath)
+
+  def _init_corpus_info(self, corpus_directories):
+    """ Initialize corpus info for fuzzing and merging. """
+    self.corpus_directories = corpus_directories
```
oliverchang: we shouldn't be storing `corpus_directories` here, since the Runner class should not be tied to any particular invocation. We can just pass this as arguments to everything that needs it.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -484,6 +480,45 @@ def _restart_qemu(self):
...
+    # By convention, the *first* corpus directory in corpus_directories is
+    # always where new corpus files from the target are stored.
+    self.new_corpus_dir_host = corpus_directories[0]
```
oliverchang: similarly for these other variables, just calculate them as needed.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -492,15 +527,20 @@ def fuzz(self,
...
     """LibFuzzerCommon.fuzz override."""
     self._test_ssh()
+    self._init_corpus_info(corpus_directories)
+    self._push_corpora_from_host_to_target()
 
-    #TODO(flowerhack): Pass libfuzzer args (additional_args) here
-    return_code = self.fuzzer.start(additional_args)
+    # Run the fuzzer.
+    return_code = self.fuzzer.start(
+        ['data/' + self.new_corpus_relative_dir_target] + additional_args)
```
oliverchang: What about the other corpus dirs? If they're still somehow taken into account.. this is still very confusing. Can we make this take in all the corpus dirs explicitly?

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -484,6 +480,45 @@ def _restart_qemu(self):
...
+    # Make the name of the "new corpus directory" on the target match that of
+    # the host.
+    self.new_corpus_relative_dir_target = os.path.join(
+        'corpus', os.path.basename(self.new_corpus_dir_host))
+    self.new_corpus_dir_target = self.fuzzer.data_path(
+        self.new_corpus_relative_dir_target)
+
+  def _push_corpora_from_host_to_target(self):
+    # Push corpus directories to the device.
+    logs.log('Push corpora from host to target.')
```
oliverchang: we should clear everything before pushing new corpora.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -484,6 +480,45 @@ def _restart_qemu(self):
...
+    return os.path.join(self.corpus_directories_target, relpath)
+
+  def _init_corpus_info(self, corpus_directories):
+    """ Initialize corpus info for fuzzing and merging. """
+    self.corpus_directories = corpus_directories
+    # By convention, the *first* corpus directory in corpus_directories is
+    # always where new corpus files from the target are stored.
+    self.new_corpus_dir_host = corpus_directories[0]
+    # corpus_directories_target is analagous to corpus_directories on the host.
+    self.corpus_directories_target = self.fuzzer.data_path('corpus')
```
oliverchang: this one we can probably initialise in __init__ since it's not specific to any particular invocation.

flowerhack: i opted to just make a function that returns this, like i've with the other corpus-related variables, lmk what you think

src/python/bot/fuzzers/libfuzzer.py
flowerhack: the new code now passes in all of the corpus directories rather than just a single one

src/python/bot/fuzzers/libfuzzer.py
flowerhack: got it, made individual functions to calculate these vars

src/python/bot/fuzzers/libfuzzer.py
flowerhack: got it, made individual functions to calculate these vars

src/python/bot/fuzzers/libfuzzer.py
flowerhack: fixed this. merge directories on host & target now sync and pull things down appropriately

flowerhack:
@oliverchang : I left in the log messages for pushing/pulling corpora since those struck me as potentially useful for debugging and not too spammy.  Let me know if they seem excessive to you.  (removed the other new log messages)

PR #1046, opened 2019-09-26
Fix issue with UBSan overriding symbolize parameter of ASan.

inferno-chromium:
In a ASan+UBSan build, ASAN_OPTIONS=symbolize=0
UBSAN_OPTIONS=symbolize=1 causes the symbolize=1 to take preference.
This breaks the symbolize=0 defaults which is needed to skip
inline frames in crash state.
Also, make sure minimize task skips current |env| value. This is
also needed to fix existing testcases.

### DIFF
--- a/src/python/bot/tasks/setup.py
+++ b/src/python/bot/tasks/setup.py
@@ -130,10 +130,10 @@ def _get_application_arguments(testcase, task_name):
   return app_args
 
 
-def _setup_memory_tools_environment(testcase):
+def _setup_memory_tools_environment(testcase, task_name):
   """Set up environment for various memory tools used."""
   env = testcase.get_metadata('env')
-  if not env:
+  if not env or task_name == 'minimize':
     environment.reset_current_memory_tool_options(redzone_size=testcase.redzone)
     return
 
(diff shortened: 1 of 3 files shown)


src/python/bot/tasks/setup.py
```diff
@@ -130,10 +130,10 @@ def _get_application_arguments(testcase, task_name):
   return app_args
 
 
-def _setup_memory_tools_environment(testcase):
+def _setup_memory_tools_environment(testcase, task_name):
   """Set up environment for various memory tools used."""
   env = testcase.get_metadata('env')
-  if not env:
+  if not env or task_name == 'minimize':
```
oliverchang: nit: we should avoid dependencies on what the task is in our code. Can we remove this after a while since you already have the clear bit above?

inferno-chromium: This is mostly for cases when we try migration script or something that calls tasks.add_task minimize to retrigger things, but it will cause older env to be used.

oliverchang: Right, I just think this makes the code harder to understand and use to accomodate a very specific use case, so would prefer removing it if possible :)

Issue #918, opened 2019-08-29
Document generic project setup cron


inferno-chromium:
Requested by @asudhak and @acatinahat

PR #1049, opened 2019-09-26
Docs: Document generic project setup cron

tapaswenipathak:
Fixes https://github.com/google/clusterfuzz/issues/918.


inferno-chromium:
This is wrong description, we will update it sometime in near future.

PR #1051, opened 2019-09-27
Use a custom label name based on issue tracker.

inferno-chromium:


src/appengine/handlers/cron/cleanup.py
```diff
@@ -666,9 +677,10 @@ def notify_closed_issue_if_testcase_is_open(policy, testcase, issue):
 
     wrong_label = policy.label('wrong')
     if wrong_label:
-      issue_comment += (
-          ' Otherwise, ignore this notification and add {label} label.'.format(
-              label=wrong_label))
+      issue_comment += ((' Otherwise, ignore this notification and add '
```
oliverchang: nit: add *the* ? similarly in other places.

src/appengine/handlers/cron/cleanup.py
```diff
@@ -81,6 +81,17 @@ def _append_generic_incorrect_comment(comment, policy, suffix):
   return comment + GENERIC_INCORRECT_COMMENT.format(label=wrong_label) + suffix
 
 
+def _label_display_name(issue):
+  """Return label display for a issue (based on its issue tracker type)."""
+  issue_tracker_name = issue.issue_tracker.name
+  if issue_tracker_name == 'monorail':
```
oliverchang: We shouldn't hardcode the issue tracker names here, as the cleanup code should have no knowledge of which issue tracker implementations are provided.

a better alternative would be to have a 

```python
@property
def label_type(self):
  return 'label'  # default
```

in the IssueTracker.

PR #1052, opened 2019-09-27
Generalize functions to account for SSE variants.

inferno-chromium:


inferno-chromium:
This is too generic, fixing

mbarbella-chromium:
LGTM

PR #1056, opened 2019-09-28
Make label text customizable

inferno-chromium:
Order of label and label_type can be different in different
issue trackers. So, make it completely customizable.

### DIFF
--- a/src/appengine/libs/issue_management/issue_tracker.py
+++ b/src/appengine/libs/issue_management/issue_tracker.py
@@ -273,6 +273,10 @@ class IssueTracker(object):
     """Label type."""
     return 'label'  # default
 
+  def label_text(self, label):
+    """Text for a label (with label type)."""
+    return label + ' ' + self.label_type
+
   def new_issue(self):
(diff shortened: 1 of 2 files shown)


inferno-chromium:
TBR simple fix.

src/appengine/libs/issue_management/issue_tracker.py
```diff
@@ -273,6 +273,10 @@ def label_type(self):
     """Label type."""
     return 'label'  # default
```
oliverchang: seems like label_type is no longer needed, so we can remove to keep things simple?

inferno-chromium: It is used in one place here - https://github.com/google/clusterfuzz/blob/feff2e0914ae136291d85e01e43257d93a179c84/src/appengine/handlers/cron/cleanup.py#L1054
We can remove it if we fix this case somehow, but i don't see this as a big deal since we have a better label_type for google issue tracker (hotlist vs hotlistid) for this case.

PR #1064, opened 2019-10-01
Filter out crashes in fuzz_task which have an empty state.

### LINKED ISSUE #1063: Don't create empty testcases for timeouts/ooms
When REPORT_OOMS_AND_HANGS is false, we report unreproducible empty crashes, because we see that there is a `Test unit written to...` even if the crash is not something we consider.

oliverchang:
Should fix #1063.


oliverchang:
(Untested and speculative -- will look more into this tomorrow morning). /gcbrun

inferno-chromium:
LGTM

PR #1065, opened 2019-10-01
Update README.md

gouravbansal26:


inferno-chromium:
There is a detailed documentation link on homepage, please use that.

PR #1070, opened 2019-10-02
Make FuchsiaBuild a subclass of RegularBuild.

oliverchang:
So that we can get builds with different revisions. 
Also in this PR:
- Get rid of RESOURCES_DIR.
- Properly plumb list of targets to fuzzer selection code to account for weighting. 
`_unpack_build`, `_get_fuzz_targets_from_dir` etc are moved in the Build class, so that they may be overridden by subclasses as is the case for Fuchsia.
- Cleanup unnecessary setup in existing integration tests.

### DIFF
--- a/src/python/build_management/build_manager.py
+++ b/src/python/build_management/build_manager.py
@@ -17,7 +17,6 @@ from builtins import object
 from builtins import range
 import datetime
 import os
-import random
 import re
 import six
 import subprocess
@@ -695,6 +694,10 @@ class RegularBuild(Build):
   def build_dir(self):
     return self._build_dir
 
+  def _get_fuzz_targets_from_dir(self):
+    """Get fuzz targets from directory."""
+    return _get_fuzz_targets_from_dir(self.build_dir)
+
   def setup(self):
     """Sets up build with a particular revision."""
     self._pre_setup()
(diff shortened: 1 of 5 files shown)


oliverchang:
@eepeep

src/python/build_management/build_manager.py
```diff
@@ -740,75 +743,52 @@ class FuchsiaBuild(Build):
...
-    # Bucket for QEMU resources.
-    fuchsia_resources_dir = fuchsia.device.initialize_resources_dir()
-    environment.set_value('FUCHSIA_RESOURCES_DIR', fuchsia_resources_dir)
+    # Need to be set before fuchsia utils is called in setup().
+    environment.set_value(
+        'FUCHSIA_DIR', os.path.join(self.build_dir, self.FUCHSIA_DIR_REL_PATH))
+    environment.set_value('FUCHSIA_RESOURCES_DIR', self.build_dir)
+
+    # Does not support partial unpack.
+    assert environment.get_value('UNPACK_ALL_FUZZ_TARGETS_AND_FILES')
```
inferno-chromium: Can we add a message after ","

oliverchang: Done!

PR #1072, opened 2019-10-03
[fuchsia] Add sanitizer extension names as needed.

flowerhack:
Zircon fuzzers have an extension indicating what type of sanitizer they
use.  However, we want to group targets with different sanitizers
together.  So, add the ability to fitler fuzzers by sanitizer type, and
dynamically modify the URL based on the sanitizer being used.

### DIFF
--- a/src/python/platforms/fuchsia/util/fuzzer.py
+++ b/src/python/platforms/fuchsia/util/fuzzer.py
@@ -55,7 +55,19 @@ class Fuzzer(object):
     pass
 
   @classmethod
-  def filter(cls, fuzzers, name):
+  def _matches_sanitizer(cls, tgt, sanitizer):
+    """ Returns whether or not a target is for a given sanitizer. """
+    tgt, ext = os.path.splitext(tgt)
+    # Remove the leading '.' from the extension.
+    ext = ext[1:]
+    if ext == sanitizer:
+      return True
+    if sanitizer == 'asan' and ext == '':
+      return True
+    return False
+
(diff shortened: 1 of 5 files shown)


src/python/platforms/fuchsia/util/fuzzer.py
```diff
@@ -55,7 +55,20 @@ class StateError(ValueError):
...
+    ext = ext[1:]
+    if ext == sanitizer:
+      return True
+    # If there's no extension, assume it's an ASAN fuzzer.
+    if sanitizer == 'asan' and ext == '':
+      return True
+    return False
+
+  @classmethod
+  def filter(cls, fuzzers, name, sanitizer=None):
```
oliverchang: question for future cleanup: is anyone actually using `name` ? It complicates logic a fair bit and removing it will make this function much cleaner.

flowerhack: In Clusterfuzz itself: no.

However, "name" is what's used by people running fuzzers locally on Fuchsia, in order to do things like "run this specific fuzzer", see https://fuchsia.googlesource.com/fuchsia/+/refs/heads/master/scripts/fuzzing/lib/fuzzer.py#88 .

We definitely don't need anything like this when we upgrade to the new Clusterfuchsia interface, but for now I'm mildly inclined to keep it, so that the "scripts in Fuchsia" and the "scripts in Clusterfuzz" are closer to parity.

PR #1073, opened 2019-10-03
Add project as an attribute on job entity.

inferno-chromium:
This is needed for dynamic filters on UI e.g. fuzzer stats.

### DIFF
--- a/src/appengine/handlers/jobs.py
+++ b/src/appengine/handlers/jobs.py
@@ -17,7 +17,6 @@ from builtins import str
 import six
 
 from base import tasks
-from base import utils
 from datastore import data_handler
 from datastore import data_types
 from datastore import ndb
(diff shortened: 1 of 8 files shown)


src/appengine/handlers/jobs.py
```diff
@@ -127,6 +131,7 @@ def post(self):
     job.description = description
     job.environment_string = environment_string
     job.templates = templates
+    job.project = project
```
oliverchang: How about setting this as a `_pre_put_hook` in data_types.Job? That will be a bit more reliable, and can prevent issues in the future when we write e.g. migration scripts.

inferno-chromium: Great idea, forgot about hook. done!

oliverchang: this line needs to be removed now?

inferno-chromium: yes fixed after last patch

PR #1076, opened 2019-10-04
Add a Job query index.

inferno-chromium:


inferno-chromium:
TBR minor index add, need for project filtering on fuzzer stats.

PR #1077, opened 2019-10-04
Add project filter to fuzzer stats.

inferno-chromium:
Also, remove route hacks and instead use query params.

### DIFF
--- a/src/appengine/handlers/fuzzer_stats.py
+++ b/src/appengine/handlers/fuzzer_stats.py
@@ -387,43 +387,68 @@ def _get_date(date_value, days_ago):
 
 
 class Handler(base_handler.Handler):
-  """Group by day handler."""
+  """Fuzzer stats main page handler."""
 
   @handler.unsupported_on_local_server
   @handler.get(handler.HTML)
   def get(self):
     """Handle a GET request."""
-    # Create a list of externally contributed fuzzers.
-    user_email = helpers.get_user_email()
-
-    if access.has_access():
-      # User is an internal user of ClusterFuzz (eg: ClusterFuzz developer).
-      fuzzers_list = (
-          data_handler.get_all_fuzzer_names_including_children(
-              include_parents=True))
-      jobs_list = data_handler.get_all_job_type_names()
-    else:
+    if not access.has_access():
       # User is an external user of ClusterFuzz (eg: non-Chrome dev who
       # submitted a fuzzer or someone with a project in OSS-Fuzz).
+      user_email = helpers.get_user_email()
       fuzzers_list = external_users.allowed_fuzzers_for_user(
           user_email, include_from_jobs=True, include_parents=True)
-
       if not fuzzers_list:
         # User doesn't actually have access to any fuzzers.
         raise helpers.AccessDeniedException()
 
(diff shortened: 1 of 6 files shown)


src/appengine/handlers/fuzzer_stats.py
```diff
@@ -387,43 +387,68 @@ def _get_date(date_value, days_ago):
...
-    user_email = helpers.get_user_email()
+    if not access.has_access():
+      # User is an external user of ClusterFuzz (eg: non-Chrome dev who
+      # submitted a fuzzer or someone with a project in OSS-Fuzz).
+      user_email = helpers.get_user_email()
+      fuzzers_list = external_users.allowed_fuzzers_for_user(
+          user_email, include_from_jobs=True, include_parents=True)
+      if not fuzzers_list:
+        # User doesn't actually have access to any fuzzers.
+        raise helpers.AccessDeniedException()
```
oliverchang: nit: add a message?

src/appengine/handlers/fuzzer_stats.py
```diff
@@ -387,43 +387,68 @@ def _get_date(date_value, days_ago):
...
+      fuzzers_list = sorted(
+          external_users.allowed_fuzzers_for_user(
+              user_email, include_from_jobs=True, include_parents=True))
       if not fuzzers_list:
         # User doesn't actually have access to any fuzzers.
         raise helpers.AccessDeniedException()
 
-      jobs_list = external_users.allowed_jobs_for_user(user_email)
+      jobs_list = sorted(external_users.allowed_jobs_for_user(user_email))
+      projects_list = sorted(
```
oliverchang: Do we need to filter out duplicates?

src/appengine/handlers/fuzzer_stats.py
```diff
@@ -387,43 +387,68 @@ def _get_date(date_value, days_ago):
...
+              include_parents=True, project=project))
+      jobs_list = data_handler.get_all_job_type_names(project=project)
     else:
       # User is an external user of ClusterFuzz (eg: non-Chrome dev who
       # submitted a fuzzer or someone with a project in OSS-Fuzz).
-      fuzzers_list = external_users.allowed_fuzzers_for_user(
-          user_email, include_from_jobs=True, include_parents=True)
+      user_email = helpers.get_user_email()
 
+      fuzzers_list = sorted(
```
oliverchang: Is it intentional that project filters aren't supported here? (e.g. if someone external has access to more than 1).

src/appengine/handlers/fuzzer_stats.py
inferno-chromium: yes right, forgot about it.

src/appengine/handlers/fuzzer_stats.py
inferno-chromium: Done!

src/appengine/handlers/fuzzer_stats.py
inferno-chromium: yes added a TODO. I couldn't figure out how to support this properly, i think a fix has to come in external_users functions, wdyt ?

PR #1078, opened 2019-10-04
Use right job type when making decisions for variant task.

inferno-chromium:
Variant task does not use testcase.job_type, but overridden
env variable JOB_NAME. Fix tests to account for it.


inferno-chromium:
another way is to pass job_type as a setup_testcase argument.

inferno-chromium:
fixing

PR #1080, opened 2019-10-05
Log command for black box fuzzers.

inferno-chromium:
Fixes #997 for black box fuzzers. Engine fuzzers already log
this in log header.


inferno-chromium:
TBR to debug see why afl failing on variant task.

inferno-chromium:
ah, can't add to output since it will duplicate command line. let me just log it.

PR #1082, opened 2019-10-06
docs: fix testcase -> test case in architecture.md

piperchester:


inferno-chromium:
this is not the only place and we need to change all places in docs for consistency. Also, UI uses testcase quite extensively, so prefer to keep as-is.

PR #1083, opened 2019-10-07
Fix regression with AFL variant task not working.

inferno-chromium:
After new engine impl was enabled in testcase_manager's
test_for_crash_with_retries, it uses the testcase.get_fuzz_target
which picks up original fuzzing engine and not the one for the
variant task. For those cases, use a cloned testcase with the
right fuzzer_name and overridden_fuzzer_name so that it sets up
AFL engine fuzzing properly.

### DIFF
--- a/src/python/bot/tasks/variant_task.py
+++ b/src/python/bot/tasks/variant_task.py
@@ -24,6 +24,34 @@ from datastore import data_types
 from system import environment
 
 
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
LGTM I guess, the solution seems a bit hacky but I don't have a better one in mind

PR #1084, opened 2019-10-07
Skip syncing corpus and new corpus items for ephemeral bots.

inferno-chromium:
Skip syncing corpus for ephemeral bots. Since |self.gcs_corpus| is not set, we skip syncing new corpus files as well from |sync_new_corpus_files|.

### DIFF
--- a/src/python/bot/tasks/fuzz_task.py
+++ b/src/python/bot/tasks/fuzz_task.py
@@ -1284,6 +1284,12 @@ class FuzzingSession(object):
 
   def sync_corpus(self, sync_corpus_directory):
     """Sync corpus from GCS."""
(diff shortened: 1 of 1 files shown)


inferno-chromium:
> Shouldn't we also disable syncing of new corpus files too? i.e. add the same check in sync_new_corpus_files. Otherwise we might see blowups in new corpora items being added.

It is already done since self.gcs_corpus is not set.

  def sync_new_corpus_files(self):
    """Sync new files from corpus to GCS."""	    """Sync new files from corpus to GCS."""
    if not self.gcs_corpus:
     return

oliverchang:
Ah right, thanks for confirming.

PR #1088, opened 2019-10-08
Fix incorrect job_type comparison in variant task.

inferno-chromium:
Save testcase original job type and use it for comparison.
This regressed in https://github.com/google/clusterfuzz/pull/1083
since we change variant testcase to use variant job type.
Also, fix a bug where last tested crash stacktrace is not
updated.


inferno-chromium:
TBR

PR #1093, opened 2019-10-09
Add bucket path in exception for targets.list failures.

inferno-chromium:


inferno-chromium:
TBR, minor fix. this makes it more clear where exception is happening and helps separate alerts.

PR #1095, opened 2019-10-10
Fix get_all_project_names to use job projection.

### LINKED ISSUE #1094: get_all_project_names doesn't include projects for which there is no testcases
E.g. on fuzzer stats page, the project filter only includes projects for which there is at least one Testcase.

inferno-chromium:
Fixes #1094.

### DIFF
--- a/src/python/datastore/data_handler.py
+++ b/src/python/datastore/data_handler.py
@@ -94,10 +94,9 @@ FuzzerDisplay = collections.namedtuple(
 @memoize.wrap(memoize.Memcache(MEMCACHE_TTL_IN_SECONDS))
 def get_all_project_names():
   """Return all project names."""
-  query = data_types.Testcase.query(
-      projection=[data_types.Testcase.project_name],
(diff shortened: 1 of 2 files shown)


src/python/datastore/data_handler.py
```diff
@@ -94,10 +94,9 @@
 @memoize.wrap(memoize.Memcache(MEMCACHE_TTL_IN_SECONDS))
 def get_all_project_names():
   """Return all project names."""
-  query = data_types.Testcase.query(
-      projection=[data_types.Testcase.project_name],
-      distinct=True).order(data_types.Testcase.project_name)
-  return [testcase.project_name for testcase in query if testcase.project_name]
+  query = data_types.Job.query(
+      projection=[data_types.Job.project], distinct=True)
+  return sorted([_.project for _ in query])
```
oliverchang: nit: s/_/job/ to be clearer?

inferno-chromium: Done!

PR #1097, opened 2019-10-10
Add more tests for the reproduce tool.

mbarbella-chromium:
In the end I don't think adding the actual libFuzzer target is necessary. This ensures we go down the proper code paths for engine fuzzers (and it caught that we weren't properly following it before).


inferno-chromium:
@oliverchang - for another pair of eyes.

PR #1105, opened 2019-10-15
Auto-shrink testcase group when too many variants are found.

inferno-chromium:


src/python/tests/appengine/handlers/cron/grouper_test.py
```diff
@@ -249,3 +251,30 @@ def test_different_project_name_with_similar_state(self):
       self.testcases[index] = data_handler.get_testcase_by_id(t.key.id())
       self.assertEqual(self.testcases[index].group_id, 0)
       self.assertTrue(self.testcases[index].is_leader)
+
+
+@test_utils.with_cloud_emulators('datastore')
+class GroupExceedMaxTestcasesTest(unittest.TestCase):
```
inferno-chromium: Different class, since general grouper testcases always create two helper testcases in setUp.

src/appengine/handlers/cron/grouper.py
```diff
@@ -80,13 +81,147 @@ def combine_testcases_into_group(testcase_1, testcase_2, testcase_map):
...
+        testcase_map[other_testcase_id].security_flag and
+        testcase.one_time_crasher_flag ==
+        testcase_map[other_testcase_id].one_time_crasher_flag):
+      return True
+    # yapf: enable
+
+  return False
+
+
+def _shink_large_groups_if_needed(testcase_map):
```
oliverchang: nit: spelling of "shrink"

src/appengine/handlers/cron/grouper.py
```diff
@@ -80,13 +81,147 @@ def combine_testcases_into_group(testcase_1, testcase_2, testcase_map):
...
+  group_id_with_testcases_map = {}
+
+  for testcase in six.itervalues(testcase_map):
+    if not testcase.group_id:
+      continue
+
+    if not testcase.group_id in group_id_with_testcases_map:
+      group_id_with_testcases_map[testcase.group_id] = []
+
+    # Prioritize testcase with issues over regular testcases.
```
oliverchang: should we also prioritize reproducible over non-reproducible?

src/appengine/handlers/cron/grouper.py
inferno-chromium: Done!

src/appengine/handlers/cron/grouper.py
inferno-chromium: Done!

inferno-chromium:
> LGTM. I wonder if another way to address this is to make the similarity algorithm a bit more aggressive in the case of crash types with user provided input like CHECKs.

I fixed some signature in crash analyzer, but there are more crazy cases.

inferno-chromium:
take another look, although i wont check this in until next week.

inferno-chromium:
@oliverchang - can you please take another look.

oliverchang:
LGTM. Sorry didn't realise you were waiting on another review!

PR #1110, opened 2019-10-17
Add fuzz target name to log entry if available

inferno-chromium:


jonathanmetzman:
Do you think it is better to have the engine be None when not available? Or better to exclude this field entirely?

inferno-chromium:
> Do you think it is better to have the engine be None when not available? Or better to exclude this field entirely?

Better to have None for debugging, wdyt ?

inferno-chromium:
> > Do you think it is better to have the engine be None when not available? Or better to exclude this field entirely?
> 
> Better to have None for debugging, wdyt ?

Removed, it is nice to not inflate log entry when not available.

jonathanmetzman:
Can you delete the part in AFL that tries to do this as well, since it will be superfluous now: https://github.com/google/clusterfuzz/blob/master/src/python/bot/fuzzers/afl/launcher.py#L1440

Also should project be included as well? Or is it implied by the job in the task payload? I guess the way AFL currently does it, it has job and engine separately even though these are also implied by task_payload. Don't know if they are worth having separately (maybe so "show matching entries"feature can be used in stack driver), OTOH it's just superfluous.

inferno-chromium:
> Can you delete the part in AFL that tries to do this as well, since it will be superfluous now: https://github.com/google/clusterfuzz/blob/master/src/python/bot/fuzzers/afl/launcher.py#L1440

It cannot be deleted since a logs.configure call is still needed (since AFL's launcher.py is called as a seperate script, whereas libFuzzer run is now just a module call in same process).

> 
> Also should project be included as well? Or is it implied by the job in the task payload? I guess the way AFL currently does it, it has job and engine separately even though these are also implied by task_payload. Don't know if they are worth having separately (maybe so "show matching entries"feature can be used in stack driver), OTOH it's just superfluous.

Yes none of those params are needed, task_payload is enough, so removed.

PR #1111, opened 2019-10-17
Automatically correct invalid fuzzing dictionary entries.

mbarbella-chromium:
PTAL when you have time.


src/python/bot/fuzzers/afl/launcher.py
```diff
@@ -110,6 +110,11 @@ def from_target_path(cls, target_path):
         config.additional_afl_arguments, constants.DICT_FLAG, remove=False)
 
     config.use_default_dict(target_path)
+
+    # Correct the dictionary if needed.
+    if config.dict_path:
+      dictionary_manager.correct_dictionary(config.dict_path)
```
inferno-chromium: s/correct_dictionary/correct_if_needed ?
can remove dictionary since it is in dictionary manager.

src/python/bot/fuzzers/dictionary_manager.py
```diff
@@ -129,6 +133,54 @@ def merge_dictionary_files(original_dictionary_path,
...
+    for character in entry:
+      if character == '"' and prev_character != '\\':
+        new_entry += '\\'
+      new_entry += character
+      prev_character = character
+
+    new_entry = '"{entry}"'.format(entry=new_entry)
+    return name_part + new_entry
+
+  content = utils.read_data_from_file(dictionary_path, eval_data=False)
```
inferno-chromium: Can you bail out on

if not dictionary_path
and not os.path.exists(dictionary_path).

src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -112,13 +112,18 @@ def prepare(self, corpus_dir, target_path, _):
...
+        arguments, constants.DICT_FLAG, remove=False)
+    if not dict_argument:
       default_dict_path = dictionary_manager.get_default_dictionary_path(
           target_path)
       if os.path.exists(default_dict_path):
         arguments.append(constants.DICT_FLAG + default_dict_path)
 
+    # If we have a dictionary, correct any items that are not formatted properly
+    # (e.g. quote items that are missing them).
+    dictionary_manager.correct_dictionary(dict_argument)
```
inferno-chromium: This does not account for default_dict_path
Change dict_argument to dict_path
change default_dict_path to dict_path
and then do a null check 
if dict_path:
 dictionary_manager.correct_dictionary(dict_path)

src/python/bot/fuzzers/dictionary_manager.py
```diff
@@ -129,6 +133,54 @@ def merge_dictionary_files(original_dictionary_path,
   utils.write_data_to_file(merged_dictionary_data, merged_dictionary_path)
 
 
+def correct_dictionary(dictionary_path):
```
inferno-chromium: nit: dict_path for consistency with other places.

src/python/bot/fuzzers/dictionary_manager.py
```diff
@@ -129,6 +133,54 @@ def merge_dictionary_files(original_dictionary_path,
...
+
+    name_part = match.group(1) or ''
+    entry = match.group(2)
+
+    # Handle quote entries as a special case. This simplifies later logic.
+    if entry == '"':
+      entry = '"\\\""'
+
+    if entry.startswith('"') and entry.endswith('"'):
+      return name_part + entry
```
inferno-chromium: just return line is safer.

src/python/bot/fuzzers/dictionary_manager.py
```diff
@@ -129,6 +133,54 @@ def merge_dictionary_files(original_dictionary_path,
...
+    new_entry = '"{entry}"'.format(entry=new_entry)
+    return name_part + new_entry
+
+  content = utils.read_data_from_file(dictionary_path, eval_data=False)
+  new_content = ''
+  for current_line in content.splitlines():
+    new_content += fix_line(current_line) + '\n'
+
+  # End of file newlines are inconsistent in dictionaries.
+  if new_content.strip('\n') != content.strip('\n'):
```
inferno-chromium: maybe just rstrip ?

src/python/bot/fuzzers/dictionary_manager.py
```diff
@@ -129,6 +133,56 @@ def merge_dictionary_files(original_dictionary_path,
...
+    if not match:
+      raise errors.BadStateError(
+          'Failed to correct dictionary line "{line}" in {path}.'.format(
+              line=line, path=dictionary_path))
+
+    name_part = match.group(1) or ''
+    entry = match.group(2)
+
+    # Handle quote entries as a special case. This simplifies later logic.
+    if entry == '"':
```
inferno-chromium: Why is this needed, i think the loop handles this automatically. Also, this should be probably after next if ?

mbarbella-chromium: This breaks with the later logic since it both startswith and endswith " but isn't valid. I could add an additional length check there, but this seemed easier to follow.

src/python/bot/fuzzers/dictionary_manager.py
mbarbella-chromium: The current behavior is needed to handle the special case above.

src/python/bot/fuzzers/dictionary_manager.py
```diff
@@ -129,6 +133,56 @@ def merge_dictionary_files(original_dictionary_path,
...
+
+    # In this case, we know the entry is invalid. Escape any unescaped quotes
+    # within it, then append quotes to the front and back.
+    new_entry = ''
+    prev_character = ''
+    for character in entry:
+      if character == '"' and prev_character != '\\':
+        new_entry += '\\'
+      new_entry += character
+      prev_character = character
```
Dor1s: can you add a few hex escaped entries to the test inputs, e.g. `\x10\xAA` to make sure we handle them as well?

src/python/bot/fuzzers/afl/launcher.py
mbarbella-chromium: Done

src/python/bot/fuzzers/dictionary_manager.py
mbarbella-chromium: Done

src/python/bot/fuzzers/dictionary_manager.py
mbarbella-chromium: Done

src/python/bot/fuzzers/dictionary_manager.py
mbarbella-chromium: Done

src/python/bot/fuzzers/libFuzzer/engine.py
mbarbella-chromium: Done. Skipped the null check (and removed it from AFL) since one was added to correct_if_needed.

src/python/bot/fuzzers/dictionary_manager.py
mbarbella-chromium: Good idea. Updated the incorrect dictionary test to include an example and switched to the JPEG dictionary for the real-world correct dictionary test.

src/python/bot/fuzzers/dictionary_manager.py
```diff
@@ -129,6 +133,59 @@ def merge_dictionary_files(original_dictionary_path,
   utils.write_data_to_file(merged_dictionary_data, merged_dictionary_path)
 
 
+def correct_if_needed(dict_path):
+  """Corrects obvious errors such as missing quotes in a dictionary."""
+
+  def fix_line(line):
+    """Correct a single dictionary line."""
+    # Ignore blank and comment lines.
+    if not line or line.startswith('#'):
```
oliverchang: how about whitespace chars that come before "#" ? should we call strip() on the line first?

src/python/bot/fuzzers/dictionary_manager.py
```diff
@@ -129,6 +133,59 @@ def merge_dictionary_files(original_dictionary_path,
   utils.write_data_to_file(merged_dictionary_data, merged_dictionary_path)
 
 
+def correct_if_needed(dict_path):
+  """Corrects obvious errors such as missing quotes in a dictionary."""
+
+  def fix_line(line):
```
oliverchang: this nested function is pretty long, so might be worth putting it outside for readability.

mbarbella-chromium:
Comments addressed in https://github.com/google/clusterfuzz/pull/1114

It's true that something like "\" would fail and there are probably a few other corner cases. It may be worth cleaning this up at some point, but the logic would start to get fairly complicated for something that would only affect a very small number of dictionaries.

PR #1113, opened 2019-10-17
Downweight fuzz targets with known issues more heavily.

mbarbella-chromium:


mbarbella-chromium:
No rush on this one, let me know if you have any feedback on this or other targets that should be changed.

At the moment it looks like about 2/3 of fuzzers fall into the bucket of being weighted at .7 for having stale coverage. It definitely seems like some of these issues should be weighted lower, and I feel like slow units and timeouts might be more severe than ooms or crashes. I think our startup crash weight of .1 is working well, but as an interesting data point it seems like these are still running often enough that they can be corrected, so perhaps we could go a bit lower on these others.

src/appengine/handlers/cron/fuzzer_and_job_weights.py
```diff
@@ -136,7 +136,7 @@ def _coverage_formatter(query_format, dataset):
 # healthy fuzzers are expected to have some crashes.
 CRASH_SPECIFICATION = QuerySpecification(
     query_format=GENERIC_QUERY_FORMAT.format(
-        field_name='crash_count', min_weight=0.70),
+        field_name='crash_count', min_weight=0.50),
```
Dor1s: I'd drop both this and oom as well a bit more

inferno-chromium: It depends how much fuzzers is crashing here, decreasing this can hurt real fuzzers which find some crashes from time to time.

mbarbella-chromium: Since we scale linearly based on how often the crashes are occurring it might be reasonable to go lower. I'm not sure what the right value would be here, but it could be lowered more.

mbarbella-chromium:
> I think you forgot to change OOM weight to 0.25 as well ?

This was intentional, but I'm not sure if it's correct. Do you think we should treat OOMs as being as sever as the timeout issues in terms of how much they affect fuzzing? Intuitively I think the timeouts are a bit more detrimental, but it's not based on much.

> Maybe change coverage weight from 0.7 to 0.75 ? (to match 0.25, 0.50, 0.75 style ?)

SGTM. I would like to have more tiers of the coverage rule (e.g. if something has been stuck for months, or if we're reasonably certain it has 100% coverage, maybe we should downweight it even more?), but that can come later.

inferno-chromium:
> > I think you forgot to change OOM weight to 0.25 as well ?
> 
> This was intentional, but I'm not sure if it's correct. Do you think we should treat OOMs as being as sever as the timeout issues in terms of how much they affect fuzzing? Intuitively I think the timeouts are a bit more detrimental, but it's not based on much.

What do @oliverchang @jonathanmetzman think of these new weight and if OOM should be weighted higher / lower ?

> 
> > Maybe change coverage weight from 0.7 to 0.75 ? (to match 0.25, 0.50, 0.75 style ?)
> 
> SGTM. I would like to have more tiers of the coverage rule (e.g. if something has been stuck for months, or if we're reasonably certain it has 100% coverage, maybe we should downweight it even more?), but that can come later.

mbarbella-chromium:
> LGTM, but I agree with @inferno-chromium on making OOMs match timeout weight (0.25).
> 
> They have the same immediate impact as timeouts in that they stop fuzzing and prevents fuzzers from making useful progress. Could you please explain why you think timeouts are more detrimental than OOMs?

I think I will go ahead and lower them, but my rationale was that they tend to go along with slow units. At least in the case where we actually hit a timeout/crash/OOM we'll stop fuzzing with this target and go on to doing something else. I've seen some cases where we hit several slow units, wasting some time, then eventually hit a timeout. I don't have any evidence of how prevalent this might be, and thinking more we should usually still hit the timeouts relatively quickly.

PR #1116, opened 2019-10-18
Attempt to handle some unquoted dictionary entries where "=" is used.

mbarbella-chromium:
PTAL. This also fixes a minor issue with whitespace around an = in valid entries.


src/python/bot/fuzzers/dictionary_manager.py
```diff
@@ -44,7 +44,7 @@
 TOKEN_USELESS_DICT_END = 'End of useless dictionary elements.'
 TOKEN_USELESS_DICT_START = 'Useless dictionary elements.'
 
-DICTIONARY_PART_PATTERN = re.compile('([^"]+=)?(.*)')
+DICTIONARY_PART_PATTERN = re.compile('([A-Za-z0-9_]+\s*=\s*)?(.*)')
```
inferno-chromium: Where do you see the need for [A-Za-z0-9_] ? Is that enforced by libfuzzer dictionary parsing

mbarbella-chromium: It actually doesn't enforce that, but this was an attempt to handle the "!=" case. I suppose another way around it would be to ensure that we don't end with "=" which might be less error-prone. I'll switch to this.

PR #1118, opened 2019-10-20
Speed up cleanup cron and avoid issue tracker token expiration.

inferno-chromium:
issue_tracker_utils.was_label_added is an expensive call that
requires pulling in all issue comments. It was already optimized
for most cases except these two:
- Assigning owner and ccs from predator results.
- Assigning ccs from owners file.

### DIFF
--- a/src/appengine/libs/issue_management/issue_tracker_utils.py
+++ b/src/appengine/libs/issue_management/issue_tracker_utils.py
@@ -178,7 +178,7 @@ def was_label_added(issue, label):
     return False
 
   # Optimization that does not require pulling in issue's actions.
-  if any([label.lower() == l.lower() for l in issue.labels]):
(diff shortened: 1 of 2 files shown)


src/appengine/libs/issue_management/issue_tracker_utils.py
```diff
@@ -177,6 +177,10 @@ def was_label_added(issue, label):
   if not label:
     return False
 
+  # Optimization that does not require pulling in issue's actions.
+  if any([label.lower() == l.lower() for l in issue.labels]):
```
oliverchang: nit: no need for surrounding []. It'll make this faster too.

inferno-chromium: done!

Issue #1001, opened 2019-09-17
Adding Label field in the Clusterfuzz form fill


anshuman-goel:
To get due credit for the bugs found from Microsoft Fuzzing, a label needs to be added while submitting the job in Clusterfuzz. It can be on the form where one can add the label and it auto adds on the bug created by the Clusterfuzz.

oliverchang:
Re-opening as we need to relax the restrictions here to let non-privileged users add this in ClusterFuzz. However, we can't do this in places like OSS-Fuzz since labels there control deadlines.

PR #1123, opened 2019-10-22
Add issue labels field to testcase upload form

inferno-chromium:
Fixes https://github.com/google/clusterfuzz/issues/1001


src/appengine/private/components/upload-testcase/upload-form.html
```diff
@@ -182,6 +182,7 @@
       <div class="inline wide">
         <paper-input value="{{uploadParams.args}}" label="Command-line arguments (optional)" title="Additional command line flags to append when running this test case. Only necessary if a flag that isn't part of the job definition is required."></paper-input>
         <paper-input value="{{uploadParams.gestures}}" label="Gestures (optional)" title="A string representing gestures to use with this test case. It's generally not possible to manually create a gesture string. This tends to be used when re-uploading existing test cases which require gestures."></paper-input>
+        <paper-input value="{{uploadParams.issue_labels}}" label="Issue labels / hotlists (optional)" title="A set of label(s) / hotlist(s) to add to the issue. Should be a string of values separated by commas, e.g. value1,value2"></paper-input>
```
Dor1s: maybe: `e.g. Label-One,Label-Two` ?

inferno-chromium: It can be hotlist and hotlist can even be an integer id, so kept it generic to value1,value2

PR #1124, opened 2019-10-23
Fix chromedriver version to 77.

oliverchang:
Ubuntu CI image is still on Chrome 77.


oliverchang:
TBR to unblock CI deployment. I think this came up before and might be a pain point every time there is a new chrome release.

oliverchang:
> Yes this is a pain. Do we need to hardcode CI with a chrome version too, otherwise in future, it breaks too.

Not sure how to do that, since the package manager will only have the latest version. Can investigate more later.

PR #1136, opened 2019-10-24
Add HWASan parsing in a missing regex.

inferno-chromium:
Most regexes do .*AddressSanitizer, but this one matches a
ERROR one (not warning), so add HWASan explicitly.


inferno-chromium:
TBR minor fix.

oliverchang:
LGTM

PR #1141, opened 2019-10-25
Make /data/local/temp dir, needed by libFuzzer merge on Android.

inferno-chromium:


src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -888,6 +888,7 @@ def cleanse_crash(self,
 class AndroidLibFuzzerRunner(new_process.ProcessRunner, LibFuzzerCommon):
   """Android libFuzzer runner."""
   DEVICE_FUZZING_DIR = '/data/fuzz'
+  LIBFUZZER_TEMP_DIR = '/data/local/tmp'
```
Dor1s: nit: maybe give it some libFuzzer specific name since we use it in libFuzzer launcher only?

Dor1s: Thanks for clarifying!

PR #1142, opened 2019-10-25
Don't run fork mode on libFuzzer for Android.

inferno-chromium:


inferno-chromium:
> LGTM as-is, but one question.
> 
> Is it reasonable to run it on Android with a lower process count, or is the battery issue a problem regardless?

Fixed comment. Mitch said even 2 processes lead to 13% degradation. Need more evaluation, will reenable once things stabilize.

PR #1143, opened 2019-10-26
Use /. in adb push directory instead of /* to avoid shell expansion.

inferno-chromium:
/* leads to shell expansion and fails when a folder is large,
e.g. corpus. This also matches the adb pull directory syntax
as well.


inferno-chromium:
TBR

PR #1144, opened 2019-10-26
Fix libFuzzer symbolization on Android

inferno-chromium:
Launcher.py is removed in #1125, so fix symbolization code in
Android to use BUILD_DIR instead of deriving it from APP_PATH
(which is no longer applicable in new engine pipeline)


inferno-chromium:
TBR minor fix.

oliverchang:
LGTM

PR #1146, opened 2019-10-28
Add a source column to stats.

oliverchang:


src/python/metrics/fuzzer_stats.py
```diff
@@ -226,6 +226,10 @@ def __init__(self, fuzzer, job, build_revision, timestamp):
         'kind': 'TestcaseRun',
     })
 
+    source = environment.get_value('STATS_SOURCE')
+    if source:
+      self._stats_data['source'] = source
```
inferno-chromium: Should there be a default ?

oliverchang: I don't think we should have a default, since all the existing stats won't have it populated. In any case I think having this blank is equivalent to any arbitrary value we pick as the default.

oliverchang:
> STATS_LOCATION sounds a little better, but upto you, location or source either is fine.

SOURCE sounds more general and more applicable to other use cases :)

PR #1147, opened 2019-10-28
Don't exception out when fuzzer build directory is not found.

inferno-chromium:
This is expected when doing variant task on a job where build
does not exist (e.g. fuzz target does not exist for that config).

### DIFF
--- a/src/python/bot/fuzzers/engine_common.py
+++ b/src/python/bot/fuzzers/engine_common.py
@@ -276,6 +276,14 @@ def dump_big_query_data(stats, testcase_file_path, fuzzer_command):
 
 def find_fuzzer_path(build_directory, fuzzer_name):
   """Find the fuzzer path with the given name."""
+  if not build_directory:
+    # Grey-box fuzzers might not have the build directory for a particular job
+    # configuration when doing variant task testing (e.g. Android on-device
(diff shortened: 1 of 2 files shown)


src/python/bot/fuzzers/engine_common.py
```diff
@@ -276,6 +276,14 @@ def dump_big_query_data(stats, testcase_file_path, fuzzer_command):
 
 def find_fuzzer_path(build_directory, fuzzer_name):
   """Find the fuzzer path with the given name."""
+  if not build_directory:
+    # Grey-box fuzzers might not have the build directory for a particular job
+    # configuration when doing variant task testing (e.g. Android on-device
+    # fuzz target might not exist on host). In this case, treat it similar to
+    # target not found by returning None.
+    logs.log_warn('No build directory found for fuzzer: %s' % fuzzer_name)
```
jonathanmetzman: If this is expected to happen under normal conditions, is "warn" a better log level than "info"?

inferno-chromium: This matches the below case when fuzz target binary is also not found.  Warn does not show on dash, so might be nice to highlight in stackdriver to inspect a potential problem.

Issue #901, opened 2019-08-28
Add signatures for HWAddressSanitizer


oliverchang:

inferno-chromium:
Done in https://github.com/google/clusterfuzz/pull/1136
Rest crash types like Invalid-free, Tag-mismatch automatically categorized as Security.

PR #1148, opened 2019-10-28
Fix a HWASan crash type, add tests.

inferno-chromium:
Fixes https://github.com/google/clusterfuzz/issues/901

### DIFF
--- a/src/python/crash_analysis/stack_parsing/stack_analyzer.py
+++ b/src/python/crash_analysis/stack_parsing/stack_analyzer.py
@@ -106,6 +106,9 @@ GENERIC_SEGV_HANDLER_REGEX = re.compile(
 GOOGLE_CHECK_FAILURE_REGEX = re.compile(GOOGLE_LOG_FATAL_PREFIX +
                                         r'\s*Check failed[:]\s*(.*)')
 GOOGLE_LOG_FATAL_REGEX = re.compile(GOOGLE_LOG_FATAL_PREFIX + r'\s*(.*)')
+HWASAN_ALLOCATION_TAIL_OVERWRITTEN_ADDRESS_REGEX = re.compile(
+    r'.*ERROR: HWAddressSanitizer: allocation-tail-overwritten; '
+    r'heap object \[([xX0-9a-fA-F]+),.*of size')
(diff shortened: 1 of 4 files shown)


src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -60,7 +60,7 @@
     r'.*(AddressSanitizer).*memcpy-param-overlap'
     r'[^\[]*([\[].*[)])')
 ASAN_REGEX = re.compile(
-    r'.*ERROR: (HWAddressSanitizer|AddressSanitizer)[: ]*[ ]*([^(:]+)')
+    r'.*ERROR: (HWAddressSanitizer|AddressSanitizer)[: ]*[ ]*([^(:;]+)')
```
Dor1s: I've checked the code, looks like `allocation-tail-overwritten` is the only signature that has `;`. Should as ask @eugenis to change it, if possible?

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_data/hwasan_allocation_tail_overwritten.txt
```diff
@@ -0,0 +1,44 @@
+==11246==ERROR: HWAddressSanitizer: allocation-tail-overwritten; heap object [0x00444b02b180,0x00444b02b182) of size 2
```
Dor1s: Would it make sense to parse `0x00444b02b180` as crash address?

src/python/crash_analysis/stack_parsing/stack_analyzer.py
inferno-chromium: It will take a while for the change to make it in build system, and this seems simple enough to add as a char.

Dor1s: yes, definitely! I meant for the long term we can ask @eugenis to get rid of it together with fixing the typo

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_data/hwasan_allocation_tail_overwritten.txt
inferno-chromium: Done!

Issue #860, opened 2019-08-19
Tracking issue: support go-fuzz


Dor1s:
We were thinking about this for a while, now started to make some progress.

There are two modes of go-fuzz:

1) "native" mode (i.e. using `go-fuzz`): https://github.com/dvyukov/go-fuzz

2) libFuzzer-compatible mode: https://github.com/dvyukov/go-fuzz#libfuzzer-support -- requires certain build actions, but the resulting binary has libFuzzer interface

Part 2) is easier to integrate, as we can rely on the existing libFuzzer fuzzing engine integration. That's what we start from.

Dor1s:
Related PRs:

1) Initial golang stacktrace support: https://github.com/google/clusterfuzz/pull/827

2) Follow up for some standard stackframes: https://github.com/google/clusterfuzz/pull/853

3) Follow up for even better parsing and unknown crash type handling: https://github.com/google/clusterfuzz/pull/854

Dor1s:
Hey @dvyukov, when you get a chance, could you please tell us more about the advantages the native `go-fuzz` has over the libFuzzer-compatible mode? Is that something CF must support eventually, in your opinion? Thanks!

Dor1s:
A relevant discussion about code coverage for Go is happening in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=16549#c2

dvyukov:
> Hey @dvyukov, when you get a chance, could you please tell us more about the advantages the native `go-fuzz` has over the libFuzzer-compatible mode? Is that something CF must support eventually, in your opinion? Thanks!

+ @josharian @thepudds

I can think of 2 things (besides the fact that obviously everybody thinks that his fuzzing engine is the best one :)):
1. We don't have comparison instrumentation in libfuzzer mode. Theoretically this is doable but as far as I understand libfuzzer does not have an interface to inject them at the moment.
2. Most likely the 2 modes will find different subsets of bugs (significantly overlapping, but still there will be some unique ones). Now I suspect lots of users will use the native go-fuzz locally and then oss-fuzz will not be in the best light when it will appear that it somehow misses easy to find bugs. Users may not understand the internal mechanics, for them it will look like they are finding some bugs very quickly locally but oss-fuzz is not reporting them.

Dor1s:
TODO from https://github.com/google/clusterfuzz/pull/1151/files#diff-64146a33e026ed0a850f80210c83b508R269

[...]

PR #1151, opened 2019-10-29
Improve log parsing to fix performance analyzer for Go targets (#860).

Dor1s:
e.g. https://oss-fuzz.com/performance-report/libFuzzer_go-attestation_parse_event_log_fuzzer/libfuzzer_asan_go-attestation/2019-10-28 reports a starup crash, basically because Go fuzz targets we're running don't have coverage edges, they use only "features", and that makes logs a bit different.

I don't really like adding `LIBFUZZER_LOG_START_INITED_NO_COVERAGE_REGEX`, but couldn't update the existing `LIBFUZZER_LOG_START_INITED_REGEX`, as we use it to extract cov. and ft. values.

I might be able to get rid of some parts of it with the new merge implementation, will add a TODO to that bug.

### DIFF
--- a/src/python/bot/fuzzers/libFuzzer/stats.py
+++ b/src/python/bot/fuzzers/libFuzzer/stats.py
@@ -47,7 +47,7 @@ LIBFUZZER_LOG_DICTIONARY_REGEX = re.compile(r'Dictionary: \d+ entries')
 LIBFUZZER_LOG_END_REGEX = re.compile(r'Done\s+\d+\s+runs.*')
 LIBFUZZER_LOG_IGNORE_REGEX = re.compile(r'.*WARNING:.*Sanitizer')
 LIBFUZZER_LOG_LINE_REGEX = re.compile(
-    r'^#\d+.*(READ|NEW|pulse|REDUCE|RELOAD|ft:)')
+    r'^#\d+.*(READ|INITED|NEW|pulse|REDUCE|RELOAD|DONE)')
 LIBFUZZER_LOG_SEED_CORPUS_INFO_REGEX = re.compile(
     r'INFO:\s+seed corpus:\s+files:\s+(\d+).*rss:\s+(\d+)Mb.*')
 LIBFUZZER_LOG_START_INITED_REGEX = re.compile(
@@ -266,6 +266,8 @@ def parse_performance_features(log_lines,
 
     match = LIBFUZZER_MODULES_LOADED_REGEX.match(line)
     if match:
+      # TODO(mmoroz): this is missing cases when fuzz target binary does not
+      # have edge coverage instrumentation (e.g. Go targets).
       stats['startup_crash_count'] = 0
       stats['edges_total'] = int(match.group(2))
 
--- a/src/python/bot/fuzzers/libFuzzer/stats.py
+++ b/src/python/bot/fuzzers/libFuzzer/stats.py
@@ -41,16 +41,19 @@ LIBFUZZER_TIMEOUT_TESTCASE_REGEX = re.compile(
     LIBFUZZER_CRASH_TYPE_REGEX.format(type='timeout'))
 
 # Regular expressions to detect different sections of logs.
+LIBFUZZER_EXTRA_COUNTERS_REGEX = re.compile(r'^INFO:\s+\d+\s+Extra\s+Counters')
 LIBFUZZER_FUZZING_STRATEGIES = re.compile(r'cf::fuzzing_strategies:\s*(.*)')
 LIBFUZZER_LOG_COVERAGE_REGEX = re.compile(r'#\d+.*cov:\s+(\d+)\s+ft:\s+(\d+).*')
 LIBFUZZER_LOG_DICTIONARY_REGEX = re.compile(r'Dictionary: \d+ entries')
-LIBFUZZER_LOG_END_REGEX = re.compile(r'Done \d+ runs.*')
+LIBFUZZER_LOG_END_REGEX = re.compile(r'Done\s+\d+\s+runs.*')
 LIBFUZZER_LOG_IGNORE_REGEX = re.compile(r'.*WARNING:.*Sanitizer')
-LIBFUZZER_LOG_LINE_REGEX = re.compile(r'^#\d+.*(READ|cov:)')
+LIBFUZZER_LOG_LINE_REGEX = re.compile(r'^#\d+.*(READ|NEW|pulse|REDUCE|ft:)')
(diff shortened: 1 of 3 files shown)


src/python/bot/fuzzers/libFuzzer/stats.py
```diff
@@ -41,16 +41,19 @@
...
+LIBFUZZER_LOG_END_REGEX = re.compile(r'Done\s+\d+\s+runs.*')
 LIBFUZZER_LOG_IGNORE_REGEX = re.compile(r'.*WARNING:.*Sanitizer')
-LIBFUZZER_LOG_LINE_REGEX = re.compile(r'^#\d+.*(READ|cov:)')
+LIBFUZZER_LOG_LINE_REGEX = re.compile(r'^#\d+.*(READ|NEW|pulse|REDUCE|ft:)')
 LIBFUZZER_LOG_SEED_CORPUS_INFO_REGEX = re.compile(
     r'INFO:\s+seed corpus:\s+files:\s+(\d+).*rss:\s+(\d+)Mb.*')
 LIBFUZZER_LOG_START_INITED_REGEX = re.compile(
     r'#\d+\s+INITED\s+cov:\s+(\d+)\s+ft:\s+(\d+).*')
+LIBFUZZER_LOG_START_INITED_NO_COVERAGE_REGEX = re.compile(
+    r'#\d+\s+INITED\s+ft:\s+(\d+).*')
```
inferno-chromium: nit: Add a "  # Applies to Go binaries with no coverage edges, only features ?"

src/python/bot/fuzzers/libFuzzer/stats.py
```diff
@@ -262,6 +266,10 @@ def parse_performance_features(log_lines,
       stats['startup_crash_count'] = 0
       stats['edges_total'] = int(match.group(2))
 
+    match = LIBFUZZER_EXTRA_COUNTERS_REGEX.match(line)
```
inferno-chromium: I am thinking if we even need this LIBFUZZER_EXTRA_COUNTERS_REGEX at all. How about just do 
stats['startup_crash_count'] = 0 as next line in libfuzzer_inited = True branch.

src/python/bot/fuzzers/libFuzzer/stats.py
```diff
@@ -41,16 +41,19 @@
...
 # Regular expressions to detect different sections of logs.
+LIBFUZZER_EXTRA_COUNTERS_REGEX = re.compile(r'^INFO:\s+\d+\s+Extra\s+Counters')
 LIBFUZZER_FUZZING_STRATEGIES = re.compile(r'cf::fuzzing_strategies:\s*(.*)')
 LIBFUZZER_LOG_COVERAGE_REGEX = re.compile(r'#\d+.*cov:\s+(\d+)\s+ft:\s+(\d+).*')
 LIBFUZZER_LOG_DICTIONARY_REGEX = re.compile(r'Dictionary: \d+ entries')
-LIBFUZZER_LOG_END_REGEX = re.compile(r'Done \d+ runs.*')
+LIBFUZZER_LOG_END_REGEX = re.compile(r'Done\s+\d+\s+runs.*')
 LIBFUZZER_LOG_IGNORE_REGEX = re.compile(r'.*WARNING:.*Sanitizer')
-LIBFUZZER_LOG_LINE_REGEX = re.compile(r'^#\d+.*(READ|cov:)')
+LIBFUZZER_LOG_LINE_REGEX = re.compile(r'^#\d+.*(READ|NEW|pulse|REDUCE|ft:)')
```
inferno-chromium: Can ft: exist on its own. Since cov: is removed, why need ft: ?

inferno-chromium: There is a RELOAD as well to add here.

Dor1s: Yes, if you compile with e.g. only `-fsanitize-coverage=inline-8bit-counters` you'll get only features without edges.

Added `RELOAD`, but I don't think it ever happened on CF, as it requires either old style `-workers=N` parallelism or any other process that could write into the output corpus directory concurrently with the fuzzer. 

The fork mode uses an isolated dir and disables reload explicitly.

src/python/bot/fuzzers/libFuzzer/stats.py
Dor1s: It's not necessarily Go binaries, can be anything, just depends on the instrumentation used.

src/python/bot/fuzzers/libFuzzer/stats.py
Dor1s: Good idea, but `calculate_log_lines` doesn't touch the stats as of now. We can rely on `libfuzzer_lines_count` being non-zero, as that's essentially the same logic. Done, thanks for the suggestion!

src/python/bot/fuzzers/libFuzzer/stats.py
inferno-chromium: Sorry, i meant if "|ft:" is needed in regex since every line has to start with (READ|NEW|pulse|REDUCE|RELOAD) and then either cov: or ft: . either you should add like
(READ|NEW|pulse|REDUCE|RELOAD)\s*(cov|ft): ? or skip "s*(cov|ft):" altogether ?

src/python/bot/fuzzers/libFuzzer/stats.py
inferno-chromium: Thanks!

src/python/bot/fuzzers/libFuzzer/stats.py
Dor1s: Yeah, can remove it, but have to add `INITED` and `DONE` as well, which makes sense I think.

inferno-chromium: yes right, those two make sense to add, then can remove ft: as it looks weird without cov:

src/python/bot/fuzzers/libFuzzer/stats.py
```diff
@@ -209,6 +213,9 @@ def parse_performance_features(log_lines,
   (stats['log_lines_unwanted'], stats['log_lines_from_engine'],
    stats['log_lines_ignored']) = calculate_log_lines(log_lines)
 
+  if stats['log_lines_from_engine'] > 0:
+    stats['startup_crash_count'] = 0
+
```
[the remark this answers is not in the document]
Dor1s: SG, then I'll leave this line for now but leave a comment that it doesn't work for cases without edge coverage instrumentation.

PR #1156, opened 2019-10-30
Upload fuzzing log before crashers.

oliverchang:
Crasher uploading may run into exceptions, so upload the log first to
aid in debugging.


oliverchang:
TBR for simple change to aid with debugging.

PR #1158, opened 2019-10-31
Exclude HWASan stack frames

inferno-chromium:


inferno-chromium:
TBR minor signature add.

PR #1159, opened 2019-10-31
Fix flakiness, limit requirement for timeout crashes

inferno-chromium:
- Increase limit for timeout processing to 60 secs, but keep
  25 secs for regular crashes.
- Don't run with -runs=100, it should help reduce flakiness of
  timeout crashes. Timeout crashes should just reproduce in
  one iteration.

### DIFF
--- a/src/python/bot/fuzzers/libFuzzer/engine.py
+++ b/src/python/bot/fuzzers/libFuzzer/engine.py
@@ -324,6 +324,7 @@ class LibFuzzerEngine(engine.Engine):
       logs.log('Retrying timeout crash with a higher timeout: %d secs' %
                constants.REPRODUCTION_TIMEOUT_LIMIT)
       fuzzer_utils.extract_argument(arguments, constants.TIMEOUT_FLAG)
+      fuzzer_utils.extract_argument(arguments, constants.RUNS_FLAG)
       arguments.append('%s%d' % (constants.TIMEOUT_FLAG,
                                  constants.REPRODUCTION_TIMEOUT_LIMIT))
       result = runner.run_single_testcase(
--- a/src/python/bot/fuzzers/libFuzzer/engine.py
+++ b/src/python/bot/fuzzers/libFuzzer/engine.py
@@ -329,7 +329,8 @@ class LibFuzzerEngine(engine.Engine):
                                  constants.REPRODUCTION_TIMEOUT_LIMIT))
       result = runner.run_single_testcase(
           input_path,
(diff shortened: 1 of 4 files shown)


src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -317,6 +317,21 @@ def reproduce(self, target_path, input_path, arguments, max_time):
...
+    if constants.TEST_TIMEOUT_SIGNATURE in result.output:
+      logs.log('Retrying timeout crash with a higher timeout: %d secs' %
+               constants.REPRODUCTION_TIMEOUT_LIMIT)
+      fuzzer_utils.extract_argument(arguments, constants.TIMEOUT_FLAG)
+      fuzzer_utils.extract_argument(arguments, constants.RUNS_FLAG)
+      arguments.append('%s%d' % (constants.TIMEOUT_FLAG,
+                                 constants.REPRODUCTION_TIMEOUT_LIMIT))
+      result = runner.run_single_testcase(
+          input_path,
+          timeout=constants.REPRODUCTION_TIMEOUT_LIMIT + 5,  # processing buffer
```
Dor1s: I'd give even more than 5 here, maybe 20?

src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -317,6 +317,22 @@ def reproduce(self, target_path, input_path, arguments, max_time):
 
     result = runner.run_single_testcase(
         input_path, timeout=max_time, additional_args=arguments)
+
+    # If this is a timeout crash, then re-run with a higher limit to make
+    # sure it reproduces.
+    if constants.TEST_TIMEOUT_SIGNATURE in result.output:
```
oliverchang: hmm, if we are going to do this, why don't we always do this for all reproduce runs?

will keep the logic simpler and less hacky.

oliverchang: And in that case, the cleanest way would be to fix up the timeout arguments near here: https://github.com/google/clusterfuzz/blob/3ed28b540e95ec81d9a0b6f0d3d33b483f29d6ec/src/python/bot/fuzzers/libFuzzer/engine.py#L271 for future testcase, and remove this hack here.

We would need to perform a migration in that case, which is more work but might be good to prevent any unforeseen breakages.

inferno-chromium:
As per discussion with @oliverchang , moving to a better non-hacky change in https://github.com/google/clusterfuzz/pull/1160

PR #1163, opened 2019-11-01
Fix missing command in untrusted runner engine_reproduce code.

### LINKED ISSUE #1162: Command field empty in OSS-Fuzz testcases
See https://oss-fuzz.com/testcase-detail/5204819744915456 for example.

inferno-chromium:
Fixes #1162

### DIFF
--- a/src/python/tests/core/bot/testcase_manager_test.py
+++ b/src/python/tests/core/bot/testcase_manager_test.py
@@ -760,6 +760,7 @@ class UntrustedEngineReproduceTest(
     """Set up."""
     super(UntrustedEngineReproduceTest, self).setUp()
     environment.set_value('JOB_NAME', 'libfuzzer_asan_job')
+    environment.set_value('USE_MINIJAIL', False)
 
     job = data_types.Job(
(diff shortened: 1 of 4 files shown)


src/python/tests/core/bot/testcase_manager_test.py
```diff
@@ -758,6 +760,7 @@ def setUp(self):
     """Set up."""
     super(UntrustedEngineReproduceTest, self).setUp()
     environment.set_value('JOB_NAME', 'libfuzzer_asan_job')
+    environment.set_value('USE_MINIJAIL', False)
```
inferno-chromium: @oliverchang - fyi, there is some flakiness in CI here. USE_MINIJAIL=True was carried behind from corpus pruning untrusted test. i could not find root cause, i see patch_environ there. locally it does not fail.

inferno-chromium: @oliverchang - there is some carryover from corpus pruning test here. if i remove this, tests fail on CI weirdly, even though locally they succeed, if you can please help to make a quick check, that would be appreciated.

PR #1164, opened 2019-11-01
[Reproduce tool] Add Android emulator support.

mbarbella-chromium:
PTAL.

At this point, running the emulator works fine as does installing a test app and the rest of our workflow. For ASanified chrome in particular, I'm seeing a consistent crash in libGLESv2_enc.so that looks emulator-specific. This is good in that it shows we're able to run and detect crashes, but bad in that we won't be able to detect the right ones yet. That issue should be unrelated to this change though, so we can land this then resolve it after the fact.


src/local/butler/reproduce_tool/android.py
```diff
@@ -0,0 +1,112 @@
...
+  # another device is connected (as we don't know the serial yet). Find a better
+  # solution.
+  time.sleep(30)
+
+  return emulator_process
+
+
+def get_devices():
+  """Get a list of all connected Android devices."""
+  adb_runner = new_process.ProcessRunner(adb.get_adb_path())
```
oliverchang: what happens if run-server is already running? will it keep accumulating?

src/local/butler/reproduce_tool/android.py
```diff
@@ -0,0 +1,112 @@
...
+        'Android device with serial {serial} that could result in data '
+        'loss.'.format(serial=serial))
+  willing_to_continue = prompts.get_boolean(
+      'Are you sure you want to continue?')
+  if not willing_to_continue:
+    raise errors.ReproduceToolUnrecoverableError(
+        'Bailing out to avoid changing settings on the connected device.')
+
+  if 'emulator' in serial:
+    adb.run_command(['remount'])
```
oliverchang: could you please add a comment here explaining why this is only needed for the emulator case?

src/local/butler/reproduce_tool/android.py
```diff
@@ -0,0 +1,112 @@
...
+from local.butler.reproduce_tool import errors
+from local.butler.reproduce_tool import prompts
+from platforms.android import adb
+from platforms.android import device
+from system import environment
+from system import new_process
+
+ADB_DEVICES_SEPARATOR_STRING = 'List of devices attached'
+EMULATOR_PATH = os.path.join(
+    os.path.dirname(__file__), os.pardir, os.pardir, os.pardir, os.pardir,
```
oliverchang: this looks a bit messy. how about using ROOT_DIR ?

src/local/butler/reproduce_tool/android.py
```diff
@@ -0,0 +1,112 @@
...
+      EMULATOR_PATH,
+      ['-avd', 'Android28', '-writable-system', '-partition-size', '2048'])
+  emulator_process = runner.run()
+
+  # If we run adb commands too soon after the emulator starts, we may see
+  # flake or errors. Delay a short while to account for this.
+  # TODO(mbarbella): This is slow and flaky, but wait-for-device isn't usable if
+  # another device is connected (as we don't know the serial yet). Find a better
+  # solution.
+  time.sleep(30)
```
oliverchang: can we start the emulator with a predictable serial?

https://developer.android.com/studio/run/emulator-commandline says there's 

```
 -prop ro.serialno=123456
```

butler.py
```diff
@@ -298,6 +298,11 @@ def main():
       '--verbose',
       action='store_true',
       help='Print additional log messages while running.')
+  parser_reproduce.add_argument(
+      '-e',
+      '--emulator',
+      action='store_true',
+      help='Run and attempt to reproduce using the Android emulator.')
```
inferno-chromium: s/Run and attempt to reproduce/Reproduce a crash ?

local/install_deps_linux.bash
```diff
@@ -152,6 +152,20 @@ pip install --upgrade pip
...
 pip install --upgrade -r src/local/requirements.txt
 
+# Install the Android emulator and its dependencies. Used in tests and as an
+# option during Android test case reproduction.
+rm -rf local/bin/android-sdk
+mkdir local/bin/android-sdk
+curl https://dl.google.com/android/repository/sdk-tools-linux-4333796.zip --output local/bin/android-sdk/sdk-tools-linux.zip
+unzip -d local/bin/android-sdk local/bin/android-sdk/sdk-tools-linux.zip
+
+tools_bin=local/bin/android-sdk/tools/bin/
```
inferno-chromium: ANDROID_TOOLS_BIN (capital) ?

local/install_deps_linux.bash
```diff
@@ -152,6 +152,20 @@ pip install --upgrade pip
 pip install --upgrade -r docker/ci/requirements.txt
 pip install --upgrade -r src/local/requirements.txt
 
+# Install the Android emulator and its dependencies. Used in tests and as an
+# option during Android test case reproduction.
```
inferno-chromium: Use local ANDROID_SDK_INSTALL_DIR=local/bin/android-sdk
Also version 28 and 4333796 should be local vars that can be easily tweaked ANDROID_SDK_VERSION=28 and ANDROID_SDK_REVISION=4333796

local/install_deps_linux.bash
```diff
@@ -152,6 +152,20 @@ pip install --upgrade pip
 pip install --upgrade -r docker/ci/requirements.txt
 pip install --upgrade -r src/local/requirements.txt
 
+# Install the Android emulator and its dependencies. Used in tests and as an
```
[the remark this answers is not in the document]
inferno-chromium: +1 probably hide this behind a command line flag.

local/install_deps_linux.bash
```diff
@@ -152,6 +152,20 @@ pip install --upgrade pip
...
+# option during Android test case reproduction.
+rm -rf local/bin/android-sdk
+mkdir local/bin/android-sdk
+curl https://dl.google.com/android/repository/sdk-tools-linux-4333796.zip --output local/bin/android-sdk/sdk-tools-linux.zip
```
inferno-chromium: nit: can wrap this line.

reproduce.sh
```diff
@@ -18,7 +18,7 @@ CLUSTERFUZZ_CONFIG_DIR=~/.config/clusterfuzz
 ROOT_DIRECTORY=$(dirname $(readlink -f "$0"))
 
 mkdir -p $CLUSTERFUZZ_CONFIG_DIR
-if [ ! -d $ROOT_DIRECTORY/ENV ]; then
+if [ ! -d $ROOT_DIRECTORY/ENV ] || [ ! -d $ROOT_DIRECTORY/local/bin/android-sdk ]; then
```
inferno-chromium: This should be optional, as per comment above.

src/local/butler/reproduce.py
```diff
@@ -423,6 +399,12 @@ def execute(args):
   # Initialize fuzzing engines.
   init.run()
 
+  # Prepare the emulator if needed.
+  emulator_process = None
+  if args.emulator:
+    print('Starting emulator...')
+    emulator_process = android.start_emulator()
```
inferno-chromium: We can fail in start_emulator if it is not installed and provide instuction to people to install it.

src/local/butler/reproduce_tool/android.py
inferno-chromium: +1

src/local/butler/reproduce_tool/android.py
```diff
@@ -0,0 +1,112 @@
...
+EMULATOR_PATH = os.path.join(
+    os.path.dirname(__file__), os.pardir, os.pardir, os.pardir, os.pardir,
+    'local', 'bin', 'android-sdk', 'emulator', 'emulator')
+
+
+def start_emulator():
+  """Return a ProcessRunner configured to start the Android emulator."""
+  runner = new_process.ProcessRunner(
+      EMULATOR_PATH,
+      ['-avd', 'Android28', '-writable-system', '-partition-size', '2048'])
```
inferno-chromium: Can we generalize Android28, we don't want to change all these when upgrading sdk.

src/local/butler/reproduce_tool/android.py
inferno-chromium: ++++1, predictable serial will be nice, user will be unhappy with 30 sec wait, too long.

src/local/butler/reproduce_tool/android.py
inferno-chromium: i think adb devices already starts it if needed, so can just remove that. Also, that addresses Oliver's concerns.

src/local/butler/reproduce_tool/constants.py
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Constants used by the reproduce tool."""
+
+import os
+
```
inferno-chromium: Feel like more things can be moved here.

src/local/butler/reproduce_tool/android.py
mbarbella-chromium: Great suggestion. Thanks!

mbarbella-chromium: Unfortunately this doesn't seem to change how ADB recognizes the device. After setting that it still only recognized it as "emulator-5554", and running commands with -s 123456 failed.

src/local/butler/reproduce_tool/constants.py
mbarbella-chromium: I deleted this file for the time being. It might make sense to refactor some things and use this later, but I originally intended to use this for a few things in this change before realizing it was unnecessary.

src/local/butler/reproduce_tool/android.py
mbarbella-chromium: This is actually unnecessary, so I've removed it. It's a leftover from an issue I was debugging.

src/local/butler/reproduce_tool/android.py
mbarbella-chromium: Done

src/local/butler/reproduce.py
mbarbella-chromium: With reproduce.sh this shouldn't happen since we'll detect it and install it automatically. There's some argument for providing instructions in every case if people try to use butler.py directly, but I don't want to special-case this one here.

local/install_deps_linux.bash
mbarbella-chromium: Done

src/local/butler/reproduce_tool/android.py
mbarbella-chromium: This probably isn't necessary. It restarts an existing server if one is already running, but I did this explicitly here to avoid issues with an already running server for a different version than what ClusterFuzz expects. ClusterFuzz will restart it automatically, but our first command will be much slower and have some extra output (though I think the code to run "adb devices" here will handle that fine).

For now I've removed this.

src/local/butler/reproduce_tool/http_utils.py
```diff
@@ -29,8 +29,8 @@
 
 CONFIG_DIRECTORY = os.path.join(
```
inferno-chromium: Shouldn't this be used from constants.py

reproduce.sh
```diff
@@ -17,13 +17,27 @@
 CLUSTERFUZZ_CONFIG_DIR=~/.config/clusterfuzz
 ROOT_DIRECTORY=$(dirname $(readlink -f "$0"))
 
+# Process command line arguments.
```
inferno-chromium: shouldn't --emulator be used here  instead of --install-android-emulator, otherwise it means we provide both arguments when running reproduce.sh. --emulator could mean that we send --install-android-emulator

src/local/butler/reproduce_tool/http_utils.py
mbarbella-chromium: I removed constants.py from this change (see the resolved comment thread). I originally intended to use CONFIG_DIRECTORY and some other variables from android.py as well, but realized it was unnecessary.

reproduce.sh
mbarbella-chromium: emulator-install-flag might be a confusing name. I'll change it to something like "additional_deps_args". I believe the logic here is what we want, so I'm not sure I understand the concern completely.

PR #1171, opened 2019-11-06
Don't check for trial args for engine fuzzer jobs

inferno-chromium:
Skips a db call for an unneeded case.


inferno-chromium:
TBR

PR #1173, opened 2019-11-06
Speed up cross pollinate fuzzer list operation.

inferno-chromium:
Previously we were query a job entity for every fuzz target. Now it only queries jobs for the selected targets+jobs. This has slight downside of getting an empty list when jobs don't have backup bucket, but in most cases, we do have one (also default bucket in project.yaml is used).
Also, use the default backup bucket from project.yaml when available, otherwise we wont pollinate for cases when a default one is used everywhere.


inferno-chromium:
This speeds up this function from 5 min to 15 sec.

jonathanmetzman:
LGTM as well.

Issue #1128, opened 2019-10-24
Add support for honggfuzz


oliverchang:
Title.

oliverchang:
This was done.

Dor1s:
Minor piece missing on OSS-Fuzz side: https://github.com/google/oss-fuzz/issues/3326

PR #1174, opened 2019-11-07
Initial implementation of honggfuzz.

oliverchang:
Part of #1128.


src/python/bot/fuzzers/honggfuzz/engine.py
```diff
@@ -0,0 +1,169 @@
...
+import glob
+import os
+import re
+
+from bot.fuzzers import dictionary_manager
+from bot.fuzzers import engine
+from system import environment
+from system import new_process
+
+_DEFAULT_ARGUMENTS = [
```
inferno-chromium: can move this and values 2048, 25 into its own vars in constants.py

src/python/bot/fuzzers/honggfuzz/engine.py
```diff
@@ -0,0 +1,169 @@
...
+    '25',
+]
+
+_CRASH_REGEX = re.compile('Crash: saved as \'(.*)\'')
+_HF_SANITIZER_LOG_PREFIX = 'HF.sanitizer.log'
+
+
+def _get_runner():
+  """Get the honggfuzz runner."""
+  honggfuzz_path = os.path.join(environment.get_value('BUILD_DIR'), 'honggfuzz')
```
inferno-chromium: nit: maybe raise an exception if hongfuzz_path does not exist.

src/python/bot/fuzzers/honggfuzz/engine.py
```diff
@@ -0,0 +1,169 @@
...
+  return new_process.ProcessRunner(honggfuzz_path)
+
+
+def _find_sanitizer_stacktrace(reproducers_dir):
+  """Find the sanitizer stacktrace from the reproducers dir."""
+  for stacktrace_path in glob.glob(
+      os.path.join(reproducers_dir, _HF_SANITIZER_LOG_PREFIX + '*')):
+    with open(stacktrace_path, 'rb') as f:
+      return f.read()
```
inferno-chromium: Don't we want to append all logs ?

inferno-chromium:
@Dor1s as fyi since he plans AFL migration.

src/python/bot/fuzzers/honggfuzz/engine.py
```diff
@@ -0,0 +1,169 @@
...
+      target_path: Path to the target.
+      build_dir: Path to the build directory.
+
+    Returns:
+      A FuzzOptions object.
+    """
+    arguments = []
+    dict_path = dictionary_manager.get_default_dictionary_path(target_path)
+    if os.path.exists(dict_path):
+      arguments.extend(['--dict', dict_path])
```
inferno-chromium: '--dict' needs to be in constants.py as DICT_FLAG

src/python/bot/fuzzers/honggfuzz/engine.py
```diff
@@ -0,0 +1,169 @@
...
+      max_time: Maximum allowed time for the fuzzing to run.
+
+   Returns:
+      A FuzzResult object.
+    """
+    runner = _get_runner()
+    arguments = _DEFAULT_ARGUMENTS[:]
+    arguments.extend(options.arguments)
+    arguments.extend([
+        '-f',
```
inferno-chromium: nit: does it have longer, more readable option, similar to --run_time ?

src/python/bot/fuzzers/honggfuzz/engine.py
```diff
@@ -0,0 +1,169 @@
...
+        '-W',
+        reproducers_dir,
+        '--run_time',
+        str(max_time),
+        '--',
+        target_path,
+    ])
+
+    fuzz_result = runner.run_and_wait(
+        additional_args=arguments, timeout=max_time + 10)
```
inferno-chromium: move 10 to global or constants.py

src/python/tests/core/bot/fuzzers/honggfuzz/honggfuzz_engine_test.py
```diff
@@ -0,0 +1,147 @@
...
+        '--run_time',
+        '10',
+        '--',
+        target_path,
+    ], results.command)
+
+    self.assertIn('Seen a crash. Terminating all fuzzing threads', results.logs)
+    self.assertEqual(1, len(results.crashes))
+    crash = results.crashes[0]
+    self.assertEqual(TEMP_DIR, os.path.dirname(crash.input_path))
```
inferno-chromium: need to add check to make sure that crash.input_input exists on disk.

src/python/bot/fuzzers/honggfuzz/engine.py
oliverchang: I don't think we need a constants.py here :) I think it's better to have vars closer to where they're used for readability, unless they're shared. In this case this file should be quite self contained. Happy to change this if we start needing these vars elsewhere or if you feel strongly.

(moved the timeout and rss limit into their own vars).

src/python/bot/fuzzers/honggfuzz/engine.py
oliverchang: There should only be one, so no need to make the code more complex.

src/python/bot/fuzzers/honggfuzz/engine.py
oliverchang: Why is dict special compared to all other arguments? :) (This is also the only occurrence).

src/python/bot/fuzzers/honggfuzz/engine.py
oliverchang: Yes, done!

src/python/bot/fuzzers/honggfuzz/engine.py
oliverchang: Put in a global.

src/python/tests/core/bot/fuzzers/honggfuzz/honggfuzz_engine_test.py
oliverchang: Done!

src/python/bot/fuzzers/honggfuzz/engine.py
inferno-chromium: it is ok, this was just a nit.

src/python/bot/fuzzers/honggfuzz/engine.py
inferno-chromium: Ok.

src/python/bot/fuzzers/honggfuzz/engine.py
inferno-chromium: Ok, here is fine.

src/python/bot/fuzzers/honggfuzz/engine.py
oliverchang: Done!

oliverchang:
/gcbrun. Also added .cc files.

PR #1175, opened 2019-11-07
Fix chrome stack parsing regexes to not allow unsymbolized functions.

inferno-chromium:
These regexes were only supposed to parse symbolized frames.
Unsymbolized frames need to be ignored, so don't use if path
chars / or \ are found.

### DIFF
--- a/src/python/crash_analysis/stack_parsing/stack_analyzer.py
+++ b/src/python/crash_analysis/stack_parsing/stack_analyzer.py
@@ -81,9 +81,9 @@ CHROME_CHECK_FAILURE_REGEX = re.compile(
 CHROME_STACK_FRAME_REGEX = re.compile(
     r'[ ]*(#(?P<frame_id>[0-9]+)[ ]'  # frame id (2)
     r'([xX0-9a-fA-F]+)[ ])'  # addr (3)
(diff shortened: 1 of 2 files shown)


inferno-chromium:
> If all tests are passing this seems reasonable to me. I'm having trouble thinking of any cases where it would be problematic.

Yes i think original intent was always to do symbolized here
```
    # Chrome symbolized stack frame regex.
    if add_frame_on_match(
        CHROME_STACK_FRAME_REGEX,
        line,
        state,
        group=4,
        frame_spec=CHROME_STACK_FRAME_SPEC):
      continue

    # Chrome symbolized stack frame regex (Mac only).
    if add_frame_on_match(
        CHROME_MAC_STACK_FRAME_REGEX,
        line,
        state,
        group=6,
        demangle=True,
        frame_spec=CHROME_MAC_STACK_FRAME_SPEC):
      continue

    # Chrome symbolized stack frame regex (Windows only).
    if add_frame_on_match(
        CHROME_WIN_STACK_FRAME_REGEX,
        line,
        state,
        group=1,
        frame_spec=CHROME_WIN_STACK_FRAME_SPEC):
      continue
```

PR #1179, opened 2019-11-08
Remove PYTHONPATH when running gsutil.

oliverchang:
gsutil may be on Python 3, which won't work with our Python 2
PYTHONPATH.

### DIFF
--- a/src/python/google_cloud_utils/gsutil.py
+++ b/src/python/google_cloud_utils/gsutil.py
@@ -96,7 +96,7 @@ class GSUtilRunner(object):
     self.gsutil_runner = _process_runner(
         _get_gsutil_path(), default_args=default_gsutil_args)
 
-  def run_gsutil(self, arguments, timeout=None, quiet=False):
+  def run_gsutil(self, arguments, quiet=False, **kwargs):
     """Run GSUtil."""
     if quiet:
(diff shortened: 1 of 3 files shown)


oliverchang:
/gcbrun. @flowerhack FYI

oliverchang:
Fixed tests and some missed cases. /gcbrun

src/python/google_cloud_utils/gsutil.py
```diff
@@ -137,7 +143,7 @@ def rsync(self,
   def download_file(self, gcs_url, file_path, timeout=None):
     """Download a file from GCS."""
     command = ['cp', _filter_path(gcs_url), file_path]
-    result = self.gsutil_runner.run_and_wait(command, timeout=timeout)
+    result = self.run_gsutil(command, timeout=timeout)
```
inferno-chromium: We should probably have something in run_gsutil that converts command into a list. in future, if we add quiet flag, it will exception out.

oliverchang: Could you please clarify this comment? `command` is already a list?

inferno-chromium: ah sorry didn't know that, ignore.

PR #1185, opened 2019-11-11
Force python2 for gcloud deployment.

oliverchang:
There is a bug causing deployment to fail on Python 3.


inferno-chromium:
LGTM

PR #1187, opened 2019-11-11
Speculative fix for unicode issues in file_path_to_file_url.

mbarbella-chromium:


mbarbella-chromium:
TBR. Will land once tests are passing.

PR #1188, opened 2019-11-11
Use newstr in file_path_to_file_url.

mbarbella-chromium:
I didn't realize this file hadn't yet been migrated to Python 3-style strings. This feels a bit hackier than I'd like but it seems like a reasonable stopgap until we migrate the rest of src/python to newstr.

### DIFF
--- a/src/python/base/utils.py
+++ b/src/python/base/utils.py
@@ -156,8 +156,8 @@ def file_path_to_file_url(path):
   # TODO(mbarbella): urljoin has several type checks for arguments. Ensure that
   # we're passing newstr on both sides while migrating to Python 3. After
   # migrating, ensure that callers pass strs to avoid this hack.
-  return urllib.parse.urljoin(str(u'file:'),
-                              str(urllib.request.pathname2url(path)))
+  return urllib.parse.urljoin(
+      str(u'file:'), str(urllib.request.pathname2url(path)))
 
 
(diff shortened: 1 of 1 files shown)


src/python/base/utils.py
```diff
@@ -156,6 +156,7 @@ def file_path_to_file_url(path):
   # TODO(mbarbella): urljoin has several type checks for arguments. Ensure that
   # we're passing newstr on both sides while migrating to Python 3. After
   # migrating, ensure that callers pass strs to avoid this hack.
+  from builtins import str
```
inferno-chromium: Why not global context for the import. 

Also, is str(u'file') needed ? Adding @oliverchang for reviewing this as well.

mbarbella-chromium: I didn't want to move the whole file over yet, only to fix the exception. I'll move this out when we fully transition the bot code.

str(u'file') ensures that we're using newstr in both places. the second str() should be safe since pathname2url % encodes the string (so even if the original string was non-unicode we should be okay). I really don't like this as a fix, but since urljoin checks that certain conditions hold true for both arguments using the same type this is the least fragile solution I can think of in the meantime.

oliverchang: Do we need str() at all? newstr should be compatible with unicode right?

oliverchang:
LGTM  overall as a temporary fix.

PR #1190, opened 2019-11-12
Add env variable to avoid system temp cleanup.

inferno-chromium:
When running fuzzing on host machine (e.g. Android with physical
devices), we can interfere with other processes maintaining state
in temp directory. Provide an environment variable to disable the
cleanup.

### DIFF
--- a/src/python/system/shell.py
+++ b/src/python/system/shell.py
@@ -169,6 +169,11 @@ def clear_system_temp_directory():
     except:
       pass
 
+  if environment.get_value('SKIP_SYSTEM_TEMP_CLEANUP'):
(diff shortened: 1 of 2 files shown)


src/python/system/shell.py
```diff
@@ -169,6 +169,11 @@ def _delete_object(path, delete_func):
     except:
       pass
 
+  if environment.get_value('SKIP_SYSTEM_TEMP_CLEANUP'):
```
Dor1s: nit: might be moved to the very beginning of the function, but I guess it's just a style preference, feel free to ignore

inferno-chromium: I think we followed this convention in other places, will keep as-is.

PR #1193, opened 2019-11-12
[fuchsia] Filter out example fuzzers.

flowerhack:
Fuchsia has a set of example fuzzers that are useful for testing, but
which we only want selected if it's explicitly set via FUZZ_TARGET.

Otherwise, we only want "real" fuzzers to be picked during standard runs
(e.g. on Clusterfuzz bots).

### DIFF
--- a/src/python/build_management/build_manager.py
+++ b/src/python/build_management/build_manager.py
@@ -720,13 +720,6 @@ class RegularBuild(Build):
 
 class FuchsiaBuild(RegularBuild):
   """Represents a Fuchsia build."""
-  # If we properly subclass RegularBuild, checking revision numbers for
-  # different builds (and other nice things) should automagically "just work."
-  # This'll mean (1) downloading in the standard _build_dir instead of our
-  # artisinal FUCHSIA_RESOURCES_DIR, and (2) create new build buckets with
-  # the name format `gs://bucket/path/fuchsia-resources-([0-9]+).zip, and
-  # (3) do some refactoring on the RegularBuild logic so we can keep using our
-  # target selection logic.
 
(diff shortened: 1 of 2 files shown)


src/python/platforms/fuchsia/util/fuzzer.py
```diff
@@ -69,7 +69,18 @@ def _matches_sanitizer(cls, tgt, sanitizer):
...
 
   @classmethod
-  def filter(cls, fuzzers, name, sanitizer=None):
+  def _is_example(cls, pkg, tgt):
+    """ Returns whether or not a given pkg/tgt pair is an example fuzzer.
+    (Helper function to prevent us from wasting cycles on example fuzzers in
+    production). """
+    # We want to remove the noop-fuzzer regardless of the extension.
+    tgt = os.path.splitext(tgt)[0]
+    if pkg == 'example_fuzzers' or tgt == 'noop-fuzzer':
```
oliverchang: nit: might be more concise as 

return (pkg == .. or tgt == )

src/python/build_management/build_manager.py
```diff
@@ -746,7 +739,7 @@ def _get_fuzz_targets_from_dir(self, build_dir):
         environment.get_value('JOB_NAME')).lower()
     return [
         str(target[0] + '/' + target[1])
-        for target in Fuzzer.filter(host.fuzzers, '', sanitizer)
+        for target in Fuzzer.filter(host.fuzzers, '', sanitizer, False)
```
oliverchang: nit: use example_fuzzers=False for better readability.

oliverchang:
Please do a reformat :) The line you just changed is now too long.

inferno-chromium:
I will do a refomat later, committing for now. CI would fail this CL is not rebased to https://github.com/google/clusterfuzz/pull/1203

PR #1194, opened 2019-11-14
Replace App Engine memcache with Cloud Memorystore Redis.

oliverchang:
Apart from NDB, this is the last dependency on the deprecated App Engine
SDK.

The default size (1GB) was picked based on our existing deployments, the
max of which is around 100~MB.

Also:
- Remove MemcacheLarge. It is no longer required as redis does not have
  such restrictive length restrictions on value size.
- Modify deployment to create the redis instance if needed, as well as
  the VPC connector.

This is currently only planned to be used on our App Engine instances,
and not bots.

### DIFF
--- a/configs/test/gae/prod/app.yaml
+++ b/configs/test/gae/prod/app.yaml
@@ -58,4 +58,4 @@ includes:
   - skip_files.yaml
 
 vpc_access_connector:
-  name: "projects/test-project/locations/us-central1/connectors/connector"
+  name: "projects/test-project/locations/gae-region/connectors/connector"
--- a/configs/test/gae/prod/app.yaml
+++ b/configs/test/gae/prod/app.yaml
@@ -56,3 +56,6 @@ basic_scaling:
 
 includes:
   - skip_files.yaml
+
+vpc_access_connector:
+  name: "projects/test-project/locations/us-central1/connectors/connector"
--- a/configs/test/redis/instance.yaml
+++ b/configs/test/redis/instance.yaml
@@ -1,0 +1,23 @@
+# Copyright 2019 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+imports:
+- path: memorystore-instance.jinja
+
+resources:
+- name: redis
+  type: memorystore-instance.jinja
+  properties:
+    region: us-central1
+    memorySizeGb: 1
--- a/configs/test/redis/instance.yaml
+++ b/configs/test/redis/instance.yaml
@@ -19,5 +19,5 @@ resources:
 - name: redis
   type: memorystore-instance.jinja
   properties:
-    region: us-central1
+    region: gae-region
     memorySizeGb: 1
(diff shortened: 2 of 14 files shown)


configs/test/gae/prod/app.yaml
```diff
@@ -56,3 +56,6 @@ basic_scaling:
 
 includes:
   - skip_files.yaml
+
+vpc_access_connector:
+  name: "projects/test-project/locations/us-central1/connectors/connector"
```
inferno-chromium: We need to use "gce-zone" instead of us-central1 (just like project.yaml). That way it gets auto-subsituted to the right zone.

configs/test/redis/instance.yaml
```diff
@@ -0,0 +1,23 @@
...
+# limitations under the License.
+
+imports:
+- path: memorystore-instance.jinja
+
+resources:
+- name: redis
+  type: memorystore-instance.jinja
+  properties:
+    region: us-central1
```
inferno-chromium: gce-zone

src/local/butler/deploy.py
```diff
@@ -65,6 +65,21 @@ def _get_services(paths):
   return services
 
 
+def _get_redis_ip(project):
+  """Get the redis IP address."""
+  _, ip = common.execute('gcloud redis instances describe redis-instance '
+                         '--project={project} --region=us-central1 '
```
inferno-chromium: some way to not hardcode us-central1

configs/test/gae/prod/app.yaml
oliverchang: Added "gae-region" and replaced with that. This needs to be in the gae region.

configs/test/redis/instance.yaml
oliverchang: Same as above :)

src/local/butler/deploy.py
oliverchang: Done!

src/local/butler/appengine.py
```diff
@@ -154,3 +157,20 @@ def symlink_config_dir():
...
+  if not location[-1].isdigit():
+    # e.g. us-central -> us-central1
+    location += '1'
+
+  return location
+
+
+def region(project):
+  """Get the App Engine region."""
+  _, location = common.execute(
```
inferno-chromium: might be good to check return code and return None if failed ?

src/local/butler/deploy.py
```diff
@@ -65,6 +65,23 @@ def _get_services(paths):
...
 
 
+def _get_redis_ip(project):
+  """Get the redis IP address."""
+  region = appengine.region(project)
+  _, ip = common.execute('gcloud redis instances describe redis-instance '
+                         '--project={project} --region={region} '
+                         '--format="value(host)"'.format(
+                             project=project, region=region))
+  return ip.strip()
```
inferno-chromium: check return code of common.execute, if failed, return None ?

src/local/butler/deploy.py
```diff
@@ -236,6 +256,25 @@ def _update_bigquery(project):
...
+                             os.path.join('redis', 'instance.yaml'))
+
+  region = appengine.region(project)
+  return_code, _ = common.execute(
+      'gcloud compute networks vpc-access connectors describe '
+      'connector --region={region} '
+      '--project={project}'.format(project=project, region=region),
+      exit_on_error=False)
+
+  if return_code != 0:
```
inferno-chromium: nit: Add a comment that this means it is already created.

src/local/butler/appengine.py
oliverchang: Raised an exception here. We can't continue in that case.

src/local/butler/deploy.py
oliverchang: Raised an exception here. We can't continue in that case.

src/local/butler/deploy.py
oliverchang: Done!

PR #1196, opened 2019-11-15
Add honggfuzz to project setup.

oliverchang:
#1128.

### DIFF
--- a/src/appengine/handlers/cron/project_setup.py
+++ b/src/appengine/handlers/cron/project_setup.py
@@ -119,8 +119,11 @@ AFL_ASAN_JOB = JobInfo(
     minimize_job_override=LIBFUZZER_ASAN_JOB)
 NO_ENGINE_ASAN_JOB = JobInfo('asan_', 'none', 'address', [])
 
-HONGGFUZZ_ASAN_JOB = JobInfo('honggfuzz_asan_', 'honggfuzz', 'address',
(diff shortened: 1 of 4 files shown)


src/appengine/handlers/cron/project_setup.py
```diff
@@ -119,6 +119,12 @@ def job_name(self, project_name):
     minimize_job_override=LIBFUZZER_ASAN_JOB)
 NO_ENGINE_ASAN_JOB = JobInfo('asan_', 'none', 'address', [])
 
+HONGGFUZZ_ASAN_JOB = JobInfo(
+    'honggfuzz_asan_',
+    'honggfuzz',
+    'address', ['honggfuzz', 'engine_asan'],
+    minimize_job_override=LIBFUZZER_ASAN_JOB)
```
inferno-chromium: Eventually we want to support native minimization using fuzzing engine itself, does honggfuzz support that ?

oliverchang: I don't think so.. (https://github.com/google/honggfuzz/blob/master/docs/USAGE.md)

oliverchang:
> LGTM.
> 
> I think we need to update environment.is_engine_fuzzer_job and any other similar places that check against libfuzzer, afl. like search for is_afl_job, is_libfuzzer_job

Done in https://github.com/google/clusterfuzz/pull/1197.

PR #1197, opened 2019-11-15
Add honggfuzz to is_engine_fuzzer_job.

oliverchang:
We should generalize this in the future instead of hardcoding these.

### DIFF
--- a/src/python/system/environment.py
+++ b/src/python/system/environment.py
@@ -597,9 +597,16 @@ def is_libfuzzer_job(job_name=None):
   return _job_substring_match('libfuzzer', job_name)
 
 
+def is_honggfuzz_job(job_name=None):
+  """Return true if the current job uses honggfuzz."""
+  return _job_substring_match('honggfuzz', job_name)
+
+
 def is_engine_fuzzer_job(job_name=None):
   """Return if true is this is an engine fuzzer."""
-  return is_afl_job(job_name) or is_libfuzzer_job(job_name)
+  # TODO(ochang): Generalize this rather than hardcoding all these engines.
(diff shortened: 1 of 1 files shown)


oliverchang:
This is a quick fix. Will generalize this with a better solution next week.

oliverchang:
> this should fix most of the places, just need to search src/appengine pages for libfuzzer, afl too.

Do you mean explicit checks to is_afl_job/is_libfuzzer_jobs ? These don't seem relevant:

```python
src/python/bot/tasks/minimize_task.py:  if environment.is_libfuzzer_job():
src/python/bot/tasks/minimize_task.py-    do_libfuzzer_minimization(testcase, testcase_file_path)
src/python/bot/tasks/minimize_task.py-    return
--
src/python/bot/tasks/setup.py:  if environment.is_afl_job(job_type):
src/python/bot/tasks/setup.py-    return '%TESTCASE%'
src/python/bot/tasks/setup.py-
--
src/python/system/environment.py:def is_afl_job(job_name=None):
src/python/system/environment.py-  """Return true if the current job uses AFL."""
src/python/system/environment.py-  # Prefix matching is not sufficient.
--
src/python/system/environment.py:def is_libfuzzer_job(job_name=None):
src/python/system/environment.py-  """Return true if the current job uses libFuzzer."""
src/python/system/environment.py-  # Prefix matching is not sufficient.
--
src/python/system/environment.py:  return (is_afl_job(job_name) or is_libfuzzer_job(job_name) or
src/python/system/environment.py-          is_honggfuzz_job(job_name))
src/python/system/environment.py-
```

PR #1202, opened 2019-11-18
Add a signature for missing library crash.

inferno-chromium:
This covers cases when we fail to archive shared library
dependencies in the build dependencies. This is prerequisite
work for capturing startup crashes in libFuzzer targets.


src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_data/missing_library.txt
```diff
@@ -0,0 +1 @@
+CANNOT LINK EXECUTABLE "/build/vts_proto_fuzzer": library "libvts_codecoverage.so" not found
```
Dor1s: where is this message from? sounds to me like a build-time error, not a runtime. In runtime I'd expect something like `Failed to load libvts_codecoverage.so`

inferno-chromium: It is a Android bionic load failure message
```
if (needed_libraries_count > 0 &&
      !find_libraries(si, needed_library_names, needed_libraries_count, nullptr,
          &g_ld_preloads, ld_preloads_count, RTLD_GLOBAL, nullptr)) {
    __libc_format_fd(2, "CANNOT LINK EXECUTABLE: %s\n", linker_get_error_buffer());
    exit(EXIT_FAILURE);
  } else if (needed_libraries_count == 0) {
    if (!si->link_image(g_empty_list, soinfo::soinfo_list_t::make_list(si), nullptr)) {
      __libc_format_fd(2, "CANNOT LINK EXECUTABLE: %s\n", linker_get_error_buffer());
      exit(EXIT_FAILURE);
    }
    si->increment_ref_count();
  }
```
Adding another signature for linux.

inferno-chromium:
@Dor1s - can you take another look.

PR #1206, opened 2019-11-20
Don't set FUZZ_TARGET_COUNT when FUZZ_TARGET is already set.

oliverchang:
In the case of split target builds,
_set_random_fuzz_target_for_fuzzing_if_needed may be called twice: first
to actually pick the target, and again when we defer to the regular
build setup, which will incorrectly set the count as there is now only
1.


src/python/build_management/build_manager.py
```diff
@@ -279,6 +276,9 @@ def _set_random_fuzz_target_for_fuzzing_if_needed(fuzz_targets, target_weights):
     logs.log_error('No fuzz targets found. Unable to pick random one.')
     return None
 
+  fuzz_targets = list(fuzz_targets)
```
inferno-chromium: probably move this line before line:275 since it can be iterator. environment.set_value can be here as-is.

PR #1212, opened 2019-11-21
Increase unused fuzz target removal threshold.

inferno-chromium:
If bots are down for a week, we don't want existing testcases to
be closed. Increase the time period for now.
Also, remove a chromium top crasher hack.


inferno-chromium:
TBR minor fix.

PR #1213, opened 2019-11-22
Add crash signature for libFuzzer overwrites-const-input.

inferno-chromium:
Also, remove an obsolete hang crash type. This was used in Chromium
long time back and is no longer present on open testcases.

### DIFF
--- a/src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_data/libfuzzer_overwrites_const_input.txt
+++ b/src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_data/libfuzzer_overwrites_const_input.txt
@@ -1,0 +1,11 @@
+==1== ERROR: libFuzzer: fuzz target overwrites its const input
+    #0 0x52a761 in __sanitizer_print_stack_trace /src/llvm/projects/compiler-rt/lib/asan/asan_stack.cpp:86:3
+    #1 0x4746f8 in fuzzer::PrintStackTrace() /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerUtil.cpp:205:5
+    #2 0x45b483 in fuzzer::Fuzzer::CrashOnOverwrittenData() /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:518:3
+    #3 0x45a49c in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:564:5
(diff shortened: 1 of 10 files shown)


src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_data/libfuzzer_overwrites_const_input.txt
```diff
@@ -0,0 +1,11 @@
+==1== ERROR: libFuzzer: fuzz target overwrites its const input
+    #0 0x52a761 in __sanitizer_print_stack_trace /src/llvm/projects/compiler-rt/lib/asan/asan_stack.cpp:86:3
+    #1 0x4746f8 in fuzzer::PrintStackTrace() /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerUtil.cpp:205:5
+    #2 0x45b483 in fuzzer::Fuzzer::CrashOnOverwrittenData() /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:518:3
```
inferno-chromium: The stack is not usable since crash stacktrace is at time of ExecuteCallback, so we dont get any useful frames. That is why we show fuzz target name as crash state instead of NULL. 

```
void Fuzzer::ExecuteCallback(const uint8_t *Data, size_t Size) {
  TPC.RecordInitialStack();
  TotalNumberOfRuns++;
  assert(InFuzzingThread());
  // We copy the contents of Unit into a separate heap buffer
  // so that we reliably find buffer overflows in it.
  uint8_t *DataCopy = new uint8_t[Size];
  memcpy(DataCopy, Data, Size);
  if (EF->__msan_unpoison)
    EF->__msan_unpoison(DataCopy, Size);
  if (EF->__msan_unpoison_param)
    EF->__msan_unpoison_param(2);
  if (CurrentUnitData && CurrentUnitData != Data)
    memcpy(CurrentUnitData, Data, Size);
  CurrentUnitSize = Size;
  {
    ScopedEnableMsanInterceptorChecks S;
    AllocTracer.Start(Options.TraceMalloc);
    UnitStartTime = system_clock::now();
    TPC.ResetMaps();
    RunningUserCallback = true;
    int Res = CB(DataCopy, Size);
    RunningUserCallback = false;
    UnitStopTime = system_clock::now();
    (void)Res;
    assert(Res == 0);
    HasMoreMallocsThanFrees = AllocTracer.Stop();
  }
  if (!LooseMemeq(DataCopy, Data, Size))
    CrashOnOverwrittenData();
  CurrentUnitSize = 0;
  delete[] DataCopy;
}
```

Dor1s: ack, thanks for clarifying!

PR #1218, opened 2019-11-25
Remove uneeded LD_LIBRARY_PATH for Android engine_test.

inferno-chromium:
This is not needed for newer HWASan builds.


inferno-chromium:
TBR minor test fix.

Issue #1220, opened 2019-11-25
Extend parsing signatures for UBSan's pointer-overflow check


Dor1s:
It's been recently extended in LLVM upstream: https://github.com/llvm/llvm-project/blame/master/compiler-rt/lib/ubsan/ubsan_handlers.cpp#L689

For an example, see https://bugs.chromium.org/p/chromium/issues/detail?id=1027893# and https://clusterfuzz.com/testcase-detail/5654796597854208

inferno-chromium:
Marty is planning to use "SUMMARY:" field for these, should just get bucketized to same "Pointer-overflow"

mbarbella-chromium:
It looks like we already have a catch-all in place for any UBSan runtime error which will result in an UNKNOWN crash type if nothing else is detected (following up on our discussion from last week). We have a few crashes in this bucket for other rules. I'm much less worried about a general solution now and it feels weird to have two different ways to detect these, so I think I'll just put a change together to update pointer overflow and some of the other ones that we're not detecting now.

inferno-chromium:
We can still update the UNKNOWN, when we see the summary field ? I think it is more future proof than using the UNKNOWN. Only think i worry about is we might tag them security if we don't know the crash type (which seems wrong). @mbarbella-chromium @Dor1s - thoughts ?

mbarbella-chromium:
The case I'm worried about is tagging inconsistently. Maybe we should log an error if we report an UNKNOWN for ubsan so that we know a signature update is needed?

Dor1s:
The summary line in the report I linked says the following:

```
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior <filepath>:<line_no>:<column_no> in
--

some UBSan issues are missing one argument (error type) and the summary line is weird and barely helpful in such cases.

I thought @hctim wanted to fix that but I might be making that up.

Logging an error sounds like a good way to monitor any future things coming!

inferno-chromium:
Yes logging error sounds good! And lets add parsing signatures for all!

hctim:
Had a grep across my emails - couldn't find anything but I do remember us discussing this.

What are you thinking you'd like here? Is it something like this:
`SUMMARY: UndefinedBehaviorSanitizer: applying non-zero offset 123 to null pointer <filepath>:<line_no>:<column_no>`?

Dor1s:
That was discussed in https://reviews.llvm.org/D67193#1658491

[...]

PR #1222, opened 2019-11-25
Add support for newer nullptr pointer-overflow checks.

mbarbella-chromium:
This will also start logging an error if we hit our catch all case for UBSan so we know that new signatures need to be added.


mbarbella-chromium:
PTAL

mbarbella-chromium:
This should fix issue #1220

Dor1s:
Nice, the logging appears to be already working!

```
Unknown UBSan crash type: null pointer passed as argument 1, which is declared to never be null
```

PR #1225, opened 2019-11-25
Add missing ubsan type regexes and improve existing ones.

mbarbella-chromium:
PTAL

### DIFF
--- a/src/python/crash_analysis/stack_parsing/stack_analyzer.py
+++ b/src/python/crash_analysis/stack_parsing/stack_analyzer.py
@@ -208,7 +208,9 @@ UBSAN_UNSIGNED_INTEGER_OVERFLOW_REGEX = re.compile(
 UBSAN_INTEGER_OVERFLOW_REGEX = re.compile(
     r'.*(integer overflow|negation of.*cannot be represented in type).*')
 UBSAN_INVALID_BOOL_VALUE_REGEX = re.compile(
-    r'.*not a valid value for type \'bool\'.*')
+    r'.*not a valid value for type \'(bool|BOOL)\'.*')
+UBSAN_INVALID_BUILTIN_REGEX = re.compile(r'.*, which is not a valid argument.*')
+UBSAN_INVALID_ENUM_VALUE_REGEX = re.compile(r'.*not a valid value for type.*')
 UBSAN_MISALIGNED_ADDRESS_REGEX = re.compile(r'.*misaligned address.*')
 UBSAN_NO_RETURN_VALUE_REGEX = re.compile(
     r'.*reached the end of a value-returning function.*')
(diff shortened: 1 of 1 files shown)


src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -208,7 +208,9 @@
 UBSAN_INTEGER_OVERFLOW_REGEX = re.compile(
     r'.*(integer overflow|negation of.*cannot be represented in type).*')
 UBSAN_INVALID_BOOL_VALUE_REGEX = re.compile(
-    r'.*not a valid value for type \'bool\'.*')
+    r'.*not a valid value for type \'(bool|BOOL)\'.*')
+UBSAN_INVALID_BUILTIN_REGEX = re.compile(r'.*, which is not a valid argument.*')
+UBSAN_INVALID_ENUM_VALUE_REGEX = re.compile(r'.*not a valid value for type.*')
```
inferno-chromium: This will conflict with UBSAN_INVALID_BOOL_VALUE_REGEX
add 
" \'(enum"

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -505,15 +511,22 @@
...
     (UBSAN_MISALIGNED_ADDRESS_REGEX, 'Misaligned-address'),
     (UBSAN_NO_RETURN_VALUE_REGEX, 'No-return-value'),
     (UBSAN_NULL_POINTER_READ_REGEX, 'Null-dereference READ'),
     (UBSAN_NULL_POINTER_REFERENCE_REGEX, 'Potential-null-reference'),
     (UBSAN_NULL_POINTER_WRITE_REGEX, 'Null-dereference WRITE'),
     (UBSAN_OBJECT_SIZE_REGEX, 'Object-size'),
     (UBSAN_POINTER_OVERFLOW_REGEX, 'Pointer-overflow'),
+    (UBSAN_RETURNS_NONNULL_ATTRIBUTE_REGEX, 'Invalid-null-return'),
     (UBSAN_SHIFT_ERROR_REGEX, 'Undefined-shift'),
+    (UBSAN_UNREACHABLE_REGEX, 'Unreachable code'),
```
inferno-chromium: Unreachable-code

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -505,15 +511,22 @@
...
     (UBSAN_VLA_BOUND_REGEX, 'Non-positive-vla-bound-value'),
+
+    # Since this covers a superset of 'Invalid-bool-value', it is intentionally
+    # placed at the end of the list.
+    (UBSAN_INVALID_ENUM_VALUE_REGEX, 'Invalid-enum-value'),
```
inferno-chromium: Why not fix UBSAN_INVALID_ENUM_VALUE_REGEX itself, see suggestion above.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
mbarbella-chromium: We use this type in other parts of the code. I'd prefer to be consistent.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
mbarbella-chromium: I believe this will actually use the enum type name. I don't think we can predict what those will be.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
mbarbella-chromium: Resolving this for now. If we can fix the enum regex that's obviously better but let's discuss on that thread.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
inferno-chromium: This is fine.

PR #1226, opened 2019-11-26
Remove the coverage module until the Python 3 migration is complete.

mbarbella-chromium:
It looks like newer versions still have this problem and I wasn't able to find a simple way to work around it. Let's just remove the module and re-enable it after migrating to Python 3.


mbarbella-chromium:
> Can you please file a tracking bug (Core feature label) for this. Also, link to python 2 to 3 migration bug.

Done!

PR #1228, opened 2019-11-26
Make unreproducible testcase text less confusing.

inferno-chromium:


inferno-chromium:
@kcc - wdyt ?

src/python/datastore/data_handler.py
```diff
@@ -74,12 +74,17 @@
     'https://{domain}/revisions?job={job_type}&revision={revision}')
 
 FILE_UNREPRODUCIBLE_TESTCASE_TEXT = (
+    '***************************************\n'
```
kcc: maybe
```
'*************** UNREPRODUCIBLE ************************\n'
```
(and the rest is great)

src/python/datastore/data_handler.py
```diff
@@ -74,12 +74,17 @@
...
     'Note: This crash might not be reproducible with the provided testcase. '
-    'That said, for the past %d days we\'ve been seeing this crash frequently. '
-    'If you are unable to reproduce this, please try a speculative fix based '
-    'on the crash stacktrace in the report. The fix can be verified by looking '
-    'at the crash statistics in the report, a day after the fix is deployed. '
-    'We will auto-close the bug if the crash is not seen for %d days.' %
+    'That said, for the past %d days, we\'ve been seeing this crash '
+    'frequently.\n\n'
+    'If you are unable to reproduce this, please try the following options:\n'
+    '- Run testcase multiple times for a longer duration.\n'
```
mbarbella-chromium: How often do we expect these first two options to work? I feel like suggesting a speculative fix might be the best advice. At the very least, I think we may want to set expectations that these first two options may still fail a bit more clearly.

inferno-chromium: Atleast some times, but we want to people to try things in that order. If running multiple times makes it consistent, then dont need to do speculative fix.

mbarbella-chromium: With the rest of the context I think it seems reasonable, thinking about it more. Maybe we could reword the bit on 81 to something like "It may be possible to reproduce by trying the following:", and split the speculative fix bit into its own paragraph below. I think that might make it a bit clearer that this could still fail.

I don't think it's too bad as-is if you have a preference for the current wording, though.

oliverchang:
Thanks! was planning on tackling this. I guess I'm a little concerned about making the text here even longer, but I guess "'*************** UNREPRODUCIBLE ************************" might make it OK.

PR #1231, opened 2019-11-27
Increase memory allocated to Fuchsia VMs to 3GB.

eepeep:
The previous value of 2GB was the same as the `rss_limit_mb`, so in
conditions with high memory usage the system would OOM before libFuzzer
could have a chance to catch it.


eepeep:
/gcbrun

cc @oliverchang

PR #1242, opened 2019-12-06
Factor out common schema columns.

oliverchang:


src/python/metrics/fuzzer_stats_schema.py
```diff
@@ -467,29 +441,7 @@
     'mode': 'NULLABLE',
     'name': 'speed',
```
Dor1s: is it different from `average_exec_per_sec` in libFuzzer / AFL?

oliverchang: Nope, but let's use the native column names to avoid confusion. In the actual fuzzer stats pages/queries etc we can give them the common names.

PR #1249, opened 2019-12-10
Add minimizer metric

mpherman2:
Add log output to report number of times minimizer runs testcase in order to compare minimizers better in the future.

### DIFF
--- a/src/python/bot/minimizer/minimizer.py
+++ b/src/python/bot/minimizer/minimizer.py
@@ -199,7 +199,7 @@ class Testcase(object):
 
   def _delete_file_if_needed(self, input_file):
     """Deletes a temporary file if necessary."""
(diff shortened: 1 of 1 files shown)


src/python/bot/minimizer/minimizer.py
```diff
@@ -216,8 +217,9 @@ def _report_progress(self):
       return
 
     self.last_progress_report_time = time.time()
-    message = '%d/%d tokens remaining.' % (len(self.get_required_tokens()),
-                                           len(self.required_tokens))
+    message = '%d/%d tokens remaining. - %d runs executed so far.' % (len(
```
inferno-chromium: no need of - in "- %d"

mpherman2: Removed the -

PR #1256, opened 2019-12-11
Revert "libFuzzer: use two step merge after fuzzing for accurate stats (#802)."

Dor1s:
Reverts google/clusterfuzz#815

Both internal and external instances seem to be reporting errors with some fuzz targets. Might be that those are using old libFuzzer, but might be a bug in the code I wrote. Reverting for now.


jonathanmetzman:
I was curious if something happened since AFL and LF are performing nearly the same this month (https://docs.google.com/spreadsheets/d/18q5JGzU2pGv4FJhRjeVkXGi3bZ2dRU7SxagC_aupzAo/edit#gid=1198170776) unlike nearly every other month.

Dor1s:
CI is happy, landing and deploying.

Dor1s:
@jonathanmetzman that is very suspicious indeed!

PR #1257, opened 2019-12-12
Add disable_ubsan attribute to control disabling UBSan in ASan builds.

inferno-chromium:


inferno-chromium:
fuzz_task plumbing will come in next cl.

inferno-chromium:
> LGTM, but any reason why this is in testcase attribute and not more generally in the job?

This will only disable ubsan for few % of runs in a ASan+UBSan build. This is similar to redzone, timeout_multiplier which is for few % of runs, we want to skip ubsan and get the more juicy ASan bugs.

PR #1258, opened 2019-12-12
Fix MergeError definition.

oliverchang:


oliverchang:
TBR

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -463,7 +463,7 @@ def run(self, initial_corpus_path, minimized_corpus_path, bad_units_path):
     except engine.TimeoutError as e:
       raise CorpusPruningException(
           'Corpus pruning timed out while merging corpus: ' + e.message)
-    except engine.MergeError as e:
+    except engine.Error as e:
       raise CorpusPruningException('Corpus pruning failed to merge corpus: ' +
```
inferno-chromium: I think this message is incorrect then s/merge corpus/minimize corpus. Also, same fix for line logs.log('Running merge...')

PR #1260, opened 2019-12-12
Start using disable_ubsan in fuzz task.

inferno-chromium:
This is follow up to #1257


src/python/bot/tasks/fuzz_task.py
```diff
@@ -665,6 +666,20 @@ def pick_redzone():
...
+  # This is only applicable in an ASan build.
+  memory_tool_name = environment.get_memory_tool_name(job_type)
+  if memory_tool_name not in ['ASAN', 'HWASAN']:
+    return False
+
+  # Check if UBSan is enabled in this ASan build. If not, can't disable it.
+  if not environment.get_value('UBSAN'):
+    return False
+
+  return not utils.random_number(0, DEFAULT_CHOOSE_PROBABILITY)
```
Dor1s: why only 10%? maybe 50?

src/python/tests/core/bot/tasks/fuzz_task_test.py
```diff
@@ -730,6 +730,7 @@ def test_existing_unreproducible_testcase(self):
             job_type='job',
             fuzz_target=data_types.FuzzTarget(engine='engine', binary='binary'),
             redzone=111,
+            disable_ubsan=False,
```
Dor1s: would it make sense to have a test to verify `get_ubsan_disabled_options` is working? All tests here and in #1257 are using `disable_ubsan=False`

src/python/bot/tasks/fuzz_task.py
inferno-chromium: Starting slow. Also, i think for major targets we hurt fuzzing efficiency and lose ubsan bugs if we disable too many times.

src/python/tests/core/bot/tasks/fuzz_task_test.py
inferno-chromium: added.

PR #1264, opened 2019-12-12
Added log output when minimizer is finished running

mpherman2:


src/python/bot/minimizer/minimizer.py
```diff
@@ -208,18 +208,21 @@ def _delete_file_if_needed(self, input_file):
     except OSError:
       pass
 
-  def _report_progress(self):
+  def _report_progress(self, is_final=False):
```
mbarbella-chromium: Nit: is_final isn't too clear a name for what this actually does. Maybe something like force_progress_report?

src/python/bot/minimizer/minimizer.py
```diff
@@ -208,18 +208,21 @@ def _delete_file_if_needed(self, input_file):
...
     """Call a function to report progress if the minimizer uses one."""
-    if not self.minimizer.progress_report_function:
+    if not self.minimizer.progress_report_function and not is_final:
```
mbarbella-chromium: In this case, we still want to bail out. I believe this would lead to an exception below since we wouldn't have a valid progress report function to call.

src/python/bot/minimizer/minimizer.py
```diff
@@ -208,18 +208,21 @@ def _delete_file_if_needed(self, input_file):
...
       return
 
-    if time.time() - self.last_progress_report_time < PROGRESS_REPORT_INTERVAL:
+    if time.time() - self.last_progress_report_time < PROGRESS_REPORT_INTERVAL \
```
mbarbella-chromium: Nit: parens around this entire condition instead of a backslash continuation.

src/python/bot/minimizer/minimizer.py
```diff
@@ -466,6 +469,8 @@ def _has_tested(self, hypothesis):
   # Result checking functions.
   def get_result(self):
     """Get the result of minimization."""
+    # Done with minimization, output log one more time
+    self._report_progress(True)
```
mbarbella-chromium: Nit: explicit parameter_name=True here.

PR #1269, opened 2019-12-12
Add newlines to corpus pruning exception messages.

oliverchang:


inferno-chromium:
@Dor1s fyi, you should stop seeing those emails with long output after this.

Dor1s:
> @Dor1s fyi, you should stop seeing those emails with long output after this.

ack, will keep an eye

Dor1s:
Looks like a line break doesn't stop all the message from going into the error: https://pantheon.corp.google.com/errors/CIn9yNWGqJaRNg?time=P1D&project=google.com:clusterfuzz

Dor1s:
should we do something like this:

```
logs.log_error('Merging new testcases timed out', output=result.output)
raise engine.TimeoutError('Merging new testcases timed out.')
```

would StackDriver associate the log message and the exception?

PR #1270, opened 2019-12-12
Fix regression in engine pipeline with setting redzone,disable-ubsan flag

inferno-chromium:
This matches what we do in blackbox fuzzing pipeline.


inferno-chromium:
Basically this will make https://github.com/google/clusterfuzz/pull/1260 work. it didn't work because engine fuzzing branch didn't set it.

PR #1272, opened 2019-12-13
Support setting testcase metadata when a testcase archive is provided.

inferno-chromium:
This makes it possible to set team labels when an archive with multiple testcases is uploaded.

### DIFF
--- a/src/python/datastore/data_types.py
+++ b/src/python/datastore/data_types.py
@@ -849,6 +849,9 @@ class TestcaseUploadMetadata(Model):
   # Flag to indicate if we are running in quiet mode (e.g. bug updates).
   quiet_flag = ndb.BooleanProperty()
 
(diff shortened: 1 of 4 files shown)


src/python/datastore/data_types.py
```diff
@@ -849,6 +849,9 @@ class TestcaseUploadMetadata(Model):
   # Flag to indicate if we are running in quiet mode (e.g. bug updates).
   quiet_flag = ndb.BooleanProperty()
 
+  # Additional testcase metadata dict stored as a string.
+  additional_metadata_string = TextProperty(indexed=False)
```
inferno-chromium: ndb.JsonProperty is not supported with our ndb patcher, that is why need to do this storing as a string and json.loads.

inferno-chromium:
This is for fixing Ned's uploads.

PR #1273, opened 2019-12-13
Corpus pruning task: fix an exception in the result logging.

Dor1s:
https://pantheon.corp.google.com/errors/CN3g59XWt7TpcQ?time=P30D&project=clusterfuzz-external

//cc @oliverchang

### DIFF
--- a/src/python/bot/tasks/corpus_pruning_task.py
+++ b/src/python/bot/tasks/corpus_pruning_task.py
@@ -497,21 +497,21 @@ class CrossPollinator(object):
 
     additional_args = self.runner.get_libfuzzer_flags()
 
+    symbolized_output = None
     try:
       result = self.runner.minimize_corpus(additional_args,
                                            [self.context.shared_corpus_path],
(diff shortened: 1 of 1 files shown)


src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -502,17 +502,16 @@ def run(self, timeout):
                                            [self.context.shared_corpus_path],
                                            self.context.minimized_corpus_path,
                                            self.context.bad_units_path, timeout)
+      symbolized_output = stack_symbolizer.symbolize_stacktrace(result.logs)
```
inferno-chromium: This code is not supposed to exception out, so can keep it at original place, but add return to the logs.log_error branch below (which causes the continue).

Dor1s: Isn't it more readable in this form? IMO it's more straightforward to follow right now, but can change if you insist

PR #1274, opened 2019-12-13
Remove buggy code that tries to get deps in bundled archives.

inferno-chromium:
This code is not used since unpack task only write one file
in every testcase from archive and also does not work with
new engine pipeline code.


inferno-chromium:
See exceptions in https://oss-fuzz.com/testcase-detail/5764706216443904

Issue #1279, opened 2019-12-16
Corpus sync file is either corrupted or empty


Dor1s:
https://pantheon.corp.google.com/errors/CIGn7fHSr4HOeg?time=P7D&project=clusterfuzz-external

```
TypeError: a float is required
at sync_from_gcs (/mnt/scratch0/bots/oss-fuzz-linux-zone3-host-qn7d-6/clusterfuzz/src/python/bot/tasks/fuzz_task.py:466)
at sync_corpus (/mnt/scratch0/bots/oss-fuzz-linux-zone3-host-qn7d-6/clusterfuzz/src/python/bot/tasks/fuzz_task.py:1305)
at do_engine_fuzzing (/mnt/scratch0/bots/oss-fuzz-linux-zone3-host-qn7d-6/clusterfuzz/src/python/bot/tasks/fuzz_task.py:1503)
at run (/mnt/scratch0/bots/oss-fuzz-linux-zone3-host-qn7d-6/clusterfuzz/src/python/bot/tasks/fuzz_task.py:1801)
at execute_task (/mnt/scratch0/bots/oss-fuzz-linux-zone3-host-qn7d-6/clusterfuzz/src/python/bot/tasks/fuzz_task.py:1881)
at run_command (/mnt/scratch0/bots/oss-fuzz-linux-zone3-host-qn7d-6/clusterfuzz/src/python/bot/tasks/commands.py:204)
at process_command (/mnt/scratch0/bots/oss-fuzz-linux-zone3-host-qn7d-6/clusterfuzz/src/python/bot/tasks/commands.py:378)
at wrapper (/mnt/scratch0/bots/oss-fuzz-linux-zone3-host-qn7d-6/clusterfuzz/src/python/bot/tasks/commands.py:150)
at task_loop (/mnt/scratch0/bots/oss-fuzz-linux-zone3-host-qn7d-6/clusterfuzz/src/python/bot/startup/run_bot.py:97)
```

The `sync_file_path` exists, but `utils.read_data_from_file` returns `None`

Trying to find any other clues why

Dor1s:
Don't see any relevant code changes or log entries. Let me upload a fix to unblock the fuzz task.

PR #1280, opened 2019-12-16
Fix an exception with an empty or malformed corpus last sync file (#1279).

Dor1s:


Dor1s:
> You can also try catch the thing, if lets say the file was corrupted, that might be better than just "if last_sync_time"

`utils.read_data_from_file` seems to be doing that and returns `None` in such case. I can do `eval_data=False`, in that case it should be fair to say that the file is empty

inferno-chromium:
> > You can also try catch the thing, if lets say the file was corrupted, that might be better than just "if last_sync_time"
> 
> `utils.read_data_from_file` seems to be doing that and returns `None` in such case. I can do `eval_data=False`, in that case it should be fair to say that the file is empty

I mean datetime.datetime.utcfromtimestamp can exception if the content is corrupt ?

Dor1s:
Yes, this is what happens right now. Let me upload another patch to look at, since the tests seem to be broken anyway:

```
selenium.common.exceptions.SessionNotCreatedException: Message: session not created: This version of ChromeDriver only supports Chrome version 77
```

src/python/bot/tasks/fuzz_task.py
```diff
@@ -419,6 +419,28 @@ def _track_testcase_run_result(fuzzer, job_type, new_crash_count,
   })
 
 
+def _last_sync_time(sync_file_path):
+  """Read and parse the last sync file for the GCS corpus."""
+  if not os.path.exists(sync_file_path):
+    return None
+
+  file_contents = utils.read_data_from_file(sync_file_path, eval_data=False)
```
oliverchang: nit: why not just let lines 432 onwards handle this case also?

Dor1s: My understanding is that empty file is ok (therefore just a warning), while corrupted means we have a problem and an error should be reported. My initial patch was to log a message like "either malformed or empty" and Abhishek suggested to catch the failure and separate these

PR #1282, opened 2019-12-16
Make butler bootstrap work on Windows

inferno-chromium:


inferno-chromium:
There is another issue with tensorflow==1.8.0 but it should go away with Python 3.

src/local/butler/common.py
```diff
@@ -322,10 +331,7 @@ def symlink(src, target):
 
 def copy_dir(src, target):
   """Copy directory."""
-  if os.path.islink(target):
-    os.remove(target)
-
-  if os.path.isdir(target):
+  if os.path.exists(target):
```
inferno-chromium: This is never a symlink, only one call site.

PR #1287, opened 2019-12-17
[Python 2->3] Fix remaining futurize issues.

mbarbella-chromium:
PTAL when you have time. Tests are passing but I still plan to do some additional local testing before considering deploying this.

### DIFF
--- a/src/local/butler/lint.py
+++ b/src/local/butler/lint.py
@@ -196,6 +196,13 @@ def execute(_):
     _execute_command_and_track_error('pylint ' + file_path)
     _execute_command_and_track_error('yapf -d ' + file_path)
 
+    py_import_order(file_path)
+    py_test_init_check(file_path)
+
+    # Skip files that cause futurize to report false positive issues.
+    if file_path.endswith('ndb_patcher.py'):
+      continue
+
     futurize_excludes = ' '.join(
         ['-x ' + exception for exception in _FUTURIZE_EXCEPTIONS])
     futurize_command = 'futurize -0 {excludes} {file_path}'.format(
@@ -205,9 +212,6 @@ def execute(_):
         'No files need to be modified' not in futurize_output):
       # Futurize doesn't modify its return code depending on the result.
       _error('Python 3 compatibility error introduced.')
-
-    py_import_order(file_path)
-    py_test_init_check(file_path)
 
   golint_path = os.path.join('local', 'bin', 'golint')
   for file_path in go_changed_file_paths:
--- a/src/local/butler/lint.py
+++ b/src/local/butler/lint.py
@@ -23,10 +23,6 @@ from local.butler import appengine
 from local.butler import common
 
 _FUTURIZE_EXCEPTIONS = [
-    # Not yet handled across the codebase.
-    'libfuturize.fixes.fix_unicode_keep_u',
-    'libfuturize.fixes.fix_future_builtins',
-
     # Causes issues if applied to the same code multiple times.
     'lib2to3.fixes.fix_dict',
 ]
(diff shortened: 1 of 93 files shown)


src/local/butler/lint.py
```diff
@@ -200,6 +196,13 @@ def execute(_):
     _execute_command_and_track_error('pylint ' + file_path)
     _execute_command_and_track_error('yapf -d ' + file_path)
 
+    py_import_order(file_path)
+    py_test_init_check(file_path)
+
+    # Skip files that cause futurize to report false positive issues.
+    if file_path.endswith('ndb_patcher.py'):
```
mbarbella-chromium: This feels pretty hacky but we won't need to keep it around too long. After the migration we'll remove the futurize checks.

src/python/base/utils.py
```diff
@@ -13,10 +13,14 @@
 # limitations under the License.
```
mbarbella-chromium: This is the only file with significant non-mechanical changes.

src/python/build_management/revisions.py
```diff
@@ -694,7 +697,7 @@ def write_revision_to_revision_file(revision_file, revision):
   """Writes a revision to the revision file."""
   try:
     with open(revision_file, 'wb') as file_handle:
-      file_handle.write(bytes(revision))
+      file_handle.write(str(revision))
```
mbarbella-chromium: This is potentially a slight behavior change, but I don't expect it to cause any issues.

src/local/butler/lint.py
oliverchang: SGTM. let's add a TODO here so we don't forget?

src/python/base/utils.py
```diff
@@ -782,16 +787,20 @@ def wait_until_timeout(threads, thread_timeout):
...
+  # TODO(mbarbella): Require callers to pass strings or bytes then reduce this
+  # back to the original retry limit.
+  for _ in range(retry_limit + 1):
     try:
       with open(file_path, file_mode) as file_handle:
-        file_handle.write(content_string)
-    except:
+        file_handle.write(content)
+    except TypeError:
+      content = str(content)
```
oliverchang: can we just always try convert it to the right type?

also, file_mode is 'ab' or 'wb'. Wouldn't writing a str type here fail on python 3 ?

```python
>>> with open('test', 'wb') as f:
...   f.write('123')
...
  File "<stdin>", line 2, in <module>
TypeError: a bytes-like object is required, not 'str'
```

src/local/butler/lint.py
inferno-chromium: +1 TODO with "python 3" word in it.

src/python/base/utils.py
mbarbella-chromium: Good catch on the type issue, I'll fix that.

As for always converting, we have a few cases where we depend on the Python 2 behavior for some of this. For example we might pass the integer 0 to this function and expect to write "0" to the file (converting to bytes in Python 3 will write an empty file).

I think the problem is really that we're just expecting this function to work a bit too magically. For now I think trying to catch the type error here is the right way to get this working with the behavior we expect, but later callers should just pass sane input to this function.

mbarbella-chromium:
> nice! LGTM with some comments.
> 
> this definitely needs to be staged on bots (including non-preemptible).
> 
> Also, do we need to fix up all our string literals to use the 'u' prefix as well to match Python 3 behaviour?

Good point, we'll still need to fix the literals. I'd rather do that in another change just to avoid making this one even more complicated, but I think that should be a bit lower risk. The future library provides a pretty good way to handle it, though it seems like it wasn't done by futurize automatically.

PR #1288, opened 2019-12-17
Catch merge failures due to timeouts.

inferno-chromium:
engine.TimeoutError is new and needs to be caught, otherwise
we lose crashes during fuzzing when merge fails due to timeouts.

### DIFF
--- a/src/python/bot/fuzzers/libFuzzer/engine.py
+++ b/src/python/bot/fuzzers/libFuzzer/engine.py
@@ -210,8 +210,8 @@ class LibFuzzerEngine(engine.Engine):
       new_units_added = new_corpus_len - old_corpus_len
 
       stat_overrides.update(result.stats)
-    except MergeError:
-      logs.log_warn('Merge failed', target=os.path.basename(target_path))
+    except (MergeError, engine.TimeoutError):
+      logs.log_warn('Merge failed.')
 
     stat_overrides['new_units_added'] = new_units_added
 


src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -210,8 +210,8 @@ def _merge_new_units(self, target_path, corpus_dir, new_corpus_dir,
       new_units_added = new_corpus_len - old_corpus_len
 
       stat_overrides.update(result.stats)
-    except MergeError:
-      logs.log_warn('Merge failed', target=os.path.basename(target_path))
+    except (MergeError, engine.TimeoutError):
+      logs.log_warn('Merge failed.')
```
inferno-chromium: This target path is unneeded since fuzz target is in log variable.

inferno-chromium:
Should fix https://github.com/google/oss-fuzz/issues/3139 fuzzing part. coverage stuff will still be broken due to merge timeout.

src/python/bot/fuzzers/libFuzzer/engine.py
oliverchang: Thanks!

Dor1s: Does this mean we lose the output from https://github.com/google/clusterfuzz/blob/7d5cec7eed73366bcbcf733df6af02f362545f8a/src/python/bot/fuzzers/libFuzzer/engine.py#L484, or is it logged anyways somehow?

inferno-chromium: Probably you can add it as part of https://github.com/google/clusterfuzz/pull/1289#pullrequestreview-333336737

Dor1s: Done!

PR #1297, opened 2019-12-19
Make honggfuzz binary executable.

oliverchang:
Some builds may not have an executable honggfuzz binary.

### DIFF
--- a/src/python/bot/fuzzers/honggfuzz/engine.py
+++ b/src/python/bot/fuzzers/honggfuzz/engine.py
@@ -118,6 +118,7 @@ class HonggfuzzEngine(engine.Engine):
     Returns:
       A FuzzOptions object.
     """
+    os.chmod(target_path, 0o775)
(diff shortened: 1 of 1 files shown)


oliverchang:
TBR. simple fix for internal builds.

src/python/bot/fuzzers/honggfuzz/engine.py
```diff
@@ -55,6 +55,7 @@ def _get_runner():
   if not os.path.exists(honggfuzz_path):
     raise HonggfuzzError('honggfuzz not found in build')
 
+  os.chmod(honggfuzz_path, 0o755)
```
inferno-chromium: If we are doing this here, why doing in prepare and reproduce ?

oliverchang: That's the target binary, while this is the honggfuzz driver binary.

PR #1298, opened 2019-12-19
Replace App Engine NDB with google cloud NDB.

oliverchang:
Part of migration to Python 3. This is our last major blocker and
dependency on the Python 2 App Engine SDK.

Also remove ndb_patcher, as that is no longer required.

Notable changes:
- indexed=False is no longer supported for StringProperty. Use TextProperty instead, which is
  equivalent.
- batch_size query keyword arg is no longer supported (Datastore backend change).
- fetch_page() is no longer supported for complex queries with !=, OR etc.
- Many import hacks/workarounds to make things work for App Engine.
  - Notable: add `grpcio` to app.yaml, cron-service.yaml
- Add workaround for googleapis/python-ndb#292 and googleapis/python-ndb#295
- Threads may be created by the new NDB implementation, which are only guaranteed to be 
  stopped after a GC. GAE doesn't like it when there are stray threads at the end of a request.
- Stub out `ndb_utils.get_all_from_query`. This is no longer required, and callers will be cleaned up in a future PR.
- All datastore operations now require:
```python
from datastore import ndb_init
with ndb_init.context():
  ...
```
This is automatically set up in tests, as well as in run_bot, run, heartbeat, run_heartbeat, and butler run, as well as for each individual request.


oliverchang:
One more blocker: https://github.com/googleapis/python-ndb/issues/292

oliverchang:
Update: I've now gotten this to run on real GAE, but now the testcases list page take literally minutes to load.

oliverchang:
Went down a pretty deep rabbit hole to figure out why this testcases list (and only this page) was hanging with this new NDB.

In case anyone is interested, the root cause is that App Engine doesn't like it when a request creates threads which are still running at the end -- this will appear as if the request never completes, without any obvious indication why.

The new NDB uses gRPC, which in turn uses google.auth.transports.grpc, which creates a [ThreadPoolExecutor with 1 thread](https://github.com/googleapis/google-auth-library-python/blob/0ca0ee55dc7f8e24728c2773f57c31aa7939207a/google/auth/transport/grpc.py#L56).

Somehow, when a [MultiQuery](https://github.com/googleapis/python-ndb/blob/master/google/cloud/ndb/_datastore_query.py#L486) happens (in the case of testcases list), things are set up such that there are references cycles which prevent the threads from getting stopped normally [here](https://github.com/googleapis/google-auth-library-python/blob/0ca0ee55dc7f8e24728c2773f57c31aa7939207a/google/auth/transport/grpc.py#L91).

So, at the end of every request, we explicitly call the GC, which fixes this.

There is also one more bug in NDB that I filed: https://github.com/googleapis/python-ndb/issues/295

oliverchang:
Don't see any exceptions on the bots I staged over the weekend. This is ready for review while I do some more testing.

oliverchang:
(CI will fail right now due to a conflicting dependency in the image, which is removed by this PR).

src/python/datastore/data_types.py
```diff
@@ -396,9 +396,9 @@ class Testcase(Model):
 
   # Blobstore keys for various things like original testcase, minimized
   # testcase, etc.
-  fuzzed_keys = StringProperty(indexed=False)
-  minimized_keys = StringProperty(indexed=False)
-  minidump_keys = StringProperty(indexed=False)
+  fuzzed_keys = TextProperty()
```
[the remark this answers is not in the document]
oliverchang: These are always stored as strings (assuming utf-8 encoding if byte are passed), per https://github.com/googleapis/python-ndb/blob/662a88a82691e69d282565b9d11706237da6f654/google/cloud/ndb/model.py#L2698

Also validated this by running the code.

src/appengine/libs/query/datastore_query.py
```diff
@@ -53,12 +55,41 @@ def _combine(q1, q2):
...
+      try:
+        self._last_cursor = self._it.cursor_after()
+      except exceptions.BadArgumentError:
+        pass
+
+    return next(self._it)
+
+  next = __next__
+
+  def last_cursor(self):
```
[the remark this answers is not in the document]
oliverchang: Done.

src/appengine/handlers/crash_stats.py
```diff
@@ -131,11 +131,7 @@ def query_testcase(project_name, crash_type, crash_state, security_flag,
       data_types.Testcase.open == is_open,
       ndb_utils.is_false(data_types.Testcase.is_a_duplicate_flag)).order(
           -data_types.Testcase.timestamp).iter(
-              limit=1,
-              projection=[
-                  data_types.Testcase.bug_information,
```
inferno-chromium: Is this really needed ? i prefer this style reference via doing typos in strings.

src/appengine/libs/query/datastore_query.py
```diff
@@ -53,12 +55,42 @@ def _combine(q1, q2):
   return result
 
 
+# TODO(ochang): Remove once bug is fixed.
```
inferno-chromium: link to bug

src/appengine/libs/query/datastore_query.py
```diff
@@ -53,12 +55,42 @@ def _combine(q1, q2):
...
+  def __next__(self):
+    if not self._it.has_next():
+      try:
+        self._last_cursor = self._it.cursor_after()
+      except exceptions.BadArgumentError:
+        pass
+
+    return next(self._it)
+
+  next = __next__
```
inferno-chromium: nite: move to start of class (not in middle of functions).

src/python/datastore/data_types.py
```diff
@@ -385,7 +385,7 @@ class Testcase(Model):
   crash_type = StringProperty()
 
   # Crashing address.
-  crash_address = StringProperty(indexed=False)
+  crash_address = TextProperty()
```
inferno-chromium: Will we lose values in old Testcase entities. Many times i have seen this if we change data type, it resets values in old entities ?

src/python/datastore/data_types.py
```diff
@@ -631,6 +631,10 @@ def _pre_put_hook(self):
     self.populate_indices()
 
   def _post_put_hook(self, _):
+    if not self.key:
+      # Failed put.
+      return
```
inferno-chromium: Do we need to log ? How is this expected, maybe expand comment.

src/appengine/handlers/crash_stats.py
oliverchang: This is https://github.com/googleapis/python-ndb/issues/292.

src/python/datastore/data_types.py
oliverchang: Nope. TextProperty is equivalent to StringProperty(indexed=False) (but the new NDB requires you to just use TextProperty instead).

src/python/datastore/data_types.py
oliverchang: This was always wrong -- _post_put_hook can be called even if there was an exception in put(), and the exception will be rethrown automatically afterwards. Expanded the comment a bit more.

src/appengine/libs/query/datastore_query.py
oliverchang: Done!

src/appengine/libs/query/datastore_query.py
oliverchang: This is a common pattern (see e.g. https://github.com/googleapis/python-ndb/blob/4ffedc7b5a2366be15dcd299052d8a46a748addd/google/cloud/ndb/_datastore_query.py#L468). Putting it before won't work either becausess __next__ won't be defined at that point. This is just to create an alias for python2/3 compatibility.

src/appengine/handlers/crash_stats.py
oliverchang: Sorry meant https://github.com/googleapis/python-ndb/issues/295

oliverchang:
phew, CI is finally passing. Merging!

PR #1299, opened 2019-12-19
Fix an exception in test case upload.

mbarbella-chromium:
PTAL

### DIFF
--- a/src/appengine/handlers/upload_testcase.py
+++ b/src/appengine/handlers/upload_testcase.py
@@ -364,7 +364,7 @@ class UploadHandlerCommon(object):
     app_launch_command = self.request.get('cmd')
     platform_id = self.request.get('platform')
     issue_labels = self.request.get('issue_labels')
-    gestures = self.request.get('gestures', '[]')
+    gestures = self.request.get('gestures') or '[]'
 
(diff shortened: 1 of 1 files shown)


src/appengine/handlers/upload_testcase.py
```diff
@@ -364,7 +364,7 @@ def do_post(self):
     app_launch_command = self.request.get('cmd')
     platform_id = self.request.get('platform')
     issue_labels = self.request.get('issue_labels')
-    gestures = self.request.get('gestures', [])
+    gestures = self.request.get('gestures') or '[]'
```
mbarbella-chromium: A different exception occurs if we pass an empty string, and self.request.get will use that rather than None since the gestures field is present when we submit the form even if it's empty. This seems like the safest way to handle the fix to me.

inferno-chromium: Why [] quoted ?

mbarbella-chromium:
literal_eval expects a string.

src/appengine/handlers/upload_testcase.py
mbarbella-chromium: Left a comment off this thread but just to follow up here it's because literal_eval expects a string as input.

PR #1307, opened 2020-01-02
Allow fuzzer weights to recover from 0.

oliverchang:
If the target count drops to 0, we may never recover otherwise.


oliverchang:
TBR for a trivial fix.

PR #1308, opened 2020-01-02
Fix grpc exception from huge libFuzzer cleanse output.

inferno-chromium:
Remove the output altogether from exception message.


inferno-chromium:
@oliverchang - i cannot think of a better stopgap fix for minimization failures. Can you think of some places where we can cap the output to grpc allowed message size.

oliverchang:
Just uploaded https://github.com/google/clusterfuzz/pull/1311 to hopefully fix this more generally. For libFuzzer, all output is already truncated to 1MB, but this was higher than the previously allowed gRPC limit.

PR #1312, opened 2020-01-03
Resurface TimeoutError more generally.

oliverchang:
Check for the error in the host's RPC wrapper, rather than in each
implementation of the RPC call.

### DIFF
--- a/src/python/bot/untrusted_runner/host.py
+++ b/src/python/bot/untrusted_runner/host.py
@@ -24,6 +24,7 @@ import grpc
 
 from base import untrusted
 from base import utils
+from bot.fuzzers import engine
 from datastore import data_types
 from datastore import ndb
 from metrics import logs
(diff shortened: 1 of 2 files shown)


src/python/bot/untrusted_runner/host.py
```diff
@@ -150,6 +153,12 @@ def wrapped(*args, **kwargs):
       try:
         return func(*args, **kwargs)
       except grpc.RpcError as e:
+        # For timeouts, which aren't fatal errors, resurface the right
```
inferno-chromium: Should add testcase_manager.TargetNotFoundError as well i think ?

oliverchang: Want to keep this more generic as with `TimeoutError`. `testcase_manager.TargetNotFoundError` is pretty specific to a single RPC, so keeping it there.

inferno-chromium: SGTM.

PR #1313, opened 2020-01-06
Add support for additional automatic components/labels/ccs.

oliverchang:
Through AUTOMATIC_CCS_1, AUTOMATIC_LABELS_1, AUTOMATIC_COMPONENTS_1 etc.
This is needed for e.g. environments where we always want to attach a
label, regardless of the job.

### DIFF
--- a/src/appengine/libs/issue_management/issue_filer.py
+++ b/src/appengine/libs/issue_management/issue_filer.py
@@ -245,7 +245,6 @@ def _get_additional_values(testcase, name):
   values = data_handler.get_additional_values_for_variable(
       name, testcase.job_type, testcase.fuzzer_name)
 
-  # TODO(ochang): Find a less hacky way of concatenating multiple values.
   return values + data_handler.get_additional_values_for_variable(
       name + '_1', testcase.job_type, testcase.fuzzer_name)
 
--- a/src/appengine/libs/issue_management/issue_filer.py
+++ b/src/appengine/libs/issue_management/issue_filer.py
@@ -240,6 +240,16 @@ def _get_from_metadata(testcase, name):
       remove_empty=True)
 
 
+def _get_additional_values(testcase, name):
+  """Get additional values from the job or fuzzer environment."""
+  values = data_handler.get_additional_values_for_variable(
+      name, testcase.job_type, testcase.fuzzer_name)
+
+  # TODO(ochang): Find a less hacky way of concatenating multiple values.
+  return values + data_handler.get_additional_values_for_variable(
+      name + '_1', testcase.job_type, testcase.fuzzer_name)
+
+
 def file_issue(testcase,
                issue_tracker,
                security_severity=None,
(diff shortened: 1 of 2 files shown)


oliverchang:
PTAL. this is needed for our internal ephemeral bots. Couldn't think of a cleaner solution, apart from rewriting our job system to support concatenating certain variables etc.

src/appengine/libs/issue_management/issue_filer.py
```diff
@@ -277,29 +287,26 @@ def file_issue(testcase,
...
-      'AUTOMATIC_COMPONENTS', testcase.job_type, testcase.fuzzer_name)
+  automatic_components = _get_additional_values(testcase,
+                                                'AUTOMATIC_COMPONENTS')
   for component in automatic_components:
     issue.components.add(component)
 
   # Add issue assignee from the job definition and fuzzer.
-  automatic_assignee = data_handler.get_additional_values_for_variable(
-      'AUTOMATIC_ASSIGNEE', testcase.job_type, testcase.fuzzer_name)
+  automatic_assignee = _get_additional_values(testcase, 'AUTOMATIC_ASSIGNEE')
```
inferno-chromium: We select automatic_assignee[0] below, so this change looks unnecessary ?

src/appengine/libs/issue_management/issue_filer.py
```diff
@@ -240,6 +240,16 @@ def _get_from_metadata(testcase, name):
...
 
 
+def _get_additional_values(testcase, name):
+  """Get additional values from the job or fuzzer environment."""
+  values = data_handler.get_additional_values_for_variable(
+      name, testcase.job_type, testcase.fuzzer_name)
+
+  # TODO(ochang): Find a less hacky way of concatenating multiple values.
+  return values + data_handler.get_additional_values_for_variable(
+      name + '_1', testcase.job_type, testcase.fuzzer_name)
```
inferno-chromium: Should we use ADDITIONAL_* , atleast that gets rid of this TODO, and matches slightly the style of like ADDITIONAL_ASAN_OPTIONS.

src/appengine/libs/issue_management/issue_filer.py
oliverchang: Just for consistency -- the code looks shorter that way anyway :)

src/appengine/libs/issue_management/issue_filer.py
oliverchang: That would be a more messy migration, since we need to update our jobs to use those vars (and handle any race conditions). 

Or are you saying to have both `AUTOMATIC_*` and `ADDITIONAL_*` ? That seems like it can be confusing too :/

inferno-chromium: yes both AUTOMATIC_* and ADDITIONAL_*. Current version is fine too. it will always remain as a hack specific to this usecase, so unsure if todo is needed ? Can be rephrased as an global override use for a specific part of lab/platform/etc

PR #1316, opened 2020-01-07
Disable fork mode completely on ephemeral bots.

oliverchang:


oliverchang:
TBR for trivial change.

PR #1317, opened 2020-01-08
Bump up libFuzzer/honggfuzz rss limit to 2.5GB.

oliverchang:


oliverchang:
@inferno-chromium do we also need to change this part? https://github.com/google/clusterfuzz/blob/master/src/python/platforms/android/adb.py#L37

inferno-chromium:
> @inferno-chromium do we also need to change this part? https://github.com/google/clusterfuzz/blob/master/src/python/platforms/android/adb.py#L37

That is only for cuttlefish bots where we are not running libFuzzing fuzzing, so not impacted. Someday will need to fix that, but unsure if those bots work with higher limit.

oliverchang:
> Will this affect OOMs that are currently open?

Nope, existing testcases have the oom limit as part of the minimized args.

Issue #389, opened 2019-04-17
Add new issue trackers based on issue tracker specification


dukebarman:
What about another issue trackers? (monorail is only for chromium project) We would like to add support gitlab issues (or trac/Bugzilla), but it's not clear what is the best way to do it. Do you have any plans to add more bug trackers or at least making that part of the code easily extendable?

inferno-chromium:
We currently don't have cycles to add support for other issue trackers. We encourage community contributions here. the package issue_management (https://github.com/google/clusterfuzz/tree/master/src/python/issue_management) should be self-contained and extendable.

jonathanmetzman:
Btw, we would love to support other trackers, so if you'd like to add support we'd be happy to work with you to do so.

inferno-chromium:
The code is totally extendable now - see e.g.monorail dir in https://github.com/google/clusterfuzz/tree/master/src/appengine/libs/issue_management/monorail. this shows how a new tracker can be added based on a generic specification by @oliverchang

dukebarman:
It's nice to hear. Thanks!

inferno-chromium:
@dukebarman - do you want to take a shot at adding gitlab or any other popular issue tracker (jira). let us know if you have any questions.

prestonvanloon:
Any chance for adding github as an issue tracker?

oliverchang:
Unfortunately we're short on cycles for adding GitHub support, but we always welcome contributions :) See https://github.com/google/clusterfuzz/issues/389#issuecomment-659745271 as an example where Jira was added.

sfiecko:
Sorry for returning to this topic after nearly 2 years - Is this PR #1321 a kind of Jira enablement? Can anyone confirm that Jira functionality is working?

PR #1321, opened 2020-01-09
Adds support for the Jira issue tracker

bhatkishor:
Partially addresses #389 

Hi! Here's the Jira integration that we discussed back in Sep/Oct. Some things came up that caused this to get put on the back-burner for a bit.

At a high level, this PR introduces the following:
- support for intake of the Jira server URL and auth credentials through /configuration
- specification of the Jira project for a job through its job definition
- client logic for interfacing with Jira

This was tested with an on-prem instance of Jira; Atlassian's docs seem to indicate that there's no difference in APIs with the cloud version.

Some caveats/requests for feedback with this:
- I've (manually) done an end-to-end issue filing test (upload job, find crash, file issue for the corresponding testcase) but it still needs unit tests; I was running into a `No api proxy found for service "datastore_v3"` assertion error when trying to fetch a value from `db_config` in a test I was writing. Do I need to mock/spoof that in my tests?
- I'd appreciate help with lines 83 and 97 in `src/appengine/libs/issue_management/jira/issue_tracker_manager.py`. With the former, I'm trying to prevent updating status if it hasn't moved from the `new` state, because Jira complains if you try a status update that doesn't actually cause a transition that's defined in the project's workflow. For the latter, I'd like to be able to grab the testcase's `is_crash()` value, but I figured I'd run it by you all before I make significant changes to the `issue.save()` invocation .
- I did change the Monorail `new_issue()` signature and invocations in order to generalize and allow the Jira tracker to read the Jira project value from the job definition. Feels like the wrong way to address this, let me know what I should do.
- Finally, I didn't want to add documentation around the issue trackers before I found out whether that was in the works for Monorail already.

Thanks for your help!

### DIFF
--- a/src/appengine/handlers/cron/oss_fuzz_build_status.py
+++ b/src/appengine/handlers/cron/oss_fuzz_build_status.py
@@ -57,223 +57,224 @@ REMINDER_INTERVAL = 3
 
 
 class OssFuzzBuildStatusException(Exception):
-  """Exceptions for the build status cron."""
+    """Exceptions for the build status cron."""
 
 
 def _get_issue_body(project_name, build_id, build_type):
-  """Return the issue body for filing new bugs."""
-  template = ('The last {num_builds} builds for {project} have been failing.\n'
-              '<b>Build log:</b> {log_link}\n'
-              'Build type: {build_type}\n\n'
-              'To reproduce locally, please see: '
-              'https://google.github.io/oss-fuzz/advanced-topics/reproducing'
-              '#reproducing-build-failures\n\n'
-              'If you have any questions, please file a bug on '
-              'https://github.com/google/oss-fuzz/issues/new.\n\n'
-              '**This bug will be automatically closed within a '
-              'day once it is fixed.**')
+    """Return the issue body for filing new bugs."""
+    template = ('The last {num_builds} builds for {project} have been failing.\n'
+                '<b>Build log:</b> {log_link}\n'
+                'Build type: {build_type}\n\n'
+                'To reproduce locally, please see: '
+                'https://google.github.io/oss-fuzz/advanced-topics/reproducing'
+                '#reproducing-build-failures\n\n'
+                'If you have any questions, please file a bug on '
+                'https://github.com/google/oss-fuzz/issues/new.\n\n'
+                '**This bug will be automatically closed within a '
+                'day once it is fixed.**')
 
-  return template.format(
-      num_builds=MIN_CONSECUTIVE_BUILD_FAILURES,
-      project=project_name,
-      log_link=_get_build_link(build_id),
-      build_type=build_type)
+    return template.format(
+        num_builds=MIN_CONSECUTIVE_BUILD_FAILURES,
+        project=project_name,
+        log_link=_get_build_link(build_id),
+        build_type=build_type)
 
 
 def _get_oss_fuzz_project(project_name):
-  """Return the OssFuzzProject entity for the given project."""
-  return ndb.Key(data_types.OssFuzzProject, project_name).get()
+    """Return the OssFuzzProject entity for the given project."""
+    return ndb.Key(data_types.OssFuzzProject, project_name).get()
 
 
 def _get_build_link(build_id):
-  """Return a link to the build log."""
-  return BUCKET_URL + '/log-' + build_id + '.txt'
+    """Return a link to the build log."""
+    return BUCKET_URL + '/log-' + build_id + '.txt'
 
 
 def _get_ndb_key(project_name, build_type):
-  """Constructs a Key literal for build failure entities."""
-  if build_type == MAIN_BUILD_TYPE:
-    return project_name
+    """Constructs a Key literal for build failure entities."""
+    if build_type == MAIN_BUILD_TYPE:
+        return project_name
 
-  # Use build type suffix for the auxiliary build (e.g. coverage).
-  return '%s-%s' % (project_name, build_type)
+    # Use build type suffix for the auxiliary build (e.g. coverage).
+    return '%s-%s' % (project_name, build_type)
 
 
 def create_build_failure(failure, build_type):
-  """Create new build failure."""
-  return data_types.OssFuzzBuildFailure(
-      id=_get_ndb_key(failure['name'], build_type),
-      project_name=failure['name'],
-      last_checked_timestamp=get_build_time(failure),
-      build_type=build_type)
+    """Create new build failure."""
+    return data_types.OssFuzzBuildFailure(
+        id=_get_ndb_key(failure['name'], build_type),
+        project_name=failure['name'],
+        last_checked_timestamp=get_build_time(failure),
+        build_type=build_type)
 
 
 def get_build_failure(project_name, build_type):
-  """Return the last build failure for the project."""
-  key = ndb.Key(data_types.OssFuzzBuildFailure,
-                _get_ndb_key(project_name, build_type))
-  return key.get()
+    """Return the last build failure for the project."""
+    key = ndb.Key(data_types.OssFuzzBuildFailure,
+                  _get_ndb_key(project_name, build_type))
+    return key.get()
 
 
 def close_build_failure(build_failure):
-  """Delete the build failure."""
-  build_failure.key.delete()
+    """Delete the build failure."""
+    build_failure.key.delete()
 
 
 def get_build_time(build):
-  """Return a datetime for when the build was done."""
-  # Strip the nanosecond precision from the timestamp, since it's not
-  # supported by Python.
-  stripped_timestamp = TIMESTAMP_PATTERN.match(build['finish_time'])
-  if not stripped_timestamp:
-    logs.log_error(
-        'Invalid timestamp %s for %s.' % (build['finish_time'], build['name']))
-    return None
+    """Return a datetime for when the build was done."""
+    # Strip the nanosecond precision from the timestamp, since it's not
+    # supported by Python.
+    stripped_timestamp = TIMESTAMP_PATTERN.match(build['finish_time'])
+    if not stripped_timestamp:
+        logs.log_error(
+            'Invalid timestamp %s for %s.' % (build['finish_time'], build['name']))
+        return None
 
-  return datetime.datetime.strptime(
-      stripped_timestamp.group(0), TIMESTAMP_FORMAT)
+    return datetime.datetime.strptime(
+        stripped_timestamp.group(0), TIMESTAMP_FORMAT)
 
 
 def file_bug(issue_tracker, project_name, build_id, ccs, build_type):
-  """File a new bug for a build failure."""
-  logs.log('Filing bug for new build failure (project=%s, build_type=%s, '
-           'build_id=%s).' % (project_name, build_type, build_id))
+    """File a new bug for a build failure."""
+    logs.log('Filing bug for new build failure (project=%s, build_type=%s, '
+             'build_id=%s).' % (project_name, build_type, build_id))
 
-  issue = issue_tracker.new_issue('job_type')
-  issue.title = '{project_name}: {build_type} build failure'.format(
-      project_name=project_name, build_type=build_type.capitalize())
-  issue.body = _get_issue_body(project_name, build_id, build_type)
-  issue.status = 'New'
-  issue.labels.add('Type-Build-Failure')
-  issue.labels.add('Proj-' + project_name)
+    issue = issue_tracker.new_issue()
+    issue.title = '{project_name}: {build_type} build failure'.format(
+        project_name=project_name, build_type=build_type.capitalize())
+    issue.body = _get_issue_body(project_name, build_id, build_type)
+    issue.status = 'New'
+    issue.labels.add('Type-Build-Failure')
+    issue.labels.add('Proj-' + project_name)
 
-  for cc in ccs:
-    issue.ccs.add(cc)
+    for cc in ccs:
+        issue.ccs.add(cc)
 
-  issue.save()
-  return str(issue.id)
+    issue.save()
+    return str(issue.id)
 
 
 def close_bug(issue_tracker, issue_id, project_name):
-  """Close a build failure bug."""
-  logs.log('Closing build failure bug (project=%s, issue_id=%s).' %
-           (project_name, issue_id))
+    """Close a build failure bug."""
+    logs.log('Closing build failure bug (project=%s, issue_id=%s).' %
+             (project_name, issue_id))
 
-  issue = issue_tracker.get_original_issue(issue_id)
-  issue.status = 'Verified'
-  issue.save(
-      new_comment='The latest build has succeeded, closing this issue.',
-      notify=True)
+    issue = issue_tracker.get_original_issue(issue_id)
+    issue.status = 'Verified'
+    issue.save(
+        new_comment='The latest build has succeeded, closing this issue.',
+        notify=True)
 
 
 def send_reminder(issue_tracker, issue_id, build_id):
-  """Send a reminder about the build still failing."""
-  issue = issue_tracker.get_original_issue(issue_id)
+    """Send a reminder about the build still failing."""
+    issue = issue_tracker.get_original_issue(issue_id)
 
-  comment = ('Friendly reminder that the the build is still failing.\n'
-             'Please try to fix this failure to ensure that fuzzing '
-             'remains productive.\n'
-             'Latest build log: {log_link}\n')
-  comment = comment.format(log_link=_get_build_link(build_id))
-  issue.save(new_comment=comment, notify=True)
+    comment = ('Friendly reminder that the the build is still failing.\n'
+               'Please try to fix this failure to ensure that fuzzing '
+               'remains productive.\n'
+               'Latest build log: {log_link}\n')
+    comment = comment.format(log_link=_get_build_link(build_id))
+    issue.save(new_comment=comment, notify=True)
 
 
 class Handler(base_handler.Handler):
-  """Build status checker."""
+    """Build status checker."""
 
-  def _close_fixed_builds(self, build_status, build_type):
-    """Close bugs for fixed builds."""
-    issue_tracker = issue_tracker_utils.get_issue_tracker()
-    if not issue_tracker:
-      raise OssFuzzBuildStatusException('Failed to get issue tracker.')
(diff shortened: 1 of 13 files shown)


bhatkishor:
@googlebot I signed it!

bhatkishor:
@googlebot I signed it!

inferno-chromium:
Thank you Kishor, really appreciate this PR. Did you sign the CLA with the right email address (one used for this PR ?)

bhatkishor:
@inferno-chromium I did; I was talking to our OSPO contact for the corporate CLA and she said that sometimes it takes up to 72 hours for the fact that I joined the Google group to land in whatever backend the bot looks at. I'm guessing that's what the delay is...

src/appengine/libs/issue_management/jira/issue_tracker_manager.py
```diff
@@ -0,0 +1,148 @@
...
+            credentials['username'], credentials['password']))
+        return jira_client
+
+    def save(self, issue):
+        """Save an issue."""
+        self._update(issue)
+
+    def _create(self, job_type):
+        """Create an issue."""
+        jira_project = data_handler.get_value_from_job_definition(
```
mbarbella-chromium: Any reason that a separate environment variable is being used here for the jira project? We already define PROJECT_NAME in some job definitions for the monorail project name, and I believe it would be set in __init__ above.

src/appengine/handlers/cron/oss_fuzz_build_status.py
```diff
@@ -139,7 +139,7 @@ def file_bug(issue_tracker, project_name, build_id, ccs, build_type):
   logs.log('Filing bug for new build failure (project=%s, build_type=%s, '
            'build_id=%s).' % (project_name, build_type, build_id))
 
-  issue = issue_tracker.new_issue()
+  issue = issue_tracker.new_issue('job_type')
```
mbarbella-chromium: I left a comment where this function is defined, but this usage feels a bit hacky to me. I think it should be possible to remove the need for it.

src/appengine/libs/issue_management/jira/__init__.py
```diff
@@ -0,0 +1,196 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Jira issue tracker."""
+
+from libs.issue_management import issue_tracker
+from libs.issue_management.jira.issue_tracker_manager import (
+    IssueTrackerManager)
+
+
+class Issue(issue_tracker.Issue):
```
mbarbella-chromium: Nit: we use two spaces for indentation, and it looks like our lint checks found a few other issues. Could you look through the travis logs and fix the issues it reported? You can also run our lint (and a few other) checks locally with "python butler.py lint".

src/appengine/libs/issue_management/jira/issue_tracker_manager.py
```diff
@@ -0,0 +1,148 @@
...
+        if 'Critical - P1' in labels:
+            update_fields['priority'] = {
+                'name': 'Critical - P1'
+            }
+        elif 'Major - P2' in labels:
+            update_fields['priority'] = {
+                'name': 'Major - P2'
+            }
+
+        print(update_fields)
```
mbarbella-chromium: Nit: delete this.

src/appengine/libs/issue_management/jira/__init__.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def find_issues(self, keywords=None, only_open=False):
+        """Find issues."""
+        search_text = _get_search_text(keywords)
+        issues = self._itm.get_issues(search_text)
+        return [Issue(self._itm, issue) for issue in issues]
+
+    def issue_url(self, issue_id):
+        """Return the issue URL with the given ID."""
+        from config import db_config
```
mbarbella-chromium: Can this be moved to a top level import?

src/appengine/libs/issue_management/jira/issue_tracker_manager.py
```diff
@@ -0,0 +1,148 @@
...
+
+        update_fields = {
+            'summary': issue.title,
+            'description': issue.body,
+            'labels': labels,
+            'components': components,
+        }
+
+        # Only add status if it has changed
+        # This is brittle - we should be pulling the equivalent of 'new' from the policy
```
mbarbella-chromium: Nit: please end comments with punctuation here and in other parts of the file/files.

mbarbella-chromium:
> * I've (manually) done an end-to-end issue filing test (upload job, find crash, file issue for the corresponding testcase) but it still needs unit tests; I was running into a `No api proxy found for service "datastore_v3"` assertion error when trying to fetch a value from `db_config` in a test I was writing. Do I need to mock/spoof that in my tests?

You likely need to add the `@test_utils.with_cloud_emulators('datastore')` decorator to your test. Search for that string in the code for examples of how/when it's used.

> * I'd appreciate help with lines 83 and 97 in `src/appengine/libs/issue_management/jira/issue_tracker_manager.py`. With the former, I'm trying to prevent updating status if it hasn't moved from the `new` state, because Jira complains if you try a status update that doesn't actually cause a transition that's defined in the project's workflow. For the latter, I'd like to be able to grab the testcase's `is_crash()` value, but I figured I'd run it by you all before I make significant changes to the `issue.save()` invocation .

I don't fully understand how having access to is_crash helps here. Could you provide an example of any usage we have now that causes a problem and how the more invasive solution would solve it?

> * I did change the Monorail `new_issue()` signature and invocations in order to generalize and allow the Jira tracker to read the Jira project value from the job definition. Feels like the wrong way to address this, let me know what I should do.

I'm about to publish some initial review comments, and I mention this there. Agreed that it probably isn't the right way to handle it, but hopefully my comment clears it up.

> * Finally, I didn't want to add documentation around the issue trackers before I found out whether that was in the works for Monorail already.

I don't think we have anything planned in the short-term, but Oliver would know better.

> 
> Thanks for your help!

src/appengine/libs/issue_management/jira/issue_tracker_manager.py
bhatkishor: Good call; using PROJECT_NAME will also let me avoid having to call `data_handler.get_value_from_job_definition` and I can revert the `new_issue` call to its old signature.

inferno-chromium:
@oliverchang - can you do a review pass as well!

src/appengine/handlers/cron/oss_fuzz_build_status.py
bhatkishor: Agreed - this will get rolled back to the current mode of calling because the only reason I made the change was to be able to call `data_handler.get_value_from_job_definition` and that's resolved by using PROJECT_NAME instead.

src/appengine/libs/issue_management/jira/__init__.py
bhatkishor: Will do

src/appengine/libs/issue_management/jira/issue_tracker_manager.py
bhatkishor: Oops, will do.

src/appengine/libs/issue_management/jira/__init__.py
bhatkishor: Yup! On it.

src/appengine/libs/issue_management/jira/issue_tracker_manager.py
bhatkishor: Roger that.

bhatkishor:
@mbarbella-chromium 

> You likely need to add the @test_utils.with_cloud_emulators('datastore') decorator to your test. Search for that string in the code for examples of how/when it's used.

Thanks, I'll check it out.

> I don't fully understand how having access to is_crash helps here. Could you provide an example of any usage we have now that causes a problem and how the more invasive solution would solve it?

From what I could see, the `is_crash` value is being used to assign priority labels, no? Jira has another dedicated field for priority and I wanted to be able to assign the right value to that instead of having everything be whatever the default for the project is.

bhatkishor:
Sorry for the force-pushes; I kept noticing things that shouldn't have been changed in my lint fix effort and I kept rebasing to roll them into a single commit.

Looks like the open items right now are:
- conversation around how to set priority (use `is_crash`? better method?)
- docs for issue tracker usage - I think I'll open a separate PR for that
- unit tests for the Jira code - this is on me

src/appengine/libs/issue_management/jira/__init__.py
```diff
@@ -0,0 +1,200 @@
...
+    search_text = _get_search_text(keywords)
+    if only_open:
+      search_text += "AND resolution = Unresolved"
+    issues = self._itm.get_issues(search_text)
+    return [Issue(self._itm, issue) for issue in issues]
+
+  def issue_url(self, issue_id):
+    """Return the issue URL with the given ID."""
+    config = db_config.get()
+    url = config.jira_url + "/browse/" + str(issue_id)
```
oliverchang: Please use single quotes for all strings that aren't docstrings (Also on line 166 and probably others).

src/appengine/libs/issue_management/jira/issue_tracker_manager.py
```diff
@@ -0,0 +1,135 @@
...
+        jira_url, auth=(credentials['username'], credentials['password']))
+    return jira_client
+
+  def save(self, issue):
+    """Save an issue."""
+    self._update(issue)
+
+  def create(self):
+    """Create an issue."""
+    jira_project = self.project_name
```
oliverchang: nit: no need for the jira_project var. Just use self.project_name directly below.

src/appengine/libs/issue_management/jira/issue_tracker_manager.py
```diff
@@ -0,0 +1,135 @@
...
+    jira_issue = self.client.create_issue(fields=default_fields)
+    return jira_issue
+
+  def _update(self, issue):
+    """Update an issue."""
+
+    # Get labels from LabelStore.
+    labels = []
+    for label in issue.labels:
+      labels.append(label)
```
oliverchang: you should just be able to do `labels = list(issue.labels)`. Similarly for components

src/appengine/libs/issue_management/jira/issue_tracker_manager.py
```diff
@@ -0,0 +1,135 @@
...
+    elif 'Major - P2' in labels:
+      update_fields['priority'] = {'name': 'Major - P2'}
+
+    # Jira weirdness, update watchers this way.
+    for watcher in watchers:
+      self.client.add_watcher(issue.jira_issue, watcher)
+
+    issue.jira_issue.update(fields=update_fields)
+
+  def get_watchers(self, issue):
```
oliverchang: add a docstring here.

src/appengine/libs/issue_management/jira/issue_tracker_manager.py
```diff
@@ -0,0 +1,135 @@
...
+    update_fields = {
+        'summary': issue.title,
+        'description': issue.body,
+        'labels': labels,
+        'components': components,
+    }
+
+    # Only add status if it has changed.
+    # Brittle - we should be pulling the equivalent of 'new' from the policy.
+    if issue.status != 'NOT STARTED':
```
oliverchang: are these statuses the same for all jira projects? Or can they be different?

src/appengine/libs/issue_management/jira/__init__.py
bhatkishor: Will do

src/appengine/libs/issue_management/jira/issue_tracker_manager.py
bhatkishor: ACK

src/appengine/libs/issue_management/jira/issue_tracker_manager.py
bhatkishor: ACK

src/appengine/libs/issue_management/jira/issue_tracker_manager.py
bhatkishor: On it

src/appengine/libs/issue_management/jira/issue_tracker_manager.py
bhatkishor: There is a default set of statuses that ships with Jira, but you can also define new ones or edit the defaults as an administrator. Different projects may have different statuses available to them.

bhatkishor: I guess the other open item to add to my list above is to figure out how to grab the list of possible statuses for a given job.

bhatkishor:
Dumb question - how do I ensure that the Jira client can be instantiated with proper creds in my tests (they're sourced from config under normal operation)? Should I just be `mock`ing any actions/functions that involve interfacing with the external Jira host?

mbarbella-chromium:
Right, it would all need to be mocked. We won't have Jira running in our test environments so properly setting it up won't be an option.

inferno-chromium:
@oliverchang @mbarbella-chromium - can you take another look

src/appengine/libs/issue_management/jira/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+    return self.jira_issue.fields.reporter
+
+  @reporter.setter
+  def reporter(self, new_reporter):
+    self.jira_issue.fields.reporter = new_reporter
+
+  @property
+  def is_open(self):
+    """Whether the issue is open."""
+    if self.jira_issue.resolution not in ['Closed', 'Done', 'Resolved']:
```
oliverchang: nit: just

```python
return self.jira_issue.resolution not in ['Closed', 'Done', 'Resolved']
```

src/appengine/libs/issue_management/jira/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+      return True
+    return False
+
+  @property
+  def closed_time(self):
+    res_date = self.jira_issue.fields.resolutiondate
+    if res_date is not None:
+      date = res_date[:res_date.index('T')]
+      return datetime.datetime.strptime(date, '%Y-%m-%d')
```
oliverchang: I'm assuming the date is the most granular we can get here?

src/appengine/libs/issue_management/jira/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+    return self._labels
+
+  @property
+  def components(self):
+    """The issue component list."""
+    return self._components
+
+  def save(self):
+    """Save the issue."""
+
```
oliverchang: nit: no need for this newline.

src/appengine/libs/issue_management/jira/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+    self._itm = itm
+
+  @property
+  def project(self):
+    return self._itm.project_name
+
+  def new_issue(self):
+    jira_issue = self._itm.create()
+    wrapped_issue = Issue(self._itm, jira_issue)
+    return wrapped_issue
```
oliverchang: nit: no need for `wrapped_issue` var.

src/appengine/libs/issue_management/jira/issue_tracker_manager.py
oliverchang: I'm assuming these are the default then? Alternatively, could we just always update the status regardless of the existing value, just to avoid this brittleness?

src/appengine/libs/issue_management/jira/issue_tracker_manager.py
```diff
@@ -0,0 +1,129 @@
...
+      status = {'name': issue.status}
+      update_fields['status'] = status
+
+    if issue.assignee is not None:
+      assignee = {'name': issue.assignee}
+      update_fields['assignee'] = assignee
+
+    # Again brittle - need to pull these strings from policy.
+    if 'Critical - P1' in labels:
```
oliverchang: Yes, the best approach here is to define these in the policy config, and save it in the IssueTrackerManager so that we can access it here. 

You can add a field to the jira policy config like so:

```
project_name:
  type: jira
  status:
    assigned: ...
    ...: ...
  priority_labels:
    Critical: 'Critical - P1'
```

etcc

If these are the default, then this PR is OK for now.

bhatkishor: Those are default priorities from what I can see.

For statuses:
> In the default workflow, issues start as being Open, progressing to In Progress, Resolved and then Closed.

I'll `s/'NOT STARTED'/'Open'` for now... but I want to play with the status update more before I set it to always update. My Jira was giving me errors if I tried to update when there was no change.

This will work for now, but I'll follow up with a PR for a more elegant solution for both of these things.

src/appengine/libs/issue_management/jira/__init__.py
bhatkishor: Ack

src/appengine/libs/issue_management/jira/__init__.py
bhatkishor: Ack

src/appengine/libs/issue_management/jira/__init__.py
bhatkishor: Ack

src/appengine/libs/issue_management/jira/__init__.py
bhatkishor: No, actually - I was looking at the Monorail tests and didn't test my assumption that you needed it to be y-m-d elsewhere in the code, heh. I'll return the hours/minutes/seconds too.

bhatkishor: Oh great, UTC conversions.

src/appengine/libs/issue_management/jira/issue_tracker_manager.py
bhatkishor: Oops, rolled my comment meant for this into https://github.com/google/clusterfuzz/pull/1321#discussion_r369697191

src/appengine/libs/issue_management/jira/issue_tracker_manager.py
oliverchang: SGTM, thanks!

oliverchang:
Thanks for addressing all our comments! Let's get this merged. Looking forward to more from you :)

PR #1326, opened 2020-01-10
Delete old tokenizer and replace them with new tokenizer

mpherman2:
I did a search to make sure there were not any more references to the old tokenizer,  but if you can think of any other place they were used please let me know.


src/python/bot/minimizer/utils.py
```diff
@@ -130,9 +131,9 @@ def single_test_run(test_path):
   return True
 
 
-def tokenize(data, level=0):
```
mbarbella-chromium: Is this function actually referenced anywhere? This was intended as a lightweight simple test, and I think the new tokenizer is probably overkill for what I intended here. If we can just delete it that would be ideal, but if it's still referenced in any of the main functions for minimizers it's fine to leave it as-is.

mpherman2: https://github.com/google/clusterfuzz/blob/3722e18b8f266a5de9d4baa068fb2aab58f4501b/src/python/bot/minimizer/chunk_minimizer.py#L53

mbarbella-chromium: Thanks for checking. Seems fine as-is, then.

PR #1330, opened 2020-01-13
Write simple unit tests for delta and chunk minimizers

mpherman2:


src/python/tests/core/bot/minimizer/chunk_minimizer_test.py
```diff
@@ -0,0 +1,51 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tests for the Chunk minimizer."""
+
+import unittest
+
+from bot.minimizer import chunk_minimizer
+
+
+class DeltaMinimizerTest(unittest.TestCase):
```
mbarbella-chromium: Nit: update class name

mpherman2: Good catch. Done

mpherman2:
I created a base class to use for both tests. Let me know if this is what you had in mind. Thanks

src/python/tests/core/bot/minimizer/base_minimizer_tester.py
```diff
@@ -0,0 +1,48 @@
+# Copyright 2019 Google LLC
```
Dor1s: nit: 2020 :)

Dor1s: and in other files too

mpherman2: Done

PR #1331, opened 2020-01-13
Minimize task: do not remove UBSAN_OPTIONS=silence_unsigned_overflow=1.

Dor1s:
PTAL. I don't fully like this because it means we will never remove that flag even for non OSS-Fuzz jobs (which may not need this option at all, as they don't enable unsigned integer overflow check).

However, I suspect that `UBSAN_OPTIONS` are always non empty anyways, therefore it might be acceptable to always have this flag in there, even when it's not used.

See https://github.com/google/oss-fuzz/issues/3227 and https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=19679 for more info.


Dor1s:
Ended up making this change OSS-Fuzz specific, just because there is already another `if utils.is_oss_fuzz():` branch in the minimize task code.

Another option could be to re-define the default value during runtime (e.g. https://cs.chromium.org/chromium/src/build/sanitizers/sanitizer_options.cc?l=161&rcl=fcc89ebb804f7efd5442985bdf12f7c8e40419e4), but we don't have a good way to link this into every fuzz target, plus a few projects intentionally enabled that check.

PR #1333, opened 2020-01-13
Add Radamsa in process mutation strategy

mpherman2:
This is just a first draft of the changes to clusterfuzz needed to get in process radamsa as a strategy. Still waiting on getting the radamsa .so to test if it works, but I would appreciate it if you took a look to see if this looks like the right place to start, or if anything is missing. Thanks.


jonathanmetzman:
The title "First draft of changes made to clusterfuzz" isn't descriptive of what this does. When this gets merged it should use a better message.
Also, see travis failure

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1497,6 +1497,20 @@ def use_mutator_plugin(target_name, extra_env):
   return True
 
 
+def use_radamsa_in_process(extra_env):
+  """Decide whether to use Radamsa in process. If yes, add the path to the
+  radamsa so to LD_PRELOAD in |extra_env| and return True."""
```
jonathanmetzman: I would write `shared object` here. `so` is confusing since it is an english word.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1497,6 +1497,20 @@ def use_mutator_plugin(target_name, extra_env):
...
+
+  #Like above. Not sure if there is a reason why this would work for windows
```
jonathanmetzman: Spaces after `#`. I'm very surprised linting/formatting didn't catch this.
I'm not sure what `Like above` means in this context. And just say that `mutator plugin doesn't work on windows` and leave out the `Not sure if there is a reason why this would...` part, doesn't add anything and it seems sarcastic :-)
I think you are actually asking a question though, so the answer is that it wont work on Windows.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1497,6 +1497,20 @@ def use_mutator_plugin(target_name, extra_env):
...
+  #When Mutator plugin does not.
+  if environment.platform() == "WINDOWS":
+    return False
+
+  radamsa_path = "HARDCODED PATH TO SO"
```
jonathanmetzman: I would add a TODO explaining that you need to get the real path to this.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1497,6 +1497,20 @@ def use_mutator_plugin(target_name, extra_env):
   return True
 
 
+def use_radamsa_in_process(extra_env):
```
jonathanmetzman: I would have named this "use_radamsa_mutator_plugin" the current name might be confusing (in which process?).

jonathanmetzman: The current interface for mutator plugins is unfinished but I'd rather radmsa reused it (extending it if needed to support a plugin that could be applied to any fuzzer) rather than added a new code path that only applied to radamsa.
@mbarbella-chromium WDYT?
You could change the naming scheme used for plugins to support wildcards (`*`). As is this has a problem where it will break mac/msan binaries because it will give them a shared object only asan binaries can use.

src/python/fuzzing/strategy.py
```diff
@@ -38,6 +38,8 @@
 FORK_STRATEGY = Strategy(name='fork', probability=0.50, manually_enable=False)
 MUTATOR_PLUGIN_STRATEGY = Strategy(
     name='mutator_plugin', probability=0.50, manually_enable=True)
+IN_PROCESS_RADAMSA_STRATEGY = Strategy(
+    name='in_process_radamsa', probability=1.0, manually_enable=False)
```
jonathanmetzman: Do we actually want an 100% probability?

src/python/bot/fuzzers/libfuzzer.py
jonathanmetzman: If you have a question for reviewers it's probably better to do so using the code review tool rather than leaving it in source. It's more likely to be seen this way and can't be accidentally left in the code.

src/python/fuzzing/strategy.py
jonathanmetzman: ACK

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_stats_test.py
```diff
@@ -99,6 +99,7 @@ def test_parse_log(self):
         'strategy_corpus_mutations_ml_rnn': 0,
         'strategy_corpus_subset': 50,
         'strategy_fork': 1,
+        'strategy_radamsa_mutator_plugin': 0,
```
Dor1s: nit: maybe `strategy_mutator_plugin_radamsa`? to be similar to `strategy_corpus_mutations_ml_rnn` and `strategy_corpus_mutations_radamsa`

src/python/fuzzing/strategy.py
```diff
@@ -38,6 +38,8 @@
 FORK_STRATEGY = Strategy(name='fork', probability=0.50, manually_enable=False)
 MUTATOR_PLUGIN_STRATEGY = Strategy(
     name='mutator_plugin', probability=0.50, manually_enable=True)
+RADAMSA_MUTATOR_PLUGIN_STRATEGY = Strategy(
+    name='radamsa_mutator_plugin', probability=1.0, manually_enable=False)
```
Dor1s: 1.0 means 100%, you probably meant `0.01` for starters

mpherman2: It was 1.0 while testing. I agree that 0.01 is where we want to start it.

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_stats_test.py
mpherman2: Done

mpherman2:
I added the radamsa shared object and responded to comments. Please take another look and let me know if this needs any more changes.

src/python/bot/fuzzers/libfuzzer.py
jonathanmetzman: Ping

mbarbella-chromium: I'm having trouble picturing exactly what reusing the existing interface would look like here. I also don't understand the point about wildcards, could you elaborate?

Either way I do agree that reusing the existing interface would be preferable if we can.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1497,6 +1497,22 @@ def use_mutator_plugin(target_name, extra_env):
   return True
 
 
+def use_radamsa_mutator_plugin(extra_env):
+  """Decide whether to use Radamsa in process. If yes, add the path to the
+  radamsa shared object to LD_PRELOAD in |extra_env| and return True."""
+
+  # Mutator Plugin doesn't work on windows
+  if environment.platform() == "WINDOWS":
```
mbarbella-chromium: As-is this should probably bail out on any non-LINUX values. We'll also want to validate that this is an asan job for the time being (based on Jonathan's previous comment).

src/python/bot/fuzzers/libfuzzer.py
jonathanmetzman: I mean the mutator plugins have a hacky naming scheme for indicating that a given plugin should be used with binaries for a certain platform, arch, sanitizer. This doesn't do that I assume it will therefore break (at least) i386 and msan binaries.

src/python/bot/fuzzers/libfuzzer.py
jonathanmetzman: And i386

jonathanmetzman:
I would test this in the oss-fuzz runner btw. Otherwise I would guess a binary built on your host will have an issue with the libcpp available on oss-fuzz.

mpherman2:
How would I go about testing this in the oss-fuzz runner?

src/python/bot/fuzzers/libfuzzer.py
mpherman2: I changed it so that it bails out of any job that isn't Linux or isn't ASAN. Is that correct?

jonathanmetzman: I 3 8 6

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1497,6 +1497,24 @@ def use_mutator_plugin(target_name, extra_env):
   return True
 
 
+def use_radamsa_mutator_plugin(extra_env):
+  """Decide whether to use Radamsa in process. If yes, add the path to the
+  radamsa shared object to LD_PRELOAD in |extra_env| and return True."""
+
+  # Radamsa will only work on LINUX ASAN jobs
```
mbarbella-chromium: Nit: punctuation

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1497,6 +1497,24 @@ def use_mutator_plugin(target_name, extra_env):
...
+  # TODO Include Architecture Info in Job Definition and exclude I386
```
mbarbella-chromium: "TODO" -> "TODO(mpherman):"

src/python/tests/core/bot/fuzzers/libFuzzer/engine_test.py
```diff
@@ -342,6 +342,7 @@ def mock_merge(*args, **kwargs):  # pylint: disable=unused-argument
         'strategy_dataflow_tracing': 0,
         'strategy_fork': 0,
         'strategy_mutator_plugin': 0,
+        'strategy_mutator_plugin_radamsa': 0,
```
Dor1s: do we also need to update `src/python/metrics/fuzzer_stats_schema.py`? //cc @oliverchang

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1612,5 +1630,9 @@ def pick_strategies(strategy_pool, fuzzer_path, corpus_directory,
       use_mutator_plugin(target_name, extra_env)):
     fuzzing_strategies.append(strategy.MUTATOR_PLUGIN_STRATEGY.name)
 
+  if (strategy_pool.do_strategy(strategy.MUTATOR_PLUGIN_RADAMSA_STRATEGY) and
+      use_radamsa_mutator_plugin(extra_env)):
+    fuzzing_strategies.append(strategy.MUTATOR_PLUGIN_RADAMSA_STRATEGY.name)
```
Dor1s: you've mentioned earlier today that the mutator might not work for some reason. If that's a real possibility, we might want to have some check to make sure we don't record `'strategy_mutator_plugin_radamsa': 1` in stats when the mutator actually failed to get loaded

jonathanmetzman: Depending on how much we want to do this, I think printing on the first run of the mutation and checking for this print could be a good solution.

src/python/tests/core/bot/fuzzers/libFuzzer/engine_test.py
oliverchang: Yep, we need to add a column there.

oliverchang: Another idea: rather than create a column for every single plugin we have, how about having a column which has a string value instead, with the value being the plugin name?

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1497,6 +1497,24 @@ def use_mutator_plugin(target_name, extra_env):
   return True
 
 
+def use_radamsa_mutator_plugin(extra_env):
+  """Decide whether to use Radamsa in process. If yes, add the path to the
+  radamsa shared object to LD_PRELOAD in |extra_env| and return True."""
+
+  # Radamsa will only work on LINUX ASAN jobs.
+  # TODO(mpherman) Include Architecture Info in Job Definition and exclude I386
+  if environment.platform() != "LINUX" or environment.get_value(
```
oliverchang: nit: single quotes for strings.

src/python/bot/fuzzers/libfuzzer.py
oliverchang: Printing can be error prone, especially if we need every plugin to conform to some format. Maybe we can modify libFuzzer to add an argument for mutator plugins (and which checks that it loads properly).

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1497,6 +1497,24 @@ def use_mutator_plugin(target_name, extra_env):
...
+      "MEMORY_TOOL") != "ASAN":
+    return False
+
+  radamsa_path = os.path.join(
+      environment.get_value('ROOT_DIR'), 'resources', 'platform', 'linux',
```
oliverchang: Use environment.get_platform_resources_directory() for this.

src/python/tests/core/bot/fuzzers/libFuzzer/engine_test.py
mpherman2: For now I just added another column. Let me know if I didn't add it right.

src/python/bot/fuzzers/libfuzzer.py
mpherman2: I have not had any success getting the libradamsa shared object to output to the log.

mpherman2:
Forgot to add the README. just added it. Let me know if that is what you had in mind.

resources/platform/linux/README.md
```diff
@@ -0,0 +1,23 @@
+In order to update the libradamsa shared object, clone the radamsa repo
+available at: 
+
+https://gitlab.com/akihe/radamsa
+
+Add the following method to the file c/lib.c
+
+```
+size_t LLVMFuzzerCustomMutator(uint8_t *Data, size_t Size,
+                                            size_t MaxSize, unsigned int Seed ){
```
mbarbella-chromium: Nit: alignment (were there any tabs in here?)

resources/platform/linux/README.md
```diff
@@ -0,0 +1,23 @@
+In order to update the libradamsa shared object, clone the radamsa repo
+available at: 
+
+https://gitlab.com/akihe/radamsa
```
mbarbella-chromium: Include a git clone command, and place in a code section.

resources/platform/linux/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+Add the following method to the file c/lib.c
```
mbarbella-chromium: Ideally include the following as a patch, or at the very least something like echo changes >> c/lib.c

resources/platform/linux/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+```
+size_t LLVMFuzzerCustomMutator(uint8_t *Data, size_t Size,
+                                            size_t MaxSize, unsigned int Seed ){
+       static bool init = false;
+       if(!init){
```
mbarbella-chromium: Spacing like "if (...) {"

resources/platform/linux/README.md
```diff
@@ -0,0 +1,23 @@
...
+available at: 
+
+https://gitlab.com/akihe/radamsa
+
+Add the following method to the file c/lib.c
+
+```
+size_t LLVMFuzzerCustomMutator(uint8_t *Data, size_t Size,
+                                            size_t MaxSize, unsigned int Seed ){
+       static bool init = false;
```
mbarbella-chromium: Nit: s/init/initialized/

resources/platform/linux/README.md
```diff
@@ -0,0 +1,23 @@
+In order to update the libradamsa shared object, clone the radamsa repo
```
mbarbella-chromium: I wasn't too clear in the previous comment, but move this and the .so to a subdirectory under this one. It's confusing for this to be included as a README for the full linux directory. That will also need a change to update the path you search for to find the .so.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1497,6 +1497,24 @@ def use_mutator_plugin(target_name, extra_env):
   return True
 
 
+def use_radamsa_mutator_plugin(extra_env):
+  """Decide whether to use Radamsa in process. If yes, add the path to the
+  radamsa shared object to LD_PRELOAD in |extra_env| and return True."""
+
+  # Radamsa will only work on LINUX ASAN jobs.
+  # TODO(mpherman) Include Architecture Info in Job Definition and exclude I386
```
mbarbella-chromium: Nit: punctuation, I386->i386

resources/platform/linux/README.md
mpherman2: Done

resources/platform/linux/README.md
mpherman2: Done

resources/platform/linux/README.md
mpherman2: Done

resources/platform/linux/README.md
mpherman2: Done

resources/platform/linux/README.md
mpherman2: Done

src/python/bot/fuzzers/libfuzzer.py
Dor1s: I think a can make a change like this in libFuzzer. Shouldn't be blocking you right now, but let's have a TODO to address in future?

src/python/tests/core/bot/fuzzers/libFuzzer/engine_test.py
Dor1s: Looks right to me, thanks!

src/python/bot/fuzzers/libfuzzer.py
Dor1s: Uploaded https://reviews.llvm.org/D73136 for review.

Dor1s: @mpherman2 but don't wait on this, even if it gets landed, it'll take a couple weeks for the change to propagate to all ecosystems where ClusterFuzz is running.

resources/platform/linux/radamsa/mutator.patch
```diff
@@ -0,0 +1,27 @@
...
+@@ -65,6 +66,16 @@ size_t radamsa(uint8_t *ptr, size_t len, uint8_t *target, size_t max, unsigned i
+    return copy_list(target, res, max);
+ }
+ 
++size_t LLVMFuzzerCustomMutator(uint8_t *Data, size_t Size,
++                                size_t MaxSize, unsigned int Seed ){
++    static bool initialized = false;
++    if (!initialized) {
++      radamsa_init();
++      init = true;
```
jonathanmetzman: Shouldn't this be `initialized`?

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1497,6 +1497,24 @@ def use_mutator_plugin(target_name, extra_env):
   return True
 
 
+def use_radamsa_mutator_plugin(extra_env):
+  """Decide whether to use Radamsa in process. If yes, add the path to the
+  radamsa shared object to LD_PRELOAD in |extra_env| and return True."""
+
+  # Radamsa will only work on LINUX ASAN jobs.
+  # TODO(mpherman) Include Architecture Info in Job Definition and exclude i386.
```
jonathanmetzman: I think we usually put a colon after the paren. 
Also this sentence has a weird amount of capitalized words.

resources/platform/linux/radamsa/README.md
```diff
@@ -0,0 +1,14 @@
+In order to update the libradamsa shared object, clone the radamsa repo
+
+`git clone https://gitlab.com/akihe/radamsa.git`
+
+Patch the changes into the directory using the mutator.patch file in this dir
```
jonathanmetzman: nit: "directory" not dir

mpherman2:
@jonathanmetzman Can you take another look another look at this when you get the chance? Thanks

PR #1336, opened 2020-01-14
Reland "[Python 2->3] Fix remaining futurize issues." (attempt 2)

mbarbella-chromium:


src/python/bot/testcase_manager.py
```diff
@@ -1001,8 +1001,17 @@ def get_command_line_for_application(file_to_run='',
...
+  # In some cases, APP_REVISION is not an integer. In these cases, we assume
+  # the revisions are recent enough that we should remove the command line.
+  # This primarily affects impact tasks.
+  try:
+    revision = int(environment.get_value('APP_REVISION', 0))
+  except ValueError:
+    revision = 0
+
   if (environment.get_value('APP_NAME') in content_shell_app_names and
-      environment.get_value('APP_REVISION', 0) >= 558998):
```
mbarbella-chromium: If bumping MIN_REVISION on these jobs to this point is an option, I'd rather just remove the hacks. WDYT?

oliverchang: That seems much better to me. 558998 was back in May 2018 and seems old enough that we don't care about bisects to back there anymore? @inferno-chromium WDYT?

oliverchang:
(Maybe add a chmod to `src/python/bot/fuzzers/afl/launcher.py`, but I don't see why your change would have broken this though..)

src/python/base/utils.py
```diff
@@ -782,16 +787,20 @@ def wait_until_timeout(threads, thread_timeout):
...
+  # TODO(mbarbella): Require callers to pass strings or bytes then reduce this
+  # back to the original retry limit.
+  for _ in range(retry_limit + 1):
     try:
       with open(file_path, file_mode) as file_handle:
-        file_handle.write(content_string)
-    except:
+        file_handle.write(content)
+    except TypeError:
+      content = str(content).encode('utf-8')
```
inferno-chromium: Dont understand what this branch does.

src/python/base/utils.py
```diff
@@ -782,16 +787,20 @@ def wait_until_timeout(threads, thread_timeout):
...
+      continue
+    except EnvironmentError:
```
inferno-chromium: Where is this EnvironmentError defined ?

src/python/bot/testcase_manager.py
inferno-chromium: Yes it might be already upped, check jobs and remove if needed.

src/python/bot/untrusted_runner/untrusted.py
```diff
@@ -210,8 +213,13 @@ def _get_tls_cert_and_key():
 
     return cert_contents, key_contents
 
-  return (str(compute_metadata.get('instance/attributes/tls-cert')),
-          str(compute_metadata.get('instance/attributes/tls-key')))
+  # TODO(mbarbella): Remove this after migrating to Python 3. The grpc library
+  # has explicit type checks against str.
+  cert_contents = str(compute_metadata.get('instance/attributes/tls-cert'))
+  cert_contents = future_utils.native(cert_contents).encode()
```
inferno-chromium: Maybe create a helper for

future_utils.native(str(something)).encode()
and use in all places.

src/python/datastore/data_handler.py
```diff
@@ -375,7 +377,7 @@ def get_formatted_reproduction_help(testcase):
   # Since this value may be in a job definition, it's non-trivial for it to
   # include newlines. Instead, it will contain backslash-escaped characters
   # that must be converted here (e.g. \n).
-  help_format = help_format.decode('unicode-escape')
+  help_format = help_format.encode().decode('unicode-escape')
```
inferno-chromium: Why both encode and decode ?

mbarbella-chromium: In Python 2 it was valid to re-encode a unicode object but that's no longer the case in 3. This transformation goes from normal utf-8 to bytes to escaped encoded unicode.

src/python/base/utils.py
mbarbella-chromium: EnvironmentError is the base class for IOError and OSError (both builtin). Added a comment to this block that should make its purpose clearer.

src/python/base/utils.py
mbarbella-chromium: Added a few comments for clarity.

There are a few problems here, but all stem from the fact that we want this to accept arbitrary input. The two specific problems here are that in Python 3 write fails if the input isn't bytes if opened as wb (or str if opened w), and that we can no longer call encode() on a bytes to keep it as-is (it's no longer defined). This hack lets us catch the type errors and encode only if it was required, without breaking when we do pass bytes.

I think the real fix would be to expect a bytes as input to this function, but that would be a very large and invasive cleanup that can come later.

src/python/bot/testcase_manager.py
mbarbella-chromium: Removed.

src/python/bot/untrusted_runner/untrusted.py
mbarbella-chromium: Done.

mbarbella-chromium:
Just to give a status update on this, I have it running on a test instance and the only remaining exception seems to be that AFL one. I'm letting it run with some additional logging, but still no obvious clues why the string changes would be related to it.

oliverchang:
I just SSHed into a test bot a see this:

```
-rw-r--r-- 1 clusterfuzz clusterfuzz    56736 Jan 17 05:43 afl-analyze
-rw-r--r-- 1 clusterfuzz clusterfuzz    41352 Jan 17 05:43 afl-as
-rw-r--r-- 1 clusterfuzz clusterfuzz    11391 Jan 17 05:43 afl-cmin
-rw-r--r-- 1 clusterfuzz clusterfuzz   347120 Jan 17 05:43 afl-fuzz
-rw-r--r-- 1 clusterfuzz clusterfuzz    24768 Jan 17 05:43 afl-gcc
-rw-r--r-- 1 clusterfuzz clusterfuzz    23896 Jan 17 05:43 afl-gotcpu
-rw-r--r-- 1 clusterfuzz clusterfuzz     4912 Jan 17 05:43 afl-plot
-rw-r--r-- 1 clusterfuzz clusterfuzz    47600 Jan 17 05:43 afl-showmap
-rw-r--r-- 1 clusterfuzz clusterfuzz    63648 Jan 17 05:43 afl-tmin
-rw-r--r-- 1 clusterfuzz clusterfuzz     3654 Jan 17 05:43 afl-whatsup
-rw-r--r-- 1 clusterfuzz clusterfuzz 27603616 Jan 17 05:43 data_context_fuzzer
-rw-r--r-- 1 clusterfuzz clusterfuzz        6 Jan 17 05:43 data_context_fuzzer.labels
-rw-r--r-- 1 clusterfuzz clusterfuzz       12 Jan 17 05:43 REVISION
```

None of these have execute bits... not sure why your change would have affected things here though :/

oliverchang:
AFAICT We were previously relying on the permissions being set correctly in the zip for AFL. So, there must be some issue in the unarchiving. 

This change changes one thing in archive.py: 

```python
+ from builtins import oct
```

`oct` is used like so:

```python
        if oct(external_attr).startswith(FILE_ATTRIBUTE):
          old_mode = external_attr & 0o7777
          new_mode = external_attr & 0o777
          new_mode |= 0o440
          needs_execute_permission = external_attr & 100

          if new_mode != old_mode or needs_execute_permission:
            # Default extract condition is 640 which is safe.
            # |new_mode| might have read+write+execute bit for
            # others, so remove those.
            new_mode &= 0o770

            os.chmod(extracted_path, new_mode)
```

Now the `oct` from `builtins` changes its behaviour: 

```python
>>> oct(123)
'0173'
>>> from builtins import oct
>>> oct(123)
'0o173'
```

This causes the `if oct(external_attr).startswith(FILE_ATTRIBUTE)` bit to fail... and thus not get the right permissions set.

mbarbella-chromium:
Thanks for the analysis! I'm going to let it run with the fix for that on the test instance over the weekend, and land on Tuesday if it sticks without any new exceptions.

mbarbella-chromium:
I think everything is resolved here. Going to try landing this again and hoping it will stick.

PR #1337, opened 2020-01-14
Always include strategies in libFuzzer stats, even if unused.

### LINKED ISSUE #1304: Default values for strategy stats in BigQuery
We used to initialize all strategy columns with 0, but we don't seem to be doing it now, e.g.:

```sql
SELECT
  count(1)
FROM
  `libFuzzer_stats.TestcaseRun`
WHERE
  strategy_dataflow_tracing = 0
  AND timestamp >= 1576195200 /* Dec 13th 00:00 UTC */
  AND timestamp <= 1577059200 /* Dec 23th 00:00 UTC */
```

returns `0`, while

```sql
...
  strategy_dataflow_tracing = 1
...
```

makes it `2,345`, and

```sql
...
  strategy_dataflow_tracing IS NULL
...
```

makes it a large number as well.

Same for e.g. `  strategy_recommended_dict = 0` and likely other strategies.

IMO this is not a big deal, but I'd prefer a default value instead of `NULL`, as it simplifies query writing plus `NULL` works as an indicator that a particular strategy wasn't supported at all. Right now it's indistinguishable whether we do not support a strategy or it just isn't being used.

This also might be the reason why Data Studio charts aren't working //cc @mbarbella-chromium

mbarbella-chromium:
Intended to fix #1304. Max, PTAL when you have time.

### DIFF
--- a/src/python/bot/fuzzers/libFuzzer/engine.py
+++ b/src/python/bot/fuzzers/libFuzzer/engine.py
@@ -292,8 +292,7 @@ class LibFuzzerEngine(engine.Engine):
         stats.parse_performance_features(
             log_lines,
             options.strategies,
-            options.arguments,
-            include_strategies=False))
+            options.arguments))
 
     # Set some initial stat overrides.
     timeout_limit = fuzzer_utils.extract_argument(
--- a/src/python/bot/fuzzers/libFuzzer/engine.py
+++ b/src/python/bot/fuzzers/libFuzzer/engine.py
@@ -289,10 +289,8 @@ class LibFuzzerEngine(engine.Engine):
 
     # Extend parsed stats by additional performance features.
     parsed_stats.update(
-        stats.parse_performance_features(
-            log_lines,
-            options.strategies,
-            options.arguments))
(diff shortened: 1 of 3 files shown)


src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -292,8 +292,7 @@ def fuzz(self, target_path, options, reproducers_dir, max_time):
         stats.parse_performance_features(
             log_lines,
             options.strategies,
-            options.arguments,
-            include_strategies=False))
```
mbarbella-chromium: It's been a while since I've looked at this code and I'm not sure why we explicitly avoided adding strategies here. Any idea why this was added initially or if this would have any negative side-effects that I missed?

Dor1s: turns out @oliverchang and I had a conversion on this: https://github.com/google/clusterfuzz/pull/835/files#r313988491 TL;DR that was a temporary feature

mbarbella-chromium:
Thanks for the review!

PR #1341, opened 2020-01-16
Enable honggfuzz by default in project setup.

oliverchang:


oliverchang:
/gcbrun. TBR for simple change as discussed.

PR #1343, opened 2020-01-16
[fuchsia] Finish corpus pruning support + add tests.

flowerhack:
Corpus pruning already mostly-worked in  Fuchsia by virtue of the
`merge` command being a thing. However, the
sync-files-between-device-and-host step was missing a beat: corpora were
sync'd properly, but "bad units" (crashes, leaks, ooms, etc) weren't
getting pulled down.  (They do get pulled down in `fuzz`, of course, no
fears there.)

This commit pulls down those "bad units", and also adds a test for the
same.

Test can be run locally via:

INTEGRATION=1 FUCHSIA_TESTS=1 python butler.py py_unittest -t core -p "*corpus_pruning*" -u


flowerhack:
@eepeep

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -617,7 +632,8 @@ def fuzz(self,
     with open(self.fuzzer.logfile) as logfile:
       symbolized_output = logfile.read()
 
-    self._pull_new_corpus_from_target_to_host(corpus_directories)
+    self._pull_new_corpus_from_target_to_host(corpus_directories,
```
oliverchang: Can we move this to before line 631? Then we can get rid of the crash artifact copying logic in `process_logs_and_crash` which is otherwise duplicated.

src/python/tests/core/bot/tasks/corpus_pruning_task_data/fuchsia/AA
```diff
@@ -0,0 +1 @@
+AA
```
oliverchang: nit: add some more duplicates to ensure they get removed? Or are AA and CC expected to reach the same coverage? If they reach the same coverage, let's just make the file contents the same for determinism.

src/python/tests/core/bot/tasks/corpus_pruning_task_test.py
```diff
@@ -94,7 +94,9 @@ def mocked_unpack_seed_corpus_if_needed(*args, **kwargs):
     self.corpus_dir = os.path.join(self.corpus_bucket, 'corpus')
     self.quarantine_dir = os.path.join(self.corpus_bucket, 'quarantine')
     self.shared_corpus_dir = os.path.join(self.corpus_bucket, 'shared')
+    self.fuchsia_corpus_dir = os.path.join(self.corpus_bucket, 'fuchsia')
```
oliverchang: Put these fuchsia specific ones into the fuchsia test subclass.

src/python/tests/core/bot/tasks/corpus_pruning_task_data/fuchsia/AA
flowerhack: huh, while fixing this i noticed something curious!

i would expect AA and CC to reach the same coverage, yes.  but i would *also* expect BB to reach the same coverage.  and yet: when i have (AA1, AA2, H, HI!), i only get *one* corpus unit ("H", as expected), but when i add BB to the mix (e.g. AA1, AA2, B, H, HI!), i get *two* corpus units ("H", "BB").

this is unintuitive and weird and making me wonder if i'm misunderstanding something.  do you have any hunches?  the fuzzer we're using is here: https://fuchsia.googlesource.com/fuchsia/+/refs/heads/master/examples/fuzzer/trap.cc

src/python/tests/core/bot/tasks/corpus_pruning_task_test.py
flowerhack: done!

src/python/tests/core/bot/tasks/corpus_pruning_task_data/fuchsia/AA
oliverchang: As discussed, there may be some non determinism here with the merge algorithm or possibly the target. Let's just go with a simpler input which gives us the expected result.

flowerhack: discussed offline--fixed by just having AA1 and AA2

oliverchang:
/gcbrun. (rebased to fix lint)

PR #1344, opened 2020-01-16
Add an extra layer of line minimization before JS and HTML minimization. 

mpherman2:
Hopefully a fix to the fact that line minimization on bots does not minimize as well as it does locally


src/python/bot/minimizer/html_minimizer.py
```diff
@@ -16,6 +16,7 @@
 from __future__ import absolute_import
 
 from builtins import object
+from builtins import range
```
Dor1s: why is this needed?

mpherman2: The linter said that I needed it.

mbarbella-chromium: More specifically this would be needed to ensure we're using range behavior compatible with Python 3's.

mbarbella-chromium:
Late comment: could you include comments before each of those to explain why we're doing this second pass?

Issue #1346, opened 2020-01-17
Tracking bug: improving DFT-based fuzzing strategy


Dor1s:
Continuation of https://github.com/google/clusterfuzz/issues/503.

The corresponding issue on OSS-Fuzz side is https://github.com/google/oss-fuzz/issues/1632

PR #1347, opened 2020-01-17
Dataflow strategy: use traces from the build, auto-focus, no fork (#1346).

Dor1s:
Instead of `-collect_data_flow=<DATAFLOW_BINARY> -fork=1`, use `-data_flow_trace=<DATAFLOW_TRACE_DIR> -focus_function=auto` and intentionally do not use the fork mode (at least for now).

### DIFF
--- a/src/python/bot/fuzzers/libfuzzer.py
+++ b/src/python/bot/fuzzers/libfuzzer.py
@@ -71,6 +71,9 @@ MAX_OUTPUT_LEN = 1 * 1024 * 1024  # 1 MB
 CRASH_TESTCASE_REGEX = (r'.*Test unit written to\s*'
                         r'(.*(crash|oom|timeout|leak)-.*)')
 
+# Currently matches oss-fuzz/infra/base-images/base-runner/collect_dft#L34.
+DATAFLOW_TRACE_DIR_SUFFIX = '_dft'
+
 
 class LibFuzzerException(Exception):
   """LibFuzzer exception."""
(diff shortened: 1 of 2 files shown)


Dor1s:
Marty, could you also review this one, and I'll deploy all recent changes at once then?

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1599,8 +1605,8 @@ def pick_strategies(strategy_pool, fuzzer_path, corpus_directory,
   is_ephemeral = environment.is_ephemeral()
 
   if (not is_fuchsia and not is_android and not is_ephemeral and
-      (use_dataflow_tracing or
-       strategy_pool.do_strategy(strategy.FORK_STRATEGY))):
+      not use_dataflow_tracing and
```
inferno-chromium: So, you don't want fork mode to be used in dataflow tracing anymore ?

Dor1s: yes

mbarbella-chromium: Could you add a comment explaining why they should be mutually exclusive here? Do they actually need to be mutually exclusive or could we still occasionally pick both?

Dor1s: Done, also fixed a stale comment (that somehow went too above in the code), thanks!

The strategies are not mutually exclusive, but in the fork mode the log is pretty poor, so I want to get readable logs from the DFT runs for now. They include coverage information, focus function selected, and some other DFT-specific stats.

Issue #1348, opened 2020-01-17
fuzzer_and_job_weights cron fails to insert data to BigQuery due to 10000 quota


Dor1s:
https://console.cloud.google.com/errors/CJ7plL-8-LTsUg?time=P30D&project=clusterfuzz-external

Dor1s:
The error is gone, cron job succeeded.

PR #1350, opened 2020-01-17
Follow up fix for merging errors from batch BigQuery inserts (#1348).

Dor1s:
Follow up! Looks like either documentation or my reading of it is not right :)

### DIFF
--- a/src/python/google_cloud_utils/big_query.py
+++ b/src/python/google_cloud_utils/big_query.py
@@ -370,7 +370,7 @@ class Client(object):
       A json explained here:
       https://cloud.google.com/bigquery/docs/reference/rest/v2/tabledata/insertAll
     """
-    result = None
+    result = {}
     for i in range(0, len(inserts), INSERT_BATCH_SIZE):
       response = self._insert_batch(
           inserts[i:min(len(inserts), i + INSERT_BATCH_SIZE)])
@@ -391,6 +391,6 @@ class Client(object):
         if result.get('insertErrors'):
           result['insertErrors'].extend(new_errors)
         else:
-          result = response
+          result['insertErrors'] = new_errors
 
     return result
--- a/src/python/google_cloud_utils/big_query.py
+++ b/src/python/google_cloud_utils/big_query.py
@@ -379,8 +379,18 @@ class Client(object):
       time.sleep(1)
 
       if not result:
+        # Use result from the first batch, appending errors from the rest.
         result = response
       else:
-        result['insertErrors'].extend(response['insertErrors'])
+        # If there are new errors from the current batch, append to the result.
+        new_errors = response.get('insertErrors')
+        if not new_errors:
+          continue
+
+        # Apparently result may not have errors, be careful.
+        if result.get('insertErrors'):
+          result['insertErrors'].extend(new_errors)
+        else:
+          result = response
 
     return result


src/python/google_cloud_utils/big_query.py
```diff
@@ -379,8 +379,18 @@ def insert(self, inserts):
...
+        # If there are new errors from the current batch, append to the result.
+        new_errors = response.get('insertErrors')
+        if not new_errors:
+          continue
+
+        # Apparently result may not have errors, be careful.
+        if result.get('insertErrors'):
+          result['insertErrors'].extend(new_errors)
+        else:
+          result = response
```
inferno-chromium: Why do we overwrite the result here ?

src/python/google_cloud_utils/big_query.py
```diff
@@ -379,8 +379,18 @@ def insert(self, inserts):
       time.sleep(1)
 
       if not result:
+        # Use result from the first batch, appending errors from the rest.
```
inferno-chromium: Why we don't capture result (non-errors) from other parts of query response. This seems weird. @oliverchang for review as well.

src/python/google_cloud_utils/big_query.py
Dor1s: let me give a try to upgrading it

Dor1s: doesn't seem like we can use the version that has it fixed (`2.3.0`: https://github.com/PyCQA/pylint/issues/2767), but I've used `{}` instead of `None` and now it's happy

src/python/google_cloud_utils/big_query.py
```diff
@@ -379,8 +379,18 @@ def insert(self, inserts):
...
         result = response
       else:
-        result['insertErrors'].extend(response['insertErrors'])
+        # If there are new errors from the current batch, append to the result.
+        new_errors = response.get('insertErrors')
+        if not new_errors:
+          continue
+
+        # Apparently result may not have errors, be careful.
+        if result.get('insertErrors'):
```
oliverchang: nit: we can do this much more concisely like so:

```python
result.setdefault('insertErrors', []).extend(new_errors)
```

src/python/google_cloud_utils/big_query.py
```diff
@@ -379,8 +379,18 @@ def insert(self, inserts):
       time.sleep(1)
 
       if not result:
+        # Use result from the first batch, appending errors from the rest.
         result = response
       else:
-        result['insertErrors'].extend(response['insertErrors'])
+        # If there are new errors from the current batch, append to the result.
```
oliverchang: I feel the else makes the flow of the logic a bit confusing here. We can probably just continue after line 383 and unindent the rest.

src/python/google_cloud_utils/big_query.py
oliverchang: Agreed that there is nothing much else in the returned values.

src/python/google_cloud_utils/big_query.py
Dor1s: beautiful, done!

src/python/google_cloud_utils/big_query.py
Dor1s: thanks, done!

PR #1352, opened 2020-01-20
updated the year

vschs007:
updated the year to 2020


Dor1s:
The year reflects when the file was added for the first time.

PR #1356, opened 2020-01-21
Speculative fix for HTML minimizer unicode exception.

mbarbella-chromium:
PTAL. Quick fix for an exception showing up after the python 3 migration CL. We essentially switched this line from using str to unicode, but really wanted raw data. Switching to bytes should give us the behavior we want (working on arbitrary test case data that isn't necessarily utf-8).

### DIFF
--- a/src/python/bot/minimizer/html_minimizer.py
+++ b/src/python/bot/minimizer/html_minimizer.py
@@ -195,7 +195,7 @@ class HTMLMinimizer(minimizer.Minimizer):  # pylint:disable=abstract-method
   @staticmethod
   def combine_tokens(tokens):
     """Combine tokens into a usable format, stripping metadata."""
-    return ''.join([bytes(t) for t in tokens])
+    return b''.join([bytes(t) for t in tokens])
 
   @staticmethod
   def run(data,
--- a/src/python/bot/minimizer/html_minimizer.py
+++ b/src/python/bot/minimizer/html_minimizer.py
@@ -15,9 +15,9 @@
    an HTML test case."""
 from __future__ import absolute_import
 
+from builtins import bytes
 from builtins import object
 from builtins import range
-from builtins import str
 
 import functools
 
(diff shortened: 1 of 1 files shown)


src/python/bot/minimizer/html_minimizer.py
```diff
@@ -190,12 +190,12 @@ def get_tokens_and_metadata(data):
   @staticmethod
   def combine_worker_tokens(tokens, prefix='', suffix=''):
     """Combine tokens for a worker minimizer."""
-    return '%s%s%s' % (prefix, ''.join(tokens), suffix)
+    return prefix + b''.join(tokens) + suffix
```
inferno-chromium: prefix, suffix need to have b'' ?

src/python/bot/minimizer/html_minimizer.py
```diff
@@ -190,12 +190,12 @@ def get_tokens_and_metadata(data):
...
 
   @staticmethod
   def combine_tokens(tokens):
     """Combine tokens into a usable format, stripping metadata."""
-    return ''.join([str(t) for t in tokens])
+    return ''.join([bytes(t) for t in tokens])
```
inferno-chromium: '' need to b'' ?

>>> '' + b'a' + ''
  File "<stdin>", line 1, in <module>
TypeError: can only concatenate str (not "bytes") to str
>>> ''.join([b'a'])
  File "<stdin>", line 1, in <module>
TypeError: sequence item 0: expected str instance, bytes found

mbarbella-chromium: Good catch.

src/python/bot/minimizer/html_minimizer.py
mbarbella-chromium: The values passed are output from combine_tokens, so we should be safe. I'd prefer to avoid explicitly converting to bytes here.

inferno-chromium: i mean the default values of prefix and suffix are not bytes, can you change those.

mbarbella-chromium: Ah, missed that. Done!

Issue #1360, opened 2020-01-22
Issue with non-ascii environment variables


eepeep:
I have some non-ascii characters in my PS1, and this causes problems with some of the code in #1336 .  For example, running some integration tests, I see:

```
Traceback (most recent call last):
  File "~/clusterfuzz/src/python/tests/core/bot/fuzzers/libFuzzer/engine_test.py", line 1017, in test_qemu_logs_returned_on_error
    build_manager.setup_build()
  File "~/clusterfuzz/src/python/build_management/build_manager.py", line 1418, in setup_build
    return setup_trunk_build(bucket_paths, target_weights=target_weights)
  File "~/clusterfuzz/src/python/build_management/build_manager.py", line 1212, in setup_trunk_build
    target_weights=target_weights)
  File "~/clusterfuzz/src/python/build_management/build_manager.py", line 1256, in setup_regular_build
    if build.setup():
  File "~/clusterfuzz/src/python/build_management/build_manager.py", line 783, in setup
    fuchsia.device.initial_qemu_setup()
  File "~/clusterfuzz/src/python/platforms/fuchsia/device.py", line 99, in initial_qemu_setup
    extend_fvm(qemu_vars['fuchsia_resources_dir'], qemu_vars['drive_path'])
  File "~/clusterfuzz/src/python/platforms/fuchsia/device.py", line 228, in extend_fvm
    result = process.run_and_wait()
  File "~/clusterfuzz/src/python/system/new_process.py", line 344, in run_and_wait
    **popen_args)
  File "~/clusterfuzz/src/python/system/new_process.py", line 283, in run
    for k, v in env.items()
  File "~/clusterfuzz/src/python/system/new_process.py", line 283, in <dictcomp>
    for k, v in env.items()
  File "~/clusterfuzz/src/python/base/utils.py", line 949, in newstr_to_native_str
    return future_utils.native(str(s)).encode()
  File "~/clusterfuzz/ENV/local/lib/python2.7/site-packages/future/types/newstr.py", line 102, in __new__
    return super(newstr, cls).__new__(cls, value)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xf0 in position 6: ordinal not in range(128)
```

[...]

PR #1368, opened 2020-01-23
Fix encoding issue in newstr_to_native_str.

oliverchang:
Should address #1360.


oliverchang:
Oh, this misses a case. fixing.

PR #1371, opened 2020-01-23
Fix remaining unicode exceptions.

mbarbella-chromium:


src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -826,7 +826,7 @@ def execute_task(full_fuzzer_name, job_type):
     _save_coverage_information(context, result)
     _process_corpus_crashes(context, result)
   except CorpusPruningException as e:
-    logs.log_error('Corpus pruning failed: %s.' % str(e))
```
mbarbella-chromium: This shouldn't be necessary since we'll log the exception info regardless. The issue in both cases is that the exception information isn't necessarily valid unicode, so when we try to convert it to a newstr it can lead to an exception. This will probably need more thought as we try to get everything running properly in Python 3, but for now this seems like a quick exception fix.

src/python/bot/untrusted_runner/host.py
```diff
@@ -157,7 +158,12 @@ def wrapped(*args, **kwargs):
       except grpc.RpcError as e:
         # For timeouts, which aren't fatal errors, resurface the right
         # exception.
-        if 'TimeoutError' in str(e, encoding='utf-8', errors='ignore'):
```
mbarbella-chromium: I tested the original fix here Python 3, but not Python 2 with newstr. It actually doesn't work in 2, so replacing it with something like the fix above. This is also just intended as a stopgap and will not work properly in Python 3, but we can likely remove it after converting.

mbarbella-chromium:
TBR. Will land this after tests are passing. PTAL when you have a chance.

PR #1373, opened 2020-01-23
Fuchsia exception fix.

oliverchang:
Check for artifact_prefix value before copying to it.


flowerhack:
Works for me locally, #shipit, etc

PR #1378, opened 2020-01-24
Remove an incorrect usage of the Python 3-style str constructor.

mbarbella-chromium:
This doesn't work with newstr, so isn't actually compatible with both Python 2 and 3.

TBR. Will land this once tests pass.

### DIFF
--- a/src/python/bot/untrusted_runner/tasks_host.py
+++ b/src/python/bot/untrusted_runner/tasks_host.py
@@ -14,8 +14,6 @@
 """Tasks host."""
 from __future__ import absolute_import
 
-from builtins import str
-
 import datetime
 
 from google.protobuf import wrappers_pb2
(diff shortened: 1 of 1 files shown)


src/python/bot/untrusted_runner/tasks_host.py
```diff
@@ -186,7 +184,7 @@ def engine_reproduce(engine_impl, target_name, testcase_path, arguments,
   try:
     response = host.stub().EngineReproduce(request)
   except grpc.RpcError as e:
-    if 'TargetNotFoundError' in str(e, encoding='utf-8', errors='ignore'):
+    if 'TargetNotFoundError' in e.message.decode('utf-8', errors='ignore'):
```
inferno-chromium: Is this impacted - https://github.com/google/clusterfuzz/blob/28fc06d463b486986b3c2695b188d4263ac510cb/src/python/base/utils.py#L109 ?

mbarbella-chromium: Tested locally, it seems like that case works fine. I'm actually a bit more confused by this case now, since I'd suspect if one works both should. I'm going to keep an eye on the exceptions related to it (there was only 1 in the past 24h) to see if this was actually working and a bot just took an unusually long time to pick up the update.

mbarbella-chromium:
Abandoning this for now based on previous comment. Will re-open if the exception continues to occur.

PR #1384, opened 2020-01-28
Handle empty JSON files in fuzzer coverage task (follow up #1372).

Dor1s:
Apparently, Go version was not crashing on this (because of the error handling)
and would just skip such fuzz targets silently. We can do the same in Python, but
IMO logging a warning and recording at empty stats is better, as it clearly indicates
that ClusterFuzz side of the things is working, while fuzz target or coverage infra
is having issues.


Dor1s:
Actually my description is a bit wrong. Fixed it. TL;DR Go was also unhappy with empty JSONs, but it would silently skip them.

PR #1385, opened 2020-01-28
Handle malformed code coverage JSON files (follow up #1372).

Dor1s:
Apparently we may have not only empty JSONs, but malformed too!

I'll try to find one why that particular one is so weird, but CF code needs to survive it anyways.

```
>>> data[208600 : 209200]
'"filename":"/src/libtiff/libtiff/tif_next.c","summary":{"functions":{"count":3,"covered":0,"percent":0},"instantiations":{"count":3,"covered":0,"percent":0},"lines":{"count":124,"covered":0,"percent":0},"regions":{"count":59,"covered":0,"notcovered":59,"percent":0}}},{"filename":"/src/libtiff/lions":{"count":0,"covered":0,"percent":0},"instantiations":{"count":0,"covered":0,"percent":0},"lines":{"count":0,"covered":0,"percent":0},"regions":{"count":0,"covered":0,"notcovered":0,"percent":0}}},{"filename":"/src/libtiff/libtiff/uvcode.h","summary":{"functions":{"count":0,"covered":0,"percent":0},'
```

`{"filename":"/src/libtiff/lions":{"count":0` -- `, "summary"` is somehow missing in this piece, right after the filename value :O

### DIFF
--- a/src/appengine/handlers/cron/fuzzer_coverage.py
+++ b/src/appengine/handlers/cron/fuzzer_coverage.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 """Cron job to get the latest code coverage stats and HTML reports."""
 
+from builtins import str
 import datetime
 import json
 import os
@@ -40,10 +41,15 @@ def _basename(gcs_path):
 def _read_json(url):
   """Returns a JSON obejct loaded from the given GCS url."""
   data = storage.read_data(url)
-  if not data:
-    return None
 
-  return json.loads(data)
+  result = None
+  try:
+    result = json.loads(data)
+  except Exception as e:
+    logs.log_warn(
+        'Empty or malformed code coverage JSON (%s): %s.' % (url, str(e)))
+
+  return result
 
 
 def _coverage_information(summary_path, name, report_info):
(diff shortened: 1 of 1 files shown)


src/appengine/handlers/cron/fuzzer_coverage.py
```diff
@@ -40,10 +41,15 @@ def _basename(gcs_path):
...
   """Returns a JSON obejct loaded from the given GCS url."""
   data = storage.read_data(url)
-  if not data:
-    return None
 
-  return json.loads(data)
+  result = None
+  try:
+    result = json.loads(data)
+  except Exception as e:
```
mbarbella-chromium: Nit: Could this be ValueError instead of exception?

inferno-chromium: usually prefer specific exception - JSONDecodeError here ?

Dor1s: It can be `JSONDecodeError` or `TypeError` at least, maybe something else. Do you think it makes sense to enumerate all of them?

Dor1s: and @mbarbella-chromium is right, in Python2 `ValueError` is also a possibility. I frankly don't think it's reasonable to distinguish given that there is a single function call inside `try` block, but would love to have an agreement between us

inferno-chromium: ok sure, just generic is fine as-is.

PR #1390, opened 2020-01-29
Use repr() for serializing OSS-Fuzz exceptions.

oliverchang:
Avoid encoding with newstr, which also incorrectly ignores the encoding
and ignores keyword arguments when given an exception.


mbarbella-chromium:
LGTM, thanks!

PR #1392, opened 2020-01-30
Add some additional logging to debug a CI hang.

oliverchang:


oliverchang:
TBR

oliverchang:
Of course, the hang goes away when I add these prints. I'll just keep this here in case the issue comes back in the future.

PR #1395, opened 2020-01-30
Remove App Engine python paths from bots.

oliverchang:
Last bot dependency on App Engine (NDB) was removed, so this can all be removed now.;

The SDK zip will be removed from the Dockerfiles in a subsequent PR.

### DIFF
--- a/butler.py
+++ b/butler.py
@@ -32,8 +32,6 @@ sys.path.insert(0, os.path.join(os.path.dirname(__file__), 'src'))
 from local.butler import guard
 guard.check()
 
-from local.butler import appengine
-
 
 class _ArgumentParser(argparse.ArgumentParser):
   """Custom ArgumentParser."""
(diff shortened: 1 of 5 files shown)


butler.py
```diff
@@ -313,9 +311,6 @@ def main():
 
 def _setup():
   """Set up configs and import paths."""
-  appengine_sdk_path = appengine.find_sdk_path()
```
inferno-chromium: Does all other butler commands (other than py_unittest) work without this ?

src/local/butler/py_unittest.py
```diff
@@ -253,6 +257,15 @@ def execute(args):
       (appengine_config)  # pylint: disable=pointless-statement
     except ImportError:
       print('Note: unable to import appengine_config.')
+
+    # google.auth uses App Engine credentials based on importability of
```
inferno-chromium: Can we call modules. _patch_appengine_modules_for_bots ?

oliverchang: That will be removed very shortly once SDK is removed from Docker image/bot environments, so duplicating here.

butler.py
oliverchang: Yes I believe so, tested run_server/run_bot/run/format/lint/bootstrap/integration_tests/package. None of the other ones should be importing from any App Engine SDK paths.

PR #1396, opened 2020-01-30
Initialize NDB in chromium-tests-syncer.

oliverchang:


oliverchang:
/gcbrun. The other one (chromium-builder) doesn't seem like it touches Datastore so didn't add that there.

PR #1397, opened 2020-01-30
Fuzzer Stats: handle missing values for corpora stats (follow up #1372).

Dor1s:
Fixes https://console.cloud.google.com/errors/CL3_lZejgoXFfg?time=PT1H&project=clusterfuzz-external

This is happening when fuzzer coverage tasks runs before the corpus pruning for a corresponding fuzz target succeeded. Not sure how it worked previously, either:
- Go initialized those fields with 0 by default
- we just never ran into this

Anyways, I think explicit `--` being shown is better than a fake `0`, so this is a good change.

Tested on staging.

### DIFF
--- a/src/python/metrics/fuzzer_stats.py
+++ b/src/python/metrics/fuzzer_stats.py
@@ -500,6 +500,9 @@ class CorpusSizeField(BaseCoverageField):
       corpus_size_bytes = coverage_info.quarantine_size_bytes
       corpus_location = coverage_info.quarantine_location
 
+    if corpus_size_units is None or corpus_size_bytes is None:
+      return None
+
     display_value = '%d (%s)' % (corpus_size_units,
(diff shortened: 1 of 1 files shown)


src/python/metrics/fuzzer_stats.py
```diff
@@ -500,6 +500,9 @@ def get(self, group_by, group_by_value):
       corpus_size_bytes = coverage_info.quarantine_size_bytes
       corpus_location = coverage_info.quarantine_location
 
+    if corpus_size_units is None or corpus_size_bytes is None:
```
inferno-chromium: Maybe add a comment that you have in description.

Dor1s: done!

Dor1s:
hm, another timeout on CI, retrying

PR #1398, opened 2020-01-31
Prevent radamsa in process strategy if mutator plugin is already being used. 

mpherman2:


src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1704,7 +1704,9 @@ def pick_strategies(strategy_pool, fuzzer_path, corpus_directory,
     fuzzing_strategies.append(strategy.MUTATOR_PLUGIN_STRATEGY.name)
 
   if (strategy_pool.do_strategy(strategy.MUTATOR_PLUGIN_RADAMSA_STRATEGY) and
-      use_radamsa_mutator_plugin(extra_env)):
+      use_radamsa_mutator_plugin(extra_env)
+     ) and not fuzzing_strategies.contains(
```
mbarbella-chromium: Nit: keep this close paren before the colon.

mpherman2: Done

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1704,7 +1704,8 @@ def pick_strategies(strategy_pool, fuzzer_path, corpus_directory,
     fuzzing_strategies.append(strategy.MUTATOR_PLUGIN_STRATEGY.name)
 
   if (strategy_pool.do_strategy(strategy.MUTATOR_PLUGIN_RADAMSA_STRATEGY) and
-      use_radamsa_mutator_plugin(extra_env)):
+      use_radamsa_mutator_plugin(extra_env) and
+      not fuzzing_strategies.contains(strategy.MUTATOR_PLUGIN_STRATEGY.name)):
```
Dor1s: I think the order here is wrong. If line 1706 is true, then line 1707 gets executed and enables the plugin (as it sets the `LD_PRELOAD` and also logs the message), but then we can end with the false on line 1708, i.e. the strategy will be used but the stats will say it isn't.

Swap these two lines, and maybe add a comment on line 1706 explaining why one strategy needs the other one to be disabled

mpherman2: This is a good catch! Thank you.

mpherman2: Done.

PR #1400, opened 2020-01-31
Use sanitizer options symbolize=1 when dataflow tracing is enabled (#1346).

Dor1s:
This is needed to fix https://github.com/google/oss-fuzz/issues/3311.

I'll also make a change in libFuzzer to break loudly when the symbolizer is not available.

### DIFF
--- a/src/python/bot/fuzzers/libfuzzer.py
+++ b/src/python/bot/fuzzers/libfuzzer.py
@@ -1534,10 +1534,6 @@ def set_sanitizer_options(fuzzer_path, fuzz_options=None):
     # Focus function feature does not work without symbolization.
     sanitizer_options['symbolize'] = 1
     environment.update_symbolizer_options(sanitizer_options)
-
-    # But we still need to disable symbolization inlined frames to make sure
-    # deduplication works correctly.
-    sanitizer_options['symbolize_inline_frames'] = 'false'
   environment.set_memory_tool_options(sanitizer_options_var, sanitizer_options)
 
 
--- a/src/python/bot/fuzzers/libfuzzer.py
+++ b/src/python/bot/fuzzers/libfuzzer.py
@@ -1522,7 +1522,7 @@ def parse_log_stats(log_lines):
   return log_stats
 
 
-def set_sanitizer_options(fuzzer_path):
+def set_sanitizer_options(fuzzer_path, fuzz_options=None):
   """Sets sanitizer options based on .options file overrides and what this
   script requires."""
   engine_common.process_sanitizer_options_overrides(fuzzer_path)
@@ -1530,6 +1530,9 @@ def set_sanitizer_options(fuzzer_path):
   sanitizer_options = environment.get_memory_tool_options(
       sanitizer_options_var, {})
   sanitizer_options['exitcode'] = constants.TARGET_ERROR_EXITCODE
+  if fuzz_options and fuzz_options.use_dataflow_tracing:
+    # Focus function feature does not work without symbolization.
+    sanitizer_options['symbolize'] = 1
   environment.set_memory_tool_options(sanitizer_options_var, sanitizer_options)
 
 
--- a/src/python/bot/fuzzers/libfuzzer.py
+++ b/src/python/bot/fuzzers/libfuzzer.py
@@ -1523,8 +1523,8 @@ def parse_log_stats(log_lines):
 
 
 def set_sanitizer_options(fuzzer_path, fuzz_options=None):
-  """Sets sanitizer options based on .options file overrides and what this
-  script requires."""
+  """Sets sanitizer options based on .options file overrides, FuzzOptions (if
+  provided), and what this script requires."""
   engine_common.process_sanitizer_options_overrides(fuzzer_path)
   sanitizer_options_var = environment.get_current_memory_tool_var()
   sanitizer_options = environment.get_memory_tool_options(
--- a/src/python/bot/fuzzers/libfuzzer.py
+++ b/src/python/bot/fuzzers/libfuzzer.py
@@ -1533,6 +1533,10 @@ def set_sanitizer_options(fuzzer_path, fuzz_options=None):
   if fuzz_options and fuzz_options.use_dataflow_tracing:
     # Focus function feature does not work without symbolization.
     sanitizer_options['symbolize'] = 1
+
+    # But we still need to disable symbolization inlined frames to make sure
+    # deduplication works correctly.
+    sanitizer_options['symbolize_inline_frames'] = 'false'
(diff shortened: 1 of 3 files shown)


Dor1s:
> just adding symbolize=1 might not work, since we add external_symbolizer_path in environment.py as well

llvm-symbolizer is already available inside the minijail:https://github.com/google/clusterfuzz/blob/18ff074ec2b631eaf707758712b64110f487cf56/src/python/bot/fuzzers/libfuzzer.py#L1324

it should be in the `$PATH` passed to the minijail

> adding symbolize=1 can cause other problems, which is crashes will have synbolized frames, so enabling/disabling inline frames will be painful.

I think Chrome OS jobs are running with `symbolize=1`, as well as `libfuzzer_chrome_asan` and `libfuzzer_chrome_asan_debug` jobs, were there any issues with the stackframes? //cc @jonathanmetzman

inferno-chromium:
> > just adding symbolize=1 might not work, since we add external_symbolizer_path in environment.py as well
> 
> llvm-symbolizer is already available inside the minijail:
> 
> https://github.com/google/clusterfuzz/blob/18ff074ec2b631eaf707758712b64110f487cf56/src/python/bot/fuzzers/libfuzzer.py#L1324
> 
> it should be in the `$PATH` passed to the minijail

What about the non-minijail case ?

> 
> > adding symbolize=1 can cause other problems, which is crashes will have synbolized frames, so enabling/disabling inline frames will be painful.
> 
> I think Chrome OS jobs are running with `symbolize=1`, as well as `libfuzzer_chrome_asan` and `libfuzzer_chrome_asan_debug` jobs, were there any issues with the stackframes? //cc @jonathanmetzman

We do in those since i remember we had to use suppressions there and that wont work without symbolize=1. symbolize=1 basically causes symbolization to always happen and then we cannot skip inline frames (hurts deduplication and even cause duplicate testcases).

Dor1s:
> What about the non-minijail case ?

It's available there as wekk, e.g. logs from `libfuzzer_chrome_asan_debug` are all symbolized:

```
libfuzzer_chrome_asan_debug/2020-01-27/23:47:05:335375.log:  NEW_FUNC: 0x55f077e10da0 in webrtc::FloatS16ToS16(float) third_party/webrtc/common_audio/include/audio_util.h:40
libfuzzer_chrome_asan_debug/2020-01-27/23:47:05:335375.log:  NEW_FUNC: 0x55f077e90b50 in WebRtcSpl_AllPassQMF third_party/webrtc/common_audio/signal_processing/splitting_filter.c:50
```

> We do in those since i remember we had to use suppressions there and that wont work without symbolize=1. symbolize=1 basically causes symbolization to always happen and then we cannot skip inline frames (hurts deduplication and even cause duplicate testcases).

Now I'm getting confused. It seems that we use `symbolize_inline_frames=false` while fuzzing anyway (i.e. inline frames aren't symbolized even with `symbolize=1`), but we do enable it in the symbolize task: https://github.com/google/clusterfuzz/search?q=symbolize_inline_frames&unscoped_q=symbolize_inline_frames

which makes it looks like we actually want to see inlined stackframes for deduplication?

Dor1s:
Added `symbolize_inline_frames=false` by re-using some code from `environment.reset_current_memory_tool_options` per @inferno-chromium's suggestion

and also filed https://github.com/google/sanitizers/issues/1190 to discuss if this internal symbolization logic can be improved in the compiler-rt

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1522,14 +1522,22 @@ def parse_log_stats(log_lines):
...
       sanitizer_options_var, {})
   sanitizer_options['exitcode'] = constants.TARGET_ERROR_EXITCODE
+  if fuzz_options and fuzz_options.use_dataflow_tracing:
+    # Focus function feature does not work without symbolization.
+    sanitizer_options['symbolize'] = 1
+    environment.update_symbolizer_options(sanitizer_options)
+
+    # But we still need to disable symbolization inlined frames to make sure
+    # deduplication works correctly.
+    sanitizer_options['symbolize_inline_frames'] = 'false'
```
inferno-chromium: Why is this needed now, can be removed.

src/python/system/environment.py
```diff
@@ -768,6 +768,19 @@ def set_environment_parameters_from_file(file_path):
       set_value(environment_variable, environment_variable_value)
 
 
+def update_symbolizer_options(tool_options, symbolize_inline_frames=False):
+  """Checks and updates the necessary symbolizer options such as
+  # `external_symbolizer_path` and `symbolize_inline_frames`."""
```
inferno-chromium: Remove the #

src/python/bot/fuzzers/libfuzzer.py
Dor1s: yeah, sorry, was removing that as you were reviewing

src/python/system/environment.py
Dor1s: done

src/python/system/environment.py
```diff
@@ -768,6 +768,19 @@ def set_environment_parameters_from_file(file_path):
       set_value(environment_variable, environment_variable_value)
 
 
+def update_symbolizer_options(tool_options, symbolize_inline_frames=False):
+  """Checks and updates the necessary symbolizer options such as
+  `external_symbolizer_path` and `symbolize_inline_frames`."""
+  if 'external_symbolizer_path' not in tool_options:
```
Dor1s: looks like this was a mistake, now this set up a wrong path for the symbolizer:

```
==1==WARNING: invalid path to external symbolizer!
==1==WARNING: Failed to use and restart external symbolizer!
```

Dor1s: should be addressed with https://github.com/google/clusterfuzz/pull/1401, will check the logs

oliverchang:
From memory `symbolize_inline_frames=false` didn't work as expected for us to ignore inline frames.. Can you please confirm if this flag actually works for our purposes (de-duplication)?

Dor1s:
> From memory `symbolize_inline_frames=false` didn't work as expected for us to ignore inline frames.. Can you please confirm if this flag actually works for our purposes (de-duplication)?

There is a test in compiler-rt that verified that inlined frames are skipped when this flag is false (https://github.com/llvm/llvm-project/blob/6ca1707b2372db13f5d6ca58b237d7f551275d2a/compiler-rt/test/sanitizer_common/TestCases/symbolize_pc_inline.cpp)

And I've just confirmed that it indeed works that way using https://oss-fuzz.com/testcase-detail/5711949293944832

As far as I understand, it shouldn't hurt deduplication, as we don't use inline frames for it. Is that right?

PR #1401, opened 2020-01-31
Do not set None to external_symbolizer_path env variable (#1346).

Dor1s:
Quick follow up.


src/python/system/environment.py
```diff
@@ -774,7 +774,10 @@ def update_symbolizer_options(tool_options, symbolize_inline_frames=False):
   if 'external_symbolizer_path' not in tool_options:
     llvm_symbolizer_path_arg = _quote_value_if_needed(
```
inferno-chromium: i think _quote_value_if_needed should be moved inside the if.
and then s/llvm_symbolizer_path_arg/llvm_symbolizer_path
we shouldn't be quoting None as then we wont fail on the if.

Dor1s: Good call, fixing!

PR #1404, opened 2020-02-03
Preliminary environment setup for Python 3. 

oliverchang:
Add Pipfile/Pipfile.lock for Python 3 dependencies. This is currently
just based on src/local/requirements.txt and docker/ci/requirements.txt, and will be cleaned up later.

Add python 3 env setup to install_deps_linux.bash. To use, set `PY3=1` before running:

```python
PY3=1 ./local/install_deps.bash
```

A separate checkout should be used from the one used for Python 2
development.

Add some minor bug fixes to get butler and core unit tests running. The vast
majority of them are still failing though.

Vendor webapp2 and Jinja2 in src/third_party, since we're no longer depending on the ones available in the App Engine SDK.

Dependent on https://github.com/google/clusterfuzz/pull/1395

### DIFF
--- a/src/local/butler/guard.py
+++ b/src/local/butler/guard.py
@@ -20,6 +20,7 @@ import sys
 def check_virtualenv():
   """Check that we're in a virtualenv."""
   if sys.version_info.major != 2:
+    # TODO(ochang): Implement check for Python 3.
     return
 
   root_path = os.path.realpath(
@@ -36,7 +37,9 @@ def check_virtualenv():
 def check_dev_requirements():
   """Check that dev requirements are installed."""
   if sys.version_info.major != 2:
+    # TODO(ochang): Implement check for Python 3.
     return
+
   freeze_output = subprocess.check_output('pip freeze', shell=True)
   installed_pips = freeze_output.strip().splitlines()
 
--- a/src/local/butler/guard.py
+++ b/src/local/butler/guard.py
@@ -14,9 +14,14 @@
 """guard.py checks virtualenv environment and dev requirements."""
 import os
 import subprocess
+import sys
 
 
 def check_virtualenv():
+  """Check that we're in a virtualenv."""
+  if sys.version_info.major != 2:
+    return
+
   root_path = os.path.realpath(
       os.path.join(os.path.dirname(__file__), '..', '..', '..'))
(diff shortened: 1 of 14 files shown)


oliverchang:
PTAL. Patch is dependent on https://github.com/google/clusterfuzz/pull/1395

src/local/butler/guard.py
```diff
@@ -14,9 +14,14 @@
 """guard.py checks virtualenv environment and dev requirements."""
 import os
 import subprocess
+import sys
 
 
 def check_virtualenv():
+  """Check that we're in a virtualenv."""
+  if sys.version_info.major != 2:
```
mbarbella-chromium: Should these have associated TODOs?

src/local/butler/py_unittest.py
```diff
@@ -241,31 +241,32 @@ def execute(args):
...
-
-    # google.auth uses App Engine credentials based on importability of
-    # google.appengine.api.app_identity.
-    try:
-      from google.auth import app_engine as auth_app_engine
-      if auth_app_engine.app_identity:
-        auth_app_engine.app_identity = None
-    except ImportError:
-      pass
+    if sys.version_info.major == 2:
```
mbarbella-chromium: And a TODO here as well?

src/local/butler/guard.py
inferno-chromium: Myabe add a comment why

src/local/butler/guard.py
```diff
@@ -30,6 +35,8 @@ def check_virtualenv():
 
 def check_dev_requirements():
   """Check that dev requirements are installed."""
+  if sys.version_info.major != 2:
```
inferno-chromium: Myabe add a comment why

src/local/butler/py_unittest.py
oliverchang: Done.

src/local/butler/guard.py
oliverchang: Done.

src/local/butler/guard.py
oliverchang: Added a TODO.

oliverchang:
Just realised I messed up the merge here (merged into the wrong branch). Fixing.

PR #1405, opened 2020-02-03
Add back some workarounds for NDB projection bug.

oliverchang:
The fix in 1.0.0 was not complete. Re-filed
https://github.com/googleapis/python-ndb/issues/319


oliverchang:
TBR

oliverchang:
> LGTM, so other similar places are not impacted ?

Right, other .query() etc uses are not affected.

PR #1406, opened 2020-02-03
Fix OSS-Fuzz bots page.

oliverchang:
batch_size argument is no longer supported.


oliverchang:
TBR for trivial fix. /gcbrun

PR #1409, opened 2020-02-04
Remove App Engine SDK from bot Docker images.

oliverchang:
The CI still has the necessary utilities for dev_appserver etc from APT.


oliverchang:
Good catch re first one! Fixed.

Second one is a bit more scary -- will remove that one in a later PR.

PR #1411, opened 2020-02-04
Remove an import hack from the base Dockerfile.

oliverchang:
No longer needed as App Engine SDK is removed.


oliverchang:
(forgot to add an reviewer). TBR for simple removal.

PR #1412, opened 2020-02-04
Remove App Engine SDK from init scripts.

oliverchang:


oliverchang:
> Probably need to remove #1411 from android-test.bash

Done!

PR #1413, opened 2020-02-04
Add support for python stacktraces

Google-Autofuzz:


src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -293,6 +293,19 @@
 GOLANG_STACK_FRAME_FUNCTION_REGEX = re.compile(
     r'^([0-9a-zA-Z\.\-\_\\\/\(\)\*]+)\([x0-9a-f\s,\.]*\)$')
 
+
+# Python specific regular expressions.
+PYTHON_UNHANDLED_EXCEPTION = re.compile(r'^ === Uncaught Python exception: ===$')
+
+PYTHON_CRASH_TYPES_MAP = [
+    (PYTHON_UNHANDLED_EXCEPTION, 'Uncaught exception.'),
```
inferno-chromium: no dot at end.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -293,6 +293,19 @@
 GOLANG_STACK_FRAME_FUNCTION_REGEX = re.compile(
     r'^([0-9a-zA-Z\.\-\_\\\/\(\)\*]+)\([x0-9a-f\s,\.]*\)$')
 
+
+# Python specific regular expressions.
+PYTHON_UNHANDLED_EXCEPTION = re.compile(r'^ === Uncaught Python exception: ===$')
```
inferno-chromium: ^\s*===

Seems slightly better, will there be always a one character space at start.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -1139,6 +1152,8 @@ def get_crash_data(crash_data, symbolize_flag=True):
 
   is_kasan = 'KASAN' in crash_stacktrace_without_inlines
   is_golang = '.go:' in crash_stacktrace_without_inlines
+  is_python = '.py", line' in crash_stacktrace_without_inlines
```
inferno-chromium: Not that "Uncaught Python exception" ?

src/python/crash_analysis/stack_parsing/stack_analyzer.py
Google-Autofuzz: Likely, yes

src/python/crash_analysis/stack_parsing/stack_analyzer.py
Google-Autofuzz: "Uncaught python exception" might not be the only thing that we want to catch, so I aimed at the most generic way to detect python crashes.

inferno-chromium:
python butler.py format
does it for you.

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -2809,3 +2809,16 @@ def test_libfuzzer_overwrites_const_input(self):
     self._validate_get_crash_data(data, expected_type, expected_address,
                                   expected_state, expected_stacktrace,
                                   expected_security_flag)
+
+  def test_python_unhandled_exception(self):
+    """Test python stacktrace with an unhandled exception."""
+    data = self._read_test_data('python_unhandled_exception.txt')
+    expected_type = 'Uncaught exception.'
```
inferno-chromium: Remove the dot from here.

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -1089,7 +1089,7 @@ def test_cfi_unrelated_vtable(self):
     """Test the CFI output format from an unrelated vtable."""
     data = self._read_test_data('cfi_unrelated_vtable.txt')
     expected_type = 'Bad-cast'
-    expected_address = '0x000000422710'
+    expected_address = '0x000000422710', PTAL!
```
oliverchang: Unintentional change?

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -2809,3 +2809,16 @@ def test_libfuzzer_overwrites_const_input(self):
     self._validate_get_crash_data(data, expected_type, expected_address,
                                   expected_state, expected_stacktrace,
                                   expected_security_flag)
+
+  def test_python_unhandled_exception(self):
+    """Test python stacktrace with an unhandled exception."""
+    data = self._read_test_data('python_unhandled_exception.txt')
+    expected_type = 'Uncaught exception'
+    expected_address = ''
+    expected_state = 'Wrapper\nTestOneInput\nopen\n'
```
oliverchang: The ordering is wrong here. For best de-duplication results, we want to take the "top" 3 frames, where "top" are the frames closest to the crash. Currently these are the bottom because of the order that Python prints them.

We should do a pre-processing of sorts on the stacktrace to reverse the frames.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -1139,11 +1151,29 @@ def get_crash_data(crash_data, symbolize_flag=True):
...
+    # Python stacktraces are reversed, and followed by an ASAN one
+    python_crash_stacktrace = []
+    for line in split_crash_stacktrace:
+      if not line:
+        continue
+      if '===========' in line:  # begining of ASAN stacktrace, skip this
+        break
+      python_crash_stacktrace.append(line)
+    # Add the "title" of the stacktrace at the top
+    split_crash_stacktrace = [python_crash_stacktrace[0]]
```
inferno-chromium: This is slightly hacky. We can have the stdout from the program in the first line (or even first few lines). I think you should add a few stdout lines in output at start like "Line1\nLine2\n" and then in the for loop, note the index for "Uncaught Python exception" marker and then only reverse lines after that.

oliverchang: +1

PR #1419, opened 2020-02-06
Fix some unicode decode issues with newstr and exceptions.

oliverchang:


oliverchang:
TBR. /gcbrun

oliverchang:
Created https://github.com/google/clusterfuzz/issues/1420 for that.

inferno-chromium:
Do these cover all the str(e) we append in errors ?

oliverchang:
> Do these cover all the str(e) we append in errors ?

These should be the majority, since they're coming from process output where this is more likely.

PR #1421, opened 2020-02-06
Initial changes for running the server using Python 3.

oliverchang:
- Add a run_server3 command for running the server using Python 3.
- Make staging deployment work for python 3 (`--staging3`).
- Use gunicorn for local server. dev_appserver no longer works well for Python 3 apps.
- Fix an encoding issue in the json encoder. 
- Add appengine/main.py as the entrypoint for GAE Python 3.

The testcases list page loads with this!


oliverchang:
Had some fun figuring out why I was seeing this on staging: 

```
status = StatusCode.UNAVAILABLE details = "DNS resolution failed"
```

After turning on some gRPC debugging logs...

```
..."target_address":"b'datastore.googleapis.com':443"}
```

Had to stare at this many times before I realised the "**b'** datastore.googleapis.com':443"

inferno-chromium:
@Dor1s - for first round review.

src/local/butler/package.py
```diff
@@ -89,7 +89,7 @@ def package(revision,
   appengine.symlink_dirs()
 
   _, ls_files_output = common.execute('git -C . ls-files', print_output=False)
-  file_paths = ls_files_output.splitlines()
+  file_paths = [path.decode('utf-8') for path in ls_files_output.splitlines()]
```
inferno-chromium: Question: Should you decode the output first, then splitlines ?

inferno-chromium:
Didn't see @mbarbella-chromium was on review, so removing @Dor1s.

src/local/butler/package.py
oliverchang: No real difference here, so leaving as is.

PR #1422, opened 2020-02-06
Upgrade waitress to 1.4.3

inferno-chromium:


inferno-chromium:
Does https://github.com/google/clusterfuzz/pull/1416 properly since that was modifying Pipfile.lock outside of just waitress changes.

inferno-chromium:
TBR

PR #1425, opened 2020-02-07
Python 3 server fixes.

oliverchang:
- Fix JSON encoding bytes for request handlers.
- Fix None and int comparison in big_query.py
- vars() no longer works on namedtuples. Use ._asdict() instead.


oliverchang:
PTAL. this makes testcase details and testcase upload pages work. every other page also seems to work from a cursory check.

PR #1426, opened 2020-02-07
Replace self.assertItemsEqual with six.assertCountEqual.

oliverchang:
Python 3 renamed assertItemsEqual to the very misleadingly named
assertCountEqual and broke compatiblity. Use six.assertCountEqual.

https://six.readthedocs.io/#six.assertCountEqual


oliverchang:
TBR since this is a test only change with no functional differences.

PR #1428, opened 2020-02-07
Remove parse_stacktrace handler.

oliverchang:
It was used by the old reproduce tool, and does not appear to be used
anymore.


oliverchang:
@mbarbella-chromium is my assumption here that the old reproduce tool was the only user of this correct? 

I do see that the endpoint was hit a very small number of times in the last week (all within the same minute) that I assume is from a user attempting to use the old reproduce tool. Does the old reproduce tool still run? If so, should we point it to the new reproduce tool?

inferno-chromium:
@mbarbella-chromium - i have make clusterfuzz-tools private now to avoid being used. can you remove the reproduce tool binaries checked in google (/move to deprecated folder).

PR #1431, opened 2020-02-07
Add a BigQuery table for cross-pollination stats

mpherman2:
I was not able to test this locally yet because run_server is down, but I would appreciate it if you took a look, and I will test it before we merge.

### DIFF
--- a/configs/test/bigquery/datasets.yaml
+++ b/configs/test/bigquery/datasets.yaml
@@ -225,7 +225,7 @@ resources:
             type: FLOAT
           - name: engine
             type: STRING
-  - name: cross-pollination-statistics-table
+  - name: cross-pollination-stats-table
     type: gcp-types/bigquery-v2:tables
     properties:
       datasetId: $(ref.main-dataset.datasetReference.datasetId)
@@ -239,15 +239,15 @@ resources:
           type: STRING
         - name: tagged
           type: BOOLEAN
-        - name: original_corpus_size
+        - name: initial_corpus_size
           type: INTEGER
-        - name: post_pollination_corpus_size
+        - name: corpus_size
           type: INTEGER
-        - name: original_edge_coverage
+        - name: initial_edge_coverage
           type: INTEGER
-        - name: post_pollination_edge_coverage
+        - name: edge_coverage
           type: INTEGER
-        - name: original_feature_coverage
+        - name: initial_feature_coverage
           type: INTEGER
-        - name: post_pollination_feature_coverage
+        - name: feature_coverage
           type: INTEGER
--- a/configs/test/bigquery/datasets.yaml
+++ b/configs/test/bigquery/datasets.yaml
@@ -225,3 +225,29 @@ resources:
             type: FLOAT
           - name: engine
             type: STRING
+  - name: cross-pollination-statistics-table
+    type: gcp-types/bigquery-v2:tables
+    properties:
+      datasetId: $(ref.main-dataset.datasetReference.datasetId)
+      tableReference:
+        tableId: cross-pollination-statistics
+      timePartitioning:
+        type: DAY
+      schema:
+        fields:
+        - name: project_qualified_name
+          type: STRING
+        - name: tagged
+          type: BOOLEAN
+        - name: original_corpus_size
+          type: INTEGER
+        - name: post_pollination_corpus_size
+          type: INTEGER
+        - name: original_edge_coverage
+          type: INTEGER
+        - name: post_pollination_edge_coverage
+          type: INTEGER
+        - name: original_feature_coverage
+          type: INTEGER
+        - name: post_pollination_feature_coverage
+          type: INTEGER
--- a/configs/test/bigquery/datasets.yaml
+++ b/configs/test/bigquery/datasets.yaml
@@ -239,6 +239,10 @@ resources:
           type: STRING
         - name: method
           type: STRING
+        - name: sources
+          type: STRING
+        - name: tags
+          type: STRING
         - name: initial_corpus_size
           type: INTEGER
         - name: corpus_size
--- a/configs/test/bigquery/datasets.yaml
+++ b/configs/test/bigquery/datasets.yaml
@@ -237,8 +237,8 @@ resources:
         fields:
         - name: project_qualified_name
           type: STRING
-        - name: tagged
-          type: BOOLEAN
+        - name: method
+          type: STRING
         - name: initial_corpus_size
           type: INTEGER
         - name: corpus_size
--- a/src/python/bot/tasks/corpus_pruning_task.py
+++ b/src/python/bot/tasks/corpus_pruning_task.py
@@ -646,10 +646,10 @@ def do_corpus_pruning(context, last_execution_failed, revision):
         'tagged': False,
         'original_corpus_size': pre_pollination_corpus_size,
         'post_pollination_corpus_size': minimized_corpus_size_units,
-        'original_edge_coverage': pruner_stats.edge_coverage,
-        'post_pollination_edge_coverage': pollinator_stats.edge_coverage,
-        'original_feature_coverage': pruner_stats.feature_coverage,
-        'post_pollination_feature_coverage': pollinator_stats.feature_coverage
+        'original_edge_coverage': pruner_stats['edge_coverage'],
+        'post_pollination_edge_coverage': pollinator_stats['edge_coverage'],
+        'original_feature_coverage': pruner_stats['feature_coverage'],
+        'post_pollination_feature_coverage': pollinator_stats['feature_coverage']
     }
 
     client = big_query.Client(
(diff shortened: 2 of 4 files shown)


mpherman2:
I am failing some unit tests. It is a bit hard to iterate on this due to the fact that unit tests are not working locally right now. I will figure out the issue once unit tests are back up.

configs/test/bigquery/datasets.yaml
```diff
@@ -225,3 +225,29 @@ resources:
             type: FLOAT
           - name: engine
             type: STRING
+  - name: cross-pollination-statistics-table
```
Dor1s: super nit: I think we use `stats` a lot, so `statistics` can be shortified to it

configs/test/bigquery/datasets.yaml
```diff
@@ -225,3 +225,29 @@ resources:
...
+      datasetId: $(ref.main-dataset.datasetReference.datasetId)
+      tableReference:
+        tableId: cross-pollination-statistics
+      timePartitioning:
+        type: DAY
+      schema:
+        fields:
+        - name: project_qualified_name
+          type: STRING
+        - name: tagged
```
Dor1s: would it make sense to record the tags as well? I guess they can change over time, so having explicit tags recorded might be useful knowledge

configs/test/bigquery/datasets.yaml
```diff
@@ -225,3 +225,29 @@ resources:
...
+    type: gcp-types/bigquery-v2:tables
+    properties:
+      datasetId: $(ref.main-dataset.datasetReference.datasetId)
+      tableReference:
+        tableId: cross-pollination-statistics
+      timePartitioning:
+        type: DAY
+      schema:
+        fields:
+        - name: project_qualified_name
```
Dor1s: how do we track what corpus was used for pollination? this field is the name of the target which we attempted to add more corpus to, i.e. it's "destination". I think tracking "sources" is also important. For example, after recording this for a while, we could try auto-deriving the tags based on what fuzz targets have helped each other through global cross polination.

configs/test/bigquery/datasets.yaml
```diff
@@ -225,3 +225,29 @@ resources:
...
+          type: STRING
+        - name: tagged
+          type: BOOLEAN
+        - name: original_corpus_size
```
Dor1s: sorry for the nit picking, I'd propose the following naming convention:
- `initial_corpus_size` for the "original" values
- `corpus_size` for the post pollination values

and the same for the coverage metrics. That's more similar to what we use in the existing stats table.

configs/test/bigquery/datasets.yaml
mpherman2: The problem with this is that multiple sources are used for pollination not just one. Do we want a field in the table that is a list of all soruce corpora?

configs/test/bigquery/datasets.yaml
mpherman2: Same as with sources, there are multiple tags for each target. Should every tag be listed?

configs/test/bigquery/datasets.yaml
```diff
@@ -225,3 +225,29 @@ resources:
...
+      datasetId: $(ref.main-dataset.datasetReference.datasetId)
+      tableReference:
+        tableId: cross-pollination-statistics
+      timePartitioning:
+        type: DAY
+      schema:
+        fields:
+        - name: project_qualified_name
+          type: STRING
+        - name: tagged
```
mbarbella-chromium: I guess the purpose here will eventually be to differentiate between the new implementation using explicit tags and normal cross pollination? Could this be something like a "method" string instead of a boolean to make it more resilient to whatever we might do in the future?

mbarbella-chromium: This may not matter if you decide to include the list of tags. Up to you how you handle this, but either way I'd like to move away from the boolean.

src/python/tests/core/bot/tasks/corpus_pruning_task_test.py
```diff
@@ -157,11 +157,23 @@ def setUp(self):
...
         'build_management.build_manager.setup_build',
+        'base.utils.get_application_id',
+        'google_cloud_utils.big_query.get_api_client',
     ])
     self.mock.setup_build.side_effect = self._mock_setup_build
+    self.mock.get_application_id.return_value = 'project'
 
   def test_prune(self):
     """Basic pruning test."""
+    underlying = mock.MagicMock()
```
mbarbella-chromium: Are these required at all? I thought return_value would default to a MagicMock, but didn't confirm.

configs/test/bigquery/datasets.yaml
mpherman2: Done.

src/python/tests/core/bot/tasks/corpus_pruning_task_test.py
mpherman2: It turns out not all of it was necessary, but I could not get the tests to pass without the code I left.

configs/test/bigquery/datasets.yaml
Dor1s: Yes, I think it's fine to have a string where we'll have multiple names listed. There is substring functionality in BQ language, should be fine if we need to query those.

mpherman2: Added. Right now the logic isn't there to populate the tags field. That will come in the next CL.

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -594,6 +600,9 @@ def do_corpus_pruning(context, last_execution_failed, revision):
   quarantine_corpus_dir_size = shell.get_directory_size(
       context.quarantine_corpus_path)
 
+  # Save the minimize corpus size before cross pollination to put in Big Query.
```
Dor1s: minor nit: `BigQuery` with no space

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -630,6 +639,33 @@ def do_corpus_pruning(context, last_execution_failed, revision):
...
 
+  sources = ','.join([
+      fuzzer.fuzz_target.project_qualified_name()
+      for fuzzer in context.cross_pollinate_fuzzers
+  ])
+
+  # TODO(mpherman): Add way to get tags string when tagged corpora are done.
+  tags = ''
+
+  if pruner_stats and pollinator_stats:
```
mbarbella-chromium: It might be a bit cleaner to factor the stats storage into a separate helper function, which will also let you simplify what you're mocking during testing (e.g. you won't need to understand that we're using the API client and mock out the relevant functions there, but can instead just mock out the new helper).

src/python/bot/tasks/corpus_pruning_task.py
mpherman2: Done

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -516,6 +519,34 @@ def run(self, timeout):
...
           'Corpus pruning failed to merge shared corpus\n' + repr(e))
 
+    return result.stats
+
+
+def record_cross_pollination_stats(
+    pruner_stats, pollinator_stats, project_qualified_name, sources, tags,
+    initial_corpus_size, minimized_corpus_size_units):
+  """Log stats about cross pollination in BigQuery."""
+  if pruner_stats and pollinator_stats and not environment.get_value(
```
mbarbella-chromium: Nit: rather than having the whole function here indented, return in the opposite case.

I also generally find it easier to reason about positive rather than negative variable names (ENABLE vs DISABLE) but don't think it will matter too much in this case once the condition is flipped.

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -516,6 +519,40 @@ def run(self, timeout):
...
 
+    return result.stats
+
+
+def record_cross_pollination_stats(
+    pruner_stats, pollinator_stats, project_qualified_name, sources, tags,
+    initial_corpus_size, minimized_corpus_size_units):
+  """Log stats about cross pollination in BigQuery."""
+  # BigQuery not available in local development.
+
```
mbarbella-chromium: Nit: remove this empty line.

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -516,6 +519,40 @@ def run(self, timeout):
...
           'Corpus pruning failed to merge shared corpus\n' + repr(e))
 
+    return result.stats
+
+
+def record_cross_pollination_stats(
+    pruner_stats, pollinator_stats, project_qualified_name, sources, tags,
+    initial_corpus_size, minimized_corpus_size_units):
+  """Log stats about cross pollination in BigQuery."""
+  # BigQuery not available in local development.
```
mbarbella-chromium: Nit: Could you give a slightly more detailed comment here? Without context it would look odd that we're checking if we're in a test here, so it's worth explaining that this is necessary because the untrusted runner is in a separate process and can't be easily mocked.

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -516,6 +519,40 @@ def run(self, timeout):
...
+
+  if environment.get_value("LOCAL_DEVELOPMENT") or environment.get_value(
```
mbarbella-chromium: Nit: single quotes.

PR #1433, opened 2020-02-10
Fix for Datastore emulator did not get ready in time in py_unittest

mpherman2:
Noticed the same problem happening in py_unittest that was happening in run_server. Looked at @oliverchangs  PR and applied the same fix.


oliverchang:
Thanks! /gcbrun

PR #1434, opened 2020-02-11
Fix all App Engine tests running under Python 3.

oliverchang:
- Make multithreaded unit tests work.
- Fix performance analyzer tests to pass unicode (str) stacktraces. This
  matches behaviour in production.
- Fix one last str/bytes issue in manage_vms_test.

### DIFF
--- a/src/python/bot/fuzzers/libfuzzer.py
+++ b/src/python/bot/fuzzers/libfuzzer.py
@@ -1498,7 +1498,7 @@ def parse_log_stats(log_lines):
   log_stats = {}
 
   # Parse libFuzzer generated stats (`-print_final_stats=1`).
-  stats_regex = re.compile(br'stat::([A-Za-z_]+):\s*([^\s]+)')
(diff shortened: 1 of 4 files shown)


src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1498,7 +1498,7 @@ def parse_log_stats(log_lines):
   log_stats = {}
 
   # Parse libFuzzer generated stats (`-print_final_stats=1`).
-  stats_regex = re.compile(br'stat::([A-Za-z_]+):\s*([^\s]+)')
+  stats_regex = re.compile(r'stat::([A-Za-z_]+):\s*([^\s]+)')
```
inferno-chromium: Is this intended to switch back ?

oliverchang: Yes, it was wrong. The stacktraces are always passed in as unicode strs, not bytes.

PR #1437, opened 2020-02-12
Integrate syzkaller fuzzing engine into clusterfuzz

hzawawy:
This change adds a new engine to run syzkaller from clusterfuzz. 
- new platform ANDROID_KERNEL
- new fuzzer/engine to run syzkaller
Another two CLs will follow to support crashes parsing/synching corpus and linkification of crashes.


hzawawy:
@googlebot I signed it!

src/appengine/handlers/cron/project_setup.py
```diff
@@ -127,6 +127,9 @@ def job_name(self, project_name):
     'address', ['honggfuzz', 'engine_asan'],
     minimize_job_override=LIBFUZZER_ASAN_JOB)
 
+SYZKALLER_KASAN_JOB = JobInfo('syzkaller_kasan_', 'syzkaller', 'address',
+                              ['syzkaller', 'engine_kasan', 'prune'])
```
inferno-chromium: Skip changing project_setup.py, this file is used for automated job setup. right now, we will manually define job and once it is stable, then move it here.

src/appengine/handlers/cron/project_setup.py
```diff
@@ -907,6 +915,12 @@ def get(self):
       logs.log_error('Failed to get honggfuzz Fuzzer entity.')
       return
 
+    syzkaller = data_types.Fuzzer.query(
+        data_types.Fuzzer.name == 'syzkaller').get()
+    if not libfuzzer:
```
inferno-chromium: Incorrect variable libfuzzer. Later when we patch project_setup, we need to add test in project_setup_test.py so that these can be easily detected. Skip changing the file for now.

src/local/butler/scripts/setup.py
```diff
@@ -37,6 +37,20 @@
...
 
+SYZKALLER_TEMPLATE = """MAX_FUZZ_THREADS = 1
+MAX_TESTCASES = 4
+FUZZ_TEST_TIMEOUT = 4800
+TEST_TIMEOUT = 65
+WARMUP_TIMEOUT = 65
+BAD_BUILD_CHECK = False
+THREAD_ALIVE_CHECK_INTERVAL = 1
+REPORT_OOMS_AND_HANGS = True
+CORPUS_FUZZER_NAME_OVERRIDE = syzkaller
```
inferno-chromium: Skip CORPUS_FUZZER_NAME_OVERRIDE as it is only needed when you need to change it.

src/local/butler/scripts/setup.py
```diff
@@ -37,6 +37,20 @@
 THREAD_DELAY = 30.0
 """
 
+SYZKALLER_TEMPLATE = """MAX_FUZZ_THREADS = 1
+MAX_TESTCASES = 4
+FUZZ_TEST_TIMEOUT = 4800
+TEST_TIMEOUT = 65
```
inferno-chromium: Is it this long needed to reproduce testcase ? Ideally this should be under 20 sec, otherwise regression testing, etc will be very slow.

src/local/butler/scripts/setup.py
```diff
@@ -223,6 +238,58 @@ def __init__(self):
     self.key_id = 1339
 
 
+class SyzkallerDefaults(BaseBuiltinFuzzerDefaults):
```
inferno-chromium: This looks copied from libFuzzer, but Syzkaller will have its own set of stats. Skip this for now or add the stat variable that syzkaller exports after a fuzzer run. This is not needed now, first we integrate fuzzing, then add stat vars.

src/python/base/tasks.py
```diff
@@ -68,6 +68,7 @@
     'MAC': 'Mac',
     'WINDOWS': 'Windows',
     'WINDOWS_WITH_GPU': 'Windows (with GPU)',
+    'ANDROID_KERNEL': 'Android Kernel',
```
inferno-chromium: nit: move between line 64,65.

src/python/bot/fuzzers/syzkaller/constants.py
```diff
@@ -0,0 +1,28 @@
...
+# limitations under the License.
+"""Constants that are meaningful to syzkaller.
+Should not have any dependancies.
+"""
+
+# syzkaller's exit code if a bug occurred in syzkaller.
+SYZKALLER_ERROR_EXITCODE = 1
+
+# syzkaller's exit code if a bug was found in the kernel code.
+TARGET_ERROR_EXITCODE = 77
```
inferno-chromium: Are these exit codes correct/used ? these seem to be libfuzzer, skip for now if unused.

src/python/bot/fuzzers/syzkaller/constants.py
```diff
@@ -0,0 +1,28 @@
...
+
+TIMEOUT_FLAG = '-timeout='
```
inferno-chromium: I dont think these ones are used, please remove.

src/python/bot/tasks/update_task.py
```diff
@@ -169,6 +169,8 @@ def run_platform_init_scripts():
   plt = environment.platform()
   if plt == 'ANDROID':
     android_init.run()
+  elif plt == 'ANDROID_KERNEL':
+    linux_init.run()
```
inferno-chromium: this should be android_init.run as we initialize device.

src/python/fuzzing/strategy.py
```diff
@@ -62,6 +62,10 @@
     MUTATOR_PLUGIN_RADAMSA_STRATEGY,
 ]
 
+SYZKALLER_STRATEGY_LIST = [
```
inferno-chromium: Add a TODO to add more.

src/python/tests/core/bot/fuzzers/syzkaller/Data/crash
```diff
@@ -0,0 +1,19 @@
+c5  11608 SELinux: unrecognized netlink message: protocol=0 nlmsg_type=5120 sclass=netlink_route_socket
```
inferno-chromium: Rename Data to test_data.
Rename crash to crash.txt

src/python/bot/fuzzers/syzkaller/stats.py
```diff
@@ -0,0 +1,332 @@
+# Copyright 2019 Google LLC
```
inferno-chromium: This file need to be removed, should be added for syzkaller case when we know which stats it exports.

src/python/tests/core/bot/fuzzers/syzkaller/Data/crash
```diff
@@ -0,0 +1,19 @@
...
+c7  11612 SELinux: unrecognized netlink message: protocol=0 nlmsg_type=5120 sclass=netlink_route_socket
+c7  11616 SELinux: unrecognized netlink message: protocol=0 nlmsg_type=46823 sclass=netlink_route_socket
+c6    907 ==================================================================
+c6    907 BUG: KASAN: slab-out-of-bounds in save_stack_trace_tsk+0x3c0/0x420 usr/local/google/buildbot/src/partner-android/o-dev-msm-wahoo-4.4/private/msm-google/arch/arm64/kernel/stacktrace.c:169
+c6    907 Read of size 4 at addr ffffffc06155b8f0 by task kworker/u16:10/907
+c6    907 
+c6    907 CPU: 6 PID: 907 Comm: kworker/u16:10 Tainted: G        WC      4.4.205-gc19cb486686f-ab6115091 #1
+c6    907 Hardware name: Qualcomm Technologies, Inc. MSM8998 v2.1 (DT)
+c6    907 usercopy: kernel memory overwrite attempt detected to ffffffc06155b3a0 (selinux_file_security) (8 bytes)
```
inferno-chromium: Are these tests used ?

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -0,0 +1,241 @@
+# Copyright 2020 Google LLC
```
inferno-chromium: @oliverchang to review syzkaller directory and give comments.

src/local/butler/scripts/setup.py
hzawawy: There are two different modes for syzkaller repro'ing, one using syz-repro that minimizes and repro (this takes time), and another using syz-execprog that only repro (this is fast). So I will change it to 20 and we will assume this will run the fast method and that minimization will be its own task.

src/local/butler/scripts/setup.py
hzawawy: Done.

src/appengine/handlers/cron/project_setup.py
hzawawy: Done.

src/local/butler/scripts/setup.py
hzawawy: Done (removed self.stats_column_descriptions = '''fuzzer: "Fuzz target")

src/python/base/tasks.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/constants.py
hzawawy: Done.

src/python/bot/tasks/update_task.py
hzawawy: Ok, I did that because we use TF which flashes the device already. But I will change as per your comment to support the general case.

src/python/fuzzing/strategy.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/stats.py
hzawawy: Done.

src/python/tests/core/bot/fuzzers/syzkaller/Data/crash
hzawawy: Not yet.

src/python/tests/core/bot/fuzzers/syzkaller/Data/crash
hzawawy: Done

src/appengine/handlers/cron/project_setup.py
hzawawy: Done.

configs/test/pubsub/queues.yaml
```diff
@@ -40,6 +40,10 @@ resources:
     type: queue.jinja
   - name: high-end-jobs-android
     type: queue.jinja
+  - name: jobs-android-kernel
+    type: queue.jinja
+  - name: high-end-jobs-android-kernel
```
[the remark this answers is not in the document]
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -0,0 +1,205 @@
...
+
+    Args:
+      corpus_dir: The main corpus directory.
+      target_path: Path to the target.
+      build_dir: Path to the build directory.
+
+    Returns:
+      A FuzzOptions object.
+    """
+    del build_dir
```
oliverchang: remove?

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -0,0 +1,205 @@
...
+    arguments = fuzzer.get_arguments(target_path)
+
+    # Add strategies here
```
oliverchang: nit: convert to a TODO? Or if this doesn't make sense in the context of syzkaller, just remove the comment?

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -0,0 +1,205 @@
...
+    """
+    profiler.start_if_needed('syzkaller_kasan')
+    runner = fuzzer.get_runner(target_path)
+
+    syzkaller_path = os.path.join(
+        environment.get_value('BUILD_DIR'), 'syzkaller')
+    if not os.path.exists(syzkaller_path):
+      raise SyzkallerError('syzkaller not found in build')
+
+    binary_path = syzkaller_path + '/bin/linux_arm64'
```
oliverchang: put /bin/linux_arm64 into a constant, and use os.path.join instead for the path components.

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -0,0 +1,205 @@
...
+      output_path: Path to the cleansed output.
+      max_time: Maximum allowed time for the cleanse.
+
+    Returns:
+      A ReproduceResult.
+    """
+    raise NotImplementedError
+
+
+class ReproduceResult(object):
```
oliverchang: From lines 181 down, they look copy-pasted and unnecesssary.

src/python/bot/fuzzers/syzkaller/fuzzer.py
```diff
@@ -0,0 +1,167 @@
...
+import tempfile
+
+# Regex to find testcase path from a crash.
+KASAN_CRASH_TESTCASE_REGEX = (r'.*Test unit written to\s*'
+                              r'(Read|Write) of .*')
+
+
+def get_arguments(fuzzer_path):
+  """Get arguments for a given fuzz target."""
+  del fuzzer_path
```
oliverchang: why the del here too?

src/python/bot/fuzzers/syzkaller/fuzzer.py
```diff
@@ -0,0 +1,167 @@
...
+  build_dir = environment.get_value('BUILD_DIR')
+  device_serial = environment.get_value('ANDROID_SERIAL')
+  json_config_path = '/tmp/' + device_serial + '/config.json'
```
oliverchang: use os.path.join.

src/python/bot/fuzzers/syzkaller/fuzzer.py
```diff
@@ -0,0 +1,167 @@
...
+  generate_config.run(
+      serial=device_serial,
+      work_dir_path='/tmp/syzkaller',
```
oliverchang: where do these paths come from? can we put them in a constant where it makes sense to?

src/python/bot/fuzzers/syzkaller/fuzzer.py
```diff
@@ -0,0 +1,167 @@
...
+        '%s%d' % (constants.TIMEOUT_FLAG, constants.REPRODUCTION_TIMEOUT_LIMIT))
+
+  def fuzz(self,
+           corpus_directories,
+           fuzz_timeout,
+           artifact_prefix=None,
+           additional_args=None,
+           extra_env=None):
+    """This is where actual syzkaller fuzzing is done."""
+    del corpus_directories, artifact_prefix, extra_env
```
oliverchang: again, why are these del calls needed?

src/python/bot/fuzzers/syzkaller/fuzzer.py
```diff
@@ -0,0 +1,167 @@
...
+      # Copy crash testcase contents into the main testcase path.
+      crashes.append(
+          Crash(crash_testcase_file_path, fuzz_logs, reproduce_arguments,
+                actual_duration))
+
+    return FuzzResult(fuzz_logs, fuzz_result.command, crashes, parsed_stats,
+                      fuzz_result.time_executed)
+
+
+class Crash(object):
```
oliverchang: Remove these and return the base engine.Crash/FuzzResult objects.

src/python/bot/fuzzers/syzkaller/generate_config.py
```diff
@@ -0,0 +1,62 @@
...
+  Args:
+    serial: (str)serial number of the device being fuzzed
+    kcov: (boolean) true if coverage is enabled
+    reproduce: (boolean) true if repro is enabled
+    vmlinux_path: (str) path to the vmlinux file
+    work_dir_path: (str) path to working directory of syzkaller
+
+  """
+  devices = {}
+  devices["devices"] = [serial]
```
oliverchang: use single quotes for all strings.

src/python/bot/tasks/update_task.py
```diff
@@ -169,6 +169,8 @@ def run_platform_init_scripts():
   plt = environment.platform()
   if plt == 'ANDROID':
     android_init.run()
+  elif plt == 'ANDROID_KERNEL':
```
oliverchang: just fold into above as 

```python
if plt in ('ANDROID', 'ANDROID_KERNEL'):
 ...
```

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+class SyzkallerEngine(object):
+  """Syzkaller fuzzing engine implementation."""
+
+  @property
+  def name(self):
+    return 'syzkaller'
+
+  def get_name(self):
```
oliverchang: what is this for?

src/python/bot/fuzzers/syzkaller/fuzzer.py
```diff
@@ -0,0 +1,167 @@
...
+  json_config_path = '/tmp/' + device_serial + '/config.json'
+  generate_config.run(
+      serial=device_serial,
+      work_dir_path='/tmp/syzkaller',
+      binary_path=build_dir + '/syzkaller',
+      vmlinux_path='/tmp/syzkaller/vmlinux',
+      config_path=json_config_path,
+      kcov=True,
+      reproduce=False)
+  arguments = ['--config', json_config_path]
```
oliverchang: nit: just 

return ['--config', json_config_path]

src/python/bot/fuzzers/syzkaller/fuzzer.py
```diff
@@ -0,0 +1,167 @@
...
+
+    # TODO: Parse stats information and add them to FuzzResult
+    parsed_stats = []
+
+    crashes = []
+    if crash_testcase_file_path:
+      arguments = additional_args[:]
+      #libfuzzer.remove_fuzzing_arguments(arguments)
+      arguments = additional_args[:]
+      reproduce_arguments = arguments[:]
```
oliverchang: there is a lot of duplication here.

Just do 

`reproduce_arguments = additional_args[:]`

That said, what do arguments look like here? Can these be passed as-is to syzkaller to reproduce the same bug? That's the intention of this argument.

src/python/bot/fuzzers/syzkaller/fuzzer.py
```diff
@@ -0,0 +1,167 @@
...
+  def get_testcase_path(self, log_lines):
+    """Get testcase path from log lines."""
+    for line in log_lines:
+      match = re.match(KASAN_CRASH_TESTCASE_REGEX, line)
+      if match:
+        return match.group(1)
+
+    return None
+
+  def fix_timeout_argument_for_reproduction(self, arguments):
```
oliverchang: this is copied from libFuzzer, but is it really necessary here? Also, TIMEOUT_FLAG is not defined in constants.py?

src/python/bot/fuzzers/syzkaller/fuzzer.py
```diff
@@ -0,0 +1,167 @@
...
+    additional_args = copy.copy(additional_args)
+    fuzz_result = self.run_and_wait(
+        additional_args=additional_args, timeout=fuzz_timeout)
+
+    log_lines = utils.decode_to_unicode(fuzz_result.output).splitlines()
+    fuzz_result.output = None
+    crash_testcase_file_path = self.get_testcase_path(log_lines)
+
+    if not crash_testcase_file_path and fuzz_result.return_code:
+      crash_testcase_file_path = self._create_empty_testcase_file()
```
oliverchang: Could you explain why this is needed?

src/python/bot/fuzzers/syzkaller/fuzzer.py
```diff
@@ -0,0 +1,167 @@
...
+  return arguments
+
+
+def get_runner(fuzzer_path):
+  """Return a suzkaller runner object."""
+  build_dir = environment.get_value('BUILD_DIR')
+  return SyzkallerRunner(fuzzer_path, build_dir)
+
+
+class Syzkaller(builtin.EngineFuzzer):
```
oliverchang: I think this is not needed or used anywhere.

src/python/bot/fuzzers/syzkaller/generate_config.py
```diff
@@ -0,0 +1,62 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Generate Syzkaller fuzzer config dynamically."""
+
+import json
+import os
+
+
+def run(serial,
```
oliverchang: Would prefer for this module to be named "config", and this method "generate".

i.e. config.generate()

src/python/bot/fuzzers/syzkaller/fuzzer.py
```diff
@@ -0,0 +1,167 @@
...
+  """Builtin syzkaller fuzzing engine."""
+
+  def generate_arguments(self, fuzzer_path):
+    """Generate arguments for fuzzer using .options file or default values."""
+    return ' '.join(get_arguments(fuzzer_path))
+
+
+class SyzkallerRunner(new_process.ProcessRunner):
```
oliverchang: nit: rename to AndroidSyzkallerRunner since this is specific to Android.

src/python/bot/fuzzers/syzkaller/fuzzer.py
```diff
@@ -0,0 +1,167 @@
...
+
+  def __init__(self, executable_path, default_args=None):
+    """Inits the SyzkallerRunner.
+
+    Args:
+      executable_path: Path to the fuzzer executable.
+      default_args: Default arguments to always pass to the fuzzer.
+    """
+    super(SyzkallerRunner, self).__init__(
+        executable_path=executable_path, default_args=None)
```
oliverchang: Could you please clarify what is actually being run here? Is this some wrapper binary which actually invokes syzkaller?

src/python/bot/fuzzers/syzkaller/engine.py
hzawawy: Review process with Oliver initiated.

src/python/bot/fuzzers/syzkaller/engine.py
hzawawy: I used the to pass the lint for unusued param (build_dir). As per your comment, I removed it but renamed the field unused in the field name.

src/python/bot/fuzzers/syzkaller/engine.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/engine.py
hzawawy: class ReproduceResult is needed as a return value and fails lint if removed. The other two methods (register and get) are removed.

src/python/bot/fuzzers/syzkaller/engine.py
hzawawy: Converted to TODO

src/python/bot/fuzzers/syzkaller/fuzzer.py
hzawawy: Same reason as earlier (path is not used). Fixed by renaming the field.

src/python/bot/fuzzers/syzkaller/fuzzer.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/fuzzer.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/fuzzer.py
hzawawy: same unusued params reason. Fixed with renaming params.

src/local/butler/scripts/setup.py
```diff
@@ -37,6 +37,19 @@
...
+SYZKALLER_TEMPLATE = """MAX_FUZZ_THREADS = 1
+MAX_TESTCASES = 4
+FUZZ_TEST_TIMEOUT = 4800
+TEST_TIMEOUT = 65
+WARMUP_TIMEOUT = 65
+BAD_BUILD_CHECK = False
+THREAD_ALIVE_CHECK_INTERVAL = 1
+REPORT_OOMS_AND_HANGS = True
+ADDITIONAL_PROCESSES_TO_KILL = syz-manager
+ENABLE_GESTURES = False
```
kalder: What does ENABLE_GESTURES do?

src/python/bot/fuzzers/syzkaller/fuzzer.py
hzawawy: Done

src/python/bot/fuzzers/syzkaller/generate_config.py
hzawawy: Done.

src/python/bot/tasks/update_task.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/engine.py
hzawawy: Removed.

src/python/bot/fuzzers/syzkaller/fuzzer.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/fuzzer.py
hzawawy: Thanks for clarifying that. I will assign it to None and add a TODO here to handle the repro params.

src/python/bot/fuzzers/syzkaller/fuzzer.py
hzawawy: Removed and will be added once the repro logic is added.

src/python/bot/fuzzers/syzkaller/fuzzer.py
hzawawy: If we exited with no crash in output, then this should create an empty testcase to store it as a  crash. I was using it mostly for testing until the syzkaller work is stable completed. I will add a todo to remove it once all syzkaller related PR are in.

src/python/bot/fuzzers/syzkaller/fuzzer.py
hzawawy: removed.

src/python/bot/fuzzers/syzkaller/generate_config.py
hzawawy: Done.

src/local/butler/scripts/setup.py
mspectorgoogle: 20 seconds may not be enough for repro because the phone has to reboot.  Plus if we don't disable ramdump you're going to need like 10 minutes,

src/python/bot/fuzzers/syzkaller/fuzzer.py
hzawawy: syzkaller consists of multiple binaries but the one that orchestrates all the work is syz-manager. Syz-manager runs and pushes a binary on the android device that will actually do the device fuzzing.

src/python/bot/fuzzers/syzkaller/generate_config.py
```diff
@@ -0,0 +1,62 @@
...
+  data["workdir"] = work_dir_path
+  data["http"] = "localhost:50001"
+  data["syzkaller"] = binary_path
+  data["suppressions"] = ["do_rt_sigqueueinfo", "do_rt_tgsigqueueinfo"]
+  data["vm"] = devices
+  data["kernel_obj"] = vmlinux_path
+  data["sandbox"] = "android"
+  data["ignores"] = ["WARNING:", "INFO:"]
+  data["type"] = "adb"
+  data["procs"] = 1
```
mspectorgoogle: Why are we keeping 1? we wont get races

hzawawy: What do you recommend as a procs value?

src/python/bot/fuzzers/syzkaller/fuzzer.py
```diff
@@ -0,0 +1,167 @@
...
+from bot.fuzzers.syzkaller import generate_config
+from builtins import object
+from system import environment
+from system import new_process
+import copy
+import re
+import tempfile
+
+# Regex to find testcase path from a crash.
+KASAN_CRASH_TESTCASE_REGEX = (r'.*Test unit written to\s*'
```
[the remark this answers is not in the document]
hzawawy: This will be addressed in the upcoming cl to create a testcase. I have added a todo for this purpose (also I fixed your other comment regarding the bin folder)

src/python/bot/fuzzers/syzkaller/engine.py
oliverchang: Just return the base engine.ReproduceResult. This shouldn't be duplicated here.

src/python/bot/fuzzers/syzkaller/engine.py
hzawawy: Fixed the path so now they are as per comment.

src/local/butler/scripts/setup.py
hzawawy: Ok so I will set it for 600 seconds.

src/python/bot/fuzzers/syzkaller/engine.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/fuzzer.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/generate_config.py
hzawawy: I will set it to 2.

src/python/bot/fuzzers/syzkaller/config.py
```diff
@@ -48,7 +48,7 @@ def generate(serial,
   data['sandbox'] = 'android'
   data['ignores'] = ['WARNING:', 'INFO:']
   data['type'] = 'adb'
-  data['procs'] = 1
+  data['procs'] = 2
```
mspectorgoogle: The original config code did:
// By default, use half of the available CPUs
        if (mProcs <= 0) {
            mProcs = getNumProcs(device) / 2;

On a pixel 2 this is 8 cores/2 = 4 procs.  I would suggest using 4, 2 will fuzz too slowly, 8 will probably cause the device to lock up

            // Ensure we use at least 1 CPU
            mProcs = Integer.max(1, mProcs);
        }

src/python/bot/fuzzers/syzkaller/config.py
```diff
@@ -35,24 +35,24 @@ def run(serial,
...
-  data["sandbox"] = "android"
-  data["ignores"] = ["WARNING:", "INFO:"]
-  data["type"] = "adb"
-  data["procs"] = 1
-  data["cover"] = kcov
+  data['reproduce'] = reproduce
+  data['workdir'] = work_dir_path
+  data['http'] = 'localhost:50001'
+  data['syzkaller'] = binary_path
+  data['suppressions'] = ['do_rt_sigqueueinfo', 'do_rt_tgsigqueueinfo']
```
mspectorgoogle: We should consider what this should be for android (its fine is its in a different cl)

src/python/bot/fuzzers/syzkaller/fuzzer.py
```diff
@@ -14,70 +14,61 @@
...
-  generate_config.run(
+  json_config_path = os.path.join('/tmp', device_serial, 'config.json')
+  config.generate(
       serial=device_serial,
-      work_dir_path='/tmp/syzkaller',
-      binary_path=build_dir + '/syzkaller',
-      vmlinux_path='/tmp/syzkaller/vmlinux',
+      work_dir_path=SYZKALLER_WORK_FOLDER,
+      binary_path=os.path.join(build_dir, 'syzkaller'),
+      vmlinux_path=VMLINUX_FOLDER,
```
mspectorgoogle: Who downloads this?

hzawawy: This is downloaded by the tradefed client ahead of fuzzing.

src/python/bot/fuzzers/syzkaller/generate_config.py
hzawawy: As per your other comment, I will update the number to 4 and add a todo to configure it dynamically.

src/python/bot/fuzzers/syzkaller/config.py
hzawawy: sg, will add a todo for this purpose.

src/python/bot/fuzzers/syzkaller/config.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/fuzzer.py
oliverchang: In that case, how are "fuzz targets" discovered by CF? We detect libFuzzer targets like so: 

https://github.com/google/clusterfuzz/blob/a53d92adb59155a5a5fa194b745389aa421394d8/src/python/bot/fuzzers/utils.py#L33

src/python/bot/fuzzers/syzkaller/fuzzer.py
```diff
@@ -0,0 +1,125 @@
...
+
+    fuzz_logs = '\n'.join(log_lines)
+
+    # TODO(hzawawy): Parse stats information and add them to FuzzResult
+    parsed_stats = []
+
+    crashes = []
+    if crash_testcase_file_path:
+      #TODO(hzawawy): add repro arguments
+      reproduce_arguments = None
```
oliverchang: This will probably cause an exception. Set it to `[]` instead.

src/python/bot/fuzzers/syzkaller/constants.py
```diff
@@ -0,0 +1,20 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Constants that are meaningful to syzkaller.
+Should not have any dependancies.
+"""
+
+REPRODUCTION_TIMEOUT_LIMIT = 600
+
+DEFAULT_TIMEOUT_LIMIT = 25
```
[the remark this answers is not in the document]
hzawawy: removed unused constant.

src/python/bot/fuzzers/syzkaller/constants.py
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Constants that are meaningful to syzkaller.
+Should not have any dependancies.
+"""
+
+REPRODUCTION_TIMEOUT_LIMIT = 600
```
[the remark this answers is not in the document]
hzawawy: removed unused constant.

src/python/bot/fuzzers/syzkaller/fuzzer.py
hzawawy: regarding first comment, Syzkaller is both the engine and the fuzzer, so this section does not apply to syzkaller.
regarding the second comment, done.

src/python/bot/fuzzers/syzkaller/fuzzer.py
```diff
@@ -0,0 +1,120 @@
+# Copyright 2020 Google LLC
```
oliverchang: nit: I know you used existing files as inspiration, but they were named for legacy reasons. Let's call this file "runner.py".

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -0,0 +1,169 @@
...
+    runner = fuzzer.get_runner(target_path)
+
+    syzkaller_path = os.path.join(
+        environment.get_value('BUILD_DIR'), 'syzkaller')
+    if not os.path.exists(syzkaller_path):
+      raise SyzkallerError('syzkaller not found in build')
+
+    binary_full_path = os.path.join(syzkaller_path, BIN_FOLDER_PATH)
+    for filename in os.listdir(binary_full_path):
+      os.chmod(os.path.join(binary_full_path, filename), 0o755)
```
oliverchang: can you do this chmod stuff in prepare() instead?

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -0,0 +1,169 @@
...
+
+    # Directory to place new units.
+    self._create_temp_corpus_dir('new')
+
+    fuzz_timeout = libfuzzer.get_fuzz_timeout(False, total_timeout=max_time)
```
oliverchang: I'm not sure it makes sense to use libfuzzer's get_fuzz_timeout here, since it's very specific to that. Can you implement a syzkaller specific one here?

src/python/bot/fuzzers/syzkaller/fuzzer.py
```diff
@@ -0,0 +1,120 @@
...
+      binary_path=os.path.join(build_dir, 'syzkaller'),
+      vmlinux_path=constants.VMLINUX_FOLDER,
+      config_path=json_config_path,
+      kcov=True,
+      reproduce=False)
+  return ['--config', json_config_path]
+
+
+def get_runner(fuzzer_path):
+  """Return a suzkaller runner object."""
```
oliverchang: nit: spelling of "suzkaller"

src/python/bot/fuzzers/syzkaller/constants.py
```diff
@@ -0,0 +1,24 @@
...
+# limitations under the License.
+"""Constants that are meaningful to syzkaller.
+Should not have any dependancies.
+"""
+
+# Regex to find testcase path from a crash.
+KASAN_CRASH_TESTCASE_REGEX = (r'.*Test unit written to\s*'
+                              r'(Read|Write) of .*')
+
+SYZKALLER_WORK_FOLDER = '/tmp/syzkaller'
```
oliverchang: This is fine for now, but how specific are these to Android conventions? Or is this a general syzkaller convention?

src/python/bot/fuzzers/syzkaller/engine.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/fuzzer.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/engine.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/fuzzer.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/constants.py
hzawawy: SYZKALLER_WORK_FOLDER is a standard param used in syzkaller config to specify where the logs and crash description/repro will be stored during fuzzing.

src/python/bot/fuzzers/syzkaller/fuzzer.py
hzawawy: As mentioned in my other note, syzkaller is the engine and the fuzzer (unlike libfuzzer).

src/python/tests/core/bot/fuzzers/syzkaller/syzkaller_fuzzer_test.py
```diff
@@ -0,0 +1,44 @@
...
+
+from tests.test_libs import helpers as test_helpers
+from tests.test_libs import test_utils
+
+TEST_PATH = os.path.abspath(os.path.dirname(__file__))
+DATA_DIR = os.path.join(TEST_PATH, 'test_data')
+TEMP_DIR = os.path.join(TEST_PATH, 'temp')
+
+
+def clear_temp_dir():
```
oliverchang: this is unused?

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -0,0 +1,169 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Fuzzing engine interface."""
+
+from bot.fuzzers import engine
+from bot.fuzzers import engine_common
+from bot.fuzzers import libfuzzer
```
oliverchang: Remove this unnecessary import. You can do a `python butler.py lint` to lint this PR.

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -0,0 +1,177 @@
...
+    return SyzkallerOptions(corpus_dir, arguments, None, None, None)
+
+  def _create_temp_corpus_dir(self, name):
+    """Create temporary corpus directory."""
+    new_corpus_directory = os.path.join(fuzzer_utils.get_temp_dir(), name)
+    engine_common.recreate_directory(new_corpus_directory)
+    return new_corpus_directory
+
+  def get_fuzz_timeout(self,
+                       unused_is_mutations_run=False,
```
oliverchang: Just remove these optional args. They're not relevant here.

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -0,0 +1,177 @@
...
+                       unused_total_timeout=None):
+    """Get the fuzz timeout."""
+    return engine_common.get_hard_timeout()
+
+  def fuzz(self,
+           target_path,
+           options,
+           unused_reproducers_dir=None,
+           unused_max_time=0):
```
oliverchang: This should be piped to engine_common.get_hard_timeout(total_timeout=...).

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -0,0 +1,177 @@
...
+    """Create temporary corpus directory."""
+    new_corpus_directory = os.path.join(fuzzer_utils.get_temp_dir(), name)
+    engine_common.recreate_directory(new_corpus_directory)
+    return new_corpus_directory
+
+  def get_fuzz_timeout(self,
+                       unused_is_mutations_run=False,
+                       unused_total_timeout=None):
+    """Get the fuzz timeout."""
+    return engine_common.get_hard_timeout()
```
oliverchang: Pipe the max_time provided to `fuzz` here.

src/python/bot/fuzzers/syzkaller/constants.py
oliverchang: So these are paths on the device, not the host where CF runs right?

src/python/bot/fuzzers/syzkaller/fuzzer.py
oliverchang: Yes, so this won't work as is, as CF expects to find a target binary and pass the path through to here. Add a TODO to make that work.

src/python/bot/fuzzers/syzkaller/engine.py
hzawawy: This version is outdated. The import is removed and the commit did pass lint.

src/python/tests/core/bot/fuzzers/syzkaller/syzkaller_fuzzer_test.py
hzawawy: This is a template for the unit test. I will add to it in the upcoming. I am fine either way removing it for now or keeping it.

src/python/bot/fuzzers/syzkaller/engine.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/constants.py
hzawawy: The paths are on the host. syz-manager handles running the fuzzing on the device and pulling the crash details into the host.

src/python/bot/fuzzers/syzkaller/fuzzer.py
hzawawy: Added. But note that using local testing the binary was downloaded and unpacked and run successfully without any change here.

src/python/bot/fuzzers/syzkaller/engine.py
hzawawy: Used the env variable RUN_TIMEOUT here.

src/python/bot/fuzzers/syzkaller/engine.py
hzawawy: done.

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -0,0 +1,164 @@
...
+from bot.fuzzers import engine_common
+from bot.fuzzers import utils as fuzzer_utils
+from bot.fuzzers.syzkaller import runner
+from builtins import object
+from metrics import profiler
+from system import environment
+import os
+
+BIN_FOLDER_PATH = 'bin'
+_ENGINES = {}
```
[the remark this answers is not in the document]
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -0,0 +1,164 @@
...
+  options."""
+
+  def __init__(self, corpus_dir, arguments, strategies, fuzz_corpus_dirs,
+               extra_env):
+    super(SyzkallerOptions, self).__init__(corpus_dir, arguments, strategies)
+    self.fuzz_corpus_dirs = fuzz_corpus_dirs
+    self.extra_env = extra_env
+
+
+class SyzkallerEngine(object):
```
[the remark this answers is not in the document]
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/runner.py
```diff
@@ -0,0 +1,120 @@
...
+    return None
+
+  def fuzz(self,
+           fuzz_timeout,
+           additional_args,
+           unused_additional_args=None,
+           unused_extra_env=None):
+    """This is where actual syzkaller fuzzing is done."""
+    additional_args = copy.copy(additional_args)
+    fuzz_result = self.run_and_wait(additional_args, timeout=fuzz_timeout)
```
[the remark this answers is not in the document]
hzawawy: Added clean exit time.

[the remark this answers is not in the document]
oliverchang: Thanks, but this isn't sufficient. Is there some way to pass `max_time` as an `additional_arg` to the syzkaller binary itself, so it knows to clean up and exit when the time is up?

PR #1439, opened 2020-02-12
Corpus backup: overwrite public.zip every time we update the public backup.

Dor1s:
Needed for CIFuzz, and I think it general we should document it in OSS-Fuzz. I don't think there is a way to list a bucket without authentication, therefore it's more of a necessity for sharing OSS-Fuzz public corpora more explicitly.

### DIFF
--- a/src/appengine/handlers/cron/corpus_backup.py
+++ b/src/appengine/handlers/cron/corpus_backup.py
@@ -27,6 +27,19 @@ from libs import handler
 from metrics import logs
 
 
+def _set_public_acl_if_needed(url):
+  """Sets public ACL on the object with given URL, if it's not public yet."""
+  if storage.get_acl(url, 'allUsers'):
+    logs.log('%s is already marked public, skipping.' % url)
+    return True
+
+  if not storage.set_acl(url, 'allUsers'):
+    logs.log_error('Failed to mark %s public.' % url)
+    return False
+
+  return True
+
+
 def _make_corpus_backup_public(target, corpus_fuzzer_name_override,
                                corpus_backup_bucket_name):
   """Identifies old corpus backups and makes them public."""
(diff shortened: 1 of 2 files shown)


src/appengine/handlers/cron/corpus_backup.py
```diff
@@ -66,6 +67,24 @@ def _make_corpus_backup_public(target, corpus_fuzzer_name_override,
...
+      corpus_manager.PUBLIC_BACKUP_TIMESTAMP + os.extsep +
+      corpus_manager.BACKUP_ARCHIVE_FORMAT)
+  public_url = os.path.join(os.path.dirname(corpus_backup_url), filename)
+
+  try:
+    result = storage.copy_blob(corpus_backup_url, public_url)
+    if result:
+      result = storage.set_acl(public_url, 'allUsers')
+  except:
+    result = None
```
oliverchang: Why not log_error here instead? That way we log any exception stacks and can avoid this extraneous `result = None` setting and `if not result`.

Dor1s: You're right, it's pointless. I've just blindly followed the existing pattern. Fixed this in my diff and two other places above. Thank you!

Dor1s: Or actually it had a point, wait a second

Dor1s: it seems the intention was to log an error in both cases:
- if there was an exception
- if `return == False`

Dor1s: Will think more about this tomorrow. Looks like we can get rid of all try-except blocks inside this function and handle an exception outside of it.

Dor1s: yeah, I think now it's much better

PR #1440, opened 2020-02-12
Fix name of cross_pollination_stats table

mpherman2:


configs/test/bigquery/datasets.yaml
```diff
@@ -230,7 +230,7 @@ resources:
     properties:
       datasetId: $(ref.main-dataset.datasetReference.datasetId)
       tableReference:
-        tableId: cross-pollination-statistics
+        tableId: cross_pollination_statistics
       timePartitioning:
         type: DAY
       schema:
```
mbarbella-chromium: Could you also indent the children under fields: below here for consistency with the other tables.

mpherman2: Didn't notice, thanks for pointing that out. Done

PR #1442, opened 2020-02-12
Fix an exception with non-unicode stack traces in testcase details.

mbarbella-chromium:
PTAL. This doesn't seem to be working on staging due to what looks like an unrelated problem (stale exception trace issue I mentioned on chat), but from a quick test I'm guessing this should work:

```
>>> from builtins import str
>>> str(b'\xca', errors='replace')
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/dist-packages/future/types/newstr.py", line 102, in __new__
    return super(newstr, cls).__new__(cls, value)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xca in position 0: ordinal not in range(128)
>>> str(b'\xca', 'utf-8', errors='replace')
'\ufffd'
>>>
```

### DIFF
--- a/src/python/datastore/data_handler.py
+++ b/src/python/datastore/data_handler.py
@@ -560,7 +560,7 @@ def get_stacktrace(testcase, stack_attribute='crash_stacktrace'):
   # For App Engine, we can't write to local file, so use blobs.read_key instead.
   if environment.is_running_on_app_engine():
     key = result[len(data_types.BLOBSTORE_STACK_PREFIX):]
-    return str(blobs.read_key(key), errors='replace')
+    return str(blobs.read_key(key), 'utf-8', errors='replace')
 
   key = result[len(data_types.BLOBSTORE_STACK_PREFIX):]
   tmpdir = environment.get_value('BOT_TMPDIR')


mbarbella-chromium:
> LGTM, but is there a way we can set the default encoding on app engine?

I'm not sure that would actually help in this case. I think it might have something to do with how newstr is implemented. For example, this also seems to avoid the exception:

```
>>> from builtins import str
>>> str('\xca', 'ascii', errors='replace')
'\ufffd'
```

oliverchang:
> > LGTM, but is there a way we can set the default encoding on app engine?
> 
> I'm not sure that would actually help in this case. I think it might have something to do with how newstr is implemented. For example, this also seems to avoid the exception:
> 
> ```
> >>> from builtins import str
> >>> str('\xca', 'ascii', errors='replace')
> '\ufffd'
> ```

Wow, that's pretty surprising.

mbarbella-chromium:
Looking at the implementation more closely they only decode with the args/kwargs if len(args) > 1 (not counting kwargs length) or 'encoding' is in kwargs. Definitely seems like a bug in newstr.

PR #1444, opened 2020-02-13
Add support for tagged cross pollination

mpherman2:
I still need to run this locally in case the unit tests missed anything, but I wanted to put this up first so you guys could see if anything that needs to change.

### DIFF
--- a/src/python/bot/tasks/corpus_pruning_task.py
+++ b/src/python/bot/tasks/corpus_pruning_task.py
@@ -878,15 +878,14 @@ def _save_coverage_information(context, result):
 def choose_cross_pollination_strategy(current_fuzzer_name):
   """Chooses cross pollination strategy. In seperate function to mock for
     predictable test behaviror."""
-  # Do not choose tagged method if there is no tag for the current fuzzer.
   method = random.choice([Pollination.RANDOM, Pollination.TAGGED])
   if method == Pollination.TAGGED:
     similar_targets = corpus_tagging.get_similarly_tagged_fuzzers(
         current_fuzzer_name)
     if similar_targets:
-      return (Pollination.TAGGED, similar_targets.keys())
+      return (Pollination.TAGGED, random.choice(similar_targets.keys()))
 
-  return (Pollination.Random, None)
+  return (Pollination.RANDOM, None)
 
 
 def execute_task(full_fuzzer_name, job_type):
@@ -913,9 +912,8 @@ def execute_task(full_fuzzer_name, job_type):
     raise CorpusPruningException(
         'Failed to set up fuzzer %s.' % fuzz_target.engine)
 
-  method_and_tag = choose_cross_pollination_strategy(full_fuzzer_name)
-  cross_pollination_method = method_and_tag[0]
-  tag = method_and_tag[1]
+  cross_pollination_method, tag = choose_cross_pollination_strategy(
+      full_fuzzer_name)
 
   # TODO(unassigned): Use coverage information for better selection here.
   cross_pollinate_fuzzers = _get_cross_pollinate_fuzzers(
--- a/src/python/bot/tasks/corpus_pruning_task.py
+++ b/src/python/bot/tasks/corpus_pruning_task.py
@@ -164,10 +164,13 @@ class Context(object):
   def __init__(self,
                fuzz_target,
                cross_pollinate_fuzzers,
-               cross_pollination_method=RANDOM_METHOD):
+               cross_pollination_method=RANDOM_METHOD,
+               tag=None):
+
     self.fuzz_target = fuzz_target
     self.cross_pollinate_fuzzers = cross_pollinate_fuzzers
     self.cross_pollination_method = cross_pollination_method
+    self.tag = tag
 
     self.merge_tmp_dir = None
     self.engine = engine.get(self.fuzz_target.engine)
@@ -531,7 +534,7 @@ class CrossPollinator(object):
 
 
 def record_cross_pollination_stats(
-    pruner_stats, pollinator_stats, project_qualified_name, sources, tags,
+    pruner_stats, pollinator_stats, project_qualified_name, sources, tag,
     initial_corpus_size, minimized_corpus_size_units, method):
   """Log stats about cross pollination in BigQuery."""
   # TODO(mpherman): Find a way to collect these stats for OSS Fuzz.
@@ -550,7 +553,7 @@ def record_cross_pollination_stats(
       'project_qualified_name': project_qualified_name,
       'method': method,
       'sources': sources,
-      'tags': tags,
+      'tags': tag if tag else '',
       'initial_corpus_size': initial_corpus_size,
       'corpus_size': minimized_corpus_size_units,
       'initial_edge_coverage': pruner_stats['edge_coverage'],
@@ -686,17 +689,9 @@ def do_corpus_pruning(context, last_execution_failed, revision):
       for fuzzer in context.cross_pollinate_fuzzers
   ])
 
-  # Leave tags blank if random method was used. Even if corpus has tags.
-  tags = ''
-  if context.cross_pollination_method == TAGGED_METHOD:
-    tags = ','.join([
-        corpus_tag.tag for corpus_tag in corpus_tagging.get_fuzz_target_tags(
-            context.fuzz_target.fully_qualified_name())
-    ])
-
   record_cross_pollination_stats(
-      pruner_stats, pollinator_stats, project_qualified_name, sources, tags,
-      pre_pollination_corpus_size, minimized_corpus_size_units,
+      pruner_stats, pollinator_stats, project_qualified_name, sources,
+      context.tag, pre_pollination_corpus_size, minimized_corpus_size_units,
       context.cross_pollination_method)
 
   return result
@@ -783,13 +778,15 @@ def _process_corpus_crashes(context, result):
 
 
 def _select_targets_and_jobs_for_pollination(engine_name, current_fuzzer_name,
-                                             method):
+                                             method, tag):
   """Select jobs to use for cross pollination."""
   target_jobs = list(fuzz_target_utils.get_fuzz_target_jobs(engine=engine_name))
-  similar_tagged_targets = corpus_tagging.get_similarly_tagged_fuzzers(
-      current_fuzzer_name)
-  # Cross Pollinate randomly if there are no similarly tagged targets.
-  if similar_tagged_targets and method == TAGGED_METHOD:
+
+  if method == TAGGED_METHOD:
+    similar_tagged_targets = [
+        target for target in corpus_tagging.get_targets_with_tag(tag)
+        if target.fully_qualified_fuzz_target_name != current_fuzzer_name
+    ]
     # Intersect target_jobs and similar_tagged_targets on fully qualified
     # fuzz target name.
     target_jobs = [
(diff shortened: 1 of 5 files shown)


src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -157,9 +158,13 @@ def __init__(self, fuzz_target, backup_bucket_name, corpus_engine_name):
 class Context(object):
   """Pruning state."""
 
-  def __init__(self, fuzz_target, cross_pollinate_fuzzers):
+  def __init__(self,
+               fuzz_target,
+               cross_pollinate_fuzzers,
+               cross_pollination_method="random"):
```
oliverchang: Use single quotes for all strings

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -679,12 +683,19 @@ def do_corpus_pruning(context, last_execution_failed, revision):
       for fuzzer in context.cross_pollinate_fuzzers
   ])
 
-  # TODO(mpherman): Add way to get tags string when tagged corpora are done.
+  # Leave tags blank if random method was used. Even if corpus has tags.
   tags = ''
+  if context.cross_pollination_method == "tagged":
```
oliverchang: single quote here too.

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -768,11 +779,25 @@ def _process_corpus_crashes(context, result):
...
-  cross_pollinate_fuzzers = []
-
+def _select_targets_and_jobs_for_pollination(engine_name, current_fuzzer_name,
+                                             method):
+  """Select jobs to use for cross pollination."""
   target_jobs = list(fuzz_target_utils.get_fuzz_target_jobs(engine=engine_name))
+  similar_tagged_targets = corpus_tagging.get_similarly_tagged_fuzzers(
+      current_fuzzer_name)
+  # Cross Pollinate randomly if there are no similarly tagged targets.
+  if similar_tagged_targets:
```
oliverchang: you can just do
`if similar_tagged_targets and method == 'tagged'` 

here to avoid an indentation level.

src/python/datastore/corpus_tagging.py
```diff
@@ -29,3 +29,16 @@ def get_targets_with_tag(tag):
   query = data_types.CorpusTag().query()
   query = query.filter(data_types.CorpusTag.tag == tag)
   return query.fetch()
+
+
+def get_similarly_tagged_fuzzers(current_fully_qualified_fuzzer_name):
+  """Get all fuzz targets with any similar tags to current fuzzer."""
+  similarly_tagged_targets = []
+  for tag in get_fuzz_target_tag(current_fully_qualified_fuzzer_name):
```
oliverchang: nit: `get_fuzz_target_tag` should be plural since it can return more than 1.

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -679,12 +683,19 @@ def do_corpus_pruning(context, last_execution_failed, revision):
       for fuzzer in context.cross_pollinate_fuzzers
   ])
 
-  # TODO(mpherman): Add way to get tags string when tagged corpora are done.
+  # Leave tags blank if random method was used. Even if corpus has tags.
   tags = ''
+  if context.cross_pollination_method == "tagged":
+    tags = ','.join([
```
Dor1s: so if a fuzz target have multiple tags, are we going to use all of them at once? I think it would be better to choose just one, because:
1) it will give us better tracking data, i.e. we should be able to see which tag works better than the other(s) for a given fuzz target
2) it will keep the number of sources for cross pollination smaller, which is also important, we shouldn't try to add too much corpora at once

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -768,11 +779,25 @@ def _process_corpus_crashes(context, result):
...
-
+def _select_targets_and_jobs_for_pollination(engine_name, current_fuzzer_name,
+                                             method):
+  """Select jobs to use for cross pollination."""
   target_jobs = list(fuzz_target_utils.get_fuzz_target_jobs(engine=engine_name))
+  similar_tagged_targets = corpus_tagging.get_similarly_tagged_fuzzers(
+      current_fuzzer_name)
+  # Cross Pollinate randomly if there are no similarly tagged targets.
+  if similar_tagged_targets:
+    # Even if there are similarly tagged targets, only use them half the time.
```
Dor1s: I don't understand the part about "half the time". Is there any random selection, or how is that ensured?

Dor1s: oh, I see, it's in `choose_cross_pollination_strategy` line 878. I don't think this comment makes sense here, better remove it

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -840,6 +875,12 @@ def _try_save_coverage_information():
         'Failed to save corpus pruning result: %s.' % repr(e))
 
 
+def choose_cross_pollination_strategy():
+  """Chooses cross pollination strategy. In seperate function to mock for
+    predictable test behaviror."""
+  return random.choice(['random', 'tagged'])
```
Dor1s: as we use these string literals (random and tagged) more than once, it might be better to put them in some global variables to make sure they're defined at a single place and use those variables (kinda constants, but not really in python) in all other places

src/python/datastore/corpus_tagging.py
```diff
@@ -29,3 +29,16 @@ def get_targets_with_tag(tag):
...
+    similarly_tagged_targets += get_targets_with_tag(tag.tag)
```
Dor1s: this can have duplicates, right? if two fuzz targets have two matching tags?

mpherman2:
Tagged cross pollination will choose just 1 tag now. The BigQuery table will also only show the tag that was selected. I don't love that the name of the field is "tags" and it will only ever have 1 tag in it, but that should change in another CL because we will need to update the config files as well.

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -768,11 +777,25 @@ def _process_corpus_crashes(context, result):
...
+  if method == TAGGED_METHOD:
+    similar_tagged_targets = [
+        target for target in corpus_tagging.get_targets_with_tag(tag)
+        if target.fully_qualified_fuzz_target_name != current_fuzzer_name
+    ]
+    # Intersect target_jobs and similar_tagged_targets on fully qualified
+    # fuzz target name.
+    target_jobs = [
+        target for target in target_jobs if target.fuzz_target_name in [
+            tagged.fully_qualified_fuzz_target_name
```
Dor1s: let's put these in a separate list or set, otherwise it'll be unnecessarily iterating, plus that code gets hard to read

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -840,6 +873,16 @@ def _try_save_coverage_information():
         'Failed to save corpus pruning result: %s.' % repr(e))
 
 
+def choose_cross_pollination_strategy(current_fuzzer_name):
+  """Chooses cross pollination strategy. In seperate function to mock for
+    predictable test behaviror."""
+  # Do not choose tagged method if there is no tag for the current fuzzer.
+  if corpus_tagging.get_similarly_tagged_fuzzers(current_fuzzer_name):
```
Dor1s: this calls into Datastore, let's not always do it unconditionally. Instead, do the `random.choice` first, if it returns random then you just return random; then check if the tagged method is possible, and branch the result based on it

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -864,11 +907,18 @@ def execute_task(full_fuzzer_name, job_type):
     raise CorpusPruningException(
         'Failed to set up fuzzer %s.' % fuzz_target.engine)
 
+  cross_pollination_method = choose_cross_pollination_strategy(full_fuzzer_name)
+  tag = None
+  if cross_pollination_method == TAGGED_METHOD:
+    tag = random.choice(
+        corpus_tagging.get_fuzz_target_tags(full_fuzzer_name)).tag
+
```
Dor1s: actually, it might be even better to give up on the `TAGGED_METHOD`. Let's say a fuzz target has two tags, one is used by other targets and the other one is unique. The call on line 880 can return some targets, while here we'll choose a tag that is unique for the current fuzzer and will not actually do any tagged cross pollination

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -106,6 +107,9 @@
     'security_flag',
 ])
 
+RANDOM_METHOD = 'random'
```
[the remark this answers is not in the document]
mbarbella-chromium: +1. It might be even cleaner to move them into a class (enum-style).

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -157,9 +161,16 @@ def __init__(self, fuzz_target, backup_bucket_name, corpus_engine_name):
 class Context(object):
   """Pruning state."""
 
-  def __init__(self, fuzz_target, cross_pollinate_fuzzers):
+  def __init__(self,
+               fuzz_target,
+               cross_pollinate_fuzzers,
```
mbarbella-chromium: Nit: I'm not sure how much this complicates the plumbing so feel free to ignore if it seems to add complexity, but could this and method be merged into one argument with a default of a new value that means "no cross pollination"?

src/python/bot/tasks/corpus_pruning_task.py
mpherman2: Done

src/python/bot/tasks/corpus_pruning_task.py
mpherman2: Done

src/python/bot/tasks/corpus_pruning_task.py
mbarbella-chromium: I misunderstood what this variable was tracking. Definitely seems fine to leave it as-is.

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -107,6 +109,11 @@
 ])
 
 
+class Pollination(Enum):
+  RANDOM = 1
```
mbarbella-chromium: It may still be useful for these values to be strings since it will be easier to understand when looking at the bigquery stats.

mpherman2: Done

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -840,6 +875,20 @@ def _try_save_coverage_information():
         'Failed to save corpus pruning result: %s.' % repr(e))
 
 
+def choose_cross_pollination_strategy(current_fuzzer_name):
+  """Chooses cross pollination strategy. In seperate function to mock for
+    predictable test behaviror."""
+  # Do not choose tagged method if there is no tag for the current fuzzer.
```
Dor1s: is this comment still applicable?

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -864,11 +913,16 @@ def execute_task(full_fuzzer_name, job_type):
     raise CorpusPruningException(
         'Failed to set up fuzzer %s.' % fuzz_target.engine)
 
+  method_and_tag = choose_cross_pollination_strategy(full_fuzzer_name)
```
Dor1s: nit: can do `method, tag = choose_cross_pollination_strategy(full_fuzzer_name`

mpherman2: Done

src/python/bot/tasks/corpus_pruning_task.py
mpherman2: Done

PR #1448, opened 2020-02-14
Fix UnboundLocalError in corpus pruning task.

mpherman2:
Still looking though logs to confirm that this is the bug, but this looks wrong to me. PTAL.

### DIFF
--- a/src/python/bot/tasks/corpus_pruning_task.py
+++ b/src/python/bot/tasks/corpus_pruning_task.py
@@ -513,8 +513,8 @@ class CrossPollinator(object):
           'Shared corpus merge finished successfully.',
           output=symbolized_output)
     except engine.TimeoutError as e:
-      logs.log_error('Corpus pruning timed out while merging shared corpus: ' +
-                     repr(e))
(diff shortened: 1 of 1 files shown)


src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -513,8 +513,8 @@ def run(self, timeout):
           'Shared corpus merge finished successfully.',
           output=symbolized_output)
     except engine.TimeoutError as e:
-      logs.log_error('Corpus pruning timed out while merging shared corpus: ' +
-                     repr(e))
+      raise CorpusPruningException(
+          'Corpus pruning timed out while merging shared corpus: ' + repr(e))
```
Dor1s: use `\n` at the end of the line, otherwise all these errors will be reported as unique

inferno-chromium:
No this is wrong, we explicitly didn't want to raise an exception for this case as it happens often.

mbarbella-chromium:
Ah, makes sense. I'll revert this.

PR #1452, opened 2020-02-14
Set minimum logging level for AppEngine to INFO.

inferno-chromium:
This avoid the noise from datastore api debug logs.


mbarbella-chromium:
Is this specific to the API on app engine? Do we also see similar logging on the bots?

inferno-chromium:
> Is this specific to the API on app engine? Do we also see similar logging on the bots?

For bots, we already do INFO. See few lines down.
  logger.setLevel(logging.INFO)

PR #1454, opened 2020-02-15
Fix log spam due to exception.

inferno-chromium:
if not original_issue.merged_into:
AttributeError: 'NoneType' object has no attribute 'merged_into'

Happens on issue tracker where some issues might be inaccessible.


inferno-chromium:
We can also do last accessible issue, but i feel like that is incorrect result.

PR #1459, opened 2020-02-19
Only run path patcher on Python 2.

mbarbella-chromium:
It relies on some builtins that are not available in Python 3, and it's
not clear that it's necessary. Similar functionality will be added if it
turns out we still need something like this after staging Python 3 on
Windows.

### DIFF
--- a/src/python/system/path_patcher.py
+++ b/src/python/system/path_patcher.py
@@ -139,6 +139,11 @@ def patch():
   if not _is_windows():
     return
 
+  # TODO(mbarbella): Verify that this is no longer necessary in Python 3 and
+  # remove it completely after transitioning.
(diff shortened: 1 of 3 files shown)


src/python/system/path_patcher.py
```diff
@@ -139,6 +139,11 @@ def patch():
   if not _is_windows():
     return
 
+  # TODO(mbarbella): Verify that this is no longer necessary in Python 3 and
```
inferno-chromium: Yes it seems not needed, see https://docs.python.org/3/using/windows.html#removing-the-max-path-limitation, can rephrase this TODO.

mbarbella-chromium: Done!

PR #1461, opened 2020-02-19
Fix cron failures due to redis connection errors.

inferno-chromium:


inferno-chromium:
This should fixup cleanup cron failure on google instance.

PR #1463, opened 2020-02-19
Fix improperly tokenized testcases causing minimization result to not reproduce error.

mpherman2:
This should stop any round of minimization where the tokenizer has failed. It logs when the tokenizer fails so that we can continue to amend the grammar as we see common missed tokens.

### DIFF
--- a/src/python/bot/minimizer/chunk_minimizer.py
+++ b/src/python/bot/minimizer/chunk_minimizer.py
@@ -19,6 +19,7 @@ import functools
 
 from . import minimizer
 from . import utils
+from metrics import logs
 
 
 class ChunkMinimizer(minimizer.Minimizer):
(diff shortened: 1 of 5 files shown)


mpherman2:
This is complicated to discuss here, so I will explain more in person. If I move the checks to minimize task the checks will be significantly more complicated, and errors will have worse consequences. HTML minimizer in particular just sends its data to other minimizers so it would be very hard to check the tokenizer until you are in the minimizer code.

mpherman2:
Done with refactoring. PTAL

src/python/bot/minimizer/chunk_minimizer.py
```diff
@@ -32,6 +33,8 @@ def __init__(self, *args, **kwargs):
   def _execute(self, data):
     """Minimize |data| using the algorithm from CF (but backwards)."""
     testcase = minimizer.Testcase(data, self)
+    if testcase.get_current_testcase_data() != data:
+      raise errors.TokenizationFailureError("JSMinimizer")
```
Dor1s: Is `JSMinimizer` a correct string here? Same for the delta minimizer class.

Also, please use single quotes for string literals

PR #1465, opened 2020-02-20
Fix NDB issues when forking with multiprocessing.Process.

oliverchang:
The NDB library does not allow re-initializing a context when one for
the thread exists, which is the case when we fork.

Removing this re-initialization is also not an option, as gRPC clients
do not work after forking.


oliverchang:
TBR. /gcbrun

PR #1468, opened 2020-02-20
Fix html minimization's tokenizer check always returning False

mpherman2:


src/python/bot/minimizer/minimizer.py
```diff
@@ -578,10 +578,27 @@ def minimize(self, data):
       testcase = error.testcase
     except errors.TokenizationFailureError:
       logs.log_error('Tokenized data did not match original data.')
-      return data
+      # In situation where the tokenizer does not work, we still want to return
+      # token_combiner([data]). This will just be the same as data unless
```
mbarbella-chromium: Nit: rather than reference token_combiner([data]) in the comments here and below, just note that we still want to use the token combiner. For the second sentence, maybe something like "In some cases the combiner will transform data in some way, such as by appending or prepending to data."

src/python/bot/minimizer/minimizer.py
```diff
@@ -578,10 +578,27 @@ def minimize(self, data):
...
+  def validate_tokenizer(self, data, testcase):
+    """Validate that the tokenizer correctly tokenized the data. This is
+    necessary because if the tokenizer does not recognize a character, it will
+    skip it."""
+    # For most token_combiners, token_combiner([data]) will be the same as data,
+    # but in situations where data is appended to the start or end of data in
+    # the token combiner, we want to make sure the same change happens to both
+    # the original data and the tokenized data.
+    data = self.token_combiner([data])
+    if testcase.get_current_testcase_data() == data:
```
mbarbella-chromium: Nit: return testcase.get_current_testcase_data == data.

src/python/bot/minimizer/minimizer.py
mpherman2: Done.

PR #1470, opened 2020-02-20
[fuchsia] Only initialize the global qemu log path when necessary

eepeep:
This change is to address some "No usable temporary directory found"
errors that were being thrown. This may not address the root cause (e.g.
lack of disk space/permission issues) but will at least avoid extra
errors for non-Fuchsia bots.

### DIFF
--- a/src/python/platforms/fuchsia/device.py
+++ b/src/python/platforms/fuchsia/device.py
@@ -122,7 +122,8 @@ class QemuProcess(object):
     self.logfile = None
 
     if not QemuProcess.LOG_PATH:
-      QemuProcess.LOG_PATH = os.path.join(tempfile.gettempdir(), 'fuchsia-qemu-log')
+      QemuProcess.LOG_PATH = os.path.join(tempfile.gettempdir(),
+                                          'fuchsia-qemu-log')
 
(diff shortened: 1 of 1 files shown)


oliverchang:
This fails lint :(

```
| C:125, 0: Line too long (84/80) (line-too-long)
| 
| -----------------------------------
| Your code has been rated at 9.91/10
| 
| Return code is non-zero (16).
Running: yapf -d src/python/platforms/fuchsia/device.py
| --- src/python/platforms/fuchsia/device.py	(original)
| +++ src/python/platforms/fuchsia/device.py	(reformatted)
| @@ -122,7 +122,8 @@
|      self.logfile = None
| 
|      if not QemuProcess.LOG_PATH:
| -      QemuProcess.LOG_PATH = os.path.join(tempfile.gettempdir(), 'fuchsia-qemu-log')
| +      QemuProcess.LOG_PATH = os.path.join(tempfile.gettempdir(),
| +                                          'fuchsia-qemu-log')
```

Please run `python butler.py format`.

eepeep:
Whoops, missed that somehow, sorry!

/gcbrun

PR #1471, opened 2020-02-20
[fuchsia] Update example fuzzer excludelist

eepeep:
An example fuzzer that intentionally OOMs was added so we can test that
ClusterFuzz is correctly catching OOMs on Fuchsia, but it was excluded
since we normally avoid running examples. This lets it run temporarily.

Additionally, this change disables a test Go fuzzer that isn't under the
examples package.

### DIFF
--- a/src/python/platforms/fuchsia/util/fuzzer.py
+++ b/src/python/platforms/fuchsia/util/fuzzer.py
@@ -73,9 +73,11 @@ class Fuzzer(object):
     """ Returns whether or not a given pkg/tgt pair is an example fuzzer.
     (Helper function to prevent us from wasting cycles on example fuzzers in
     production). """
-    # We want to remove the noop-fuzzer regardless of the extension.
+    # Strip any sanitizer extensions
     tgt = os.path.splitext(tgt)[0]
-    return pkg == 'example_fuzzers' or tgt == 'noop-fuzzer'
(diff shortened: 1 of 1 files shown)


src/python/platforms/fuchsia/util/fuzzer.py
```diff
@@ -73,9 +73,11 @@ def _is_example(cls, pkg, tgt):
     """ Returns whether or not a given pkg/tgt pair is an example fuzzer.
     (Helper function to prevent us from wasting cycles on example fuzzers in
     production). """
-    # We want to remove the noop-fuzzer regardless of the extension.
+    # Strip any sanitizer extensions
     tgt = os.path.splitext(tgt)[0]
-    return pkg == 'example_fuzzers' or tgt == 'noop-fuzzer'
+    return (pkg == 'example_fuzzers' and tgt != 'oom_fuzzer') \
```
oliverchang: nit: we generally don't use \ for line continuations. Use parentheses instead.

i.e.

```python
return ((pkg == 'example_fuzzers' and tgt != 'oom_fuzzer')
        or (pkg == 'zircon_fuzzers' and tgt == 'noop-fuzzer')...
```

eepeep: Done!

PR #1472, opened 2020-02-20
[fuchsia] Add runtime snapshot information to qemu failure logs

eepeep:
A warning will now be logged if stop_qemu is called but qemu isn't
running.  This will help to narrow down causes of connection failures.

### DIFF
--- a/src/python/platforms/fuchsia/device.py
+++ b/src/python/platforms/fuchsia/device.py
@@ -35,12 +35,6 @@ from system import environment
 from system import new_process
 from system import process_handler
 
-# Import will fail on AppEngine
-try:
-  import psutil
-except ImportError:
-  pass
-
 _QEMU_WAIT_SECONDS = 60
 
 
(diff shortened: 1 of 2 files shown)


src/python/platforms/fuchsia/device.py
```diff
@@ -214,6 +220,20 @@ def start_qemu():
 
 def stop_qemu():
   """Stop qemu."""
+
+  # Report whether qemu is running, for diagnostics
```
oliverchang: Could we do this in fuzzers/libfuzzer.py instead, in `_restart_qemu`, as that's closer to the problem we're trying to diagnose?

Also, I just realised we already have a method using psutil to dump processes, so we can just do this instead:

```python
  def _restart_qemu(self):
    """Restart QEMU."""
    logs.log_warn(
        'Connection to fuzzing VM lost. Restarting.',
        processes=process_handler.get_runtime_snapshot())
```

eepeep: Ah, that's handy, thanks for pointing it out!

PR #1476, opened 2020-02-21
Fix a unicode exception in the HTML minimizer.

mbarbella-chromium:
The combine_worker_functions function was not actually properly
converting the tokens to a bytes object. |tokens| were unicode objects
since Antlr converts the bytes we pass to it to unicode, which caused
the entire string to be treated as unicode. If the suffix contained
invalid utf-8, we'd hit a UnicodeDecodeError here.


mbarbella-chromium:
TBR

PR #1480, opened 2020-02-25
Integrate Peach Mutation strategy

mpherman2:


src/python/bot/fuzzers/libFuzzer/peach/pits.py
```diff
@@ -0,0 +1,37 @@
...
+from system import environment
+import os
+
+PIT_DIR = os.path.join(environment.get_platform_resources_directory(), 'peach',
+                       'pits')
+PATH = 0
+TITLE = 1
+# The key is the name of the grammar as defined in the libfuzzer options file.
+# The value is the path to the pit, and the title of the pit.
+PIT_INFORMATION = {'PDF': (os.path.join(PIT_DIR, 'pdf.xml'), 'PDF')}
```
mbarbella-chromium: Having this mapping seems like overkill to me. I'd prefer that get path function just adds ".xml" to its argument and returns None if the file doesn't exist. Callers could check this and log error so that we know if things are improperly configured.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -49,6 +51,7 @@
 from system import minijail
 from system import new_process
 from system import shell
+from zipfile import ZipFile
```
mbarbella-chromium: Nit: instead of this, "from system import archive" and use archive.unpack instead. It's fine to pass trusted=True given that the file is checked in.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1589,6 +1592,55 @@ def use_radamsa_mutator_plugin(extra_env):
...
+  new_path = new_path + sys.path
+  extra_env['PYTHONPATH'] = ':'.join(new_path)
+
+  # Select grammar.
+  grammar = fuzzer_utils.extract_argument(arguments, 'grammar')
+  # If the target has no grammar, do not use this strategy.
+  if not grammar:
+    return False
+
+  if not pits.validate(grammar):
```
mbarbella-chromium: Can we do this check earlier, ideally before extracting?

src/python/bot/fuzzers/libFuzzer/peach/pits.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def validate(grammar):
```
mbarbella-chromium: This and get_title could be removed if the comment above is addressed. The return value of get_path would be used for validation.

src/python/fuzzing/strategy.py
```diff
@@ -46,6 +46,8 @@
     name='recommended_dict', probability=0.10, manually_enable=False)
 VALUE_PROFILE_STRATEGY = Strategy(
     name='value_profile', probability=0.33, manually_enable=False)
+PEACH_GRAMMAR_MUTATION_STRATEGY = Strategy(
+    name='peach_grammar_mutation', probability=0.10, manually_enable=True)
```
mbarbella-chromium: This is fine to start, but we should likely increase it later. Since relatively few jobs will support this, we'll want to run it more often so that it's actually used when it can be used.

src/python/bot/fuzzers/libFuzzer/peach/pits.py
mpherman2: Done

src/python/bot/fuzzers/libFuzzer/peach/pits.py
mpherman2: Done

src/python/bot/fuzzers/libfuzzer.py
mpherman2: Done

src/python/bot/fuzzers/libfuzzer.py
mpherman2: Done

src/python/fuzzing/strategy.py
mpherman2: Ack

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1589,6 +1592,58 @@ def use_radamsa_mutator_plugin(extra_env):
...
+  """Decide whether or not to use peach mutator, and set up all of the
+  environment variables necessary to do so."""
+  # TODO(mpherman): Include architecture info in job definition and exclude
+  #  i386.
+  if environment.platform() != 'LINUX' or environment.get_value(
+      'MEMORY_TOOL') != 'ASAN':
+    return False
+
+  # Select grammar.
+  grammar = fuzzer_utils.extract_argument(arguments, 'grammar')
```
mbarbella-chromium: Could we move more of this logic into get_path? Essentially everything from here to through the log_error, returning None instead of false, then returning False from this function if not pit_path.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1589,6 +1592,58 @@ def use_radamsa_mutator_plugin(extra_env):
...
+  source = os.path.join(peach_dir, 'peach_mutator.zip')
+
+  archive.unpack(source, unzipped, trusted=True)
+
+  # Set LD_PRELOAD.
+  peach_path = os.path.join(unzipped, 'peach_mutator', 'src', 'peach.so')
+  extra_env['LD_PRELOAD'] = peach_path
+
+  # Set Python path.
+  # TODO(mpherman): Change this to explicitly point to Python 2
```
mbarbella-chromium: Nit: end with period.

src/python/fuzzing/strategy.py
mbarbella-chromium: Thinking more, could you add a TODO here so we don't forget about that? Sorry for not mentioning it last time.

src/python/bot/fuzzers/libfuzzer.py
mpherman2: Done

src/python/bot/fuzzers/libfuzzer.py
mpherman2: Done

src/python/fuzzing/strategy.py
mpherman2: Done

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1589,6 +1592,52 @@ def use_radamsa_mutator_plugin(extra_env):
   return True
 
 
+def use_peach_mutator(extra_env, arguments):
+  """Decide whether or not to use peach mutator, and set up all of the
+  environment variables necessary to do so."""
+  # TODO(mpherman): Include architecture info in job definition and exclude
+  #  i386.
+  if environment.platform() != 'LINUX' or environment.get_value(
```
oliverchang: nit: This line continuation looks a bit strange.

I'd do:

```python
if (environment.platform() != 'LINUX' or
    environment.get_value('MEMORY_TOOL') != 'ASAN'):
  ...

File: src/python/bot/fuzzers/libfuzzer.py [hunk 9]
```diff
@@ -1589,6 +1592,52 @@ def use_radamsa_mutator_plugin(extra_env):
...
+
+  archive.unpack(source, unzipped, trusted=True)
+
+  # Set LD_PRELOAD.
+  peach_path = os.path.join(unzipped, 'peach_mutator', 'src', 'peach.so')
+  extra_env['LD_PRELOAD'] = peach_path
+
+  # Set Python path.
+  # TODO(mpherman): Change this to explicitly point to Python 2.
+  new_path = []
```
oliverchang: nit: just do

```python
new_path = [
  os.path.join(unzipped, 'peach_mutator', 'src'),
  os.path.join(unzipped, 'peach_mutator', 'third_party', 'peach')
] + sys.path
```

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1589,6 +1592,52 @@ def use_radamsa_mutator_plugin(extra_env):
...
+  new_path.append(os.path.join(unzipped, 'peach_mutator', 'src'))
+  new_path.append(
+      os.path.join(unzipped, 'peach_mutator', 'third_party', 'peach'))
+
+  # Append the mutator to the begining of the Python path.
+  new_path = new_path + sys.path
+  extra_env['PYTHONPATH'] = ':'.join(new_path)
```
oliverchang: us os.pathsep here.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1589,6 +1592,52 @@ def use_radamsa_mutator_plugin(extra_env):
...
+  """Decide whether or not to use peach mutator, and set up all of the
+  environment variables necessary to do so."""
+  # TODO(mpherman): Include architecture info in job definition and exclude
+  #  i386.
+  if environment.platform() != 'LINUX' or environment.get_value(
+      'MEMORY_TOOL') != 'ASAN':
+    return False
+
+  # Select grammar.
+  grammar = fuzzer_utils.extract_argument(arguments, 'grammar')
```
oliverchang: How is the grammar piped to here? I.e. how is the grammar specified for a target such that at this point there is an argument here?

Also, we shouldn't be using `arguments` for these (the usage seems wrong anyway, as we need to specify the argument of the format `-flag=`). That should be libFuzzer-specific arguments. If we want to add additional metadata, we can add a field to StrategyInfo.

src/python/bot/fuzzers/libfuzzer.py
mpherman2: Done

src/python/bot/fuzzers/libfuzzer.py
mpherman2: Done

src/python/bot/fuzzers/libfuzzer.py
mpherman2: Done

src/python/bot/fuzzers/libfuzzer.py
mpherman2: We add this "grammar=..." in the libfuzzer options. This it how it ends up in the arguments. Is this not how we should do it?

oliverchang: IMO we should create a new section in .options for grammar specific stuff, or even an entirely new file like TARGET.grammar or something similar.

We shouldn't combine libFuzzer specific and non-libFuzzer specific options under the one `arguments`, as that creates confusion.

@mbarbella-chromium WDYT?

oliverchang: Another reason not to do this is if we don't choose peach strategy, this argument will be passed to the libFuzzer target, which is wrong and can potentially break things.

Again, we can avoid all of this confusion if we pipe it through via a different mechanism.

mbarbella-chromium: Good point. We definitely don't want to be passing this along if we aren't picking this strategy. Having the separate .grammar file sounds like the cleanest solution to me.

src/python/bot/fuzzers/libFuzzer/peach/pits.py
```diff
@@ -0,0 +1,38 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Dictionary to keep track of pit information."""
+
+from metrics import logs
+from system import environment
+import os
```
Dor1s: nit: move this system import above the local ones

src/python/bot/fuzzers/libFuzzer/peach/pits.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def get_path(grammar):
+  """Return the path of the peach pit for the given grammar. Return None if the
+  Pit does not exist or the grammar is None."""
```
Dor1s: do you anticipate it ever being called with an empty `grammar`? I'd rather make sure on the caller side that the grammar is always non-empty and then log an error here if it ever gets call with an empty argument. Otherwise, all these cases will go silent and may hide some other issue from us

src/python/bot/fuzzers/libFuzzer/peach/pits.py
```diff
@@ -0,0 +1,38 @@
...
+  # If the target has no grammar, do not use this strategy.
+  if not grammar:
+    return None
+
+  pit_dir = os.path.join(environment.get_platform_resources_directory(),
+                         'peach', 'pits')
+  pit_path = os.path.join(pit_dir, grammar + '.xml')
+
+  if not os.path.exists(pit_path):
+    logs.log_error('Peach Mutator is being used with the grammar named ' +
```
Dor1s: nit: I'd write `'Pit file for "%s" grammar is not found.' % grammar`

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1589,6 +1592,50 @@ def use_radamsa_mutator_plugin(extra_env):
   return True
 
 
+def use_peach_mutator(extra_env, arguments):
+  """Decide whether or not to use peach mutator, and set up all of the
+  environment variables necessary to do so."""
+  # TODO(mpherman): Include architecture info in job definition and exclude
+  #  i386.
```
Dor1s: super nit: extra space after `#`

Dor1s: seems to be copied from 1582, please fix there too :)

src/python/bot/fuzzers/libfuzzer.py
Dor1s: why do we restrict peach mutator to this configuration only?

src/python/metrics/fuzzer_stats_schema.py
```diff
@@ -363,6 +363,10 @@
     'mode': 'NULLABLE',
     'name': 'strategy_mutator_plugin_radamsa',
     'type': 'INTEGER'
+}, {
+    'mode': 'NULLABLE',
+    'name': 'strategy_peach_grammar_mutation',
+    'type': 'INTEGER'
```
Dor1s: would it make sense to record name of the grammar used?

mbarbella-chromium: +1, that makes more sense as the value here.

mpherman2:
PTAL. I pushed a first draft of clusterfuzz side changes to transition to a grammar file instead of using the options file. The code still needs some unit tests, and I would like to run it locally as well, but I wanted to see if anyone had any high level comments.

src/python/bot/fuzzers/options.py
```diff
@@ -157,22 +158,59 @@ def get_ubsan_options(self):
...
+    return None
+
+  config = configparser.ConfigParser()
+  with open(grammar_file_path, 'r') as f:
+    try:
+      config.read_file(f)
+    except configparser.Error:
+      raise FuzzerOptionsException('Failed to parse fuzzer grammar file.')
+
+  if not config.has_section('grammar'):
```
oliverchang: this seems fine, but any reason why we want to complicate this with a config format, rather than say a plain string? Do we anticipate that we need to attach more metadata here?

src/python/bot/fuzzers/libfuzzer.py
oliverchang: Looks like you updated lines 1582-1584, but not here. We should move this into a shared function and call it from both places.

mpherman2: Done

src/python/bot/fuzzers/options.py
mpherman2: Not presently, but it is likely in the future it will be helpful to have a place to put more grammar metadata. All of the infrastructure was there already anyway, so might as well make our lives easier in the future.

src/python/tests/core/bot/fuzzers/libFuzzer/engine_test.py
```diff
@@ -344,6 +344,7 @@ def mock_merge(*args, **kwargs):  # pylint: disable=unused-argument
         'strategy_fork': 0,
         'strategy_mutator_plugin': 0,
         'strategy_mutator_plugin_radamsa': 0,
+        'strategy_peach_grammar_mutation': 0,
```
oliverchang: The type defined in fuzzer_stats_schema is a STRING, while this is an int?

src/python/fuzzing/strategy.py
```diff
@@ -72,6 +76,7 @@
 LIBFUZZER_STRATEGIES_WITH_PREFIX_VALUE = [
     CORPUS_SUBSET_STRATEGY,
     FORK_STRATEGY,
+    PEACH_GRAMMAR_MUTATION_STRATEGY,
```
oliverchang: I don't think this will work as expected unfortunately. 

This (https://github.com/google/clusterfuzz/blob/5e9b02204428cb1a81993d512844966d2bee0a80/src/python/bot/fuzzers/libFuzzer/stats.py#L148) assumes that the part after the prefix is an int, which is not the case here.

We need to do some refactoring to support this. While the ideal refactor is a bit complicated (I can do this later), to unblock you you can probably just refactor LIBFUZZER_STRATEGIES_WITH_PREFIX_VALUE such that it's a dict where the values indicate the type.

src/python/tests/core/bot/fuzzers/libFuzzer/engine_test.py
mpherman2: I am not sure where this value is being set. It seems to be 0 by default. Do you know what I need to change to make this the tag?

src/python/fuzzing/strategy.py
mpherman2: Changing LIBFUZZER_STRATEGIES_WITH_PREFIX_VALUE to a dict is not going to work because all that is available is the strategy.name, not the strategy itself. Instead I can make another dict that maps strategy.name -> type. Is that sufficient?

oliverchang: That also works.

src/python/tests/core/bot/fuzzers/libFuzzer/engine_test.py
oliverchang: Most likely

https://github.com/google/clusterfuzz/blob/5e9b02204428cb1a81993d512844966d2bee0a80/src/python/bot/fuzzers/libFuzzer/stats.py#L205

src/python/fuzzing/strategy.py
mpherman2: Done

src/python/bot/fuzzers/libFuzzer/stats.py
```diff
@@ -202,7 +203,10 @@ def parse_performance_features(log_lines, strategies, arguments):
 
   # Initialize all strategy stats as disabled by default.
   for strategy_type in strategy.LIBFUZZER_STRATEGY_LIST:
-    stats[strategy_column_name(strategy_type.name)] = 0
+    if strategy_type.name == 'peach_grammar_mutation':
```
oliverchang: This is a bit too hacky.  Let's use STRATEGIES_WITH_PREFIX_VALUE_TYPE here to decide the default value.

mpherman2: Let me know if this is what you had in mind. The other way I considered doing it was to store the default value of the prefix in the STRATEGIES_WITH_PREFIX_VALUE_TYPE dict as well. Which would you prefer?

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1589,6 +1597,50 @@ def use_radamsa_mutator_plugin(extra_env):
...
+
+  # Set title and pit environment variables
+  extra_env['PIT_PATH'] = pit_path
+  extra_env['PIT_TITLE'] = grammar
+
+  # Extract zip of peach mutator code.
+  peach_dir = os.path.join(environment.get_platform_resources_directory(),
+                           'peach')
+  unzipped = os.path.join(peach_dir, 'mutator')
+  source = os.path.join(peach_dir, 'peach_mutator.zip')
```
Dor1s: why do we keep it in zip and extract every time?

src/python/bot/fuzzers/options.py
```diff
@@ -157,22 +158,59 @@ def get_ubsan_options(self):
     return self._get_option_section('ubsan')
 
 
+def get_fuzz_target_support_file(fuzz_target_path, file_extension):
+  """Helper function. returns a file path for a fuzz target support file
```
Dor1s: nit: capitalize "r" in "returns"

Dor1s: also can skip "Helper function. "

src/python/bot/fuzzers/options.py
```diff
@@ -157,22 +158,59 @@ def get_ubsan_options(self):
...
+  that has the given file extension"""
```
Dor1s: and add trailing period

src/python/fuzzing/strategy.py
```diff
@@ -72,8 +77,15 @@
...
 ]
 
+STRATEGIES_WITH_PREFIX_VALUE_TYPE = {
```
Dor1s: nit: append `LIBFUZZER_` to the variable name for consistency

src/python/bot/fuzzers/options.py
```diff
@@ -157,22 +158,59 @@ def get_ubsan_options(self):
...
+
   try:
     return FuzzerOptions(options_file_path, cwd=options_cwd)
   except FuzzerOptionsException:
     logs.log_error('Invalid options file: %s.' % options_file_path)
     return None
+
+
+def get_fuzz_target_grammar(fuzz_target_path):
+  """Return the grammra for a given target or None if it does not exist"""
```
Dor1s: typo in `grammra` and missing trailing period

src/python/bot/fuzzers/options.py
Dor1s: yeah, this confused me in Chromium side CL as well, I don't see why we need a `.grammar` file at all if we just need to pass an option that goes into a separate section of the `.option` file

src/python/bot/fuzzers/libfuzzer.py
Dor1s: I frankly prefer having a new section in the `.options` file instead.
- We already have there support for libFuzzer, AFL, env variables and various sanitizer options.
- That would also keeps things a bit more simple for OSS-Fuzz, as some projects use `.options` files.
- On Chromium side we will re-use the same script for generating `.options` file instead of duplicating it.

mpherman2: @mbarbella-chromium @oliverchang Thoughts?

mbarbella-chromium: A new section seems fine to me as well.

Would this be the first time we're using a new section for something other than arguments? If so, what happens if a header is omitted? I just want to confirm that we won't suddenly have issues if any OSS-Fuzz targets had improperly formed options files, but I suspect the risk would be very low.

src/python/bot/fuzzers/libfuzzer.py
mpherman2: Because the peach code has to be python 2, and we don't want python 2 code in the repo. @mbarbella-chromium and I figure this would be the best way around that.

src/python/bot/fuzzers/options.py
mpherman2: Done

src/python/bot/fuzzers/options.py
mpherman2: Done

src/python/bot/fuzzers/options.py
mpherman2: Done

src/python/fuzzing/strategy.py
mpherman2: Done

src/python/bot/fuzzers/libfuzzer.py
Dor1s: > Would this be the first time we're using a new section for something other than arguments?

No, only `[libfuzzer]` and `[AFL]` sections' values go into arguments.

`[env]` elements go into `VARIABLE=VALUE` env vars

`[asan]` and other sanitizers sections go into `ASAN_OPTIONS=VARIABLE=VALUE:...` and such

I think these are all different use cases, and having something new isn't harmful.

> If so, what happens if a header is omitted? 

what do you mean by "header"?

mpherman2:
I changed getting the grammar to use an environment variable. We should just be able to add LIBFUZZER_GRAMMAR to the environment variables section of the options file without any extra changes needed to chromium. PTAL.

src/python/bot/fuzzers/options.py
```diff
@@ -31,6 +31,7 @@
 from system import environment
 
 OPTIONS_FILE_EXTENSION = '.options'
+GRAMMAR_FILE_EXTENSION = '.grammar'
```
mbarbella-chromium: This should be unused now, right?

src/python/bot/fuzzers/options.py
```diff
@@ -157,20 +158,32 @@ def get_ubsan_options(self):
     return self._get_option_section('ubsan')
 
 
+def get_fuzz_target_support_file(fuzz_target_path, file_extension):
```
mbarbella-chromium: It's probably best to inline this back into the other function. I suspect we won't be adding new files later based on the discussions in this PR.

src/python/bot/fuzzers/options.py
mpherman2: Yes. Removed

src/python/bot/fuzzers/options.py
mpherman2: Done

src/python/bot/fuzzers/libFuzzer/peach/pits.py
```diff
@@ -0,0 +1,34 @@
...
+def get_path(grammar):
+  """Return the path of the peach pit for the given grammar. Return None if the
+  Pit does not exist or the grammar is None."""
+
+  pit_dir = os.path.join(environment.get_platform_resources_directory(),
+                         'peach', 'pits')
+  pit_path = os.path.join(pit_dir, grammar + '.xml')
+
+  if not os.path.exists(pit_path):
+    logs.log_error('Pit file for "%s" grammar is not found.' % grammar)
```
Dor1s: can you also log `pit_path`? I think the following should work:

```python
  if not os.path.exists(pit_path):
    logs.log_error('Pit file for "%s" grammar is not found.' % grammar,
                   pit_path=pit_path)
```

and the value will be included in the log metadata

src/python/bot/fuzzers/libFuzzer/stats.py
```diff
@@ -144,8 +144,10 @@ def parse_line_for_strategy_prefix(line, strategy_name):
     if not line.startswith(strategy_prefix):
       return
 
+    prefix_type = strategy.LIBFUZZER_STRATEGIES_WITH_PREFIX_VALUE_TYPE[
```
Dor1s: it's `suffix_type` I believe, that prefix is always a string like `corpus_subset_` followed by a value we parse here

src/python/bot/fuzzers/libFuzzer/stats.py
```diff
@@ -202,7 +204,13 @@ def parse_performance_features(log_lines, strategies, arguments):
 
   # Initialize all strategy stats as disabled by default.
   for strategy_type in strategy.LIBFUZZER_STRATEGY_LIST:
-    stats[strategy_column_name(strategy_type.name)] = 0
+    if (strategy_type.name in
+        strategy.LIBFUZZER_STRATEGIES_WITH_PREFIX_VALUE_TYPE and
+        strategy.LIBFUZZER_STRATEGIES_WITH_PREFIX_VALUE_TYPE[strategy_type.name]
+        == str):
+      stats[strategy_column_name(strategy_type.name)] = 'None'
```
Dor1s: @oliverchang is there any different whether we use `'None'` or an empty string, WDYT? I definitely don't want `None` (aka `NULL`) as that should indicate that strategy didn't existed at the time of a particular run

src/python/bot/fuzzers/libfuzzer.py
Dor1s: ack

src/python/bot/fuzzers/libFuzzer/peach/pits.py
mpherman2: Done

src/python/bot/fuzzers/libFuzzer/stats.py
mpherman2: Done

src/python/bot/fuzzers/libFuzzer/stats.py
oliverchang: Good point, empty string may be better here. 'None' shouldn't result in a NULL value, but empty ('') is still cleaner as the "default" value for a string.

src/python/bot/fuzzers/options.py
```diff
@@ -160,19 +160,22 @@ def get_ubsan_options(self):
 def get_fuzz_target_options(fuzz_target_path):
   """Return a FuzzerOptions for the given target, or None if it does not
   exist."""
-  options_file_path = fuzzer_utils.get_supporting_file(fuzz_target_path,
-                                                       OPTIONS_FILE_EXTENSION)
-  options_cwd = os.path.dirname(options_file_path)
+  file_path = fuzzer_utils.get_supporting_file(fuzz_target_path,
```
oliverchang: What's the reason for the changes here? It seems like we're just effectively renaming a variable? I prefer the old one tbh.

src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -101,12 +101,13 @@ def prepare(self, corpus_dir, target_path, _):
       A FuzzOptions object.
     """
     arguments = fuzzer.get_arguments(target_path)
+    grammar = environment.get_value('LIBFUZZER_GRAMMAR', None)
```
oliverchang: Why was this changed to an env var? 

There are a number of issues with this:
1. This doesn't actually propagated anywhere for libFuzzer. There is a [whitelist](https://github.com/google/clusterfuzz/blob/5e9b02204428cb1a81993d512844966d2bee0a80/src/python/bot/fuzzers/options.py#L36), and the propagation currently only happens for [AFL](https://github.com/google/clusterfuzz/search?q=get_env&unscoped_q=get_env).

2. Using env vars for this seems a bit hacky -- env vars are very general, pollute the namespace of multiple processes, and we should avoid adding any more to our (already very large list) if we can.. An explicit section in the options file (or the .grammar) is better.

src/python/bot/fuzzers/libFuzzer/stats.py
```diff
@@ -202,7 +204,13 @@ def parse_performance_features(log_lines, strategies, arguments):
 
   # Initialize all strategy stats as disabled by default.
   for strategy_type in strategy.LIBFUZZER_STRATEGY_LIST:
-    stats[strategy_column_name(strategy_type.name)] = 0
+    if (strategy_type.name in
+        strategy.LIBFUZZER_STRATEGIES_WITH_PREFIX_VALUE_TYPE and
+        strategy.LIBFUZZER_STRATEGIES_WITH_PREFIX_VALUE_TYPE[strategy_type.name]
```
oliverchang: nit: just do

```python
strategy.LIBFUZZER_STRATEGIES_WITH_PREFIX_VALUE_TYPE.get(strategy_type.name) == str
```

To avoid the "in" check.

src/python/bot/fuzzers/libFuzzer/engine.py
mpherman2: We thought that it would get piped in from the environment variables section of the libfuzzer options file without any need for a change on the chromium side. These are good points though, so now it works with a separate grammar section in the libfuzzer options file.

src/python/bot/fuzzers/options.py
```diff
@@ -171,6 +174,11 @@ def get_fuzz_target_options(fuzz_target_path):
   if not os.path.exists(options_file_path):
     return None
 
+  if not options_file_path:
```
oliverchang: nit: move this to immediately after line 167-168.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1708,6 +1763,14 @@ def pick_strategies(strategy_pool, fuzzer_path, corpus_directory,
     fuzzing_strategies.append(strategy.MUTATOR_PLUGIN_STRATEGY.name)
 
   if (strategy.MUTATOR_PLUGIN_STRATEGY.name not in fuzzing_strategies and
+      strategy_pool.do_strategy(strategy.PEACH_GRAMMAR_MUTATION_STRATEGY) and
```
oliverchang: nit: this is getting a bit difficult to read and will get even more so if we add more.

maybe do something like

```python
MUTATOR_STRATAGIES = [
  strategy.PEACH_GRAMMAR_MUTATION_STRATEGY.name,
  strategy.MUTATOR_PLUGIN_STRATEGY.name,
  strategy.MUTATOR_PLUGIN_RADAMSA_STRATEGY.name,
]

def has_existing_mutator_strategy(fuzzing_strategies):
  return any(strategy in fuzzing_strategies for strategy in MUTATOR_STRATEGIES)

...

if (not has_existing_mutator_strategy(fuzzing_strategies) and 
    strategy_pool.do_strategy(..)):

...

src/python/bot/fuzzers/options.py
mpherman2: Done

src/python/bot/fuzzers/libfuzzer.py
mpherman2: Done

PR #1481, opened 2020-02-25
Some fixes for untrusted tests on Python 3.

oliverchang:
- Move proto import paths to make them work under Python 3.
- Various bytes/str and other minor fixes.
- Convert engine RPCs to use string stacktraces instead of bytes


src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -488,10 +489,11 @@ def minimize_corpus(self, target_path, arguments, input_dirs, output_dir,
     if result.return_code != 0:
       raise MergeError('Merging new testcases failed: ' + result.output)
 
-    merge_stats = stats.parse_stats_from_merge_log(result.output.splitlines())
+    merge_output = utils.decode_to_unicode(result.output)
```
inferno-chromium: I keep thinking if we should decode_to_unicode the output in all cases (in run_and_wait ?), so that we don't forget to explicitly do it, thoughts ?

oliverchang: Not sure if that makes sense more generally, since processes can return non-unicode/text content. For libFuzzer/engines, it makes sense because that's what we expect.

oliverchang: I think it may make sense for process_handler.run_process, since that's mostly used for running blackbox targets which should output text content.

Issue #1477, opened 2020-02-21
Fuzzing Strategies not being added to Testcase Metadata


mpherman2:
In fuzz_task.py it appears as if we are attempting to add the fuzzing strategies to testcase metadata: 
https://github.com/google/clusterfuzz/blob/2c9b41570dbcfca3e8bbf1266b1b8ed5e2cef7e4/src/python/bot/tasks/fuzz_task.py#L992-L1003

When we do this though, we try to find the strategies by parsing through crash.stacktrace. We believe the problem is that crash.stacktrace does not have the fuzzing strategy info. This is contained in the logs.

We need to either parse through the logs, or add the fuzzing strategy info to the crash stacktrace.

Dor1s:
Thanks @mpherman2 !

> We need to either parse through the logs, or add the fuzzing strategy info to the crash stacktrace.

By the logs we mean this:

https://github.com/google/clusterfuzz/blob/fb4d1aedce2fb0d93c2d6e5cb280f6f187055101/src/python/bot/tasks/fuzz_task.py#L1303-L1306

PR #1486, opened 2020-02-26
Fix missing strategy info in testcase metadata

mpherman2:
Not sure if this is the cleanest way to send the result logs to the crash. Let me know what you think.

### DIFF
--- a/src/python/bot/tasks/fuzz_task.py
+++ b/src/python/bot/tasks/fuzz_task.py
@@ -996,7 +996,8 @@ def create_testcase(group, context):
     testcase.put()
 
   fuzzing_strategies = (
-      libfuzzer_stats.LIBFUZZER_FUZZING_STRATEGIES.search(crash.fuzzing_log))
+      libfuzzer_stats.LIBFUZZER_FUZZING_STRATEGIES.search(
+          crash.fuzzing_log if crash.fuzzing_log else ''))
 
   if fuzzing_strategies:
     assert len(fuzzing_strategies.groups()) == 1
--- a/src/python/bot/tasks/fuzz_task.py
+++ b/src/python/bot/tasks/fuzz_task.py
@@ -202,6 +202,12 @@ class Crash(object):
     self.crash_frames = state.frames
     self.crash_info = None
 
+    # Fuzzing log containing strategy is stored here.
+    self.fuzzing_log = None
+
+  def set_fuzzing_log(self, log):
+    self.fuzzing_log = log
+
   def is_archived(self):
     """Return true if archive_testcase_in_blobstore(..) was
       performed."""
@@ -990,8 +996,7 @@ def create_testcase(group, context):
     testcase.put()
 
   fuzzing_strategies = (
-      libfuzzer_stats.LIBFUZZER_FUZZING_STRATEGIES.search(
-          crash.crash_stacktrace))
+      libfuzzer_stats.LIBFUZZER_FUZZING_STRATEGIES.search(crash.fuzzing_log))
 
   if fuzzing_strategies:
     assert len(fuzzing_strategies.groups()) == 1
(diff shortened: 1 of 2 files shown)


src/python/tests/core/bot/tasks/fuzz_task_test.py
```diff
@@ -790,6 +787,10 @@ def test_create_many_groups(self, project_name):
         self._make_crash('u4', state='unreproducible3')
     ]
 
+    # TODO(metzman): Add a seperate test for strategies.
```
mbarbella-chromium: Instead of moving this, could you add a separate test to ensure that we're able to get the strategy information from the log correctly?

src/python/bot/tasks/fuzz_task.py
```diff
@@ -991,7 +997,7 @@ def create_testcase(group, context):
 
   fuzzing_strategies = (
       libfuzzer_stats.LIBFUZZER_FUZZING_STRATEGIES.search(
-          crash.crash_stacktrace))
+          crash.fuzzing_log if crash.fuzzing_log else ''))
```
mbarbella-chromium: Instead of doing this here, could you just initialize fuzzing_log to ''?

mpherman2:
@oliverchang This change is trying to solve this problem: https://github.com/google/clusterfuzz/issues/1477. Do you think this is the best way to go about it, or is there a better way?

src/python/bot/tasks/fuzz_task.py
```diff
@@ -202,6 +202,12 @@ def __init__(self,
     self.crash_frames = state.frames
     self.crash_info = None
 
+    # Fuzzing log containing strategy is stored here.
+    self.fuzzing_log = ''
```
oliverchang: Sorry for the delayed review! I don't think this is the right way to fix this. See my other comments.

src/python/bot/tasks/fuzz_task.py
```diff
@@ -990,8 +996,7 @@ def create_testcase(group, context):
     testcase.put()
 
   fuzzing_strategies = (
```
oliverchang: So this cf::fuzzing_strategies thing is a holdout from our old (deprecated) way of piping strategies. We still need this support this for the time being until AFL is migrated. In the new pipeline, we already have strategies available without needing to parse them. 

Let's move this to the `from_testcase_manager_crash` constructor (which is what the deprecated way is using), and make `fuzzing_strategies` an instance variable of `Crash`. Then, we can do something like this:

```python
  
def run_engine_fuzzer(...):
   ...
   return result, fuzzer_metadata, options.strategies

def do_engine_fuzzing(...):
  ... pipe strategies from run_engine_fuzzer and return here...
   
  @classmethod
  def from_testcase_manager_crash(...):
    fuzzing_strategies = (
        libfuzzer_stats.LIBFUZZER_FUZZING_STRATEGIES.search(orig_unsymbolized_crash_stacktrace))
    return Crash( 
        ...
        fuzzing_strategies=fuzzing_strategies)

  def from_engine_crash(cls, crash, fuzzing_strategies):
    return Crash(
      ...
      fuzzing_strategies=fuzzing_strategies)

  # Crash class
  def __init__(self,
               file_path,
               crash_time,
               return_code,
               resource_list,
               gestures,
               unsymbolized_crash_stacktrace,
               arguments,
               application_command_line,
               http_flag=False,
               fuzzing_strategies=None):
  ...
  self.fuzzing_strategies = fuzzing_strategies
```

Does this make sense?

oliverchang:
Since what I proposed is a bit complicated, I uploaded https://github.com/google/clusterfuzz/pull/1487 to fix this :)

oliverchang:
Closing as this is superseded by #1487

Dor1s:
Thanks a lot, @oliverchang !

PR #1488, opened 2020-02-29
Don't error when there are no stats.

inferno-chromium:


inferno-chromium:
TBR

src/appengine/handlers/fuzzer_stats.py
```diff
@@ -525,18 +525,20 @@ def get(self):
...
       if not job_filter:
         # Group by job only makes sense for queries that do not specify job.
         group_by = 'by-job'
         try:
           build_results(fuzzer, job_filter, group_by, date_start, date_end)
-        except Exception:
-          logs.log_error('Failed to preload %s %s %s %s %s.' %
-                         (fuzzer, job_filter, group_by, date_start, date_end))
+        except Exception as e:
+          if 'No stats.' not in repr(e):
```
mbarbella-chromium: This feels a bit hacky as-is. Is there a specific exception type that could be caught for the no stats case? If not, a comment explaining why this is safe would be helpful.

inferno-chromium: i think this gets shown to user in message, so uses generic exception is used

  if 'rows' not in results:
    raise helpers.EarlyExitException('No stats.', 404)

PR #1489, opened 2020-03-02
Use mmap for utils.search_bytes_in_file.

oliverchang:
Renamed from utils.search_strings_in_file, and use mmap instead of attempting
to read the file line by line.

This also fixes a Python 3 exception.

### DIFF
--- a/src/python/tests/core/bot/fuzzers/utils_test.py
+++ b/src/python/tests/core/bot/fuzzers/utils_test.py
@@ -97,3 +97,10 @@ class IsFuzzTargetLocalTest(unittest.TestCase):
     path = self._create_file(
         'a_fuzzer', contents=b'anything\nLLVMFuzzerTestOneInput')
     self.assertTrue(utils.is_fuzz_target_local(path))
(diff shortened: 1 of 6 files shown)


oliverchang:
(This is a bit of a risky change, so will test carefully before submitting).

src/python/tests/core/bot/fuzzers/utils_test.py
```diff
@@ -12,73 +12,88 @@
...
+    shutil.rmtree(self.temp_dir, ignore_errors=True)
+
+  def _create_file(self, name, contents=b''):
+    path = os.path.join(self.temp_dir, name)
+    with open(path, 'wb') as f:
+      f.write(contents)
+
+    return path
 
   def test_not_a_fuzzer_invalid_name(self):
```
inferno-chromium: Can you add a test where a file handle is passed (and not a string path)?

oliverchang: Done!

PR #1490, opened 2020-03-02
Fix untrusted minimize tests.

oliverchang:


oliverchang:
trivial change. TBR /gcbrun

PR #1493, opened 2020-03-02
search_bytes_in_file: Don't mmap if file handle size is 0.

oliverchang:
Windows doesn't allow mmaping empty files.


oliverchang:
Actually, reverting original CL since it seems it doesn't work in all cases (i.e. when the "file handle" comes from the archive for partial extraction).

PR #1494, opened 2020-03-02
Revert "Use mmap for utils.search_bytes_in_file."

oliverchang:
`Reverts` google/clusterfuzz#1489. Doesn't work for archive file handles.


inferno-chromium:
LGTM

PR #1495, opened 2020-03-02
Run bigquery queries in cleanup only for project's platforms.

inferno-chromium:


inferno-chromium:
Causing 3x more queries for google instance [3 platforms] for every project. Should shorten cron.

src/appengine/handlers/cron/cleanup.py
```diff
@@ -178,12 +181,11 @@ def cleanup_unused_fuzz_targets_and_jobs():
   ndb_utils.delete_multi(to_delete)
 
 
-def get_jobs_and_platforms_for_top_crashes():
-  """Return list of jobs and platforms to use for picking top crashes."""
-  jobs = set()
-  platforms = set()
-
+def get_projects_to_jobs_and_platforms_map_for_top_crashes():
```
oliverchang: nit: this is a little long and difficult to comprehend.

Maybe rename to `get_jobs_and_platforms_for_project`. The "top_crashes" part isn't really necessary context that needs to be included here in the name.

src/appengine/handlers/cron/cleanup.py
```diff
@@ -178,12 +181,11 @@ def cleanup_unused_fuzz_targets_and_jobs():
...
+  """Return a map of projects to jobs and platforms map to use for picking top
+  crashes."""
   all_jobs = ndb_utils.get_all_from_model(data_types.Job)
+  projects_to_jobs_and_platforms_map = {}
```
oliverchang: nit: omit '_map' for brevity?

src/appengine/handlers/cron/cleanup.py
inferno-chromium: Done!

src/appengine/handlers/cron/cleanup.py
inferno-chromium: Done!

PR #1496, opened 2020-03-02
Fix encoding issues with utils.search_bytes_in_file.

oliverchang:
Renamed from search_string_in_file.

### DIFF
--- a/src/python/base/utils.py
+++ b/src/python/base/utils.py
@@ -626,15 +626,21 @@ def restart_machine():
     os.system('sudo shutdown -r now')
 
 
-def search_string_in_file(search_string, file_handle):
-  """Helper to search for a string in a large binary file without memory
+def search_bytes_in_file(search_bytes, file_handle):
+  """Helper to search for bytes in a large binary file without memory
   issues.
   """
-  # TODO(aarya): This is too brittle and will fail if we have a very large line.
-  search_string = search_string.encode('utf-8')
-  for line in file_handle:
-    if search_string in line:
-      return True
+  old_pos = file_handle.tell()
+  file_handle.seek(0, os.SEEK_SET)
+
+  try:
+    # TODO(aarya): This is too brittle and will fail if we have a very large
+    # line.
+    for line in file_handle:
+      if search_bytes in line:
+        return True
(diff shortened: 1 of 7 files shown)


src/python/base/utils.py
```diff
@@ -626,15 +626,21 @@ def restart_machine():
...
+  old_pos = file_handle.tell()
+  file_handle.seek(0, os.SEEK_SET)
+
+  try:
+    # TODO(aarya): This is too brittle and will fail if we have a very large
+    # line.
+    for line in file_handle:
+      if search_bytes in line:
+        return True
+  finally:
```
inferno-chromium: do we need to log_error or log_warn on failures ?

src/python/base/utils.py
```diff
@@ -626,15 +626,21 @@ def restart_machine():
...
+def search_bytes_in_file(search_bytes, file_handle):
+  """Helper to search for bytes in a large binary file without memory
   issues.
   """
-  # TODO(aarya): This is too brittle and will fail if we have a very large line.
-  search_string = search_string.encode('utf-8')
-  for line in file_handle:
-    if search_string in line:
-      return True
+  old_pos = file_handle.tell()
```
inferno-chromium: Do we have usecase where file_handle position is not zero. anyway good to do this.

src/python/base/utils.py
oliverchang: The finally isn't for exception handling. It's just to make sure we reset the location.

src/python/base/utils.py
oliverchang: Not that I know of. I'll leave this here just in case.

inferno-chromium:
I think this is causing

```
UnsupportedOperation: seek
at search_bytes_in_file (/mnt/scratch0/clusterfuzz/src/python/base/utils.py:633)
at is_fuzz_target_local (/mnt/scratch0/clusterfuzz/src/python/bot/fuzzers/utils.py:69)
at _get_fuzz_targets_from_archive (/mnt/scratch0/clusterfuzz/src/python/build_management/build_manager.py:577)
```
on chromium instance.

PR #1497, opened 2020-03-03
More untrusted runner fixes.

oliverchang:
Also make all fuzz task tests pass while we're here.


oliverchang:
TBR for trivial fixes.

inferno-chromium:
LGTM

PR #1500, opened 2020-03-04
Fix honggfuzz for Python 3.

oliverchang:
- Introduce UnicodeProcessRunner, which always returns unicode decoded
  output.

### DIFF
--- a/src/python/system/new_process.py
+++ b/src/python/system/new_process.py
@@ -359,3 +359,15 @@ class ProcessRunner(object):
     result.command = process.command
 
     return result
+
+
+class UnicodeProcessRunner(ProcessRunner):
+  """ProcessRunner which always returns unicode output."""
+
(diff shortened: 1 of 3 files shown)


src/python/system/new_process.py
```diff
@@ -359,3 +359,15 @@ def run_and_wait(self,
     result.command = process.command
 
     return result
+
+
+class UnicodeProcessRunner(ProcessRunner):
```
oliverchang: Will migrate libFuzzer in another PR, as that change is a bit scarier.

oliverchang:
> So, we cannot return unicode in all cases in ProcessRunner for output ? Too scary ?

I don't think that's the best interface for a generic ProcessRunner, as we shouldn't assume that all processes return text/utf-8 output.

PR #1501, opened 2020-03-04
Fix testcase manager on Python 3.

oliverchang:


src/python/bot/testcase_manager.py
```diff
@@ -301,7 +301,7 @@ def get_additional_command_line_flags(testcase_path):
   flags_file_content = utils.read_data_from_file(
       flags_file_path, eval_data=False)
   if flags_file_content:
-    additional_command_line_flags += ' %s' % flags_file_content
```
oliverchang: '%s' is dangerous in Python 3, and results in `b''` being included in the result.

PR #1514, opened 2020-03-05
Refactor to allow cross pollination stats in oss fuzz

mpherman2:


src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -682,21 +686,26 @@ def do_corpus_pruning(context, last_execution_failed, revision):
...
+  if not pruner_stats or not pollinator_stats:
+    cross_pollination_stats = None
+  else:
+    cross_pollination_stats = CrossPollinationStats(
+        project_qualified_name, context.cross_pollination_method, sources,
+        context.tag, initial_corpus_size, pre_pollination_corpus_size,
+        pruner_stats['edge_coverage'], pollinator_stats['edge_coverage'],
+        pruner_stats['feature_coverage'], pollinator_stats['feature_coverage'])
+
+  result = CorpusPruningResult(
```
oliverchang: nit: no need for result variable now. can just return the CorpusPruningResult directly.

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -537,10 +544,10 @@ def run(self, timeout):
...
 
 
-def record_cross_pollination_stats(
-    pruner_stats, pollinator_stats, project_qualified_name, sources, tag,
-    initial_corpus_size, minimized_corpus_size_units, method):
+def record_cross_pollination_stats(stats):
   """Log stats about cross pollination in BigQuery."""
+  if not stats:
+    return
   # TODO(mpherman): Find a way to collect these stats for OSS Fuzz.
```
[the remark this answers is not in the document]
oliverchang: Also remove the is_untrusted_worker check.

src/python/tests/core/bot/tasks/corpus_pruning_task_test.py
```diff
@@ -499,6 +498,36 @@ def test_prune(self):
...
+              stats.edge_coverage == edge_coverage and
+              stats.initial_feature_coverage == initial_feature_coverage and
+              stats.feature_coverage == feature_coverage)
+
+    return compare
+
+  def test_untrusted_cross_pollination_stats(self):
+    """Test that the untrusted runner gets correct stats to BigQuery"""
+    self.mock.record_cross_pollination_stats.side_effect = (
+        self.get_mock_record_compare('test_fuzzer', 'random', 'test2_fuzzer',
```
oliverchang: I don't think this is testing anything as you might expect? .side_effect = Func allows you to mock out a function with another implementation, but in this case you're doing some comparisons which have no effect. Did you mean to make them self.assertEqual() checks instead rather than plain == comparison?

Also, we can just stick the pollination test in the main test_prune in the interests of time as this integration test can take a while to run.

mpherman2: Yea, good catch. I meant to do self.assertEqual() :).

src/protos/untrusted_runner.proto
```diff
@@ -220,12 +220,27 @@ message PruneCorpusResponse {
...
+  optional string method = 2;
+  optional string sources = 3;
+  optional string tags = 4;
+  optional int32 initial_corpus_size = 5;
+  optional int32 corpus_size = 6;
+  optional int32 initial_edge_coverage = 7;
+  optional int32 edge_coverage = 8;
+  optional int32 initial_feature_coverage = 9;
+  optional int32 feature_coverage = 10;
+
```
Dor1s: nit: remove this line

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -537,33 +544,27 @@ def run(self, timeout):
...
 
 
-def record_cross_pollination_stats(
-    pruner_stats, pollinator_stats, project_qualified_name, sources, tag,
-    initial_corpus_size, minimized_corpus_size_units, method):
+def _record_cross_pollination_stats(stats):
   """Log stats about cross pollination in BigQuery."""
-  # TODO(mpherman): Find a way to collect these stats for OSS Fuzz.
-  if environment.is_untrusted_worker():
+  if not stats:
```
Dor1s: is this expected to happen regularly? maybe we should log an error or a warning at least?

src/python/tests/core/bot/tasks/corpus_pruning_task_test.py
```diff
@@ -319,16 +319,15 @@ def setUp(self):
     environment.set_value('JOB_NAME', 'libfuzzer_asan_job')
 
     helpers.patch(self, [
-        'bot.fuzzers.engine.get',
-        'bot.tasks.setup.get_fuzzer_directory',
+        'bot.fuzzers.engine.get', 'bot.tasks.setup.get_fuzzer_directory',
```
Dor1s: nit: did auto-format do that?

src/python/tests/core/bot/tasks/corpus_pruning_task_test.py
```diff
@@ -421,6 +420,10 @@ def tearDown(self):
   def test_prune(self):
     """Test pruning."""
     self._setup_env(job_type='libfuzzer_asan_job')
+    self.mock._record_cross_pollination_stats.side_effect = (
+        self.get_mock_record_compare('test_fuzzer', 'random', 'test2_fuzzer',
```
Dor1s: nit: IMO it would be more readable if you created an object here, like the one in `src/python/bot/untrusted_runner/tasks_impl.py` line 56 and then asserted its fields

src/protos/untrusted_runner.proto
mpherman2: Done

src/python/bot/tasks/corpus_pruning_task.py
mpherman2: This will happen when corpus pruning timeouts or has no corpus and stats could not be collected. These will have already been logged. I added a comment to clarify this case.

src/python/tests/core/bot/tasks/corpus_pruning_task_test.py
mpherman2: I believe so, fixed.

src/python/tests/core/bot/tasks/corpus_pruning_task_test.py
mpherman2: Discussed on chat: Used named arguments to make this more readable.

Dor1s: or just use named arguments

src/python/tests/core/bot/tasks/corpus_pruning_task_test.py
mpherman2: Turns out the linter wants these two to be on the same line. Changed it back

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -682,23 +684,26 @@ def do_corpus_pruning(context, last_execution_failed, revision):
...
       for fuzzer in context.cross_pollinate_fuzzers
   ])
 
-  record_cross_pollination_stats(
-      pruner_stats, pollinator_stats, project_qualified_name, sources,
-      context.tag, pre_pollination_corpus_size, minimized_corpus_size_units,
-      context.cross_pollination_method)
-
-  return result
+  if not pruner_stats or not pollinator_stats:
```
oliverchang: nit: Start your if with the positive case, to avoid a mental double negative when reasoning about the "else".

This can be further simplified to e.g. 

```python

cross_pollination_stats = None
if pruner_stats and pollinator_stats:
  cross_pollination_stats = ...

```

PR #1516, opened 2020-03-06
Fix crash uploader tests on Python 3.

oliverchang:


src/python/chrome/crash_uploader.py
```diff
@@ -354,7 +359,7 @@ def parse_mime_to_crash_report_info(local_minidump_mime_path):
     # Extract from the MIME part the key-value pairs used by report uploading.
     if key_match is not None:
       report_key = key_match.group(1)
-      report_value = mime_part.get_payload()
+      report_value = mime_part.get_payload(decode=True)
```
oliverchang: This is to avoid nonascii characters getting removed in Python 3: https://github.com/python/cpython/blob/da9ee00ffea9120bea4b5f0cbed20f9fb584d2a2/Lib/email/message.py#L267

On Python 2, there should be no functional difference either: https://github.com/python/cpython/blob/249706c1fbce04125d81bd9993e6c010ae30f8e4/Lib/email/message.py#L197

oliverchang:
After this and #1515, only 20 tests remaining fail (excluding AFL integration).

PR #1517, opened 2020-03-07
Exception fixes.

inferno-chromium:
- Fix json.loads exception from an empty file read.
- Wifi connection can be flaky, move to log_warn.
- For android >=O, factory reset command changed.
- Move radamsa mutation failure to warn as it is expected.


inferno-chromium:
TBR, some top exception fixes.

PR #1519, opened 2020-03-08
Fix chromium bad build exceptions.

inferno-chromium:
Don't set FONTCONFIG_SYSROOT before fix revision r635076. Otherwise,
this causes a startup crash and build detected as a bad build (when it is not).

Check failed: InitDefaultFont(). Could not find the default font


inferno-chromium:
TBR top exception fix.

PR #1524, opened 2020-03-09
Cloud IAP Authentication support.

oliverchang:


src/appengine/libs/auth.py
```diff
@@ -54,6 +59,66 @@ def is_current_user_admin():
   return bool(key.get())
 
 
+@memoize.wrap(memoize.FifoInMemory(1))
+def _project_number_from_id(project_id):
+  """Get the project number from project ID."""
+  resource_manager = build('cloudresourcemanager', 'v1')
+  result = resource_manager.projects().get(projectId=project_id).execute()
+  return result['projectNumber']
```
inferno-chromium: nit:maybe add
assert result and 'projectNumber' in result ?

src/appengine/libs/auth.py
```diff
@@ -54,6 +59,66 @@ def is_current_user_admin():
...
+    decoded_jwt = jwt.decode(
+        iap_jwt,
+        key,
+        algorithms=['ES256'],
+        issuer='https://cloud.google.com/iap',
+        audience=expected_audience)
+    return decoded_jwt['sub'], decoded_jwt['email']
+  except (jwt.exceptions.InvalidTokenError,
+          requests.exceptions.RequestException) as e:
+    raise AuthError('JWT assertion decode error ' + str(e))
```
inferno-chromium: error:

src/appengine/libs/auth.py
```diff
@@ -54,6 +59,66 @@ def is_current_user_admin():
...
+
+
+def get_iap_email(current_request):
+  """Get Cloud IAP email."""
+  jwt_assertion = current_request.headers.get('X-Goog-IAP-JWT-Assertion')
+  if not jwt_assertion:
+    return None
+
+  _, email = _validate_iap_jwt(jwt_assertion)
```
inferno-chromium: Since we are only user of _validate_iap_jwt, we can skip returning first argument ?

src/appengine/libs/auth.py
oliverchang: Added a check to raise an error.

src/appengine/libs/auth.py
oliverchang: Done!

src/appengine/libs/auth.py
oliverchang: Done!

urbanenomad:
This is awesome!!!  I have been doing the same code change on my own instance but will switch to using this branch.  I ran into some issue on my changes because it would not provide the IAP session JWT in the x-goog-iap-jwt-assertion cookie.  I was wondering if having the firebase auth still in place affects what cookies and headers are passed to the application?  Also with this code change do I still need to configure the Firebase Authentication?

oliverchang:
> This is awesome!!! I have been doing the same code change on my own instance but will switch to using this branch. I ran into some issue on my changes because it would not provide the IAP session JWT in the x-goog-iap-jwt-assertion cookie. I was wondering if having the firebase auth still in place affects what cookies and headers are passed to the application? Also with this code change do I still need to configure the Firebase Authentication?

Firebase auth shouldn't be affecting the header set by IAP here... Let me know if my change makes things work for you.

You should not need to configure Firebase auth after this.

urbanenomad:
ok...so I am running into issues with my current deployment of clusterfuzz with the updated IAP code.  If I want to create a new cluster without Firebase can I still use the create script?  Do I not need to add the firebase and OAuth keys to the create script?

python butler.py create_config --project-id=$CLOUD_PROJECT_ID $CONFIG_DIR

oliverchang:
Unfortunately for now you will still need the firebase keys to re-run the configuration creation script. What issues are you running into with your current deployment?

urbanenomad:
Ok I turned off firebase in my current instance and now I had to modify the _is_privileged_user(email) in the access.py to always return true.  Otherwise I can't see anything even though I am IAP authenticated.  I have tested that a user without IAP_user role can't get access to the app which is great.  And my access is based on google group rather than individual users in the project.  One thing that I can't seem to get into still is the Configuration page which I am still shown as unauthorized.  Not sure why that is the case.

urbanenomad:
Ok I figured out how to get access to Configuration page.  I just added an entity into the datastore table for Admin kind with key as email address and email property as email address as well.  That allows that specific user Configuration access!!  So now I think this should work for us.  I can allow the clusterfuzz access via IAP and we can add people manually as admins to the configuration page.  Maybe the is_privileged_user method should also use the datastore rather than the db_config.

PR #1529, opened 2020-03-10
Fix all AFL integration tests for Python 3.

oliverchang:
Use UnicodeProcessRunner for AFL runners.

### DIFF
--- a/src/python/system/new_process.py
+++ b/src/python/system/new_process.py
@@ -361,7 +361,19 @@ class ProcessRunner(object):
     return result
 
 
-class UnicodeProcessRunner(ProcessRunner):
+class UnicodeProcessRunnerMixin(object):
+  """Mixin for process runner subclasses to output unicode output."""
+
+  def run_and_wait(self, *args, **kwargs):  # pylint: disable=arguments-differ
+    """Overridden run_and_wait which always decodes the output."""
+    result = ProcessRunner.run_and_wait(self, *args, **kwargs)
+    if result.output is not None:
+      result.output = utils.decode_to_unicode(result.output)
+
+    return result
+
+
+class UnicodeProcessRunner(UnicodeProcessRunnerMixin, ProcessRunner):
   """ProcessRunner which always returns unicode output."""
 
   def run_and_wait(self, *args, **kwargs):  # pylint: disable=arguments-differ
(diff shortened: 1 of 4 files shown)


src/python/tests/core/bot/fuzzers/afl/afl_launcher_integration_test.py
```diff
@@ -364,7 +364,7 @@ def _run_with_sudo(command):
       stdin=subprocess.PIPE,
       stdout=subprocess.PIPE,
       stderr=subprocess.STDOUT)
-  popen.communicate(input=SUDO_PASSWORD + '\n')
+  popen.communicate(input=(SUDO_PASSWORD + '\n').encode())
```
inferno-chromium: does utf-8 need to be specified ?

src/python/system/new_process.py
```diff
@@ -361,7 +361,19 @@ def run_and_wait(self,
...
+    if result.output is not None:
+      result.output = utils.decode_to_unicode(result.output)
+
+    return result
+
+
+class UnicodeProcessRunner(UnicodeProcessRunnerMixin, ProcessRunner):
   """ProcessRunner which always returns unicode output."""
 
   def run_and_wait(self, *args, **kwargs):  # pylint: disable=arguments-differ
```
inferno-chromium: Why did we create UnicodeProcessRunnerMixin ? it has the same run_and_wait as UnicodeProcessRunner ?

oliverchang: Oops, forgot to delete the run_and_wait in the Runner.

This is needed for subclasses of ProcessRunner (e.g. for minijail). We cannot just do e.g.

```python
class MinijailLibFuzzerRunner(UnicodeProcessRunner, MinijailProcessRunner):
```

As `UnicodeProcessRunner` may override things that we need from `MinijailProcessRunner`

src/python/tests/core/bot/fuzzers/afl/afl_launcher_integration_test.py
oliverchang: This is for input from keyboard, so not really needed. Also on Python 3 the default is utf-8, so leaving it out here to avoid another cleanup in the future.

PR #1530, opened 2020-03-10
Python 3 fixes for Fuchsia code.

oliverchang:


src/python/tests/core/bot/tasks/corpus_pruning_task_test.py
```diff
@@ -300,7 +300,7 @@ def test_prune(self):
     corpus = os.listdir(self.corpus_dir)
     self.assertEqual(2, len(corpus))
     six.assertCountEqual(self, [
-        '31836aeaab22dc49555a97edb4c753881432e01d',
+        '801c34269f74ed383fc97de33604b8a905adb635',
```
oliverchang: This was failing on Python 2 as well, and there is no testcase in the test input which matches this hash.

src/python/platforms/fuchsia/util/device.py
```diff
@@ -231,9 +231,9 @@ def _guess_pid(self):
         """
     out = self._dump_log(['--only', 'reset,Fuzzer,Sanitizer'])
     pid = -1
-    for line in out.split('\n'):
+    for line in out.split(b'\n'):
```
inferno-chromium: why not decode here ?

src/python/platforms/fuchsia/util/device.py
```diff
@@ -193,13 +193,13 @@ def ls(self, path):
...
       out, _ = p.communicate()
-      for line in str(out).split('\n'):
+      for line in out.decode('utf-8').split('\n'):
         # Line ~= '-rw-r--r-- 1 0 0 8192 Mar 18 22:02 some-name'
         parts = line.split()
         # When we're running ls over ssh, we may get a note about
         # "Warning: Permanently added [address] to the list of known hosts"
         # Don't try to treat those as file paths
-        if len(parts) > 8 and "Warning:" not in parts:
+        if len(parts) > 8 and 'Warning:' not in parts:
```
inferno-chromium: b'Warning:' ?

oliverchang: parts is a str (unicode) so that would be wrong here.

src/python/platforms/fuchsia/util/device.py
oliverchang: Seems simpler to work on bytes here, as this is just a few lines and the resulting bytes isn't propagated to anywhere.

PR #1531, opened 2020-03-10
Increase backoff for monitoring.

oliverchang:


src/python/metrics/monitor.py
```diff
@@ -47,7 +47,9 @@
 
 CUSTOM_METRIC_PREFIX = 'custom.googleapis.com/'
 FLUSH_INTERVAL_SECONDS = 10 * 60  # 10 minutes.
-RETRY_DEADLINE_SECONDS = 3 * 60  # 3 minutes.
+RETRY_DEADLINE_SECONDS = 5 * 60  # 5 minutes.
+RETRY_INITIAL_SECONDS = 16
```
inferno-chromium: INITIAL_DELAY_SECONDS
and MAXIUM_DELAY_SECONDS
for var name to match default names (+clarity) in lib.

_DEFAULT_INITIAL_DELAY = 1.0  # seconds
_DEFAULT_MAXIMUM_DELAY = 60.0  # seconds
_DEFAULT_DELAY_MULTIPLIER = 2.0
_DEFAULT_DEADLINE = 60.0 * 2.0  # seconds

oliverchang: Done!

PR #1533, opened 2020-03-10
Speculative fixes for disk space issues.

inferno-chromium:
- Fix #1376 by storing crashes during merge in a reproducers dir.
- Create merge corpus directory early-on to avoid running into:
```
Exception: Failed to recreate directory: /mnt/scratch0/clusterfuzz/bot/inputs/fuzzer-testcases-disk/temp-368/merge-corpus
at recreate_directory (/mnt/scratch0/clusterfuzz/src/python/bot/fuzzers/engine_common.py:500)
at _create_temp_corpus_dir (/mnt/scratch0/clusterfuzz/src/python/bot/fuzzers/libFuzzer/engine.py:166)
at _create_merge_corpus_dir (/mnt/scratch0/clusterfuzz/src/python/bot/fuzzers/libFuzzer/engine.py:171)
at _merge_new_units (/mnt/scratch0/clusterfuzz/src/python/bot/fuzzers/libFuzzer/engine.py:188)
at fuzz (/mnt/scratch0/clusterfuzz/src/python/bot/fuzzers/libFuzzer/engine.py:316)
```

PR #1536, opened 2020-03-10
Fix corpus pruning: MCF should persist across merge steps (follow-up #1533).

Dor1s:


Dor1s:
Not needed as we've debugged this more with Abhishek and ended up reverting in https://github.com/google/clusterfuzz/pull/1537

Issue #1540, opened 2020-03-10
Update TensorFlow and other dependencies after Python3 migration


Dor1s:
We haven't updated it in a while and while working on https://github.com/google/clusterfuzz/issues/1535 it became evident that we're using a too old version. For instance, there is not build of it for Python3.7.

I'm hoping to fix https://github.com/google/clusterfuzz/issues/1535 with some old-but-new-enough version, but we'll definitely should upgrade all ML pieces after the migration.

mihaimaruseac:
I didn't know TF is being used as backend here. But if there's some help needed from TF team, I can do the work.

Dor1s:
Thank you @mihaimaruseac! It's currently used by a small feature only, which isn't even enabled in OSS-Fuzz yet.

Dor1s:
@mihaimaruseac could you please advise whether it makes sense to upgrade to 1.15.10 or should we target 2.2.0 instead? How much code changes would major upgrade involve?

mihaimaruseac:
We recommend using the latest update for longer support. We are currently in the process of releasing 2.3.0, so maybe let's wait a few more weeks and then switch to that?

The main upgrade cost would be in converting from 1.15 to 2.0. There is a script to do the API conversion but that won't cover the transition from Session and graph mode to eager mode.

If you file a bug against me internally I can try to do the conversion in a week (or less) after the 2.3 release.

oliverchang:
I recall running into some issues with tensorflow.contrib.{layers,rnn} no longer being available in tensorflow 2.x when we tried the automated conversion a few months ago. Are there drop in replacements for those?

mihaimaruseac:
Keras should have support for them now.

mihaimaruseac:
Is there a dockerfile I can use to test this? Or should I attempt to build following the existing install instructions?

mihaimaruseac:
I looked through the code and it seems the main issue with conversion would be converting from TF v1 graph API / control flow to the v2 eager one. The APIs used seem to be available. I can send a PR with the fix, though I'm thinking we should wait until 2.3 is fully released. But if that's too long to wait for, I can attempt a fix earlier, against either 2.2.0 or the RC for 2.3.

Also, there are two docker files configured to install TF 1.15. Plus, the CUDA version used is 9.0 but TF 2.2 and 2.3 support only CUDA 10.1. But I think these changes are not a priority, first we need to convert the rnn fuzzer.

Dor1s:
[...]

Issue #1535, opened 2020-03-10
Fix ML RNN tests for Python 3.


oliverchang:
6 out of the remaining 14 failing tests are in ML RNN code:

```
======================================================================
FAIL: test_empty_corpus (tests.core.bot.fuzzers.ml.rnn.ml_rnn_generator_test.MLRnnGeneratorIntegrationTest)
Test generation should abort for empty corpus.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/google/home/ochang/clusterfuzz/src/python/tests/core/bot/fuzzers/ml/rnn/ml_rnn_generator_test.py", line 89, in test_empty_corpus
    self.assertEqual(result.return_code, constants.ExitCode.CORPUS_TOO_SMALL)
AssertionError: 1 != 2

======================================================================
FAIL: test_generate (tests.core.bot.fuzzers.ml.rnn.ml_rnn_generator_test.MLRnnGeneratorIntegrationTest)
Test generate specified number of inputs.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/google/home/ochang/clusterfuzz/src/python/tests/core/bot/fuzzers/ml/rnn/ml_rnn_generator_test.py", line 68, in test_generate
    self.assertEqual(result.return_code, constants.ExitCode.SUCCESS)
AssertionError: 1 != 0

======================================================================
FAIL: test_invalid_model (tests.core.bot.fuzzers.ml.rnn.ml_rnn_generator_test.MLRnnGeneratorIntegrationTest)
Test TensorFlow should throw exception if model does not match.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/google/home/ochang/clusterfuzz/src/python/tests/core/bot/fuzzers/ml/rnn/ml_rnn_generator_test.py", line 114, in test_invalid_model
    self.assertEqual(result.return_code, constants.ExitCode.TENSORFLOW_ERROR)
AssertionError: 1 != 3

======================================================================
FAIL: test_timeout (tests.core.bot.fuzzers.ml.rnn.ml_rnn_generator_test.MLRnnGeneratorIntegrationTest)
Test timeout case in generation.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/google/home/ochang/clusterfuzz/src/python/tests/core/bot/fuzzers/ml/rnn/ml_rnn_generator_test.py", line 140, in test_timeout
    self.assertTrue(result.timed_out)
[...]

PR #1541, opened 2020-03-10
Fix TensorFlow dependency and ML RNN tests for Python3 (#1535).

Dor1s:
Besides the typical py2->3 problem in src/python/bot/fuzzers/ml/rnn/utils.py, another issue here is that tensorflow 1.8.0 is way too old and wasn't even built for Python3.7, and official tensorflow GCS bucket doesn't have it.

1.13.2 seems to be the first version built for Python3.7, and it's also not that new to break code compatibility. Should be acceptable in the meantime. After migration we'll update TF and its dependencies altogether (https://github.com/google/clusterfuzz/issues/1540).

### DIFF
--- a/src/python/bot/fuzzers/ml/rnn/utils.py
+++ b/src/python/bot/fuzzers/ml/rnn/utils.py
@@ -94,7 +94,7 @@ def encode_text(bytes_data):
   if sys.version_info.major == 3:
     return list(bytes_data)
 
-  result = list(map(ord, bytes_data))
+  return list(map(ord, bytes_data))
 
 
 def decode_to_text(encoded_list):
--- a/src/python/bot/fuzzers/ml/rnn/utils.py
+++ b/src/python/bot/fuzzers/ml/rnn/utils.py
@@ -82,19 +82,19 @@ def get_files_info(directory):
   return files_info_list
 
 
-def encode_text(string_data):
-  """Encode string data to a list of integers.
+def encode_text(bytes_data):
+  """Encode byte string to a list of integers.
 
   Args:
-    string_data: String data.
+    bytes_data: Byte string.
 
   Returns:
     An encoded list of integers representing code points.
   """
   if sys.version_info.major == 3:
-    return list(string_data)
+    return list(bytes_data)
 
-  result = list(map(ord, string_data))
+  result = list(map(ord, bytes_data))
 
 
 def decode_to_text(encoded_list):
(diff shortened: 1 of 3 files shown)


Dor1s:
hey @oliverchang, I wonder if you know any `pipenv` magic. When I do the following:

0) Pipfile has no mention of tensorflow, but the rest of it is the same
1) Run `PY3=1 ./local/install_deps.bash`
2) run any of the following:
   - `pip install https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-1.13.2-cp37-cp37m-linux_x86_64.whl`
   - `pipenv install https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-1.13.2-cp37-cp37m-linux_x86_64.whl`
   - `pipenv install -d https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-1.13.2-cp37-cp37m-linux_x86_64.whl`

the tests work fine (`INTEGRATION=1 python butler.py py_unittest -t core -p '*ml*'`)

but whenever I try to embed this dependency into Pipfile, it doesn't get installed properly.

Looks like there was some bug with transitive dependencies for such cases, but it was fixed a while ago: https://github.com/pypa/pipenv/pull/1944

Dor1s:
Explicitly specifying those dependencies seems to work. Should be fine for now, PTAL

src/python/bot/fuzzers/ml/rnn/utils.py
```diff
@@ -91,7 +91,13 @@ def encode_text(string_data):
   Returns:
     An encoded list of integers representing code points.
   """
-  return list(map(ord, string_data))
+  result = None
+  try:
+    result = list(map(ord, string_data))
```
oliverchang: which one is the python2 case, and which one is python 3? maybe we should be explicit and use

```python
if sys.version_info.major == 3:
 etc
```

Dor1s: nice, thanks!

oliverchang: Another super nit now that I understand what's going on :) 

We should rename "string_data" to "bytes_data", as strings in python 3 are different. In this case we're reading as 'rb' before passing it to this function so we're working on bytes (hence the exception when you try to call `ord`).

Dor1s: done!

mihaimaruseac:
I would recommend 1.15 or 2.1 instead of 1.13. The 2 mentioned ones are long term releases and will receive updates for 3 years. 1.13 is nearing its end of life soon.

Dor1s:
wow, thanks @mihaimaruseac I can try 1.15 tomorrow and see if it's still compatible with our code

oliverchang:
Btw I also had some issues with my local pipenv  -- it wasn't installing tensorflow properly and I had to install it manually. Hopefully this goes away with upgrade.

PR #1542, opened 2020-03-10
Allow evicting data bundles when trying to make space for build.

inferno-chromium:


inferno-chromium:
Thoughts on this ?
Basically we will have all data bundles in one dir (not mixed with FUZZ_DATA) and we always sync data bundle post build setup. in build setup, if we run out of space, we clear data bundles first, then builds. that way, we have less exceptions and can actually continue task.

src/python/bot/tasks/setup.py
```diff
@@ -604,19 +604,19 @@ def get_data_bundle_directory(fuzzer_name):
...
+
   # Check if we have a fuzzer-specific data bundle. Use it to calculate the
   # data directory we will fetch our testcases from.
   data_bundle = data_types.DataBundle.query(
       data_types.DataBundle.name == fuzzer.data_bundle_name).get()
   if not data_bundle:
     # Generic data bundle directory. Available to all fuzzers if they don't
     # have their own data bundle.
-    return environment.get_value('FUZZ_DATA')
+    return local_data_bundles_directory
```
oliverchang: won't this break fuzzers that rely on web tests, since this is what's passed to the fuzzer --input_dir ? 

web tests are downloaded explicitly to FUZZ_DATA in update_task.py.

inferno-chromium: ah right, forgot they will go into wrong dir. closing.

oliverchang:
Let me think on a good solution for this... I'll file a bug to track it so we don't forget.

PR #1546, opened 2020-03-11
Set up CI for Python 3.

oliverchang:


src/local/butler/integration_tests.py
```diff
@@ -28,14 +29,21 @@
 
 def execute(_):
   """Run integration tests."""
+  if sys.version_info.major == 3:
+    command = 'run_server3'
+    indicator = b'Booting worker'
```
inferno-chromium: Why is the indicator different here ?

oliverchang: We're no longer using dev_appserver, but a different server framework.

PR #1547, opened 2020-03-11
Generalize python 3 deployment.

oliverchang:
- Automatically figure out whether we're deploying Python 2 or 3 based
  on environment.
- Prepare for production App Engine deployment via the prod3 config dir.
- Prepare for zip deployment in the form of e.g. linux-3.zip
- Skip deployment of prod/prod3 if it's not specified, to allow for gradual migration.

### DIFF
--- a/src/local/butler/deploy.py
+++ b/src/local/butler/deploy.py
@@ -319,7 +319,8 @@ def _staging_deployment_helper(deploy_go, python3=False):
 def _prod_deployment_helper(config_dir,
                             package_zip_paths,
                             deploy_go=True,
-                            deploy_appengine=True):
+                            deploy_appengine=True,
+                            python3=False):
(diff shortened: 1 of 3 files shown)


src/local/butler/deploy.py
```diff
@@ -91,7 +91,7 @@ def _deploy_app_prod(project,
                      package_zip_paths,
                      deploy_appengine=True):
   """Deploy app in production."""
-  if deploy_appengine:
+  if deploy_appengine and yaml_paths:
```
inferno-chromium: nit: "deploy_appengine and yaml_paths" can be moved to a local and use in three places

oliverchang: Just set `deploy_appengine = False` if there are no yaml paths.

PR #1548, opened 2020-03-11
Bump TensorFlow to 1.15.2 (a long term release), follow-up #1541.

Dor1s:


Dor1s:
rebased, tests succeeded locally, should be good to go after GCB run finishes

Issue #893, opened 2019-08-27
Implement proper corpus eviction.


oliverchang:
Right now, the only way we deal with our corpora filling up disk is to call shell.clear_data_directories and exit with a log_error.

For cases where there are many targets which can run on a single bot, we reach this state easily. We should have a non-fatal way of handling this by cleaning up corpora we don't need.

oliverchang:
Actually there is another related issue here -- we don't clear FUZZ_DATA. For some deployments the libFuzzer and AFL entities won't have a data bundle set so it'll default to FUZZ_DATA. This leads to a buildup that's never cleaned up properly.

oliverchang:
Per #1543 this issue is coming back so we should fix this.

oliverchang:
We should probably also take the chance to clean up FUZZ_DATA logic -- it makes cleaning it difficult as web tests can live there.

oliverchang:
Re corpora filling up disk, maybe we can even do a rolling model where we store at most N corpora on disk. When we try to set up the N+1'th corpora, we clear the oldest. 

For small OSS-Fuzz projects with a small number of targets, this would ensure we don't waste time constantly re-syncing corpora. For larger deployments with hundreds/thousands, this would stop the buildup.

PR #1549, opened 2020-03-11
Move builtin fuzzers corpora to local data bundles directory.

inferno-chromium:
This makes it easy to implement a corpus eviction strategy
when we run out of disk space.


inferno-chromium:
This lays the ground for https://github.com/google/clusterfuzz/issues/893

PR #1551, opened 2020-03-11
Fix Attribute error in record cross pollination stats. 

mpherman2:
There was a check that was stopping this method from running during local development and py_unittest so we did not catch these errors. I moved the check to just before the bigquery call so that the function could at least be tested up to that point.

### DIFF
--- a/src/python/bot/tasks/corpus_pruning_task.py
+++ b/src/python/bot/tasks/corpus_pruning_task.py
@@ -563,9 +563,9 @@ def _record_cross_pollination_stats(stats):
       'feature_coverage': stats.feature_coverage
   }
 
-  # BigQuery not available in local development.This is necessary because the
+  # BigQuery not available in local development. This is necessary because the
   # untrusted runner is in a separate process and can't be easily mocked.
-  # Check here instead of earlier to test for attribute errors above.
+  # Check here instead of earlier to test as much of the function as we can.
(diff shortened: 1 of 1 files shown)


src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -549,25 +549,27 @@ def _record_cross_pollination_stats(stats):
...
-      'corpus_size': stats.minimized_corpus_size_units,
+      'corpus_size': stats.corpus_size,
       'initial_edge_coverage': stats.initial_edge_coverage,
       'edge_coverage': stats.edge_coverage,
-      'initial_feature_coverage': stats.initial_feature_coverate,
+      'initial_feature_coverage': stats.initial_feature_coverage,
       'feature_coverage': stats.feature_coverage
   }
 
+  # BigQuery not available in local development.This is necessary because the
```
mbarbella-chromium: Nit: space after period.

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -549,25 +549,27 @@ def _record_cross_pollination_stats(stats):
...
+  # untrusted runner is in a separate process and can't be easily mocked.
+  # Check here instead of earlier to test for attribute errors above.
```
mbarbella-chromium: Nit: rather than call out the attribute errors specifically, just mention that we return as late as possible to test as much of the function as we can.

PR #1555, opened 2020-03-12
Don't lint in Python 3 CI.

oliverchang:
pylint for Python 3 seems to introduce a many new (helpful) lint checks which currently fail.
We'll fix these after the migration is complete.


inferno-chromium:
Landed to fix CI.

PR #1556, opened 2020-03-12
optimization debian package manager tweaks

Rajpratik71:
By default, Ubuntu or Debian based "apt" or "apt-get" system installs recommended but not suggested packages .

By passing "--no-install-recommends" option, the user lets apt-get know not to consider recommended packages as a dependency to install.

This results in smaller downloads and installation of packages .

Refer to blog at [Ubuntu Blog](https://ubuntu.com/blog/we-reduced-our-docker-images-by-60-with-no-install-recommends) .

Signed-off-by: Pratik Raj <rajpratik71@gmail.com>


inferno-chromium:
These are self-contained small packages without dependencies, so this does not apply to us. Thanks for the information on this flag.

PR #1562, opened 2020-03-12
Deploy clusterfuzz-source.manifest with a .3 suffix for Python 3.

oliverchang:
Otherwise the versions will clash and cause bots to constantly try to
update.

### DIFF
--- a/src/local/butler/deploy.py
+++ b/src/local/butler/deploy.py
@@ -218,9 +218,14 @@ def _deploy_zip(bucket_name, zip_path):
 
 def _deploy_manifest(bucket_name, manifest_path):
   """Deploy source manifest to GCS."""
-  common.execute(
-      'gsutil cp -a public-read %s '
-      'gs://%s/clusterfuzz-source.manifest' % (manifest_path, bucket_name))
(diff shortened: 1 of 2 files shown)


mbarbella-chromium:
I'm a little bit confused as to what the deployment process will look like now. Are you still planning on deploying to app engine using the Python 3 runtime while we're still running 2 on the bots? If so, are we just expecting the bots to skip the source code update in that case?

oliverchang:
> I'm a little bit confused as to what the deployment process will look like now. Are you still planning on deploying to app engine using the Python 3 runtime while we're still running 2 on the bots? If so, are we just expecting the bots to skip the source code update in that case?

Deployment happens twice and are independent of one another

From Python 2 -> We deploy to App Engine if the config for Python 2 exists. Sources uploaded to gs://bucket/PLATFORM.zip and gs://bucket/clusterfuzz-source.manifest

From Python 3 -> We deploy to App Engine if the config for Python 3 exists. Sources uploaded to gs://bucket/PLATFORM-3.zip and gs://bucket/clusterfuzz-source.manifest.3

PR #1563, opened 2020-03-12
Fix more Python 3 issues.

oliverchang:


oliverchang:
TBR for trivial change. /gcbrun

PR #1564, opened 2020-03-13
Preparation for staging Python 3 bots

oliverchang:
- Move Python 3 installation from CI to base image
- Prepare for supporting running ClusterFuzz bots with Python 3


oliverchang:
FYI added some more small changes in this PR for preparing for staging bots.

PR #1567, opened 2020-03-13
Revert "Fix docker build error"

oliverchang:
Reverts google/clusterfuzz#1566


oliverchang:
Reverting as this seems to cause another error later on.

PR #1568, opened 2020-03-13
Fix ml-with-gpu image build.

oliverchang:


oliverchang:
Should fix `AssertionError: mock>=2.0.0 .dist-info directory not found`

oliverchang:
TBR

PR #1570, opened 2020-03-13
Added IAP check in the access check for is_privileged_user. 

urbanenomad:
All IAP users are privileged to allow for access controlled by google groups.  This requires that IAP be enabled and turned on to the AppEngine

### DIFF
--- a/src/appengine/libs/access.py
+++ b/src/appengine/libs/access.py
@@ -32,17 +32,16 @@ def _is_privileged_user(email):
   privileged_user_emails = (db_config.get_value('privileged_users') or
                             '').splitlines()
   # IAP users are automatically priviledged
-  # This requires IAP to be turned on 
+  # This requires IAP to be turned on
   # and a google group in IAM to have the "IAP-secured Web App User" role
   current_request = auth.get_current_request()
   iap_email = auth.get_iap_email(current_request)
   if iap_email:
     return True
-  else:
-    return any(
-      utils.emails_equal(email, privileged_user_email)
-      for privileged_user_email in privileged_user_emails)
 
+  return any(
+    utils.emails_equal(email, privileged_user_email)
+    for privileged_user_email in privileged_user_emails)
 
 def _is_blacklisted_user(email):
   """Check if an email is in the privileged users list."""
--- a/src/appengine/libs/access.py
+++ b/src/appengine/libs/access.py
@@ -35,7 +35,8 @@ def _is_privileged_user(email):
 
   current_request = auth.get_current_request()
   iap_email = auth.get_iap_email(current_request)
-  if iap_email:
+  iap_ispriv = local_config.AuthConfig().get('iap_is_privileged', default=False)
+  if iap_email and iap_ispriv:
     return True
 
   privileged_user_emails = (db_config.get_value('privileged_users') or
--- a/src/appengine/libs/access.py
+++ b/src/appengine/libs/access.py
@@ -42,8 +42,8 @@ def _is_privileged_user(email):
   privileged_user_emails = (db_config.get_value('privileged_users') or
                             '').splitlines()
   return any(
-    utils.emails_equal(email, privileged_user_email)
-    for privileged_user_email in privileged_user_emails)
+      utils.emails_equal(email, privileged_user_email)
+      for privileged_user_email in privileged_user_emails)
 
 def _is_blacklisted_user(email):
   """Check if an email is in the privileged users list."""
--- a/src/appengine/libs/access.py
+++ b/src/appengine/libs/access.py
@@ -45,6 +45,7 @@ def _is_privileged_user(email):
       utils.emails_equal(email, privileged_user_email)
       for privileged_user_email in privileged_user_emails)
 
+
 def _is_blacklisted_user(email):
   """Check if an email is in the privileged users list."""
   blacklisted_user_emails = (db_config.get_value('blacklisted_users') or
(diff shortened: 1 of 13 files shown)


inferno-chromium:
Why do you think IAP users be privileged by default. I think an admin for project should go to /configuration page and add those emails explicitly in privileged users list.

urbanenomad:
I have not investigated IAP deep enough to find an authorization mechanism in IAP.  So far it is only binary whether you are part of a google group that has "IAP-secured Web App User" role or not.  I need to be able to control privilege at the google group IAM level so that it can be tied back to our ldap groups.  Is there another way you know of to give privileged access by IAM controls?

urbanenomad:
@googlebot I fixed it.

src/appengine/libs/access.py
```diff
@@ -31,10 +31,17 @@ def _is_privileged_user(email):
   """Check if an email is in the privileged users list."""
   privileged_user_emails = (db_config.get_value('privileged_users') or
```
inferno-chromium: move this to after line 40.

src/appengine/libs/access.py
```diff
@@ -31,10 +31,17 @@ def _is_privileged_user(email):
...
                             '').splitlines()
-  return any(
-      utils.emails_equal(email, privileged_user_email)
-      for privileged_user_email in privileged_user_emails)
+  # IAP users are automatically priviledged
```
inferno-chromium: typo priviledged
End comments with dot(.).

src/appengine/libs/access.py
urbanenomad: done

src/appengine/libs/access.py
urbanenomad: done

urbanenomad:
@googlebot I fixed it.

oliverchang:
I also don't think we should automatically consider all IAP users to be privileged. We currently have a cron job that syncs Admin entities: https://github.com/google/clusterfuzz/blob/7468b0de57c962694a1029655bd662ec59583b4f/src/appengine/handlers/cron/sync_admins.py#L26 based on IAM permissions. Would that work for you?

urbanenomad:
is admins from iam policy work with google groups?  We can't managed access to a project based on individuals in IAM...matter of fact it is a security policy that all IAM members be managed by google groups that can be tied back to ldap groups.  We don't allow for individual users into IAM services.  Will this function work with google groups?

oliverchang:
Ok, thanks for explaining your use case. Before we go down this route though, let's figure out if you need privileged access at all, as that's quite powerful. As a general auth mechanism, I certainly don't think we should automatically consider all users privileged by default. 

We have the ability to grant all users in a domain that are authenticated access to security bugs: https://github.com/google/clusterfuzz/blob/master/configs/test/gae/auth.yaml#L18 and the checkbox to allow security bugs on /configuration. Would this suffice?

urbanenomad:
Yes I don’t mind removing this request as long as their is a mechanism for controlling access through google groups and not have to add people individually into IAM. We are planning to offer this to our engineers and we have automated tools that can add and remove access via LDAP groups.  We need to control access via this mechanism which can be synced by googles directory sync between LDAP and google groups.

oliverchang:
> Domain name is also too broad for us. Development teams will get their own clusterfuzz to do their testing and only that team will need access to everything. We don’t plan to let everyone in our company to see the test or bug results.

Re being too broad, IAP will still limit this. It just means that anyone who gets through IAP with @domain.com will have access to security bugs etc. Do they need access to perform privileged operations as well?

Could you clarify what you mean by "get their own clusterfuzz"? Do you mean every team gets their own instance?

urbanenomad:
Yes if you are part of the IAP access which is controlled by google group access in this project they will get access to privileged parts.

And yes we plan to give each team their own clusterfuzz cluster.

oliverchang:
> Yes if you are part of the IAP access which is controlled by google group access in this project they will get access to privileged parts.
> 
> And yes we plan to give each team their own clusterfuzz cluster.

Ok, so I don't think we can accept this PR as is (or if we do, we can't promise we won't change the behaviour). 

What we can do is to support a switch to allow everyone authenticated be privileged users. i.e. add a flag to https://github.com/google/clusterfuzz/blob/2591c16e5b20db4425cb3ab78bafea8ce7f23d6d/configs/test/gae/auth.yaml#L33 like "all_users_privileged" or something similar. @inferno-chromium WDYT?

urbanenomad:
That sounds fine to me, as long as I can use google groups in IAM as the means of authentication and not individual users in IAM.

inferno-chromium:
> > Yes if you are part of the IAP access which is controlled by google group access in this project they will get access to privileged parts.
> > And yes we plan to give each team their own clusterfuzz cluster.
> 
> Ok, so I don't think we can accept this PR as is (or if we do, we can't promise we won't change the behaviour).
> 
> What we can do is to support a switch to allow everyone authenticated be privileged users. i.e. add a flag to
> 
> https://github.com/google/clusterfuzz/blob/2591c16e5b20db4425cb3ab78bafea8ce7f23d6d/configs/test/gae/auth.yaml#L33
> 
> like "all_users_privileged" or something similar. @inferno-chromium WDYT?

SGTM

urbanenomad:
Added flag in auth.yaml file

oliverchang:
Thanks for the changes. I think this can be done a bit more simply, so I uploaded https://github.com/google/clusterfuzz/pull/1669

oliverchang:
https://github.com/google/clusterfuzz/pull/1669 is now merged. Let us know if this works for you.

urbanenomad:
This is fine with me.  Thanks for the update!

PR #1574, opened 2020-03-13
Clear temp directory on untrusted worker.

inferno-chromium:


inferno-chromium:
This is not needed i think, this is just a 90mb temp. Bigger one is another cl.

PR #1576, opened 2020-03-13
Reland Peach Mutator

mpherman2:
I am still not sure what caused the syntax error the first time this landed. The line it was pointing to seems fine, all unit tests pass, and it runs locally fine. If you could PTAL and see if anything sticks out.

I also changed one environment variable and updated the peach mutator. I still need to make changes so the peach mutator will run in Python 2. Let me know if you figure out a way to do this.


Dor1s:
Could you please post the link to the exception which made us revert this originally?

mpherman2:
https://pantheon.corp.google.com/errors/CMGShYf91ZvuUg?time=P30D&project=clusterfuzz-external

PR #1577, opened 2020-03-13
Clean fuzzer temp dir between engine fuzz iterations.

inferno-chromium:


inferno-chromium:
This should fix ton of out of space exceptions on OSS-Fuzz.

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1316,6 +1317,10 @@ def run_engine_fuzzer(engine_impl, target_name, sync_corpus_directory,
 
   fuzzer_metadata.update(engine_common.get_all_issue_metadata(target_path))
   _add_issue_metadata_from_environment(fuzzer_metadata)
+
+  # Cleanup fuzzer temporary artifacts (e.g. mutations dir, merge dirs. etc).
+  shell.remove_directory(fuzzer_utils.get_temp_dir())
```
oliverchang: We should move this to fuzz_task. That way, we don't have to do this for all engine impls.

oliverchang: Oops, completely misread this code. Ignore.

PR #1583, opened 2020-03-16
Fixes for Python 3 bot running.

oliverchang:
- Create a chromium/base3 image for testing.
- Remove call to sys.clear_exc() in Python 3. It does not exist there.
- Fix encoding issue when logging to stackdriver.

### DIFF
--- a/src/python/bot/tasks/commands.py
+++ b/src/python/bot/tasks/commands.py
@@ -84,9 +84,7 @@ def cleanup_task_state():
   environment.reset_current_memory_tool_options()
 
   # Clear exceptions.
-  if sys.version_info.major == 2:
-    # TODO(ochang): Remove this once migrated to Python 3.
-    sys.exc_clear()
+  utils.exc_clear()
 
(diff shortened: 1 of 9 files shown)


src/python/bot/tasks/commands.py
```diff
@@ -84,7 +84,9 @@ def cleanup_task_state():
   environment.reset_current_memory_tool_options()
 
   # Clear exceptions.
-  sys.exc_clear()
+  if sys.version_info.major == 2:
+    # TODO(ochang): Remove this once migrated to Python 3.
+    sys.exc_clear()
```
inferno-chromium: i think we use this in 3 places, feel free to create a helper in utils.py and use it

src/python/bot/startup/heartbeat.py:        sys.exc_clear()
src/python/bot/startup/run_bot.py:    sys.exc_clear()
src/python/bot/tasks/commands.py:  sys.exc_clear()

oliverchang: Done!

PR #1591, opened 2020-03-16
Fix a unicode exception in libFuzzer/engine.py.

mbarbella-chromium:


oliverchang:
> There are many more places with this pattern where we add result.output, see many places in this file, i guess in afl, honggfuzz too and also corpus pruning task. @oliverchang - thoughts ? should we use repr ?

The more general fix here would be to use UnicodeProcessRunner for all the LibFuzzerRunner classes, which is already done for honggfuzz and AFL.

inferno-chromium:
> This change LGTM itself, but agreed there's a more general fix needed.

Lets do this UnicodeProcessRunner fix, i dont like this fix please, it is incomplete i feel like. @oliverchang - is it easy to change the remaining places, can you do this ?

oliverchang:
This should be fixed more generally by https://github.com/google/clusterfuzz/pull/1595

PR #1592, opened 2020-03-16
Properly handle bytes objects passed to utils.decode_to_unicode.

mbarbella-chromium:
If we pass a bytes to this function, instead of iterating over 1-character strings we'll be iterating over the integer values. This makes the ord() call unnecessary.

### DIFF
--- a/src/python/base/utils.py
+++ b/src/python/base/utils.py
@@ -108,7 +108,11 @@ def decode_to_unicode(obj, encoding='utf-8'):
     try:
       obj = str(obj, encoding)
     except:
-      obj = str(''.join(char for char in obj if ord(char) < 128), encoding)
+      obj = str(
+          ''.join(
+              char for char in obj
+              if (char if isinstance(char, int) else ord(char)) < 128),
+          encoding)
 
   return obj
 


src/python/base/utils.py
```diff
@@ -108,7 +108,11 @@ def decode_to_unicode(obj, encoding='utf-8'):
     try:
       obj = str(obj, encoding)
     except:
-      obj = str(''.join(char for char in obj if ord(char) < 128), encoding)
+      obj = str(
+          ''.join(
+              char for char in obj
+              if (char if isinstance(char, int) else ord(char)) < 128),
```
oliverchang: should we just use errors='ignore' instead of doing our own filtering here?

mbarbella-chromium: I don't recall why we were doing this, but I think there was some reason. Ideally I'd like to get rid of this function, so I'd rather address it by doing that.

inferno-chromium: @oliverchang @mbarbella-chromium - can we really get rid of this function ??

i like Oliver's idea, i dont remember why we implemented this way.
obj = obj.encode(encoding, errors='ignore') ? that should run into exceptions.

mbarbella-chromium: I still don't like the type checks we're doing here. It seems cleaner to me to have callers that need to use decode directly with errors=ignore. This does seem odd as-is though, so I'll give it a try with errors='ignore' here, and if that works we can factor it out later if we decide to.

PR #1593, opened 2020-03-16
[fuchsia] Update names of example fuzzers

eepeep:
These recently changed on the Fuchsia side, so the integration tests and
exclusions needed to be updated.


eepeep:
@oliverchang Please take a look when you get a chance, thanks!

PR #1594, opened 2020-03-16
Restrict Honggfuzz max input to 5 Mb (default is 128 MB).

inferno-chromium:
This avoids blowing corpus quickly and failing gsutil rsyncs.


inferno-chromium:
AFL is 1 mb which feels too low to me, so keeping 5 mb. Thoughts - higher, lower ? libFuzzer is based on max seed input size which makes more sense (otherwise we pick 10000 which also feels slightly low)).

jonathanmetzman:
>(otherwise we pick 1-10000 which also feels slightly low)

Is this done in FuzzBench? My understanding of libFuzzer is that it uses a max len of 1<<20 (1 MB) if there is none and there is no corpus https://cs.chromium.org/chromium/src/third_party/libFuzzer/src/FuzzerLoop.cpp?type=cs&q=kMaxSaneLen+file:%5Esrc/third_party/libFuzzer/src/+package:%5Echromium$&g=0&l=727

I think 1 MB is large enough in most cases. I think the larger the inputs, the less likely the fuzzer is to produce an interesting mutation.

inferno-chromium:
Discussed with @jonathanmetzman , keeping 5 mb just to keep some variance. will move to 1 mb in some time (to match afl, and libfuzzer in no-corpus,no-max len mode).

robertswiecki:
Yup, it'd break the size autodiscovery. I lowered the maximum maximum limit to 1MB - https://github.com/google/honggfuzz/commit/0414a9934a507d3334f3101a14630ec15acda566 - the previous one is an artifcact from older days.

A suggestion, maybe you can have some filter in clusterfuzz, which will trim the size of individual corpus file to a fixed value (say 1MB), possibly all fuzzers do similar heuristics (at least libfuzzer does), so such change can align them. But it's optional now I guess. Dunno about AFL.

PR #1596, opened 2020-03-17
Revert "Restrict Honggfuzz max input to 5 Mb (default is 128 MB)."

inferno-chromium:
Reverts google/clusterfuzz#1594

Robert says this is wrong and breaks auto-discovery algo. Default limit is 8 kb or largest file in seed corpus. So, looks like something else broken in corpus with large items.


inferno-chromium:
he also reduced that max limit to 1 mb - https://github.com/google/honggfuzz/commit/0414a9934a507d3334f3101a14630ec15acda566

PR #1597, opened 2020-03-17
Create Python 3 docker images for oss-fuzz.

oliverchang:
Move start_python3.sh to the root base image, and generalize it to
accept an argument pointing to the python entrypoint.


oliverchang:
TBR

PR #1600, opened 2020-03-17
honggfuzz_engine: Make honggfuzz not delete *SAN report files.

robertswiecki:
Currently honggfuzz doesn't touch *SAN log/report files if these are
produced, but I intend to change that behavior in the future (delete by
default). Set a flag for honggfuzz, which will prevent deleting those,
so clusterfuzz can use those report-files for its own crash analysis.

### DIFF
--- a/src/python/bot/fuzzers/honggfuzz/engine.py
+++ b/src/python/bot/fuzzers/honggfuzz/engine.py
@@ -37,6 +37,7 @@ _DEFAULT_ARGUMENTS = [
     '-z',  # use clang instrumentation
     '-P',  # persistent mode
     '-S',  # enable sanitizers
+    '--sanitizers_del_report=false',  # keep ASAN log files around for parsing
     '--rlimit_rss',
     str(_RSS_LIMIT),
     '--timeout',


robertswiecki:
Not sure about those presubmit checks, is it something on my or the infra side?

inferno-chromium:
I dont think we need this. We dont save ASan report files to disk, but parse it from stderrr. Also, libFuzzer runs in its own job, so its artifacts are retained in its own job.

That test is failing since we have honggfuzz tests in https://github.com/google/clusterfuzz/blob/ea66c6e2667f0914166843575413217772bfe6d7/src/python/tests/core/bot/fuzzers/honggfuzz/honggfuzz_engine_test.py#L109 [makes sure that fuzzing works with code changes].
To fix and run tests:
git clone https://github.com/google/clusterfuzz
cd clusterfuzz
./local/install_deps.bash
source ENV/bin/activate
INTEGRATION=1 python butler.py py_unittest -t core -p honggfuzz*
once done,
python butler.py format
python butler.py lint
send pr :)

inferno-chromium:
For Honggfuzz crashes, we write to workspace dir
```
        '--workspace',
        reproducers_dir,
```
https://github.com/google/clusterfuzz/blob/master/src/python/bot/fuzzers/honggfuzz/engine.py#L151

and then parse stacks from it,
https://github.com/google/clusterfuzz/blob/master/src/python/bot/fuzzers/honggfuzz/engine.py#L68

With your future change of --sanitizers_del_report=true, will it not write crashes to workspace dir ?

robertswiecki:
It's just enforcing the current default value (it's 'false' anyway). I want to change it in the future, cause ASAN logs are useful only for limited subset of cases (e.g. for clusterfuzz), but generally they eat up disk space, as there's one report for crash.

I'll take a look at the indent.

oliverchang:
Thanks for proactively fixing this on our side before changing the default @robertswiecki :)

oliverchang:
@robertswiecki I think you need to update the test here (https://github.com/google/clusterfuzz/blob/ea66c6e2667f0914166843575413217772bfe6d7/src/python/tests/core/bot/fuzzers/honggfuzz/honggfuzz_engine_test.py#L109) as @inferno-chromium pointed out:

Could you please do this when you get the chance so we can merge this? :)

robertswiecki:
Hi, sorry for the delay.

The last time I tried to fix this PR, I had some problems with python tooling in clusterfuzz (some intriguing python panic()s when linting/formatting code :).

Would you mind taking over this PR for me? I promise to learn more python in the coming days, but some quick hacking wasn't enough IIRC.

inferno-chromium:
Sure, i will take over.

inferno-chromium:
> Hi, sorry for the delay.
> 
> The last time I tried to fix this PR, I had some problems with python tooling in clusterfuzz (some intriguing python panic()s when linting/formatting code :).
> 
> Would you mind taking over this PR from me? I promise to learn more python in the coming days, but some quick hacking wasn't enough IIRC.

Done in https://github.com/google/clusterfuzz/pull/1674

robertswiecki:
Thank you!

PR #1601, opened 2020-03-17
Don't file crash if its job is removed.

inferno-chromium:
Clean up cron closes testcases when job is removed. However,
there can be a race and this fix prevent that crash from
getting filed.


src/appengine/handlers/cron/recurring_tasks.py
```diff
@@ -44,7 +44,7 @@ def get(self):
               testcase.key.id(),
               testcase.job_type,
               queue=tasks.queue_for_testcase(testcase))
-        except tasks.Error:
+        except Exception:
```
inferno-chromium: unrelated cron exception fix.

PR #1603, opened 2020-03-17
Suppress unwanted warnings in googleapiclient.

inferno-chromium:
Should fix warnings like
```
ImportError: file_cache is unavailable when using oauth2client >= 4.0.0 or google-auth
```

### DIFF
--- a/src/python/metrics/logs.py
+++ b/src/python/metrics/logs.py
@@ -63,6 +63,12 @@ def _console_logging_enabled():
   return bool(os.getenv('LOG_TO_CONSOLE'))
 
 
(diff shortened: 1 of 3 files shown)


src/python/metrics/logs.py
```diff
@@ -89,6 +95,7 @@ def get_handler_config(filename, backup_count):
 def get_logging_config_dict(name):
   """Get config dict for the logger `name`."""
   logging_handler = {
+      'appengine': get_handler_config('bot/logs/appengine.log', 1),
```
inferno-chromium: This is for local tests, which need to write this.

PR #1605, opened 2020-03-17
Reland "Suppress unwanted warnings in googleapiclient.""

inferno-chromium:


src/python/metrics/logs.py
```diff
@@ -89,6 +96,7 @@ def get_handler_config(filename, backup_count):
 def get_logging_config_dict(name):
   """Get config dict for the logger `name`."""
   logging_handler = {
+      'appengine': get_handler_config('bot/logs/appengine.log', 1),
```
oliverchang: why is this needed? what happens if we don't include it? 

Please include a comment if it's really needed :)

inferno-chromium: Fixed tests to mock as if running in appengine. PTAL.

PR #1607, opened 2020-03-18
Fix log entry corelation in App Engine logs.

inferno-chromium:


inferno-chromium:
Tested on chromium instances, seems to work, can search for this in logs.

```
resource.type="gae_app"
protoPayload.resource="/schedule-progression-tasks"
```

PR #1608, opened 2020-03-18
Use google.cloud.logging for logging on App Engine.

oliverchang:
- Install google.cloud.logging handler to root logger. This is necessary for severity logging to work.
- Also attach trace information to each log to allow correlating logs with requests.

### DIFF
--- a/src/python/metrics/logs.py
+++ b/src/python/metrics/logs.py
@@ -252,6 +252,20 @@ def uncaught_exception_handler(exception_type, exception_value,
   sys.__excepthook__(exception_type, exception_value, exception_traceback)
 
 
+def configure_appengine():
+  """Configure logging for App Engine."""
+  logging.getLogger().setLevel(logging.INFO)
+
+  if os.getenv('LOCAL_DEVELOPMENT') or sys.version_info.major == 2:
+    # TODO(ochang): Remove Python 2 check once all migrated to Python 3.
+    return
+
+  import google.cloud.logging
+  client = google.cloud.logging.Client()
+  handler = client.get_default_handler()
+  logging.getLogger().addHandler(handler)
+
+
 def configure(name, extras=None):
   """Set logger. See the list of loggers in bot/config/logging.yaml.
   Also configures the process to log any uncaught exceptions as an error.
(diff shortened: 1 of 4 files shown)


src/python/metrics/logs.py
```diff
@@ -252,14 +252,28 @@ def uncaught_exception_handler(exception_type, exception_value,
   sys.__excepthook__(exception_type, exception_value, exception_traceback)
 
 
+def configure_appengine():
+  """Configure logging for App Engine."""
+  logging.getLogger().setLevel(logging.INFO)
+
+  if os.getenv('LOCAL_DEVELOPMENT') or sys.version_info.major == 2:
```
inferno-chromium: split into two checks to make it easy to not to remove os.getenv('LOCAL_DEVELOPMENT') check.

src/python/metrics/logs.py
```diff
@@ -317,12 +331,37 @@ def get_source_location():
   return 'Unknown', '-1', 'Unknown'
 
 
+def _add_appengine_trace(extras):
+  """Add App Engine tracing information."""
+  if sys.version_info.major != 3 or not _is_running_on_app_engine():
+    # TODO(ochang): Remove Python 3 check once all migrated to Python 3.
```
inferno-chromium: Does it break anything on python 2, if not can simplify this check and one above to remove sys.version_info.major != 3

src/python/metrics/logs.py
```diff
@@ -317,12 +331,37 @@ def get_source_location():
...
+          project_id=project_id, trace_id=trace_id)
+
+
 def emit(level, message, exc_info=None, **extras):
   """Log in JSON."""
   logger = get_logger()
   if not logger:
     return
 
+  if exc_info == (None, None, None):
```
inferno-chromium: Why this, can you add a comment, is this needed for appengine (not bot)?

src/python/metrics/logs.py
inferno-chromium: more of a nit, can skip.

src/python/metrics/logs.py
oliverchang: Want to avoid any unexpected breakages there. Will be removed very soon anyway.

oliverchang:
Btw I think with this logging library we can get rid of fluentd completely. Will file a bug for that.

PR #1612, opened 2020-03-18
Remove some Python 3 dependencies.

oliverchang:
Some of these seem to be unused.


oliverchang:
These also need to be changed to be installed on Docker build, as GCE instances may not have full network access.

Issue #1616, opened 2020-03-18
radamsa generates a lot of large testcases blowing up corpus


Dor1s:
I think @inferno-chromium was investigating a similar issue recently, but I'm afraid we may need to do a huge clean up now

While debugging https://github.com/google/oss-fuzz/issues/3449 I've noticed that 99% of large testcases are produced with radamsa:

```
$ ls -lS ghostscript_gstoraster_fuzzer/ | head -n 1000 | egrep radamsa | wc -l
990
```

inferno-chromium:
This should fix it - https://github.com/google/clusterfuzz/pull/1581, waiting for some corpus pruning tasks to actually complete.

inferno-chromium:
Ah i see a bug specific to OSS-Fuzz.
i think there is a bug in our untrusted corpus pruning task handling. Basically we dont propagate the last execution failure.

oliverchang:
Let's strongly enforce this with a limit on our side. Since engines have a size limit anyway, it does not make sense to try running with larger inputs generated by external mutators etc.

inferno-chromium:
gs://ghostscript-corpus.clusterfuzz-external.appspot.com/libFuzzer/ghostscript_gstoraster_fuzzer
it was not a file size issue here i think, but overall corpora was so big and kept growing since corpus pruning was exiting with exception and we never set last execution failed (OSS-Fuzz only).

but anyway, we should add some filter on upload_files in fuzz task, so that big files cannot sneak anytime.

oliverchang:
+1 to a limit enforced before we upload.

Dor1s:
1MB? or 5MB?

inferno-chromium:
Since this is a global upload limit, i think higher than 1 mb makes sense [5/10/15mb ?]. Wdyt @Dor1s @oliverchang  ? I am really worried if seed corpuses have big files [e.g. media stuff] and we break those with this limit. this is reason i fixed everything else first rather than enforcing this limit.

Regarding this particular issue, it is currently getting fixed, corpus pruning task is clearing stuff currently, right now WIP, down from 19 gb to like 13 gb uptil now
gs://ghostscript-corpus.clusterfuzz-external.appspot.com/libFuzzer/ghostscript_gstoraster_fuzzer

Dor1s:
i don't think anything greater 5MB is reasonable

oliverchang:
+1 to 5MB. Anything larger probably won't be effective during fuzzing anyway.

[...]

PR #1617, opened 2020-03-18
Catch all exceptions from corpus pruning tasks.

inferno-chromium:
This catches exceptions that come over grpc channel from OSS-Fuzz
worker bots for the same CorpusPruningException. This helps to mark
last execution failed properly instead of task failure.


inferno-chromium:
Should fix https://github.com/google/clusterfuzz/issues/1616

PR #1629, opened 2020-03-19
Unconditionally encode directory name in remove_directory.

mbarbella-chromium:


inferno-chromium:
@oliverchang - Travis has stopped working recently, maybe some CI change affected this.

src/python/system/shell.py
```diff
@@ -469,10 +470,7 @@ def clear_read_only(func, path, _):
     # File paths are not necessarily valid unicode, so we must encode the
     # directory here to prevent exceptions when shutil tries to build
     # subdirectory paths.
-    encoded_directory = directory
-    if isinstance(encoded_directory, str):
-      encoded_directory = encoded_directory.encode('utf-8')
-
+    encoded_directory = directory.encode('utf-8')
```
oliverchang: was this fixing another exception? or just cleanup?

inferno-chromium: It didnt fix the original exception, since on python 2, that string was basestring, not str

```
UnicodeDecodeError: 'utf8' codec can't decode byte 0xfd in position 84: invalid start byte
at clear_read_only (/mnt/scratch0/clusterfuzz/src/python/system/shell.py:453)
at rmtree (/usr/lib/python2.7/shutil.py:252)
at rmtree (/usr/lib/python2.7/shutil.py:247)
at rmtree (/usr/lib/python2.7/shutil.py:247)
at rmtree (/usr/lib/python2.7/shutil.py:247)
at rmtree (/usr/lib/python2.7/shutil.py:247)
at remove_directory (/mnt/scratch0/clusterfuzz/src/python/system/shell.py:476)
at remove_directory (/mnt/scratch0/clusterfuzz/src/python/bot/untrusted_runner/file_impl.py:34)
at RemoveDirectory (/mnt/scratch0/clusterfuzz/src/python/bot/untrusted_runner/untrusted.py:129)
at wrapper (/mnt/scratch0/clusterfuzz/src/python/bot/untrusted_runner/untrusted.py:84)
at _call_behavior (/mnt/scratch0/clusterfuzz/src/third_party/grpc/_server.py:389)
```

oliverchang: This could potentially cause exceptions on Python 3 right? If bytes are already passed to this function? Unlike Python 2, there's no bytes.encode

mbarbella-chromium: Yes it could. The motivation behind the original implementation was to be resilient, but I'm still torn about the idea of trying to arbitrarily handle either str or bytes in many of our helper functions. It looked at a glance (at least from tests) that we were always passing str, but I'm not 100% sure this is safe. After we've migrated away from python 2 the check could be restored if we want it to be resilient.

PR #1630, opened 2020-03-19
Invoke blackbox fuzzers with Python 2 explicitly.

mbarbella-chromium:


mbarbella-chromium:
PTAL. I'm a little bit worried this won't work on Windows and haven't had a chance to test that. Does anyone happen to have easy access to one of the windows bots to see if python2 is in their path?

oliverchang:
> Haven't checked Windows yet. I can check this today and merge if this is the case.

Just checked a Windows bot:
```
 Directory of c:\python27

11/06/2018  10:26 AM    <DIR>          .
11/06/2018  10:26 AM    <DIR>          ..
11/06/2018  10:26 AM    <DIR>          DLLs
11/06/2018  10:26 AM    <DIR>          Doc
11/06/2018  10:26 AM    <DIR>          include
11/06/2018  10:26 AM    <DIR>          Lib
11/06/2018  10:26 AM    <DIR>          libs
04/30/2018  05:41 PM            38,586 LICENSE.txt
07/29/2008  06:28 AM             1,859 Microsoft.VC90.CRT.manifest
07/29/2008  06:23 AM           626,688 msvcr90.dll
04/30/2018  05:37 PM           493,404 NEWS.txt
04/30/2018  05:31 PM            28,160 python.exe
04/30/2018  05:31 PM         3,428,352 python27.dll
04/30/2018  05:31 PM            28,160 pythonw.exe
04/14/2018  11:47 PM            56,951 README.txt
01/02/2020  02:15 PM    <DIR>          Scripts
11/06/2018  10:26 AM    <DIR>          tcl
11/06/2018  10:26 AM    <DIR>          Tools
```

Looks like there's just `python.exe`..

src/python/system/shell.py
```diff
@@ -350,14 +350,25 @@ def get_interpreter(file_to_execute):
...
+    interpreter = interpreter_extension_map[extension]
   except KeyError:
     return None
 
+  # TODO(mbarbella): Remove this when fuzzers have been migrated to Python 3.
+  # TODO(mbarbella): Remove the Windows platform check when bots have a
+  # python2.exe available.
+  if (invoke_with_python2 and interpreter == 'python' and
+      environment.platform() != 'WINDOWS'):
+    interpreter = 'python2'
```
oliverchang: Please rebase this to after my change here: https://github.com/google/clusterfuzz/pull/1647

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1453,7 +1453,10 @@ def generate_blackbox_testcases(self, fuzzer, fuzzer_directory,
         return error_occurred, None, None, None, None
 
       # Build the fuzzer command execution string.
-      command = shell.get_execute_command(fuzzer_executable)
+      # TODO(mbarbella): Conditionally invoke with Python 2 depending on
+      # the fuzzer. At present, all were implemented to work with Python 2.
+      command = shell.get_execute_command(
```
inferno-chromium: Can you make this rely on an env variable
USE_PYTHON2_FOR_BLACKBOX_FUZZERS
and then set that env in project.yaml for chromium.
I don't want this to happen for other CF users.

src/python/system/shell.py
inferno-chromium: +1

src/python/bot/tasks/fuzz_task.py
mbarbella-chromium: Good call. Done.

src/python/system/shell.py
mbarbella-chromium: Done

src/python/system/shell.py
```diff
@@ -350,14 +350,23 @@ def get_interpreter(file_to_execute):
...
   try:
-    return interpreters[os.path.splitext(file_to_execute)[1]]
+    interpreter = interpreters[os.path.splitext(file_to_execute)[1]]
   except KeyError:
     return None
 
+  # TODO(mbarbella): Remove this when fuzzers have been migrated to Python 3.
+  if (is_blackbox_fuzzer and interpreter == sys.executable and
+      environment.get_value('USE_PYTHON2_FOR_BLACKBOX_FUZZERS') and
+      environment.platform() != 'WINDOWS'):
```
inferno-chromium: Can we remove this "environment.platform() != 'WINDOWS'", since we copy python2.exe in windows startup script (from @oliverchang )

mbarbella-chromium: We still need it for now since the Python 2 bots would break in this case.

.pylintrc
```diff
@@ -30,6 +30,7 @@ disable=
   suppressed-message,
   ungrouped-imports,
   unsubscriptable-object,
+  useless-object-inheritance, # Remove once all bots are on Python 3.
```
oliverchang: There's quite a few extra lint checks in Python 3. Maybe we should just remove the lint step from Travis until then. We're already doing that for the Python 3 Cloud build CI.

mbarbella-chromium: If they get to be a problem it might make sense. The other check that I disabled in one spot in fuzz task actually looked good to me, though. I just didn't want to make a more invasive change as part of this fix. I wouldn't object to disabling them if you think it makes sense, but I also don't mind keeping them around so we can slowly clean things up as we transition.

src/python/system/shell.py
oliverchang: Can we do this check instead:


```python
if (environment.get_value('USE_PYTHON2_FOR_BLACKBOX_FUZZERS') 
     and sys.version_info.major == 3):
  ...
```

This will make Windows Py 3 work. We should be using the `python2` binary explicitly only in the Python 3 case anyway.

mbarbella-chromium: Very good point, I'll make this change in a new PR.

PR #1634, opened 2020-03-20
Speculative fix for Travis CI.

oliverchang:


oliverchang:
TBR. looks like this works.

PR #1636, opened 2020-03-20
Revert shell.remove_directory utf-8 exception fixes.

inferno-chromium:


inferno-chromium:
Didn't work, needs rethought. it only happens for unrar on OSS-Fuzz which creates weird files all over disk.

PR #1644, opened 2020-03-21
Fix Python 3 bots to use pipenv run, as python3.7 is not inherited.

inferno-chromium:
python3.7 call on run.py still causes python 2 to be loaded on
run_bot.py, which is the child process. Use "pipenv run" instead.

### DIFF
--- a/docker/chromium/base3/Dockerfile
+++ b/docker/chromium/base3/Dockerfile
@@ -14,4 +14,4 @@
 FROM gcr.io/clusterfuzz-images/chromium/base
 
 ENV DEPLOYMENT_ZIP "linux-3.zip"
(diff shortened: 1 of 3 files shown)


inferno-chromium:
This didn't work out, it fixed the child process issue (run_bot runs with python 3.7), but Pipfile deps are not in pipenv shell. Will let @oliverchang fix this (i am unfamiliar with pipenv).

oliverchang:
> This didn't work out, it fixed the child process issue (run_bot runs with python 3.7), but Pipfile deps are not in pipenv shell. Will let @oliverchang fix this (i am unfamiliar with pipenv).

will look at fixing this

PR #1645, opened 2020-03-22
Fix exception in deps_to_revisions_dict.

oliverchang:


oliverchang:
PTAL. not sure what changed here, but this seems to make things work.

PR #1652, opened 2020-03-23
Increase test timeout.

oliverchang:


oliverchang:
For some reason Python 3 tests consistently pass in a very short amount of time. Python 2 is slow/hanging most of the time.

PR #1655, opened 2020-03-23
Only encode environment entries on Python 2.

mbarbella-chromium:
PTAL


inferno-chromium:
Need to import sys module to fix CI failure.

mbarbella-chromium:
Travis should pass once the pylint changes from #1630 have landed. I'll land this right after.

oliverchang:
Thanks!

PR #1658, opened 2020-03-23
Make build setup result checking more strict in corpus pruning.

oliverchang:
Otherwise we may proceed even if the build actually failed to be set up
(e.g. a broken build on Fuchsia).

### DIFF
--- a/src/python/bot/tasks/corpus_pruning_task.py
+++ b/src/python/bot/tasks/corpus_pruning_task.py
@@ -595,9 +595,9 @@ def do_corpus_pruning(context, last_execution_failed, revision):
     return tasks_host.do_corpus_pruning(context, last_execution_failed,
                                         revision)
 
-  build_manager.setup_build(revision=revision)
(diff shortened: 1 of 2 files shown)


src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -595,9 +595,9 @@ def do_corpus_pruning(context, last_execution_failed, revision):
     return tasks_host.do_corpus_pruning(context, last_execution_failed,
                                         revision)
 
-  build_manager.setup_build(revision=revision)
+  build_setup_result = build_manager.setup_build(revision=revision)
   build_directory = environment.get_value('BUILD_DIR')
-  if not build_directory:
+  if not build_setup_result or not build_directory:
```
inferno-chromium: do we even need to check build_directory ?

oliverchang: Let me try removing it.

PR #1662, opened 2020-03-24
Build python 3 with bz2 and lzma support.

inferno-chromium:
Fixes "tarfile.ReadError: file could not be opened successfully"


inferno-chromium:
Tested, it fixes things,
list used in fuzzbench,
mentioned in other places like https://linuxize.com/post/how-to-install-python-3-8-on-ubuntu-18-04/

PR #1664, opened 2020-03-24
Hacky fix for unicode exceptions in AFL launcher.

mbarbella-chromium:
PTAL. My original intent was to land this along with a change back to having fuzzer_stderr return a bytes, but it looks like a str is expected in other parts of the file and that seems like a riskier change. I also wonder if some of the other print statements in the file are safe.

### DIFF
--- a/src/python/bot/fuzzers/afl/launcher.py
+++ b/src/python/bot/fuzzers/afl/launcher.py
@@ -1334,7 +1334,7 @@ def load_testcase_if_exists(fuzzer_runner, testcase_file_path):
            | stat.S_IXGRP)
 
   fuzzer_runner.run_single_testcase(testcase_file_path)
-  print(fuzzer_runner.fuzzer_stderr)
+  sys.stdout.buffer.write(fuzzer_runner.fuzzer_stderr.encode('utf-8'))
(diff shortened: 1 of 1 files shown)


mbarbella-chromium:
Looking at the test failures I'm likely going to abandon this in favor of another change. Hold off on review here for now.

inferno-chromium:
Feels too hacky to me, @oliverchang do you have any better ideas.

@Dor1s - afl code is getting crazy hacky since we run a launcher in a different python process, i think moving to new engine pipeline might be good if we get chance.

mbarbella-chromium:
It's likely sufficient to set PYTHONIOENCODING=utf_8 before running the launcher script, but I'm having trouble finding the cleanest place to set it. I don't think we need to do this for all of ClusterFuzz, but I suppose that's also an option. Either way I don't think this change makes sense, so abandoning it.

Dor1s:
> @Dor1s - afl code is getting crazy hacky since we run a launcher in a different python process, i think moving to new engine pipeline might be good if we get chance.

ack, I'll keep this in mind and prioritize AFL migration and it becomes possible for me

PR #1668, opened 2020-03-25
Use Python 2 for blackbox launchers.

oliverchang:


oliverchang:
afl is still using Python 3, as the launcher.py for that comes in via APP_NAME and not Fuzzer.launcher_path.

inferno-chromium:
But AFL's launcher.py should Py3 compatible, and shouldn't we run this in python 3 ?

oliverchang:
> But AFL's launcher.py should Py3 compatible, and shouldn't we run this in python 3 ?

Yes, that's exactly the case even after this PR :)

inferno-chromium:
oh right right, looked at it now more closely this is about general launchers. and then it is still based on sys.executable. Thanks for fix.
LGTM!

PR #1672, opened 2020-03-25
Fix an exception when running Antlr under Python 3.

mbarbella-chromium:
Mitchel, would you mind taking a look at this? I think we may have had some other check in place to deal with Antlr failing to handle invalid utf-8, but this seems like a cleaner way to handle it which also fixes an issue we're seeing when running with Python 3. If there was another check, I'd like to remove it.

### DIFF
--- a/src/python/base/utils.py
+++ b/src/python/base/utils.py
@@ -99,14 +99,20 @@ def utc_datetime_to_timestamp(dt):
   return (dt - datetime.datetime.utcfromtimestamp(0)).total_seconds()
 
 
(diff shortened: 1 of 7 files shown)


mbarbella-chromium:
Thanks for the feedback. I switched the new check to return an AntlrDecodeError so that the existing fallbacks for line minimization should still work, and removed the old places where it was returned.

mbarbella-chromium:
TBR. The Travis failures are a bunch of python 3 lint errors that would be better to clean up in another change, since this is already larger than I planned.

inferno-chromium:
@oliverchang - can you do a review pass please.

PR #1673, opened 2020-03-26
Speculative fix for AFL minimization issues.

oliverchang:
The fuzzer override wasn't propagating to the right places.


oliverchang:
TBR

PR #1674, opened 2020-03-26
Make honggfuzz not delete *SAN report files.

inferno-chromium:
Patch by Robert Swiecki.

Currently honggfuzz doesn't touch *SAN log/report files if these are
produced, but the author intends to change that behavior in the future
(delete by default). Set a flag for honggfuzz, which will prevent deleting those,
so ClusterFuzz can use those report-files for its own crash analysis.

Binary is updated to account for flag rename.


inferno-chromium:
Actually thinking again @robertswiecki - can we keep this option disabled by default [as it-is]. This CL will break regression testing on OSS-Fuzz and google instance since older builds had the sanitizer_del_report (without -s) [due to https://github.com/google/honggfuzz/commit/8393e6bb12a761bce4899e66142f189f54125d17#diff-42ec6b5f0f84c374ed83ccc21fcc13a6], so this breaks reproduction on older builds.

I would prefer to keep things backward compatible as other devs can get confused with wrong regression range. unsure if there is any command line option to ignore flags that honggfuzz does not understand.

robertswiecki:
> Actually thinking again @robertswiecki - can we keep this option disabled by default [as it-is]. 

Sure, what I can probably do is to set this flag by default to 'false' if ```BUILD_OSSFUZZ_STATIC``` is used? Or, would be be too fragile of a hack for you?

https://github.com/google/oss-fuzz/blob/81c465662194fefeca2a80cc4426789f7683b7e3/infra/base-images/base-builder/precompile_honggfuzz#L19

robertswiecki:
Ok, let's keep it this way for now. I guess I can revisit the topic in a year or so, when the risk of breaking something relatively older than one year will be v. low.

inferno-chromium:
> Ok, let's keep it this way for now. I guess I can revisit the topic in a year or so, when the risk of breaking something older than one year will be v. low.

SGTM.

inferno-chromium:
> > Actually thinking again @robertswiecki - can we keep this option disabled by default [as it-is].
> 
> Sure, what I can probably do is to set this flag by default to 'false' if `BUILD_OSSFUZZ_STATIC` is used? Or, would be be too fragile of a hack for you?
> 
> https://github.com/google/oss-fuzz/blob/81c465662194fefeca2a80cc4426789f7683b7e3/infra/base-images/base-builder/precompile_honggfuzz#L19
> 
> Not sure how this will affect our internal builds (which can be updated anyway) though.

Yes the problem is older internal builds, so best to lets wait a year as you said.

PR #1675, opened 2020-03-26
Migrate certs cron to Python 3.

oliverchang:
- Remove Go deployment code and server code.
- Split src/requirements.txt into src/requirements.txt and
  src/appengine/requirements.txt for App Engine only dependencies.


src/local/butler/common.py
```diff
@@ -301,6 +301,13 @@ def install_dependencies(platform_name=None, is_reproduce_tool_setup=False):
         'src/third_party',
         platform_name=platform_name)
 
+  with tempfile.NamedTemporaryFile() as f:
+    f.write(open('src/requirements.txt', 'rb').read())
+    f.write(open('src/appengine/requirements.txt', 'rb').read())
+    f.flush()
+
+    _install_pip(f.name, 'src/appengine/third_party')
```
inferno-chromium: nit: why not two _install_pip calls rather than temp file approach.

src/python/tests/appengine/handlers/cron/oss_fuzz_generate_certs_test.py
```diff
@@ -0,0 +1,77 @@
...
+import webapp2
+import webtest
+
+from datastore import data_types
+from handlers.cron import oss_fuzz_generate_certs
+from tests.test_libs import helpers as test_helpers
+from tests.test_libs import test_utils
+
+
+@unittest.skipIf(sys.version_info.major != 3, 'Python 3 only')
```
inferno-chromium: there was some helper decorator for this ?

src/appengine/requirements.txt
```diff
@@ -0,0 +1,8 @@
+firebase-admin==2.16.0
+Jinja2==2.11.1
+jira==2.0.0
+PyJWT==1.7.1
+pyOpenSSL==19.1.0
```
inferno-chromium: Isnt pyOpenSSL coming from requirements.txt.

src/python/tests/appengine/handlers/cron/oss_fuzz_generate_certs_test.py
mbarbella-chromium: We have python2_only but not a python 3 equivalent. It might make sense to for consistency, but hopefully both will be irrelevant soon. It would make cleanup after the migration is complete a little easier to have our own decorators in both cases, though.

src/local/butler/common.py
oliverchang: That didn't work for some reason. The second call would overwrite/remove some things in the first call.

src/appengine/requirements.txt
oliverchang: The only other place this was getting installed was on bots (which is platform specific). We can maybe even get rid of that there.

src/python/tests/appengine/handlers/cron/oss_fuzz_generate_certs_test.py
oliverchang: Added a python3_only decorator.

PR #1679, opened 2020-03-27
py2->3: Fold chromium/base3 image into chromium/base.

oliverchang:


docker/chromium/ml-with-gpu/Dockerfile
```diff
@@ -38,10 +38,10 @@ RUN wget https://storage.googleapis.com/clusterfuzz-data/cudnn/7.1.4_cuda9.0/lib
 
 # Replace TensorFlow CPU version with GPU version. Also the version number
 # needs to match cuda and cuDNN version above.
```
oliverchang: @Dor1s how do I tell that this new version (1.15.2) is compatible with the cuda and cuDNN versions above?

Dor1s: There are some known tested combinations listed in https://www.tensorflow.org/install/source#gpu

It doesn't even list 1.15.2, but seems like it would require cuDNN bump (and possibly drivers as well)

I've checked that when working on #1535 and ended up thinking that we should migrate to 2+, update all other dependencies as well as the code (it's not compatible either): https://github.com/google/clusterfuzz/issues/1540

oliverchang: Thanks! I won't be deploying this new image today. If you get the chance, please look at updating this to the right versions. Otherwise, I can deploy next week and see if this actually breaks anything.

Dor1s: Ack, but sadly can't promise any ETA on this :( maybe next week will be better on my end.

What's the issue with leaving this as 1.8.0 for now and then migrating everything to 2+ sometime in Q2 when I get around #1540?

oliverchang: We need it for Python 3.7. I believe that was the [reason](https://github.com/google/clusterfuzz/pull/1541#issue-386359607) for the upgrade from 1.8.0->1.15.0 in the first place? Everywhere else was changed except for the GPU package install.

Dor1s: I see. I'm also not optimistic that model generation script would run as Python3, since I remember we asked Huan to convert the code from 3 to 2 back then, and we don't have an integration test for it because of the GPU requirement :/

In the worst case scenario, we can disable the training task and set number of GPU bots to 0.

oliverchang: Ok let me try this and monitor things next week. I can also fix anything obvious that comes up :)

Dor1s: SG, thanks!

Yeah, bumping revisions and fixing common code compatibility issues is likely not a big deal, but last time I had a lot of trouble finding correct version of the drivers -- if that's still the case then it'll be time consuming and totally unpleasant thing to do :)

PR #1684, opened 2020-03-27
Upgrade gRPC to fix encoding issue.

oliverchang:
First release that contains:
https://github.com/grpc/grpc/commit/a04a928cf387a845931ad10490f2ba4b0f472a16

We upgrade conservatively for now to avoid any other issues that may
arise while we migrate from Python 2 -> 3.


oliverchang:
PTAL. this is causing exceptions in some cases when exceptions for untrusted workers contain unicode characters: https://github.com/google/clusterfuzz/blob/cdeb160c5b1dc5a03d701d5ef1d9f055375bf196/src/python/bot/untrusted_runner/untrusted.py#L88

/gcbrun

PR #1687, opened 2020-03-28
Fix chromium builder image startup exception.

inferno-chromium:


inferno-chromium:
TBR, also current image is updated and this works.

PR #1688, opened 2020-03-28
Upgrade old git on chromium builder and tests sync image.

inferno-chromium:
Otherwise we fail on gclient sync.
https://groups.google.com/a/chromium.org/g/chromium-dev/c/BwLhfb6-8_8/m/eojZ4fh0BQAJ


inferno-chromium:
TBR, image is upgraded in place

PR #1689, opened 2020-03-29
Fix low disk space issues on OSS-Fuzz.

oliverchang:
Lower free disk space threshold back to 2GB on trusted hosts.


oliverchang:
PTAL. some oss-fuzz hosts are continously running exiting due to this.

PR #1692, opened 2020-03-31
Fix libFuzzer minimization in minijail.

oliverchang:
We were passing incorrectly passing -artifact_prefix=-artifact_prefix=/tmp/.


oliverchang:
TBR

PR #1693, opened 2020-03-31
Remove most Go code.

oliverchang:
Only Go code left are the mock metadata server for local Docker testing,
and the local GCS emulator for uploads. These are moved to local/.


local/gcs.go
```diff
@@ -143,6 +143,7 @@ func handleGet(w http.ResponseWriter, r *http.Request) {
 
 // handler is the root http request handler.
 func handler(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Access-Control-Allow-Origin", "http://localhost:9000")
```
inferno-chromium: Is this needed, as this hardcodes 9000 ?

oliverchang: Yeah, otherwise errors are very difficult to debug (as was my case while making this PR). It's not possible to control this port anyway via butler.py CLI interface, so keeping this.

oliverchang:
/gcbrun. Also removed generate_datastore_models, which I missed originally.

local/run_metadata.bash
```diff
@@ -20,5 +20,4 @@ set +x
 docker_ip=$(ip -4 addr show docker0 | grep -P inet | head -1 | awk '{print $2}' | cut -d/ -f1)
```
inferno-chromium: Probably lets move metadata.go and gcs.go in a emulator subdir, don't want too many files in local/ root dir ? thoughts ?

oliverchang: subdir sounds better. Done!

Issue #1691, opened 2020-03-30
Release new reproduce-tool-stable and make reproduce.sh use python 3


inferno-chromium:

mbarbella-chromium:
Fixes landed and the stable tag has been updated.

PR #1695, opened 2020-03-31
Update the reproduce tool to run with Python 3.

mbarbella-chromium:
Small change to the reproduce tool to address issue #1691. PTAL when you have time.

### DIFF
--- a/reproduce.sh
+++ b/reproduce.sh
@@ -32,12 +32,14 @@ while [ "$1" != "" ]; do
 done
 
 mkdir -p $CLUSTERFUZZ_CONFIG_DIR
-if [ ! -d $ROOT_DIRECTORY/ENV ] || ([ $additional_deps_args ] && [ ! -d $ROOT_DIRECTORY/local/bin/android-sdk ]); then
+installed_version=$(cat $CLUSTERFUZZ_CONFIG_DIR/version 2>/dev/null || echo -n)
+version=1
+if pipenv graph 2>&1 | grep -q "Please run within a Pipenv project." || [ "$installed_version" != "$version" ] || ([ $additional_deps_args ] && [ ! -d $ROOT_DIRECTORY/local/bin/android-sdk ]); then
   echo "Running first time setup. This may take a while, but is only required once."
   echo "You may see several password prompts to install required packages."
   sleep 5
-  $ROOT_DIRECTORY/local/install_deps.bash --only-reproduce $additional_deps_args || { rm -rf $ROOT_DIRECTORY/ENV && exit 1; }
+  PY3=1 $ROOT_DIRECTORY/local/install_deps.bash --only-reproduce $additional_deps_args || { exit 1; }
+  echo -n $version > $CLUSTERFUZZ_CONFIG_DIR/version
 fi
 
-source ENV/bin/activate
-python $ROOT_DIRECTORY/butler.py reproduce $original_args
+pipenv run python $ROOT_DIRECTORY/butler.py reproduce $original_args
--- a/reproduce.sh
+++ b/reproduce.sh
@@ -31,10 +31,28 @@ while [ "$1" != "" ]; do
   shift
 done
 
+# Later checks will set this to false if dependencies are missing.
+dependencies_installed="true"
+
+# Create the config directory if needed and check the ClusterFuzz version.
 mkdir -p $CLUSTERFUZZ_CONFIG_DIR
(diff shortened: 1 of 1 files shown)


inferno-chromium:
@oliverchang - can you take a review pass. i dont understand the pipenv part well.

reproduce.sh
```diff
@@ -32,12 +32,14 @@ while [ "$1" != "" ]; do
 done
 
 mkdir -p $CLUSTERFUZZ_CONFIG_DIR
-if [ ! -d $ROOT_DIRECTORY/ENV ] || ([ $additional_deps_args ] && [ ! -d $ROOT_DIRECTORY/local/bin/android-sdk ]); then
+installed_version=$(cat $CLUSTERFUZZ_CONFIG_DIR/version 2>/dev/null || echo -n)
+version=1
```
inferno-chromium: This should be VERSION in capital at start of file ? Also, should it be 1.0.0 to match CF other release style ?

reproduce.sh
```diff
@@ -32,12 +32,14 @@ while [ "$1" != "" ]; do
 done
 
 mkdir -p $CLUSTERFUZZ_CONFIG_DIR
-if [ ! -d $ROOT_DIRECTORY/ENV ] || ([ $additional_deps_args ] && [ ! -d $ROOT_DIRECTORY/local/bin/android-sdk ]); then
+installed_version=$(cat $CLUSTERFUZZ_CONFIG_DIR/version 2>/dev/null || echo -n)
```
inferno-chromium: This should check file exists in an if rather than 2>&1 stuff

inferno-chromium: Use version_file for $CLUSTERFUZZ_CONFIG_DIR/version maybe use .version for filename

reproduce.sh
```diff
@@ -32,12 +32,14 @@ while [ "$1" != "" ]; do
...
+version=1
+if ! pipenv graph 2>&1 > /dev/null || [ "$installed_version" != "$version" ] || ([ $additional_deps_args ] && [ ! -d $ROOT_DIRECTORY/local/bin/android-sdk ]); then
```
inferno-chromium: maybe put these four into local var so that it explains what they do
e.g.
pipenv_exists 
version_changed
?

reproduce.sh
mbarbella-chromium: It won't likely be seen and the other files in the directory are lowercase. As for the style, this is basically just something I'll be using to force updates if we ever make any breaking changes. The reproduce tool versions don't necessarily align with ClusterFuzz releases.

reproduce.sh
mbarbella-chromium: Good call on the change to version_file, but since this is already a hidden directory (~/.config/clusterfuzz) I don't think there's a good reason to rename it to have the dot prefix.

mbarbella-chromium: Done

reproduce.sh
mbarbella-chromium: Good call, this was getting very hard to read. Broke it up into a few sections.

reproduce.sh
```diff
@@ -31,13 +31,33 @@ while [ "$1" != "" ]; do
...
+  dependencies_installed="false"
+fi
+
+if [ "$dependencies_installed" != "true" ]; then
   echo "Running first time setup. This may take a while, but is only required once."
   echo "You may see several password prompts to install required packages."
   sleep 5
-  $ROOT_DIRECTORY/local/install_deps.bash --only-reproduce $additional_deps_args || { rm -rf $ROOT_DIRECTORY/ENV && exit 1; }
+  PY3=1 $ROOT_DIRECTORY/local/install_deps.bash --only-reproduce $additional_deps_args || { exit 1; }
+  echo -n $version > $CLUSTERFUZZ_CONFIG_DIR/version
```
oliverchang: where is `$version` set?

reproduce.sh
```diff
@@ -31,13 +31,33 @@ while [ "$1" != "" ]; do
...
 fi
 
-source ENV/bin/activate
-python $ROOT_DIRECTORY/butler.py reproduce $original_args
+pipenv run python $ROOT_DIRECTORY/butler.py reproduce $original_args
```
oliverchang: We should run `pipenv sync --dev` before here to update any dependencies that might have changed.

reproduce.sh
mbarbella-chromium: Will fix this and the other comments in another change I'm putting together now.

PR #1699, opened 2020-04-01
Fix launcher rebase to worker root.

inferno-chromium:
Rebase has to happen on direct path and not one with interpreter.
Add interpreter path (e.g. python) at the point of command line
calculation.

Fixes exception
```
AssertionError: Bad relative path ../../../../../usr/local/bin/python3.7 /mnt/scratch0/bots/oss-fuzz-linux-zone1-host-high-end-1fr1-7/clusterfuzz/bot/inputs/fuzzers/graphicsfuzz_spirv_fuzzer/spirv_launcher.py
```

### DIFF
--- a/src/python/bot/tasks/setup.py
+++ b/src/python/bot/tasks/setup.py
@@ -547,9 +547,8 @@ def update_fuzzer_and_data_bundles(fuzzer_name):
 
   # Setup environment variable for launcher script path.
   if fuzzer.launcher_script:
-    fuzzer_launcher_path = shell.get_execute_command(
-        os.path.join(fuzzer_directory, fuzzer.launcher_script),
-        is_blackbox_fuzzer=True)
+    fuzzer_launcher_path = os.path.join(fuzzer_directory,
+                                        fuzzer.launcher_script)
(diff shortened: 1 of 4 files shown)


src/python/bot/tasks/setup.py
```diff
@@ -547,9 +547,8 @@ def update_fuzzer_and_data_bundles(fuzzer_name):
 
   # Setup environment variable for launcher script path.
   if fuzzer.launcher_script:
-    fuzzer_launcher_path = shell.get_execute_command(
-        os.path.join(fuzzer_directory, fuzzer.launcher_script),
-        is_blackbox_fuzzer=True)
+    fuzzer_launcher_path = os.path.join(fuzzer_directory,
```
oliverchang, line 550: this change will probably also require a change in process_handler.py due to:

https://github.com/google/clusterfuzz/blob/d99cfb0e3710c46fb9322ce0d244de8070cf51b8/src/python/system/process_handler.py#L310

inferno-chromium: Good catch, fixed!

PR #1701, opened 2020-04-01
Allow specifying fuzzer name during testcase uploads.

oliverchang:
Necessary for launcher scripts to work.

### DIFF
--- a/src/appengine/handlers/upload_testcase.py
+++ b/src/appengine/handlers/upload_testcase.py
@@ -43,6 +43,7 @@ from libs import helpers
 from libs.issue_management import issue_tracker_utils
 from libs.query import datastore_query
 from system import archive
+from system import environment
 
 MAX_RETRIES = 50
 RUN_FILE_PATTERNS = ['run', 'fuzz-', 'index.', 'crash.']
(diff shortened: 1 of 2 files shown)


src/appengine/handlers/upload_testcase.py
```diff
@@ -165,6 +166,17 @@ def filter_target_names(targets, engine):
   return [t[len(prefix):] for t in targets if t.startswith(prefix)]
 
 
+def filter_blackbox_fuzzers(fuzzers):
+  """Filter out fuzzers such that only blackbox fuzzers are included."""
+
+  def is_greybox_fuzzer(name):
+    # TODO(ochang): Generalize this check.
+    return (name.startswith('libFuzzer') or name.startswith('afl') or
```
mbarbella-chromium, line 174: Do we have a constant anywhere else that would cover these three? This seems like something we'd easily forget to update when new things are added.

inferno-chromium: i thought we had a builtin_fuzzers that we can check for names ?

src/appengine/handlers/upload_testcase.py
```diff
@@ -221,10 +233,14 @@ def get(self):
...
     self.render(
-        'upload.html', {
+        'upload.html',
+        {
             'fieldValues': {
+                'blackboxFuzzers':
+                    filter_blackbox_fuzzers(allowed_fuzzers),
                 'jobs':
                     allowed_jobs,
+                # TODO(unassigned): Add honggfuzz.
```
inferno-chromium, line 243: lets just add it, it is 2 line change here and another 2 line one in appengine/private/components/upload-testcase/upload-form.html

src/appengine/handlers/upload_testcase.py
oliverchang: It's a mess right now because of engine vs builtin fuzzers. I will clean this all up once AFL is migrated.

src/appengine/handlers/upload_testcase.py
oliverchang: Done!

PR #1706, opened 2020-04-03
Update Windows startup script for Python 3.

oliverchang:


configs/test/gce/windows-init.ps1
```diff
@@ -266,31 +281,40 @@ $deploymentBucket = $webClient.DownloadString('http://metadata.google.internal/c
...
-} # !(Test-Path ($scriptExecutedFilePath))
+} # !(Test-Path ($packageSetupFilePath))
 
-Set-Content $scriptExecutedFilePath "skipped"
+Set-Content $packageSetupFilePath "Skipped package install"
 
 # Schedule chkdsk on every reboot.
 echo y | chkdsk C: /F /I /C
 
+# Install Pipfile dependencies
```
inferno-chromium, line 308: Any reason why these two blocks are not inside packageSetupFilePath ? They should be done once i think , not everytime.

oliverchang: This needs to happen after we download the source.

inferno-chromium: Makes sense as we change clusterfuzz source! LGTM

PR #1707, opened 2020-04-03
Remove Python 2 development dependencies.

oliverchang:
Also remove guard check for installed dependencies. There is no easy way
of doing this with pipenv.


oliverchang:
src/appengine/requirements.txt is needed as they contain App Engine specific requirements. 

src/requirements.txt is also still needed as well as Pipfile does not handle vendoring very well.

Keeping the Docker requirements.txt for now to avoid any Python 2 breakages in the meantime.

PR #1710, opened 2020-04-03
Migrate install_deps_macos.bash to python 3.

inferno-chromium:


docs/getting-started/prerequisites.md
```diff
@@ -19,7 +19,7 @@ This page explains how to set up your environment for using ClusterFuzz.
 Many features of ClusterFuzz depend on [Google Cloud
 Platform](https://cloud.google.com) services, but it's possible to run it locally without these dependencies for testing purposes. See the [Architecture page]({{ site.baseurl }}/architecture/#requirements) for more details.
 
-**Note:** Local development is only supported on **Linux** platform.
+**Note:** Local development is only supported on **Linux** and **macOS** platforms.
```
oliverchang, line 22: I think deployment is still broken, so not sure if we should list macOS as supported.

inferno-chromium: ok removed in https://github.com/google/clusterfuzz/commit/d221494d9278626027aa5af432867a623bc9eb8c
there are other issues too, need to add java in brewfile, also java path is not in PATH from homebrew, pain!

PR #1711, opened 2020-04-03
Update peach mutator to python3

mpherman2:


docker/base/requirements.txt
```diff
@@ -1,8 +1,11 @@
 crcmod==1.7
 future==0.17.1
+lxml==4.5.0
```
Dor1s, line 3: I'm not sure if this is needed after adding the same deps to `src/requirements.txt`

inferno-chromium: https://github.com/google/clusterfuzz/pull/1712

Dor1s: awesome!

oliverchang: These deps here are also deprecated and Python 2 only.

PR #1716, opened 2020-04-06
Fix an encoding error when downloading from fake GCS and local server.

oliverchang:


oliverchang:
/gcbrun. Affects local run_server only.

PR #1722, opened 2020-04-07
[fuchsia] Enable an example Rust fuzzer

eepeep:
Modifies the allowlist to let one of the example fuzzers written in Rust
run. This is so we can ensure that Rust crashes are handled properly.


eepeep:
cc @oliverchang

PR #1724, opened 2020-04-08
Implement an unshare runner for libFuzzer.

oliverchang:


src/python/system/new_process.py
```diff
@@ -376,3 +377,35 @@ def run_and_wait(self, *args, **kwargs):  # pylint: disable=arguments-differ
...
+    unshare_path = spawn.find_executable('unshare')
+    if not unshare_path:
+      raise RuntimeError('unshare not found')
+
+    command = [unshare_path]
+    command.extend([
+        '-U',  # Enter user namespace.
+        '--map-root-user',  # Map user to root user.
+        '-n',  # Enter network namespace.
+    ])
```
inferno-chromium, line 399: Does the permission propogate to child processes ?

src/python/system/new_process.py
```diff
@@ -376,3 +377,35 @@ def run_and_wait(self, *args, **kwargs):  # pylint: disable=arguments-differ
...
+    if environment.platform() != 'LINUX':
+      raise RuntimeError('UnshareProcessRunner only supported on Linux')
+
+    unshare_path = spawn.find_executable('unshare')
+    if not unshare_path:
+      raise RuntimeError('unshare not found')
+
+    command = [unshare_path]
+    command.extend([
+        '-U',  # Enter user namespace.
```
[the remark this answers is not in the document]
oliverchang: "enter" is the terminology for namespaces. see e.g. from http://man7.org/linux/man-pages/man1/unshare.1.html:

> Set the user ID which will be used in the **entered** namespace.

src/python/system/new_process.py
oliverchang: Yep!

PR #1728, opened 2020-04-08
[Reproduce tool] Allow APP_NAME overrides and other fixes.

mbarbella-chromium:
PTAL

### DIFF
--- a/src/local/butler/guard.py
+++ b/src/local/butler/guard.py
@@ -24,8 +24,12 @@ def check_virtualenv():
   root_path = os.path.realpath(
       os.path.join(os.path.dirname(__file__), '..', '..', '..'))
   is_in_virtualenv = bool(os.getenv('VIRTUAL_ENV'))
+  is_reproduce_tool = bool(os.getenv('REPRODUCE_TOOL'))
 
-  if not is_in_virtualenv:
+  # Check that we're in a virtual enviornment unless we're running the
+  # reproduce tool. The tool's method of running does not set the VIRTUAL_ENV
+  # environment variable.
+  if not is_in_virtualenv and not is_reproduce_tool:
     raise Exception(
         'You are not in a virtual env environment. Please install it with'
         ' `./local/install_deps.bash` or load it with'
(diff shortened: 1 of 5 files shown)


src/local/butler/reproduce.py
```diff
@@ -257,11 +257,15 @@ def _verify_target_exists(build_directory):
             build_directory=build_directory))
 
 
-def _update_environment_for_testcase(testcase, build_directory):
+def _update_environment_for_testcase(testcase, build_directory, application):
```
inferno-chromium, line 260: application_override sounds better here. in args.application, that seems ok.

src/local/butler/guard.py
```diff
@@ -24,8 +24,12 @@ def check_virtualenv():
   root_path = os.path.realpath(
       os.path.join(os.path.dirname(__file__), '..', '..', '..'))
   is_in_virtualenv = bool(os.getenv('VIRTUAL_ENV'))
+  is_reproduce_tool = bool(os.getenv('REPRODUCE_TOOL'))
 
-  if not is_in_virtualenv:
+  # Check that we're in a virtual enviornment unless we're running the
+  # reproduce tool. The tool's method of running does not set the VIRTUAL_ENV
+  # environment variable.
+  if not is_in_virtualenv and not is_reproduce_tool:
```
inferno-chromium, line 32: I dont understand why this is an exception for reproduce tool, butler.py reproduce is run in pipenv, see

$ pipenv run printenv | grep VIRTUAL
VIRTUAL_ENV=/usr/local/google/home/aarya/.local/share/virtualenvs/clusterfuzz-qlNIpp8O

mbarbella-chromium: Did you run this from within a pipenv shell? I don't see the same result.

src/local/butler/reproduce.py
```diff
@@ -257,11 +257,15 @@ def _verify_target_exists(build_directory):
...
 
 
-def _update_environment_for_testcase(testcase, build_directory):
+def _update_environment_for_testcase(testcase, build_directory, application_override):
   """Update environment variables that depend on the test case."""
   commands.update_environment_for_job(testcase.job_definition)
   environment.set_value('JOB_NAME', testcase.job_type)
 
+  # Override app name if explicitly specified.
+  if application:
```
inferno-chromium, line 266: s/application/application_override

PR #1731, opened 2020-04-10
Fix two-step merge breakage for Android libFuzzer.

inferno-chromium:
device_merge_control_file was not set.


inferno-chromium:
TBR, simple typo fix.

PR #1734, opened 2020-04-10
Enforce max_len in corpus pruning task (follow up #1616).

Dor1s:
Otherwise the large elements living in GCS (or added there manually) will remain.

### DIFF
--- a/src/python/bot/tasks/corpus_pruning_task.py
+++ b/src/python/bot/tasks/corpus_pruning_task.py
@@ -84,8 +84,8 @@ RSS_LIMIT_MB_FLAG = '-rss_limit_mb=%d'
 # Flag to enforce length limit for a single corpus element.
 MAX_LEN_FLAG = '-max_len=%d'
 
-# Flag to disable leak checking.
-DISABLE_LEAK_CHECK_FLAG = '-detect_leaks=0'
+# Flag to control memory leaks detection.
+DETECT_LEAKS_FLAG = '-detect_leaks=%d'
 
 # Flag to do value profile during merges.
 USE_VALUE_PROFILE_FLAG = '-use_value_profile=%d'
@@ -343,6 +343,7 @@ class Runner(object):
     """Get default libFuzzer options."""
     rss_limit = RSS_LIMIT
     max_len = engine_common.CORPUS_INPUT_SIZE_LIMIT
+    detect_leaks = 1
     arguments = [TIMEOUT_FLAG]
 
     if self.fuzzer_options:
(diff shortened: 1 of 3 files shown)


Dor1s:
Haven't found any tests for this, let me see if GCB fails, maybe there is something I missed. Otherwise will have to add a new test.

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -338,25 +341,35 @@ def __init__(self, build_directory, context):
 
   def get_libfuzzer_flags(self):
```
oliverchang, line 342: we should find a way to merge this with https://github.com/google/clusterfuzz/blob/32ce4d06bf75dabff18f009b5eacd3ae88e7fc51/src/python/bot/fuzzers/libFuzzer/fuzzer.py#L31.

The detect_leaks bit is also missing there.

src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -81,8 +81,11 @@
 RSS_LIMIT = 2560
 RSS_LIMIT_MB_FLAG = '-rss_limit_mb=%d'
 
-# Flag to disable leak checking.
-DISABLE_LEAK_CHECK_FLAG = '-detect_leaks=0'
+# Flag to enforce length limit for a single corpus element.
+MAX_LEN_FLAG = '-max_len=%d'
```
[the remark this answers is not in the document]
Dor1s: yep, there even seems to be a todo for that https://github.com/google/clusterfuzz/blob/e0c6ad071951c403d17ef11a3020ca75b4bf9a09/src/python/bot/tasks/corpus_pruning_task.py#L52

src/python/bot/tasks/corpus_pruning_task.py
Dor1s: The libFuzzer's one uses all the options provided in the `.options` file, with hich value sanitization for `timeout` and `rss_limit_mb`. The pruning one only extracts certain known options, nothing else.

I'm not sure if we want to use all the options for pruning as well. I mean, it seems reasonable in general, but I recall we had some targets that set a particular value for the `-runs` option (in Chromium, due to some threading issue), I don't think we would want that to be used in the pruning, so I guess there might be some other corner cases.

OTOH it's probably fault of the fuzz target developer who added some unlucky flag to the `.options` file, so maybe we shouldn't care. Having `close_fd_mask` to be turned out during pruning would be definitely better than not having it, for an opposite example.

PR #1735, opened 2020-04-10
Android libfuzzer fixes

inferno-chromium:
- Clean device and local directory before copy. Otherwise, we
   can keep stale files.
- For merge control dir, use sync_directories, so that it is
   synced back to local after merge.


inferno-chromium:
Thanks @Dor1s, fix worked. Thanks for spotting the problem.

PR #1739, opened 2020-04-14
Actually retry in google_cloud_utils/storage.py

### LINKED ISSUE #1737: Retry logic in storage.py
Filing this to clarify whether I misunderstand it or we need to fix something.

Once in a while we get the following exception: https://console.cloud.google.com/errors/CIuh6-K0uKWFNw?time=P30D&project=clusterfuzz-external

It comes from https://github.com/google/clusterfuzz/blob/e0c6ad071951c403d17ef11a3020ca75b4bf9a09/src/appengine/handlers/cron/corpus_backup.py#L65

Which calls into https://github.com/google/clusterfuzz/blob/e0c6ad071951c403d17ef11a3020ca75b4bf9a09/src/python/google_cloud_utils/storage.py#L786

which calls into https://github.com/google/clusterfuzz/blob/e0c6ad071951c403d17ef11a3020ca75b4bf9a09/src/python/google_cloud_utils/storage.py#L253

and the particular failing operation is https://github.com/google/clusterfuzz/blob/e0c6ad071951c403d17ef11a3020ca75b4bf9a09/src/python/google_cloud_utils/storage.py#L263

If I check the logs, I see the request that was sent prior to the copy operation (setting ACL): https://console.cloud.google.com/logs/viewer?resource=gae_app%2Fmodule_id%2Fcron-service%2Fversion_id%2F20200410t180004&logName=projects%2Fclusterfuzz-external%2Flogs%2Fapp&expandAll=false&timestamp=2020-04-12T03%3A26%3A25.013000000Z&project=clusterfuzz-external&minLogLevel=0&customFacets&limitCustomFacetWidth=true&advancedFilter=resource.type%3D%22gae_app%22%0Aresource.labels.project_id%3D%22clusterfuzz-external%22%0Aresource.labels.version_id%3D%2220200410t180004%22%0Aresource.labels.module_id%3D%22cron-service%22%0Atimestamp%3D%222020-04-12T03%3A25%3A44.899340000Z%22%0AinsertId%3D%22gddy5ufk382zt%22&interval=JUMP_TO_TIME&scrollTimestamp=2020-04-12T03%3A25%3A55.876537987Z

and that request happens 10 second earlier than when the copy error is recorded, even though we use the following constants:

```
# The number of retries to perform some GCS operation.
DEFAULT_FAIL_RETRIES = 8

# The time to wait between retries while performing GCS operation.
DEFAULT_FAIL_WAIT = 2
```

and the time interval between the previous request and the error should be 14+ seconds at least, but actually even more given that the copy error is `google.api_core.exceptions.GatewayTimeout` which is also a few seconds at least?

**Question**

Does the retry logic actually work? How can I verify it? Or do I miss something important in my debugging?

oliverchang:
Re-raise any exceptions for the retry to actually work.

Fixes #1737

### DIFF
--- a/src/python/google_cloud_utils/storage.py
+++ b/src/python/google_cloud_utils/storage.py
@@ -164,7 +164,7 @@ class GcsProvider(StorageProvider):
       client.buckets().insert(project=project_id, body=request_body).execute()
     except HttpError as e:
       logs.log_error('Failed to create bucket %s: %s' % (name, e))
-      return False
+      raise
 
     return True
 
@@ -223,7 +223,7 @@ class GcsProvider(StorageProvider):
     except google.cloud.exceptions.GoogleCloudError:
       logs.log_error('Failed to copy cloud storage file %s to local file %s.' %
                      (remote_path, local_path))
-      return False
+      raise
 
     return True
 
@@ -246,7 +246,7 @@ class GcsProvider(StorageProvider):
     except google.cloud.exceptions.GoogleCloudError:
       logs.log_error('Failed to copy local file %s to cloud storage file %s.' %
                      (local_path_or_handle, remote_path))
-      return False
+      raise
 
     return True
 
@@ -264,7 +264,7 @@ class GcsProvider(StorageProvider):
     except google.cloud.exceptions.GoogleCloudError:
       logs.log_error('Failed to copy cloud storage file %s to cloud storage '
                      'file %s.' % (remote_source, remote_target))
-      return False
+      raise
 
     return True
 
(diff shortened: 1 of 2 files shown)


src/python/google_cloud_utils/storage.py
```diff
@@ -294,7 +294,7 @@ def write_data(self, data, remote_path, metadata=None):
       blob.upload_from_string(data)
     except google.cloud.exceptions.GoogleCloudError:
       logs.log_error('Failed to write cloud storage file %s.' % remote_path)
-      return False
+      raise
```
inferno-chromium, line 297: Since you are raising error here, should these log_error be log_warn ? and then let retry do the log_error when it completely failed ?

Dor1s: yeah, good point!

oliverchang: Done!

PR #1743, opened 2020-04-15
fuchsia: Return symbolizer output as a string

nopsledder:
Python3 if more opinionated about regex type matching than Python2. The
output of the FuchsiaQemuLiFuzzerRunner's fuzz() result needs to be a
string to be handled properly downstream.


oliverchang:
Thanks! I have a slightly more robust fix here: https://github.com/google/clusterfuzz/pull/1744 which handles unicode errors and extends that to some additional places.

Closing in favour of that one.

PR #1746, opened 2020-04-21
Fix another Fuchsia encoding issue.

oliverchang:


oliverchang:
TBR. /gcbrun @eepeep FYI

Issue #1750, opened 2020-04-27
Important: Chromium Linux bots are hanging randomly in fuzzing task after py2->3 migration


inferno-chromium:
E.g. clusterfuzz-linux-0d16
clusterfuzz-linux-0188

This one is from clusterfuzz-linux-0d16
```
sudo pystack -v 411278
Standard Output:
[New LWP 411290]
[New LWP 411335]
[New LWP 411362]
[New LWP 411363]
[New LWP 411630]
[New LWP 411631]
[New LWP 411632]
[New LWP 411633]
[New LWP 411634]
[New LWP 411635]
[New LWP 415632]
[New LWP 555768]
[New LWP 1073068]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
0x00007f554929698d in pthread_join (threadid=140003423549184, thread_return=0x0) at pthread_join.c:90
$1 = (void *) 0x0

Standard Error:
90	pthread_join.c: No such file or directory.

Dumping Threads....


  File "/usr/local/lib/python3.7/threading.py", line 890, in _bootstrap
    self._bootstrap_inner()
  File "/usr/local/lib/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.7/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/lib/python3.7/concurrent/futures/thread.py", line 78, in _worker
    work_item = work_queue.get(block=True)

---------------

  File "/usr/local/lib/python3.7/threading.py", line 890, in _bootstrap
    self._bootstrap_inner()
  File "/usr/local/lib/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.7/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/lib/python3.7/concurrent/futures/thread.py", line 78, in _worker
    work_item = work_queue.get(block=True)

---------------

  File "/usr/local/lib/python3.7/threading.py", line 890, in _bootstrap
    self._bootstrap_inner()
  File "/usr/local/lib/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.7/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
[...]

PR #1755, opened 2020-04-28
Set multiprocessing start method to 'spawn'. …

oliverchang:
This makes things consistent across our platforms and may prevent fork
issues with gRPC (#1750)

### DIFF
--- a/src/python/bot/startup/run_bot.py
+++ b/src/python/bot/startup/run_bot.py
@@ -37,6 +37,7 @@ from future import standard_library
 standard_library.install_aliases()
 from builtins import object
 import os
+import multiprocessing
 import time
 import traceback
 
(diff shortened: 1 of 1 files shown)


src/python/bot/startup/run_bot.py
```diff
@@ -189,6 +190,10 @@ def main():
 
 
 if __name__ == '__main__':
+  if sys.version_info.major == 3:
+    # TODO(ochang): Remove check once all migrated to Python 3.
+    multiprocessing.set_start_method('spawn')
```
inferno-chromium, line 195: Will this impact OSS-Fuzz bots ? I was thinking on process_handler.get_process ? Thoughts ?

oliverchang: I guess this will impact JS fuzzers there, but there shouldn't be negative consequences. Grey box fuzzers don't use multiprocessing.Process there.

PR #1767, opened 2020-05-07
Upgrade dependencies for Python 3.8 support.

oliverchang:
Tensorflow 2.x had some API changes. Used tf_upgrade_v2 for these.

Also disable ml_train_task_test, as training depends on modules in tensorflow.contrib which
no longer exist.


oliverchang:
Also had to disable ml train tests (training bots are broken right now anyway), as they require more involved migrations. /gcbrun

PR #1770, opened 2020-05-08
Add crash address parsing for GDB.

mspectorgoogle:


src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -2867,3 +2867,29 @@ def test_python_unhandled_exception(self):
...
+                                  
+  def test_gdb_sigtrap(self):
+    """Test for GDB stack."""
+    data = self._read_test_data(
+        'gdb_sigtrap.txt')
+    expected_type = 'SIGTRAP'
+    expected_address = '0x000000000ac8'
+    expected_state = ('__aeabi_memcpy\nxymodem_trnasfer\nLoadImageFromUsb30\n')
+    expected_stacktrace = data
+    expected_security_flag = True
```
inferno-chromium, line 2879: This seems incorrect since the crash address is near null, expected security flag should be False. I think this is because of needing to add that signature in
https://github.com/google/clusterfuzz/blob/191ee1fce1351c0016abbae1186d0792de83aeae/src/python/crash_analysis/crash_analyzer.py#L38 and then it will come out as False.

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -2867,3 +2867,29 @@ def test_python_unhandled_exception(self):
...
+    self._validate_get_crash_data(data, expected_type, expected_address,
+                                  expected_state, expected_stacktrace,
+                                  expected_security_flag)
+                                  
+  def test_gdb_sigtrap_and_libfuzzer(self):
+    """Test for GDB stack with libfuzzer."""
+    data = self._read_test_data(
+        'gdb_sigtrap_and_libfuzzer.txt')
+    expected_type = 'SIGTRAP'
+    expected_address = '0x000000000ac8'
```
inferno-chromium, line 2889: In one test/stacktrace, increase the address to a high one like 
0x007000000ac8,
then security flag should be True as expected. That way we test for both cases.

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -2867,3 +2867,29 @@ def test_python_unhandled_exception(self):
...
                                   expected_state, expected_stacktrace,
                                   expected_security_flag)
+                                  
+  def test_gdb_sigtrap(self):
+    """Test for GDB stack."""
+    data = self._read_test_data(
+        'gdb_sigtrap.txt')
+    expected_type = 'SIGTRAP'
+    expected_address = '0x000000000ac8'
+    expected_state = ('__aeabi_memcpy\nxymodem_trnasfer\nLoadImageFromUsb30\n')
```
inferno-chromium, line 2877: __aeabi_* should not in crash state since memcpy is not useful for deduplication, can you add that __aeabi_ to ignore list here - https://github.com/google/clusterfuzz/blob/master/src/python/crash_analysis/stack_parsing/stack_analyzer.py#L363

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
inferno-chromium: Maybe this requires this
https://github.com/google/clusterfuzz/blob/master/src/python/crash_analysis/crash_analyzer.py#L360
instead of return true,
int_address = address_to_integer(crash_address)
return is_null_dereference(int_address)

mspectorgoogle: This is tough because the addresses for PC are all very low so it will  set the security flag to false incorrectly

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -132,6 +132,7 @@
     r'cannot open shared object file')
 LINUX_GDB_CRASH_TYPE_REGEX = re.compile(r'Program received signal ([a-zA-Z]+),')
 LINUX_GDB_CRASH_ADDRESS_REGEX = re.compile(r'rip[ ]+([xX0-9a-fA-F]+)')
+LINUX_GDB_CRASH_ADDRESS_NO_REGISTERS_REGEX = re.compile(r'^([xX0-9a-fA-F]+)\s+in\s+')
```
inferno-chromium, line 135: Since this regex is too generic, like it can match "abc in def", can you change to like
re.compile(r'^0x([0-9a-fA-F]+)\s+in\s+')

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
mspectorgoogle: We discussed we will do this in a separate CL because these addresses will be near null but they may be security crashes still.

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
mspectorgoogle: We will fix this in a separate cl

PR #1787, opened 2020-05-16
Minor syzkaller code cleanup.

inferno-chromium:


src/python/bot/fuzzers/syzkaller/runner.py
```diff
@@ -118,7 +120,7 @@ def fuzz(self,
           visited.add(unique_crash)
           log_lines = utils.read_data_from_file(
               os.path.join(subdir, file), eval_data=False)
-          fuzz_result.output = str(log_lines)
+          fuzz_logs += str(log_lines) + '\n'
```
inferno-chromium, line 123: fuzz_logs should have logs from all crashes. and engine.Crash is fine as-is with log_lines from each crash.

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -122,9 +122,9 @@ def reproduce(self, target_path, input_path, arguments, max_time):
     Returns:
       A ReproduceResult.
     """
-    binary_path = self.prepare_binary_path()
+    binary_dir = self.prepare_binary_path()
```
oliverchang, line 125: nit: also rename prepare_binary_path ?

inferno-chromium: It actually does prepare binary path by fixing execute perms on the binary. returning binary dir is just a helper.

PR #1793, opened 2020-05-20
OSS-Fuzz: Lower cpu weight for memory safe languages.

inferno-chromium:


src/appengine/handlers/cron/project_setup.py
```diff
@@ -72,6 +72,10 @@
 
 PUBSUB_PLATFORMS = ['linux']
 
+MEMORY_SAFE_LANGUAGES = ['go', 'java', 'python', 'rust']
```
jonathanmetzman, line 75: Since there are fewer memory unsafe languages than memory safe ones, why don't we just use that instead?

src/appengine/handlers/cron/project_setup.py
```diff
@@ -527,10 +532,16 @@ def create_project_settings(project, info, service_account):
       oss_fuzz_project.ccs = ccs
       oss_fuzz_project.put()
   else:
+    cpu_weight = (
```
jonathanmetzman, line 535: I would convert this to a normal `if-else` the one line version is not very readable when it's more than one line.

src/appengine/handlers/cron/project_setup.py
jonathanmetzman, line 75: Why not make this a set?

inferno-chromium: It is better to be explicit on which memory unsafe languages we are lowering the weight down for.

src/appengine/handlers/cron/project_setup.py
inferno-chromium: Done!

src/appengine/handlers/cron/project_setup.py
inferno-chromium: Done!

PR #1795, opened 2020-05-20
Fixing regression in testcase support for syzkaller

hzawawy:


src/python/system/environment.py
```diff
@@ -492,7 +492,7 @@ def get_memory_tool_name(job_name):
 
   if platform() == 'ANDROID_KERNEL':
     from platforms import android
-    if 'KASAN' in android.settings.get_product_name():
+    if 'KASAN' in  get_value('JOB_NAME'):
```
inferno-chromium, line 495: Actually you can remove this block entirely 493-496. Just have kasan in lowercase in your job name and then check above at 489 just works.

hzawawy: done.

PR #1798, opened 2020-05-27
Add ignore frames for Fuchsia.

oliverchang:


src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -511,8 +511,12 @@
     r'^panic$',
     r'^runtime\.',
 
-    # Fuchsia libFuzzer specific.
+    # Fuchsia specific.
     r'^CrashTrampolineAsm',
+    r'^libc_io_functions_not_implemented_use_fdio_instead',
+    r'^\<libclang_rt.asan.so\>',
+    r'^__zx_panic',
```
inferno-chromium, line 518: nit: for python 2, _ needs to be escaped in all these entries.

oliverchang: Done!

PR #1799, opened 2020-05-27
Fix exception output in run_bot.task_loop.

oliverchang:
Otherwise we may not terminate as expected in some cases.


oliverchang:
TBR. not sure when this broke, as this part hasn't been changed in ages. 

This is needed for this bit to work:

```python
    should_terminate = (
        clean_exit or errors.error_in_list(error_stacktrace,
                                           errors.BOT_ERROR_TERMINATION_LIST))
```

PR #1802, opened 2020-05-29
Update Android KASAN parsing.

mspectorgoogle:
This adds modifies Android KASAN parsing to properly get the crash
address and crash state.

### DIFF
--- a/src/python/crash_analysis/stack_parsing/stack_analyzer.py
+++ b/src/python/crash_analysis/stack_parsing/stack_analyzer.py
@@ -115,8 +115,12 @@ HWASAN_ALLOCATION_TAIL_OVERWRITTEN_ADDRESS_REGEX = re.compile(
     r'heap object \[([xX0-9a-fA-F]+),.*of size')
 JAVA_EXCEPTION_CRASH_STATE_REGEX = re.compile(r'\s*at (.*)\(.*\)')
 KASAN_ACCESS_TYPE_REGEX = re.compile(r'(Read|Write) of size ([0-9]+)')
+KASAN_ACCESS_TYPE_ADDRESS_REGEX = re.compile(
+    r'(Read|Write) of size ([0-9]+) at (addr|address) ([a-f0-9]+)')
 KASAN_CRASH_TYPE_ADDRESS_REGEX = re.compile(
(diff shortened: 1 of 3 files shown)


src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -1560,11 +1563,23 @@ def get_crash_data(crash_data, symbolize_flag=True):
 
     # For KASan crashes, additional information about a bad access may come
     # from a later line. Update the type if this happens.
+    if is_kasan and not state.crash_type:
```
inferno-chromium, line 1566: We should not do this as this would mean we match any line and change crash type to 'Kernel failure'. I think 
this is a new KASAN pattern. So, similar to KASAN_CRASH_TYPE_ADDRESS_REGEX call, add a KASAN_CRASH_TYPE_REGEX, to account for your example "BUG: KASAN: null-ptr-deref in sockfs_setattr+0x68/0x90 net/socket.c:544" example.

update_state_on_match(
        KASAN_CRASH_TYPE_REGEX,
        line,	
        state,
        new_type='Kernel failure',
)

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -1578,10 +1593,16 @@ def get_crash_data(crash_data, symbolize_flag=True):
     # For KASan crashes, additional information about a bad access may come
     # from a later line. Update the type if this happens.
     if state.crash_type.startswith('Kernel failure'):
-      kasan_access_match = KASAN_ACCESS_TYPE_REGEX.match(line)
+      kasan_access_match = KASAN_ACCESS_TYPE_ADDRESS_REGEX.match(line)
```
oliverchang, line 1596: nit: move to a helper function like `update_kasan_crash_details`.

mspectorgoogle:
Why does pylint on travis complain but my pylint doesn't seems weird.  Looking at 1432, its not a long line

inferno-chromium:
> Why does pylint on travis complain but my pylint doesn't seems weird. Looking at 1432, its not a long line

I will take a look, unrelated to your cl anyway.

PR #1804, opened 2020-05-29
Skip crash state check for flaky stacks during libFuzzer minimization

inferno-chromium:
This matches blackbox fuzzer testcases where we ignore crash
signature matching if testcase was flaky on initial testcase.

### DIFF
--- a/src/python/bot/tasks/minimize_task.py
+++ b/src/python/bot/tasks/minimize_task.py
@@ -1194,13 +1194,16 @@ def _run_libfuzzer_tool(tool_name,
   security_flag = crash_result.is_security_issue()
   if (security_flag != testcase.security_flag or
       state.crash_state != expected_crash_state):
-    logs.log_warn('Ignoring unrelated crash.\n'
+    logs.log_warn('Unrelated crash occurred:\n'
                   'State: %s (expected %s)\n'
                   'Security: %s (expected %s)\n'
                   'Output: %s\n' %
                   (state.crash_state, expected_crash_state, security_flag,
                    testcase.security_flag, state.crash_stacktrace))
-    return None, None
+    if not testcase.flaky_stack:
+      logs.log_warn(
+          'Skipping minimization result due to different crash signature.')
+      return None, None
 
   with open(output_file_path, 'rb') as file_handle:
     minimized_keys = blobs.write_blob(file_handle)
--- a/src/python/bot/tasks/minimize_task.py
+++ b/src/python/bot/tasks/minimize_task.py
@@ -56,11 +56,13 @@ from system import process_handler
 from system import shell
 
 IPCDUMP_TIMEOUT = 60
-COMBINED_IPCDUMP_TIMEOUT = 60 * 3
+COMBINED_IPCDUMP_TIMEOUT = 3 * 60
 MAX_DEADLINE_EXCEEDED_ATTEMPTS = 3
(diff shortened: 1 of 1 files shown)


src/python/bot/tasks/minimize_task.py
```diff
@@ -1353,19 +1351,26 @@ def handle_unreproducible():
     if cleansed_testcase_path:
       current_testcase_path = cleansed_testcase_path
 
+  # Store the minimized testcase.
+  testcase = data_handler.get_testcase_by_id(testcase_id)
+  with open(current_testcase_path, 'rb') as file_handle:
+    minimized_keys = blobs.write_blob(file_handle)
```
inferno-chromium, line 1357: This avoids creating 6 blob entries instead of 1.

inferno-chromium:
Do you think this is the right thing to do ? Flaky stacks are controversial. libfuzzer minimizes without stack comparison.

src/python/bot/tasks/minimize_task.py
oliverchang: From memory this was intentional, so we can see minimize results earlier while task is still running. No strong opinions about keeping this though.

src/python/bot/tasks/minimize_task.py
```diff
@@ -1194,19 +1191,16 @@ def _unset_dedup_flags():
...
                   (state.crash_state, expected_crash_state, security_flag,
                    testcase.security_flag, state.crash_stacktrace))
-    return None, None
-
-  with open(output_file_path, 'rb') as file_handle:
-    minimized_keys = blobs.write_blob(file_handle)
-
-  testcase.minimized_keys = minimized_keys
-  testcase.put()
+    if not testcase.flaky_stack:
```
oliverchang, line 1200: I don't think this is ever set for engine tests. As far as I can see, it's only set after calling check_for_initial_crash, which is not called in the engine case.

inferno-chromium:
Ok closing then.

evverx:
> Not sure if we should do this, as it can have worse consequences down the line

Another option would be not to use minimized testcases with flaky stacks down the line and just keep them somewhere along with the "minimized" testcases that are less likely to break anything in OSS-Fuzz so that projects like `systemd` that keep all the testcases that have ever triggered any issues in their repositories wouldn't have to minimize them manually every time. I mean, I already kind of manually deduplicate issues. I wouldn't want to also start minimize every testcase manually.

oliverchang:
> > Not sure if we should do this, as it can have worse consequences down the line
> 
> Another option would be not to use minimized testcases with flaky stacks down the line and just keep them somewhere along with the "minimized" testcases that are less likely to break anything in OSS-Fuzz so that projects like `systemd` that keep all the testcases that have ever triggered any issues in their repositories wouldn't have to minimize them manually every time. I mean, I already kind of manually deduplicate issues. I wouldn't want to also start minimize every testcase manually.

This would be a pretty involved change on ClusterFuzz side. In any case, this seems to address an edge case, as the majority of minimization should still work for non-flaky testcases. Is this not the case for systemd?

evverx:
> as the majority of minimization should still work for non-flaky testcases

As long as bugs aren't found with honggfuzz. Though I hope that issue will be addressed soon.

> Is this not the case for systemd?

To be honest, I don't know. Until I downloaded large testcases twice last week I wouldn't say I paid attention to the size of testcases.

PR #1806, opened 2020-06-01
Support multiple project setup configurations.

oliverchang:
- Make the project_setup section of project.yaml be a list instead.
- Add a config_suffix option as well to append to job names.

### DIFF
--- a/src/appengine/handlers/cron/project_setup.py
+++ b/src/appengine/handlers/cron/project_setup.py
@@ -933,7 +933,7 @@ class Handler(base_handler.Handler):
           BUILD_BUCKET_PATH_TEMPLATE,
           REVISION_URL,
           setup_config.get('build_type'),
-          config_suffix=setup_config.get('suffix', ''),
+          config_suffix=setup_config.get('job_suffix', ''),
           segregate_projects=setup_config.get('segregate_projects', False),
           engine_build_buckets={
               'libfuzzer': bucket_config.get('libfuzzer'),
--- a/src/appengine/handlers/cron/project_setup.py
+++ b/src/appengine/handlers/cron/project_setup.py
@@ -100,8 +100,9 @@ class JobInfo(object):
     self.experimental = experimental
     self.minimize_job_override = minimize_job_override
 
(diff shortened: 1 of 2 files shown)


src/appengine/handlers/cron/project_setup.py
```diff
@@ -919,47 +922,46 @@ def get(self):
...
-        additional_vars=project_setup_config.get('additional_vars'))
-
-    projects_source = project_setup_config.get('source')
-    if projects_source == 'oss-fuzz':
-      projects = get_oss_fuzz_projects()
-    elif projects_source.startswith(storage.GS_PREFIX):
-      projects = get_projects_from_gcs(projects_source)
-    else:
-      raise ProjectSetupError('Invalid projects source: ' + projects_source)
+    project_setup_configs = local_config.ProjectConfig().get('project_setup')
```
inferno-chromium, line 925: We probably need to change the example project.yaml file in configs/test directory ? maybe add example for suffix (commented) ?

src/python/tests/appengine/handlers/cron/project_setup_test.py
```diff
@@ -1607,41 +1623,66 @@ def setUp(self):
...
+                }
             },
-            'additional_vars': {
-                'all': {
-                    'STRING_VAR': 'VAL',
-                    'BOOL_VAR': True,
-                    'INT_VAR': 0,
+            {
+                'source': 'gs://bucket-dbg/projects.json',
+                'suffix': '_dbg',
```
inferno-chromium, line 1659: suffix feels too generic ? maybe name_suffix or config_suffix or something else ?

oliverchang: Renamed to job_suffix.

src/appengine/handlers/cron/project_setup.py
oliverchang: Hmm, I don't think we need to document this for now. Would make it more difficult to change the format (like I'm doing now) in the future. Can add this later if there is demand.

PR #1810, opened 2020-06-03
Switch to using json key format for build artifacts.

inferno-chromium:


inferno-chromium:
will remove build_apiary_service_account_email in another cl, will let this stabilize.

PR #1818, opened 2020-06-09
Extend crash comparer to catch some cases of shifting frames.

oliverchang:
If 2 out of 3 frames match (in the same order), consider the crash similar. This can catch
cases where a function is inlined in one build but not in another.

### DIFF
--- a/src/python/crash_analysis/crash_comparer.py
+++ b/src/python/crash_analysis/crash_comparer.py
@@ -53,9 +53,20 @@ def _similarity_ratio(string_1, string_2):
       1.0 * length_sum)
 
 
+def count_same_frames(first_frames, second_frames):
+  """Count number of frames which are the same (disregarding order)."""
+  same_count = 0
+  for frame in first_frames:
+    if frame in second_frames:
+      same_count += 1
+
+  return same_count
+
+
 class CrashComparer(object):
   """Compares two crash results."""
   COMPARE_THRESHOLD = 0.8
+  SAME_FRAMES_THRESHOLD = 2
 
   def __init__(self, crash_state_1, crash_state_2, compare_threshold=None):
     self.crash_state_1 = crash_state_1
@@ -81,6 +92,10 @@ class CrashComparer(object):
     # stacktrace.
     crash_state_lines_1 = self.crash_state_1.splitlines()
     crash_state_lines_2 = self.crash_state_2.splitlines()
+
+    if (count_same_frames(crash_state_lines_1, crash_state_lines_2) >=
+        self.SAME_FRAMES_THRESHOLD):
+      return True
 
     lines_compared = 0
     similarity_ratio_sum = 0.0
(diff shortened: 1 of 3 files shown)


src/python/crash_analysis/crash_comparer.py
```diff
@@ -53,9 +53,20 @@ def _similarity_ratio(string_1, string_2):
       1.0 * length_sum)
 
 
+def count_same_frames(first_frames, second_frames):
+  """Count number of frames which are the same (disregarding order)."""
```
inferno-chromium, line 57: We should account for order, the matches should be top-down. I can never think of a case of out-of-order and it decreases false positives.
also, i am worried about cases when assertion stack gets grouped with functional crashes [like two frames match]. Can you please run this over current open chromium, oss-fuzz crashes and see if the new groups look correct. What also happens sometimes is we overgroup, like a1 and a2 get group, and a2 and a3 get group, and unrelated a1, a3 get grouped (a1, a2, a3 all in one group). this overgroup case has happened a lot in past and needs to be monitored. The reason being that only one testcase is filed in overgroup, so we try to lower false positive as much as possible.

oliverchang: Re matching ordering, I can think of stack overflows here. At the same time I can't think of many cases where two frames match and it's a false positive, but I can add this to be safe.

Will see if I can generate new groups.

src/python/crash_analysis/crash_comparer.py
```diff
@@ -53,9 +53,27 @@ def _similarity_ratio(string_1, string_2):
       1.0 * length_sum)
 
 
+def longest_common_subsequence(first_frames, second_frames):
+  """Count number of frames which are the same (taking into account order)."""
+  first_len = len(first_frames)
+  second_len = len(second_frames)
+
+  solution = [[0 for _ in range(second_len + 1)] for _ in range(first_len + 1)]
```
inferno-chromium, line 61: shouldn't this be
[[0 for _ in range(second_len)] for _ in range(first_len)]
without the +1

oliverchang: The solution array needs +1 in both dimentions for the extra [0][0] as the base case.

src/python/tests/core/crash_analysis/crash_comparer_test.py
```diff
@@ -67,3 +67,21 @@ def test_is_not_similar(self):
...
+  def test_only_one_frame_matching(self):
+    """Test not similar when 1/3 frames match."""
+    crash_state_1 = ('first\n' 'second\n' 'third\n')
+    crash_state_2 = ('second\n' 'fourth\n' 'fifth\n')
+    self.is_similar_helper(crash_state_1, crash_state_2, False)
+
+  def test_is_same_frames_wrong_order(self):
+    """Test not similar when some frames match but the order doesn't match."""
+    crash_state_1 = ('first\n' 'second\n' 'third\n')
+    crash_state_2 = ('second\n' 'first\n' 'fourth\n')
```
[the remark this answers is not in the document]
oliverchang: Done! Also added a few more tests.

PR #1821, opened 2020-06-10
Decode byte fuzz output properly instead of str in syzkaller.

inferno-chromium:


src/python/bot/fuzzers/syzkaller/runner.py
```diff
@@ -119,8 +119,8 @@ def fuzz(self,
         if fnmatch.fnmatch(file, 'report*') and unique_crash not in visited:
           visited.add(unique_crash)
           log_lines = utils.read_data_from_file(
-              os.path.join(subdir, file), eval_data=False)
-          fuzz_logs += str(log_lines) + '\n'
+              os.path.join(subdir, file), eval_data=False).decode('utf-8')
```
inferno-chromium, line 122: This is a bug, you can see an example log
gsutil cat gs://fuzz-logs.internal.clusterfuzz.com/syzkaller_syzkaller/syzkaller_kasan_android_device/2020-06-10/03:11:36:378571.log

Basically,
>>> a = b'123'
>>> str(a)
"b'123'"
>>> a.decode('utf-8')
'123'

src/python/bot/fuzzers/syzkaller/runner.py
```diff
@@ -87,7 +86,7 @@ def repro(self, repro_timeout, repro_args):
     result = self.run_and_wait(additional_args, timeout=repro_timeout)
     logs.log('Syzkaller testcase stopped.')
     return engine.ReproduceResult(result.command, result.return_code,
-                                  result.time_executed, str(result.output))
+                                  result.time_executed, result.output)
```
inferno-chromium, line 89: Similar to below, this should be wrong, as str shouldn't be called on a byte output. Do you remember where it failed ?

src/python/bot/fuzzers/syzkaller/runner.py
inferno-chromium: I forgot to tell utils.read_data_from_file gives bytes in my last review.

PR #1829, opened 2020-06-15
Cache fuzz target entities and job names in uploads.

inferno-chromium:
Speculative fix to save some datastore reads.

### DIFF
--- a/src/appengine/handlers/upload_testcase.py
+++ b/src/appengine/handlers/upload_testcase.py
@@ -28,6 +28,7 @@ import os
 from google.cloud import ndb
 
 from base import external_users
+from base import memoize
 from base import tasks
 from base import utils
 from datastore import data_handler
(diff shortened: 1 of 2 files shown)


src/appengine/handlers/upload_testcase.py
```diff
@@ -177,18 +179,19 @@ def is_greybox_fuzzer(name):
...
+  not be prefixed with project), and job."""
   project_name = data_handler.get_project_name(job_name)
   candidate_name = data_types.fuzz_target_fully_qualified_name(
       engine, project_name, target_name)
 
   target = data_handler.get_fuzz_target(candidate_name)
-  if target:
-    return target.fully_qualified_name(), target.binary
+  if not target:
+    raise helpers.EarlyExitException('Fuzz target does not exist.', 400)
```
inferno-chromium, line 192: This avoids memcache to save (None, None) tuple and we raise exception in caller anyway, so just moved it here.

PR #1835, opened 2020-06-18
Add a USE_PYTHON3_FOR_FUZZER variable.

mbarbella-chromium:
This is meant to simplify our transition to Python 3 for fuzzers in addition to ClusterFuzz itself. It will help us deprecate then eventually remove USE_PYTHON2_FOR_BLACKBOX_FUZZERS, and will be removed along with that variable once all fuzzers are updated.

### DIFF
--- a/src/python/system/shell.py
+++ b/src/python/system/shell.py
@@ -368,6 +368,7 @@ def get_interpreter(file_to_execute, is_blackbox_fuzzer=False):
   # TODO(mbarbella): Remove this when fuzzers have been migrated to Python 3.
   if (is_blackbox_fuzzer and interpreter == sys.executable and
       environment.get_value('USE_PYTHON2_FOR_BLACKBOX_FUZZERS') and
+      not environment.get_value('USE_PYTHON3_FOR_FUZZER') and
       sys.version_info.major == 3):
     interpreter = 'python2'
 


mbarbella-chromium:
PTAL when you have time. I know we mentioned having the variable defined the other way around (3 by default, 2 explicitly set), but I forgot that we're already explicitly setting a variable in our config to enable this behavior at all. This new variable doesn't seem too clean, but it's the easiest way for us to transition away from needing any special casing.

src/python/system/shell.py
```diff
@@ -368,6 +368,7 @@ def get_interpreter(file_to_execute, is_blackbox_fuzzer=False):
   # TODO(mbarbella): Remove this when fuzzers have been migrated to Python 3.
   if (is_blackbox_fuzzer and interpreter == sys.executable and
       environment.get_value('USE_PYTHON2_FOR_BLACKBOX_FUZZERS') and
+      not environment.get_value('USE_PYTHON3_FOR_FUZZER') and
```
inferno-chromium, line 371: We should remove USE_PYTHON2_FOR_BLACKBOX_FUZZERS from global and then add in environment attribute of all fuzzers
.environment += '\nUSE_PYTHON2_FOR_BLACKBOX_FUZZERS = True'
that way, nobody using clusterfuzz does not have to set any of these env vars and are on python 3 by default.

mbarbella-chromium: If we plan to set them per-fuzzer I'd rather just have an override for the interpreter. I'm still not totally sure this is worth it though, since the default is python 3 and other users won't run into this at all. Might be easier to discuss in person, but I'm fine with either option.

mbarbella-chromium:
While discussing this with Abhishek I realized that there's no reason to use a new variable here. We can just set "USE_PYTHON2_FOR_BLACKBOX_FUZZERS = False" in the job definitions for fuzzers we want to run with Python 3, and delete the check once all are converted as originally planned.

PR #1836, opened 2020-06-18
Remove python 2 requirements install from docker images.

inferno-chromium:


src/python/system/archive.py
```diff
@@ -295,14 +290,8 @@ def unpack(archive_path,
       logs.log_error(
           'Failed to extract everything from archive %s.' % archive_filename)
 
-  elif archive_type == ArchiveType.TAR or archive_type == ArchiveType.TAR_LZMA:
+  elif archive_type in [ArchiveType.TAR, ArchiveType.TAR_LZMA]:
```
oliverchang, line 293: nit: use a tuple instead of list. i.e. `c in (A, B)`.

inferno-chromium: Done!

PR #1839, opened 2020-06-19
Use bot local temp for TEST_TEMPDIR.

inferno-chromium:


inferno-chromium:
I thought about using FUZZ_INPUTS_DISK but right now we use it for testcases as well, this feels less hacky and gets cleared between iterations with clear_temp_directory.

PR #1840, opened 2020-06-20
Remove some datastore calls from Fuzzer, TestcaseUploadMetadata.

inferno-chromium:


inferno-chromium:
TBR, yet another set of CLs to decrease datastore calls in fuzz task.

PR #1841, opened 2020-06-21
Remove redundant fuzzer entity query call in fuzz task.

inferno-chromium:
Make setup.update_fuzzer_and_data_bundles return a fuzzer entity
on success and None on failure. This helps to remove a redundant
fuzzer entity query call.


inferno-chromium:
TBR minor datastore read reduction fix.

PR #1842, opened 2020-06-21
Don't do platform id check if running fuzz task.

inferno-chromium:
Saves datastore read of testcase entity since it does not apply
in fuzz task.


inferno-chromium:
TBR simple fix to reduce an obvious unnecessary datastore query.

inferno-chromium:
Revert to earlier fix #1 in https://github.com/google/clusterfuzz/pull/1843 since testcase entity needs to be defined for variant task.

PR #1846, opened 2020-06-22
Batch FuzzerJob entities to reduce Datastore read ops.

### LINKED ISSUE #1844: Optimize FuzzerJob selection.
Currently the FuzzerJob selection algorithm relies on a |Fuzzer| x |Job| query. For a large number of jobs, this results in a pretty big blowup in number of reads required.

We should make this such that we pick the Fuzzer first, then the Job to go with that fuzzer. This will reduce reads down to |Fuzzer| + |Job|.

oliverchang:
Batch FuzzerJobs in a cron, and have bots read the batched version
instead.

Fixes #1844

### DIFF
--- a/configs/test/gae/prod3/cron.yaml
+++ b/configs/test/gae/prod3/cron.yaml
@@ -55,6 +55,11 @@ cron:
     schedule: every day 03:00
     target: cron-service
 
+  - description: Batch FuzzerJobs.
+    url: /batch-fuzzer-jobs
+    schedule: every 30 minutes
+    target: cron-service
+
   - description: Adjust strategy selection weights based on strategy stats.
     url: /fuzz-strategy-selection
     schedule: every day 05:00
--- a/configs/test/gae/prod3/cron.yaml
+++ b/configs/test/gae/prod3/cron.yaml
@@ -13,16 +13,15 @@
 # limitations under the License.
 
 cron:
-  # Run fuzzer coverage task right before the corpus pruning task.
-  - description: Obtain code coverage information from the coverage GCS bucket.
-    url: /fuzzer-coverage
-    schedule: every day 18:30
-    target: cron-service
-
   # Python crons start here.
   - description: Backup
     url: /backup
     schedule: every day 01:47
+    target: cron-service
+
+  - description: Batch FuzzerJobs.
+    url: /batch-fuzzer-jobs
+    schedule: every 30 minutes
     target: cron-service
 
   - description: Build crash stats.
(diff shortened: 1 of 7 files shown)


src/python/fuzzing/fuzzer_selection.py
```diff
@@ -80,10 +80,13 @@ def get_fuzz_task_payload(platform=None):
     queue_override = environment.get_value('QUEUE_OVERRIDE')
     platform = queue_override if queue_override else environment.platform()
 
-  query = data_types.FuzzerJob.query()
-  query = query.filter(data_types.FuzzerJob.platform == platform)
+  query = data_types.FuzzerJobs.query()
```
oliverchang, line 83: This can introduce a 30 min delay for a job to run which can be confusing when run locally. I can add a condition here to check for local server to use the individual FuzzerJobs instead, but that can be a bit hacky. Alternatively we can have something a bit smarter which flips some dynamic flag in Config based on how many FuzzerJobs we detect. @mbarbella-chromium @inferno-chromium WDYT?

oliverchang:
/gcbrun. PTAL.

Went with a batching approach here rather than converting everything to use the batch directly as that introduces complexities with races and querying/modification etc. It would also be a much more substantial change.

configs/test/gae/prod3/cron.yaml
```diff
@@ -55,6 +55,11 @@ cron:
     schedule: every day 03:00
     target: cron-service
 
+  - description: Batch FuzzerJobs.
+    url: /batch-fuzzer-jobs
```
inferno-chromium, line 59: i thought this list was sorted by url values, so maybe move it after /backup (might also move fuzzer-coverage while you are here).

src/python/fuzzing/fuzzer_selection.py
```diff
@@ -80,10 +80,13 @@ def get_fuzz_task_payload(platform=None):
...
+  query = query.filter(data_types.FuzzerJobs.platform == platform)
+
+  mappings = []
+  for fuzzer_jobs in query:
```
inferno-chromium, line 87: s/fuzzer_jobs/entity to avoid confusion with fuzzer_jobs.fuzzer_jobs.

src/python/fuzzing/fuzzer_selection.py
inferno-chromium: What about local development, maybe we need a bailout that does regular way for environment.is_local. for remote, waiting 30 min is fine.

inferno-chromium:
Thanks so much for doing this!

configs/test/gae/prod3/cron.yaml
oliverchang: Done!

src/python/fuzzing/fuzzer_selection.py
oliverchang: Just added an is_local check.

src/python/fuzzing/fuzzer_selection.py
oliverchang: Done.

PR #1849, opened 2020-06-22
Remove superfluous reads from the severity analyzer unit test

renatahodovan:
The analyze method of the SeverityAnalyzerSanitizer class doesn't use
its second parameter, however the tests in SeverityAnalyzerTest still
read files to fill this parameter, which is a waste of resources,
hence they are removed.

### DIFF
--- a/src/python/tests/core/crash_analysis/severity_analyzer_test.py
+++ b/src/python/tests/core/crash_analysis/severity_analyzer_test.py
@@ -35,23 +35,20 @@ class SeverityAnalyzerTest(unittest.TestCase):
     """Tests severity analysis of a use after free report."""
     self.assertEqual(
         severity_analyzer.get_analyzer('sanitizer_generic').analyze(
-            'Heap-use-after-free\nREAD 8', self._read_test_data('asan_uaf.txt'),
-            False), SecuritySeverity.HIGH)
+            'Heap-use-after-free\nREAD 8', None, False), SecuritySeverity.HIGH)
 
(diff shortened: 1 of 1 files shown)


oliverchang:
Thanks for the PR. However, I think it still makes sense to keep these as:

- The stacktrace is part of the API. Whether or not it's used is an implementation detail.
- If for example we had a bug where we were returning the wrong analyzer (e.g. Chrome's which does read this output), this would not be caught by the test.
- This is a very minimal use of resources.

renatahodovan:
@oliverchang Thanks for the review! I may have started to solve a potential issue from a wrong direction. What I originally spotted was that there were several unit tests in severity_analyzer_test that loaded wrong stacktraces. However, this was not caught by the test framework since they were passed as unused parameters:

1) In `test_asan_heap_overflow_write` asan_heap_overflow_write.txt should be loaded instead of asan_heap_overflow_read.txt.
2) In `test_asan_heap_overflow_write_gestures` asan_heap_overflow_write.txt should be loaded instead of asan_heap_overflow_read.txt.
3) In `test_asan_container_overflow_write` the loaded asan_container_overflow_read.txt doesn't seem to be fitting either, but there is no asan_container_overflow_write.txt output among the test outputs.

Should these inconsistencies be fixed or are they good as they are now?

oliverchang:
Ah, good point :) In that case, please feel free to upload a new PR with those fixes. Re 3. just ignore that one for now.

PR #1856, opened 2020-06-23
Fix bug with batch FuzzerJobs update removing existing jobs.

inferno-chromium:


inferno-chromium:
TBR to fix entity deletion

oliverchang:
Thanks! LGTM

PR #1862, opened 2020-06-24
Update ml-with-gpu image to use Python 3 pip.

oliverchang:
To match chromium/ml-with-gpu

We no longer install Python 2 dependencies.


oliverchang:
TBR to get builds going.

inferno-chromium:
LGTM

PR #1865, opened 2020-06-25
Speed up web fonts load.

inferno-chromium:
Fixes https://github.com/google/oss-fuzz/issues/3919


evverx:
I've just tried to visit that page and the browser still says that it's unresponsive
Screenshot 2020-06-25 at 20 13 23

Could it be that the change hasn't been rolled out yet?

inferno-chromium:
> I've just tried to visit that page and the browser still says that it's unresponsive
> Screenshot 2020-06-25 at 20 13 23
> 
> Could it be that the change hasn't been rolled out yet?

Still in deployment, should finish in 15-30 minutes. I don't see this page unresponsive thing on https://oss-fuzz.com/testcases, maybe some other bug with db data cache. can you check in network console if this is still the font wait ?

evverx:
Last time I checked it took the page about 15s to load so it seems to be working

Screenshot 2020-06-25 at 21 22 27

inferno-chromium:
Thanks so much @evverx for checking!

Issue #1864, opened 2020-06-25
install_deps.bash: support Ubuntu 20.04 LTS


l2dy:
Would Ubuntu 20.04 become a supported distro?

inferno-chromium:
Yes we want to do this.

@iamarshsingh - can you help with this. Just try making local/install_deps_linux.bash run correctly on a Ubuntu 20.04 docker container.

iamarshsingh:
Sure @inferno-chromium!

inferno-chromium:
Check out related 18.04 which we fixed some time back.
https://github.com/google/clusterfuzz/issues/1781

Basically python butler.py bootstrap
python butler.py run_server should work (localhost:9000 works).

inferno-chromium:
Should be fixed now, use master.

PR #1867, opened 2020-06-25
Added support for Ubuntu 20.04 and root usage

iamarshsingh:
Hi, Adding support for Ubuntu 20.04 and also allow script to work in root environment as by default bower and npm faces installation issues in sudo.
Tested changes in docker image of Ubuntu 20.04, 18.04 and 16.04.

Fixes issue #1864 

Please let me know if any changes are required.


local/install_deps_linux.bash
```diff
@@ -183,8 +184,8 @@ fi
 if [ ! $only_reproduce ]; then
   # Install other dependencies (e.g. bower).
   nodeenv -p --prebuilt
-  npm install -g bower polymer-bundler
-  bower install
+  npm install --unsafe-perm -g bower polymer-bundler
+  bower --allow-root install
```
iamarshsingh, line 188: These two flags allowed to support installation in root mode, they are not generally recommended. Can be removed too if we make it explicit to not run install deps as a root user.

inferno-chromium: Add a comment like "# Allow bower and polymer-bundler install for root users as well."

iamarshsingh: Sure!

inferno-chromium:
Thanks for the super quick fix!

PR #1869, opened 2020-06-26
Store past crashes in corpus regressions folder, test in pruning task.

inferno-chromium:
Also, remove _static urls which are unused and just increase unnecessary storage calls.

### DIFF
--- a/src/python/bot/tasks/progression_task.py
+++ b/src/python/bot/tasks/progression_task.py
@@ -234,9 +234,13 @@ def _store_testcase_for_regression_testing(testcase, testcase_file_path):
   regression_testcase_url = os.path.join(
       corpus.get_regressions_corpus_gcs_url(),
       os.path.basename(testcase_file_path))
-  storage.copy_file_to(testcase_file_path, regression_testcase_url)
-  logs.log('Successfully stored testcase for regression testing: ' +
-           regression_testcase_url)
+
+  if storage.copy_file_to(testcase_file_path, regression_testcase_url):
+    logs.log('Successfully stored testcase for regression testing: ' +
+             regression_testcase_url)
+  else:
+    logs.log_error('Failed to store testcase for regression testing: ' +
+                   regression_testcase_url)
 
 
 def find_fixed_range(testcase_id, job_type):
--- a/src/python/bot/tasks/progression_task.py
+++ b/src/python/bot/tasks/progression_task.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 """Test to see if test cases are fixed."""
 
+import os
 import six
 import time
 
@@ -28,7 +29,9 @@ from build_management import revisions
 from chrome import crash_uploader
 from datastore import data_handler
 from datastore import data_types
+from fuzzing import corpus_manager
 from google_cloud_utils import big_query
+from google_cloud_utils import storage
 from metrics import logs
 from system import environment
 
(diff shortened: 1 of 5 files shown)


src/python/bot/tasks/progression_task.py
```diff
@@ -210,6 +213,32 @@ def _save_fixed_range(testcase_id, min_revision, max_revision):
...
+    # No work to do, only applicable for engine fuzzers.
+    return
+
+  corpus = corpus_manager.FuzzTargetCorpus(fuzz_target.engine,
+                                           fuzz_target.project_qualified_name())
+  regression_testcase_url = os.path.join(
+      corpus.get_regressions_corpus_gcs_url(),
+      os.path.basename(testcase_file_path))
+  storage.copy_file_to(testcase_file_path, regression_testcase_url)
+  logs.log('Successfully stored testcase for regression testing: ' +
```
Dor1s, line 238: nit: move this under `if storage.copy_file_to(...):`

Dor1s: can also log a warning / error in case of a failure

src/python/fuzzing/corpus_manager.py
```diff
@@ -40,7 +40,7 @@
 CORPUS_FILES_SYNC_TIMEOUT = 60 * 60
 LATEST_BACKUP_TIMESTAMP = 'latest'
 PUBLIC_BACKUP_TIMESTAMP = 'public'
-STATIC_CORPUS_GCS_PATH_SUFFIX = '_static'
+REGRESSIONS_GCS_PATH_SUFFIX = '_regressions'
```
Dor1s, line 43: do we need to automatically create these buckets somehow?

src/python/bot/tasks/progression_task.py
inferno-chromium: Done!

src/python/fuzzing/corpus_manager.py
inferno-chromium: Nope, sub-dir get created automatically.

Dor1s: ah, right! awesome

src/python/bot/tasks/progression_task.py
```diff
@@ -210,6 +213,36 @@ def _save_fixed_range(testcase_id, min_revision, max_revision):
   task_creation.request_bisection(testcase, 'fixed')
 
 
+def _store_testcase_for_regression_testing(testcase, testcase_file_path):
```
jonathanmetzman, line 216: Are regressions made public right after they are fixed? If not how can we use them in CIFuzz?

inferno-chromium: This is not complete, i am thinking to do hooks into corpus backup cron that makes things public. Still thinking if these regressions should be part of corpus backup or its own archive/folder.

Dor1s: IMO backup is better, as that way we'll have a single good regression test suite for each fuzz target.

inferno-chromium: Yes i agree, @jonathanmetzman - wdyt ?

inferno-chromium: This is now done and working - https://storage.cloud.google.com/clusterfuzz-libfuzzer-backup/corpus/libfuzzer/grpc_api_fuzzer/latest.zip (see regressions sub-dir). @jonathanmetzman fyi!

PR #1870, opened 2020-06-26
Fix bug with wrong testcase open status.

inferno-chromium:
_store_testcase_for_regression_testing relies on testcase.open which is only updated in local version of testcase in _save_fixed_range.


inferno-chromium:
TBR minor fix.

PR #1874, opened 2020-07-01
Remove various newstr hacks.

mbarbella-chromium:


mbarbella-chromium:
Not yet ready for review.

mbarbella-chromium:
> Does this work on appengine staging ?

Haven't tested it yet, but will check before landing.

PR #1875, opened 2020-07-01
Remove explicit gc for ndb cleanup.

mbarbella-chromium:
This may no longer be necessary due to the Python 3 migration.

### DIFF
--- a/src/appengine/handlers/base_handler.py
+++ b/src/appengine/handlers/base_handler.py
@@ -21,7 +21,6 @@ standard_library.install_aliases()
 import base64
 import cgi
 import datetime
-import gc
 import json
 import logging
 import os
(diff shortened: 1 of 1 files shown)


mbarbella-chromium:
Agreed. I wanted to separate this one out so it would be easily revertible as it looks likely to cause trouble.

oliverchang:
Thanks! this should indeed be no longer required after https://github.com/googleapis/python-ndb/issues/336#issuecomment-611505353

inferno-chromium:
> Thanks! this should indeed be no longer required after [googleapis/python-ndb#336 (comment)](https://github.com/googleapis/python-ndb/issues/336#issuecomment-611505353)

We are using google-auth==1.14.3 which was released in may, so should have this fix. i think we are good to go then. Marty, can you test these two cls on staging and then can land.

Issue #780, opened 2019-08-07
Paginate jobs on /jobs page.


oliverchang:
Currently everything is rendered at once on /jobs, leading to very slow loads when there are many jobs/projects.

PR #1877, opened 2020-07-02
Added Pagination support to Jobs Page

iamarshsingh:
Update (6th July, 2020): Keywords search requires a re-puts for all jobs, so @oliverchang has suggested to make a migration command for the same. I have added the script in src/local/butler/scripts/migration and the instructions in Next section in Changelog.

Hi, This PR is to add pagination support to Jobs page(Issue #780). Along with that some other changes that are made:
1. Requests for search and page change will be using AJAX. Added a JsonHandler in page handler.
2. Search shifted from on-browser to server side using keywords present in name. (Due to this, the existing builds' jobs data should be modified as a new attribute is added to Job model)
3. Added new test for changed APIs.

I will share the design doc for new jobs dataflow and these changes in the core group.

Screenshots:
Number of pages are set to 5 by default, each page will have 10 jobs.
Screenshot 2020-07-02 at 12 03 57 PM

Please have a look and let me know if any changes are required!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 Change Log
 ==========
 
+## Next
+- Next version requires new DB migrations due to addition of search keywords in Jobs.
+- To perform migrations, please use:
+```
+python butler.py run -c=path/to/config --non-dry-run migration.jobs_keywords
+```
+
 ## Version 2.0.2
 - Improved Syzkaller support.
 - Support narrower bisection for regression/fix ranges.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ Change Log
 - Next version requires new DB migrations due to addition of search keywords in Jobs.
 - To perform migrations, please use:
 ```
-python butler.py run -c=path/to/config --non-dry-run migration.jobs_keywords
+python butler.py run -c path/to/config --non-dry-run migration.jobs_keywords
 ```
 
 ## Version 2.0.2
--- a/src/appengine/handlers/jobs.py
+++ b/src/appengine/handlers/jobs.py
@@ -24,10 +24,23 @@ from datastore import data_types
 from datastore import ndb_utils
 from fuzzing import fuzzer_selection
 from handlers import base_handler
+from libs import filters
 from libs import form
 from libs import gcs
 from libs import handler
 from libs import helpers
+from libs.query import datastore_query
+
+PAGE_SIZE = 10
+MORE_LIMIT = 50 - PAGE_SIZE  # exactly 5 pages
+
+KEYWORD_FILTERS = [
+    filters.String('name', 'name'),
+]
+
+FILTERS = [
+    filters.Keyword(KEYWORD_FILTERS, 'keywords', 'q'),
+]
 
 
 def get_queues():
@@ -43,6 +56,36 @@ def get_queues():
   queues.sort(key=lambda q: q['display_name'])
   return queues
 
+def get_results(this):
+  """Get results for the jobs page."""
+  params = {k: v for k, v in this.request.iterparams()}
+
+  query = datastore_query.Query(data_types.Job)
+
+  # Results sorted alphabatically by name.
+  # is_desc reverses the sorting.
+  query.order('name', is_desc=False)
+
+  page = helpers.cast(
+      this.request.get('page') or 1, int, "'page' is not an int.")
+
+  # Adding Filters for Search Query
+  filters.add(query, params, FILTERS)
+
+  items, total_pages, total_items, has_more = query.fetch_page(
+      page=page, page_size=PAGE_SIZE, projection=None, more_limit=MORE_LIMIT)
+
+  result = {
+      'hasMore': has_more,
+      'items': items,
+      'page': page,
+      'pageSize': PAGE_SIZE,
+      'totalItems': total_items,
+      'totalPages': total_pages,
+  }
+
+  return result, params
+
 
 class Handler(base_handler.Handler):
   """View job handler."""
@@ -53,19 +96,14 @@ class Handler(base_handler.Handler):
     """Handle a get request."""
     helpers.log('Jobs', helpers.VIEW_OPERATION)
 
-    template_values = self.get_results()
-    self.render('jobs.html', template_values)
-
-  @staticmethod
-  def get_results():
-    """Get results for the jobs page."""
-    jobs = list(data_types.Job.query().order(data_types.Job.name))
     templates = list(data_types.JobTemplate.query().order(
         data_types.JobTemplate.name))
     queues = get_queues()
 
-    return {
-        'jobs': jobs,
+    result, params = get_results(self)
+
+    self.render('jobs.html', {
+        'result': result,
         'templates': templates,
         'fieldValues': {
             'csrf_token': form.generate_csrf_token(),
@@ -74,7 +112,8 @@ class Handler(base_handler.Handler):
             'update_job_template_url': '/update-job-template',
             'upload_info': gcs.prepare_blob_upload()._asdict(),
         },
-    }
+        'params': params,
+    })
 
 
 class UpdateJob(base_handler.GcsUploadHandler):
@@ -232,3 +271,14 @@ class DeleteJobHandler(base_handler.Handler):
 
     helpers.log('Deleted job %s' % job.name, helpers.MODIFY_OPERATION)
     self.redirect('/jobs')
+
+
+class JsonHandler(base_handler.Handler):
+  """Handler that gets the jobs when user clicks on next page."""
+
+  @handler.check_user_access(need_privileged_access=True)
+  @handler.post(handler.JSON, handler.JSON)
+  def post(self):
+    """Get and render the jobs in JSON."""
+    result, _ = get_results(self)
+    self.render_json(result)
--- a/src/appengine/handlers/jobs.py
+++ b/src/appengine/handlers/jobs.py
@@ -56,9 +56,10 @@ def get_queues():
   queues.sort(key=lambda q: q['display_name'])
   return queues
 
+
 def get_results(this):
   """Get results for the jobs page."""
-  params = {k: v for k, v in this.request.iterparams()}
+  params = {k: v for k, v in this.request.iterparams()}  # pylint: disable=unnecessary-comprehension
 
   query = datastore_query.Query(data_types.Job)
 
@@ -102,18 +103,19 @@ class Handler(base_handler.Handler):
 
     result, params = get_results(self)
 
-    self.render('jobs.html', {
-        'result': result,
-        'templates': templates,
-        'fieldValues': {
-            'csrf_token': form.generate_csrf_token(),
-            'queues': queues,
-            'update_job_url': '/update-job',
-            'update_job_template_url': '/update-job-template',
-            'upload_info': gcs.prepare_blob_upload()._asdict(),
-        },
-        'params': params,
-    })
+    self.render(
+        'jobs.html', {
+            'result': result,
+            'templates': templates,
+            'fieldValues': {
+                'csrf_token': form.generate_csrf_token(),
+                'queues': queues,
+                'update_job_url': '/update-job',
+                'update_job_template_url': '/update-job-template',
+                'upload_info': gcs.prepare_blob_upload()._asdict(),
+            },
+            'params': params,
+        })
 
 
 class UpdateJob(base_handler.GcsUploadHandler):
--- a/src/appengine/handlers/jobs.py
+++ b/src/appengine/handlers/jobs.py
@@ -34,12 +34,8 @@ from libs.query import datastore_query
 PAGE_SIZE = 10
 MORE_LIMIT = 50 - PAGE_SIZE  # exactly 5 pages
 
-KEYWORD_FILTERS = [
-    filters.String('name', 'name'),
-]
-
 FILTERS = [
-    filters.Keyword(KEYWORD_FILTERS, 'keywords', 'q'),
+    filters.Keyword([], 'keywords', 'q'),
 ]
 
 
@@ -59,20 +55,14 @@ def get_queues():
 
 def get_results(this):
   """Get results for the jobs page."""
-  params = {k: v for k, v in this.request.iterparams()}  # pylint: disable=unnecessary-comprehension
-
+  params = dict(this.request.iterparams())
+  # Return jobs sorted alphabetically by name
   query = datastore_query.Query(data_types.Job)
-
-  # Results sorted alphabatically by name.
-  # is_desc reverses the sorting.
   query.order('name', is_desc=False)
+  filters.add(query, params, FILTERS)
 
   page = helpers.cast(
       this.request.get('page') or 1, int, "'page' is not an int.")
-
-  # Adding Filters for Search Query
-  filters.add(query, params, FILTERS)
-
   items, total_pages, total_items, has_more = query.fetch_page(
       page=page, page_size=PAGE_SIZE, projection=None, more_limit=MORE_LIMIT)
 
@@ -84,7 +74,6 @@ def get_results(this):
       'totalItems': total_items,
       'totalPages': total_pages,
   }
-
   return result, params
 
 
@@ -102,7 +91,6 @@ class Handler(base_handler.Handler):
     queues = get_queues()
 
     result, params = get_results(self)
-
     self.render(
         'jobs.html', {
             'result': result,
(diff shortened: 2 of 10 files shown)


src/appengine/handlers/jobs.py
```diff
@@ -44,6 +57,37 @@ def get_queues():
   return queues
 
 
+def get_results(this):
+  """Get results for the jobs page."""
+  params = {k: v for k, v in this.request.iterparams()}  # pylint: disable=unnecessary-comprehension
```
oliverchang, line 62: Let's just write this as `dict(this.request.iterparams())` to avoid the lint error.

src/appengine/handlers/jobs.py
```diff
@@ -44,6 +57,37 @@ def get_queues():
...
+
+  query = datastore_query.Query(data_types.Job)
+
+  # Results sorted alphabatically by name.
+  # is_desc reverses the sorting.
```
oliverchang, line 67: not need for this second part of the comment.

src/appengine/handlers/jobs.py
```diff
@@ -44,6 +57,37 @@ def get_queues():
...
+  query.order('name', is_desc=False)
+
+  page = helpers.cast(
+      this.request.get('page') or 1, int, "'page' is not an int.")
+
+  # Adding Filters for Search Query
```
oliverchang, line 73: No need for this comment, as it's pretty self evident.

src/appengine/handlers/jobs.py
```diff
@@ -44,6 +57,37 @@ def get_queues():
...
+
+  result = {
+      'hasMore': has_more,
+      'items': items,
+      'page': page,
+      'pageSize': PAGE_SIZE,
+      'totalItems': total_items,
+      'totalPages': total_pages,
+  }
+
```
oliverchang, line 87: nit: remove this newline.

src/appengine/handlers/jobs.py
```diff
@@ -44,6 +57,37 @@ def get_queues():
   return queues
 
 
+def get_results(this):
+  """Get results for the jobs page."""
+  params = {k: v for k, v in this.request.iterparams()}  # pylint: disable=unnecessary-comprehension
+
```
oliverchang, line 63: remove this newline.

src/appengine/handlers/jobs.py
```diff
@@ -44,6 +57,37 @@ def get_queues():
...
+  # Results sorted alphabatically by name.
+  # is_desc reverses the sorting.
+  query.order('name', is_desc=False)
+
+  page = helpers.cast(
+      this.request.get('page') or 1, int, "'page' is not an int.")
+
+  # Adding Filters for Search Query
+  filters.add(query, params, FILTERS)
```
oliverchang, line 74: Let's group all the these lines like so to reduce vertical whitespace.

```python
# Return jobs sorted alphabetically by name
query = datastore_query.Query(data_types.Job)
query.order('name', is_desc=False)
filters.add(query, params, FILTERS)

page = helpers.cast(
      this.request.get('page') or 1, int, "'page' is not an int.")
items, total_pages, total_items, has_more = query.fetch_page(
      page=page, page_size=PAGE_SIZE, projection=None, more_limit=MORE_LIMIT)
```

src/appengine/handlers/jobs.py
```diff
@@ -53,28 +97,25 @@ def get(self):
...
-        'fieldValues': {
-            'csrf_token': form.generate_csrf_token(),
-            'queues': queues,
-            'update_job_url': '/update-job',
-            'update_job_template_url': '/update-job-template',
-            'upload_info': gcs.prepare_blob_upload()._asdict(),
-        },
-    }
+    result, params = get_results(self)
+
```
oliverchang, line 105: remove this newline.

src/python/tests/appengine/handlers/jobs_test.py
```diff
@@ -54,54 +57,20 @@ def _create_job(self,
...
-    }
-    results = jobs.Handler.get_results()
-    self.assertEqual(expected, results)
+    self.mock.has_access.return_value = True
+    job1 = self._create_job('test_job1', 'APP_NAME = launcher1.py\n')
+    job2 = self._create_job('test_job2', 'APP_NAME = launcher2.py\n')
+    job3 = self._create_job('test_job3', 'APP_NAME = launcher3.py\n')
+    job4 = self._create_job('test_job4', 'APP_NAME = launcher4.py\n')
+
+    expected_items = [
```
oliverchang, line 66: Please test with more jobs such that we test pagination with more than 1 page.

src/python/tests/appengine/handlers/jobs_test.py
```diff
@@ -54,54 +57,20 @@ def _create_job(self,
...
+        job1.key.id(),
+        job2.key.id(),
+        job3.key.id(),
+        job4.key.id()
+    ]
+
+    resp = self.app.post_json('/', {'page': 1})
```
oliverchang, line 73: Can you add a test for the searching part too?

src/python/datastore/data_types.py
```diff
@@ -942,8 +945,13 @@ def get_environment_string(self):
...
     return environment_string
 
+  def populate_indices(self):
+    """Populate keywords for fast test case list searching."""
+    self.keywords = list(search_tokenizer.tokenize(self.name))
+
   def _pre_put_hook(self):
     """Pre-put hook."""
+    self.populate_indices()
     self.project = self.get_environment().get('PROJECT_NAME',
```
oliverchang, line 955: actually, let's make it so that we can search by project as well.

src/python/datastore/data_types.py
```diff
@@ -909,6 +909,9 @@ class Job(Model):
   # Project name.
   project = StringProperty()
 
+  # keywords is used for searching.
```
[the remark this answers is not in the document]
iamarshsingh: sure, will fix this

src/appengine/handlers/jobs.py
iamarshsingh: thanks! Will replace this

src/appengine/handlers/jobs.py
iamarshsingh: sure, will remove this

src/appengine/handlers/jobs.py
iamarshsingh: sure, will remove this

src/appengine/handlers/jobs.py
iamarshsingh: sure, thanks!

src/python/tests/appengine/handlers/jobs_test.py
iamarshsingh: sure, will add it and update the PR.

src/python/tests/appengine/handlers/jobs_test.py
iamarshsingh: Should I make a separate unit-test for searching or add it in the same?

src/python/datastore/data_types.py
iamarshsingh: sure, will add that too.

src/python/tests/appengine/handlers/jobs_test.py
oliverchang: let's make a separate one.

src/appengine/handlers/jobs.py
```diff
@@ -24,10 +24,23 @@
...
 from libs import gcs
 from libs import handler
 from libs import helpers
+from libs.query import datastore_query
+
+PAGE_SIZE = 10
+MORE_LIMIT = 50 - PAGE_SIZE  # exactly 5 pages
+
+KEYWORD_FILTERS = [
+    filters.String('name', 'name'),
```
oliverchang, line 38: Do we actually need this? I think we can skip this and keep things simple. We don't need to do property: value filtering for the name.

iamarshsingh: yeah, we don't need this as all tokens will be in keywords. Thanks for pointing this out.

src/python/datastore/data_types.py
iamarshsingh: should we tokenize project name as well? It will be useful in cases such as `abseil-cpp` which will be tokenized to `abseil` and `cpp`

oliverchang: Yes, good idea :)

src/python/tests/appengine/handlers/jobs_test.py
```diff
@@ -52,56 +56,77 @@ def _create_job(self,
...
+
+    resp = self.app.post_json('/', {'q': "asan"})
+    self.assertListEqual([job_asan.key.id()],
+                         [item['id'] for item in resp.json['items']])
+
+    resp = self.app.post_json('/', {'q': "ubsan"})
+    self.assertListEqual([job_ubsan.key.id()],
+                         [item['id'] for item in resp.json['items']])
+
+    resp = self.app.post_json('/', {'q': "test"})
```
oliverchang, line 129: add a test for searching by project as wsell

src/python/datastore/data_types.py
```diff
@@ -942,8 +945,15 @@ def get_environment_string(self):
...
 
+  def populate_indices(self):
+    """Populate keywords for fast test case list searching."""
+    self.keywords = list(
+        search_tokenizer.tokenize(self.name)
+        | search_tokenizer.tokenize(self.project))
+
   def _pre_put_hook(self):
     """Pre-put hook."""
+    self.populate_indices()
```
oliverchang, line 956: One other thing: this needs to come after .project is set. Otherwise it could be empty.

oliverchang:
You also need to add this:

```
- kind: Job
  properties:
    - name: keywords
    - name: name
```

to `src/appengine/index.yaml` for this to work. This is because you're making a query filtering on `keywords` **and** sorting by `name`.

src/appengine/private/components/jobs/jobs.html
```diff
@@ -164,64 +183,96 @@ <h2>Jobs</h2>
     <div class="section">
       <div class="title">Existing jobs</div>
       <div class="body">
-        <paper-input class="search-box" label="Enter search string here" value="{{jobSearchString::input}}">
+        <paper-input class="search-box" label="Enter search string here" type="search" value="{{params.q}}" on-keyup="searchButtonTapped">
```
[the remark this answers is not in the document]
iamarshsingh: Sure! There is another event which we can use i.e. [onchange](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event). It was mentioned in the note of on-input that if we want to fire a action on an event such as ENTER key pressed, we can use change. Here: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event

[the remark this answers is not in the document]
oliverchang: Not sure if change will be a direct replacement for input, as that only triggers when the value is "committed". Since we want to handle when the "enter" key is pressed, it seem to me the best way to do this is to have two handlers: one for `input` events which just fires the timeout, and one for the keyup event which just tests for "enter" key. This will catch all cases of inputs we care about.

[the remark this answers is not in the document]
iamarshsingh: yes, that would be best option. Thanks :)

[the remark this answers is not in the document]
iamarshsingh: done

src/python/tests/appengine/handlers/jobs_test.py
iamarshsingh: done

src/python/datastore/data_types.py
iamarshsingh: done

iamarshsingh:
Latest commit adds the migration script required to re-put all jobs so that keywords works correctly in Jobs search.

src/appengine/handlers/jobs.py
```diff
@@ -44,6 +53,30 @@ def get_queues():
   return queues
 
 
+def get_results(this):
+  """Get results for the jobs page."""
+  params = dict(this.request.iterparams())
+  # Return jobs sorted alphabetically by name
```
Dor1s, line 59: nit: add an empty line before a comment line

Dor1s: actually it would be even better to move line 58 (`params = ...`) to line 62, as `params` isn't used anywhere before that

src/appengine/handlers/jobs.py
```diff
@@ -44,6 +53,30 @@ def get_queues():
...
+  query = datastore_query.Query(data_types.Job)
+  query.order('name', is_desc=False)
+  filters.add(query, params, FILTERS)
+
+  page = helpers.cast(
```
Dor1s, line 64: probably makes sense to check that the value is not negative

src/local/butler/scripts/migration/__init__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2019 Google LLC
```
Dor1s, line 1: nit: 2020

src/local/butler/scripts/migration/jobs_keywords.py
```diff
@@ -0,0 +1,35 @@
+# Copyright 2019 Google LLC
```
Dor1s, line 1: nit: 2020

src/local/butler/scripts/migration/jobs_keywords.py
```diff
@@ -0,0 +1,35 @@
...
+from datastore import data_types
+from google.cloud import ndb
+
+
+def execute(args):
+  """Build keywords for jobs."""
+  jobs = list(data_types.Job.query())
+  if args.non_dry_run:
+    try:
+      ndb.put_multi(jobs)
```
Dor1s, line 27: would `ndb_utils.put_multi` be a better choice here? it does batching to avoid the max keys limitation, so you shouldn't need the loop inside the exception handler

oliverchang: That was for an older version of ndb which had a bug in the implementation. Should be no longer needed here :) 

That being said.. what is the exception handling for? Did you run into actual issues here?

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 Change Log
 ==========
 
+## Next
+- Next version requires new DB migrations due to addition of search keywords in Jobs.
+- To perform migrations, please use:
+```
+python butler.py run -c=path/to/config --non-dry-run migration.jobs_keywords
```
oliverchang, line 8: nit: no need for the = after the "-c"

src/appengine/handlers/jobs.py
iamarshsingh: Yes, actually I have opened a issue(#1883) for the same. This issue exists everywhere in testcase-list, crash and here as well. I was actually planning to add a fix for this datastore_query.Query class' fetch_page method, as this would ensure that it will be fixed everywhere and we wouldn't have to add explicit validations in all pages.

src/local/butler/scripts/migration/jobs_keywords.py
iamarshsingh: @oliverchang thanks for the clarification. No I didn't run into any actual issues. I add exception handling because it was used in build_attributes.py, which serves a similar purpose. I will remove the for loop and just keep a print message incase something goes wrong.

src/appengine/handlers/jobs.py
iamarshsingh: done

src/local/butler/scripts/migration/__init__.py
iamarshsingh: done

src/local/butler/scripts/migration/jobs_keywords.py
iamarshsingh: done

iamarshsingh:
Thank you for the constructive and detailed reviews @oliverchang and @Dor1s I have addressed all review issues, please have a look!

src/local/butler/scripts/migration/jobs_keywords.py
```diff
@@ -0,0 +1,28 @@
...
+      print("Done building keywords for jobs.")
+    except Exception:
+      print("Error during building keywords for jobs.")
```
oliverchang, line 28: Better to just let the exception surface normally, rather than handling. As is, the error message is not very helpful as it hides the stacktrace.

iamarshsingh: done

src/appengine/handlers/jobs.py
Dor1s: sg!

PR #1885, opened 2020-07-06
Upgrade CPU tensorflow to 2.3.0-rc0 and re-enable training test (#1540).

Dor1s:
I was preparing this for @mihaimaruseac as discussed in https://github.com/google/clusterfuzz/issues/1540, but the test seems to pass locally now. Maybe something's wrong with my environment, let's see what CI thinks about this.


Dor1s:
Done in https://github.com/google/clusterfuzz/pull/1921

PR #1887, opened 2020-07-06
Store past crash regressions in corpus backup archive.

inferno-chromium:
See https://github.com/google/clusterfuzz/pull/1869#discussion_r447182042

### DIFF
--- a/src/python/bot/tasks/corpus_pruning_task.py
+++ b/src/python/bot/tasks/corpus_pruning_task.py
@@ -646,9 +646,18 @@ def do_corpus_pruning(context, last_execution_failed, revision):
   context.sync_to_gcs()
 
   # Create corpus backup.
+  # Temporarily move the past crash regressions folder into the minimized corpus
+  # so that corpus backup archive can have both.
+  regressions_input_dir = os.path.join(context.initial_corpus_path,
+                                       'regressions')
+  regressions_output_dir = os.path.join(context.minimized_corpus_path,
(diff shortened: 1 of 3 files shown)


src/python/bot/tasks/corpus_pruning_task.py
```diff
@@ -646,9 +646,18 @@ def do_corpus_pruning(context, last_execution_failed, revision):
   context.sync_to_gcs()
 
   # Create corpus backup.
+  # Temporarily copy the past crash regressions folder into the minimized corpus
+  # so that corpus backup archive can have both.
+  regressions_input_dir = os.path.join(context.initial_corpus_path,
+                                       'regressions')
+  regressions_output_dir = os.path.join(context.minimized_corpus_path,
```
Dor1s, line 653: so this would be non-minimized, but we don't expect a lot of files in there?

inferno-chromium: i dont think we can minimize crash regressions as the input itself is important for regression testing. over time, size can get large, but this should be only for crashes with bugs attached, so number will be very low. over time, i can think of removing maybe ooms and timeout from being archived if needed, but it feels weird to exclude them right now, i think a project maintainer cares about all regressions.

Dor1s: makes sense! Should be fine since it's only the ones with bugs attached. Thanks!

Issue #1883, opened 2020-07-06
Missing Validation checks in Pagination Request Handlers


iamarshsingh:
There is lack of Validation checks for negative page numbers in Testcase-list and Crash-stats.

**To Reproduce:** (Enter a negative page number in the URL)
1. https://oss-fuzz.com/testcases?open=yes&page=-5
2. https://oss-fuzz.com/crash-stats?block=day&days=7&end=442789&group=platform&number=count&page=-2&sort=total_count

**Screenshot**
![Error in Negative Page number](https://user-images.githubusercontent.com/30122295/86598654-0ca2b200-bfbb-11ea-9e0b-ad06404a3ec1.png)

I discussed with @oliverchang and he suggested to put validation checks and display empty list.

oliverchang:
Should be fixed now.

PR #1888, opened 2020-07-07
Added validation checks for negative page numbers.

iamarshsingh:
Hi, This PR fixes the issues mentioned in #1883. The approach is to display `Didn't find anything with the current search.` instead of the error message. 

Please let me know if there are any changes required.

### DIFF
--- a/src/appengine/libs/query/datastore_query.py
+++ b/src/appengine/libs/query/datastore_query.py
@@ -281,14 +281,35 @@ class Query(base.Query):
 
   def fetch_page(self, page, page_size, projection, more_limit):
     """Return the items, total_pages, total_items, and has_more."""
+
+    # Validation check to confirm positive page number.
+    if page > 0:
+      items, total_items, has_more = self.fetch(
+          offset=(page - 1) * page_size,
+          limit=page_size,
+          projection=projection,
+          more_limit=more_limit)
+
+      total_pages = total_items // page_size
+      if (total_items % page_size) > 0:
+        total_pages += 1
+
+      return items, total_pages, total_items, has_more
+
+    # Retriving deatils for page 1 to get other details correct.
     items, total_items, has_more = self.fetch(
-        offset=(page - 1) * page_size,
+        offset=0,
         limit=page_size,
         projection=projection,
         more_limit=more_limit)
+
+    if total_items > 0:
+      has_more = False
 
     total_pages = total_items // page_size
     if (total_items % page_size) > 0:
       total_pages += 1
 
+    # Non-positive page number gets no resulting items found.
+    items = []
     return items, total_pages, total_items, has_more
--- a/src/appengine/libs/query/datastore_query.py
+++ b/src/appengine/libs/query/datastore_query.py
@@ -298,10 +298,7 @@ class Query(base.Query):
 
     # Retriving deatils for page 1 to get other details correct.
     items, total_items, has_more = self.fetch(
-        offset=0,
-        limit=page_size,
-        projection=projection,
-        more_limit=more_limit)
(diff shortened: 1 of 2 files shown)


src/appengine/libs/query/datastore_query.py
```diff
@@ -281,14 +281,32 @@ def fetch(self, offset, limit, projection, more_limit):
 
   def fetch_page(self, page, page_size, projection, more_limit):
     """Return the items, total_pages, total_items, and has_more."""
+
+    # Validation check to confirm positive page number.
+    if page > 0:
```
oliverchang, line 286: Actually, let's just set it to 1 by default to keep the implementation simple. There's no real benefit of showing nothing vs showing just the first page in this case.

iamarshsingh: if we would like to do that, it would be better to that on the front-end side i.e. change the negative requested page number before sending posting it. As if we just handle it here, the page number in the front-end side of the stack will remain the same. So, if user requests for page -5 and we show data for page 1, and then he clicks next. His page number will shift to -4.

oliverchang: That's fine. This is not a case that anyone would run into normally. I think we are spending too much engineering effort to solve this here :)

iamarshsingh: yes, I agree. I have updated this to keep 1 by default and added tests.

src/python/tests/appengine/libs/query/datastore_query_test.py
```diff
@@ -171,6 +171,24 @@ def test_greater_or_equal(self):
...
+    query.order('datetime_value', is_desc=True)
+
+    items, total_pages, total_items, has_more = query.fetch_page(
+        page=-5, page_size=2, projection=['tokens'], more_limit=4)
+    self.assertListEqual([self.mocks[0].key.id(), self.mocks[2].key.id()],
+                         [item.key.id() for item in items])
+    self.assertListEqual([['a'], ['a']], [item.tokens for item in items])
+    with self.assertRaises(ndb.UnprojectedPropertyError):
+      _ = [item.boolean_value for item in items]
+    self.assertEqual(6, total_items)
```
inferno-chromium, line 188: Compared to test_third_page, why did the total_items and total_pages decrease here ?

iamarshsingh: The page we are loading here is 1st which is followed by 2 pages(2nd and 3rd) each containing 2 items each. So total_items becomes 6 and total_pages becomes 3. 

In the 3rd page test case, 2 pages are before the 3rd one(1st and 2nd) and 2 are followed by it(4th and 5th) which results in total_items and total_pages as 10 and 5 respectively.

inferno-chromium: How do their total items differ if the query is the same. that is what confused me.

iamarshsingh: Oh, actually this wrapper doesn't returns the total_items that are present in the datastore, this just returns how many are there in the particular page limit we are looking at. Such as if the if we are at page 10 and it has 2 pages before this in display and 2 after and each page have 20 items. Then total_items will return 5*20=100. 

In the above test case of 3rd page you can also see it is 10 but in the datastore prepration, it actually adds 30 entries to the datastore.

inferno-chromium: ah great, thanks for explaining!

iamarshsingh: no problem :)

PR #1889, opened 2020-07-07
Lower down corpus backup public marking time

inferno-chromium:
Change corpus backup public marking time from 90 to 30 days.

Now that we archive crash regressions in corpus backup, it makes sense for projects to use this corpus backup sooner than later. Also, anyone can generate this corpus with enough cpus, so 30 days seems reasonable.


jonathanmetzman:
>Now that we archive crash regressions in corpus backup, it makes sense for projects to use this corpus backup sooner than later.

Not sure I follow this.

PR #1891, opened 2020-07-08
Android Kernel repro.prop acquisition

mspectorgoogle:
This is the first part of 2 changes that will allow linkification of
Android kernel crashes.  This change simply gets the repo.prop from the
Android build server and places it as a new element of the
StackAnalyzerState.

### DIFF
--- a/src/python/crash_analysis/stack_parsing/stack_analyzer.py
+++ b/src/python/crash_analysis/stack_parsing/stack_analyzer.py
@@ -27,6 +27,7 @@ from config import local_config
 from crash_analysis import crash_analyzer
 from crash_analysis.stack_parsing import stack_parser
 from metrics import logs
+from platforms.android import settings as android_settings
 from system import environment
 
 C_CPP_EXTENSIONS = ['c', 'cc', 'cpp', 'cxx', 'h', 'hh', 'hpp', 'hxx']
@@ -634,6 +635,7 @@ class StackAnalyzerState(object):
 
     # Additional tracking for Android bugs.
     self.found_java_exception = False
+    self.android_kernel_repo = None
 
     # Additional tracking for bad casts.
     self.found_bad_cast_crash_end_marker = False
@@ -1224,6 +1226,7 @@ def get_crash_data(crash_data, symbolize_flag=True):
   is_python = '.py", line' in crash_stacktrace_without_inlines
   found_python_crash = False
   found_golang_crash = False
+  found_android_kernel_crash = False
   ubsan_disabled = 'halt_on_error=0' in environment.get_value(
       'UBSAN_OPTIONS', '')
 
@@ -1927,6 +1930,8 @@ def get_crash_data(crash_data, symbolize_flag=True):
     android_kernel_match = add_frame_on_match(
         ANDROID_KERNEL_STACK_FRAME_REGEX, line, state, group=3)
     if android_kernel_match:
+      found_android_kernel_crash = True
+
       # Update address from the first stack frame unless we already have
       # more detailed information from KASan.
       if state.frame_count == 1 and not is_kasan:
@@ -1955,6 +1960,13 @@ def get_crash_data(crash_data, symbolize_flag=True):
     if is_python and add_frame_on_match(
         PYTHON_STACK_FRAME_FUNCTION_REGEX, line, state, group=3):
       continue
+
+  # Only get repo.prop if we have an Android kernel or KASAN crash
+  if not state.android_kernel_repo and (found_android_kernel_crash or is_kasan):
+    repo_prop_path = android_settings.get_repo_prop_path()
+    if repo_prop_path and os.path.exists(repo_prop_path):
+      state.android_kernel_repo = utils.read_data_from_file(
+          repo_prop_path, eval_data=False).decode()
 
   # Detect cycles in stack overflow bugs and update crash state.
   update_crash_state_for_stack_overflow_if_needed(state)
--- a/src/python/crash_analysis/stack_parsing/stack_symbolizer.py
+++ b/src/python/crash_analysis/stack_parsing/stack_symbolizer.py
@@ -45,6 +45,7 @@ from metrics import logs
 from platforms.android import adb
 from platforms.android import fetch_artifact
 from platforms.android import settings
+from platforms.android import symbols_downloader
 from system import archive
 from system import environment
 from system import shell
@@ -556,7 +557,7 @@ def filter_binary_path(binary_path):
 
     # We didn't find the library locally in the build directory.
     # Try finding the library in the local system library cache.
-    download_system_symbols_if_needed(symbols_directory)
+    symbols_downloader.download_system_symbols_if_needed(symbols_directory)
     local_binary_path = utils.find_binary_path(symbols_directory, binary_path)
     if local_binary_path:
       return local_binary_path
@@ -636,74 +637,3 @@ def symbolize_stacktrace(unsymbolized_crash_stacktrace,
       unsymbolized_crash_stacktrace)
 
   return symbolized_crash_stacktrace
-
-
-def download_system_symbols_if_needed(symbols_directory):
-  """Download system libraries from |SYMBOLS_URL| and cache locally."""
-  # For local testing, we do not have access to the cloud storage bucket with
-  # the symbols. In this case, just bail out.
-  if environment.get_value('LOCAL_DEVELOPMENT'):
-    return
-
-  # When running reproduce tool locally, we do not have access to the cloud
-  # storage bucket with the symbols. In this case, just bail out.
-  if environment.get_value('REPRODUCE_TOOL'):
-    return
-
-  # We have archived symbols for google builds only.
-  if not settings.is_google_device():
-    return
-
-  # Get the build fingerprint parameters.
-  build_params = settings.get_build_parameters()
-  if not build_params:
-    logs.log_error('Unable to determine build parameters.')
-    return
-  build_id = build_params.get('build_id')
-  target = build_params.get('target')
-  type = build_params.get('type')
-  if not build_id or not target or not type:
-    logs.log_error('Null build parameters found, exiting.')
-    return
-
-  # Check if we already have the symbols in cache.
-  build_params_check_path = os.path.join(symbols_directory,
-                                         '.cached_build_params')
-  cached_build_params = utils.read_data_from_file(
-      build_params_check_path, eval_data=True)
-  if cached_build_params and cached_build_params == build_params:
-    # No work to do, same system symbols already in cache.
-    return
-
-  symbols_archive_filename = '%s-symbols-%s.zip' % (target, build_id)
-  symbols_archive_path = os.path.join(symbols_directory,
-                                      symbols_archive_filename)
-
-  # Delete existing symbols directory first.
-  shell.remove_directory(symbols_directory, recreate=True)
-
-  # Fetch symbol file from cloud storage cache (if available).
-  found_in_cache = storage.get_file_from_cache_if_exists(
-      symbols_archive_path, update_modification_time_on_access=False)
-  if not found_in_cache:
-    # Include type and sanitizer information in the target.
-    target_with_type_and_san = '%s-%s' % (target, type)
-    tool_suffix = environment.get_value('SANITIZER_TOOL_NAME')
-    if tool_suffix and not tool_suffix in target_with_type_and_san:
-      target_with_type_and_san += '_%s' % tool_suffix
-
-    # Fetch the artifact now.
-    fetch_artifact.get(build_id, target_with_type_and_san,
-                       symbols_archive_filename, symbols_directory)
-
-  if not os.path.exists(symbols_archive_path):
-    logs.log_error(
-        'Unable to locate symbols archive %s.' % symbols_archive_path)
-    return
-
-  # Store the artifact for later use or for use by other bots.
-  storage.store_file_in_cache(symbols_archive_path)
-
-  archive.unpack(symbols_archive_path, symbols_directory, trusted=True)
-  shell.remove_file(symbols_archive_path)
-  utils.write_data_to_file(build_params, build_params_check_path)
--- a/src/python/platforms/android/fetch_artifact.py
+++ b/src/python/platforms/android/fetch_artifact.py
@@ -50,7 +50,8 @@ def execute_request_with_retries(request):
   return result
 
 
-def download_artifact(client, bid, target, attempt_id, name, output_directory):
+def download_artifact(client, bid, target, attempt_id, name, output_directory,
+                      output_filename):
   """Download one artifact."""
   artifact_query = client.buildartifact().get(
       buildId=bid, target=target, attemptId=attempt_id, resourceId=name)
@@ -71,9 +72,13 @@ def download_artifact(client, bid, target, attempt_id, name, output_directory):
   dl_request = client.buildartifact().get_media(
       buildId=bid, target=target, attemptId=attempt_id, resourceId=name)
 
-  # If the artifact already exists, then bail out.
-  file_name = name.replace('signed/', '')
+  if output_filename:
+    file_name = output_filename
+  else:
+    file_name = name.replace('signed/', '')
+
   output_path = os.path.join(output_directory, file_name)
+  # If the artifact already exists, then bail out.
   if os.path.exists(output_path) and os.path.getsize(output_path) == size:
     logs.log('Artifact %s already exists, skipping download.' % name)
     return output_path
@@ -159,7 +164,7 @@ def get_latest_artifact_info(branch, target, signed=False):
   return {'bid': bid, 'target': target}
 
 
-def get(bid, target, regex, output_directory):
+def get(bid, target, regex, output_directory, output_filename=None):
   """Return artifact for a given build id, target and file regex."""
   client = get_client()
   if not client:
@@ -171,10 +176,11 @@ def get(bid, target, regex, output_directory):
       bid=bid,
       target=target,
       regex=regex,
-      output_directory=output_directory)
+      output_directory=output_directory,
+      output_filename=output_filename)
 
 
-def run_script(client, bid, target, regex, output_directory):
+def run_script(client, bid, target, regex, output_directory, output_filename):
   """Download artifacts as specified."""
   artifacts = get_artifacts_for_build(
       client=client, bid=bid, target=target, attempt_id='latest')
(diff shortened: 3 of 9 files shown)


src/python/crash_analysis/stack_parsing/stack_symbolizer.py
```diff
@@ -45,6 +45,7 @@
 from platforms.android import adb
 from platforms.android import fetch_artifact
 from platforms.android import settings
+from platforms.android import symbols_downloader
```
oliverchang, line 48: I think you forgot to add this file?

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -1956,6 +1961,14 @@ def get_crash_data(crash_data, symbolize_flag=True):
         PYTHON_STACK_FRAME_FUNCTION_REGEX, line, state, group=3):
       continue
 
+  # Only get repo.prop if we have and Android kernel or KASAN crash
```
oliverchang, line 1964: s/and/an/

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -27,6 +27,7 @@
 from crash_analysis import crash_analyzer
 from crash_analysis.stack_parsing import stack_parser
 from metrics import logs
+from platforms.android import settings
```
oliverchang, line 30: import settings as android_settings.

src/python/platforms/android/fetch_artifact.py
```diff
@@ -72,7 +73,7 @@ def download_artifact(client, bid, target, attempt_id, name, output_directory):
       buildId=bid, target=target, attemptId=attempt_id, resourceId=name)
 
   # If the artifact already exists, then bail out.
-  file_name = name.replace('signed/', '')
+  file_name = output_filename.replace('signed/', '')
```
oliverchang, line 76: output_filename has a default of None in many places. This will exception in this case. Does this need to default to `name`?

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -1956,6 +1961,14 @@ def get_crash_data(crash_data, symbolize_flag=True):
         PYTHON_STACK_FRAME_FUNCTION_REGEX, line, state, group=3):
       continue
 
+  # Only get repo.prop if we have and Android kernel or KASAN crash
+  if environment.platform() in ('ANDROID', 'ANDROID_KERNEL') and (
```
inferno-chromium, line 1965: I don't think we need to check environment.platform here, just "found_android_kernel_crash or is_kasan" should be enough.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -1956,6 +1961,14 @@ def get_crash_data(crash_data, symbolize_flag=True):
...
+      found_android_kernel_crash or is_kasan):
+    repo_prop_path = settings.get_repo_prop_path()
+    if os.path.exists(repo_prop_path):
+      state.android_kernel_repo = utils.read_data_from_file(
```
inferno-chromium, line 1969: Where is android_kernel_repo used ?

src/python/crash_analysis/stack_parsing/stack_analyzer.py
inferno-chromium: +1

src/python/crash_analysis/stack_parsing/stack_symbolizer.py
inferno-chromium: +1, need to review this.
also, symbols_downloader as android_symbols_downloader to be more explicit.

src/python/platforms/android/settings.py
```diff
@@ -150,6 +167,36 @@ def get_security_patch_level():
...
+  product_name = get_product_name()
+  build_product = get_build_product()
+
+  # Some devices have a different kernel name than product_name, if so use the
+  # kernel name, but don't forget the kasan suffix if it exists.
+  if product_name in PRODUCT_TO_KERNEL:
+    return PRODUCT_TO_KERNEL[build_product]
+
+  # Strip _kasan off of the end as we will add it later if needed.
+  if product_name.endswith('_kasan'):
```
inferno-chromium, line 194: from base import utils
use
utils.strip_from_right(

src/python/platforms/android/settings.py
```diff
@@ -194,3 +241,14 @@ def set_database_setting(database_path, table, key, value):
   sql_command_string = ('"UPDATE %s SET value=\'%s\' WHERE name=\'%s\'"') % (
       table, str(value), key)
   adb.run_shell_command(['sqlite3', database_path, sql_command_string])
+
+
+def get_repo_prop_path():
+  symbols_directory = environment.get_value('SYMBOLS_DIR')
```
inferno-chromium, line 247: function description.

src/python/platforms/android/settings.py
```diff
@@ -194,3 +241,14 @@ def set_database_setting(database_path, table, key, value):
...
+  _, _, repro_filename = symbols_downloader.get_kernel_info()
+
+  # Grab repo.prop, it is not on the device nor in the build_dir
```
inferno-chromium, line 250: end comment with period.

src/python/platforms/android/settings.py
```diff
@@ -18,13 +18,25 @@
...
+from base import utils
+from platforms.android import symbols_downloader
 from system import environment
 
 BUILD_FINGERPRINT_REGEX = re.compile(
     r'(?P<vendor>.+)\/(?P<target>.+)'
     r'\/(?P<flavor>.+)\/(?P<name_name>.+)'
     r'\/(?P<build_id>.+):(?P<type>.+)\/(?P<keys>.+)')
 
+PRODUCT_TO_KERNEL = {
```
inferno-chromium, line 30: probably lets move this to existing file android/constants.py , so that it is more prominent and will be updated over time.

src/python/platforms/android/settings.py
```diff
@@ -150,6 +167,36 @@ def get_security_patch_level():
...
+
+def get_kernel_name():
+  """Returns the kernel name for the device, since some kernels are shared."""
+  product_name = get_product_name()
+  build_product = get_build_product()
+
+  # Some devices have a different kernel name than product_name, if so use the
+  # kernel name, but don't forget the kasan suffix if it exists.
+  if product_name in PRODUCT_TO_KERNEL:
+    return PRODUCT_TO_KERNEL[build_product]
```
inferno-chromium, line 191: What if device family is not in list, maybe use attribute.get() so that it returns None.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
mspectorgoogle: It will be used in the second CL to build the srcmap.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
mspectorgoogle: The theory here is a non Android KASAN kernel should not get repro

inferno-chromium: Right now, it will do environment.platform() which checks env var on every line parsed in stacktrace, very slow. I think found_android_kernel_crash and is_kasan should indicate android kernel and in other cases should bail out.

mspectorgoogle: When in non kasan (just straight kernel) crashes we want the repo.prop as well hence the or.  I can remove the platform check, it just means we will get the repro every time.  We should also probably check if the state.android_kernel_repo  is still none otherwise we'll download it more than once

src/python/platforms/android/settings.py
```diff
@@ -194,3 +228,15 @@ def set_database_setting(database_path, table, key, value):
   sql_command_string = ('"UPDATE %s SET value=\'%s\' WHERE name=\'%s\'"') % (
       table, str(value), key)
   adb.run_shell_command(['sqlite3', database_path, sql_command_string])
+
+
+def get_repo_prop_path():
+  """Download repo.prop and return path of it on local machine."""
+  symbols_directory = environment.get_value('SYMBOLS_DIR')
```
inferno-chromium, line 235: Lets not use the same directory as userspace symbols, as otherwise we will constantly keep deleting symbols dir content, first with userspace symbols, then delete, then have kernel symbols, then delete, etc.
kernel_symbols_directory = os.path.join(symbols_directory, 'kernel')

src/python/platforms/android/symbols_downloader.py
```diff
@@ -0,0 +1,144 @@
...
+from system import shell
+
+
+def get_kernel_info():
+  """Get the build_id, target, and symbols_archive_filename."""
+  build_id = settings.get_kernel_build_id()
+  target = settings.get_kernel_name()
+  if not build_id or not target:
+    return None, None, None
+  symbols_archive_filename = '%s-%s-repo.prop' % (target, build_id)
```
inferno-chromium, line 35: symbol archive filename should be a different function, that just take target, build_id.
and then get_kernel_info should in settings.py as well

src/python/platforms/android/settings.py
```diff
@@ -194,3 +228,15 @@ def set_database_setting(database_path, table, key, value):
...
+  _, _, repro_filename = symbols_downloader.get_kernel_info()
```
inferno-chromium, line 236: see comment below, instead of unused args in unrelated function
build_id, target = settings.get_kernel_info()
repro_filename = settings.get_symbols_filename('repro.prop') or some other helper function.

src/python/platforms/android/symbols_downloader.py
```diff
@@ -0,0 +1,144 @@
...
+  # When running reproduce tool locally, we do not have access to the cloud
+  # storage bucket with the symbols. In this case, just bail out.
+  if environment.get_value('REPRODUCE_TOOL'):
+    return
+
+  # We have archived symbols for google builds only.
+  if not settings.is_google_device():
+    return
+
+  # For Android kernel we want to get the repro.prop
```
inferno-chromium, line 55: End comment with period (dot).

src/python/platforms/android/symbols_downloader.py
```diff
@@ -0,0 +1,144 @@
...
+    target = build_params.get('target')
+    build_type = build_params.get('type')
+    if not build_id or not target or not build_type:
+      logs.log_error('Null build parameters found, exiting.')
+      return
+
+    symbols_archive_filename = '%s-symbols-%s.zip' % (target, build_id)
+    artifact_file_name = symbols_archive_filename
+    output_filename_override = None
+  # Check if we already have the symbols in cache.
```
inferno-chromium, line 85: New line before this line.

src/python/platforms/android/symbols_downloader.py
```diff
@@ -0,0 +1,144 @@
...
+    tool_suffix = environment.get_value('SANITIZER_TOOL_NAME')
+
+    if is_kernel:
+      # Some kernels are just 'kernel', some are kernel_target
+      if tool_suffix:
+        targets_with_type_and_san = [
+            'kernel_%s' % tool_suffix,
+            'kernel_%s_%s' % (tool_suffix, target)
+        ]
+
```
inferno-chromium, line 113: i think you are missing an else here.

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -42,11 +42,13 @@ def setUp(self):
     helpers.patch(self, [
         'crash_analysis.stack_parsing.stack_symbolizer.symbolize_stacktrace',
         'metrics.logs.log_error',
+        'platforms.android.settings.get_repo_prop_path'
     ])
 
     os.environ['JOB_NAME'] = TEST_JOB_NAME
 
     self.mock.symbolize_stacktrace.side_effect = self._mock_symbolize_stacktrace
```
inferno-chromium, line 50: better to just do self.mock.symbolize_stacktrace.return_value = 'value' directly instead of defining a new function.

src/python/tests/core/platforms/android/stack_analyzer_test.py
```diff
@@ -0,0 +1,118 @@
...
+class AndroidStackAnalyzerTest(unittest.TestCase):
+  """Android specific Stack analyzer tests."""
+  # pylint: disable=unused-argument
+  @staticmethod
+  def _mock_symbolize_stacktrace(stacktrace, enable_inline_frames=True):
+    """No-op mocked version of symbolize_stacktrace."""
+    return stacktrace
+
+  @staticmethod
+  def _fetch_artifact_get(bid,
```
inferno-chromium, line 58: _mock_fetch_artifact_get

src/python/tests/core/platforms/android/stack_analyzer_test.py
```diff
@@ -0,0 +1,118 @@
...
+                          target,
+                          regex,
+                          output_directory,
+                          output_filename_override=None):
+    if output_filename_override:
+      artifact_path = os.path.join(output_directory, output_filename_override)
```
inferno-chromium, line 64: else: branch missing ?
 what is artifact_path ?

src/python/tests/core/platforms/android/stack_analyzer_test.py
```diff
@@ -0,0 +1,118 @@
...
+    with open(artifact_path, 'w') as artifact_file:
+      artifact_file.write(KERNEL_REPRO)
+
+  # pylint: enable=unused-argument
+
```
inferno-chromium, line 69: why empty line, can be removed.

src/python/tests/core/platforms/android/stack_analyzer_test.py
```diff
@@ -0,0 +1,118 @@
...
+platform/prebuilts/misc u'15560bb32cdb9b47db48eb4865b736df9708a8fe'
+platform/tools/repohooks u'233b8010f7f5e3c544b47c68ffae781860156945'
+"""
+_real_read_data_from_file = None
+
+
+class AndroidStackAnalyzerTest(unittest.TestCase):
+  """Android specific Stack analyzer tests."""
+  # pylint: disable=unused-argument
+  @staticmethod
```
inferno-chromium, line 52: staticmethod are not encourage anymore, can remove - https://google.github.io/styleguide/pyguide.html#217-function-and-method-decorators

src/python/tests/core/platforms/android/stack_analyzer_test.py
```diff
@@ -0,0 +1,118 @@
...
+
+from base import utils
+from crash_analysis.stack_parsing import stack_analyzer
+from system import environment
+from tests.test_libs import helpers as test_helpers
+
+DATA_DIRECTORY = os.path.join(os.path.dirname(__file__), 'stack_analyzer_data')
+TEST_JOB_NAME = 'test'
+
+KERNEL_REPRO = """kernel/build u'c059b39e1caf2b96aa376582eeb93062b43d69d5'
```
inferno-chromium, line 27: I need to understand how you plan to hook this information for srcmap/link hyperlinking. Will you change stacktrace lines to include this in next CL (i remember you said part 2).

src/python/tests/core/platforms/android/stack_analyzer_test.py
mspectorgoogle: We're just mocking so we write to disk, the only place we call fetch_artifact_get with output_filename_override is when we are calling it for the repo.prop.  We don't want to disturb other calls to fetch_artifact_get (there should be none in this call chain) but I add this check just in case.  We are basically faking the call and just writing to disk where the rest of the code can read it as we are not interested in setting up everything to call out to ab/ for real

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
mspectorgoogle: Well you need to return the first element passed to it, I don't think you can do that with return_value, you could do a lambda I guess.  I copied this from the other stack_analyzer_test.py

src/python/platforms/android/settings.py
mspectorgoogle: I added it as a new constant but yes

src/python/tests/core/platforms/android/stack_analyzer_test.py
mspectorgoogle: Hmm, I need to look at how the srcmap code works, I will probably extract the correct hash and project from this.  Yeah the next cl will use this data

src/python/tests/core/platforms/android/stack_analyzer_test.py
```diff
@@ -0,0 +1,120 @@
...
+    self.mock.symbolize_stacktrace.side_effect = _mock_symbolize_stacktrace
+
+  def _read_test_data(self, name):
+    """Helper function to read test data."""
+    with open(os.path.join(DATA_DIRECTORY, name)) as handle:
+      return handle.read()
+
+  def test_syzkaller_kasan_android_with_env(self):
+    """Test syzkaller kasan."""
+    global _real_read_data_from_file
```
oliverchang, line 96: Let's avoid this use of global here. We can just save this in self._real_read_data_from_file, and define _mock_read_data_from_file as a member function. You can still do ` self.mock.read_data_from_file.side_effect = self._mock_read_data_from_file` and things will work as expected.

src/python/system/environment.py
```diff
@@ -913,6 +913,9 @@ def set_bot_environment():
                                          'fuzzer-common-data-bundles')
   os.environ['IMAGES_DIR'] = os.path.join(inputs_dir, 'images')
   os.environ['SYMBOLS_DIR'] = os.path.join(inputs_dir, 'symbols')
+  os.environ['KERNEL_SYMBOLS_DIR'] = os.path.join(inputs_dir, 'symbols',
```
inferno-chromium, line 916: It is just used in one place, so best to not define an environment var for it, but just add it locally in that function.

src/python/tests/core/platforms/android/stack_analyzer_test.py
```diff
@@ -0,0 +1,120 @@
...
+                             output_directory,
+                             output_filename_override=None):
+  if output_filename_override:
+    artifact_path = os.path.join(output_directory, output_filename_override)
+    with open(artifact_path, 'w') as artifact_file:
+      artifact_file.write(KERNEL_REPRO)
+
+
+# pylint: enable=unused-argument
+
```
inferno-chromium, line 67: Two empty line, do you need these.

src/python/tests/core/platforms/android/stack_analyzer_test.py
inferno-chromium: +1

Issue #1892, opened 2020-07-08
Fuzzer Stats: show logs link even when the job isn't specified


Dor1s:
Reported and discussed in https://github.com/google/oss-fuzz/issues/4092

PR #1894, opened 2020-07-08
Fuzzer Stats: show logs link even when Job is not specified (#1892).

Dor1s:
I will probably test on staging first, if others think it's a reasonable change


Dor1s:
Seems to be working :) https://staging-dot-clusterfuzz-external.appspot.com/fuzzer-stats?group_by=by-day&date_start=2020-07-02&date_end=2020-07-07&fuzzer=libFuzzer_qt_qtbase_corelib_serialization_qcborstreamreader_next&project=qt

PR #1898, opened 2020-07-10
Support repro of syzkaller changes using syz-crush

hzawawy:


src/python/bot/fuzzers/syzkaller/runner.py
```diff
@@ -95,7 +95,20 @@ def repro(self, repro_timeout, repro_args):
     logs.log('Running Syzkaller testcase.')
     additional_args = copy.copy(repro_args)
     result = self.run_and_wait(additional_args, timeout=repro_timeout)
-    logs.log('Syzkaller testcase stopped.')
+    reproducible = False
+    if 'all done.' in result.output:
+      regex = re.compile(r'reproduced (\d+) crashes')
```
oliverchang, line 100: compile this once and save this as a class constant.

src/python/bot/fuzzers/syzkaller/runner.py
```diff
@@ -95,7 +95,20 @@ def repro(self, repro_timeout, repro_args):
...
+      search = regex.search(result.output)
+      if search and search.group(1):
+        if search.group(1) > '0':
+          reproducible = True
+
+    if reproducible:
+      result.return_code = 0
+      logs.log('Repro success.')
+    else:
+      logs.log('Failed to repro.')
```
oliverchang, line 110: Do we need to set return_code in this case too?

src/python/bot/fuzzers/syzkaller/runner.py
```diff
@@ -95,7 +95,20 @@ def repro(self, repro_timeout, repro_args):
     logs.log('Running Syzkaller testcase.')
     additional_args = copy.copy(repro_args)
     result = self.run_and_wait(additional_args, timeout=repro_timeout)
-    logs.log('Syzkaller testcase stopped.')
+    reproducible = False
+    if 'all done.' in result.output:
+      regex = re.compile(r'reproduced (\d+) crashes')
+      search = regex.search(result.output)
+      if search and search.group(1):
+        if search.group(1) > '0':
```
oliverchang, line 103: fold this if into the above.

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -130,7 +131,8 @@ def reproduce(self, target_path, input_path, arguments, max_time):  # pylint: di
     syzkaller_runner = runner.get_runner(
         os.path.join(binary_dir, constants.SYZ_REPRO))
     repro_args = runner.get_config()
-    repro_args.extend(input_path)
+    repro_args.extend(
+        ['-infinite=false', '-restart_time={}s'.format(max_time), input_path])
```
oliverchang, line 135: may be worth setting a slightly lowered time than `max_time` here, as `max_time` is a hard limit at which we kill the process. If we set it slightly lower in the argument here, we can give it some time to exit cleanly (if necessary).

src/python/bot/fuzzers/syzkaller/runner.py
```diff
@@ -95,7 +95,20 @@ def repro(self, repro_timeout, repro_args):
...
+          reproducible = True
+
+    if reproducible:
```
oliverchang, line 106: put lines 99-104 into a helper method. so this can look much simpler as:

```python
if self._crash_was_reproducible(result.output):
  logs.log('Repro success')
  ...
else:
 ...

File: src/python/bot/fuzzers/syzkaller/engine.py (hunk 4 above)
Comment (inferno-chromium) [reply src/python/bot/fuzzers/syzkaller/engine.py:135]: Is restart_time to try things multiple times ? maybe even a constant restart time is better than near the limit as process will be killed anyway. this constant can go in class constant or android/constants.py

File: src/python/bot/fuzzers/syzkaller/runner.py (hunk 1 above)
Comment (inferno-chromium) [reply src/python/bot/fuzzers/syzkaller/runner.py:100]: +1, class constant compiled var once.

File: src/python/bot/fuzzers/syzkaller/runner.py (hunk 5 above)
Comment (inferno-chromium) [reply src/python/bot/fuzzers/syzkaller/runner.py:106]: +1
Successfully reproduced crash
and
Failed to reproduce crash
slightly better version of message.

File: src/python/bot/fuzzers/syzkaller/runner.py (hunk 2 above)
Comment (inferno-chromium) [reply src/python/bot/fuzzers/syzkaller/runner.py:110]: +1

Comment (hzawawy) [reply src/python/bot/fuzzers/syzkaller/runner.py:110]: Based on https://github.com/google/clusterfuzz/blob/cdfe140489e85896a3e5f5747acc9ab3d0df7fdb/src/python/crash_analysis/crash_analyzer.py#L139, the return code is only needed when a crash is found.

File: src/python/bot/fuzzers/syzkaller/engine.py (hunk 4 above)
Comment (hzawawy) [reply src/python/bot/fuzzers/syzkaller/engine.py:135]: So experimentally, 120 seconds seems to work well. It will look for one crash and may exit before 120s is completed. I will make it as a constant as per the comment.

File: src/python/bot/fuzzers/syzkaller/runner.py (hunk 2 above)
Comment (inferno-chromium) [reply src/python/bot/fuzzers/syzkaller/runner.py:110]: So, this means from crash_analyzer is_crash looks for a non-zero return to identify this as a crash. so you need a non-zero one (and not 0). you can do result.return_code = int(reproducible).

File: src/python/bot/fuzzers/syzkaller/constants.py [hunk 6]
```diff
@@ -15,6 +15,6 @@
 Should not have any dependancies.
 """
 
-SYZ_REPRO = 'syz-repro'
```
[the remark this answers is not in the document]
inferno-chromium: SYZ_CRUSH is a very ambiguous name, production projects shouldn't even use such a binary name. For variable name, i still like SYZ_REPRO or SYZ_REPRO_BINARY

src/python/bot/fuzzers/syzkaller/runner.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/runner.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/runner.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/engine.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/runner.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/constants.py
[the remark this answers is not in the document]
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/runner.py
```diff
@@ -27,6 +27,8 @@
 from system import environment
 from system import new_process
 
+regex = re.compile(r'reproduced (\d+) crashes')
```
inferno-chromium, line 30: nit: capital for global
s/regex/REPRODUCE_REGEX

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -130,7 +132,8 @@ def reproduce(self, target_path, input_path, arguments, max_time):  # pylint: di
     syzkaller_runner = runner.get_runner(
         os.path.join(binary_dir, constants.SYZ_REPRO))
     repro_args = runner.get_config()
-    repro_args.extend(input_path)
+    repro_args.extend(
+        ['-infinite=false', '-restart_time={}s'.format(REPRO_TIME), input_path])
```
inferno-chromium, line 136: As per Max's comment, infinite=false means restart time is the max run time, so i think restart_time argument does not apply. if you want to keep restart_time, i think you need to set infinite=true. i am fine with removing restart_time arg if we are only trying to reproduce it once (and wont have more time).

hzawawy: From syzkaller:
	flagRestartTime = flag.Duration("restart_time", time.Hour, "how long to run the test")
	flagInfinite    = flag.Bool("infinite", true, "by default test is run for ever, -infinite=false to stop on crash")
Running under infinite=false and restart_time=120 means running until crash is found or till 2m0s (so possibly few tries). If you remove restart_time, it will only run once. I think we can afford to let it run few times, that s better than running once and failing to repro but maybe second or third could work.
Now infinite default value is true, if i remove it, and keep restart_time=120, this is the behavior:
2020/07/13 10:57:29 running infinitely and restarting VM every 2m0s
So even after a crash is found, it will still run.
2020/07/13 10:57:29 running infinitely and restarting VM every 2m0s
2020/07/13 10:57:29 reproducing from log file: /usr/local/google/home/hamzeh/HT75H1A00107/syzkaller/crashes/c684119dc4cf71e8d82ee041b591c3a09139a6cf/log0
2020/07/13 10:57:29 booting 1 test machines...
2020/07/13 10:57:29 vm-0: starting
2020/07/13 10:58:20 failed to associate adb device HT75H1A00107 with console: no console is associated with this device
2020/07/13 10:58:20 falling back to 'adb shell dmesg -w'
2020/07/13 10:58:20 note: some bugs may be detected as 'lost connection to test machine' with no kernel output
2020/07/13 10:58:21 device HT75H1A00107: battery level 98%, OK
2020/07/13 10:58:22 vm-0: crushing...
2020/07/13 10:58:32 vm-0: crash: panic:LINE, dump:LINE
2020/07/13 10:58:32 vm-0: starting
2020/07/13 10:58:32 saving crash panic:LINE, dump:LINE to /usr/local/google/home/hamzeh/HT75H1A00107/syzkaller/crashes/ea52b31f6e5d74acc123f0a3e7b8d77c021aa5bd
2020/07/13 10:58:32 instances executed: 1, crashes: 1
2020/07/13 10:59:20 device HT75H1A00107: battery level 98%, OK
2020/07/13 10:59:21 vm-0: crushing...

hzawawy: 2020/07/13 10:59:31 saving crash panic:LINE, dump:LINE to /usr/local/google/home/hamzeh/HT75H1A00107/syzkaller/crashes/ea52b31f6e5d74acc123f0a3e7b8d77c021aa5bd
2020/07/13 10:59:31 instances executed: 2, crashes: 2

inferno-chromium: You are right, sorry for creating this confusion. 

I think we might want to lower down restart_time=120 in future as it wont get change to retry as TEST_TIMEOUT=120 as well. so, it will only execute it once.

PR #1907, opened 2020-07-18
Fix android platform check to account for ANDROID_KERNEL.

inferno-chromium:


src/python/system/process_handler.py
```diff
@@ -144,7 +144,8 @@ def run_process(cmdline,
   # This is used when running scripts on native linux OS and not on the device.
   # E.g. running a fuzzer to generate testcases or launcher script.
   plt = environment.platform()
-  if plt in ['ANDROID', 'FUCHSIA'] and (not testcase_run or launcher):
+  is_android = environment.is_android(plt)
```
oliverchang, line 147: nit: can we do something like

```python
runs_on_device = environment.is_android(plt) or plt == 'FUCHSIA'
if runs_on_device and (not testcase_run or launcher):
 ...
```

I think that's a bit clearer?

inferno-chromium: Done!

inferno-chromium:
@hzawawy as fyi. this makes it easy to add any ANDROID_* platform.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -1962,7 +1962,7 @@ def get_crash_data(crash_data, symbolize_flag=True):
       continue
 
   # Only get repo.prop if we have an Android kernel or KASAN crash
-  if not state.android_kernel_repo and (found_android_kernel_crash or is_kasan):
+  if environment.is_android() and (found_android_kernel_crash or is_kasan):
```
inferno-chromium, line 1965: Just fyi @mspectorgoogle, this would need a rebase in your cl, since this block runs outside loop, there is no need to check state.android_kernel_repo. in your cl, you can even remove this var from constructor and then just use a local var inside this block.

Issue #1899, opened 2020-07-12
build_afl.bash fails to download correct files because cs.chromium.org does a 302 redirect without a specific cookie


phwd:
[build_afl.bash](https://google.github.io/clusterfuzz/setting-up-fuzzing/libfuzzer-and-afl/#afl) tries to download files from cs.chromium.org however that site does a 302 to a newer "code search"

adding `-H "cookie:REDIRECT_STATUS=optout" ` to each curl command in the bash script will fix the download issue (until Google decides to change it again)

`curl -H "cookie:REDIRECT_STATUS=optout" -O "https://cs.chromium.org/codesearch/f/chromium/src/third_party/afl/src/$source_file"`

`curl -H "cookie:REDIRECT_STATUS=optout" "https://cs.chromium.org/codesearch/f/chromium/src/third_party/afl/src/llvm_mode/afl-llvm-rt.o.c" > "llvm_mode/afl-llvm-rt.o.c"`

`curl -H "cookie:REDIRECT_STATUS=optout" -O "https://cs.chromium.org/codesearch/f/chromium/src/third_party/libFuzzer/src/afl/afl_driver.cpp"`

Dor1s:
Thanks for reporting this. We should probably use https://github.com/google/AFL now. Would you be interested in uploading a pull request and contributing to ClusterFuzz? :)

PR #1908, opened 2020-07-18
Update script to ignore 302 redirects for curl

phwd:
build_afl.bash tries to download files from cs.chromium.org however that site does a 302 to a newer "code search"
https://github.com/google/clusterfuzz/issues/1899

### DIFF
--- a/docs/setting-up-fuzzing/build_afl.bash
+++ b/docs/setting-up-fuzzing/build_afl.bash
@@ -36,15 +36,15 @@ declare -a afl_sources=(
 )
 for source_file in "${afl_sources[@]}"
 do
-  curl -O "https://cs.chromium.org/codesearch/f/chromium/src/third_party/afl/src/$source_file"
+  curl -H "cookie:REDIRECT_STATUS=optout" -O "https://cs.chromium.org/codesearch/f/chromium/src/third_party/afl/src/$source_file"
 done
 make afl-fuzz afl-showmap
 
 # Build AFL runtime sources needed to link against the fuzz target.
 mkdir -p llvm_mode
-curl "https://cs.chromium.org/codesearch/f/chromium/src/third_party/afl/src/llvm_mode/afl-llvm-rt.o.c" > "llvm_mode/afl-llvm-rt.o.c"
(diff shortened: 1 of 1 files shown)


phwd:
@googlebot I signed it!

docs/setting-up-fuzzing/build_afl.bash
```diff
@@ -36,15 +36,15 @@ declare -a afl_sources=(
 )
 for source_file in "${afl_sources[@]}"
 do
-  curl -O "https://cs.chromium.org/codesearch/f/chromium/src/third_party/afl/src/$source_file"
+  curl -H "cookie:REDIRECT_STATUS=optout" -O "https://cs.chromium.org/codesearch/f/chromium/src/third_party/afl/src/$source_file"
```
inferno-chromium, line 39: Instead, can you update the link to the new location.

Dor1s: Would it be better to use google/AFL GitHub repo instead?

phwd: Can you provide the link that points to the new raw location at CodeSearchV2?

The 302 is incorrect to get the raw file. It wants to point to https://source.chromium.org/chromium/chromium/src/+/master:third_party/afl/src/afl-fuzz.c

which means to get the raw I'd have to go to the gitiles version and append format TEXT https://chromium.googlesource.com/chromium/src/+/master/third_party/afl/src/afl-fuzz.c?format=TEXT then base64 decode the response to get the file.

The reason why I chose the cookie was because CodeSearchV2 doesn't offer a nice way to download the file without base64 decoding it first. https://github.com/google/gitiles/issues/106.

Using google AFL repo it would be https://raw.githubusercontent.com/google/AFL/master/afl-fuzz.c

So the switch would be something like

curl -I "https://raw.githubusercontent.com/google/AFL/master/$source_file"

inferno-chromium: https://raw.githubusercontent.com/google/AFL/* links SGTM, no need to use chromium ones.

phwd: Github changed as the source and I added "android-ashmem.h" to the list of the files as there was an update https://github.com/google/AFL/pull/46. I also chose the Github LLVM project for the AFL driver cpp file.

PR #1909, opened 2020-07-19
Fix JSON serialization exception in cleanup task.

inferno-chromium:
For _get_crash_occurrence_platforms_from_crash_parameters,
Memoize module caches result as json, so revert to using lists
than sets.


inferno-chromium:
TBR exception fix.

PR #1910, opened 2020-07-19
Fix stack frame linkification issue with /proc/self/cwd paths.

inferno-chromium:
Fixes b/158773995


src/python/build_management/source_mapper.py
```diff
@@ -225,6 +225,9 @@ def normalize_source_path(path):
   # out/<build_dir>/ at time of build generation (chromium only).
   path = utils.remove_prefix(path, '../../')
 
+  # Remove /proc/self/cwd prefix if exists.
```
oliverchang, line 228: Can you mention that this is something bazel can add?

inferno-chromium: Done!

PR #1911, opened 2020-07-20
Fix log parse failure on newer versions of Android.

inferno-chromium:


inferno-chromium:
Don't see any other less hacky way to fix regex.

inferno-chromium:
TBR

PR #1915, opened 2020-07-20
Remove entropic and value profile argument from repro stack.

inferno-chromium:
These are fuzzing arguments and not needed during reproduction.
Also create confusion for developer.

### DIFF
--- a/src/python/bot/fuzzers/libFuzzer/engine.py
+++ b/src/python/bot/fuzzers/libFuzzer/engine.py
@@ -311,16 +311,19 @@ class LibFuzzerEngine(engine.Engine):
     })
 
     # Remove fuzzing arguments before merge and dictionary analysis step.
-    arguments = options.arguments[:]
-    libfuzzer.remove_fuzzing_arguments(arguments)
+    merge_arguments = options.arguments[:]
+    libfuzzer.remove_fuzzing_arguments(merge_arguments, is_merge=True)
     self._merge_new_units(target_path, options.corpus_dir, new_corpus_dir,
-                          options.fuzz_corpus_dirs, arguments, parsed_stats)
+                          options.fuzz_corpus_dirs, merge_arguments,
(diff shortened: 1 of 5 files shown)


src/python/tests/core/bot/fuzzers/libFuzzer/engine_test.py
```diff
@@ -268,6 +272,8 @@ def mock_merge(*args, **kwargs):  # pylint: disable=unused-argument
             '-timeout=123',
             '-dict=blah.dict',
             '-max_len=9001',
+            '-entropic=1',
```
inferno-chromium, line 275: See crash.reproduce_args result as test check
six.assertCountEqual(self, ['-arg=1', '-timeout=60'], crash.reproduce_args)

inferno-chromium:
@Dor1s - can you take another look please, any ideas on making this less hacky.

src/python/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -311,16 +311,19 @@ def fuzz(self, target_path, options, reproducers_dir, max_time):
     })
 
     # Remove fuzzing arguments before merge and dictionary analysis step.
-    arguments = options.arguments[:]
-    libfuzzer.remove_fuzzing_arguments(arguments)
+    merge_arguments = options.arguments[:]
+    libfuzzer.remove_fuzzing_arguments(merge_arguments, is_merge=True)
```
Dor1s, line 315: my only other alternative is to get rid of `is_merge` parameter and add value profile flag here explicitly, but I lean towards your implementation as it keeps all the code modifying the arguments inside `remove_fuzzing_arguments`

inferno-chromium: yes right, explicit for is_merge feels slightly better.

PR #1919, opened 2020-07-22
Request bisection for regression earlier.

oliverchang:
Previously, a regressed and fixed bisect were requested at the same time
when a Testcase is detected to be fixed in progression_task.

This PR changes this: A regressed bisection can be requested earlier,
even if the Testcase isn't fixed. To ensure that tasks are only
requested once we add corresponding testcase metadata.

### DIFF
--- a/src/python/bot/tasks/task_creation.py
+++ b/src/python/bot/tasks/task_creation.py
@@ -252,11 +252,13 @@ def _get_commits(commit_range, job_type):
   return old_commit, new_commit
 
 
-def request_bisection(testcase):
+def request_bisection(testcase_id):
   """Request precise bisection."""
   pubsub_topic = local_config.ProjectConfig().get('bisect_service.pubsub_topic')
   if not pubsub_topic:
     return
+
+  testcase = data_handler.get_testcase_by_id(testcase_id)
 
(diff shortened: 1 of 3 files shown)


src/python/bot/tasks/task_creation.py
```diff
@@ -271,12 +273,19 @@ def request_bisection(testcase):
   if not target:
     return
 
-  _make_bisection_request(pubsub_topic, testcase, target, 'regressed')
-  _make_bisection_request(pubsub_topic, testcase, target, 'fixed')
+  # Only make 1 request of each type per testcase.
```
inferno-chromium, line 276: I think you should be refetch testcase to avoid updating a stale one.
testcase = data_handler.get_testcase_by_id(testcase.key.id())

inferno-chromium, line 276: What happens if that request fails ? Should we check pubsub_client.publish return value in _make_bisection_request

oliverchang: Only way that can fail will be via an exception.

oliverchang: Done!

PR #1924, opened 2020-07-24
Upgrade ML GPU bot config to run TF 2.3.0rc2 and other appropriate deps (#1540).

Dor1s:
I'm not quite sure how to test this. I did try to install things manually on a bot and that worked, but we have slightly different setup for deploying CF bots.

One idea might be to update Chromium's ml-with-gpu image and just re-create those bots. Even if they break, we won't lose much.

Another option is to set up a separate GCP project with a ml-with-gpu bot. It'll take a bit longer but won't be "testing in prod".

Maybe there are any other tricks that I don't remember?

### DIFF
--- a/docker/ml-with-gpu/Dockerfile
+++ b/docker/ml-with-gpu/Dockerfile
@@ -17,31 +17,31 @@ WORKDIR /data
 
 # Note that CUDA Toolkit version might work with only a particular version of
 # NVidia drivers, e.g. cuda 9.2 needs drivers 396.37 and cuda 9.0 needs 384.130.
-# Drivers are installed in
-# clusterfuzz/configs/gce/linux-init-internal-ml-with-gpu.yaml.
+# The drivers are installed in
+# configs/test/gce/linux-init-ml-with-gpu.yaml.
 # cuDNN packages installed later in this script must match the CUDA version too.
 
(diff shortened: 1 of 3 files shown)


docker/ml-with-gpu/Dockerfile
```diff
@@ -17,31 +17,31 @@ WORKDIR /data
...
+
+# Install development and runtime libraries (~4GB)
+RUN apt-get install -y --no-install-recommends \
+    cuda-10-1 \
+    libcudnn7=7.6.4.38-1+cuda10.1  \
+    libcudnn7-dev=7.6.4.38-1+cuda10.1
+
+# Install TensorRT. Requires that libcudnn7 is installed above.
+RUN apt-get install -y --no-install-recommends \
+    libnvinfer6=6.0.1-1+cuda10.1 \
```
inferno-chromium, line 35: nit: CUDA_VERSION=6.0.1-1+cuda10.1 at start of command and then use in three places as $CUDA_VERSION ?
Same for libcudnn7 above.

Dor1s: thanks for the suggestion! I don't think that would improve readability a lot, plus it's not CUDA version (cuda is 10.1), so I'm just leaving this like suggested in the docs: https://www.tensorflow.org/install/gpu#ubuntu_1604_cuda_101

PR #1936, opened 2020-07-31
Allow honggfuzz targets in testcase upload page.

inferno-chromium:


inferno-chromium:
Some day we will refactor these checks, till then this one.

oliverchang:
I think we are just waiting on AFL refactor :) Then we will have a single source of truth for engine names.

PR #1942, opened 2020-08-03
Do not use Entropic and Focus Function together (OSS-Fuzz#4242).

Dor1s:


src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1789,7 +1789,9 @@ def pick_strategies(strategy_pool,
                                  fuzzer_path)):
     fuzzing_strategies.append(strategy.RECOMMENDED_DICTIONARY_STRATEGY.name)
 
-  if strategy_pool.do_strategy(strategy.ENTROPIC_STRATEGY):
+  # Entropic isn't compatible with focus function.
+  if (strategy_pool.do_strategy(strategy.ENTROPIC_STRATEGY) and not any(
```
inferno-chromium, line 1793: Should we prioritize dataflow over entropic or entropic over dataflow. i think entropic over dataflow, so move the block up and then check entropic not in any arguments ?

jonathanmetzman: It might help me slightly in my analysis if entropic is actually running 50% of the time so I have a slight preference for prioritizing entropic.

Dor1s: Just for the sake of discussion, I think it's better to give the preference to dataflow, as it's supposed to be useful more often (i.e. when the corpus is saturated) compared to entropic (which presumably is most effective in the first days / weeks of a fuzz target lifetime).

But I'll swap these as per your suggestion, since anyway dataflow isn't yet deployed and instead of having this case-by-case preference decisions we need to refactor the strategy selection algorithm: https://github.com/google/clusterfuzz/issues/1943

PR #1949, opened 2020-08-04
Remove support for Python 2 fuzzers.

mbarbella-chromium:
PTAL. There are a few minor breakages left which I'll try to resolve (3 minor issues with external fuzzers), but overall I don't think it's worth keeping Python 2 around for them. After this lands there are no remaining Python 2 dependencies on bots that I know of.

Oliver, are there any other Python 2 dependencies you can think of? I don't think it's urgent that we remove Python 2 from the images, but it should be safe to do so after this.

### DIFF
--- a/src/python/system/shell.py
+++ b/src/python/system/shell.py
@@ -351,7 +351,7 @@ def get_free_disk_space(path='/'):
   return psutil.disk_usage(path).free
 
 
-def get_interpreter(file_to_execute, is_blackbox_fuzzer=False):
+def get_interpreter(file_to_execute):
   """Gives the interpreter needed to execute |file_to_execute|."""
   interpreters = {
       '.bash': 'bash',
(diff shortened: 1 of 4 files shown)


src/python/system/shell.py
```diff
@@ -368,12 +368,6 @@ def get_interpreter(file_to_execute, is_blackbox_fuzzer=False):
   except KeyError:
     return None
 
-  # TODO(mbarbella): Remove this when fuzzers have been migrated to Python 3.
-  if (is_blackbox_fuzzer and interpreter == sys.executable and
```
inferno-chromium, line 372: Can you clean up use of is_blackbox_fuzzer argument as it is not needed.

mbarbella-chromium: Good catch. Done!

oliverchang:
Nice! Can't think of any other Python 2 dependencies. We can try go ahead with removing it.

PR #1956, opened 2020-08-06
Only retry GCS functions on GoogleCloudErrors.

### LINKED ISSUE #1955: google_cloud_utils.storage should not retry on out of disk exceptions.

oliverchang:
Fixes #1955.

Only functions which use the google.cloud.storage client library (and not the discovery APIs) are changed.

### DIFF
--- a/src/python/google_cloud_utils/storage.py
+++ b/src/python/google_cloud_utils/storage.py
@@ -220,7 +220,7 @@ class GcsProvider(StorageProvider):
       bucket = client.bucket(bucket_name)
       blob = bucket.blob(path, chunk_size=self._chunk_size())
       blob.download_to_filename(local_path)
-    except google.cloud.exceptions.GoogleCloudError:
+    except google.cloud.google.cloud.exceptions.GoogleCloudError:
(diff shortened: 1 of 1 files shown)


oliverchang:
> There seem to be more functions that need changing, like get_acl, set_acl, get, can you take a quick repass through.

As mentioned in description, those don't use the google.cloud.storage client library, so didn't change those for now :)

inferno-chromium:
> > There seem to be more functions that need changing, like get_acl, set_acl, get, can you take a quick repass through.
> 
> As mentioned in description, those don't use the google.cloud.storage client library, so didn't change those for now :)

Sorry about that, should read things more carefully.

oliverchang:
> > > There seem to be more functions that need changing, like get_acl, set_acl, get, can you take a quick repass through.
> > 
> > 
> > As mentioned in description, those don't use the google.cloud.storage client library, so didn't change those for now :)
> 
> Sorry about that, should read things more carefully.

We should probably also look at converting those at some point. Previously we needed them to use a different library due to App Engine Py2, but this is no longer an issue :)

PR #1957, opened 2020-08-06
Shifted several handlers to flask.

iamarshsingh:
1. Configuration handlers.
2. Fuzzers hanlders.
3. Login handlers.
4. Corpora handlers.

Please have a look!

### DIFF
--- a/src/appengine/handlers/base_handler_flask.py
+++ b/src/appengine/handlers/base_handler_flask.py
@@ -181,7 +181,7 @@ class Handler(MethodView):
     response.headers['X-Frame-Options'] = 'deny'
     return response
 
-  def render(self, path, values=None, status=200, response=None):
+  def render(self, path, values=None, status=200):
     """Write HTML response."""
     if values is None:
       values = {}
@@ -203,8 +203,7 @@ class Handler(MethodView):
 
     template = _JINJA_ENVIRONMENT.get_template(path)
 
-    if not response:
-      response = Response()
+    response = Response()
     response = self._add_security_response_headers(response)
     response.headers['Content-Type'] = 'text/html'
     response.data = template.render(values)
@@ -216,10 +215,9 @@ class Handler(MethodView):
     """A hook for modifying values before render_json."""
     return response
 
-  def render_json(self, values, status=200, response=None):
+  def render_json(self, values, status=200):
     """Write JSON response."""
-    if not response:
-      response = Response()
+    response = Response()
     response = self._add_security_response_headers(response)
     response.headers['Content-Type'] = 'application/json'
     response = self.before_render_json(values, status, response)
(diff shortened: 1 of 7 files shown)


src/appengine/handlers/base_handler_flask.py
```diff
@@ -203,7 +203,8 @@ def render(self, path, values=None, status=200):
 
     template = _JINJA_ENVIRONMENT.get_template(path)
 
-    response = Response()
+    if not response:
+      response = Response()
```
Dor1s, line 207: possibly dumb question: can `Response()` be used as a default value (on line 184)?

oliverchang: I don't think we can, since that instance would be shared for all invocations.

src/appengine/handlers/base_handler_flask.py
```diff
@@ -181,7 +181,7 @@ def _add_security_response_headers(self, response):
     response.headers['X-Frame-Options'] = 'deny'
     return response
 
-  def render(self, path, values=None, status=200):
+  def render(self, path, values=None, status=200, response=None):
```
oliverchang, line 184: Could you explain a bit why we need to add the response argument? Seems like we can always just add stuff to the response after the call to render() to keep the interface clean.

src/appengine/handlers/login.py
```diff
@@ -81,5 +84,6 @@ def get(self):
       # Even if the revoke failed, remove the cookie.
       logs.log_error('Failed to revoke session cookie.')
 
-    self.response.delete_cookie('session')
-    self.redirect(self.request.get('dest'))
+    response = Response()
+    response.delete_cookie('session')
+    return self.redirect(request.get('dest'), response=response)
```
oliverchang, line 89: I'm a bit confused how this works here. redirect doesn't take in a response argument? 

And again, we can delete the cookie from after getting the return value from self.redirect instead of doiung this.

src/appengine/handlers/login.py
```diff
@@ -16,63 +16,66 @@
...
+    response = Response()
+    response.set_cookie(
         'session',
         session_cookie,
         expires=expires,
         httponly=True,
         secure=True,
         overwrite=True)
-    self.render_json({'status': 'success'})
+    return self.render_json({'status': 'success'}, response=response)
```
oliverchang, line 70: Can we just set_cookie on the return value from render_json, then return it?

src/appengine/handlers/base_handler_flask.py
iamarshsingh: yes, sorry. we actually don't need it. I will clean this approach.

src/appengine/handlers/login.py
iamarshsingh: yes, fixed in the latest commit.

PR #1960, opened 2020-08-07
Fix NDB transaction exception type.

oliverchang:
Use the base Error type, as TransactionFailedError no longer exists.


oliverchang:
TBR. /gcbrun

PR #1962, opened 2020-08-07
Fix OSS-Fuzz build status cron after json format update.

inferno-chromium:
Fixes https://github.com/google/oss-fuzz/issues/4278

### DIFF
--- a/src/appengine/handlers/cron/oss_fuzz_build_status.py
+++ b/src/appengine/handlers/cron/oss_fuzz_build_status.py
@@ -15,7 +15,6 @@
 
 from builtins import str
 import datetime
-import itertools
 import json
 import re
 import requests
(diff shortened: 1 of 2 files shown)


inferno-chromium:
> awesome!

Thanks for reviewing!

src/appengine/handlers/cron/oss_fuzz_build_status.py
```diff
@@ -185,19 +184,26 @@ def send_reminder(issue_tracker, issue_id, build_id):
...
+      project_name = project['name']
+      builds = project['history']
+      if not builds:
+        continue
 
+      build_failure = get_build_failure(project_name, build_type)
       if not build_failure:
         continue
 
+      build = builds[0]
```
Dor1s, line 203: while trying to understand why this part is needed, I realized that we might want to run the cron more often now, given that more frequent builds are supported. It runs daily, we can probably run it 4 times a day now?

inferno-chromium: Sure, feel free to make it 4 times.

PR #1966, opened 2020-08-10
Migrated several UI handlers to flask.

iamarshsingh:
1. bots
2. crash_stats
3. home
4. download
5. help_redirector
6. issue_redirector
7. all testcase handlers

Please have a look!


src/appengine/server_flask.py
```diff
@@ -70,25 +94,67 @@ def register_routes(flask_app, routes):
...
     ('/delete-external-user-permission',
      configuration.DeleteExternalUserPermission),
+    ('/crash-stats/load', crash_stats.JsonHandler),
+    ('/crash-stats', crash_stats.Handler),
     ('/corpora', corpora.Handler),
     ('/corpora/create', corpora.CreateHandler),
     ('/corpora/delete', corpora.DeleteHandler),
+    ('/docs', help_redirector.DocumentationHandler),
+    ('/download', download.Handler),
+    ('/download/<resource>', download.Handler),
```
iamarshsingh, line 118: @oliverchang here in the webapp2 url routing there was a feature for optional param but I was not able to find any similar feature for flask. The alternative was to add two rules, with and without variable. Is there any other alternative you would suggest or should we go with this?

src/appengine/handlers/download.py
```diff
@@ -79,18 +82,15 @@ def handle_public_testcase(self, blob_info, testcase, fuzzer_binary_name):
...
 
-  def get(self, resource):
+  @handler_flask.get(handler_flask.HTML)
+  def get(self, resource=None):
     """Handle a get request with resource."""
     testcase = None
-    testcase_id = self.request.get('testcase_id')
-
+    testcase_id = request.get('testcase_id')
+    print(request.url)
```
oliverchang, line 92: Remove prints.

src/appengine/server_flask.py
oliverchang: Seems fine to me.

iamarshsingh:
> This looks mostly mechanical so nothing much to comment, but will need to test this careful on staging since it touches quite a few things. In particular the testcase upload one is important since that's relied on by some internal clients.

yes, make sense. Please share the staging link once that's done and I can help in testing.

src/appengine/handlers/upload_testcase.py
```diff
@@ -581,24 +580,25 @@ def do_post(self):
...
+    return self.render_json({'id': '%s' % testcase_id})
 
 
-class UploadHandler(UploadHandlerCommon, base_handler.GcsUploadHandler):
+class UploadHandler(UploadHandlerCommon, base_handler_flask.GcsUploadHandler):
   """Handler that uploads the testcase file."""
 
   # pylint: disable=unused-argument
-  def before_render_json(self, values, status):
+  def before_render_json(self, values, status, response):
```
oliverchang, line 590: Is anything actually modifying `response`? We can just fix up the base_handler's before_render_json to not return anything and therefore not need this argument at all.

src/appengine/handlers/download.py
```diff
@@ -79,18 +82,13 @@ def handle_public_testcase(self, blob_info, testcase, fuzzer_binary_name):
...
 
-    self._send_blob(
-        blob_info,
-        testcase.key.id(),
-        is_minimized=True,
-        fuzzer_binary_name=fuzzer_binary_name)
     return True
 
-  def get(self, resource):
+  @handler_flask.get(handler_flask.HTML)
```
oliverchang, line 87: Why do we need to add HTML ?

oliverchang:
I'm also seeing some weirdness with the testcase list page hanging on staging.

src/appengine/libs/auth.py
```diff
@@ -178,6 +178,7 @@ def create_session_cookie(id_token, expires_in):
 
 def get_current_request():
   """Get the current request."""
+  # TODO(singharshdeep): Remove webapp.get_request() after flask migration.
```
oliverchang, line 181: We call setattr() on this. Does this work as expected on flask.request, which does some magic to make flask.request proxy the actual request, or can this lead to inconsistent states?

oliverchang: https://werkzeug.palletsprojects.com/en/1.0.x/wrappers/#mutability-and-reusability-of-wrappers mentions that requests are immutable.

src/appengine/handlers/upload_testcase.py
```diff
@@ -609,25 +609,26 @@ def __init__(self, name, value):
...
     """Add upload info when the request fails."""
-    values['uploadUrl'] = self.request.host_url + UPLOAD_URL
+    values['uploadUrl'] = request.host_url + UPLOAD_URL
+    return response
 
   def get_upload(self):
     """Get the upload."""
-    uploaded_file = self.request.POST.get('file')
+    uploaded_file = request.POST.get('file')
```
oliverchang, line 623: This doesn't work in flask. 

```
Request' object has no attribute 'POST'
```

src/appengine/handlers/upload_testcase.py
iamarshsingh: no, there are no current handlers which modifies response. yes, probably its better to not keep it. I will remove this in next commit

src/appengine/handlers/download.py
iamarshsingh: yes, here we should create another type such as handler.REDIRECT or something. Actually calling get handler is important to make request.get works.

src/appengine/libs/auth.py
iamarshsingh: yes, as we discussed I have used flask.g for all such setattr() and getattr() on requests.

src/appengine/libs/auth.py
```diff
@@ -137,11 +137,17 @@ def get_current_user():
   if iap_email:
     return User(iap_email)
 
-  oauth_email = getattr(current_request, '_oauth_email', None)
+  if flask.request:
+    oauth_email = getattr(flask.g, '_oauth_email', None)
```
oliverchang, line 141: rather than all these ifs, you can just do it once

if flask.request:
  cache_backing = flask.request
else:
  cache_backing = get_current_request()

and use cache_backing everywhere.

src/appengine/libs/request_cache.py
```diff
@@ -31,18 +32,24 @@ def __init__(self, cache_key, capacity):
...
     if not request:
       # Not a request (e.g. in a unit test). Should not happen in production.
       logs.log_error('No request found for cache.')
       return None
 
     key = '__cache:' + self._cache_key
 
-    backing = getattr(request, key, None)
+    if flask.request:
+      backing = getattr(flask.g, key, None)
```
oliverchang, line 44: same here, just have 1 if statement for this.

src/appengine/libs/auth.py
```diff
@@ -137,17 +137,13 @@ def get_current_user():
   if iap_email:
     return User(iap_email)
 
-  if flask.request:
-    oauth_email = getattr(flask.g, '_oauth_email', None)
-  else:
-    oauth_email = getattr(current_request, '_oauth_email', None)
+  cache_backing = get_cache_backing()
+
```
oliverchang, line 141: nit: remove newline.

src/appengine/libs/request_cache.py
```diff
@@ -40,16 +39,11 @@ def cache(self):
 
     key = '__cache:' + self._cache_key
 
-    if flask.request:
-      backing = getattr(flask.g, key, None)
-    else:
-      backing = getattr(request, key, None)
+    cache_backing = auth.get_cache_backing()
```
oliverchang, line 42: get_cache_backing() shouldn't be part of the `auth` module.  It should be part of this module instead.

src/appengine/libs/request_cache.py
```diff
@@ -40,16 +39,11 @@ def cache(self):
```
oliverchang, line 39: also remove the request checks and check on cache_backing instead.

src/appengine/libs/request_cache.py
```diff
@@ -15,12 +15,21 @@
...
 from libs import auth
 from metrics import logs
 
 
+def get_cache_backing():
+  """Get the cache backing for saving current context data."""
+  # TODO(singharshdeep): Remove get_current_request() after flask migration.
+  if flask.request:
+    return flask.g
+  return auth.get_current_request()
```
oliverchang, line 30: nit: move get_current_request here too. auth.get_current_request wasn't intended to be used from elsewhere and doesn't make sense as a place for this.

src/python/tests/appengine/libs/request_cache_test.py
```diff
@@ -53,8 +53,7 @@ class CacheTest(unittest.TestCase):
 
   def setUp(self):
     test_helpers.patch(self, [
-        'libs.auth.get_current_request',
-        'libs.auth.get_cache_backing'
+        'libs.auth.get_current_request', 'libs.request_cache.get_cache_backing'
```
oliverchang, line 56: do you still need to mock get_current_request?

iamarshsingh: no, we don't need it. Removed in latest commit

src/appengine/handlers/upload_testcase.py
```diff
@@ -609,25 +608,25 @@ def __init__(self, name, value):
...
   # pylint: disable=unused-argument
   def before_render_json(self, values, status):
     """Add upload info when the request fails."""
-    values['uploadUrl'] = self.request.host_url + UPLOAD_URL
+    values['uploadUrl'] = request.host_url + UPLOAD_URL
 
   def get_upload(self):
     """Get the upload."""
-    uploaded_file = self.request.POST.get('file')
+    uploaded_file = request.get('file')
```
oliverchang, line 621: Actually, seems like I didn't test this properly last time. 

This needs to be updated:

```
-    uploaded_file = request.get('file')
-    if not isinstance(uploaded_file, cgi.FieldStorage):
+    uploaded_file = request.files.get('file')
+    if not uploaded_file:
       raise helpers.EarlyExitException('File upload not found.', 400)

-    bytes_io = NamedBytesIO(uploaded_file.filename, uploaded_file.file.read())
+    bytes_io = NamedBytesIO(uploaded_file.filename, uploaded_file.stream.read())
     key = blobs.write_blob(bytes_io)
     return blobs.get_blob_info(key)
```

oliverchang: (Would be a good idea to check documentation before making changes in the future).

oliverchang:
In the interests of time, I'll merge this and commit the upload fix separately.

PR #1967, opened 2020-08-11
Fix alert deployment.

oliverchang:
- Only replace BOT_COUNT:region(space or end of line), to prevent issues
  with overlapping region names.
- Add retries.


oliverchang:
TBR

PR #1969, opened 2020-08-13
Pass --coverfile to syz-manager to get raw coverage.

mspectorgoogle:


src/python/bot/fuzzers/syzkaller/runner.py
```diff
@@ -64,6 +64,11 @@ def get_config():
   return ['-config', json_config_path]
 
 
+def get_cover_file_path():
+  """Return location of cover file for Syzkaller."""
```
inferno-chromium, line 68: s/cover/coverage ? for description

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -79,13 +79,10 @@ def prepare(self, corpus_dir, target_path, unused_build_dir):  # pylint: disable
     Returns:
       A FuzzOptions object."""
     self.prepare_binary_path()
-    config = runner.get_config()
+    args = runner.get_config()
+    args += ['--coverfile', runner.get_cover_file_path()]
```
inferno-chromium, line 83: Since we are calling a helper in runner, why not just add this in get_config directly. We need this coveragefile in all cases.

mspectorgoogle: We only want it for fuzz, doesn't make sense to get it other places

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -79,10 +79,9 @@ def prepare(self, corpus_dir, target_path, unused_build_dir):  # pylint: disable
     Returns:
       A FuzzOptions object."""
     self.prepare_binary_path()
-    config = runner.get_config()
     return SyzkallerOptions(
         corpus_dir,
-        config,
+        runner.get_cover_file_path(),
```
inferno-chromium, line 84: I think you meant runner.get_config(), but good to just revert to local var as before.

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -111,8 +110,10 @@ def fuzz(self, target_path, options, unused_reproducers_dir=None, max_time=0):
 
     # Directory to place new units.
     self._create_temp_corpus_dir('new')
+    args = options.arguments
```
inferno-chromium, line 113: newline before this line as it is not related to block "# Directory to place new units."

PR #1971, opened 2020-08-14
Fix upload API endpoint.

oliverchang:
Flask API has differences from webapp2.


oliverchang:
TBR

oliverchang:
@iamarshsingh FYI.

PR #1972, opened 2020-08-14
Fix testcase cache handler.

oliverchang:


oliverchang:
/gcbrun. TBR

@iamarshsingh please check for similar cases in future PRs. Also, try running cron jobs by just using curl locally.

PR #1973, opened 2020-08-14
Fix home cache handler.

oliverchang:
Flask methods require a return value. Also simplify the return value of
testcase list cache to be consistent.

### DIFF
--- a/src/appengine/handlers/home.py
+++ b/src/appengine/handlers/home.py
@@ -14,13 +14,12 @@
 """Home page handler_flask."""
 from __future__ import absolute_import
 
-from . import base_handler_flask
-
 from base import external_users
 from base import memoize
 from base import utils
 from datastore import data_handler
 from datastore import data_types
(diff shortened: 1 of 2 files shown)


oliverchang:
TBR

oliverchang:
> LGTM

Please remember to check this on other handlers that you migrate in the future :)

iamarshsingh:
> Please remember to check this on other handlers that you migrate in the future :)
yes I was also confused about the cron jobs return, will take care of this in future and thanks for the fix.

src/appengine/handlers/home.py
```diff
@@ -152,3 +151,4 @@ def get(self):
     """GET handler_flask."""
     # pylint: disable=unexpected-keyword-arg
     _get_all_project_results(__memoize_force__=True)
+    return 'OK'
```
inferno-chromium, line 154: Should we define this in constants somewhere and use everywhere.

PR #1974, opened 2020-08-16
Migrated rest UI handlers to Flask.

iamarshsingh:
1. commit_range
2. coverage_report
3. fuzzer_stats
4. gcd_redirector
5. performance_report
6. report_csv_failure
7. reproduce_tool handlers
8. revisions_info

Please have a look!

### DIFF
--- a/src/appengine/handlers/commit_range.py
+++ b/src/appengine/handlers/commit_range.py
@@ -15,11 +15,12 @@
 
 import json
 
+from flask import request
 from google_cloud_utils import big_query
-from handlers import base_handler
+from handlers import base_handler_flask
 from libs import crash_access
 from libs import filters
-from libs import handler
+from libs import handler_flask
 from libs import helpers
 from libs.query import big_query_query
 
@@ -135,12 +136,11 @@ def get(params, query, offset, limit):
   return result.rows, result.total_count
 
 
-def get_result(this):
+def get_result():
   """Get the result for the crash stats page."""
-  params = {k: v for k, v in this.request.iterparams()}
+  params = dict(request.iterparams())
   params['type'] = params.get('type', 'regression')
-  page = helpers.cast(
-      this.request.get('page') or 1, int, "'page' is not an int.")
+  page = helpers.cast(request.get('page') or 1, int, "'page' is not an int.")
 
   is_revision_empty = 'revision' not in params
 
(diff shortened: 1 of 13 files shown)


src/appengine/server_flask.py
```diff
@@ -47,12 +57,22 @@
...
 from handlers.testcase_detail import update_issue
+from libs import helpers
 from metrics import logs
 from system import environment
 
 ndb_client = ndb.Client()
 
 
+class EmptyHandler(base_handler_flask.Handler):
+  """Empty hanlder for flask migration."""
```
oliverchang, line 68: nit: spelling.

src/appengine/handlers/revisions_info.py
```diff
@@ -51,7 +52,8 @@ def get(self):
       raise helpers.EarlyExitException('Failed to get component revisions.',
                                        400)
 
-    self.render('revisions-info.html',
-                {'info': {
-                    'componentRevisionsList': component_revisions_list
-                }})
+    return self.render(
+        'revisions-info.html',
+        {'info': {
```
oliverchang, line 57: nit:  this indentation looks a bit weird. Can you space things out a bit more suchj that this looks like:

```python
{
   'info': {
     'componentRevisionsList...':
   },
})
```

src/appengine/server_flask.py
```diff
@@ -47,12 +57,22 @@
...
+
+  def get(self):
+    # TODO(singharshdeep): Remove routes based on this handler
+    # after flask migration.
+    raise helpers.EarlyExitException('URL not complete.', 404)
```
oliverchang, line 73: Can you explain why this is needed?

iamarshsingh: the middleware redirects on the basis of the first path before the second '/' so paths like '/corpus-backup/make-public' won't get redirected to flask but to webapp, so just to redirect these URLs to flask I am using this.
I also considered to modify the middleware, but as currently its working well and many handlers are already migrated, I didn't wanted to modify it. Also, this hack will removed in a few days.

src/appengine/handlers/commit_range.py
```diff
@@ -15,11 +15,12 @@
 
 import json
 
+from flask import request
```
oliverchang, line 18: Forgot to mention this, but there should be a newline in between third party dependencies and our own libraries.

iamarshsingh: Sure, will fix this as part of next PR.

PR #1975, opened 2020-08-16
Testing rendering fix for show testcase-detail.

iamarshsingh:
This is a probable fix for [this](https://github.com/google/oss-fuzz/issues/4326).
Couldn't test it locally, we can host on staging and check.


oliverchang:
Fixed this in https://github.com/google/clusterfuzz/pull/1977 with the correct code.

PR #1976, opened 2020-08-16
Migrated CRON handlers to flask.

iamarshsingh:
This PR migrates all cron handlers to flask. We can first work on #1974 and then continue on this.

### DIFF
--- a/src/appengine/handlers/cron/backup.py
+++ b/src/appengine/handlers/cron/backup.py
@@ -24,7 +24,6 @@ import googleapiclient
 
 from base import utils
 from config import local_config
-from flask import make_response
 from handlers import base_handler_flask
 from libs import handler_flask
 from metrics import logs
@@ -42,7 +41,7 @@ def _datastore_client():
 class Handler(base_handler_flask.Handler):
   """Handler for triggering the backup URL."""
 
-  @handler_flask.check_cron()
+  @handler_flask.cron()
   def get(self):
     """Handle a cron job."""
     backup_bucket = local_config.Config(
@@ -81,4 +80,4 @@ class Handler(base_handler_flask.Handler):
       status_code = e.resp.status
       logs.log_error(message, error=str(e))
 
-    return make_response(message, status_code, {'Content-Type': 'text/plain'})
+    return (message, status_code, {'Content-Type': 'text/plain'})
--- a/src/appengine/handlers/cron/backup.py
+++ b/src/appengine/handlers/cron/backup.py
@@ -24,8 +24,9 @@ import googleapiclient
 
 from base import utils
 from config import local_config
-from handlers import base_handler
-from libs import handler
+from flask import make_response
+from handlers import base_handler_flask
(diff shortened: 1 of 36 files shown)


src/appengine/handlers/cron/backup.py
```diff
@@ -80,6 +81,4 @@ def get(self):
       status_code = e.resp.status
       logs.log_error(message, error=str(e))
 
-    self.response.headers['Content-Type'] = 'text/plain'
-    self.response.out.write(message)
-    self.response.set_status(status_code)
+    return make_response(message, status_code, {'Content-Type': 'text/plain'})
```
oliverchang, line 84: nit: turns out the make_response call isn't actually necessary. You can just return the tuple.

src/appengine/handlers/cron/backup.py
```diff
@@ -38,17 +39,17 @@ def _datastore_client():
...
-  @handler.check_cron()
+  @handler_flask.check_cron()
   def get(self):
     """Handle a cron job."""
     backup_bucket = local_config.Config(
         local_config.PROJECT_PATH).get('backup.bucket')
     if not backup_bucket:
       logs.log('No backup bucket is set, skipping.')
-      return
+      return 'OK'
```
oliverchang, line 52: This is coming up a lot. I think we should make the decorator for cron handlers return 'OK' by default (and be overrideable by actual return values). We can rename `check_cron` to `cron` as well to make the name more general.

iamarshsingh: yes, nice approach. Will update the PR.

src/appengine/handlers/cron/backup.py
```diff
@@ -38,17 +38,17 @@ def _datastore_client():
...
-  @handler.check_cron()
+  @handler_flask.cron()
   def get(self):
     """Handle a cron job."""
     backup_bucket = local_config.Config(
         local_config.PROJECT_PATH).get('backup.bucket')
     if not backup_bucket:
       logs.log('No backup bucket is set, skipping.')
-      return
+      return 'OK'
```
oliverchang, line 51: Remove 'OK'

src/appengine/libs/handler_flask.py
```diff
@@ -93,7 +93,10 @@ def wrapper(self):
       if not self.is_cron():
         raise helpers.AccessDeniedException('You are not a cron.')
 
-      return func(self)
+      response = func(self)
+      if not response:
```
oliverchang, line 97: Check for None instead, as we could possibly have a '' value. 


```python
if response is None:
  return 'OK'

return response
```

src/appengine/handlers/cron/backup.py
iamarshsingh: Here actually I added 'OK'  because we are returning a tuple in the same function, if we don't return anything here there will be a lint check fail of inconsistent-return-statements

PR #1977, opened 2020-08-17
Fix unicode separator in testcase-details title.

oliverchang:


oliverchang:
TBR. @iamarshsingh FYI. this is the correct unicode code.

iamarshsingh:
Thanks for fixing this. I had mistakenly added the octal code instead of hexadecimal code in #1975.

PR #1980, opened 2020-08-18
Flask fixes.

oliverchang:
Add missing return value for /commit-range/load handler.

Also call make_response in the various wrappers in handler_flask to
ensure that we have a Response object.

### DIFF
--- a/src/appengine/libs/handler_flask.py
+++ b/src/appengine/libs/handler_flask.py
@@ -21,12 +21,14 @@ import json
 import re
 import requests
 
+from flask import g
+from flask import make_response
+from flask import request
+
 from base import utils
 from config import db_config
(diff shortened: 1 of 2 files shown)


src/appengine/libs/handler_flask.py
```diff
@@ -21,12 +21,14 @@
 import re
 import requests
 
+from flask import g
```
inferno-chromium, line 24: remove ?

oliverchang: This is needed for per-request caching: https://flask.palletsprojects.com/en/1.1.x/api/#flask.g

inferno-chromium: :)

PR #1983, opened 2020-08-19
Add name to syzkaller config when using hub.

mspectorgoogle:


src/python/bot/fuzzers/syzkaller/config.py
```diff
@@ -63,6 +63,7 @@ def generate(serial,
     data['hub_addr'] = syzhub_address
     data['hub_client'] = syzhub_client
     data['hub_key'] = syzhub_key
+    data['name'] = 'sk_haiku-{}'.format(serial)
```
inferno-chromium, line 66: Is this tied to sk_haiku prefix or can it be generic ? Is some service setup on -118 with sk_haiku (just thinking in terms of documentation change in case someone else wants to use clusterfuzz+syzkaller).

PR #1984, opened 2020-08-20
Complete migration to flask and webapp2 removal.

### LINKED ISSUE #1394: Migrate to flask
[webapp2](https://github.com/GoogleCloudPlatform/webapp2) is unsupported and the latest release (which was from 2016) has "Provisional support for Python 3". It's likely there'll be issues with this. 

We should just migrate to Flask or something similar which is more actively maintained.

iamarshsingh:
This PR removes all the webapp2 code and fixes naming for all handlers. This completes flask migration and fixes #1394.

Currently there is some problem with redirect_domain to domain routing, there are no errors but redirection doesn't work.


oliverchang:
I also re-implemented the redirection in a much simpler way. /gcbrun

PR #1992, opened 2020-08-27
GradientFuzz Integration [1]

RyanBCao:
# Integration Sanitycheck
1. Modified `ml_train.py` to differentiate between ML-RNN and GradientFuzz jobs. 
    - Creates a task with `gradientfuzz` as task name (to eventually invoke `gradientfuzz_train.py`).
2. Modified `ml_train_test.py` unit test to queue up both types of jobs, ensuring both tasks get added.
3. Renamed `ml_train_task.py` to `rnn_generator_task.py` and edited naming usage.


src/appengine/handlers/cron/ml_train.py
```diff
@@ -28,17 +28,22 @@ class Handler(base_handler.Handler):
...
     """Handle a GET request."""
     for job in data_types.Job.query():
+
       if not utils.string_is_true(
           job.get_environment().get('USE_CORPUS_FOR_ML')):
         continue
 
+      task_name = 'gradientfuzz' if utils.string_is_true(
+          job.get_environment().get(
+              'USE_LIBFUZZER_FOR_GRADIENTFUZZ')) else 'ml_train'
```
mbarbella-chromium, line 38: We're going to need to support the case where both are set for the same job. This will be especially true as we add support for the other ML jobs that are likely coming soon.

I'm also not sure I like the name of the environment variable. Could you rename this to USE_GRADIENTFUZZ? It'll be defined in a job definition, so that's clear enough when set. You'll probably also need to add a separate environment variable for something like USE_RNN, though the fact that that's how we refer to it throughout the codebase is a bit unfortunate in hindsight (and, if possible, we should rename that task to something other than ml_train to something more specific).

src/python/bot/fuzzers/ml/gradientfuzz/run_constants.py
```diff
@@ -0,0 +1,77 @@
+# Copyright 2019 Google LLC
```
mbarbella-chromium, line 1: Let's hold off on landing the actual code until a later change. For now just include a skeleton of a task which logs something when it runs so that we can confirm it's working. I don't want this CL to get too large, and don't want any ambiguity when we test to see if the scheduling portion is working.

src/appengine/handlers/cron/ml_train.py
RyanBCao: Gotchu -- I'll put separate tasks for any job rolling through with `USE_CORPUS_FOR_ML`. Rename incoming!

src/appengine/handlers/cron/ml_train.py
```diff
@@ -28,17 +28,22 @@ class Handler(base_handler.Handler):
   def get(self):
     """Handle a GET request."""
     for job in data_types.Job.query():
+
       if not utils.string_is_true(
           job.get_environment().get('USE_CORPUS_FOR_ML')):
         continue
 
+      task_name = 'gradientfuzz' if utils.string_is_true(
```
oliverchang, line 36: given the length of this condition, this would bve more readable as:

```python
if utils.string_is_true(...):
  task_name = 'gradientfuzz'
else:
  task_name = 'ml_train'
```

src/python/bot/fuzzers/ml/gradientfuzz/run_constants.py
oliverchang: +1

src/appengine/handlers/cron/ml_train.py
RyanBCao: Got it -- fixed :)

src/python/bot/fuzzers/ml/gradientfuzz/run_constants.py
RyanBCao: Yup -- file removed until later commit!

RyanBCao:
Renamed `ml_train_task.py` and all uses/mentions to `rnn_generator_task.py`. Changed unit tests accordingly as well.

src/appengine/handlers/cron/ml_train.py
```diff
@@ -28,17 +28,25 @@ class Handler(base_handler.Handler):
   def get(self):
     """Handle a GET request."""
     for job in data_types.Job.query():
+
       if not utils.string_is_true(
           job.get_environment().get('USE_CORPUS_FOR_ML')):
         continue
 
+      if utils.string_is_true(job.get_environment().get('USE_GRADIENTFUZZ')):
+        task_name = 'gradientfuzz'
```
mbarbella-chromium, line 37: Nit: could you rename each of these tasks to train_gradientfuzz and train_rnn_generator? That seems to clear up any ambiguity. Sorry to ask you to rename those files again.

src/appengine/handlers/cron/ml_train.py
```diff
@@ -28,17 +28,25 @@ class Handler(base_handler.Handler):
   def get(self):
     """Handle a GET request."""
     for job in data_types.Job.query():
+
       if not utils.string_is_true(
           job.get_environment().get('USE_CORPUS_FOR_ML')):
         continue
 
+      if utils.string_is_true(job.get_environment().get('USE_GRADIENTFUZZ')):
```
mbarbella-chromium, line 36: It still looks like this wouldn't work properly for the case where both variables were set. Could you replace task_name with a task_names list above, and add one task of each?

src/python/bot/tasks/commands.py
```diff
@@ -49,7 +49,7 @@
     'fuzz': fuzz_task,
     'impact': impact_task,
     'minimize': minimize_task,
-    'ml_train': ml_train_task,
+    'rnn_generator': rnn_generator_task,
```
mbarbella-chromium, line 52: Can you also add an entry for train_gradientfuzz? We shouldn't add the full file for it but I'd still like to have a simple dummy that logs that it's running so that we can confirm things work before landing the future changes.

src/appengine/handlers/cron/ml_train.py
RyanBCao: Yeah -- makes sense, given that training and generation are separate.

src/appengine/handlers/cron/ml_train.py
RyanBCao: Ah I see -- one job, but multiple tasks. Got it!

src/python/bot/tasks/commands.py
RyanBCao: Yeah -- sounds good. I'll add the dummy as well!

RyanBCao:
Sorry about that -- I tried running `python3 butler.py format` but I guess it doesn't help pylint standards? I'll manually try pylint on all the changes files.

mbarbella-chromium:
I'm going to deploy this now and kick off a manual run of the cron job. If it looks like anything goes wrong I'll file a bug.

Issue #1611, opened 2020-03-18
Add hwasan support to .options file


Dor1s:
Similarly to e.g. https://github.com/google/clusterfuzz/blob/cadf520d1329d80ac337c9282f72582b4452b470/src/python/bot/fuzzers/options.py#L147

the test to update: https://github.com/google/clusterfuzz/blob/f594d3e46fecf28720ef4258d947e135aa243e0c/src/python/tests/core/bot/fuzzers/options_test.py#L34

PR #1996, opened 2020-08-31
Support HWASan in .options file with [hwasan] section.

inferno-chromium:
Fixes https://github.com/google/clusterfuzz/issues/1611

### DIFF
--- a/src/python/bot/fuzzers/options.py
+++ b/src/python/bot/fuzzers/options.py
@@ -157,7 +157,7 @@ class FuzzerOptions(object):
     return self._get_option_section('ubsan')
 
   def get_hwasan_options(self):
-    """Return a list of UBSAN_OPTIONS overrides."""
+    """Return a list of HWSAN_OPTIONS overrides."""
     return self._get_option_section('hwasan')
 
   def get_grammar_options(self):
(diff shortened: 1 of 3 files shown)


inferno-chromium:
Leon from skia team was asking on this.

src/python/bot/fuzzers/options.py
```diff
@@ -156,6 +156,10 @@ def get_ubsan_options(self):
     """Return a list of UBSAN_OPTIONS overrides."""
     return self._get_option_section('ubsan')
 
+  def get_hwasan_options(self):
+    """Return a list of UBSAN_OPTIONS overrides."""
```
Dor1s, line 160: nit: typo in `UBSAN_OPTIONS`

inferno-chromium:
> The change LGTM, thanks! But I'm surprised that we want to re-use `[asan]` section for HWASan. I think the original request for this came in from Mitch, who had something in mind for Android.
> 
> Why don't we want to use an explicit `[hwasan]` section?

yes, thinking again, it feels much cleaner to use hwasan section rather than re-use asan. i was thinking cross-platform stuff, but i think build pipelines should be different per-platform. So, lets keep it cleaner. so fixed it now.

src/python/bot/fuzzers/options.py
inferno-chromium: Done!

Dor1s:
and PR title can be changed now (since we don't rely on `[asan]` anymore). Thanks a lot!

PR #1997, opened 2020-08-31

HOLMEScdk:
Added another environment for further using RL method to generate more corpus for libfuzzer.


HOLMEScdk:
@googlebot I fixed it.

PR #1999, opened 2020-09-01

HOLMEScdk:


oliverchang:
Hi, before we can accept this, can you share what this is, and what your plans are for contributing support for this?

HOLMEScdk:
Hi Oliver, 
sorry for an unexpected pull request. This work actually is my current internship project at Google, trying to use reinforcement learning method to learn the policy in order to generate more useful initial seed corpus for libfuzzer whose functionality is same as current RNN-ML and Ryan's work but with different approach. 
Sorry for several pull request, because I don't know why it said I need another CLA though I already got approved I think.
Abhishek and Max know this project too. 
Thanks!

HOLMEScdk:
For the integration plan, I actually finished the model part, I am now working with Ryan to figure it out how we can integrate our work smoothly, after integration, we can have more options on generating corpus using ML.

PR #2000, opened 2020-09-01

HOLMEScdk:


HOLMEScdk:
Hi,
This work actually is my current internship project at Google, trying to use reinforcement learning method to learn the policy in order to generate more useful initial seed corpus for libfuzzer whose functionality is same as current RNN-ML and Ryan's work but with different approach.

For the integration plan, I actually finished the model part, I am now working with Ryan to figure it out how we can integrate our work smoothly, after integration, we can have more options on generating corpus using ML.

Thanks!

PR #2003, opened 2020-09-02
Check for 'blackbox' field in project.yaml for high end project setup.

oliverchang:
Rather than a 'none' value in the fuzzing_engines field.


oliverchang:
CF counterpart to https://github.com/google/oss-fuzz/pull/4405. This is needed for quickjs to use our blackbox JS fuzzer.

PR #2006, opened 2020-09-02
Test two new dependencies for training reinforcement learning based models to generate corpus.

HOLMEScdk:
Add torch==1.4.0 and gym==0.17.2 as two new dependencies.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -11,6 +11,8 @@ protobuf = "==3.11.3"
 psutil = "==5.7.0"
 numpy = "==1.18.4"
 tensorflow = "==2.3.0"
+torch = "==1.4.0"
+gym = "==0.17.2"
 
 [dev-packages]
 pylint = "~=2.4"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "beabbd6e083fb18cebd027caf59ca712c6c597439033b416ee4d95e9ff38088f"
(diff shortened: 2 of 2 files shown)


Dor1s:
Thanks Dongkai! We'll discuss the feasibility of this in the team meeting (I mean ClusterFuzz team -- maybe my teammates have other ideas on how to proceed).

In the meantime, could you please give concrete numbers on how long (in minutes) it takes to train a model (with some fixed default parameters) and the reasonable number of epochs. Try to fix all the parameters to imitate a training run in the production and give us the concrete numbers. That's important because different fuzzing bots have different lifetime and long lasting tasks may require to be enabled for a certain subset of the bots or even require specially configured bots.

HOLMEScdk:
Thanks Max, 
For the runtime,  specifically, for each epoch, it relies on the number of episodes we want to execute and the length of each episode, called steps.  # of  episodes X # of steps is the total number of training data we will collect in each epoch where we need libfuzzer to provide reward. 
So if we set # of episodes = 40, # of steps = 20, it will average take 15-20min in my machine on 4 fuzz targets. Mupdf: 15 mins/epoch,  openjpeg: 25mins/epoch, libpng: 10mins/epoch, libmpeg2: 30~40 mins/epoch ( can't remember very well). 

And for generation task, it will take around 10 mins to generate 1k new files in average.

Dor1s:
@HOLMEScdk thanks a lot for the performance numbrers. I'm closing this issue since we decided to do the evaluation outside of ClusterFuzz for now.

PR #2016, opened 2020-09-08
Implement an API endpoint for checking if a crash is known.

oliverchang:
Input (JSON POST):
{
  "stacktrace": "stacktrace as string",
  "project": "project name",
  "fuzz_target": "fuzz target if applicable (optional)"
}

Output (JSON):
{
  "result": "new/duplicate",
  "state": "crash state",
  "type": "crash type",
  "security": "true/false",
}

Also modify stack_analyzer.get_crash_data to:
- remove reliance on environment variables for getting the fuzz target
  and whether or not to parse OOMs/timeouts.
- add "already_symbolized" argument to indicate that a stacktrace is
  already symbolized.

### DIFF
--- a/src/appengine/handlers/crash_query.py
+++ b/src/appengine/handlers/crash_query.py
@@ -1,0 +1,61 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Handler for the crash query api."""
+
+from flask import request
+
+from crash_analysis import crash_analyzer
+from crash_analysis.stack_parsing import stack_analyzer
+from datastore import data_handler
+from handlers import base_handler
+from libs import auth
+from libs import handler
+from libs import helpers
+
+
+class Handler(base_handler.Handler):
+  """Handler that gets the crash stats when user first lands on the page."""
+
+  @handler.post(handler.JSON, handler.JSON)
+  @handler.oauth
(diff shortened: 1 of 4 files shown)


src/appengine/handlers/crash_query.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+class Handler(base_handler.Handler):
+  """Handler that gets the crash stats when user first lands on the page."""
+
+  @handler.post(handler.JSON, handler.JSON)
+  @handler.oauth
+  def post(self):
+    """Handle a post request."""
+    if not auth.get_current_user():
```
inferno-chromium, line 34: Does handler.oauth decorator not check this ?

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -633,6 +632,11 @@ def __init__(self, symbolized=True):
     self.raw_frames = []
     self.last_frame_id = -1
 
+    if fuzz_target:
+      self.fuzz_target = fuzz_target
+    else:
+      self.fuzz_target = environment.get_value('FUZZ_TARGET')
```
inferno-chromium, line 638: environment.get_value('FUZZ_TARGET') or fuzz_target ?

inferno-chromium (approved):

src/appengine/handlers/crash_query.py
oliverchang: Nope. Was thinking to change this, but would prefer to do this in a separate PR in case things break.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
oliverchang: Done! (but changed the order since fuzz_target argument should take precedence).

src/appengine/handlers/crash_query.py
oliverchang: Looking at this again. We can't change handler.oauth to enforce this, as it's used on handlers for web stuff too (e.g. the testcase refresh handler). It was intended to add an additional way for authenticating against an endpoint, but does not make it exclusive.

inferno-chromium: Thanks for checking, yes it looks like it just adds additional headers but does not validate them.

PR #2017, opened 2020-09-10
[fuchsia] Work around bug in older builds with newer KVM

eepeep:
Newer KVM versions exposed a bug that caused Fuchsia to fail to boot.
This was previously fixed on the Fuchsia side, but requires a
ClusterFuzz-side workaround in order to be able to run builds from
before that fix rolled (e.g. when bisecting).

This commit also adds the `invtsc` feature to the QEMU CPU config to
bring it in line with Fuchsia defaults and improve performance, and
makes it slightly easier to reproduce the QEMU commands run by
ClusterFuzz.

### DIFF
--- a/src/python/platforms/fuchsia/device.py
+++ b/src/python/platforms/fuchsia/device.py
@@ -164,10 +164,18 @@ class QemuProcess(object):
     ]
     # yapf: enable
 
-    # Detecing KVM is tricky, so use an environment variable to determine
+    # Detecting KVM is tricky, so use an environment variable to determine
     # whether to turn it on or not.
     if environment.get_value('FUCHSIA_USE_KVM'):
-      qemu_args.extend(['-cpu', 'host,migratable=no'])
+      # In builds before fxrev.dev/375343, a bug prevents booting with newer
+      # versions of KVM. On some of these older builds,
+      # `kernel.x86.disable-spec-mitigations` also doesn't work as
+      # expected, so we work around this by selecting a CPU type where the
+      # speculation mitigation will not applied.
+      if environment.get_value('APP_REVISION') < 20200414210423:
+        qemu_args.extend(['-cpu', 'Opteron_G5,+invtsc'])
+      else:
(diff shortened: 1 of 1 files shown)


eepeep:
/gcbrun

+@oliverchang

src/python/platforms/fuchsia/device.py
```diff
@@ -185,7 +193,7 @@ def create(self):
     # Fuzzing jobs that SSH into the QEMU VM need access to this env var.
     environment.set_value('FUCHSIA_PKEY_PATH', qemu_vars['pkey_path'])
     logs.log('Ready to run QEMU. Command: ' + qemu_vars['qemu_path'] + ' ' +
-             str(qemu_args))
+             ' '.join('"{}"'.format(arg) for arg in qemu_args))
```
oliverchang, line 196: nit: use shlex.quote instead (https://docs.python.org/3/library/shlex.html#shlex.quote) of just putting double quotes around t

oliverchang (approved):
Not the prettiest solution, but don't see a better way without the new interface :)

Note that our bots are running a pretty old image (ubuntu 16.04). Is that new enough to be affected? Also, could this be related at all to the qemu deaths we've been seeing?

src/python/platforms/fuchsia/device.py
eepeep: Good call; thanks for bringing that library to my attention!

eepeep:
Ah, yeah, I think this only manifests on kernel 5.3+, so it shouldn't be an issue right now if that's the current configuration. I don't think it's related to the qemu deaths, for that reason and because it's an extremely-early boot failure which would look different in the logs.

oliverchang:
> Ah, yeah, I think this only manifests on kernel 5.3+, so it shouldn't be an issue right now if that's the current configuration. I don't think it's related to the qemu deaths, for that reason and because it's an extremely-early boot failure which would look different in the logs.

Ah of course, KVM is in the kernel.. in that case that's decoupled from the docker image and dependent on the host image we're using, so better to keep this for now :)

PR #2019, opened 2020-09-10
Cap RUN_TIMEOUT to avoid broken bots.

inferno-chromium:
If a client provides a large timeout, cap it to avoid
bot failures. Maximum allowed value is int(2**31 / 1000).

### DIFF
--- a/src/python/bot/startup/run.py
+++ b/src/python/bot/startup/run.py
@@ -48,6 +48,11 @@ def start_bot(bot_command):
 
   # Wait until the process terminates or until run timed out.
   run_timeout = environment.get_value('RUN_TIMEOUT')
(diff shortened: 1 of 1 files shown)


src/python/bot/startup/run.py
```diff
@@ -48,6 +48,11 @@ def start_bot(bot_command):
 
   # Wait until the process terminates or until run timed out.
   run_timeout = environment.get_value('RUN_TIMEOUT')
+  max_timeout = int(2**31 / 1000)  # https://bugs.python.org/issue20493
```
oliverchang, line 51: nit: just `2**31 // 1000`?

oliverchang (approved):

PR #2021, opened 2020-09-11
Close testcase whose fuzzers got deleted automatically.

### LINKED ISSUE #1825: Gracefully handle tasks when build isn't found.
For cases when a bucket has a lifecycle policy, if a target is removed/renamed we will repeatedly try tasks on the same testcase even though we'll never get a build.

inferno-chromium:
GCS Bucket can have lifecycle policy. For deleted fuzzers,
this would mean empty revision list. So, rather than retry,
just close the testcase. It should never happen in regular
case and instead result in exception.

Fixes #1825


mbarbella-chromium (approved):

oliverchang (commented):
Thanks!

PR #2022, opened 2020-09-11
Gradientfuzz Integration [3]

RyanBCao:
------------------------------------ Integration of GradientFuzz Generator Script ------------------------------------
- Added `generator.py`, which downloads pre-trained models from GCS and calls three separate scripts from `bot/fuzzers/ml/gradientfuzz`.
    - `libfuzzer_to_numpy.py` -- Called to convert raw inputs into Numpy arrays without generating coverage information (new command-line flag added to allow for this).
    - `gradient_gen_critical_locs.py` -- Generates a suggested critical locations for each input.
    - `gen_mutations.py` -- Generates mutations based on suggested critical locations.

- Modified a few other CF files to pick GradientFuzz as a mutation strategy option and use it a fixed proportion of the time.


RyanBCao:
Sadly I didn't have time to write unit tests for the above :( But the integration scripts should be fairly self-explanatory and hopefully not a massive amount of work to fix up.

oliverchang:
Closing as obsolete. We may revisit this in the future.

PR #2026, opened 2020-09-15
Better heuristic for getting main repo in get_components_list.

oliverchang:
If no exact match is found, use a repo with the project name in it as a
substring.

### DIFF
--- a/src/python/build_management/revisions.py
+++ b/src/python/build_management/revisions.py
@@ -316,12 +316,17 @@ def get_components_list(component_revisions_dict, job_type):
     # No project name found in job environment, return list as-is.
     return components
 
-  project_src = '/src/%s' % project_name
-  for component in components:
+  project_src = '/src/' + project_name
+  for component in components.copy():
     if component == project_src:
       components.remove(component)
       components.insert(0, component)
       break
+
+    if project_name in os.path.basename(component):
+      components.remove(component)
+      components.insert(0, component)
+      # Keep trying in case an exact match is found later.
 
   return components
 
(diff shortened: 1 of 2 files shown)


oliverchang:
This is needed for more accurate bisection etc.

src/python/build_management/revisions.py
```diff
@@ -316,13 +316,18 @@ def get_components_list(component_revisions_dict, job_type):
...
-  project_src = '/src/%s' % project_name
-  for component in components:
+  project_src = '/src/' + project_name
+  for component in components.copy():
     if component == project_src:
       components.remove(component)
       components.insert(0, component)
       break
 
+    if project_name in os.path.basename(component):
```
inferno-chromium, line 326: Should this be startswith to be avoid unexpected matches ?

src/python/tests/core/build_management/revisions_test.py
```diff
@@ -453,6 +453,37 @@ def test_get_components_list(self):
...
+        name='libfuzzer_asan_php',
+        environment_string=('PROJECT_NAME = php\n'
+                            'HELP_URL = help_url\n')).put()
+    revisions_dict = {
+        '/src/afl': {
+            'type': 'git',
+            'url': 'https://github.com/google/AFL.git',
+            'rev': '82b5e359463238d790cadbe2dd494d6a4928bff3'
+        },
+        '/src/php-src': {
```
inferno-chromium, line 468: Do we want to have a test that has both

/src/php
and /src/php-src ?

inferno-chromium (approved):

src/python/build_management/revisions.py
oliverchang: Want to catch cases like NAME vs libNAME. Precedence goes to exact matches anyway, so this would be only in cases where that fails.

src/python/tests/core/build_management/revisions_test.py
oliverchang: Done!

PR #2027, opened 2020-09-15
skip battery/temprature check if device is android automotive

stplaydog:


src/python/platforms/android/adb.py
```diff
@@ -320,6 +320,10 @@ def is_gce():
   android_serial = environment.get_value('ANDROID_SERIAL')
   return android_serial.startswith('127.0.0.1:')
 
+def is_automotive():
+  """Returns if we are running in Android Automotive OS, currently only for Osprey."""
+  model = get_property('ro.product.model')
```
inferno-chromium, line 325: Define get_property('ro.product.model') as a helper function in settings.py (get_product_model function)

src/python/platforms/android/adb.py
```diff
@@ -320,6 +320,10 @@ def is_gce():
...
+  return model.startswith('Osprey')
```
inferno-chromium, line 326: Why startswith, why not == ?, do we have model examples of like Osprey_ ?

inferno-chromium (changes requested):

src/python/platforms/android/adb.py
stplaydog: I agree to use ==, because there is not variations for now

src/python/platforms/android/adb.py
stplaydog: Thanks for pointing out, I think I can move is_automotive  to setting.py and use helper functions in settings.py directly

src/python/platforms/android/settings.py
```diff
@@ -174,6 +174,12 @@ def is_google_device():
   return product_brand in ('google', 'generic')
 
 
+def is_automotive():
+  """Returns if we are running in Android Automotive OS, currently only for Osprey."""
+  product_name = get_product_name()
+  return product_name == 'osprey_hwasan'
```
inferno-chromium, line 180: i don't think hwasan is the right match, as we can have devices without hwasan, so either use model (define a function) or use startswith.

inferno-chromium (changes requested):

src/python/platforms/android/settings.py
stplaydog: sure, let's use model.

inferno-chromium (approved):

PR #2002, opened 2020-09-01
Gradientfuzz Integration [2]

RyanBCao:
- Added GradientFuzz training task script.
  - Functionality: Generates input/output pairs from raw inputs, trains model, and uploads weights to GCS.
- Added GradientFuzz training task unit tests.
  - Tests mock function calls, input/output pair generation, and training on a small corpus using a small, custom-built binary.
- Added auxiliary test files for unit tests.
- Refactored functions from RNN generator training script for general ML-fuzz training scripts.

PR #2031, opened 2020-09-17
Reland "Gradientfuzz Integration [2] (#2002)"

mbarbella-chromium:
This is essentially just a reland but includes updates to the sample Windows configuration to handle the tensorflow dependency. It looks like it would be difficult to break the dependency completely, and we have other tensorflow imports that I'm worried about in some of the older ML code that it looks like we managed to avoid accidentally.


mbarbella-chromium:
TBR for a reland. Will address any comments on the startup script changes if needed.

Issue #2028, opened 2020-09-15
Allow job types to specify configurable task lease timeout instead of hardcoded max 6 hours.


inferno-chromium:

inferno-chromium:
Maybe with preemptible bots, not a great idea.

PR #2033, opened 2020-09-17
Make task lease timeout configurable.

inferno-chromium:
This is needed for Syzkaller usecase where we need to run
long jobs.

### DIFF
--- a/src/python/base/tasks.py
+++ b/src/python/base/tasks.py
@@ -51,7 +51,7 @@ MAX_TASKS_LIMIT = 100000
 TASK_COMPLETION_BUFFER = 90 * 60
 TASK_CREATION_WAIT_INTERVAL = 2 * 60
 TASK_EXCEPTION_WAIT_INTERVAL = 5 * 60
-TASK_LEASE_SECONDS = 6 * 60 * 60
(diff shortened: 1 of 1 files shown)


mbarbella-chromium (approved):

src/python/base/tasks.py
```diff
@@ -376,12 +376,16 @@ def add_task(command, argument, job_type, queue=None, wait_time=None):
       [task.to_pubsub_message()])
 
 
+def get_task_lease_timeout():
+  """Return the task lease timeout."""
+  return environment.get_value('TASK_LEASE_SECONDS', TASK_LEASE_SECONDS)
```
oliverchang, line 381: should this be restricted to fuzzing tasks only? seems like there is an inconsistency here since because whatever is in TASK_LEASE_SECONDS_BY_COMMAND (i.e. regression, corpus_pruning) may be unintentionally overridden by this.

inferno-chromium: Do you have a fix in mind ? I ideally wanted this configurable in job definition [https://github.com/google/clusterfuzz/issues/2028] but this code gets triggered much before.

oliverchang: actually never mind, i misread the code before. TASK_LEASE_SECONDS_BY_COMMAND still has precedence, so I don't have any more concerns here :)

PR #2034, opened 2020-09-17
Implement the engine interface for blackbox fuzzers.

mbarbella-chromium:
No rush on this. PTAL when you have time.

This change only implements the interface. In a follow-up change, I'll update fuzz_task to use this for certain fuzzers.


oliverchang (commented):
nice! some initial comments.

src/python/bot/fuzzers/blackbox/engine.py
```diff
@@ -0,0 +1,171 @@
...
+    test cases and pass them to a binary that is managed by the infrastructure.
+
+    Args:
+      corpus_dir: The main corpus directory.
+      target_path: Path to the target.
+      build_dir: Path to the build directory.
+    Returns:
+      A FuzzOptions object.
+    """
+    app_path = environment.get_value('APP_PATH')
```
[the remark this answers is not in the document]
inferno-chromium: +1

src/python/tests/core/bot/fuzzers/blackbox/test_data/sample_fuzz_output/fuzz-noncrash-testcase
```diff
@@ -0,0 +1 @@
+2
```
inferno-chromium, line 1: Where is this used ?

src/python/tests/core/bot/fuzzers/blackbox/test_data/sample_fuzz_output/output-noncrash-testcase
```diff
@@ -0,0 +1 @@
+Boring program output.
```
inferno-chromium, line 1: Where is this used ?

src/python/tests/core/bot/fuzzers/blackbox/test_data/sample_fuzz_output/output-real-crash
```diff
@@ -0,0 +1,4 @@
+#
```
inferno-chromium, line 1: Where is this used ?

src/python/tests/core/bot/fuzzers/blackbox/blackbox_engine_test.py
```diff
@@ -0,0 +1,77 @@
...
+    result = blackbox_engine.prepare('/input/corpus', 'unused', '/build')
+    self.assertEqual(result.arguments, [
+        '--app_path=/build/test_binary', '--app_args=-a -b',
+        '--input_dir=/input/corpus'
+    ])
+
+  def test_fuzz(self):
+    blackbox_engine = engine.BlackboxEngine()
+    options = blackbox_engine.prepare('/input/corpus', 'unused', '/build')
+    result = blackbox_engine.fuzz('/build/test_binary', options,
```
inferno-chromium, line 62: How are we running the test_binary ? Does this need to be checked in ?

src/python/tests/core/bot/fuzzers/blackbox/blackbox_engine_test.py
```diff
@@ -0,0 +1,77 @@
...
+                                  FUZZ_OUTPUT_DIR, 10)
+    self.assertEqual(len(result.crashes), 1)
+    self.assertEqual(result.crashes[0].input_path,
+                     os.path.join(FUZZ_OUTPUT_DIR, 'fuzz-real-crash'))
+
+  def test_reproduce(self):
+    blackbox_engine = engine.BlackboxEngine()
+    blackbox_engine.reproduce('/build/test_binary', '/testcase', None, 10)
```
inferno-chromium, line 70: add some args, instead of None ?

src/python/bot/testcase_manager.py
```diff
@@ -903,9 +904,14 @@ def get_command_line_for_application(file_to_run='',
   if ' ' in app_path:
     app_path = '"%s"' % app_path
 
-  # Prepend command with interpreter if it is a script.
   interpreter = shell.get_interpreter(app_name)
-  if interpreter:
+  if get_arguments_only:
+    # If we are only returning the arguments, leave do not return the
```
inferno-chromium, line 909: fix language "leave do not return"

src/python/bot/fuzzers/blackbox/engine.py
```diff
@@ -0,0 +1,171 @@
...
+      arguments: Additional arguments needed for reproduction.
+      max_time: Maximum allowed time for the reproduction.
+    Returns:
+      A ReproduceResult.
+    """
+    os.chmod(target_path, 0o775)
+
+    app_path = environment.get_value('APP_PATH')
+    app_args = testcase_manager.get_command_line_for_application(
+        get_arguments_only=True).strip()
```
inferno-chromium, line 142: Why strip ?

src/python/bot/fuzzers/blackbox/engine.py
```diff
@@ -0,0 +1,171 @@
...
+      # Filter obviously non-crashing test cases. Crashes still follow the
+      # normal flow in fuzz task to ensure that the state should not be ignored
+      # for other reasons, but we don't want to log every test case for the
+      # fuzzers that don't do their own crash processing.
+      state = stack_analyzer.get_crash_data(output)
+      if not state.crash_type:
+        continue
+
+      crash = engine.Crash(
+          os.path.join(reproducers_dir, testcase_path), output, [],
```
inferno-chromium, line 118: Why [] and not args ?

move os.path.join(reproducers_dir, testcase_path) into a local

src/python/bot/fuzzers/blackbox/engine.py
```diff
@@ -0,0 +1,171 @@
...
+    for testcase_path in os.listdir(reproducers_dir):
+      if not testcase_path.startswith(TESTCASE_PREFIX):
+        continue
+
+      output_path = OUTPUT_PREFIX + testcase_path[len(TESTCASE_PREFIX):]
+      absolute_output_path = os.path.join(reproducers_dir, output_path)
+
+      # If no output was written for a test case, skip it.
+      if not os.path.exists(absolute_output_path):
+        continue
```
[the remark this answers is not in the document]
inferno-chromium: +1, how is this backwards compatible, or is it not possible to do ?

inferno-chromium (changes requested):

src/python/bot/fuzzers/blackbox/engine.py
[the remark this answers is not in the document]
mbarbella-chromium: At a high level it would just be a matter of moving some of the logic we have for test case processing and running to the fuzz function. That would also let us share the post-processing logic in a cleaner way, but I don't think it should be a blocker for the initial version.

If we were to change the existing blackbox pipeline to write the output to files in the same way we're doing here and port it over (have some environment variable set which determines whether we run in this new way or have CF process everything) it should be fairly easy to share the logic between both. I'd like to start with this for the case where the fuzzer processes test cases on its own for now, but I don't see any inherent problems with extending this to work with the old pipeline as well.

src/python/bot/fuzzers/blackbox/engine.py
mbarbella-chromium: They're currently unused in reproduce, but they could be useful in the future. I'll make the change here. Also moved that expression to a local.

src/python/bot/fuzzers/blackbox/engine.py
mbarbella-chromium: Because of how get_command_line_for_application is implemented we're left with extra whitespace. It's not usually a problem since we won't add extra whitespace after the testcase path, but if we don't we might be left with an intermediate string like "-a -b ". It didn't seem like a good idea to pass the trailing whitespace to the fuzzer scripts, but honestly it probably wouldn't hurt to include it.

src/python/bot/testcase_manager.py
mbarbella-chromium: Done.

src/python/tests/core/bot/fuzzers/blackbox/blackbox_engine_test.py
```diff
@@ -0,0 +1,77 @@
...
+from bot.fuzzers.blackbox import engine
+from system import environment
+from tests.test_libs import helpers as test_helpers
+
+TEST_PATH = os.path.abspath(os.path.dirname(__file__))
+DATA_DIR = os.path.join(TEST_PATH, 'test_data')
+FUZZ_OUTPUT_DIR = os.path.join(DATA_DIR, 'sample_fuzz_output')
+
+
+class Any:
```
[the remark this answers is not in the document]
mbarbella-chromium: Done

src/python/tests/core/bot/fuzzers/blackbox/blackbox_engine_test.py
mbarbella-chromium: Args are currently unused in reproduce. This is mostly because it expects to run with a different command format and otherwise has all of the information it needs to build up the command line. I'll pass an empty list here though, just so that it's closer to what will actually be provided.

src/python/tests/core/bot/fuzzers/blackbox/blackbox_engine_test.py
mbarbella-chromium: We're not actually running anything in this test since run_and_wait is mocked. We're just ensuring that it would be executed by checking the values passed to that function.

src/python/tests/core/bot/fuzzers/blackbox/test_data/sample_fuzz_output/fuzz-noncrash-testcase
mbarbella-chromium: All of these are used by test_fuzz. It's expected to find that real-crash is an interesting crash because of its output, and to ignore fuzz-noncrash-testcase because it has no crash stack in its output.

src/python/tests/core/bot/fuzzers/blackbox/test_data/sample_fuzz_output/output-noncrash-testcase
mbarbella-chromium: See above.

src/python/tests/core/bot/fuzzers/blackbox/test_data/sample_fuzz_output/output-real-crash
mbarbella-chromium: See above.

src/python/bot/fuzzers/blackbox/engine.py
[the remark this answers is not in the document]
oliverchang: I'm more envisioning just having every single blackbox fuzzer go through this pipeline, and completely deleting the old one. We shouldn't need to maintain 3 separate fuzzing pipelines.

Is there any problem with CF just always verifying/processing testcase in the output directory? The blackbox fuzzer can always do its own processing as well to trim down on results etc. That way it will be perfectly backwards compatible. As is, this PR expects the fuzzer to always process testcases since it expects this output file. We should relax this requirement.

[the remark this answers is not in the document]
inferno-chromium: +1 to what @oliverchang said.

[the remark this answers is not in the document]
mbarbella-chromium: I'm having trouble finding a clean way to support this in this change without either duplicating or, more likely, re-implementing a lot of the existing test case processing logic, but I agree that we should do this. Here's my current plan, since this CL has been quiet for a bit too long:

1. I'm going to include the usage in fuzz task in this change to pull all of this together. With that, this should actually be usable after this PR lands. I'll include a few TODOs along with that which I think will make the next steps for transitioning a bit clearer.
2. If an environment variable is set, we'll use this new pipeline for blackbox fuzzers which are capable of processing test cases on their own.
3. Once this seems stable, I'll convert the entire blackbox fuzzing pipeline to use the engine interface. This will avoid some of the code duplication issues we've been running into, and resolve the backwards compatibility issues noted here.
4. We should be able to clean up some of the other code in follow up changes, such as the use of different crash types between the two interface types, and hopefully get rid of some of the branching logic for handling both cases (though I'm not totally sure the latter will be easy to resolve).

WDYT?

src/python/tests/core/bot/fuzzers/blackbox/blackbox_engine_test.py
```diff
@@ -0,0 +1,73 @@
...
+    options = blackbox_engine.prepare('/input/corpus', 'unused', '/build')
+    result = blackbox_engine.fuzz('/build/test_binary', options,
+                                  FUZZ_OUTPUT_DIR, 10)
+    self.assertEqual(len(result.crashes), 1)
+    self.assertEqual(result.crashes[0].input_path,
+                     os.path.join(FUZZ_OUTPUT_DIR, 'fuzz-real-crash'))
+
+  def test_reproduce(self):
+    blackbox_engine = engine.BlackboxEngine()
+    blackbox_engine.reproduce('/build/test_binary', '/testcase', [], 10)
```
oliverchang, line 66: same here

oliverchang (approved):

src/python/tests/core/bot/fuzzers/blackbox/blackbox_engine_test.py
mbarbella-chromium: Good catch.

src/python/tests/core/bot/fuzzers/blackbox/blackbox_engine_test.py
mbarbella-chromium: Done.

mbarbella-chromium:
I'm going to land this change as-is after all to keep what's required for the usage in fuzz task self-contained. It should be up for review soon.

PR #2046, opened 2020-09-24
Adding support to sync corpus.db for syzkaller

hzawawy:
The corpus.db is downloaded from the cloud for each bot. When fuzzing is
concluded, the resulting corpus.db is uploaded and name after the bot.
This each bot will have its own corpus and will continue fuzzing from
where it stopped last. The exchange of corpus between individual bots
is handled by syzhub.

### DIFF
--- a/src/python/bot/fuzzers/syzkaller/engine.py
+++ b/src/python/bot/fuzzers/syzkaller/engine.py
@@ -13,17 +13,22 @@
 # limitations under the License.
 """Fuzzing engine interface."""
 
+import os
+import shutil
+
 from bot.fuzzers import engine
 from bot.fuzzers import engine_common
 from bot.fuzzers import utils as fuzzer_utils
 from bot.fuzzers.syzkaller import constants
 from bot.fuzzers.syzkaller import runner
 from metrics import profiler
+from platforms.android import settings
 from system import environment
-import os
+from system import shell
 
 BIN_FOLDER_PATH = 'bin'
 REPRO_TIME = 70
+CORPUS_FILENAME = 'corpus.db'
 
 
 class SyzkallerError(Exception):
@@ -93,6 +98,35 @@ class SyzkallerEngine(engine.Engine):
     engine_common.recreate_directory(new_corpus_directory)
     return new_corpus_directory
 
+  def save_corpus(self, source, destination):
+    """Saves syzkaller to folder so it is backed up to the cloud.
+
+    Args:
+      source: Folder where syzkaller corpus is.
+      destination: Folder where the corpus is synced with the cloud.
+    """
+    source_file = os.path.join(source, CORPUS_FILENAME)
+    shell.create_directory(destination)
+    target_file = os.path.join(destination,
+                               settings.get_device_codename() + '.db')
+    if os.path.isfile(source_file):
+      shutil.copy(source_file, target_file)
+
+  def init_corpus(self, source, destination):
+    """Uses corpus from the ckoud to initialize syzkaller corpus.
+
+    Args:
+      source: Folder where the corpus is downloaded from the cloud.
+      destination: Folder where syzkaller will be looking for corpus.
+    """
+    source_file = os.path.join(source, settings.get_device_codename() + '.db')
+    shell.create_directory(destination)
+    destination_file = os.path.join(destination, CORPUS_FILENAME)
+    if os.path.isfile(source_file) and (
+        not os.path.exists(destination_file) or
+        (os.path.getsize(source_file) > os.path.getsize(destination_file))):
+      shutil.copy(source_file, destination_file)
+
   def fuzz(self, target_path, options, unused_reproducers_dir=None, max_time=0):
     """Run a fuzz session.
 
@@ -115,7 +149,10 @@ class SyzkallerEngine(engine.Engine):
     args = options.arguments
     args += ['--coverfile', runner.get_cover_file_path()]
 
-    return syzkaller_runner.fuzz(max_time, additional_args=args)
+    self.init_corpus(options.corpus_dir, runner.get_work_dir())
+    fuzz_result = syzkaller_runner.fuzz(max_time, additional_args=args)
+    self.save_corpus(runner.get_work_dir(), options.corpus_dir)
+    return fuzz_result
 
   def reproduce(self, target_path, input_path, arguments, max_time):  # pylint: disable=unused-argument
     """Reproduce a crash given an input.
--- a/src/python/bot/fuzzers/syzkaller/engine.py
+++ b/src/python/bot/fuzzers/syzkaller/engine.py
@@ -113,7 +113,7 @@ class SyzkallerEngine(engine.Engine):
       shutil.copy(source_file, target_file)
 
   def init_corpus(self, source, destination):
-    """Uses corpus from the ckoud to initialize syzkaller corpus.
+    """Uses corpus from the cloud to initialize syzkaller corpus.
 
     Args:
       source: Folder where the corpus is downloaded from the cloud.


src/python/bot/fuzzers/syzkaller/runner.py
```diff
@@ -74,6 +79,20 @@ def get_runner(fuzzer_path):
   return AndroidSyzkallerRunner(fuzzer_path)
 
 
+def download_corpus():
```
inferno-chromium, line 82: Don't need this, see comment below.

src/python/bot/fuzzers/syzkaller/runner.py
```diff
@@ -74,6 +79,20 @@ def get_runner(fuzzer_path):
...
+  corpus_gcs_path = os.path.join(
+      'gs://' + CORPUS_BUCKET, GCS_PATH, DEVICE_SERIAL + '.db')
+  target_path = os.path.join(get_work_dir(), CORPUS_FILENAME)
+  return gsutil.GSUtilRunner().download_file(corpus_gcs_path, target_path)
+
+
+def upload_corpus():
```
inferno-chromium, line 89: Don't need this, see comment below.

src/python/bot/fuzzers/syzkaller/runner.py
```diff
@@ -23,11 +23,16 @@
...
+from google_cloud_utils import gsutil
 from metrics import logs
 from system import environment
 from system import new_process
 
 REPRODUCE_REGEX = re.compile(r'reproduced (\d+) crashes')
+CORPUS_BUCKET = environment.get_value('CORPUS_BUCKET')
+GCS_PATH = 'corpus'
+CORPUS_FILENAME = 'corpus.db'
+DEVICE_SERIAL = environment.get_value('ANDROID_SERIAL')
```
inferno-chromium, line 35: Not recommended to set environment vars from global vars, since it might not be set that time.

src/python/bot/fuzzers/syzkaller/runner.py
```diff
@@ -23,11 +23,16 @@
...
 from bot.fuzzers.syzkaller import config
+from google_cloud_utils import gsutil
 from metrics import logs
 from system import environment
 from system import new_process
 
 REPRODUCE_REGEX = re.compile(r'reproduced (\d+) crashes')
+CORPUS_BUCKET = environment.get_value('CORPUS_BUCKET')
+GCS_PATH = 'corpus'
+CORPUS_FILENAME = 'corpus.db'
```
inferno-chromium, line 34: I think Max spector said to include device family in corpus filename so that when we have multiple bots, we don't override corpus across familieis.

src/python/bot/fuzzers/syzkaller/runner.py
```diff
@@ -23,11 +23,16 @@
 from bot.fuzzers import engine
 from bot.fuzzers import utils as fuzzer_utils
 from bot.fuzzers.syzkaller import config
+from google_cloud_utils import gsutil
 from metrics import logs
 from system import environment
 from system import new_process
 
 REPRODUCE_REGEX = re.compile(r'reproduced (\d+) crashes')
+CORPUS_BUCKET = environment.get_value('CORPUS_BUCKET')
```
inferno-chromium, line 32: Don't need this var, see comment below.

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1581,7 +1581,9 @@ def do_engine_fuzzing(self, engine_impl):
     # Synchronize corpus files with GCS
     sync_corpus_directory = builtin.get_corpus_directory(
         self.data_directory, self.fuzz_target.project_qualified_name())
-    self.sync_corpus(sync_corpus_directory)
+    # For syzkaller, it is one file corpus.db and is being synced in its runner.
+    if environment.platform() != 'ANDROID_KERNEL':
```
inferno-chromium, line 1585: Don't need this, see comment above.

src/python/bot/fuzzers/syzkaller/runner.py
```diff
@@ -148,12 +167,18 @@ def _filter_log(content):
         result += strip_regex.sub('', line) + '\n'
       return result
 
+    # Download corpus ahead of fuzzing
+    download_corpus()
```
inferno-chromium, line 171: We don't need to do this, as ClusterFuzz automatically synces corpus before fuzzing and synces it back automatically after fuzzing. All you need is to put things in 
options.corpus_dir
https://github.com/google/clusterfuzz/blob/master/src/python/bot/fuzzers/honggfuzz/engine.py#L150
and everything happens automatically.

Now you still need to put the corpus.db in the right directory and tell syzkaller to load corpus.db. Can you add code that use this corpus.db file. Also, another thing Max wanted to not sync corpus db if it is smaller than previous one, so you should save file size at start of function and then only sync back if file increased in size. we can even ignore this or put in next cl.

inferno-chromium (changes requested):

src/python/bot/fuzzers/syzkaller/runner.py
hzawawy: Not sure i understood your comment: Is the suggestion to:
1) before fuzzing, copy corpus.db from options.corpus_dir (e.g. ~/clusterfuzz/872Y00ELP/clusterfuzz/bot/inputs/data-bundles/syzkaller) to the workdir where syzkaller will be using corpus.db i.e. ~/clusterfuzz/876X00GUE/bot/clusterfuzz/bot/inputs/fuzzer-testcases-disk/syzkaller
2) after fuzzing, copy back from workdir to options.corpus_dir?

Regarding copying using family (e.g. blueline) for device running blueline and then copying the file only if it is larger, that sounds like we will lose some corpus since even if it is smaller, it could contain different corpus. Shouldnt this be done by syzhub?

inferno-chromium: Yes you are right.
1. before fuzzing, check if corpus.db exists in options.corpus_dir and copy to work dir.
2. after fuzzing, check if size of corpus.db in work dir is bigger than one in options.corpus_dir (if exists) and if yes, copy it to options.corpus_dir. 
rest of gcs sync back and forth is done automatically by clusterfuzz from options.corpus_dir.

Regarding your second question, what can happen is if there are two hosts running two different family of devices for syzkaller fuzzing, then their bots can have different corpus.db files in gcs. since name is the same, they will keep overriding each other's one. regarding syzhub, i dont know you are setting it per host or across multiple hosts. Also, this can be fixed later once you have more device families, for now, we can do the more straightforward cl.

src/python/bot/fuzzers/syzkaller/runner.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/runner.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/runner.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/runner.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/runner.py
hzawawy: Done.

src/python/bot/tasks/fuzz_task.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/runner.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -93,6 +99,27 @@ def _create_temp_corpus_dir(self, name):
...
     return new_corpus_directory
 
+  def get_phone_family(self):
+    fingerprint = settings.get_build_fingerprint()
+    matcher = re.search(FINGERPRINT_REGEX, fingerprint)
+    return matcher.group(1)
+
+  def save_corpus(self, source, destination):
+    source_file = os.path.join(source, CORPUS_FILENAME)
+    if not os.path.exists(destination):
```
inferno-chromium, line 109: for these two lines, we already have a helper in
from system import shell
shell.create_directory(destination)
it auto does the directory creation check. same for init_corpus.

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -93,6 +99,27 @@ def _create_temp_corpus_dir(self, name):
...
+      os.mkdir(destination)
+    target_file = os.path.join(destination, self.get_phone_family() + '.db')
+    if os.path.isfile(source_file):
+      shutil.copy(source_file, target_file)
+
+  def init_corpus(self, source, destination):
+    source_file = os.path.join(source, self.get_phone_family() + '.db')
+    if not os.path.exists(destination):
```
inferno-chromium, line 117: use shell.create_directory(destination) for these two lines.

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -93,6 +99,27 @@ def _create_temp_corpus_dir(self, name):
...
+  def save_corpus(self, source, destination):
+    source_file = os.path.join(source, CORPUS_FILENAME)
+    if not os.path.exists(destination):
+      os.mkdir(destination)
+    target_file = os.path.join(destination, self.get_phone_family() + '.db')
+    if os.path.isfile(source_file):
+      shutil.copy(source_file, target_file)
+
+  def init_corpus(self, source, destination):
+    source_file = os.path.join(source, self.get_phone_family() + '.db')
```
inferno-chromium, line 116: function description.

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -93,6 +99,27 @@ def _create_temp_corpus_dir(self, name):
...
+    if not os.path.exists(destination):
+      os.mkdir(destination)
+    destination_file = os.path.join(destination, CORPUS_FILENAME)
+    if os.path.isfile(source_file):
```
inferno-chromium, line 120: we need to add
if os.path.isfile(source_file) and (not os.path.exists(destination_file) or os.path.getsize(source_file) > os.path.getsize(destination_file)):

this ensures that we only update corpus if it is updated.

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -113,9 +140,19 @@ def fuzz(self, target_path, options, unused_reproducers_dir=None, max_time=0):
     self._create_temp_corpus_dir('new')
 
     args = options.arguments
-    args += ['--coverfile', runner.get_cover_file_path()]
+    args += [
+        '--coverfile',
+        runner.get_cover_file_path(), '--input', options.corpus_dir
```
inferno-chromium, line 145: i dont think we want to use
'--input', options.corpus_dir
since you are copying corpus to work dir and not using the one from corpus dir.

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -113,9 +140,19 @@ def fuzz(self, target_path, options, unused_reproducers_dir=None, max_time=0):
...
+    ]
+
+    # Init corpus ahead of fuzzing
```
inferno-chromium, line 148: These comments are not required, can be in the function description for those function.

inferno-chromium (changes requested):
almost there, just some minor changes needed.

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -93,6 +99,27 @@ def _create_temp_corpus_dir(self, name):
     engine_common.recreate_directory(new_corpus_directory)
     return new_corpus_directory
 
+  def get_phone_family(self):
```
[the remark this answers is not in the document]
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -13,17 +13,23 @@
...
 from bot.fuzzers.syzkaller import runner
 from metrics import profiler
+from platforms.android import settings
 from system import environment
-import os
 
 BIN_FOLDER_PATH = 'bin'
 REPRO_TIME = 70
+CORPUS_FILENAME = 'corpus.db'
+FINGERPRINT_REGEX = r'.\/([a-z]+)_*'
```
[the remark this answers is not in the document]
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -93,6 +99,27 @@ def _create_temp_corpus_dir(self, name):
     engine_common.recreate_directory(new_corpus_directory)
     return new_corpus_directory
 
+  def get_phone_family(self):
+    fingerprint = settings.get_build_fingerprint()
+    matcher = re.search(FINGERPRINT_REGEX, fingerprint)
+    return matcher.group(1)
+
+  def save_corpus(self, source, destination):
+    source_file = os.path.join(source, CORPUS_FILENAME)
```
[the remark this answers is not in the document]
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/engine.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/engine.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/engine.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/engine.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/engine.py
hzawawy: Done.

src/python/bot/fuzzers/syzkaller/engine.py
hzawawy: Done.

inferno-chromium (approved):

inferno-chromium:
If you can run a bot locally to make sure this works, that would be great, otherwise, i will commit this on monday.

hzawawy:
> If you can run a bot locally to make sure this works, that would be great, otherwise, i will commit this on monday.

I am having issues with the test device. I will test on Monday and confirm everything is ok. Let us wait until I can see the end to end scenario with the latest code version.

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -93,6 +98,35 @@ def _create_temp_corpus_dir(self, name):
...
+    """
+    source_file = os.path.join(source, CORPUS_FILENAME)
+    shell.create_directory(destination)
+    target_file = os.path.join(destination,
+                               settings.get_device_codename() + '.db')
+    if os.path.isfile(source_file):
+      shutil.copy(source_file, target_file)
+
+  def init_corpus(self, source, destination):
+    """Uses corpus from the ckoud to initialize syzkaller corpus.
```
oliverchang, line 116: nit: s/ckoud/cloud/

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -93,6 +98,35 @@ def _create_temp_corpus_dir(self, name):
...
+      destination: Folder where the corpus is synced with the cloud.
+    """
+    source_file = os.path.join(source, CORPUS_FILENAME)
+    shell.create_directory(destination)
+    target_file = os.path.join(destination,
+                               settings.get_device_codename() + '.db')
+    if os.path.isfile(source_file):
+      shutil.copy(source_file, target_file)
+
+  def init_corpus(self, source, destination):
```
oliverchang, line 115: nit: s/source/source_dir/ and s/destination/destionation_dir/ to be clearer. Same for save_corpus.

src/python/bot/fuzzers/syzkaller/engine.py
```diff
@@ -93,6 +98,35 @@ def _create_temp_corpus_dir(self, name):
...
+    Args:
+      source: Folder where the corpus is downloaded from the cloud.
+      destination: Folder where syzkaller will be looking for corpus.
+    """
+    source_file = os.path.join(source, settings.get_device_codename() + '.db')
+    shell.create_directory(destination)
+    destination_file = os.path.join(destination, CORPUS_FILENAME)
+    if os.path.isfile(source_file) and (
+        not os.path.exists(destination_file) or
+        (os.path.getsize(source_file) > os.path.getsize(destination_file))):
```
oliverchang, line 127: Maybe just use "!=" rather than > ? Does the corpus ever get smaller (e.g. minimized) during fuzzing ?

oliverchang (commented):
sorry for the late review! just some minor comments

src/python/bot/fuzzers/syzkaller/engine.py
inferno-chromium: Yes there is a smaller corpus case when device is in bad state, like console cable not working. Max spector gave that.

Issue #2043, opened 2020-09-23
Pub/Sub feed for bug updates


ravi50041:
Hi ,

Is there any option to send emails for each new crash to configured email-address ?

inferno-chromium:
This could be a crazy simple feature to add, you want to give it a shot and propose PR

Basically needs to be call like this
https://github.com/google/clusterfuzz/blob/master/src/appengine/handlers/cron/cleanup.py#L773

in this code
https://github.com/google/clusterfuzz/blob/master/src/appengine/handlers/cron/triage.py#L327

hpathuri:
It would good to add this feature. Creating JIRA tickets from ClusterFuzz cloud to a company hosted internal JIRA needs to clear lot of security challenges.

If an email can be generated with the unique crash with a URL, companies can develop Bots to cosume these emails , pull additional info such as test case info, binary file and create internal tickets.

This is one of the issues we are trying to figure out with our eval.

As you may have noticed, oneFuzz seems to have better integration into Teams. So At least ClusterFuzz team should consider supporting this quickly.

oliverchang:
Thanks for the feedback. Would a [pub/sub](https://cloud.google.com/pubsub/docs/overview) topic work for you as well? Pub/Sub can give you structured messages that are much easier to script on in your bots.

Emails can be tricky to parse and relies on third party services (sendgrid) which may cause additional security concerns.

inferno-chromium:
@hpathuri - see @oliverchang suggestion.

Regarding ClusterFuzz, we do automatic bug filing, triage (e.g. discovery of bugs on other platforms, tell if something is top crashes, see full list here - https://github.com/google/clusterfuzz/blob/master/src/appengine/handlers/cron/cleanup.py#L133 , etc), and automatic bug verification. Filing the bug itself is just one piece of the puzzle, ClusterFuzz value addition is a lot of other triage stuff that ClusterFuzz helps developers with in a fully automated way.

If talking with internal issue tracker is an issue, we can do other publishing mechanisms like pub/sub as Oliver mentioned. Email feels insecure to do from a third-party service like sendgrid. Do you have any preferred way of getting this info in a secure way ?

hpathuri:
pub/sub will work.

I assume, we need to develop subscriber that runs on ClusterFuzz and use gsutil to export to our internal system of sort?

Where would be the subscriber need to be deployed?

oliverchang:
[...]

PR #2047, opened 2020-09-24
Notify new issues filed or closed via Pub/Sub.

oliverchang:
For #2043.

### DIFF
--- a/configs/test/project.yaml
+++ b/configs/test/project.yaml
@@ -119,8 +119,8 @@ env:
 
   # Any of the above variables can be overridden in a job definition.
 
-bisect_service:
+#bisect_service:
   #pubsub_topic: projects/project/topics/topic
 
-issue_updates:
+#issue_updates:
   #pubsub_topic: projects/project/topics/topic
--- a/configs/test/project.yaml
+++ b/configs/test/project.yaml
@@ -120,4 +120,7 @@ env:
   # Any of the above variables can be overridden in a job definition.
 
 bisect_service:
-  #pubsub_topic: /projects/project/topics/topic
+  #pubsub_topic: projects/project/topics/topic
+
+issue_updates:
+  #pubsub_topic: projects/project/topics/topic
--- a/configs/test/project.yaml
+++ b/configs/test/project.yaml
@@ -119,8 +119,10 @@ env:
 
   # Any of the above variables can be overridden in a job definition.
 
+# Pub/Sub topic for a custom bisection service.
 #bisect_service:
-  #pubsub_topic: projects/project/topics/topic
+  #pubsub_topic: projects/test-project/topics/bisection
 
+# Pub/Sub topic for receiving issue updates.
 #issue_updates:
-  #pubsub_topic: projects/project/topics/topic
+  #pubsub_topic: projects/test-project/topics/issue-updates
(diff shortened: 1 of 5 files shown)


configs/test/project.yaml
```diff
@@ -119,5 +119,8 @@ env:
 
   # Any of the above variables can be overridden in a job definition.
 
-bisect_service:
-  #pubsub_topic: /projects/project/topics/topic
+#bisect_service:
```
inferno-chromium, line 122: Probably add comments on both these pubsub topics on the purpose of these.
Also, s/project/test-project so that create_config substitution works ? or can you create directly usable names.

e.g. projects/test-project/topics/bisect-updates ?
projects/test-project/topics/issues-updates

src/appengine/handlers/cron/triage.py
```diff
@@ -307,6 +307,7 @@ def get(self):
       issue_tracker = issue_tracker_utils.get_issue_tracker_for_testcase(
           testcase)
       if not issue_tracker:
+        issue_filer.notify_issue_update(testcase, 'new')
```
inferno-chromium, line 310: This isn't fully correct as _check_and_update_similar_bug check is done below. Should we just move the issue_tracker check below this ?

src/appengine/libs/issue_management/issue_filer.py
```diff
@@ -257,6 +259,26 @@ def _get_from_metadata(testcase, name):
...
+
+  pubsub_client = pubsub.PubSubClient()
+  pubsub_client.publish(
+      topic, [
+          pubsub.Message(
+              attributes={
+                  'status': status,
+                  'testcase_id': str(testcase.key.id()),
+                  'crash_type': testcase.crash_type,
+                  'crash_state': testcase.crash_state,
```
inferno-chromium, line 276: Add crash_address and security flag as well (i think this matches our crash params list).

inferno-chromium: Maybe add testcase download url as well (i think they asked for reproducer as well).

inferno-chromium (approved):
Thank you so much!

configs/test/project.yaml
oliverchang: Done! Will add some documentation on how to use this after I verify this works.

src/appengine/handlers/cron/triage.py
oliverchang: _check_and_update_similar_bug needs issue_tracker to exist as well, so we can't move it.

src/appengine/libs/issue_management/issue_filer.py
oliverchang: Added crash_address and security flag.

Download URL can be easily constructed by the consumer of this feed by concating domain + /download?testcase_id=ID. We wouldn't be adding much value by including this? 

In any case I don't think this feed can provide all the needed details. People will inevitably need to go to the detailed report to get all the details necessary to triage and fix the bug. This can just serve as a notification system for that.

oliverchang: Friendly ping @inferno-chromium  :) Any thoughts here?

inferno-chromium: Yes no more changes needed, and fine to skip download url.

inferno-chromium (approved):

PR #2049, opened 2020-09-24
Use the new blackbox fuzzer engine implementation.

mbarbella-chromium:
PTAL. This needs to be tested a bit more carefully, but the majority of the changes are limited to the test cases where the blackbox engine is used, or to fuzzing with USE_BLACKBOX_ENGINE set to True. I expect that some follow-up fixes will be needed.

### DIFF
--- a/src/python/bot/fuzzers/blackbox/engine.py
+++ b/src/python/bot/fuzzers/blackbox/engine.py
@@ -26,6 +26,13 @@ TESTCASE_PREFIX = 'fuzz-'
 OUTPUT_PREFIX = 'output-'
 
 
+def _get_fuzzer_path(fuzzer_name, executable_path):
+  """Get the path to the fuzzer binary."""
+  fuzzer_directory = environment.get_value('FUZZERS_DIR')
+  fuzzer_directory = os.path.join(fuzzer_directory, fuzzer_name)
+  return os.path.join(fuzzer_directory, executable_path)
+
+
 def _get_arguments(app_path, app_args):
   """Get arguments shared between multiple run types."""
   return [f'--app_path={app_path}', f'--app_args={app_args}']
@@ -81,18 +88,12 @@ class BlackboxEngine(engine.Engine):
     Returns:
       A FuzzOptions object.
     """
-    app_path = environment.get_value('APP_PATH')
-    app_args = testcase_manager.get_command_line_for_application(
-        get_arguments_only=True).strip()
-
-    arguments = _get_arguments(app_path, app_args)
-    arguments.append(f'--input_dir={corpus_dir}')
-
+    executable_path = environment.get_value('FUZZER_EXECUTABLE_PATH')
+    arguments = [executable_path]
     return engine.FuzzOptions(corpus_dir, arguments, {})
 
   # TODO(mbarbella): As implemented, this will not work for untrusted workers.
-  # For this support we will also need to copy fuzzer binaries and possibly
-  # shared corpora to the workers.
+  # We would need to copy fuzzer binaries to workers.
   def fuzz(self, target_path, options, reproducers_dir, max_time):
     """Run a fuzzing session.
     Args:
@@ -104,15 +105,22 @@ class BlackboxEngine(engine.Engine):
    Returns:
       A FuzzResult object.
     """
-    # For blackbox fuzzers, |target_path| refers to the fuzzer script itself
-    # rather than a specific target in the build archive.
-    fuzzer_path = target_path
+    # For blackbox fuzzers, |target_path| supplies the fuzzer name rather than
+    # the path to a target within the build archive.
+    fuzzer_name = target_path
+    executable_path = options.arguments[0]
+    fuzzer_path = _get_fuzzer_path(fuzzer_name, executable_path)
     os.chmod(fuzzer_path, 0o775)
 
-    args = options.arguments
-    args.append(f'--output_dir={reproducers_dir}')
+    app_path = environment.get_value('APP_PATH')
+    app_args = testcase_manager.get_command_line_for_application(
+        get_arguments_only=True).strip()
+    corpus_dir = options.corpus_dir
+    command_line_args = _get_arguments(app_path, app_args)
+    command_line_args.append(f'--input_dir={corpus_dir}')
 
-    result = _run_with_interpreter_if_needed(fuzzer_path, args, max_time)
+    result = _run_with_interpreter_if_needed(fuzzer_path, command_line_args,
+                                             max_time)
     crashes = []
     for testcase_path in os.listdir(reproducers_dir):
       if not testcase_path.startswith(TESTCASE_PREFIX):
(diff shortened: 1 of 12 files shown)


src/python/bot/fuzzers/blackbox/engine.py
```diff
@@ -81,18 +88,12 @@ def prepare(self, corpus_dir, target_path, build_dir):  # pylint: disable=unused
...
       A FuzzOptions object.
     """
-    app_path = environment.get_value('APP_PATH')
-    app_args = testcase_manager.get_command_line_for_application(
-        get_arguments_only=True).strip()
-
-    arguments = _get_arguments(app_path, app_args)
-    arguments.append(f'--input_dir={corpus_dir}')
-
+    executable_path = environment.get_value('FUZZER_EXECUTABLE_PATH')
```
oliverchang, line 91: Rather than use an environment variable to hack around the interface, can we just pass the full path to the fuzzer as target_path instead? 

Right now target_path is fuzzer_name, but there's really no reason we need to care about what the name of the fuzzer is here. It's only used to get the full path to the fuzzer anyway, so let's just compute the full path beforehand and pass it here.

src/python/bot/fuzzers/blackbox/engine.py
```diff
@@ -26,6 +26,13 @@
 OUTPUT_PREFIX = 'output-'
 
 
+def _get_fuzzer_path(fuzzer_name, executable_path):
```
oliverchang, line 29: As discussed below, let's not do this here.

src/python/bot/fuzzers/blackbox/engine.py
```diff
@@ -137,7 +145,7 @@ def fuzz(self, target_path, options, reproducers_dir, max_time):
         continue
 
       full_testcase_path = os.path.join(reproducers_dir, testcase_path)
-      crash = engine.Crash(full_testcase_path, output, args,
+      crash = engine.Crash(full_testcase_path, output, options.arguments,
```
oliverchang, line 148: Do we need to set options.arguments here at all? It's just the relative path to the fuzzer. We can just pass [] here?

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1873,14 +1881,20 @@ def run(self):
...
     # apply some of the time. Adjust APP_ARGS for them if needed.
     trials.setup_additional_args_for_app()
 
-    engine_impl = engine.get(fuzzer.name)
+    # TODO(mbarbella): Remove the environment variable check when the old
+    # blackbox pipeline has been converted to the engine interface.
+    if environment.get_value('USE_BLACKBOX_ENGINE'):
+      engine_impl = engine.get('blackbox')
+    else:
+      engine_impl = engine.get(self.fuzzer.name)
```
oliverchang, line 1896: Let's do it like this to prevent potential issues with never picking the grey box engine if we somehow bungle up our handling of env vars and always keep USE_BLACKBOX_ENGINE set.

```python
engine_impl = engine.get(self.fuzzer.name)
if not engine_impl and environment.get_value('USE_BLACKBOX_ENGINE'):
  engine_impl = engine.get('blackbox')
```

oliverchang (changes requested):

src/python/bot/fuzzers/engine_common.py
```diff
@@ -281,19 +282,28 @@ def dump_big_query_data(stats, testcase_file_path, fuzzer_command):
   fuzzer_stats.TestcaseRun.write_to_disk(testcase_run, testcase_file_path)
 
 
-def find_fuzzer_path(build_directory, fuzzer_name):
+def find_fuzzer_path(build_directory, engine_name, target_name):
```
oliverchang, line 285: Rather than passing `engine_name`, can we just call it `blackbox` since the value doesn't matter beyond it being 'blackbox'? This function shouldn't need to know about the names of the implemented engines either (as that's an implementation detail). We can also give it a default value of False and make it easier for non-blackbox callers to use.

oliverchang (commented):
Thanks for the changes! A few more suggestions.

src/python/bot/fuzzers/engine_common.py
mbarbella-chromium: That makes sense. I was a bit put off by this initially since it seemed error prone for callers to have to remember to do this, but looking at the usage it seems unlikely that we'll really be adding new calls where this matters.

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1571,6 +1573,10 @@ def generate_blackbox_testcases(self, fuzzer, fuzzer_directory,
 
   def do_engine_fuzzing(self, engine_impl):
     """Run fuzzing engine."""
+    # For blackbox fuzzers, |FUZZ_TARGET| should be set to the fuzzer name.
+    if engine_impl.name == 'blackbox':
+      environment.set_value('FUZZ_TARGET', self.fuzzer_name)
+
     # Record fuzz target.
     fuzz_target_name = environment.get_value('FUZZ_TARGET')
     self.fuzz_target = record_fuzz_target(engine_impl.name, fuzz_target_name,
```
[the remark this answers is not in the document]
mbarbella-chromium: This makes some of the rest of this function and a bit of the plumbing a lot messier. Any specific concerns here, or did you just want to avoid writing potentially confusing entries like blackbox_fuzzer_name that might show up in lists?

I worry that this might cause some issues down the road if we try to handle these as a special case, but if you have some specific concerns I can try to address them.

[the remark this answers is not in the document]
mbarbella-chromium: Looking more, I'm less convinced we can actually get rid of this. It looks like we need the entity for functions like engine_common.get_all_issue_metadata_for_testcase.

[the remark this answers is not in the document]
oliverchang: OK we can leave for now for testing if it's going to end up causing exceptions. However this doesn't really make sense to write so we should figure out a story for not doing this -- e.g. even for get_all_issue_metadata_for_testcase that's just the get the name of the binary so we can get TARGET.labels etc metadata files, which doesn't really apply here.

There might be more cases where things don't fit as-is or in a weird way, so maybe we should take a step back and see if there's some more general refactoring we can do to make things fit better rather than leaving things hackily like this.

oliverchang (approved):
LGTM to test this. But as mentioned in https://github.com/google/clusterfuzz/pull/2049#discussion_r496346946 it's not ideal that things like FuzzTarget entities are written when it doesn't make sense to. There may be other unintended side effects too so we should look at if there's some way to refactor/generalize these to make things fit a bit better.

Perhaps it might make more sense to keep the interface, but separate out some of the blackbox specific pipeline from the rest of the engine pipeline to make things cleaner. Otherwise inevitably we'll end up in the same situation we were in before where we shoehorned engine fuzzers into the blackbox interface we had -- this is just the other way around.

mbarbella-chromium:
Sounds good. Let's see what things look like with this version, and I'll plan on a few follow up PRs to deal with whatever issues we find.

PR #2053, opened 2020-09-28
Remove entropic strategy (fixes #2052).

### LINKED ISSUE #2052: Entropic is now the default in libFuzzer
Matt landed it 9 days ago in https://reviews.llvm.org/D87476

I think we should probably delete that strategy at all from CF, to keep things simple. Does anyone think we should invert the existing strategy or do anything else with it?

Dor1s:


src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_stats_test.py
```diff
@@ -95,7 +95,7 @@ def test_parse_log(self):
         'slow_units_count': 0,
         'startup_crash_count': 0,
         'strategy_dataflow_tracing': 0,
-        'strategy_entropic': 1,
+        'strategy_entropic': 0,
```
Dor1s, line 98: @oliverchang should I remove this columns from the BQ schema? Would that work, or we still shouldn't do that?

If not, do we always need to have the default value dumped in the stats, or it'll be auto-assigned to NULL?

I recall you've been fixing that issue with the columns some time ago, so I'm hoping you remember what we can and cannot do now :)

oliverchang: I think we can remove it here, as the schema is actually defined in `python/metrics/fuzzer_stats_schema.py` now  which still has it. It should be NULL by default

Now the question is can we remove it from fuzzer_stats_schema.py ? I'm not sure :) And https://cloud.google.com/bigquery/docs/managing-table-schemas also seems to say that deleting a column requires manual intervention which might indicate it's safest to leave it there for now.

oliverchang (approved):

jonathanmetzman (approved):
LGTM

src/python/tests/core/bot/fuzzers/libFuzzer/libfuzzer_stats_test.py
Dor1s: Thanks, Oliver!

PR #2058, opened 2020-10-01
Syzkaller corpus sync fixes.

inferno-chromium:
- In fuzz task, force corpus sync as other userspace fuzzing relies
  on unique sha hash filenames to make sync decisions.
- To avoid bad corpus overwrites, keep corpus db for different
  serials seperate for now.
- Using serial over device family is advantageous as device can
  disappear after syzkaller fuzzing. So, we just use serial or
  could have cached device family before fuzzing.


oliverchang (approved):

inferno-chromium:
Just fyi for @mspectorgoogle, this fixes the issue with ".db" in corpus bucket [since device disappears after fuzzing, so device family name=''], and the issue with blueline.db not updated [add force code in cf to always update corpus for that device serial]. Right now, to avoid corruption with concurrent updates, i am using <SERIAL>.db.

PR #2060, opened 2020-10-01
Request bisection for whole range again if range is invalid.

oliverchang:
In some cases, regression and progression tasks may bisect to the commit
range X..X, which is really invalid as it represents an empty range.
Rather than assume X is the correctly bisected commit, re-request
bisection for the whole range.

### DIFF
--- a/src/python/bot/tasks/task_creation.py
+++ b/src/python/bot/tasks/task_creation.py
@@ -287,7 +287,7 @@ def request_bisection(testcase_id):
 
 def _check_commits(testcase, bisect_type, old_commit, new_commit):
   """Check old and new commit validity."""
-  if old_commit != new_commit:
+  if old_commit != new_commit or build_manager.is_custom_binary():
     return old_commit, new_commit
 
   # Something went wrong during bisection for the same commit to be chosen for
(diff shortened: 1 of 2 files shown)


mbarbella-chromium (approved):

src/python/bot/tasks/task_creation.py
```diff
@@ -285,6 +285,35 @@ def request_bisection(testcase_id):
...
 
+def _check_commits(testcase, bisect_type, old_commit, new_commit):
+  """Check old and new commit validity."""
+  if old_commit != new_commit:
+    return old_commit, new_commit
+
+  # Something went wrong during bisection for the same commit to be chosen for
+  # both the start and end range.
+  # Get the bisection infrastructure to re-bisect.
+  bucket_path = build_manager.get_primary_bucket_path()
```
inferno-chromium, line 296: What if this is a custom binary job, do we need any bailout for that ?

inferno-chromium (approved):

src/python/bot/tasks/task_creation.py
oliverchang: Added a check for this.

PR #2061, opened 2020-10-01
Fix syzkaller check in init_corpus and save_corpus.

inferno-chromium:


inferno-chromium:
The order of the check needs fixing, that check is for save_corpus.

hzawawy (approved):

PR #2062, opened 2020-10-01
Check fuzzer before job when preparing fuzzer logs links.

mbarbella-chromium:
PTAL. This should fix an issue where we were generating report links for parent fuzzers, which would be invalid (and which don't really make sense). By switching the order of this check, if the fuzzer is a parent fuzzer we'll bail out early. As of the time of posting, this change is live on OSS-Fuzz staging if you want to test it out.

### DIFF
--- a/src/python/metrics/fuzzer_stats.py
+++ b/src/python/metrics/fuzzer_stats.py
@@ -566,11 +566,11 @@ class FuzzerRunLogsContext(BuiltinFieldContext):
 
   def get_logs_bucket(self, fuzzer_name=None, job_type=None):
     """Return logs bucket for the job."""
-    if job_type:
-      return self._get_logs_bucket_from_job(job_type)
-
     if fuzzer_name:
       return self._get_logs_bucket_from_fuzzer(fuzzer_name)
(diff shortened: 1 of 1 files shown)


inferno-chromium:
Dont know this code, will let @oliverchang review.

oliverchang (approved):
LGTM, but I *think* we made a conscious decisions to have the previous order for performance reasons. Does this impact how long it takes for the stats page to load?

mbarbella-chromium:
I didn't notice any major difference in the load time. I'll land this after fixing the test failures.

mbarbella-chromium:
Looking at the test failures more carefully I'm less convinced that this is the actual fix. Oliver, do we have a clean way to check that a given fuzzer name is a parent fuzzer? If so, do you think it would be reasonable to simply add a check for that to the top of this function and bail out in that case?

oliverchang:
From App Engine side, the easiest way might be to check the 'builtin' property of the relevant Fuzzer entity.

oliverchang:
Closing as obsolete.

PR #2063, opened 2020-10-01
Minor reproduce tool fixes.

mbarbella-chromium:
1) Increment version so that new dependency updates are picked up by
users.
2) Do not assume that pipenv is installed before deps have been
installed.


mbarbella-chromium:
TBR to get the new version of the tool out. Should be fairly straightforward.

PR #2064, opened 2020-10-02
Remove unnecessary builtins imports.

oliverchang:
Also fix some minor remaining lint issues for the affected files.

### DIFF
--- a/docker/oss-fuzz/host/start_host.py
+++ b/docker/oss-fuzz/host/start_host.py
@@ -14,9 +14,6 @@
 """Start host."""
 from __future__ import print_function
 
-from builtins import range
-from builtins import str
-
 from future import standard_library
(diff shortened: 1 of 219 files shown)


docker/oss-fuzz/host/start_host.py
```diff
@@ -14,9 +14,6 @@
 """Start host."""
 from __future__ import print_function
```
inferno-chromium, line 15: Should all these from __future__ go away too.

inferno-chromium (approved):
Amazing!

docker/oss-fuzz/host/start_host.py
oliverchang: Will do this in another PR :)

PR #2067, opened 2020-10-02
Make ML RNN generator work without legacy .meta files.

Dor1s:
@mihaimaruseac after I've deleted old `.meta` files, the generation stopped working (and it didn't work for new fuzzers whose models didn't have `.meta` files).

I've deleted code handling `.meta` files, but we also need to migrate from the legacy session restore and run to the new `model` thing.

I've found the following guide: https://www.tensorflow.org/guide/migrate#converting_models

But it isn't straightforward to me, as we have slightly different function arguments and output values.

Would you be able to guide me please in fixing the generation code?

### DIFF
--- a/src/python/bot/fuzzers/ml/rnn/constants.py
+++ b/src/python/bot/fuzzers/ml/rnn/constants.py
@@ -59,7 +59,6 @@ TRAINING_SCRIPT_NAME = 'train.py'
 GENERATION_SCRIPT_NAME = 'generate.py'
 
 # Model files suffix.
-MODEL_META_SUFFIX = '.meta'
 MODEL_DATA_SUFFIX = '.data-00000-of-00001'
 MODEL_INDEX_SUFFIX = '.index'
 
--- a/src/python/bot/fuzzers/ml/rnn/generate.py
+++ b/src/python/bot/fuzzers/ml/rnn/generate.py
@@ -73,9 +73,11 @@ def main(args):
     print('\nusing model {} to generate {} inputs...'.format(model_path, count))
 
     # Restore the model.
-    new_saver = tf.compat.v1.train.import_meta_graph(
-        model_path + constants.MODEL_META_SUFFIX)
-    new_saver.restore(session, model_path)
+    # Build the RNN model.
+    model = utils.build_model(
+        constants.HIDDEN_STATE_SIZE * constants.HIDDEN_STATE_SIZE,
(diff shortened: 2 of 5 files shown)


mihaimaruseac (approved):
This looks good to me, I cannot see the failing build's error.

The only culprit I can think of before seeing it is that `model_path` might not be the same as the one where the checkpoints are. I could go into debugging that, if I have the build error.

Dor1s:
> This looks good to me, I cannot see the failing build's error.
> 
> The only culprit I can think of before seeing it is that `model_path` might not be the same as the one where the checkpoints are. I could go into debugging that, if I have the build error.

The problem is that I've replaced `import_meta_graph` and `restore()` here https://github.com/google/clusterfuzz/blob/438909e4a0a3ebaa528614bd4f61a97a7889abb8/src/python/bot/fuzzers/ml/rnn/generate.py#L75

with the new `model.load_weights` convention, but later on as we do the actual generation via `session.run`: https://github.com/google/clusterfuzz/blob/master/src/python/bot/fuzzers/ml/rnn/generate.py#L124

we need to replace it with some `forward` function (which can be optionally marked with `@tf.function` decorator, from https://www.tensorflow.org/guide/migrate#converting_models) -- this stuff isn't clear to me. I can figure it out after spending more time, but I suspect that it might be something very obvious for you, since you've been converting the rest of the code.

mihaimaruseac:
I'll look after this this week. Sorry for the delay it causes :(

Dor1s:
Thanks a lot @mihaimaruseac, no worries. I'll try to give this another shot too later this week -- maybe I'll get it pretty quickly :)

mihaimaruseac:
Sorry for the delay, I just got a chance to look at this now.

I recall doing a generation during training too, it is at https://github.com/mihaimaruseac/clusterfuzz/blob/8559f727b4396558442f448d20b41c40535544d7/src/python/bot/fuzzers/ml/rnn/train.py#L264-L290

We can extract this to a separate function and call it from here too. I'll make a PR based on this one.

mihaimaruseac:
#2083 should build on top of this branch.

inferno-chromium:
Can you 'git pull' and 'git rebase' to tip of tree trunk, unsure why those lint error show up in unrelated files.

Dor1s:
This was incorporated into https://github.com/google/clusterfuzz/pull/2083

PR #2068, opened 2020-10-04
Android fixes

inferno-chromium:
- Don't run initialize_device or try to configure system build.prop
  changes on libFuzzer and syzkaller jobs.
- Initialize environment early on, and make sure that stacktrace
  generation uses environment vars rather than trying to query a
  possibly dead device.


inferno-chromium:
TBR minor fixes

mbarbella-chromium (commented):
LGTM for debugging.

PR #2076, opened 2020-10-07
Fix recursion issue with Android usb reset.

inferno-chromium:
wait_for_device inside usb_reset can try to recover a bad device
recursing into usb_reset again. Avoid stack recursion and don't
recover in wait_for_device call for that instance.


inferno-chromium:
TBR minor fix.

mbarbella-chromium (commented):
LGTM

PR #2078, opened 2020-10-08
Begin refactoring stack_analyzer for libClusterFuzz.

oliverchang:
- Move majority of code to lib/clusterfuzz/stacktraces.
- Introduce StackParser class to better store/keep state.
- Rename StackAnalyzerState to stacktraces.CrashInfo.  Also move several
  attributes that are better suited to the StackParser class.
- stack_analyzer becomes an embedder, and does some additional steps
  such as symbolisation and Android linkification.
- Remove StackAnalyzerState.tool and tool_from_group functionality. They
  were unused.


oliverchang:
PTAL. There should be no functional differences.

Will be doing a lot of testing on actual testcases to make sure states don't change before landing this.

mbarbella-chromium (approved):
LGTM overall. I have some concerns about leaving the tests as they are now, especially if the end goal is to factor this code out. That's probably better for another PR after you're confident this is working, but we'll probably want the old stack analyzer test to test the integration between these two pieces, with most of the existing tests factored out to some new file.

oliverchang:
Thanks! Could you please clarify some of your concerns here? As is the tests represent an end-to-end usage (with CF being the embedder) so in terms of test coverage it seems OK. It certainly might be neater to move some of the tests over, but did you see foresee any other issues?

That said, I'm not 100% sure if things will stay in the same lib/clusterfuzz directory as is, depending on how easy it is to untie our dependencies for our engine related stuff

src/python/lib/clusterfuzz/stacktraces/__init__.py
```diff
@@ -0,0 +1,1362 @@
...
+    # Remove non-printable chars from crash state.
+    state.crash_state = ''.join(
+        s for s in state.crash_state if s in string.printable)
+
+    # Shorten JNI messages.
+    if JNI_ERROR_STRING in state.crash_state:
+      state.crash_state = state.crash_state.replace(JNI_ERROR_STRING, 'JNI:')
+
+    if self.symbolized:
+      # 1. Normalize addresses and numbers in crash_state.
```
inferno-chromium, line 303: nit: maybe lets remove these 1., 2., since missing at end, can remove.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -1199,789 +48,56 @@ def get_crash_data(crash_data,
...
-        ASAN_MEMCPY_OVERLAP_REGEX,
-        line,
-        state,
-        new_type='Memcpy-param-overlap',
-        reset=True,
-        address_from_group=2)
+  stack_parser = stacktraces.StackParser(
+      symbolized=symbolize_flag or already_symbolized,
+      detect_ooms_and_hangs=detect_ooms_and_hangs,
+      detect_v8_runtime_errors=detect_v8_runtime_errors,
```
inferno-chromium, line 76: This detect_v8_runtime_errors is very chromium specific, wonder if we should keep it just inside stackparser as a hack.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -1199,789 +48,56 @@ def get_crash_data(crash_data,
...
-      for python_crash_regex, python_crash_type in PYTHON_CRASH_TYPES_MAP:
-        if update_state_on_match(
-            python_crash_regex, line, state, new_type=python_crash_type):
-          found_python_crash = True
-          state.crash_state = ''
-          state.frame_count = 0
-          continue
+  # Use stacktrace with inlines for the result.
+  if result.crash_stacktrace:
+    result.crash_stacktrace = crash_stacktrace_with_inlines
```
inferno-chromium, line 85: Is it += ?

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -1199,789 +48,56 @@ def get_crash_data(crash_data,
...
-      continue
-
-    # Python stack frames.
-    if is_python and add_frame_on_match(
-        PYTHON_STACK_FRAME_FUNCTION_REGEX, line, state, group=3):
-      continue
-
-  # Only get repo.prop if we have an Android kernel or KASAN crash
-  if environment.is_android() and (found_android_kernel_crash or is_kasan):
+  # Linkify Android stacktrace.
```
inferno-chromium, line 87: nit: maybe move this block into its own function. (this file is fine since hacky anyway).

src/python/lib/clusterfuzz/stacktraces/__init__.py
```diff
@@ -0,0 +1,1362 @@
...
+
+    self.is_kasan = False
+    self.is_golang = False
+    self.is_python = False
+    self.found_python_crash = False
+    self.found_golang_crash = False
+    self.found_android_kernel_crash = False
+
+
+def _filter_stack_frame(stack_frame):
```
inferno-chromium, line 66: nit: can be moved after StackParser, next to other funvtions

inferno-chromium (approved):
Great!

src/python/crash_analysis/stack_parsing/stack_analyzer.py
oliverchang: Previously it was, but that was just a ''  += so it's the same.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
oliverchang: Yes thought about it, but it needs a a bit more refactoring to make it work. Will keep this in mind for future work here.

src/python/lib/clusterfuzz/stacktraces/__init__.py
oliverchang: This can't be moved as it has to come before the default argument value in add_frame_on_match.

src/python/lib/clusterfuzz/stacktraces/__init__.py
oliverchang: Removed the numbers!

src/python/crash_analysis/stack_parsing/stack_analyzer.py
oliverchang: Done!

PR #2087, opened 2020-10-12
Assign a default Pri-1 when filing monorail issues.

mbarbella-chromium:


inferno-chromium (changes requested):
i think real issue might be missing pri label in clusterfuzz-config for chromium project.
configs/internal/issue_trackers/config.yaml#25

mbarbella-chromium:
Good catch, that does seem a bit cleaner. I still wouldn't mind landing this if we decide we want the failsafe, but for now I'll assume we don't want to do this.

PR #2088, opened 2020-10-13
Add support for parsing LKL style crashes

mspectorgoogle:
LKL or Linux Kernel Library allows libfuzzer to be used to fuzz the
linux kernel by including it as a userspace library.  Crashes appear as
a libfuzzer "deadly signal" with a KASAN crash inside.  This change
allows for parsing these KASAN crashes with higher priority than the
libfuzzer crash to insure the correct crash_type and state.

### DIFF
--- a/src/python/lib/clusterfuzz/stacktraces/__init__.py
+++ b/src/python/lib/clusterfuzz/stacktraces/__init__.py
@@ -56,6 +56,7 @@ class CrashInfo:
     self.fatal_error_occurred = False
 
     self.is_kasan = False
+    self.is_lkl = False
     self.is_golang = False
     self.is_python = False
     self.found_python_crash = False
@@ -359,13 +360,27 @@ class StackParser:
 
     return state
 
+  def remove_lkl_kernel_times(self, crash_data):
+    """Filter kernel time from crash_data if LKL."""
+    result = ''
+    for line in crash_data.splitlines():
+      result += ANDROID_KERNEL_TIME_REGEX.sub('', line) + '\n'
+    return result
+
   def parse(self, stacktrace: str) -> CrashInfo:
     """Parse a stacktrace."""
     state = CrashInfo()
     state.crash_stacktrace = stacktrace
     state.is_kasan = 'KASAN' in stacktrace
+    state.is_lkl = 'Linux Kernel Library Stack Trace:' in stacktrace
     state.is_golang = '.go:' in stacktrace
     state.is_python = '.py", line' in stacktrace
+
+    # For Android LKL (and potentially kernel output), the KASAN crash may start
+    # with the time since boot.  We need to remove this so that our regexes
+    # match.
+    if state.is_lkl:
+      stacktrace = self.remove_lkl_kernel_times(stacktrace)
 
     split_crash_stacktrace = stacktrace.splitlines()
 
@@ -389,10 +404,13 @@ class StackParser:
           continue
 
       # Assertions always come first, before the actual crash stacktrace.
-      self.match_assert(line, state, ASSERT_REGEX)
-      self.match_assert(line, state, ASSERT_REGEX_GOOGLE, group=2)
-      self.match_assert(line, state, ASSERT_REGEX_GLIBC)
-      self.match_assert(line, state, RUST_ASSERT_REGEX)
+      # However if we already have a kernel crash, we don't want to
+      # replace it with the ASSERT.
+      if state.crash_type.startswith('Kernel failure'):
+        self.match_assert(line, state, ASSERT_REGEX)
+        self.match_assert(line, state, ASSERT_REGEX_GOOGLE, group=2)
+        self.match_assert(line, state, ASSERT_REGEX_GLIBC)
+        self.match_assert(line, state, RUST_ASSERT_REGEX)
 
       # ASSERT_NOT_REACHED prints a single line error then triggers a crash. We
       # set the crash state here, but look for the stack after a crash on an
--- a/src/python/lib/clusterfuzz/stacktraces/__init__.py
+++ b/src/python/lib/clusterfuzz/stacktraces/__init__.py
@@ -406,7 +406,7 @@ class StackParser:
       # Assertions always come first, before the actual crash stacktrace.
       # However if we already have a kernel crash, we don't want to
       # replace it with the ASSERT.
-      if state.crash_type.startswith('Kernel failure'):
+      if not state.crash_type.startswith('Kernel failure'):
         self.match_assert(line, state, ASSERT_REGEX)
         self.match_assert(line, state, ASSERT_REGEX_GOOGLE, group=2)
         self.match_assert(line, state, ASSERT_REGEX_GLIBC)
--- a/src/python/lib/clusterfuzz/stacktraces/constants.py
+++ b/src/python/lib/clusterfuzz/stacktraces/constants.py
@@ -34,9 +34,10 @@ ANDROID_KERNEL_ERROR_REGEX = re.compile(
     r'.*Internal error: (Oops)?( -|:) (BUG|[0-9a-fA-F]+)')
 ANDROID_KERNEL_STACK_FRAME_REGEX = re.compile(
     # e.g. "[ 1998.156940] [<c0667574>] "
-    r'[^(]*\[\<([0-9a-fA-F]+)\>\]\s+'
+    r'[^(]*\[\<([x0-9a-fA-F]+)\>\]\s+'
     # e.g. "(msm_vidc_prepare_buf+0xa0/0x124)"; function (3), offset (4)
     r'\(?(([\w]+)\+([\w]+)/[\w]+)\)?')
+ANDROID_KERNEL_TIME_REGEX = re.compile(r'^\[\s*\d+\.\d+\]\s')
 # Parentheses are optional.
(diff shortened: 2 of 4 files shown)


src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Stack analyzer module."""
+import re
```
inferno-chromium, line 15: This is not needed here, see comment below.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -37,6 +38,15 @@ def linkify_android_stacktrace(crash_info):
     crash_info.crash_stacktrace = temp_crash_stacktrace
 
 
+def remove_kernel_times(crash_data):
+  """Filter kernel time from crash_data."""
+  result = ''
+  strip_regex = re.compile(r'^\[\s*\d+\.\d+\]\s*')
```
inferno-chromium, line 44: this compiled regex should be in constants.py

ANDROID_KERNEL_TIME_REGEX

Also "]\s*" should be just "]\s", i think we don't match multiple space after time as it strips more than needed.

src/python/lib/clusterfuzz/stacktraces/constants.py
```diff
@@ -328,6 +328,8 @@
     r'^pthread\_kill$',
     r'^raise$',
     r'^tgkill$',
+    r'^print_address_description$',
```
inferno-chromium, line 331: Add a comment line before this line.
"# Android kernel stack frame ignores."

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -1591,7 +1591,7 @@ def test_syzkaller_kasan(self):
                       'sock_def_readable\n'
                       'unix_dgram_sendmsg\n')
     expected_address = '0xffffffc01640e9d0'
-    expected_stacktrace = data
+    expected_stacktrace = stack_analyzer.remove_kernel_times(data)
```
inferno-chromium, line 1594: We shouldnt call function here in test code, this should be an "kasan_syzkaller_stripped.txt" file and then we read it here for raw contents. if the files are same (no times), then use same file kasan_syzkaller.txt.

src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -2950,3 +2950,16 @@ def test_rust_oom(self):
     self._validate_get_crash_data(data, expected_type, expected_address,
                                   expected_state, expected_stacktrace,
                                   expected_security_flag)
+
+  def test_linux_kernel_library_libfuzzer(self):
+    """Test Linux Kernel Library fuzzed with libfuzzer ."""
+    data = self._read_test_data('lkl_libfuzzer.txt')
```
inferno-chromium, line 2956: We shouldnt call function here in test code, this should be an "lkl_libfuzzer_stripped.txt" file and then we read it here for raw contents.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -37,6 +38,15 @@ def linkify_android_stacktrace(crash_info):
     crash_info.crash_stacktrace = temp_crash_stacktrace
 
 
+def remove_kernel_times(crash_data):
```
oliverchang, line 41: s/remove_kernel_times/remove_android_kernel_times/

src/python/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -48,6 +58,12 @@ def get_crash_data(crash_data,
   unless |symbolize| flag is set to False. Symbolized stacktrace will contain
   inline frames, but we do exclude them for purposes of crash state generation
   (helps in testcase deduplication)."""
+
+  # For Android LKL (and potentially kernel output), the KASAN crash may start
+  # with the time since boot.  We need to remove this so that our regexes match.
+  if '] BUG: KASAN:' in crash_data:
```
oliverchang, line 64: Add a check to make sure this is Android.

inferno-chromium (approved):
LGTM with some minor changes

oliverchang (commented):
Some comments re the current implementation, but would it be possible to modify the existing regexes to allow for this optional kernel timestamp, rather than removing it? That would be a much cleaner approach.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
inferno-chromium: +1

src/python/crash_analysis/stack_parsing/stack_analyzer.py
inferno-chromium: +1

inferno-chromium:
> Some comments re the current implementation, but would it be possible to modify the existing regexes to allow for this optional kernel timestamp, rather than removing it? That would be a much cleaner approach.

I think this applies to all output and not some frames. E.g. these are sanitizer style frames - "[    0.058680] #0 [<0x000000000062a66a>] print_address_description+0x6a/0x5c0", so stripping seems better.

oliverchang:
> > Some comments re the current implementation, but would it be possible to modify the existing regexes to allow for this optional kernel timestamp, rather than removing it? That would be a much cleaner approach.
> 
> I think this applies to all output and not some frames. E.g. these are sanitizer style frames - "[ 0.058680] #0 [<0x000000000062a66a>] print_address_description+0x6a/0x5c0", so stripping seems better. or thoughts ?

I see, so this requires modifying the main `SAN_STACK_FRAME_REGEX`, which is a bit risky? 

In that case this approach seems fine, but I would prefer that we don't return the stripped output in the `CrashInfo` as a "parsing" function should not be modifying the stacktrace as a sideffect.

inferno-chromium:
> > > Some comments re the current implementation, but would it be possible to modify the existing regexes to allow for this optional kernel timestamp, rather than removing it? That would be a much cleaner approach.
> > 
> > 
> > I think this applies to all output and not some frames. E.g. these are sanitizer style frames - "[ 0.058680] #0 [<0x000000000062a66a>] print_address_description+0x6a/0x5c0", so stripping seems better. or thoughts ?
> 
> I see, so this requires modifying the main `SAN_STACK_FRAME_REGEX`, which is a bit risky?
> 
> In that case this approach seems fine, but I would prefer that we don't return the stripped output in the `CrashInfo` as a "parsing" function should not be modifying the stacktrace as a sideffect.

ah right, stacktrace shouldn't be modified when returned, this code needs to be moved inside stack_parser.parse.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
mspectorgoogle: LKL runs on host so its not Android platform sadly.  I am asking xuanxing@ if they could add "LKL" or something to the crash stack so we can identify it.

mspectorgoogle:
The problem is we would have to update A LOT of regex (all KASAN all kernel etc).  I wanted to make sure the symbolizer will work correctly later hence why I wanted to change it pre-symbolize.  Let me look at symbolizer code and see if it will break without stripping.

src/python/crash_analysis/stack_parsing/stack_analyzer.py
inferno-chromium: The job type is selected to use android platform right, so environment.is_android() should still work.

mspectorgoogle: No its a host based fuzzer, does not run on an android phone runs on x86 GCE host

mspectorgoogle: We have added 'Linux Kernel Library Stack Trace:' so we know explicitly now

src/python/crash_analysis/stack_parsing/stack_analyzer.py
mspectorgoogle: remove_lkl_kernel_times

src/python/lib/clusterfuzz/stacktraces/__init__.py
```diff
@@ -389,10 +404,15 @@ def parse(self, stacktrace: str) -> CrashInfo:
           continue
 
       # Assertions always come first, before the actual crash stacktrace.
-      self.match_assert(line, state, ASSERT_REGEX)
-      self.match_assert(line, state, ASSERT_REGEX_GOOGLE, group=2)
-      self.match_assert(line, state, ASSERT_REGEX_GLIBC)
-      self.match_assert(line, state, RUST_ASSERT_REGEX)
+      # However if we have and LKL crash and already have a kernel crash, we
+      # don't want to replace it with the ASSERT.
+      if not state.is_lkl and (
```
inferno-chromium, line 409: I think we should simplify this

if not state.crash_type.startswith('Kernel failure'):

should be sufficient. If we have kernel failure, we dont want to overwrite with assert.

inferno-chromium (approved):
LGTM with minor change needed.

PR #2089, opened 2020-10-13
Start adding engine support to libClusterFuzz.

oliverchang:
Move engine.py to lib/clusterfuzz/fuzz/. Engine implementations are
imported from their existing locations in ClusterFuzz.

Also:

- Add initial setup.py
- Move redis imports to be deferred as a start to reducing the number of dependencies
  required by this package.

### DIFF
--- a/src/python/lib/setup.py
+++ b/src/python/lib/setup.py
@@ -1,3 +1,17 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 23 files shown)


src/python/lib/setup.py
```diff
@@ -0,0 +1,38 @@
...
+import setuptools
+
+with open('README.md', 'r') as fh:
+  long_description = fh.read()
+
+setuptools.setup(
+    name='clusterfuzz',
+    version='0.0.1',
+    author='ClusterFuzz authors',
+    author_email='clusterfuzz-announce@googlegroups.com',
```
inferno-chromium, line 24: should this be clusterfuzz-dev@ ?

inferno-chromium (approved):

src/python/lib/setup.py
oliverchang: Sure, created clusterfuzz-dev@googlegroups.com

PR #2090, opened 2020-10-13
[Do not land] Use a single environment variable to specify ML models for fuzzing.

mbarbella-chromium:
PTAL. I'll need to update the job definitions before landing this.

### DIFF
--- a/src/appengine/handlers/cron/ml_train.py
+++ b/src/appengine/handlers/cron/ml_train.py
@@ -19,6 +19,12 @@ from datastore import data_types
 from datastore import fuzz_target_utils
 from handlers import base_handler
 from libs import handler
+from metrics import logs
+
+MODEL_NAME_TO_TASK = {
+    'gradientfuzz': 'train_gradientfuzz',
+    'rnn_generator': 'train_rnn_generator',
+}
 
 
 class Handler(base_handler.Handler):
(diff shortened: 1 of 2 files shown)


src/appengine/handlers/cron/ml_train.py
```diff
@@ -19,6 +19,12 @@
 from datastore import fuzz_target_utils
 from handlers import base_handler
 from libs import handler
+from metrics import logs
+
+MODEL_NAME_TO_TASK = {
+    'gradientfuzz': 'train_gradientfuzz',
```
inferno-chromium, line 25: Should we rename gradientfuzz to something like gradient_generator or something similar to rnn_generator ?
Same for train_gradientfuzz files ?

src/appengine/handlers/cron/ml_train.py
```diff
@@ -29,17 +35,18 @@ def get(self):
...
+      models = job.get_environment().get('ML_MODELS_TO_USE')
+      if not models:
         continue
 
       task_list = []
-      if utils.string_is_true(job.get_environment().get('USE_GRADIENTFUZZ')):
-        task_list.append('train_gradientfuzz')
-      if utils.string_is_true(job.get_environment().get('USE_RNN_GENERATOR')):
-        task_list.append('train_rnn_generator')
+      for model_name in models.split(','):
```
inferno-chromium, line 43: models.strip().split(',')

inferno-chromium (approved):

src/appengine/handlers/cron/ml_train.py
mbarbella-chromium: I think it's reasonable to leave it as-is, I don't think it's necessary for everything to be prefixed with generator. We made it a point to rename rnn to rnn_generator because it was starting to look like "rnn" was too generic a name to make sense for what we were doing (we can and probably will use RNNs for other things, so we had to be more specific).

src/appengine/handlers/cron/ml_train.py
mbarbella-chromium: I think the strip below should cover any cases this would handle, and that one will be necessary regardless for cases like "... = model1, model2".

mbarbella-chromium:
I've updated the jobs for Chrome and didn't see any usage on our other deployments. Hopefully that's everything, but if we are using this outside of Chrome those jobs will need to be updated as well.

Dor1s (commented):
LGTM! Sorry for the delay. Glad we got it working!

PR #2091, opened 2020-10-14
Don't ignore exceptions in issue_tracker_utils.get_issue_for_testcase.

oliverchang:
This can lead to bad updates in cleanup task, as it may mark testcases
as triaged even though there was an error in accessing issues. This can
lead to issues getting left open indefinitely.

Prefer to exception out early here.

### DIFF
--- a/src/appengine/libs/issue_management/issue_tracker_utils.py
+++ b/src/appengine/libs/issue_management/issue_tracker_utils.py
@@ -20,7 +20,6 @@ from libs import request_cache
 from libs.issue_management import issue_tracker_policy
 from libs.issue_management import jira
 from libs.issue_management import monorail
-from metrics import logs
 
 _ISSUE_TRACKER_CACHE_CAPACITY = 8
 _ISSUE_TRACKER_CONSTRUCTORS = {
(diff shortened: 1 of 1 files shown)


oliverchang:
PTAL. noticed a few issues that are left open (indefinitely) due to that internal issue tracker breakage a few weeks back.

inferno-chromium (approved):
We don't want exceptions in clean up task as it prevent furthur testcase processing ?

oliverchang:
> We don't want exceptions in clean up task as it prevent furthur testcase processing ?

There is still this top level try/except: https://github.com/google/clusterfuzz/blob/cc88dcd57dd3f60fa4e4a3757cda03da3d3ff98c/src/appengine/handlers/cron/cleanup.py#L126

inferno-chromium:
> > We don't want exceptions in clean up task as it prevent furthur testcase processing ?
> 
> There is still this top level try/except:
> 
> https://github.com/google/clusterfuzz/blob/cc88dcd57dd3f60fa4e4a3757cda03da3d3ff98c/src/appengine/handlers/cron/cleanup.py#L126

yes right, was double checking.

PR #2095, opened 2020-10-19
Remove some dependencies for libClusterFuzz.

oliverchang:
- Move selenium to Pipfile.
- Move python-dateutil to appengine requirements.
- Remove hyperlink pin.
- Remove pytz.
- Remove configparser. This was a Python 2 backport of a Python 3
  module.
- Remove peach, antlr dependencies for Pip package.
- Also some import fixes for the pip package.


src/requirements.txt
```diff
@@ -12,17 +11,13 @@ google-cloud-profiler==1.0.3
...
 oauth2client==4.1.3
 protobuf==3.11.3
-python-dateutil==2.8.1
-pytz==2018.5
 PyYAML==5.1
 redis==3.3.11
 requests==2.21.0
-selenium==3.141.0
 six==1.14.0
 tlslite-ng==0.7.5
```
oliverchang, line 22: tlslite-ng, twisted, and lxml are depended on by peach. This is not obvious at all, since the peach code is a zip in the resources dir. 

@mbarbella-chromium would it be possible to move these dependencies into that zip instead of having them here?

inferno-chromium (approved):

src/requirements.txt
mbarbella-chromium: I think I'd rather remove the zip. I'll file a bug for this.

mbarbella-chromium (approved):

PR #2100, opened 2020-10-20
libClusterFuzz progress.

oliverchang:
- Add test for basic stacktrace parsing.
- Add main module for clusterfuzz.fuzz to allow CLI invocation.
- Add environment.is_lib() to disable certain features that don't make
  sense for the pip package.
- Add config into pip package as a data dependency.

### DIFF
--- a/src/python/lib/clusterfuzz/fuzz/__main__.py
+++ b/src/python/lib/clusterfuzz/fuzz/__main__.py
@@ -1,0 +1,68 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Fuzzing module."""
+
+import argparse
+import os
+
+import clusterfuzz.fuzz
+
+
+def main():
+  os.environ['CONFIG_DIR_OVERRIDE'] = os.path.join(
+      os.path.dirname(os.path.dirname(os.path.abspath(__file__))), 'lib-config')
+  os.environ['PROJECT_NAME'] = 'libClusterFuzz'
+
+  parser = argparse.ArgumentParser(description='Fuzzing tool')
+  parser.add_argument('-t', '--target', help='Path to target.', required=True)
+  parser.add_argument(
+      '-e',
+      '--engine',
+      help='Fuzzing engine.',
+      choices=['libFuzzer', 'honggfuzz'],
+      default='libFuzzer')
+  parser.add_argument(
+      '-s',
+      '--sanitizer',
+      help='Sanitizer.',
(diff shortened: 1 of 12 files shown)


src/python/lib/clusterfuzz/fuzz/__main__.py
```diff
@@ -0,0 +1,68 @@
...
+      type=int,
+      default=3600)
+  args = parser.parse_args()
+
+  # TODO(ochang): Find a cleaner way to propagate this.
+  if args.sanitizer == 'address':
+    os.environ['JOB_NAME'] = 'libfuzzer_asan'
+  elif args.sanitizer == 'memory':
+    os.environ['JOB_NAME'] = 'libfuzzer_msan'
+  elif args.sanitizer == 'memory':
```
mbarbella-chromium, line 57: memory -> undefined

mbarbella-chromium (approved):

src/python/lib/clusterfuzz/fuzz/__main__.py
inferno-chromium: +1

src/python/lib/clusterfuzz/fuzz/__main__.py
```diff
@@ -0,0 +1,68 @@
...
+    os.environ['JOB_NAME'] = 'libfuzzer_ubsan'
+
+  engine_impl = clusterfuzz.fuzz.get_engine(args.engine)
+  options = engine_impl.prepare(args.corpus, args.target, None)
```
inferno-chromium, line 61: use named argument for None

src/python/lib/clusterfuzz/fuzz/__main__.py
```diff
@@ -0,0 +1,68 @@
...
+  result = engine_impl.fuzz(args.target, options, args.output,
+                            args.max_duration)
+  print(result.logs)
```
inferno-chromium, line 64: anything to print if it succeeded or failed, before we print the logs.

src/python/system/environment.py
```diff
@@ -971,7 +971,7 @@ def set_value(environment_variable, value):
   """Set an environment variable."""
   value_str = str(value)
   environment_variable_str = str(environment_variable)
-  value_str = value_str.replace('%ROOT_DIR%', os.environ['ROOT_DIR'])
+  value_str = value_str.replace('%ROOT_DIR%', os.getenv('ROOT_DIR', ''))
```
inferno-chromium, line 974: SHould this be empty string or "."

inferno-chromium (approved):

src/python/lib/clusterfuzz/fuzz/__main__.py
oliverchang: Fixed!

src/python/lib/clusterfuzz/fuzz/__main__.py
oliverchang: Actually this shouldn't be none, so fixed that.

src/python/lib/clusterfuzz/fuzz/__main__.py
oliverchang: Not sure what the definition of succeeded or failed be should be here. Also think there are other improvements to be made here which will come in future PRs (e.g. handling KeyboardInterrupt, streaming logs).

src/python/system/environment.py
oliverchang: For libClusterFuzz, we don't expect ROOT_DIR to be set at all, so have the default of ''.

PR #2104, opened 2020-10-20
add fields from .metadata file to issue description

kalder:
This reads an optional .metadata file, which contains a JSON-formatted
dictionary. These fields will be appended to the issue description.

### DIFF
--- a/src/appengine/libs/issue_management/issue_filer.py
+++ b/src/appengine/libs/issue_management/issue_filer.py
@@ -16,7 +16,6 @@
 import itertools
 import json
 import re
-import six
 
 from base import external_users
 from base import utils
@@ -390,11 +389,6 @@ def file_issue(testcase,
     for result in apply_substitutions(policy, label, testcase,
                                       security_severity):
       issue.labels.add(result)
-
-  # Add additional body text from metadata.
-  additional_fields = json.loads(testcase.get_metadata('additional_fields', ''))
-  for key, value in six.iteritems(additional_fields):
-    issue.body += '\n\n%s: %s' % (key, value)
 
   issue.body += data_handler.format_issue_information(
       testcase, properties.issue_body_footer)
--- a/src/appengine/libs/issue_management/issue_filer.py
+++ b/src/appengine/libs/issue_management/issue_filer.py
@@ -16,6 +16,7 @@
 import itertools
 import json
 import re
+import six
 
 from base import external_users
 from base import utils
--- a/src/appengine/libs/issue_management/issue_filer.py
+++ b/src/appengine/libs/issue_management/issue_filer.py
@@ -14,6 +14,7 @@
 """Helper functions to file issues."""
 
 import itertools
+import json
 import re
 
 from base import external_users
@@ -388,6 +389,11 @@ def file_issue(testcase,
     for result in apply_substitutions(policy, label, testcase,
                                       security_severity):
       issue.labels.add(result)
+
+  # Add additional body text from metadata.
+  additional_fields = json.loads(testcase.get_metadata('additional_fields', ''))
+  for key, value in six.iteritems(additional_fields):
+    issue.body += '\n\n%s: %s' % (key, value)
 
   issue.body += data_handler.format_issue_information(
       testcase, properties.issue_body_footer)
--- a/src/appengine/libs/issue_management/issue_filer.py
+++ b/src/appengine/libs/issue_management/issue_filer.py
@@ -14,7 +14,6 @@
 """Helper functions to file issues."""
 
 import itertools
-import json
 import re
 
 from base import external_users
--- a/src/python/bot/fuzzers/engine_common.py
+++ b/src/python/bot/fuzzers/engine_common.py
@@ -472,9 +472,9 @@ def get_all_issue_metadata(fuzz_target_path):
   if issue_owners:
     metadata['issue_owners'] = ','.join(issue_owners)
 
-  additional_fields = get_additional_fields(fuzz_target_path)
-  if additional_fields:
-    metadata['additional_fields'] = additional_fields
+  additional_issue_fields = get_additional_fields(fuzz_target_path)
+  if additional_issue_fields:
+    metadata['additional_issue_fields'] = additional_issue_fields
 
   return metadata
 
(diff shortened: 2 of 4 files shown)


src/appengine/libs/issue_management/issue_filer.py
```diff
@@ -389,6 +391,11 @@ def file_issue(testcase,
                                       security_severity):
       issue.labels.add(result)
 
+  # Add additional body text from metadata.
+  additional_fields = json.loads(testcase.get_metadata('additional_fields', ''))
```
oliverchang, line 395: lets call the metadata field 'additional_label_fields' to be less ambiguous.

src/appengine/libs/issue_management/issue_filer.py
```diff
@@ -389,6 +391,11 @@ def file_issue(testcase,
...
+  for key, value in six.iteritems(additional_fields):
```
oliverchang, line 396: not need to use six here. just do `additional_fields.items()`. We no longer support python 2.

src/appengine/libs/issue_management/issue_filer.py
```diff
@@ -389,6 +391,11 @@ def file_issue(testcase,
...
+    issue.body += '\n\n%s: %s' % (key, value)
```
oliverchang, line 397: nit: use python 3 f-strings.

e.g. `f'\n\n{key}: {value}`

Also, let's minimize new newlines introduced by appending the double newline once. and then doing a join on the rest on a single newline so that the final result looks lke:

```
...prev part of report

Field1:
Field2: 
```

src/appengine/libs/issue_management/issue_filer.py
oliverchang, line 395: json.loads() does not work with an empty string. please fix this check.

oliverchang, line 395: move this functionality into data_handler.get_issue_description instead. then, add a test similar to the ones here: 

https://github.com/google/clusterfuzz/blob/20a4de8c85595566e4f3ebdfbd49bde7fb090bfd/src/python/tests/core/datastore/data_handler_test.py#L250

oliverchang (changes requested):

src/python/bot/fuzzers/engine_common.py
```diff
@@ -67,6 +67,9 @@
 # Extension for per-fuzz target components to be added to issue tracker.
 COMPONENTS_FILE_EXTENSION = '.components'
 
+# Extension for additional metadata to be added to issue descriptions.
+METADATA_FILE_EXTENSION = '.metadata'
```
oliverchang, line 71: Could you call this `.issue_metadata` instead ?

src/python/bot/fuzzers/engine_common.py
```diff
@@ -436,6 +439,23 @@ def get_issue_components(fuzz_target_path):
   return get_issue_metadata(fuzz_target_path, COMPONENTS_FILE_EXTENSION)
 
 
+def get_additional_fields(fuzz_target_path):
```
oliverchang, line 442: s/get_additional_fields/get_additional_issue_fields/

src/python/bot/fuzzers/engine_common.py
```diff
@@ -452,6 +472,10 @@ def get_all_issue_metadata(fuzz_target_path):
   if issue_owners:
     metadata['issue_owners'] = ','.join(issue_owners)
 
+  additional_issue_fields = get_additional_fields(fuzz_target_path)
+  if additional_issue_fields:
```
oliverchang, line 476: Let's do this in a way that's more extensible in the future. We can make the format of the ".issue_metadata" file be a json that looks like this:

```
{
  "additional_fields": ...,
}
```

That way, if we need to add additional metadata, we can just add to this json object without breaking the format.

oliverchang (commented):
Thanks for addressing the comments! In the future please ping the reviewers again once the PR is ready for review again.

oliverchang:
Please fix the failing tests: https://github.com/google/clusterfuzz/pull/2104/checks?check_run_id=1300439429

inferno-chromium:
tests should be easy to run, like
 ```python butler.py py_unittest -t core -m```
failure is
```
ERROR: test_get_issue_description_additional_issue_fields (tests.core.datastore.data_handler_test.with_cloud_emulators.<locals>.decorator.<locals>.Wrapped)
Test get_issue_description with a additional fields set in metadata.
  File "/workspace/src/python/tests/core/datastore/data_handler_test.py", line 378, in test_get_issue_description_additional_issue_fields
    description = data_handler.get_issue_description(testcase)
  File "/workspace/src/python/datastore/data_handler.py", line 552, in get_issue_description
    issue_metadata = json.loads(testcase.get_metadata('issue_metadata', '{}'))
  File "/usr/local/lib/python3.7/json/__init__.py", line 341, in loads
    raise TypeError(f'the JSON object must be str, bytes or bytearray, '
TypeError: the JSON object must be str, bytes or bytearray, not dict
```

ping me / @oliverchang / @hzawawy if you have any questions here.

src/python/bot/fuzzers/engine_common.py
```diff
@@ -436,6 +439,23 @@ def get_issue_components(fuzz_target_path):
...
+
+  if environment.is_trusted_host():
+    metadata_file_path = fuzzer_utils.get_file_from_untrusted_worker(
+        metadata_file_path)
+
+  if not os.path.exists(metadata_file_path):
+    return {}
+
+  with open(metadata_file_path) as handle:
+    return handle.read()
```
oliverchang, line 456: For consistency with the other return value ({}) in this function on line 453, and other get_* functions, this should return the parsed value (i.e. the dictionary).

Something like this (with extra checking so that bad formats don't propagate all the way into Testcase entities).
 
```python

with open(metadata_file_path) as handle:
  try:
    return json.load(handle)
  except (ValueError, TypeError):
    logs.log_error('Invalid metadata file format.', path=metadata_file_path)
    return {}
```

oliverchang (commented):
Nice! Just one more comment.

inferno-chromium (approved):

PR #2107, opened 2020-10-22
libClusterFuzz fixes.

oliverchang:
- Move internal imports into explicit `_internal` dir when packaged as a pip
  package to prevent clashes and accidental imports. This is guarded
  with a try/catch to fall back to importing them normally when called
  from a normal ClusterFuzz context (where the symlinks don't exist).
- Set zip_safe=False to not zip up the resulting built egg. Otherwise
  local_config does not work.
- Fix glob format to include all config files.

### DIFF
--- a/src/python/lib/clusterfuzz/fuzz/__init__.py
+++ b/src/python/lib/clusterfuzz/fuzz/__init__.py
@@ -19,10 +19,17 @@ _initialized = False
 
 
 def _initialize():
+  """Initialize the engine implementations."""
   global _initialized
 
-  from bot.fuzzers.honggfuzz import engine as honggfuzz_engine
-  from bot.fuzzers.libFuzzer import engine as libFuzzer_engine
+  try:
+    from clusterfuzz._internal.bot.fuzzers.honggfuzz \
+        import engine as honggfuzz_engine
+    from clusterfuzz._internal.bot.fuzzers.libFuzzer \
+        import engine as libFuzzer_engine
+  except:
+    from bot.fuzzers.honggfuzz import engine as honggfuzz_engine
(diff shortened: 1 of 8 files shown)


src/python/lib/clusterfuzz/fuzz/__init__.py
```diff
@@ -19,10 +19,17 @@
 
 
 def _initialize():
+  """Initialize the engine implementations."""
   global _initialized
```
inferno-chromium, line 24: Maybe add a comment next to three 3 try-except that this is temporary until internal clusterfuzz dependencies are untangled from the python package.

src/python/lib/setup.py
```diff
@@ -55,7 +55,8 @@
         'six',
     ],
     package_data={
-        'clusterfuzz': ['lib-config/**',],
+        'clusterfuzz': ['lib-config/*', 'lib-config/**/*'],
```
inferno-chromium, line 58: is 'lib-config/*' needed still ?

inferno-chromium (approved):

mbarbella-chromium (approved):

src/python/lib/setup.py
oliverchang: Yes unfortunately both are needed. lib-config/**/* only includes files that are at least 1 directory deep.

src/python/lib/clusterfuzz/fuzz/__init__.py
oliverchang: Realistically I'm not sure internal clusterfuzz dependencies can be completely removed, there are just too many things that e.g. libFuzzer engine implementation depends on. The best we can do is to hide them, as I've tried to do here.

Issue #2101, opened 2020-10-20
Move engine timeout subtraction to outside the Engine implementation. 


oliverchang:
Right now, it's very confusing exactly how long a target will run when a timeout is passed to Engine.fuzz. 

e.g. https://github.com/google/clusterfuzz/blob/20a4de8c85595566e4f3ebdfbd49bde7fb090bfd/src/python/bot/fuzzers/libFuzzer/engine.py#L247.

We should move this subtraction outside, and make the actual time for fuzzing closer to the actual value passed.

PR #2108, opened 2020-10-22
Move timeout subtraction outside Engine.fuzz.

oliverchang:
Part of #2101 

Currently it's not obvious at all to users of Engine.fuzz how long
fuzzing will go for when a max_time is passed.

Add Engine.fuzz_additional_processing_timeout so that the caller can
explicitly subtract the necessary processing time themselves if needed.

Remove engine_common.POSTPROCESSING_TIME. This was a holdout from the
launcher days. This is moved to the AFL launcher and can be removed
there as well once that's migrated to the Engine interface.

### DIFF
--- a/src/python/bot/tasks/fuzz_task.py
+++ b/src/python/bot/tasks/fuzz_task.py
@@ -1301,6 +1301,14 @@ def run_engine_fuzzer(engine_impl, target_name, sync_corpus_directory,
   options = engine_impl.prepare(sync_corpus_directory, target_path, build_dir)
 
   fuzz_test_timeout = environment.get_value('FUZZ_TEST_TIMEOUT')
+  additional_processing_time = engine_impl.fuzz_additional_processing_timeout(
+      options)
+  fuzz_test_timeout -= additional_processing_time
+  if fuzz_test_timeout <= 0:
+    raise FuzzTaskException(
+        f'Invalid engine timeout: '
+        f'{fuzz_test_timeout} - {additional_processing_time}')
+
   result = engine_impl.fuzz(target_path, options, testcase_directory,
                             fuzz_test_timeout)
 
--- a/src/python/lib/clusterfuzz/fuzz/engine.py
+++ b/src/python/lib/clusterfuzz/fuzz/engine.py
@@ -70,6 +70,19 @@ class Engine(object):
     """Get the name of the engine."""
     raise NotImplementedError
 
+  def fuzz_additional_processing_timeout(self, options):
+    """Return the maximum additional timeout in seconds for additional
+    operations in fuzz() (e.g. merging back new items).
+
+    Args:
+      options: A FuzzOptions object.
+
+    Returns:
+      An int representing the number of seconds required.
+    """
+    del options
+    return 0
+
   def prepare(self, corpus_dir, target_path, build_dir):
     """Prepare for a fuzzing session, by generating options. Returns a
     FuzzOptions object.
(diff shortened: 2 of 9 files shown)


src/python/lib/clusterfuzz/fuzz/engine.py
```diff
@@ -70,6 +70,19 @@ def name(self):
...
+    """Return the maximum additional timeout in seconds for additional
+    operations in fuzz() (e.g. merging back new items).
+
+    Args:
+      options: A FuzzOptions object.
+
+    Returns:
+      An int representing the number of seconds required.
+    """
+    del options
```
mbarbella-chromium, line 83: What is this needed for?

mbarbella-chromium (approved):

src/python/lib/clusterfuzz/fuzz/engine.py
oliverchang: This is the new recommended approach to dealing with unused arguments: 

https://google.github.io/styleguide/pyguide.html#214-decision

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1301,6 +1301,14 @@ def run_engine_fuzzer(engine_impl, target_name, sync_corpus_directory,
   options = engine_impl.prepare(sync_corpus_directory, target_path, build_dir)
 
   fuzz_test_timeout = environment.get_value('FUZZ_TEST_TIMEOUT')
+  additional_processing_time = engine_impl.fuzz_additional_processing_timeout(
+      options)
+  fuzz_test_timeout -= additional_processing_time
```
inferno-chromium, line 1306: Why do we do -(-X), why not just return positive val from this function and add it here.

src/python/tests/core/bot/fuzzers/libFuzzer/engine_test.py
```diff
@@ -783,7 +812,7 @@ def mocked_create_merge_directory(_):
     engine_impl = engine.LibFuzzerEngine()
     options = engine_impl.prepare(corpus_path, target_path, DATA_DIR)
     options.arguments.append('-runs=10')
-    engine_impl.fuzz(target_path, options, TEMP_DIR, 10)
+    engine_impl.fuzz(target_path, options, TEMP_DIR, get_fuzz_timeout(5))
```
inferno-chromium, line 815: This get_fuzz_timeout call is confusing, isnt it this signature ?
src/python/bot/fuzzers/libfuzzer.py:def get_fuzz_timeout(is_mutations_run, total_timeout=None):

inferno-chromium (approved):

src/python/tests/core/bot/fuzzers/libFuzzer/engine_test.py
oliverchang: This is a different get_fuzz_timeout, local to this file: https://github.com/google/clusterfuzz/blob/dc099ac6e2eb2e170803a21640acb70b29826580/src/python/tests/core/bot/fuzzers/libFuzzer/engine_test.py#L441

src/python/bot/tasks/fuzz_task.py
oliverchang: This is a positive value. We negate it in the implementation of libFuzzer's fuzz_additional_processing_timeout because libfuzzer.get_fuzz_timeout() returned a negative value. 

We can clean up libfuzzer.get_fuzz_timeout to return a positive value in the first place, but that needs to wait for further refactoring (including AFL->engien refactor) so that we can get rid of some env var overrides used for testing  (e.g. HARD_TIMEOUT_OVERRIDE).

inferno-chromium: ok sounds good.

PR #1901, opened 2020-07-13
Added fuzzer-to-job configurability on /jobs page

iamarshsingh:
Hi, This PR is to add Fuzzer Selection option to /jobs page. The relationship defined between the jobs and fuzzer is many-to-many and the ground truth for the relationships is considered to be Fuzzer.jobs with automatic updation to JobFuzzer as well. This PR will fixes #596. 

Screenshot of the location of the menu:
Screenshot 2020-07-13 at 4 47 09 PM

Screenshot of the drop-down:
Screenshot 2020-07-13 at 4 44 39 PM

Please have a look and let me know if there are any changes required!

PR #2109, opened 2020-10-23
[docs] Describe how to enable a fuzzer for a job on the Job page.

Dor1s:
This makes the doc and the process simpler. The feature was added in https://github.com/google/clusterfuzz/pull/1901

### DIFF
--- a/docs/setting-up-fuzzing/libfuzzer_and_afl.md
+++ b/docs/setting-up-fuzzing/libfuzzer_and_afl.md
@@ -100,23 +100,23 @@ To create a job for libFuzzer or AFL:
 1. Navigate to the *Jobs* page.
 2. Go to the "ADD NEW JOB" form.
 3. Fill out the "Name" and "Platform" (LINUX).
-  1. If setting up an **AFL** job, use the templates **"afl"** and
-     **"engine_asan"**.
-  2. If setting up a **libFuzzer** job, use the templates **"libfuzzer"** and
-     **"engine_$SANITIZER"** depending on which sanitizer you are using (e.g.
-     **"engine_asan"**).
-4. Select your build (your zip containing the fuzz target binary) to upload as a
+4. Enable the desired fuzzer in the "Select/modify fuzzers" field, e.g. **libFuzzer** or **afl**.
+5. If setting up an **AFL** job, use the templates **"afl"** and
+   **"engine_asan"**.
+6. If setting up a **libFuzzer** job, use the templates **"libfuzzer"** and
+   **"engine_$SANITIZER"** depending on which sanitizer you are using (e.g.
+   **"engine_asan"**).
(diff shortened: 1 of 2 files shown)


docs/setting-up-fuzzing/libfuzzer_and_afl.md
```diff
@@ -100,23 +100,23 @@ To create a job for libFuzzer or AFL:
...
 2. Go to the "ADD NEW JOB" form.
 3. Fill out the "Name" and "Platform" (LINUX).
-  1. If setting up an **AFL** job, use the templates **"afl"** and
-     **"engine_asan"**.
-  2. If setting up a **libFuzzer** job, use the templates **"libfuzzer"** and
-     **"engine_$SANITIZER"** depending on which sanitizer you are using (e.g.
-     **"engine_asan"**).
-4. Select your build (your zip containing the fuzz target binary) to upload as a
+4. Enable the desired fuzzer in the "Select/modify fuzzers" field, e.g. **libFuzzer** or **afl**.
+5. If setting up an **AFL** job, use the templates **"afl"** and
```
Dor1s, line 104: not sure why we had the indentation here, it is rendered as a straight list anyway: https://google.github.io/clusterfuzz/setting-up-fuzzing/libfuzzer-and-afl/#creating-a-job-type

docs/setting-up-fuzzing/libfuzzer_and_afl.md
```diff
@@ -100,23 +100,23 @@ To create a job for libFuzzer or AFL:
...
-
-Next we must let ClusterFuzz know which fuzzer the job can be used with:
-1. Navigate to the *Fuzzers* page.
-2. Click "EDIT" for the desired fuzzer (afl or libFuzzer).
-3. Click "Select/modify jobs".
-4. Mark the desired job.
-5. Click "SUBMIT".
+8. Use the "ADD" button to add the job to ClusterFuzz.
+
+You can also control which fuzzer the job can be used with using the *Fuzzers* page:
```
oliverchang, line 115: I would actually just remove this completely to keep the documentation lean. I think it's actually quite uncommon to want to modify associations from the fuzzer page instead.

oliverchang (approved):
thanks!

docs/setting-up-fuzzing/libfuzzer_and_afl.md
inferno-chromium: I am fine with either. Can we expand this page to include honggfuzz as well.

inferno-chromium (approved):

docs/setting-up-fuzzing/libfuzzer_and_afl.md
Dor1s: cool, I was also thinking maybe we should remove it -- done!

also added a step for honggfuzz

PR #2110, opened 2020-10-25
Refactor the AFL launcher to share code with the new engine interface.

mbarbella-chromium:
PTAL. I plan on landing this in 3 parts:
1) This change, which I want to ensure doesn't break the existing AFL code before anything else. I include a basic skeleton of how things will be used by the new engine interface here to make it clearer why the refactoring is necessary, but if it doesn't seem clean to land it in this form I'm happy to leave it out to part 2.
2) A change including the rest of the engine interface implementation, tests, and which enables the new pipeline selectively.
3) A change which removes the old AFL implementation, removes obsolete code, and merges what we do need from launcher.py into engine.py. At this point we'll enable the new AFL implementation unconditionally.


src/python/bot/fuzzers/afl/engine.py
```diff
@@ -0,0 +1,109 @@
...
+      corpus_dir: The main corpus directory.
+      target_path: Path to the target.
+      build_dir: Path to the build directory.
+
+    Returns:
+      A FuzzOptions object.
+    """
+    afl_config = launcher.AFLConfig.from_target_path(target_path)
+    arguments = afl_config.additional_afl_arguments
+    strategies = launcher.FuzzingStrategies(target_path).to_strategy_dict()
```
oliverchang, line 43: currently i think mutations using radamsa etc are done in the fuzz() method of the runner. This should be moved to prepare() here as well -- maybe add a TODO?

src/python/bot/fuzzers/afl/launcher.py
```diff
@@ -256,29 +254,60 @@ class FuzzingStrategies(object):
...
+  # to_strategy_dict function should be removed when everything is fully
+  # converted to the engine pipeline. For now this allows the code to be shared
+  # between both cases while still adhering to the new pipeline's API.
+  def __init__(self, target_path, strategy_dict=None):
+    self.corpus_subset_size = None
+
+    # If we have already generated a strategy dict, use that in favor of
+    # creating a new pool and picking randomly.
+    if strategy_dict:
+      self.candidate_generator = strategy_dict['candidate_generator']
```
oliverchang, line 267: should we use .get() here, or are these always expected to exist?

src/python/bot/fuzzers/afl/launcher.py
```diff
@@ -256,29 +254,60 @@ class FuzzingStrategies(object):
...
+      strategy_pool = strategy_selection.generate_weighted_strategy_pool(
+          strategy_list=strategy.AFL_STRATEGY_LIST,
+          use_generator=True,
+          engine_name='afl')
+
+      # Select a generator to attempt to use for existing testcase mutations.
+      self.candidate_generator = engine_common.select_generator(
+          strategy_pool, target_path)
+
+      self.use_corpus_subset = strategy_pool.do_strategy(
```
oliverchang, line 283: while we're here, might be worth thinking if we can move the corpus subset parts (and possibly even mutations) into a common place in fuzz_task rather than requiring each engine to implement its own logic.

oliverchang (approved):
nice! LGTM with some suggestions for future PRs

src/python/bot/fuzzers/afl/engine.py
```diff
@@ -0,0 +1,109 @@
+# Copyright 2019 Google LLC
```
jonathanmetzman, line 1: nit: 2020

src/python/bot/fuzzers/afl/launcher.py
jonathanmetzman: +1

jonathanmetzman (approved):
LGTM

src/python/bot/fuzzers/afl/engine.py
mbarbella-chromium: Done.

src/python/bot/fuzzers/afl/engine.py
mbarbella-chromium: Good point. Added a TODO.

src/python/bot/fuzzers/afl/launcher.py
mbarbella-chromium: This will always exist. When we no longer need to support both pipelines I plan on simplifying this workflow a bit, so it will only matter in the interim anyway.

src/python/bot/fuzzers/afl/launcher.py
mbarbella-chromium: Good idea. I filed #2112 to track this.

I definitely think the mutation strategies are worth moving as well. It would be nice to get all of this for free for each new engine we implement.

mbarbella-chromium (commented):
Thanks for the review! Comments addressed.

PR #2122, opened 2020-11-02
Use python3 in libClusterFuzz build.sh,add wheel

inferno-chromium:


inferno-chromium:
This also need pip3 install wheel==0.35.1, should we be installing this in build.sh ? See https://github.com/google/fuzzbench/pull/929

oliverchang:
I think this currently works from `pipenv shell`? Were you seeing these issues even within pipenv?

inferno-chromium:
> I think this currently works from `pipenv shell`? Were you seeing these issues even within pipenv?

I am thinking more of an external user perspective like fuzzbench or any other user who wants to use this lib.

oliverchang:
> > I think this currently works from `pipenv shell`? Were you seeing these issues even within pipenv?
> 
> I am thinking more of an external user perspective like fuzzbench or any other user who wants to use this lib.

Ah right, I think for external users, we will publish onto pypi instead so users can just do `pip install clusterfuzz`. I can see if I can build/do this now.

inferno-chromium:
> > > I think this currently works from `pipenv shell`? Were you seeing these issues even within pipenv?
> > 
> > 
> > I am thinking more of an external user perspective like fuzzbench or any other user who wants to use this lib.
> 
> Ah right, I think for external users, we will publish onto pypi instead so users can just do `pip install clusterfuzz`. I can see if I can build/do this now.

Sounds good, i will wait on it then. can go after tmrw too since i know you are busy.

oliverchang:
Initially upload is done :) check https://pypi.org/project/clusterfuzz/

`pip install clusterfuzz` should now work.

PR #2131, opened 2020-11-05
Reproduce tool fixes.

mbarbella-chromium:
PTAL. If the installed version of distutils is the Python 3.8 version, this seems to cause issues if we attempt to run under Python 3.7. Switching the order of this check appears to work fine.

### DIFF
--- a/local/install_deps_linux.bash
+++ b/local/install_deps_linux.bash
@@ -28,12 +28,12 @@ while [ "$1" != "" ]; do
 done
 
 # Check that python 3.7 or 3.8 is installed.
-if python3.7 --help > /dev/null; then
-  PYTHON='python3.7'
-  PYTHON_VERSION='3.7'
-elif python3.8 --help > /dev/null; then
+if python3.8 --help > /dev/null; then
   PYTHON='python3.8'
   PYTHON_VERSION='3.8'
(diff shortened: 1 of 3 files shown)


inferno-chromium:
i think this will break deployment since we still need to deploy as 3.7, need to discuss a plan with oliver.

mbarbella-chromium:
Ah, good catch. I'll update this to favor 3.8 in the reproduce tool case only.

mbarbella-chromium:
Though if it's not much work to support 3.8 in deployment, maybe it's worth considering. I'll hold off until we hear from Oliver.

oliverchang:
Supporting 3.8 deployment will be a bit of work and needs more thought. It's difficult to upgrade some of our bots (internal chromium ones etc) so the proper fix here would be to fix our deployment so that we build multiple packages for 3.7, 3.8 (and 3.9 and whatever comes after...) as these aren't backwards compatible.  Let's just go with the quick fix for reproduce tool first.

mbarbella-chromium:
SGTM.

mbarbella-chromium:
Made a few changes here and added another batch of fixes. PTAL.

inferno-chromium (approved):

Issue #2137, opened 2020-11-09
GradiantFuzz training failures


inferno-chromium:
```
2020-11-09 05:42:04.633057: I tensorflow/stream_executor/platform/default/dso_loader.cc:48] Successfully opened dynamic library libcudart.so.10.1
2020-11-09 05:42:05.752552: I tensorflow/stream_executor/platform/default/dso_loader.cc:48] Successfully opened dynamic library libcuda.so.1
2020-11-09 05:42:06.056471: I tensorflow/stream_executor/cuda/cuda_gpu_executor.cc:982] successful NUMA node read from SysFS had negative value (-1), but there must be at least one NUMA node, so returning NUMA node zero
2020-11-09 05:42:06.057069: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1716] Found device 0 with properties: 
pciBusID: 0000:00:04.0 name: Tesla P100-PCIE-16GB computeCapability: 6.0
coreClock: 1.3285GHz coreCount: 56 deviceMemorySize: 15.90GiB deviceMemoryBandwidth: 681.88GiB/s
2020-11-09 05:42:06.057110: I tensorflow/stream_executor/platform/default/dso_loader.cc:48] Successfully opened dynamic library libcudart.so.10.1
2020-11-09 05:42:06.058975: I tensorflow/stream_executor/platform/default/dso_loader.cc:48] Successfully opened dynamic library libcublas.so.10
2020-11-09 05:42:06.060802: I tensorflow/stream_executor/platform/default/dso_loader.cc:48] Successfully opened dynamic library libcufft.so.10
2020-11-09 05:42:06.061175: I tensorflow/stream_executor/platform/default/dso_loader.cc:48] Successfully opened dynamic library libcurand.so.10
2020-11-09 05:42:06.063109: I tensorflow/stream_executor/platform/default/dso_loader.cc:48] Successfully opened dynamic library libcusolver.so.10
2020-11-09 05:42:06.064219: I tensorflow/stream_executor/platform/default/dso_loader.cc:48] Successfully opened dynamic library libcusparse.so.10
2020-11-09 05:42:06.068539: I tensorflow/stream_executor/platform/default/dso_loader.cc:48] Successfully opened dynamic library libcudnn.so.7
[...]

PR #2139, opened 2020-11-09
Speculative fix for gradientfuzz training errors (issue #2137).

mbarbella-chromium:
PTAL. If this works I think a follow-up fix is needed, but want to test it before making more significant changes.

### DIFF
--- a/src/python/bot/fuzzers/ml/gradientfuzz/model_utils.py
+++ b/src/python/bot/fuzzers/ml/gradientfuzz/model_utils.py
@@ -38,7 +38,7 @@ class BitmapAcc(keras.metrics.Metric):
     """
 
   def __init__(self, name='bitmap_acc', **kwargs):
-    super(BitmapAcc, self).__init__(name=name, **kwargs)
+    super().__init__(name=name, **kwargs)
     self.total_correct = self.add_weight(
         name='total_correct', initializer='zeros')
     self.total_branches = self.add_weight(
@@ -84,7 +84,7 @@ class NeuzzJaccardAcc(keras.metrics.Metric):
     """
 
   def __init__(self, name='neuzz_jaccard_acc', **kwargs):
-    super(NeuzzJaccardAcc, self).__init__(name=name, **kwargs)
+    super().__init__(name=name, **kwargs)
     self.true_positives = self.add_weight(
         name='true_positives', initializer='zeros')
     self.total_errors = self.add_weight(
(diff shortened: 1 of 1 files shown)


src/python/bot/fuzzers/ml/gradientfuzz/model_utils.py
```diff
@@ -51,6 +51,9 @@ def update_state(self, *args, **kwargs):
       Computes (total matches) / (total branches).
       """
     y_true, y_pred = args
+    # TODO(mbarbella): Verify that this works properly and refactor this file
+    # for consistency between types if so.
```
inferno-chromium, line 55: Maybe update the other update_state function in this file as well. can remove the todo as can refactor anyway if needed.

inferno-chromium (approved):
LGTM with lets fix the other call site as well.

src/python/bot/fuzzers/ml/gradientfuzz/model_utils.py
mbarbella-chromium: Done!

inferno-chromium (approved):
LGTM

mbarbella-chromium:
I'm able to repro the error that CI was hitting before locally, I expect it is related to this change. Still investigating.

mbarbella-chromium (commented):
Ended up handling this differently in the two separate update_state functions since the second is working with bools. Both now explicitly match types.

src/python/bot/fuzzers/ml/gradientfuzz/model_utils.py
```diff
@@ -95,6 +96,7 @@ def update_state(self, *args, **kwargs):
       Computes (true positives) / (true positives + errors).
       """
     y_true, y_pred = args
+    y_true = tf.cast(y_true, tf.bool)
     y_pred = tf.cast(tf.round(y_pred), tf.bool)
```
inferno-chromium, line 100: Is this one supposed to casted to bool ?

mbarbella-chromium: Yeah, since it's used below in logical_and.

mbarbella-chromium: I should also point out I don't think the cast above is actually doing anything (and it may be cleaner not to check) but I don't mind having the explicit cast while we're getting stuff working initially.

PR #2141, opened 2020-11-10
Add seahawk target and fix a minor logger issue

stplaydog:


src/python/platforms/android/settings.py
```diff
@@ -179,9 +179,9 @@ def is_google_device():
 
 def is_automotive():
   """Returns if we are running in Android Automotive OS, currently only for
-  Osprey."""
+  Osprey or Seahawk."""
   product_model = get_product_model()
-  return product_model == 'Osprey'
+  return product_model == 'Osprey' or product_model == 'Seahawk'
```
oliverchang, line 184: Please fix the lint failure:

```
src/python/platforms/android/settings.py:184:9: R1714: Consider merging these comparisons with "in" to "product_model in ('Osprey', 'Seahawk')" (consider-using-in)
```

inferno-chromium: +1.

src/python/platforms/android/logger.py
```diff
@@ -22,8 +23,10 @@
 
 def clear_log():
   """Clear log."""
+  adb.run_as_root()
   adb.run_shell_command(['stop', 'logd'])
   adb.run_shell_command(['start', 'logd'])
+  time.sleep(1)
```
inferno-chromium, line 29: Can you check if anything smaller is possible here like 0.5 or 0.1/0.2 in a reliable way ?

inferno-chromium (changes requested):
LGTM with slight changes needed.

src/python/platforms/android/settings.py
stplaydog: done

src/python/platforms/android/logger.py
stplaydog: 0.1 works fine, I've ran it for hours and no warnings in logs.

PR #2144, opened 2020-11-12
Remove unmaintainable gradientfuzz code.

inferno-chromium:


src/python/tests/appengine/handlers/cron/ml_train_test.py
```diff
@@ -48,21 +48,11 @@ def setUp(self):
     data_types.Job(name='libfuzzer_msan', environment_string='').put()
```
mbarbella-chromium, line 48: Some of the cases here are still potentially interesting if we end up adding more ML engines later. It might be worth modifying MODEL_NAME_TO_TASK in the test so that we're still covering the multiple model case.

mbarbella-chromium (approved):
LGTM with a nit.

src/python/tests/appengine/handlers/cron/ml_train_test.py
inferno-chromium: i think we need another train_*task for this, that is why skipping for now.

PR #2146, opened 2020-11-13
Don't replace LKL crashes with ABRT crashes.

mspectorgoogle:
The libfuzzer portion of the LKL stack may contain an ABRT don't replace
the crash_type similar to other aborts.

### DIFF
--- a/src/python/crash_analysis/crash_analyzer.py
+++ b/src/python/crash_analysis/crash_analyzer.py
@@ -344,11 +344,14 @@ def is_security_issue(crash_stacktrace, crash_type, crash_address):
   if crash_type == 'Unexpected-exit':
     return False
 
+  # Kernel Failures are security bugs
+  if crash_type.startswith('Kernel failure'):
+    return True
+
   # No crash type, can't process.
   if not crash_type:
     return False
 
-  # Check signal types, some of them indicate not a security bug.
   if has_signal_for_non_security_bug_type(crash_stacktrace):
     return False
 
--- a/src/python/lib/clusterfuzz/stacktraces/__init__.py
+++ b/src/python/lib/clusterfuzz/stacktraces/__init__.py
@@ -673,7 +673,10 @@ class StackParser:
         continue
 
       # Sanitizer regular crash (includes ills, abrt, etc).
-      if not state.found_golang_crash and not state.found_python_crash:
+      # However if we have a Go, Python, or Kernel crash, don't replace
+      # this with sanitizer signal handler failure.
+      if (not state.crash_type.startswith('Kernel failure') and
+          not state.found_golang_crash and not state.found_python_crash):
         self.update_state_on_match(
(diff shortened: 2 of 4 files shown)


src/python/crash_analysis/crash_analyzer.py
```diff
@@ -348,8 +349,11 @@ def is_security_issue(crash_stacktrace, crash_type, crash_address):
   if not crash_type:
     return False
 
+  is_lkl = lkl_constants.LINUX_KERNEL_MODULE_STACK_TRACE in crash_stacktrace
   # Check signal types, some of them indicate not a security bug.
-  if has_signal_for_non_security_bug_type(crash_stacktrace):
+  # LKL stacks may have ABRTs and gsignal etc in them on the libfuzzer side,
+  # This should not affect whether they are marked security.
+  if not is_lkl and has_signal_for_non_security_bug_type(crash_stacktrace):
```
oliverchang, line 356: this makes the logic a bit difficult to follow, and couples this code to lkl which is not ideal. We should just be relying on the crash_type and other testcase information here. 

What's the crash type for these bugs? Is it 'Kernel failure' ? Can you just add a crash_type check above to return True?

src/python/lib/clusterfuzz/stacktraces/__init__.py
```diff
@@ -673,7 +673,10 @@ def parse(self, stacktrace: str) -> CrashInfo:
         continue
 
       # Sanitizer regular crash (includes ills, abrt, etc).
-      if not state.found_golang_crash and not state.found_python_crash:
+      # However if we already have a kernel crash, we don't want to
+      # replace it with the ASSERT.
```
oliverchang, line 677: This comment seems a bit confusing. This isn't replacing anything with 'ASSERT'? 

Also can make this comment more general. Something along the lines of: "If we have a Go, Python, or Kernel crash, don't replace this with sanitizer signal handler failures."

oliverchang (changes requested):

src/python/crash_analysis/crash_analyzer.py
```diff
@@ -348,8 +348,10 @@ def is_security_issue(crash_stacktrace, crash_type, crash_address):
   if not crash_type:
     return False
 
-  # Check signal types, some of them indicate not a security bug.
-  if has_signal_for_non_security_bug_type(crash_stacktrace):
+  # Check signal types for userspace crashes, some of them indicate
+  # not a security bug.
+  if (not crash_type.startswith('Kernel failure') and
```
oliverchang, line 353: Can we just separate this out into a 
```python
# Assume kernel crashes are security bugs.
if crash_type.startswith('Kernel failure'):
  return True
```

above?

oliverchang (approved):

oliverchang (approved):

inferno-chromium (approved):

PR #2149, opened 2020-11-19
Relax Python 3.7 requirement for local development.

oliverchang:
Allow any Python 3 local dev environment as long as it's >=3.7.

For deployment, add a check for 3.7 to ensure it can only occur from
there.


oliverchang:
This is a shorter term solution to let people do local development on >3.7. For deployment our CI images are still on 3.7 so this should still work.

inferno-chromium (approved):

mbarbella-chromium (approved):

PR #2155, opened 2020-11-25
Remove incorrect options quoting on Windows.

inferno-chromium:
Also, remove unneeded quoting for posix.
See https://bugs.chromium.org/p/chromium/issues/detail?id=1152549
Keep ' ' and ':' in _quote_value_if_needed as ':' parsing fails
when decoding options value. Use ' ' as it is needed for both
linux and windows.

### DIFF
--- a/src/python/system/environment.py
+++ b/src/python/system/environment.py
@@ -22,11 +22,6 @@ import socket
 import sys
 import yaml
 
-try:
-  from shlex import quote
-except ImportError:
-  from pipes import quote
-
 # Tools supporting customization of options via ADDITIONAL_{TOOL_NAME}_OPTIONS.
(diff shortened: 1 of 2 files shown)


src/python/system/environment.py
```diff
@@ -111,9 +106,7 @@ def _parse_memory_tool_options(options_str):
 def _quote_value_if_needed(value):
   """Quote environment value as needed for certain platforms like Windows."""
   result = value
-
-  bot_platform = platform()
-  if bot_platform == 'WINDOWS':
+  if ' ' in result or ':' in result:
```
oliverchang, line 109: What about other whitespace (e.g. tabs)?

oliverchang (approved):

src/python/system/environment.py
inferno-chromium: We only run this on paths (symbolizer path) and it shouldnt have anything other than spaces. unsure if adding regex is worth it here.

PR #2156, opened 2020-11-26
Set <meta charset="utf-8"> on all pages.

oliverchang:
Pointed out in https://bugs.chromium.org/p/chromium/issues/detail?id=920160#c7

### DIFF
--- a/src/appengine/private/templates/layout.html
+++ b/src/appengine/private/templates/layout.html
@@ -24,6 +24,7 @@
     <link rel="preconnect" href="https://fonts.gstatic.com/" crossorigin>
     <meta name="msapplication-config" content="/favicon/browserconfig.xml?v=A0RYylnML6">
     <meta name="theme-color" content="#ffffff">
+    <meta charset="utf-8">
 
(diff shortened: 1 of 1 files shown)


src/appengine/private/templates/layout.html
```diff
@@ -24,6 +24,7 @@
     <link rel="preconnect" href="https://fonts.gstatic.com/" crossorigin>
     <meta name="msapplication-config" content="/favicon/browserconfig.xml?v=A0RYylnML6">
     <meta name="theme-color" content="#ffffff">
+    <meta charset="utf-8">
```
mathiasbynens, line 27: FWIW, the best place to do this is as early as possible in the `<head>`, since `<meta charset>` must appear in the first 1024 bytes. See https://github.com/h5bp/html5-boilerplate/blob/master/dist/doc/html.md#the-order-of-the-title-and-meta-tags.

I had submitted a patch before I had seen yours: https://github.com/google/clusterfuzz/pull/2158 Feel free to either merge that one instead, or close it.

PR #2157, opened 2020-11-26
Add DOCTYPE to force standards mode rendering

mathiasbynens:
This avoids rendering the page in quirks mode.

https://quirks.spec.whatwg.org/


oliverchang (approved):
Thanks!

PR #2158, opened 2020-11-26
Specify the HTML document’s encoding

mathiasbynens:
This avoids rendering non-ASCII symbols as mojibake.

Fixes: https://bugs.chromium.org/p/chromium/issues/detail?id=920160


oliverchang:
Thanks!

PR #2159, opened 2020-11-26
Install python3-pip

zetafunction:
Later steps depend on this package.

### DIFF
--- a/local/install_deps_linux.bash
+++ b/local/install_deps_linux.bash
@@ -77,6 +77,7 @@ sudo apt-get update
 sudo apt-get install -y \
     blackbox \
     curl \
+    python3-pip \
     unzip \
     xvfb
 


local/install_deps_linux.bash
```diff
@@ -160,6 +161,7 @@ fi
 # Setup pipenv and install python dependencies.
 $PYTHON -m pip install --user pipenv
 $PYTHON -m pipenv --python $PYTHON
+$PYTHON -m pipenv lock
```
oliverchang, line 164: Is this step needed? We already have a Pipfile.lock file in this repo.

zetafunction: You are right; this isn't needed due to https://github.com/google/clusterfuzz/commit/59de121ad4465f15e19a47ae9e58973af12ca367, but I missed that because reproduce-tool-stable had not been updated yet.

inferno-chromium:
@mbarbella-chromium - can you do a reproduce tool stable release (also add process to release doc).

mbarbella-chromium:
Updated the reproduce-tool-stable tag. Thanks for the fix!

PR #2162, opened 2020-12-01
cleanup: Close issues for testcases with fixed == 'NA'.

### LINKED ISSUE #2113: Testcase can be untriaged indefinitely if fixed == 'NA' and a bug is attached.
A testcase can have its `fixed` status marked as 'NA' and `open = False` in some cases.

e.g. https://github.com/google/clusterfuzz/blob/537b73cdc5aec7320af39562badebfe7c9b470c2/src/python/bot/tasks/progression_task.py#L399

If a bug is attached, that will never be closed in our current cleanup task logic: https://github.com/google/clusterfuzz/blob/8aa74e1b932b870792e095136ce14a20f2b8fee4/src/appengine/handlers/cron/cleanup.py#L435

`triaged` will also be False indefinitely as the bug remains open.

oliverchang:
Fixes #2113.


inferno-chromium (approved):
Nice!

PR #2163, opened 2020-12-02
Determine trials for every single test run

mi-ac:
Determining trials is moved to the main test loop of blackbox fuzzing. The trial configuration from DB is set up once in the beginning of the loop. The set of additional flags is then re-determined for every test run.

It's not needed anymore to bail out for engine fuzzing, as we now only use this logic in blackbox fuzzing.

### DIFF
--- a/src/python/bot/tasks/fuzz_task.py
+++ b/src/python/bot/tasks/fuzz_task.py
@@ -1732,6 +1732,9 @@ class FuzzingSession(object):
       testcases_metadata[testcase_file_path]['gestures'] = pick_gestures(
           test_timeout)
 
+    # Prepare selecting trials in main loop below.
+    trial_selector = trials.Trials()
+
     logs.log('Starting to process testcases.')
     logs.log('Redzone is %d bytes.' % self.redzone)
     logs.log('Timeout multiplier is %s.' % str(self.timeout_multiplier))
@@ -1755,6 +1758,11 @@ class FuzzingSession(object):
         gestures = testcases_metadata[testcase_file_path]['gestures']
 
         env_copy = environment.copy()
+
+        # For some binaries, we specify trials, which are sets of flags that we
+        # only apply some of the time. Adjust APP_ARGS for them if needed.
+        trial_selector.setup_additional_args_for_app(env_copy)
+
         thread = process_handler.get_process()(
             target=testcase_manager.run_testcase_and_return_result_in_queue,
             args=(temp_queue, thread_index, testcase_file_path, gestures,
@@ -1895,10 +1903,6 @@ class FuzzingSession(object):
       logs.log_error(
           'Unable to setup data bundle %s.' % self.fuzzer.data_bundle_name)
       return
-
-    # For some binaries, we specify trials, which are sets of flags that we only
-    # apply some of the time. Adjust APP_ARGS for them if needed.
-    trials.setup_additional_args_for_app()
 
     # TODO(mbarbella): Remove the environment variable check when the old
     # blackbox pipeline has been converted to the engine interface.
--- a/src/python/bot/tasks/trials.py
+++ b/src/python/bot/tasks/trials.py
@@ -20,34 +20,38 @@ from datastore import data_types
 from system import environment
 
 
-def setup_additional_args_for_app():
-  """Select additional args for the specified app at random."""
-  if environment.is_engine_fuzzer_job():
-    # Not applicable to engine fuzzers.
-    return
+class Trials:
+  """Helper class for selecting app-specific extra flags."""
 
-  app_name = environment.get_value('APP_NAME')
-  if not app_name:
-    return
+  def __init__(self):
+    self.trials = []
 
-  # Convert the app_name to lowercase. Case may vary by platform.
-  app_name = app_name.lower()
+    app_name = environment.get_value('APP_NAME')
+    if not app_name:
+      return
 
-  # Hack: strip file extensions that may be appended on various platforms.
-  extensions_to_strip = ['.exe', '.apk']
-  for extension in extensions_to_strip:
-    app_name = utils.strip_from_right(app_name, extension)
+    # Convert the app_name to lowercase. Case may vary by platform.
+    app_name = app_name.lower()
 
-  trials = data_types.Trial.query(data_types.Trial.app_name == app_name)
-  trials = [trial for trial in trials if random.random() < trial.probability]
-  if not trials:
-    return
+    # Hack: strip file extensions that may be appended on various platforms.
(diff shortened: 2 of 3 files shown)


inferno-chromium:
Can you please do python butler.py format and then verify lint failures are fixed with python butler.py lint.

src/python/tests/core/bot/tasks/trials_test.py
```diff
@@ -44,45 +44,52 @@ def test_no_effect_on_no_match(self):
...
+    trial_selector.setup_additional_args_for_app(environment)
     self.assertEqual(environment.get_value('APP_ARGS'), '-x --a1')
     self.assertEqual(environment.get_value('TRIAL_APP_ARGS'), '--a1')
 
   def test_trial_not_selected(self):
-    """Ensure that no additional flags a trial was not selected."""
+    """Ensure no additional flags if a trial was not selected."""
     self.mock.random.return_value = 0.5
     environment.set_value('APP_NAME', 'app_2')
+    trial_selector = trials.Trials(environment)
```
mi-ac, line 65: Here it seems there was a line missing before that calls trials.setup_additional_args_for_app()

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1732,6 +1732,9 @@ def do_blackbox_fuzzing(self, fuzzer, fuzzer_directory, job_type):
       testcases_metadata[testcase_file_path]['gestures'] = pick_gestures(
           test_timeout)
 
+    # Prepare selecting trials in main loop below.
+    trial_selector = trials.Trials(environment)
```
inferno-chromium, line 1736: Remove environment module from argument, it should be added to import in trials.py

src/python/bot/tasks/trials.py
```diff
@@ -17,37 +17,38 @@
...
-  """Select additional args for the specified app at random."""
-  if environment.is_engine_fuzzer_job():
-    # Not applicable to engine fuzzers.
-    return
+class Trials:
 
-  app_name = environment.get_value('APP_NAME')
-  if not app_name:
-    return
+  def __init__(self, environment):
```
inferno-chromium, line 24: Remove environment arg, add "from system import environment" in this file.

inferno-chromium (approved):
LGTM with minor fixes needed.

src/python/tests/core/bot/tasks/trials_test.py
inferno-chromium: Yes, thanks for fixing that.

src/python/bot/tasks/fuzz_task.py
mi-ac: Done

src/python/bot/tasks/trials.py
mi-ac: Done. But now I also renamed the environment used in the method below as there it needs to be a separate copied test environment now. Calling it also "environment" reduces readability...

src/python/bot/tasks/fuzz_task.py
```diff
@@ -1755,6 +1758,11 @@ def do_blackbox_fuzzing(self, fuzzer, fuzzer_directory, job_type):
         gestures = testcases_metadata[testcase_file_path]['gestures']
 
         env_copy = environment.copy()
+
+        # For some binaries, we specify trials, which are sets of flags that we
+        # only apply some of the time. Adjust APP_ARGS for them if needed.
+        trial_selector.setup_additional_args_for_app(env_copy)
```
inferno-chromium, line 1764: This seems wrong, i should have looked closely before. Trials are done once per fuzzing session and not every testcase, why do you do this?

inferno-chromium (changes requested):

src/python/bot/tasks/fuzz_task.py
mi-ac: That's exactly what I'm trying to change here. See change description. I've also shared a doc with you with more discussion about this.

src/python/tests/core/bot/tasks/trials_test.py
```diff
@@ -44,45 +44,52 @@ def test_no_effect_on_no_match(self):
     """Ensure that no additional flags are added if a binary has no trials."""
     self.mock.random.return_value = 0.0
     environment.set_value('APP_NAME', 'app_0')
-    trials.setup_additional_args_for_app()
+    trial_selector = trials.Trials()
+    trial_selector.setup_additional_args_for_app(environment)
```
inferno-chromium, line 48: This needs to be environment.copy(), or os.environ
Using module is wrong here. In fuzz_task, it is environment.copy() which is a dict.

src/python/bot/tasks/trials.py
```diff
@@ -20,34 +20,37 @@
...
+    trial_args = [
+        trial.app_args
+        for trial in self.trials
+        if random.random() < trial.probability
+    ]
+    if not trial_args:
+      return
+
+    trial_app_args = ' '.join(trial_args)
+    app_args = test_env.get_value('APP_ARGS', '')
```
inferno-chromium, line 54: this needs to get and set in the next command. test_env is a dict and not an environment module.

inferno-chromium (changes requested):

src/python/tests/core/bot/tasks/trials_test.py
mi-ac: Good catch! Changed it now to resemble the code in fuzz_task. Now it also caught the bug below.

Wonder how hard it would be to add a test for fuzz_task which would have caught this... I'll experiment with this separately.

src/python/bot/tasks/trials.py
mi-ac: Done. The improved test now also found this.

src/python/bot/tasks/trials.py
```diff
@@ -20,34 +20,37 @@
...
+
+    # Convert the app_name to lowercase. Case may vary by platform.
+    app_name = app_name.lower()
+
+    # Hack: strip file extensions that may be appended on various platforms.
+    extensions_to_strip = ['.exe', '.apk']
+    for extension in extensions_to_strip:
+      app_name = utils.strip_from_right(app_name, extension)
+
+    self.trials = data_types.Trial.query(data_types.Trial.app_name == app_name)
```
inferno-chromium, line 41: add list() around "data_types.Trial.query(data_types.Trial.app_name == app_name)" so that it computes the query result.

inferno-chromium (approved):
LGTM

src/python/bot/tasks/trials.py
mi-ac: Done

PR #2164, opened 2020-12-07
Android fixes

inferno-chromium:
- Don't keep rebooting android device on constant task leases.
- Add task retry wait when a particular device family is not available.


inferno-chromium:
@mspectorgoogle fyi, this fixes task leasing issues, but please file ticket with android lab to get devices back [also some bots are still android r instead of master]

oliverchang (approved):

PR #2165, opened 2020-12-07
Allow privileged testcase users to set security flag.

oliverchang:
Add a dialog to set severity as well.


jonathanmetzman (approved):
LGTM. Thanks!
Does this affect the bug tracker at all?

oliverchang:
> LGTM. Thanks!
> Does this affect the bug tracker at all?

Nope, that will come in another PR :)

PR #2167, opened 2020-12-09
Enable the new engine pipeline for AFL.

mbarbella-chromium:
This seems to be working fine. Fuzzing was finding crashes (according to crash stats, though nothing was new) overnight, and after enabling it on the existing jobs old test cases seem to reproduce correctly. I'm still waiting for a few OSS-Fuzz jobs to finish to be sure, but it's essentially good to go.


mbarbella-chromium:
Don't land this quite yet, investigating a potential issue.

inferno-chromium (approved):

inferno-chromium:
Need some additions.

1. Remove APP_NAME=launcher.py from setup template.
src/local/butler/scripts/setup.py:AFL_TEMPLATE = """APP_NAME = launcher.py

2. Add AFL_SKIP_CRASHES (see https://github.com/google/fuzzbench/blob/bce8826f82c57bdc3a3412f2c511beeb7ace269d/fuzzers/afl/fuzzer.py#L76)
in launcher.py set_environment_variables

3. Probably exclude -6 exit code in https://github.com/google/clusterfuzz/blob/master/src/python/bot/fuzzers/afl/launcher.py#L672
that seems expected.

inferno-chromium (changes requested):
See comment above.

inferno-chromium (approved):

inferno-chromium:
Can we modify the result of that test to make sure it passes, that might be better than removing the test ?

PR #2172, opened 2020-12-11
Make the random-seed argument run-specific

mi-ac:
This chooses a fresh random-seed argument for every test-case run in blackbox fuzzing for a better variation. Previously, it was chosen only once per session.

The environment variable RANDOM_SEED is only used by V8 jobs and, hence, this has no effect on engine fuzzing or other non-V8 blackbox jobs.

The two environment variables WINDOW_ARG and ADDITIONAL_COMMAND_LINE_FLAGS are processed consecutively in this order by get_command_line_for_application of testcase_manager. Previously the random seed was appended to WINDOW_ARG, now it is prepended to ADDITIONAL_COMMAND_LINE_FLAGS, leaving the flag in the same position as before in the resulting command line.

### DIFF
--- a/src/python/bot/tasks/fuzz_task.py
+++ b/src/python/bot/tasks/fuzz_task.py
@@ -770,17 +770,20 @@ def pick_window_argument():
       window_argument = window_argument.replace('$LEFT', str(left))
       window_argument = window_argument.replace('$TOP', str(top))
 
-  # FIXME: Random seed is currently passed along to the next job
-  # via WINDOW_ARG. Rename it without breaking existing tests.
-  random_seed_argument = environment.get_value('RANDOM_SEED')
-  if random_seed_argument:
-    if window_argument:
-      window_argument += ' '
-    seed = utils.random_number(-2147483648, 2147483647)
-    window_argument += '%s=%d' % (random_seed_argument.strip(), seed)
-
   environment.set_value('WINDOW_ARG', window_argument)
   return window_argument
+
+
+def pick_seed_argument(env_dict):
+  """Prepend a random-seed argument to the additional flags in the copied
+  environment.
+  """
+  random_seed_argument = environment.get_value('RANDOM_SEED')
+  if random_seed_argument:
+    flags = env_dict.get('ADDITIONAL_COMMAND_LINE_FLAGS', '')
+    seed = utils.random_number(-2147483648, 2147483647)
+    flags = '%s=%d %s' % (random_seed_argument.strip(), seed, flags)
+    env_dict['ADDITIONAL_COMMAND_LINE_FLAGS'] = flags.strip()
 
 
 @retry.wrap(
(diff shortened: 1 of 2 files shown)


src/python/bot/tasks/fuzz_task.py
```diff
@@ -1763,6 +1766,9 @@ def do_blackbox_fuzzing(self, fuzzer, fuzzer_directory, job_type):
         # only apply some of the time. Adjust APP_ARGS for them if needed.
         trial_selector.setup_additional_args_for_app(env_copy)
 
+        # Pick a run-specific random-seed if specified.
+        pick_seed_argument(env_copy)
```
oliverchang, line 1770: Missed this in https://github.com/google/clusterfuzz/pull/2163 since I didn't get a chance to review :)

As is, these arguments are passed to the testcase run, but they do not get propagated properly to the Testcase entities for reproduction. These args are reset after each run, but they need to propagate all the way to here: https://github.com/google/clusterfuzz/blob/9b086bed9b9894ce12fef1aed6acbcf62c3aff92/src/python/bot/tasks/fuzz_task.py#L125

Short of a more comprehensive refactoring to attach more general metadata to each testcase, the easiest way may be to append these flags to the flags files instead, such that they're picked up in https://github.com/google/clusterfuzz/blob/9b086bed9b9894ce12fef1aed6acbcf62c3aff92/src/python/bot/testcase_manager.py#L294.

oliverchang (changes requested):

src/python/bot/tasks/fuzz_task.py
mi-ac: Hmm, I guess I'm running into some gotchas here... lemme try to understand:
- The environment in https://github.com/google/clusterfuzz/blob/9b086bed9b9894ce12fef1aed6acbcf62c3aff92/src/python/bot/tasks/fuzz_task.py#L125 is not the one we copied when running test cases?
- The call testcase_manager.get_command_line_flags(crash.file_path) there is followed by testcase_manager.get_command_line_for_application, which also indirectly calls get_additional_command_line_flags as well. Do we get those twice?
- Is this now already broken for trials?
- Is there an existing test that could be tweaked to exercise the problem you describe?

I have the feeling the devil lies in using Python's "environment" for the application's data flow. It's sort-of invisible in the program code and hard to reason about as it's part of the global state. Would there be a long-term plan to refactor this into an explicit data abstraction (e.g. a class storing all the cmd-line arguments and env vars - that then can be passed around as explicit objects for each test case)?

mi-ac: I think I understand better now. https://clusterfuzz.com/testcase-detail/5728059199389696 shows, how this is already broken. There a crash was found that was triggered by --simulate-errors chosen by a trial, but the command line doesn't show the flag. The crash is also not marked as reproducible.

Maybe I can make the fuzz task test case a bit richer to also show this problem. I think I understand this enough to work on a short-term fix.

PR #2181, opened 2020-12-26
Improve fuzz-task test for crashes

mi-ac:
Before, threads were mocked out, now pseudo threads execute more code, only the actual execution of the tests is mocked out.
Two crashes are faked, and the command-lines for the crashes are verified.

This prepares making changes to the handling of trial arguments and the random-seed argument. It will ease the verification of the command-lines of crashes, which is not covered by any other test.

### DIFF
--- a/src/python/tests/core/bot/tasks/fuzz_task_test.py
+++ b/src/python/tests/core/bot/tasks/fuzz_task_test.py
@@ -18,8 +18,10 @@ import datetime
 import mock
 import os
 import parameterized
+import queue
 import shutil
 import tempfile
+import threading
 import time
 import unittest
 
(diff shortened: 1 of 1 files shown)


src/python/tests/core/bot/tasks/fuzz_task_test.py
```diff
@@ -1343,23 +1344,45 @@ def setUp(self):
...
     os.environ['MAX_TESTCASES'] = '3'
     os.environ['RANDOM_SEED'] = '-r'
+    os.environ['ROOT_DIR'] = '/root'
     os.environ['THREAD_ALIVE_CHECK_INTERVAL'] = '0.001'
     os.environ['THREAD_DELAY'] = '0.001'
+    os.environ['USER_PROFILE_IN_MEMORY'] = 'True'
+
+    test_utils.set_up_pyfakefs(self)
+    self.fs.create_dir('/crash')
+    self.fs.create_dir('/root/bot/logs')
```
mi-ac, line 1385: This line is required when running tests without -v. The log handler takes a different path in this case.

src/python/tests/core/bot/tasks/fuzz_task_test.py
```diff
@@ -1368,14 +1391,29 @@ def setUp(self):
...
+    }]
     self.mock.current_timestamp.return_value = 0.0
 
-    # Mock out threads with immedietely ending dummies.
-    self.process = mock.MagicMock()
-    self.mock.get_process.return_value = self.process
-    thread = mock.MagicMock()
-    self.process.return_value = thread
-    thread.is_alive.return_value = False
+    crash_queue = queue.Queue()
```
oliverchang, line 1400: Can we just not mock get_queue if we're gonna expect it to return a real Queue?

oliverchang (commented):
nice!

src/python/tests/core/bot/tasks/fuzz_task_test.py
mi-ac: Not sure if you meant that, but I now simplified it a bunch by just switching things to threading.Thread. Another way would be to mock is_trusted_host=True, which has the same effect, but it also activates a few other code paths, which looks like something that should be tested in separation.

Using multiprocessing would be complicated as the fake filesystem (and maybe other testing setup) isn't properly initialized in the other processes.

oliverchang (approved):

PR #2195, opened 2021-01-12
Use HTTP basic auth for GitHub.

oliverchang:
For query param deprecation: https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param/


inferno-chromium (approved):
What is this needed for, can you add in description.

oliverchang:
> What is this needed for, can you add in description.

Added. Query params are deprecated: https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param/

PR #2206, opened 2021-01-25
Use explicit instance dir for afl-fuzz, needed for afl++

inferno-chromium:
afl++ defaults to -S default now. So, explicitly provide one.


jonathanmetzman:
> afl++ defaults to -S default now. So, explicitly provide one.

Is there some kind of error this fixes?
I think we need to swap out the afl-fuzz binaries for afl++ ones for this to pass tests.

inferno-chromium:
TBR to unblock bots, and see if we need to revert oss-fuzz commit/it sticks.

jonathanmetzman (commented):
LGTM

jonathanmetzman:
Isn't this going to break internal ClusterFuzz?

inferno-chromium:
> Isn't this going to break internal ClusterFuzz?

-S is supported by regular afl.

jonathanmetzman:
> > Isn't this going to break internal ClusterFuzz?
> 
> -S is supported by regular afl.

Nice!

PR #2208, opened 2021-01-26
Tweak settings for AFL++

inferno-chromium:
- Change fuzz duration to 4 hr.
- Use AFL_EXPAND_HAVOC_NOW as per Marc's recommendation since we
  are working on saturated corpus most of the time.
  "Setting AFL_EXPAND_HAVOC_NOW will start in the extended havoc mode
  that includes costly mutations."


jonathanmetzman (approved):
LGTM

PR #2211, opened 2021-01-27
Fix AFL stats.

oliverchang:
- Rename strategies to ones used in schema.
- Remove deprecated strategy printing code.

### DIFF
--- a/src/python/bot/fuzzers/afl/launcher.py
+++ b/src/python/bot/fuzzers/afl/launcher.py
@@ -310,12 +310,16 @@ class FuzzingStrategies(object):
     # The decision on whether or not fast cal should be used is made during
     # fuzzing. This function is expected to be called whe preparing for fuzzing.
     assert self.fast_cal == strategies.FastCal.NOT_SET
+    strategies_dict = {}
 
-    strategies_dict = {
-        'candidate_generator': self.candidate_generator,
-    }
+    if self.generator_strategy == engine_common.Generator.RADAMSA:
+      strategies_dict[strategy.CORPUS_MUTATION_RADAMSA_STRATEGY.name] = 1
+    elif self.generator_strategy == engine_common.Generator.ML_RNN:
+      strategies_dict[strategy.CORPUS_MUTATION_ML_RNN_STRATEGY.name] = 1
+
     if self.use_corpus_subset:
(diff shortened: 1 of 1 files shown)


src/python/bot/fuzzers/afl/launcher.py
```diff
@@ -365,27 +369,6 @@ def decide_fast_cal_random(self, input_dir):
...
-    """Print the strategies used for logging purposes."""
-    fuzzing_strategies = []
-
-    if self.generator_strategy == engine_common.Generator.RADAMSA:
-      fuzzing_strategies.append('strategy_' +
-                                strategy.CORPUS_MUTATION_RADAMSA_STRATEGY.name)
-    elif self.generator_strategy == engine_common.Generator.ML_RNN:
-      fuzzing_strategies.append('strategy_' +
-                                strategy.CORPUS_MUTATION_ML_RNN_STRATEGY.name)
-    if self.fast_cal == strategies.FastCal.RANDOM:
```
oliverchang, line 378: This is no longer surfaced in the new stats pipeline (see line 312). However since stats were completely broken anyway this isn't necessarily a regression.

@jonathanmetzman is the assumption on line 312 correct that we can only get these when we fuzz? Are these stats important enough for us to make interface changes to support strategies obtained during fuzz() ?

src/python/bot/fuzzers/afl/launcher.py
```diff
@@ -310,12 +310,16 @@ def to_strategy_dict(self):
     # The decision on whether or not fast cal should be used is made during
     # fuzzing. This function is expected to be called whe preparing for fuzzing.
     assert self.fast_cal == strategies.FastCal.NOT_SET
+    strategies_dict = {}
+
+    if self.generator_strategy == engine_common.Generator.RADAMSA:
```
inferno-chromium, line 315: Do we need the FastCal ones here too ?

inferno-chromium (approved):

src/python/bot/fuzzers/afl/launcher.py
oliverchang: See line 312 and https://github.com/google/clusterfuzz/pull/2211#discussion_r564941258

Issue #2214, opened 2021-01-27
Fallback to CF's llvm-symbolizer if the one in build fails to run.


oliverchang:

oliverchang:
@inferno-chromium to prevent future breakage.

PR #2219, opened 2021-01-27
Fall-back to own llvm-symbolizer if one from archive fails.

inferno-chromium:
Fixes https://github.com/google/clusterfuzz/issues/2214

### DIFF
--- a/src/python/system/environment.py
+++ b/src/python/system/environment.py
@@ -19,6 +19,7 @@ import os
 import re
 import six
 import socket
+import subprocess
 import sys
 import yaml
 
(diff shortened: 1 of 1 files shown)


src/python/system/environment.py
```diff
@@ -332,13 +333,25 @@ def get_llvm_symbolizer_path():
...
+  if llvm_symbolizer_path and os.path.exists(llvm_symbolizer_path):
+    # Make sure that llvm symbolizer binary is executable.
+    os.chmod(llvm_symbolizer_path, 0o750)
+
+    return_code = subprocess.call(
+        [llvm_symbolizer_path, '--help'],
+        stdout=subprocess.DEVNULL,
+        stderr=subprocess.DEVNULL)
+    if return_code == 0:
+      # llvm-symbolize works, return it.
```
oliverchang, line 345: nit: log_error so we know about it?

oliverchang (approved):
thanks!

src/python/system/environment.py
inferno-chromium: environment is such a self-contained module, that is why skipped adding logging dependency to avoid cycles.

oliverchang: Ah right, thanks for clarifying.

PR #2224, opened 2021-01-29
Speculative fix for fuzzer_and_job_weights exception.

oliverchang:


oliverchang:
tbr

inferno-chromium (commented):
dont understand this, but lets see how it goes.

oliverchang:
> dont understand this, but lets see how it goes.

This can happen when columns are missing in the given time period for a fuzzer. The weight result from this query: https://github.com/google/clusterfuzz/blob/17dc87e1a41b5d2cffe7702d5528a18a50b42c33/src/appengine/handlers/cron/fuzzer_and_job_weights.py#L96 will be null in that case.

PR #2230, opened 2021-02-04
pubsub: create subscriptions that don't expire.

oliverchang:


inferno-chromium (approved):
Thanks a ton! this has bitten us some times now!

andrei-near:
got bitten by this today: jobs-linux subscription isn't created even after 3rd project deploy tentative. 
Double checked queues.yaml and is there. 
Fixed by adding it manually to GCP

PR #2231, opened 2021-02-05
[fuchsia] Add support for undercoat-managed Fuchsia instances

eepeep:
Rollout is controlled via the `FUCHSIA_UNDERCOAT_ROLLOUT_LEVEL` environment variable (0-1).

Running the integration tests requires a Fuchsia build bundle that includes the `undercoat` binary (which will soon be all of them).


eepeep:
cc @oliverchang @nopsledder

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -413,6 +414,201 @@ class UnshareLibFuzzerRunner(new_process.UnshareProcessRunnerMixin,
...
+  def _target_corpus_path(self, corpus_name):
+    """ Returns the path of a given corpus directory on the target. """
+    return 'data/corpus/' + corpus_name
+
+  def _push_corpora_from_host_to_target(self, corpus_directories):
+    # Push corpus directories to the device.
+    self._clear_all_target_corpora()
+    logs.log('Push corpora from host to target.')
+    for corpus_dir in corpus_directories:
+      # TODO(eep): Was the previous implementation's behavior of merging all the
```
oliverchang, line 454: better to match host behaviour (separate corpora) if possible. 

that said what are the semantics of put_data here? do the contents of each `corpus_dir` end up merged into `data/corpus`, do do they get copied as `data/corpus/corpus_dir_0`, `data/corpus/corpus_dir_1` etc... ?

src/python/platforms/fuchsia/undercoat.py
```diff
@@ -0,0 +1,153 @@
...
+  # TODO(fxbug.dev/47490): Pass back raw return code from libFuzzer?
+  undercoat_args = ['-fuzzer', fuzzer]
+  if outdir:
+    undercoat_args += ['-artifact-dir', outdir]
+  args = undercoat_args + ['--'] + args
+  return undercoat_instance_command('run_fuzzer', handle, *args)
+
+
+def put_data(handle, fuzzer, src, dst):
+  """Put files for a fuzzer onto an instance, via undercoat."""
```
oliverchang, line 145: Can you add some documentation on the semantics here when `src` or `dst` is a directory (or both)?

src/python/platforms/fuchsia/undercoat.py
```diff
@@ -0,0 +1,153 @@
...
+  return undercoat_instance_command('put_data', handle, '-fuzzer', fuzzer,
+                                    '-src', src, '-dst', dst).output
+
+
+def get_data(handle, fuzzer, src, dst):
```
oliverchang, line 150: same on semantics here too.

src/python/platforms/fuchsia/undercoat.py
```diff
@@ -0,0 +1,153 @@
...
+import tempfile
+
+from metrics import logs
+from system import environment
+from system import new_process
+from system import shell
+
+# In order to properly clean up any stale instances, we need to keep track of
+# handles in a way that persists between runs
+UNDERCOAT_HANDLE_DIR = os.path.join(tempfile.gettempdir(), 'undercoat-handles')
```
oliverchang, line 28: put this into a function rather than a global, and use environment.get_value('BOT_TMPDIR') instead.

i.e.

```python
def get_handle_dir():
  return os.path.join(...)
```

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -413,6 +414,201 @@ class UnshareLibFuzzerRunner(new_process.UnshareProcessRunnerMixin,
...
+    undercoat.prepare_fuzzer(self.handle, self.executable_path)
+    self._push_corpora_from_host_to_target(corpus_directories)
+
+    max_total_time = self.get_max_total_time(fuzz_timeout)
+    if any(arg.startswith(constants.FORK_FLAG) for arg in additional_args):
+      max_total_time -= self.LIBFUZZER_FORK_MODE_CLEAN_EXIT_TIME
+    assert max_total_time > 0
+
+    additional_args.extend([
+        '%s%d' % (constants.MAX_TOTAL_TIME_FLAG, max_total_time),
```
oliverchang, line 496: nit: just `constants.MAX_TOTAL_TIME_FLAG + str(max_total_time)` here.

oliverchang (commented):
nice! some initial comments.

src/python/build_management/build_manager.py
```diff
@@ -738,8 +739,21 @@ class FuchsiaBuild(RegularBuild):
...
+    rollout_level = environment.get_value('FUCHSIA_UNDERCOAT_ROLLOUT_LEVEL', 0)
+    self.use_undercoat = random.random() < rollout_level
+    logs.log('Undercoat enabled: %s' % self.use_undercoat)
+
   def _get_fuzz_targets_from_dir(self, build_dir):
     """Overridden to get targets list from fuchsia."""
+    if self.use_undercoat:
+      # list_fuzzers requires a running instance, so we can't list these until
+      # later in setup() below
+      return []
```
oliverchang, line 755: Taking a closer look here it's a rather difficult to untie the target selection logic from the unpacking process.

Something like this is reasonable, but will result in a log_error every setup() because we return this empty list.

Here's a rough approach that I can think of:

Add a `_pick_fuzz_target` **method** to the `Build` class which just calls the top level `_set_random_fuzz_target_for_fuzzing_if_needed` by default. Then replace all the existing calls where appropriate to call the class method instead.  We still need the top-level function because it's called from outside the context of a Build as well. 

Then for fuchsia, we can just override `_pick_fuzz_target` to **always** be a no-op.

Then in `setup()` we can decide whether or not to use undercoat after the unpacking, and call a custom `_pick_fuzz_target_for_fuchsia` of some sort in both cases.

Does that make sense/seem reasonable?

src/python/platforms/fuchsia/undercoat.py
eepeep: Ah, nice! The only thing `persistent_cache` doesn't allow that I was doing before is enumeration, but I think it's unnecessary if I don't need to worry about multiple processes acting on the same cache key simultaneously.  I assume this is a reasonable assumption to make?

src/python/build_management/build_manager.py
eepeep: Thanks for the detailed suggestions! I did something similar to what you described, with a little bit of modification: mostly a) leaving out the extra layer of `_pick_fuzz_target_for_fuchsia` because it would simply be another bare wrapper to `_set_random_fuzz_target_for_fuzzing_if_needed` and b) it seemed slightly cleaner for `_pick_fuzz_target` to be an instance method instead of a class method, even though it doesn't need to be.  Happy to change those if you'd prefer otherwise, though.  I also split the setup function up a bit more in the process of implementing fallback logic, which hopefully makes the distinction between the two codepaths (legacy/undercoat) slightly clearer.

eepeep:
I used the github interface to resolve a minor merge conflict but I see it has created a merge commit and I'm not sure that was the best choice... lmk if you'd prefer me to rebase.

src/python/platforms/fuchsia/undercoat.py
```diff
@@ -46,14 +55,14 @@ class UndercoatError(Exception):
...
   undercoat = new_process.ProcessRunner(undercoat_path, args)
-  # TODO(eep): Is there a more useful way to connect stderr to the
-  # logging system?
-  result = undercoat.run_and_wait(stderr=sys.stderr)
+  error_log = tempfile.TemporaryFile()
+  result = undercoat.run_and_wait(stderr=error_log)
   result.output = result.output.decode('utf-8')
 
   if result.return_code != 0:
+    logs.log_warn(utils.read_from_handle_truncated(error_log, 1024 * 1024))
```
oliverchang, line 65: nit: add `'Error output from undercoat: ' + `

src/python/platforms/fuchsia/undercoat.py
```diff
@@ -46,14 +55,14 @@ class UndercoatError(Exception):
...
     raise UndercoatError(
         'Error running undercoat command %s: %s' % (args, result.output))
```
oliverchang, line 67: Does `result.output` actually make sense here when errors are written to stderr?

src/python/build_management/build_manager.py
```diff
@@ -739,19 +744,17 @@ class FuchsiaBuild(RegularBuild):
...
+  def _pick_fuzz_target(self, fuzz_targets, target_weights):
+    """No-op, since Fuchsia builds pick targets later than other build types
+    and we aren't ready at the point that this is called by the superclass's
+    setup()."""
 
   def _get_fuzz_targets_from_dir(self, build_dir):
-    """Overridden to get targets list from fuchsia."""
+    """Returns a list of fuzz targets in a legacy Fuchsia build. For
+    undercoat-driven Fuchsia builds, a running instance is required to
+    enumerate targets so this is a no-op."""
```
oliverchang, line 757: Can we raise an exception here now if `use_undercoat` is true, since this no now never be called in that case?

oliverchang (approved):
nice!

src/python/platforms/fuchsia/undercoat.py
eepeep: The clusterfuchsia API is defined as returning an error message via stdout when the exit code is non-zero. stderr, on the other hand, is used for more detailed logging (i.e. log messages from undercoat itself, not directly related to the API) and is echoed here to assist in debugging. Does that address your concern here?

src/python/build_management/build_manager.py
eepeep: Unfortunately this is still called by `RegularBuild.setup()` via `Build._unpack_build()`, so at least as things are now I think we don't want to throw an error in this case.

eepeep:
/gcbrun

Not sure what happen with the previous GCB run; the error didn't seem related to the area this code touched.

src/python/platforms/fuchsia/undercoat.py
oliverchang: I see, it was just a bit confusing that we are logging stderr and then raising an exception with stdout. Perhaps add this in a comment?

oliverchang (approved):
Thanks for addressing all the comments! LGTM with a nit.

src/python/platforms/fuchsia/undercoat.py
eepeep: Good point! I've added clarifying comments.

eepeep:
And thank you for your feedback!

PR #2238, opened 2021-02-09
Support making reproduction requests to external infrastructure.

oliverchang:


src/python/base/tasks.py
```diff
@@ -364,6 +365,14 @@ def add_task(command, argument, job_type, queue=None, wait_time=None):
   if wait_time is None:
     wait_time = random.randint(1, TASK_CREATION_WAIT_INTERVAL)
 
+  job = data_types.Job.query(data_types.Job.name == job_type).get()
+  if not job:
+    raise Error(f'Job {job_type} not found.')
+
+  if job.is_external():
+    external_tasks.add_external_task(command, argument, job)
```
inferno-chromium, line 373: nit: it might be nice if we can pass job name here, and move all the checks like job exists, is external, etc in that function as bail out conditions.

src/python/base/external_tasks.py
```diff
@@ -0,0 +1,62 @@
...
+  topic_name = job.external_reproduction_topic
+  assert topic_name is not None
+
+  testcase = data_handler.get_testcase_by_id(testcase_id)
+  fuzz_target = testcase.get_fuzz_target()
+
+  memory_tool_name = environment.get_memory_tool_name(job.name)
+  sanitizer = environment.SANITIZER_NAME_MAP.get(memory_tool_name)
+  job_environment = job.get_environment()
+  build_path = (
```
inferno-chromium, line 39: Do we need a bail out for custom binaeries.

src/python/tests/appengine/common/tasks_test.py
```diff
@@ -266,3 +266,113 @@ def test_high_end_queue(self):
     t = data_types.Testcase(job_type='job_project', queue='high-end-jobs')
     self.assertEqual('high-end-jobs-project-linux-lib',
                      tasks.queue_for_testcase(t))
+
+
+@test_utils.with_cloud_emulators('datastore', 'pubsub')
+class ExternalTasksTest(unittest.TestCase):
```
inferno-chromium, line 272: Very nice tests!

inferno-chromium (approved):

src/python/base/external_tasks.py
oliverchang: This should never happen (it doesn't make sense for an "external" infra to know about custom binaries). I can add an exception here.

src/python/tests/appengine/common/tasks_test.py
oliverchang: :)

src/python/base/tasks.py
oliverchang: Done!

oliverchang: i think that makes some conditions in is_external more confusing because we need to skip adding tasks for all external jobs, even if that command is unsupported.. Also the job exists check seems beneficial in general if we're adding tasks.

If we do this the bailout conditions will look like:

```python
def add_external_task(...):
  if not job.is_external():
    return False

  if command != 'progression':
    return True   # this is confusing but necessary
```

PR #2240, opened 2021-02-11
Implement external reproduction updates.

oliverchang:
This is implemented as an App Engine endpoint which receives Pub/Sub
push notifications.

### DIFF
--- a/src/appengine/handlers/external_update.py
+++ b/src/appengine/handlers/external_update.py
@@ -60,6 +60,7 @@ def handle_update(testcase, revision, stacktrace, error):
                   f'revision {last_tested_revision}.')
     return
 
+  # TODO(ochang): Record fuzz target.
   fuzz_target = testcase.get_fuzz_target()
   if fuzz_target:
     fuzz_target_name = fuzz_target.binary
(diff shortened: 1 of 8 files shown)


src/python/tests/appengine/handlers/external_update_test.py
```diff
@@ -0,0 +1,211 @@
...
+        'email': 'test-clusterfuzz@appspot.gserviceaccount.com',
+    }
+
+    data_types.Job(
+        name='external_job',
+        external_reproduction_topic='topic',
+        external_updates_subscription='subscription').put()
+    data_types.Job(name='job').put()
+
+    self.testcase_0 = data_types.Testcase(
```
inferno-chromium, line 62: nit: just testcase instead of testcase_0

src/appengine/handlers/external_update.py
```diff
@@ -0,0 +1,124 @@
...
+
+  is_security = crash_analyzer.is_security_issue(
+      state.crash_stacktrace, state.crash_type, state.crash_address)
+  if is_security != testcase.security_flag:
+    logs.log('Security flag no longer matches.', testcase_id=testcase.key.id())
+    _mark_as_fixed(testcase, revision)
+    return
+
+  logs.log('Still crashes.', testcase_id=testcase.key.id())
+  testcase.last_tested_crash_stacktrace = stacktrace
```
inferno-chromium, line 92: Do we need to refetch testcase to avoid race updates. Same for some branches like line 88, line 81

inferno-chromium (approved):

src/appengine/handlers/external_update.py
oliverchang: Shouldn't be needed. Unlike native jobs these don't run any other tasks and this is the only place updates happen.

src/python/tests/appengine/handlers/external_update_test.py
oliverchang: Fixed!

PR #2248, opened 2021-02-18
Install libpython3-all-dev in install_deps_linux.bash

oliverchang:
This is needed for many pip packages to install. Also drop support for
trusty (14.04), where this package does not exist.


jonathanmetzman (approved):
LGTM

PR #2252, opened 2021-02-22
Change afl hang timeout from 1000+ to 5000+

inferno-chromium:
Some targets can be slow, so need higher timeout, like 5 sec.
1000+ or 1 sec is not enough as all inputs in a corpus can
hang on a slow target and then afl-fuzz will crash. So, 5 sec
seems to work well.

See https://github.com/google/oss-fuzz/pull/5223#issuecomment-782325020

### DIFF
--- a/src/python/bot/fuzzers/afl/launcher.py
+++ b/src/python/bot/fuzzers/afl/launcher.py
@@ -516,7 +516,7 @@ class AflRunnerCommon(object):
 
   # The timeout we will use if autocalibrating results in too many hangs. This
   # is the maximum autocalibrated timeout afl-fuzz can set.
-  MANUAL_TIMEOUT_MILLISECONDS = 1000
+  MANUAL_TIMEOUT_MILLISECONDS = 5000
 
   # Regexes used to determine which file caused AFL to quit.
   CRASH_REGEX = re.compile(


oliverchang (approved):

jonathanmetzman:
I don't love this.
I think it's good that AFL skips inputs that take longer than a second.
Doing this makes AFL more similar to libFuzzer since now AFL doesn't get as much performance benefit of skipping slow inputs.
My ideal would be for the max time limit AFL autocalibrates to be configurable, so that on targets where less than 1 second is not viable AFL uses a higher timeout, but on other ones it uses a lower timeout.
I guess if @vanhauser-thc thinks it's OK than LGTM.

vanhauser-thc:
> I don't love this.
> I think it's good that AFL skips inputs that take longer than a second.
> Doing this makes AFL more similar to libFuzzer since now AFL doesn't get as much performance benefit of skipping slow inputs.
> I guess if @vanhauser-thc thinks it's OK than LGTM.

as far as I understood there are projects that have such slow runtimes that this is required so the actually can be run.

In the issue where this was discussed I proposed to make this rather configurable by project as I see the same issue that it hurts overall performance and is just necessary on a handful of targets.

> My ideal would be for the max time limit AFL autocalibrates to be configurable, so that on targets where less than 1 second is not viable AFL uses a higher timeout, but on other ones it uses a lower timeout.

maybe something like  `-t +` where this is calculated after running the initial seeds?
it could gather the average runtime, multiply by 2 and use that as max value maybe?

inferno-chromium:
> I don't love this.
> I think it's good that AFL skips inputs that take longer than a second.
> Doing this makes AFL more similar to libFuzzer since now AFL doesn't get as much performance benefit of skipping slow inputs.
> My ideal would be for the max time limit AFL autocalibrates to be configurable, so that on targets where less than 1 second is not viable AFL uses a higher timeout, but on other ones it uses a lower timeout.
> I guess if @vanhauser-thc thinks it's OK than LGTM.

this should only happen if there are too many timeouts in corpus, indicating slow target. But as @vanhauser-thc, would prefer if we has -t + or even default that auto-calibrates timeout. Also, what happens if there is no corpus on a slow target.

vanhauser-thc:
> > I don't love this.
> > I think it's good that AFL skips inputs that take longer than a second.
> > Doing this makes AFL more similar to libFuzzer since now AFL doesn't get as much performance benefit of skipping slow inputs.
> > My ideal would be for the max time limit AFL autocalibrates to be configurable, so that on targets where less than 1 second is not viable AFL uses a higher timeout, but on other ones it uses a lower timeout.
> > I guess if @vanhauser-thc thinks it's OK than LGTM.
> 
> this should only happen if there are too many timeouts in corpus, indicating slow target. 

this impacts also normal fuzzing where a fuzz entry hangs. it is not terminated after 1 seconds but after 5.

> But as @vanhauser-thc, would prefer if we has -t + or even default that auto-calibrates timeout. Also, what happens if there is no corpus on a slow target.

maybe `-t 5000++` the extra `+` meaning auto-calibrate, but max is 5000ms

vanhauser-thc:
If anyone wants to try it: https://github.com/AFLplusplus/AFLplusplus/commit/ee0ca07f3c9f7c5971440f5dca70a2ee6f37584d
(current dev state).
no need to update the -t 5000+ option to anything else, the '+' now means to auto-calculate the timeout based on the initial seeds with 5000 being the max value. Note that if only 1 seed is present then it is fixed at 5000 without auto-calculating.
feedback welcome.

vanhauser-thc:
#5258

PR #2261, opened 2021-02-25
Fix an exception for testcases with missing queue.

oliverchang:
These apply to external testcases. This is the only use of this variable
so adding a check here rather than setting an empty string default.


oliverchang:
tbr

inferno-chromium (commented):
lgtm

PR #2265, opened 2021-02-25
Fix logging for external updates.

oliverchang:
Extra attributes don't work on App Engine, so format them into the
message body instead.


oliverchang:
TBR

inferno-chromium (commented):
lgtm

PR #2266, opened 2021-02-26
Upgrade protobuf.

oliverchang:
Something changed/broke:

AttributeError
module 'google.protobuf.descriptor' has no attribute
'_internal_create_key'


inferno-chromium (commented):
thanks a lot!

PR #2267, opened 2021-02-26
add EMULATED_ANDROID platform

kalder:
This allows downloading an emulator bundle (from the
ANDROID_EMULATOR_BUCKET_PATH variable) and running it (using the
ANDROID_EMULATOR_SCRIPT_PATH variable) before fuzzing.

The emulated Android instance is controlled using 'adb', just as a
physical Android device would. The new EmulatedAndroidLibFuzzerRunner
shares most logic with the existing AndroidLibFuzzerRunner.

### DIFF
--- a/src/python/base/tasks.py
+++ b/src/python/base/tasks.py
@@ -63,7 +63,7 @@ TASK_QUEUE_DISPLAY_NAMES = {
     'ANDROID_KERNEL': 'Android Kernel',
     'ANDROID_AUTO': 'Android Auto',
     'ANDROID_X86': 'Android (x86)',
-    'EMULATED_ANDROID': 'Android (Emulated)',
+    'ANDROID_EMULATOR': 'Android (Emulated)',
     'CHROMEOS': 'Chrome OS',
     'FUCHSIA': 'Fuchsia OS',
     'MAC': 'Mac',
--- a/src/python/base/tasks.py
+++ b/src/python/base/tasks.py
@@ -63,6 +63,7 @@ TASK_QUEUE_DISPLAY_NAMES = {
     'ANDROID_KERNEL': 'Android Kernel',
     'ANDROID_AUTO': 'Android Auto',
     'ANDROID_X86': 'Android (x86)',
+    'EMULATED_ANDROID': 'Android (Emulated)',
     'CHROMEOS': 'Chrome OS',
     'FUCHSIA': 'Fuchsia OS',
     'MAC': 'Mac',
--- a/src/python/bot/fuzzers/libfuzzer.py
+++ b/src/python/bot/fuzzers/libfuzzer.py
@@ -1298,6 +1298,7 @@ class AndroidLibFuzzerRunner(new_process.UnicodeProcessRunner, LibFuzzerCommon):
 
       return result
 
+
 class EmulatedAndroidLibFuzzerRunner(AndroidLibFuzzerRunner):
   """Android libFuzzer runner."""
 
@@ -1312,7 +1313,8 @@ class EmulatedAndroidLibFuzzerRunner(AndroidLibFuzzerRunner):
     start_emulator()
 
     super().__init__(
-        executable_path=executable_path, build_directory=build_directory,
+        executable_path=executable_path,
+        build_directory=build_directory,
         default_args=default_args)
 
   def __del__(self):
--- a/src/python/bot/fuzzers/libfuzzer.py
+++ b/src/python/bot/fuzzers/libfuzzer.py
@@ -37,8 +37,6 @@ from fuzzing import strategy
 from metrics import logs
 from platforms import android
 from platforms import fuchsia
-from platforms.android.emulator import start_emulator
-from platforms.android.emulator import stop_emulator
 from platforms.fuchsia.device import QemuProcess
 from platforms.fuchsia.device import start_qemu
 from platforms.fuchsia.device import stop_qemu
@@ -1299,28 +1297,6 @@ class AndroidLibFuzzerRunner(new_process.UnicodeProcessRunner, LibFuzzerCommon):
       return result
 
 
-class EmulatedAndroidLibFuzzerRunner(AndroidLibFuzzerRunner):
-  """Android libFuzzer runner."""
-
-  def __init__(self, executable_path, build_directory, default_args=None):
-    """Inits the EmulatedAndroidLibFuzzerRunner.
-
-    Args:
-      executable_path: Path to the fuzzer executable.
-      build_directory: A MinijailChroot.
-      default_args: Default arguments to always pass to the fuzzer.
-    """
-    start_emulator()
-
-    super().__init__(
-        executable_path=executable_path,
-        build_directory=build_directory,
-        default_args=default_args)
-
-  def __del__(self):
-    stop_emulator()
-
-
 def get_runner(fuzzer_path, temp_dir=None, use_minijail=None, use_unshare=None):
   """Get a libfuzzer runner."""
   if use_minijail is None:
@@ -1390,10 +1366,7 @@ def get_runner(fuzzer_path, temp_dir=None, use_minijail=None, use_unshare=None):
   elif is_fuchsia:
     runner = FuchsiaQemuLibFuzzerRunner(fuzzer_path)
   elif is_android:
-    if environment.is_android_emulator():
-      runner = EmulatedAndroidLibFuzzerRunner(fuzzer_path, build_dir)
-    else:
-      runner = AndroidLibFuzzerRunner(fuzzer_path, build_dir)
+    runner = AndroidLibFuzzerRunner(fuzzer_path, build_dir)
   elif use_unshare:
     runner = UnshareLibFuzzerRunner(fuzzer_path)  # pylint: disable=too-many-function-args
   else:
--- a/src/python/bot/fuzzers/libfuzzer.py
+++ b/src/python/bot/fuzzers/libfuzzer.py
@@ -37,6 +37,8 @@ from fuzzing import strategy
 from metrics import logs
 from platforms import android
 from platforms import fuchsia
+from platforms.android.emulator import start_emulator
+from platforms.android.emulator import stop_emulator
 from platforms.fuchsia.device import QemuProcess
 from platforms.fuchsia.device import start_qemu
 from platforms.fuchsia.device import stop_qemu
@@ -1296,6 +1298,26 @@ class AndroidLibFuzzerRunner(new_process.UnicodeProcessRunner, LibFuzzerCommon):
 
       return result
 
+class EmulatedAndroidLibFuzzerRunner(AndroidLibFuzzerRunner):
+  """Android libFuzzer runner."""
+
+  def __init__(self, executable_path, build_directory, default_args=None):
+    """Inits the EmulatedAndroidLibFuzzerRunner.
+
+    Args:
+      executable_path: Path to the fuzzer executable.
+      build_directory: A MinijailChroot.
+      default_args: Default arguments to always pass to the fuzzer.
+    """
+    start_emulator()
+
+    super().__init__(
+        executable_path=executable_path, build_directory=build_directory,
+        default_args=default_args)
+
+  def __del__(self):
+    stop_emulator()
+
 
 def get_runner(fuzzer_path, temp_dir=None, use_minijail=None, use_unshare=None):
   """Get a libfuzzer runner."""
@@ -1366,7 +1388,10 @@ def get_runner(fuzzer_path, temp_dir=None, use_minijail=None, use_unshare=None):
   elif is_fuchsia:
     runner = FuchsiaQemuLibFuzzerRunner(fuzzer_path)
   elif is_android:
-    runner = AndroidLibFuzzerRunner(fuzzer_path, build_dir)
+    if environment.platform() == 'EMULATED_ANDROID':
+      runner = EmulatedAndroidLibFuzzerRunner(fuzzer_path, build_dir)
+    else:
+      runner = AndroidLibFuzzerRunner(fuzzer_path, build_dir)
   elif use_unshare:
     runner = UnshareLibFuzzerRunner(fuzzer_path)  # pylint: disable=too-many-function-args
   else:
--- a/src/python/bot/fuzzers/libfuzzer.py
+++ b/src/python/bot/fuzzers/libfuzzer.py
@@ -1390,7 +1390,7 @@ def get_runner(fuzzer_path, temp_dir=None, use_minijail=None, use_unshare=None):
   elif is_fuchsia:
     runner = FuchsiaQemuLibFuzzerRunner(fuzzer_path)
   elif is_android:
-    if environment.platform() == 'EMULATED_ANDROID':
+    if environment.is_android_emulator():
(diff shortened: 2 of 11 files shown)


src/python/platforms/android/emulator.py
```diff
@@ -0,0 +1,126 @@
...
+class EmulatorProcess(object):
+  """A EmulatorProcess encapsulates the creation, running, and destruction
+  of Android emulator processes."""
+
+  def __init__(self):
+    self.process_runner = None
+    self.popen = None
+    self.logfile = None
+
+    log_path = os.path.join(tempfile.gettempdir(), 'android-emulator-log')
```
inferno-chromium, line 48: android-emulator.log is slightly better name.

src/python/platforms/android/emulator.py
```diff
@@ -0,0 +1,126 @@
...
+    self.logfile = open(log_path, 'wb')
+
+  def create(self):
+    """Configures a emulator process which can subsequently be `run`."""
+    # Download emulator image
```
inferno-chromium, line 53: nit: end with period.

src/python/platforms/android/emulator.py
```diff
@@ -0,0 +1,126 @@
...
+    temp_directory = environment.get_value('BOT_TMPDIR')
+
```
inferno-chromium, line 55: remove newline to align to comment.

src/python/platforms/android/emulator.py
```diff
@@ -0,0 +1,126 @@
...
+    archive_src_path = environment.get_value('ANDROID_EMULATOR_BUCKET_PATH')
+    archive_dst_path = os.path.join(temp_directory, 'emulator_bundle.zip')
+
```
inferno-chromium, line 58: remove newline to align to comment.

src/python/platforms/android/emulator.py
```diff
@@ -0,0 +1,126 @@
...
+    log_path = os.path.join(tempfile.gettempdir(), 'android-emulator-log')
+    self.logfile = open(log_path, 'wb')
+
+  def create(self):
+    """Configures a emulator process which can subsequently be `run`."""
+    # Download emulator image
+    temp_directory = environment.get_value('BOT_TMPDIR')
+
+    archive_src_path = environment.get_value('ANDROID_EMULATOR_BUCKET_PATH')
```
inferno-chromium, line 56: What if this is not set, we should log_error and bail out.

src/python/platforms/android/emulator.py
```diff
@@ -0,0 +1,126 @@
...
+    archive_dst_path = os.path.join(temp_directory, 'emulator_bundle.zip')
+
+    storage.copy_file_from(archive_src_path, archive_dst_path)
+
+    # Extract emulator image.
+    self.emu_path = os.path.join(temp_directory, 'emulator')
```
inferno-chromium, line 62: s/emu_path/emulator_path [we prefer full var names]

src/python/platforms/android/emulator.py
```diff
@@ -0,0 +1,126 @@
...
+    archive.unpack(archive_dst_path, self.emu_path)
+    os.remove(archive_dst_path)
+
+    # Run emulator
+    script_path = os.path.join(
+        self.emu_path, environment.get_value('ANDROID_EMULATOR_SCRIPT_PATH'))
```
inferno-chromium, line 68: Do we need another ANDROID_EMULATOR_SCRIPT_PATH, will this change often ?

src/python/platforms/android/emulator.py
```diff
@@ -0,0 +1,126 @@
...
+
+    self.process_runner = new_process.ProcessRunner(script_path, [])
```
inferno-chromium, line 70: Can remove [] arg, not needed.

src/python/platforms/android/emulator.py
```diff
@@ -0,0 +1,126 @@
...
+      logs.log('Stopping emulator.')
+      self.popen.kill()
+      self.popen = None
+
+    if self.logfile:
+      self.logfile.close()
+      self.logfile = None
+
+    if self.emu_path:
+      shutil.rmtree(self.emu_path)
```
inferno-chromium, line 105: use shell.remove_directory

src/python/platforms/android/emulator.py
```diff
@@ -0,0 +1,126 @@
...
+
+    archive_src_path = environment.get_value('ANDROID_EMULATOR_BUCKET_PATH')
+    archive_dst_path = os.path.join(temp_directory, 'emulator_bundle.zip')
+
+    storage.copy_file_from(archive_src_path, archive_dst_path)
+
+    # Extract emulator image.
+    self.emu_path = os.path.join(temp_directory, 'emulator')
+    archive.unpack(archive_dst_path, self.emu_path)
+    os.remove(archive_dst_path)
```
inferno-chromium, line 64: use shell.remove_file

src/python/platforms/android/emulator.py
```diff
@@ -0,0 +1,126 @@
...
+def start_emulator():
+  """Start emulator."""
+  global _emu_proc
+  global _emu_users
+  if _emu_users == 0:
+    _emu_proc = EmulatorProcess()
+    _emu_proc.create()
+    _emu_proc.run()
+    adb.run_as_root()
+  _emu_users += 1
```
inferno-chromium, line 117: Will we have more emu_users ? this logic seems to only account for one.

src/python/platforms/android/emulator.py
```diff
@@ -0,0 +1,126 @@
...
+
+    devices_before = adb.get_devices()
+    new_device = False
+
+    logs.log('Starting emulator.')
+    self.popen = self.process_runner.run(
+        stdout=self.logfile, stderr=subprocess.PIPE)
+
+    logs.log('Waiting for emulated device to come online.')
+    while not new_device:
```
inferno-chromium, line 85: Will we be starting one emulator at a time, this looks racy if another script is running and trying to start at same time. There should be some way to specify our own serial for emulator or prefix, then can just wait on it.

inferno-chromium (changes requested):

src/python/platforms/android/emulator.py
kalder: The problem here is that we don't or can't know in advance what the serial of the emulated device will be, since it's determined when the emulator starts.

src/python/base/tasks.py
```diff
@@ -63,6 +63,7 @@
     'ANDROID_KERNEL': 'Android Kernel',
     'ANDROID_AUTO': 'Android Auto',
     'ANDROID_X86': 'Android (x86)',
+    'EMULATED_ANDROID': 'Android (Emulated)',
```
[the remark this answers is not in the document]
kalder: It actually just checks for if the string contains 'ANDROID', but I think it makes sense to change for consistency anyway.

src/python/platforms/android/emulator.py
kalder: Will there be more than one task running per bot? I assumed that we had 1 bot/task per process, and that we would thus be okay with a single emulator.

src/python/platforms/android/emulator.py
kalder: I suppose we can always expect that the emulator bundle includes a top-level script called "start" or "run" and just hard-code it.

src/python/platforms/android/emulator.py
inferno-chromium: not per bot, but there can be multiple cf bots per host right, wont they conflict with one another.

inferno-chromium: why we need a counter on _emu_users when it can only have 0 or 1 ?

src/python/platforms/android/emulator.py
```diff
@@ -0,0 +1,125 @@
...
+    self.logfile = open(log_path, 'wb')
+
+  def create(self):
+    """Configures a emulator process which can subsequently be `run`."""
+    # Download emulator image.
+    if not environment.get_value('BOT_TMPDIR'):
+      logs.log_error('BOT_TMPDIR is not set.')
+      return
+    temp_directory = environment.get_value('BOT_TMPDIR')
+    archive_src_path = environment.get_value('ANDROID_EMULATOR_BUCKET_PATH')
```
oliverchang, line 58: can we do this in build_manager instead, similar to FuchsiaBuild.

kalder: Moved it.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1297,6 +1299,28 @@ def cleanse_crash(self,
...
+    """
+    start_emulator()
+
+    super().__init__(
+        executable_path=executable_path,
+        build_directory=build_directory,
+        default_args=default_args)
+
+  def __del__(self):
+    stop_emulator()
```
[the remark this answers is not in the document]
kalder: I think it makes sense to do it in build_manager. I've moved it there.

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -37,6 +37,8 @@
 from metrics import logs
 from platforms import android
 from platforms import fuchsia
+from platforms.android.emulator import start_emulator
```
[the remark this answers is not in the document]
kalder: It isn't imported here since I moved it to build_manager.py.

src/python/platforms/android/emulator.py
kalder: Good point. I've modified the startup script to print out the device serial number, which is determined at runtime, and added code CF to read it.

src/python/platforms/android/emulator.py
kalder: The counter has been removed in the update that reads the child process's stdout.

kalder:
Please take another look. I've moved things into build_manager.py and made some changes so that we know which DEVICE_SERIAL to wait on and use.

src/python/platforms/android/adb.py
```diff
@@ -571,7 +571,7 @@ def run_command(cmd,
   if (output in [
       DEVICE_HANG_STRING, DEVICE_OFFLINE_STRING,
       device_not_found_string_with_serial
-  ]):
+  ]) and not environment.is_android_emulator():
```
inferno-chromium, line 574: i think this needs to be moved up,

if not recover or environment.is_android_emulator():
on line 564
since basically we don't want this to trigger on emulator?

src/python/bot/tasks/update_task.py
```diff
@@ -171,7 +171,8 @@ def run_platform_init_scripts():
 
   plt = environment.platform()
   if environment.is_android():
-    android_init.run()
+    if not environment.is_android_emulator():
```
inferno-chromium, line 174: Forgot to ask, why are we doing this, this bailout needs to be added in the specific init function call, like what breaks with emulator ? Is it this ? https://github.com/google/clusterfuzz/blob/4290129370d0d3c6198ae752e4e4eff2b59621a6/src/python/bot/init_scripts/android.py#L25

maybe bailout needs to be added at start of flash_to_latest_build_if_needed ?
basically some functions in that android_init might be useful, like restarting device? so we shouldn't just skip it all.

src/python/system/shell.py
```diff
@@ -144,7 +144,7 @@ def clear_data_directories_on_low_disk_space():
 
 def clear_device_temp_directories():
   """Clear device specific temp directories."""
-  if environment.is_android():
+  if environment.is_android() and not environment.is_android_emulator():
```
inferno-chromium, line 147: I dont understand this, why won't we want to clear temp dirs on android emulator, we might not be spinning up emulator on every task. same for clear_testcase_directory below.

src/python/system/environment.py
```diff
@@ -471,6 +471,8 @@ def get_platform_id():
   """Return a platform id as a lowercase string."""
   bot_platform = platform()
   if is_android(bot_platform):
+    if is_android_emulator():
```
inferno-chromium, line 474: move it to start, why put it inside "if is_android" ?

src/python/platforms/android/emulator.py
```diff
@@ -0,0 +1,110 @@
...
+      line = self.process.popen.stdout.readline().decode()
+      match = DEVICE_SERIAL_RE.match(line)
+      if match:
+        device_serial = match.group(1)
+
+    logs.log('Found serial ID: %s.' % device_serial)
+    environment.set_value('ANDROID_SERIAL', device_serial)
+
+    logs.log('Waiting on device')
+    while adb.get_device_state() != 'device':
```
inferno-chromium, line 92: instead of this, i think android.adb.wait_until_fully_booted()

src/python/platforms/android/emulator.py
inferno-chromium: I meant to check on ANDROID_EMULATOR_BUCKET_PATH and not on BOT_TMPDIR. BOT_TMPDIR is always set.

inferno-chromium (changes requested):
Some more questions, fixes needed.
@oliverchang PTAL as well.

src/python/platforms/android/emulator.py
kalder: That depends on some things that don't exist for emulated devices, specifically the 'sys.boot_completed' property.

kalder: I also considered using adb.wait_for_device(), but that fails if the device hasn't been registered with ADB yet.

src/python/system/shell.py
kalder: I added this because it can be called before the device is started (commands.py:382) which will cause errors. Instead, I can more directly check for the emulator being started or the device being available by checking for the ANDROID_SERIAL variable being set.

src/python/platforms/android/adb.py
kalder: I agree. I've made the change.

src/python/platforms/android/emulator.py
inferno-chromium: Still need to fix this @kalder

src/python/system/environment.py
inferno-chromium: still needs fixing.

src/python/system/shell.py
```diff
@@ -216,7 +216,7 @@ def clear_testcase_directories():
   remove_directory(environment.get_value('FUZZ_INPUTS'), recreate=True)
   remove_directory(environment.get_value('FUZZ_INPUTS_DISK'), recreate=True)
 
-  if environment.is_android():
+  if environment.is_android() and not environment.is_android_emulator():
```
inferno-chromium, line 219: this need "and environment.get_value('ANDROID_SERIAL')" as well

src/python/bot/tasks/update_task.py
inferno-chromium: Ok!

src/python/system/shell.py
inferno-chromium: Ok!

inferno-chromium:
ok see the 4 above comments from this round. once those are fixed, this looks ready to go.

src/python/system/environment.py
kalder: It should be in the latest commit (10).

PR #2270, opened 2021-03-03
Add support for manual security/severity markers

fmeum:
Any stack trace that contains the substring "Fuzzer-Security-Issue"
(ignoring case) is marked as a security issue. By default, the severity
is set to High. If desired, the severity can be specified by including
a substring of the form "Fuzzer-Security-Issue-Medium" (again ignoring
case).


src/python/crash_analysis/severity_analyzer.py
```diff
@@ -104,8 +104,20 @@ def get_security_severity(crash_type, crash_output, job_name,
...
+  def analyze(self, crash_type, crash_output, requires_gestures):
     """Return a security severity based on the ASan crash output."""
+
+    manual_severity_match = re.search(
+        r'fuzzer-security-issue(-critical|-high|-medium|-low|)', crash_output,
+        re.IGNORECASE)
+    if manual_severity_match:
+      manual_severity = manual_severity_match.groups()[0].lower()
+      if manual_severity:
+        # Skip over the leading dash.
```
jonathanmetzman, line 116: This seems a bit hacky. Why is this needed?

src/python/crash_analysis/severity_analyzer.py
```diff
@@ -104,8 +104,20 @@ def get_security_severity(crash_type, crash_output, job_name,
 class SeverityAnalyzerSanitizer(object):
   """Generic ASan severity analyzer."""
 
-  def analyze(self, crash_type, _, requires_gestures):
+  def analyze(self, crash_type, crash_output, requires_gestures):
```
jonathanmetzman, line 107: @inferno-chromium what do you think of this?
Maybe it's better to have exceptions for SecurityExceptionCritical, SecurityExceptionHigh, SecurityExceptionMedium, SecurityExceptionLow rather than make this API more complex.

src/python/crash_analysis/severity_analyzer.py
```diff
@@ -104,8 +104,20 @@ def get_security_severity(crash_type, crash_output, job_name,
...
+    manual_severity_match = re.search(
+        r'fuzzer-security-issue(-critical|-high|-medium|-low|)', crash_output,
+        re.IGNORECASE)
+    if manual_severity_match:
+      manual_severity = manual_severity_match.groups()[0].lower()
+      if manual_severity:
+        # Skip over the leading dash.
+        return string_to_severity(manual_severity[1:])
+      # Assign severity HIGH if none is specified explicitly.
+      return SecuritySeverity.HIGH
```
jonathanmetzman, line 119: Why don't you do this for low, critical, and medium severitys?

src/python/crash_analysis/severity_analyzer.py
fmeum: This is done to support both `fuzzer-security-bug` (no severity specified, just that this is a security issue) and `fuzzer-security-bug-critical` (severity specified). I could either

1. make the code less condensed by checking for `fuzzer-security-bug-<severity>` first and if not found checking for `fuzzer-security-bug` or
2. just drop the `fuzzer-security-bug` feature so that fuzz target authors need to specify a severity as well.

src/python/crash_analysis/severity_analyzer.py
fmeum: Could you clarify your question? This line returns a default severity if no severity is specified via the marker. I chose HIGH because that is also the default in the UI.

Should this rather be a `MISSING` or just `None`?

oliverchang: Let's use None here. We shouldn't try to guess the severity and instead encourage explicit severities in the string from the target.

oliverchang (commented):
thank you for contributing this change!

src/python/crash_analysis/severity_analyzer.py
fmeum: The problem went away per your other comment ;-)

src/python/crash_analysis/severity_analyzer.py
fmeum: I simplified the design as per your other comment. As the severity has to be specified explicitly now, this special case no longer exists.

src/python/crash_analysis/crash_analyzer.py
```diff
@@ -240,6 +240,10 @@ def has_signal_for_non_security_bug_type(stacktrace):
 def is_security_issue(crash_stacktrace, crash_type, crash_address):
   """Based on unsymbolized crash parameters, determine whether it has security
   consequences or not."""
+  # Stack traces of any type can be manually labelled as a security issue.
+  if re.search('fuzzer-security-issue', crash_stacktrace, re.IGNORECASE):
```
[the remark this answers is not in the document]
fmeum: Makes sense. I dropped the `re.IGNORECASE` to enforce better readability of the marker now that it is all letters.

src/python/crash_analysis/severity_analyzer.py
jonathanmetzman: For the target.
I mean that right now we are treating SecurityExceptionCritical, SecurityExceptionHigh etc all as `uncaught exception` crash type.
So to determine the severity of uncaught exception we need to pass crash_output to `analyze`.
But instead of doing that, when determining crash type, we can make the type `SecurityExceptionCritical` or `SecurityExceptionHigh` etc if needed.

src/python/crash_analysis/severity_analyzer.py
```diff
@@ -104,8 +104,15 @@ def get_security_severity(crash_type, crash_output, job_name,
...
   """Generic ASan severity analyzer."""
 
-  def analyze(self, crash_type, _, requires_gestures):
+  def analyze(self, crash_type, crash_output, requires_gestures):
     """Return a security severity based on the ASan crash output."""
+
+    manual_severity_match = re.search(
+        r'FuzzerSecurityIssue(Critical|High|Medium|Low)', crash_output)
+    if manual_severity_match:
+      manual_severity = manual_severity_match.groups()[0]
```
oliverchang, line 113: nit: use `group(1)` instead.

oliverchang (approved):
Thanks! LGTM with a last nit.

src/python/crash_analysis/severity_analyzer.py
fmeum: Done, thanks!

PR #2277, opened 2021-03-17
Fix testcase uploads for OSS-Fuzz.

oliverchang:
Previously platform IDs were being incorrectly set to `project-linux`
when then should just be e.g. `linux`. This only affects OSS-Fuzz.

This was reported in https://github.com/google/oss-fuzz/issues/5388

### DIFF
--- a/src/python/datastore/data_handler.py
+++ b/src/python/datastore/data_handler.py
@@ -1272,6 +1272,10 @@ def create_user_uploaded_testcase(key,
         utils.current_date_time(), uploader_email)
     # External jobs never get minimized.
     testcase.minimized_keys = 'NA'
+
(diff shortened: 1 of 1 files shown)


oliverchang:
/gcbrun. TBR

inferno-chromium (approved):
Do any testcase need migration?

oliverchang:
> Do any testcase need migration?

Yeah there's a small number (2-3 pages) of uploaded testcases I'll fix up.

inferno-chromium:
> > Do any testcase need migration?
> 
> Yeah there's a small number (2-3 pages) of uploaded testcases I'll fix up.

not a big deal, can even just delete. (just uploads)

PR #2282, opened 2021-03-22
ignore symbolicated libc with versions

demoray:
As is, `libc.so` is ignored, but `libc-2.31.so` is not.  This PR adds ignoring libc versions used by glibc.

Ref: [glibc release timeline](https://sourceware.org/glibc/wiki/Glibc%20Timeline)

### DIFF
--- a/src/python/lib/clusterfuzz/stacktraces/constants.py
+++ b/src/python/lib/clusterfuzz/stacktraces/constants.py
@@ -534,6 +534,7 @@ STACK_FRAME_IGNORE_REGEXES_IF_SYMBOLIZED = [
     r'.*libc\+\+\.so',
     r'.*libc\+\+_shared\.so',
     r'.*libstdc\+\+\.so',
+    r'.*libc-\d+\.\d+(?:\.\d+)?\.so',
 ]
 
 IGNORE_CRASH_TYPES_FOR_ABRT_BREAKPOINT_AND_ILLS = [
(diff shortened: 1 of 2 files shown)


src/python/lib/clusterfuzz/stacktraces/constants.py
```diff
@@ -534,6 +534,7 @@
     r'.*libc\+\+\.so',
     r'.*libc\+\+_shared\.so',
     r'.*libstdc\+\+\.so',
+    r'.*libc-\d+\.\d+(?:\.\d+)?\.so',
```
inferno-chromium, line 537: Isnt this too specific, how about just
i think libc-.*\.so should be better?

demoray: I was assuming more conservative on the validation, but it's fine with me either way.

I've updated it per your request.

inferno-chromium (approved):

demoray:
I've updated the unit test that expected `/lib/x86_64-linux-gnu/libc-2.15.so` in the minimized stack.

Note, I left the parenthesis, `expected_state = ('/tmp/coredump\n')`, even though this is now just a simple string, as multiple other entries followed this style:

https://github.com/google/clusterfuzz/blob/03aeb6662f4c99264f2bc8ec3dd0415bd70965e7/src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py#L200
https://github.com/google/clusterfuzz/blob/03aeb6662f4c99264f2bc8ec3dd0415bd70965e7/src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py#L214
https://github.com/google/clusterfuzz/blob/03aeb6662f4c99264f2bc8ec3dd0415bd70965e7/src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py#L353

demoray:
I've updated the unit test to match the generated output.  I mistakenly thought the check for cycles would have ignored one of the `/tmp/coredump\n` entries. 

ref:
https://github.com/google/clusterfuzz/blob/03aeb6662f4c99264f2bc8ec3dd0415bd70965e7/src/python/tests/core/crash_analysis/stack_parsing/stack_analyzer_data/sanitizer_signal_abrt.txt#L3-L7

PR #2283, opened 2021-03-23
change android-emulator to write to build_dir, not $BOT_TMPDIR

kalder:
The call to self._pre_setup() is needed to create the build directory as needed.

### DIFF
--- a/src/python/build_management/build_manager.py
+++ b/src/python/build_management/build_manager.py
@@ -864,8 +864,9 @@ class AndroidEmulatorBuild(RegularBuild):
 
   def setup(self):
     """Android Emulator build setup."""
+    self._pre_setup()
     emu_proc = android.emulator.EmulatorProcess()
-    emu_proc.create()
+    emu_proc.create(self.build_dir)
     emu_proc.run()
     android.adb.run_as_root()
 
--- a/src/python/platforms/android/emulator.py
+++ b/src/python/platforms/android/emulator.py
@@ -51,19 +51,18 @@ class EmulatorProcess(object):
     log_path = os.path.join(tempfile.gettempdir(), 'android-emulator.log')
     self.logfile = open(log_path, 'wb')
 
-  def create(self):
+  def create(self, work_dir):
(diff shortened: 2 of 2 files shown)


src/python/build_management/build_manager.py
```diff
@@ -864,8 +864,9 @@ class AndroidEmulatorBuild(RegularBuild):
 
   def setup(self):
     """Android Emulator build setup."""
+    self._pre_setup()
     emu_proc = android.emulator.EmulatorProcess()
-    emu_proc.create()
+    emu_proc.create(self.build_dir)
```
inferno-chromium, line 869: We don't clean up build directory if revision does not change. Will this cause build directory to swell causing out of disk space?

oliverchang: I think this always gets unpacked to the same place, so should not build up right @kalder  ?

kalder: Yes, it is always unpacked to the same place and there already exists cleanup code (in emulator.py) to remove the archive immediately after unpacking it and the resulting emulator image and data during the EmulatorProcess's kill() method.

src/python/platforms/android/emulator.py
```diff
@@ -51,19 +51,18 @@ def __init__(self):
...
     archive_src_path = environment.get_value('ANDROID_EMULATOR_BUCKET_PATH')
-    archive_dst_path = os.path.join(temp_directory, 'emulator_bundle.zip')
+    archive_dst_path = os.path.join(work_dir, 'emulator_bundle.zip')
     storage.copy_file_from(archive_src_path, archive_dst_path)
 
     # Extract emulator image.
-    self.emulator_path = os.path.join(temp_directory, 'emulator')
+    self.emulator_path = os.path.join(work_dir, 'emulator')
     archive.unpack(archive_dst_path, self.emulator_path)
     shell.remove_file(archive_dst_path)
```
kalder, line 67: This is where the archive gets cleaned up.

inferno-chromium (approved):

PR #2284, opened 2021-03-23
Store emulator images in base_build_path instead of build_path

kalder:


src/python/build_management/build_manager.py
```diff
@@ -866,7 +866,7 @@ def setup(self):
     """Android Emulator build setup."""
     self._pre_setup()
     emu_proc = android.emulator.EmulatorProcess()
-    emu_proc.create(self.build_dir)
+    emu_proc.create(self.base_build_dir)
```
oliverchang, line 869: things in `base_build_dir` live for longer and things don't get cleaned up in between revisions.

Can we delete any previously extracted files prior to extracting as well in https://github.com/google/clusterfuzz/blob/master/src/python/platforms/android/emulator.py#L66 ?

kalder: Done.

oliverchang (approved):

PR #2259, opened 2021-02-24
afl++ integration

vanhauser-thc:
This integrates afl++ into clusterfuzz.
Review of @jonathanmetzman has been incorporated, unit tests pass.

PR #2293, opened 2021-04-01
[AFL] AFL++ integration

jonathanmetzman:


jonathanmetzman:
Superseeds: https://github.com/google/clusterfuzz/pull/2259

inferno-chromium (approved):

oliverchang:
Nice! Just curious here: Is this integration compatible with previous vanilla AFL builds?

vanhauser-thc:
@oliverchang yes you can keep old builds. but afl++ supports better instrumentation, especially for large targets, so rebuilding with afl-clang-fast from afl++ is recommended. note that oss-fuzz was moved to afl++ 8 weeks ago.

PR #2296, opened 2021-04-07
Add 'killall' Package to base/Dockerfile

kalder:


oliverchang (approved):
thanks!

inferno-chromium:
why do you need this, cant you use https://github.com/google/clusterfuzz/blob/master/src/python/system/process_handler.py#L601 ?

oliverchang:
> why do you need this, cant you use https://github.com/google/clusterfuzz/blob/master/src/python/system/process_handler.py#L601 ?

I believe this is called from an external shell script. Seems like something we should have in the base image (and we call it from other places too like https://github.com/google/clusterfuzz/blob/f9b0b8c684cdf57cdf999c34eff4aea07dffa8f6/src/python/system/process_handler.py#L330)

inferno-chromium:
> > why do you need this, cant you use https://github.com/google/clusterfuzz/blob/master/src/python/system/process_handler.py#L601 ?
> 
> I believe this is called from an external shell script. Seems like something we should have in the base image (and we call it from other places too like
> 
> https://github.com/google/clusterfuzz/blob/f9b0b8c684cdf57cdf999c34eff4aea07dffa8f6/src/python/system/process_handler.py#L330
> 
> )

ah in chromium, we get it from its install script. makes sense to push in base image.

PR #2298, opened 2021-04-08
Fix Android Emulator Hangs

kalder:
In some cases, Android Emulators will hang because of output buffering. Closing the pipe fixes this.

There's also some unused code (stderr pipe, log tempfile) that should be removed.


oliverchang (approved):
nice! do we need to prevent stderr from filling up too?

src/python/platforms/android/emulator.py
```diff
@@ -82,8 +77,7 @@ def run(self):
       raise EmulatorError('Attempted to `run` emulator before calling `create`')
 
     logs.log('Starting emulator.')
-    self.process = self.process_runner.run(
-        stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    self.process = self.process_runner.run(stdout=subprocess.PIPE)
```
oliverchang, line 80: Since we don't care about stderr, let's set stderr to open(os.devnull, 'w').

kalder: Done, but I'm using subprocess.DEVNULL.

PR #2303, opened 2021-04-13
Export Datastore backups to a /latest/ directory as well.

oliverchang:
For BQ exports.

Also clean up previous error handling logic. If an exception occurs, just let it bubble up
and have it handled by our error reporting.

### DIFF
--- a/src/appengine/handlers/cron/backup.py
+++ b/src/appengine/handlers/cron/backup.py
@@ -47,7 +47,7 @@ class Handler(base_handler.Handler):
         local_config.PROJECT_PATH).get('backup.bucket')
     if not backup_bucket:
       logs.log('No backup bucket is set, skipping.')
-      return 'OK'
+      return
 
     kinds = [
         kind for kind in ndb.Model._kind_map  # pylint: disable=protected-access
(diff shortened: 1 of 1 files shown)


src/appengine/handlers/cron/backup.py
```diff
@@ -47,36 +47,31 @@ def get(self):
...
-      message = 'Datastore export succeeded.'
-      status_code = 200
-      logs.log(message, response=response)
-    except googleapiclient.errors.HttpError as e:
-      message = 'Datastore export failed.'
-      status_code = e.resp.status
-      logs.log_error(message, error=str(e))
-
-    return (message, status_code, {'Content-Type': 'text/plain'})
+      logs.log('Datastore export succeeded.', response=response)
```
inferno-chromium, line 77: I think you remove the error so that cron failure gets seen more prominently in error reporting right ?

inferno-chromium (approved):

src/appengine/handlers/cron/backup.py
oliverchang: Yep, if we get an exception we'll just let it show in error reporting.

PR #2304, opened 2021-04-13
Add test for swift stacktrace parsing.

oliverchang:
Also add a ignore frame pattern to ignore stdlib functions.


inferno-chromium (approved):
Thanks!

PR #2316, opened 2021-04-22
Support chrome extended stable in issue filling logic

yuanjunh-git:
1. Add extended stable attributes in datastore schema
2. Support extended stable impact label in issue filling logic

### DIFF
--- a/src/appengine/libs/issue_management/issue_filer.py
+++ b/src/appengine/libs/issue_management/issue_filer.py
@@ -164,8 +164,9 @@ def update_issue_impact_labels(testcase, issue):
 
   if testcase.regression.startswith('0:'):
     # If the regression range starts from the start of time,
-    # then we assume that the bug impacts extended stable.
-    new_impact = data_types.SecurityImpact.EXTENDED_STABLE
+    # then we assume that the bug impacts stable.
+    # TODO(yuanjunh): change to extended stable label when it's fully supported.
+    new_impact = data_types.SecurityImpact.STABLE
   elif testcase.is_impact_set_flag:
     # Add impact label based on testcase's impact value.
     if testcase.impact_extended_stable_version:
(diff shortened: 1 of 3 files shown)


src/python/datastore/data_types.py
```diff
@@ -190,6 +190,7 @@ class SecurityImpact(object):
   HEAD = 2
   NONE = 3
   MISSING = 4
+  EXTENDED_STABLE = 5
```
oliverchang, line 193: nit: put this as the first value. since these values aren't actually stored anywhere, we can change this to make EXTENDED_STABLE = 0.

oliverchang (approved):
Nice! LGTM with some nits

src/appengine/libs/issue_management/issue_filer.py
```diff
@@ -161,19 +164,21 @@ def update_issue_impact_labels(testcase, issue):
 
   if testcase.regression.startswith('0:'):
     # If the regression range starts from the start of time,
-    # then we assume that the bug impacts stable.
-    new_impact = data_types.SecurityImpact.STABLE
+    # then we assume that the bug impacts extended stable.
+    new_impact = data_types.SecurityImpact.EXTENDED_STABLE
```
oliverchang, line 168: Also: do we need to give a head's up to any PMs before we do this? 

They may have scripts today that depend on `Security_Impact-Stable` and not yet ExtendedStable.

src/python/tests/appengine/libs/issue_filer_test.py
```diff
@@ -646,7 +646,19 @@ def test_update_impact_stable_from_regression(self):
     mock_issue = self._make_mock_issue()
 
     issue_filer.update_issue_impact_labels(self.testcase, mock_issue)
-    six.assertCountEqual(self, ['Security_Impact-Stable'],
+    six.assertCountEqual(self, ['Security_Impact-ExtendedStable'],
```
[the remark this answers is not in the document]
yuanjunh-git: undo this change because we still label it as 'Stable' for regression range which starts from 0.

src/appengine/libs/issue_management/issue_filer.py
yuanjunh-git: Reversed this piece of code as there are some scripts reply this label.

oliverchang (approved):

PR #2317, opened 2021-04-26
Support Python 3.9 for development.

oliverchang:
- Remove tensorflow from Pipfile as it repeatedly causes issues when a
  new Python version needs to be supported. It's not required for the
  majority of ClusterFuzz devs and most things run fine without it.
- Upgrade grpcio.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -9,12 +9,10 @@ crcmod = "==1.7"
 future = "==0.17.1"
 protobuf = "==3.15.3"
 psutil = "==5.7.0"
-numpy = "==1.18.4"
-tensorflow = "==2.3.0"
 
 [dev-packages]
 Fabric = "==1.14.1"
-grpcio-tools = "==1.28.1"
+grpcio-tools = "==1.37.0"
 gunicorn = "*"
 mock = "==2.0.0"
 nodeenv = "==1.0.0"
--- a/local/install_deps_linux.bash
+++ b/local/install_deps_linux.bash
@@ -29,6 +29,8 @@ done
 
 if [ -z "$PYTHON" ]; then
   if which python3.8 > /dev/null; then
+    PYTHON='python3.9'
+  elif which python3.8 > /dev/null; then
     PYTHON='python3.8'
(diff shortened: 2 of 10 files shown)


local/install_deps_linux.bash
```diff
@@ -43,8 +45,8 @@ if ! which "$PYTHON" > /dev/null; then
 fi
 
 version=$($PYTHON --version 2>&1 | cut -f2 -d' ')
-if [[ "$version" < "3.7" || ! "$version" < "3.9" ]]; then
-  echo "You need Python 3.8 or 3.7. Try \`export PYTHON=python3.8\` (or 3.7)."
+if [[ "$version" < "3.7" || ! "$version" < "3.a" ]]; then
```
oliverchang, line 48: 3.a is just a hack for the comparison to work (as 3.10 comes before 3.9 lexicographically).

Pipfile
```diff
@@ -9,12 +9,10 @@ crcmod = "==1.7"
 future = "==0.17.1"
 protobuf = "==3.15.3"
 psutil = "==5.7.0"
-numpy = "==1.18.4"
-tensorflow = "==2.3.0"
```
inferno-chromium, line 13: How will testcase generation work if this is removed - https://github.com/google/clusterfuzz/blob/481c78b28fb2d9a401a6f3fe302746a8f6110954/src/python/bot/fuzzers/ml/rnn/generate.py#L21 ?

inferno-chromium (approved):

Pipfile
oliverchang: Good point! I mistakenly thought we were installing this somewhere else in the Docker process, but that's using this Pipfile as well. 

I added a line to install tensorflow in the base image separately from this Pipfile install.

inferno-chromium: might be still a problem for chromium mac,android bots, but we can even disable it there ? thoughts like try catch it.

oliverchang: I added a try/catch to the place importing the generator to log_warn and exit if we can't import it.

PR #2319, opened 2021-04-28
Report generic fatal errors as 'Fatal error'

backes:
We were previously overwriting the type with 'CHECK failure' even if
there was no 'Check failed' line in the output.


oliverchang (approved):
thanks!

PR #2320, opened 2021-04-30
Add chrome extended stable support in impact task executing flow

yuanjunh-git:
It will try to get extended stable build and calculate the impact. But it will silence any exceptions if it fails to get extend stable build. So it won't break the current CF flows even this code is deployed.

### DIFF
--- a/src/python/bot/tasks/impact_task.py
+++ b/src/python/bot/tasks/impact_task.py
@@ -24,6 +24,7 @@ from build_management import build_manager
 from build_management import revisions
 from datastore import data_handler
 from datastore import data_types
+from metrics import logs
 from system import environment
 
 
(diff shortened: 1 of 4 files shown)


src/python/bot/tasks/impact_task.py
```diff
@@ -125,12 +130,17 @@ def get_component_impacts_from_url(component_name,
...
 
+  build_revision_mappings = revisions.get_build_to_revision_mappings(platform)
+  if not build_revision_mappings:
+    return Impacts()
+
   found_impacts = dict()
-  for build in ['stable', 'beta']:
-    build_revision_mappings = revisions.get_build_to_revision_mappings(platform)
-    if not build_revision_mappings:
-      return Impacts()
```
yuanjunh-git, line 132: get_build_to_revision_mappings() doesn't need to be called multiple times. So I extracted it out of the for-loop.

yuanjunh-git:
@oliverchang PTAL, thanks!

src/python/build_management/build_manager.py
```diff
@@ -1426,7 +1425,10 @@ def setup_custom_binary(target_weights=None):
 def setup_production_build(build_type):
   """Sets up build with a particular revision."""
   # Bail out if there are not stable and beta build urls.
-  if build_type == 'stable':
+  if build_type == 'extended_stable':
+    build_bucket_path = environment.get_value(
+        'EXTENDED_STABLE_BUILD_BUCKET_PATH')
```
oliverchang, line 1430: Can you `return None` here for now if this is None? Otherwise we'll log_error below all the time (resulting in email alerts).

oliverchang (approved):
nice! just one comment.

src/python/build_management/build_manager.py
yuanjunh-git: Done.

yuanjunh-git:
@oliverchang addressed the comment. Can you help to merge it?

oliverchang (approved):

PR #2321, opened 2021-05-03
Find CHECK failures also without source file

backes:
The block of code to replace a generic 'Fatal error' by the more
specific 'CHECK failure' was only executed if we had a source file for
the fatal error. In d8 / chrome release builds, we won't have a source
file though, but we still want to extract the CHECK failure.


backes:
The diff looks bigger than it is. The patch is removing an outer "if state.check_failure_source_file:" and instead adding two such conditions further down.

oliverchang (approved):

PR #2323, opened 2021-05-04
Update Testcase timestamp when analyze_task completes.

oliverchang:
Triage task waits data_types.MIN_ELAPSED_TIME_SINCE_REPORT for a
testcase to make sure it is considered for any grouping before we file
a bug. This is based on the `timestamp` attribute.

However, for uploaded testcases, the timestamp was previously set to the
time of upload, rather than the time when analyze_task completes and
marks the Testcase as "Processed". This means these testcases could get
filed without being grouped.

Also increase MIN_ELAPSED_TIME_SINCE_REPORT by an hour to be safe, as
the grouper can be quite slow with lots of open testcases.

### DIFF
--- a/src/appengine/handlers/cron/triage.py
+++ b/src/appengine/handlers/cron/triage.py
@@ -293,8 +293,12 @@ class Handler(base_handler.Handler):
       if not data_handler.critical_tasks_completed(testcase):
         continue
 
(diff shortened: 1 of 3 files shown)


inferno-chromium (approved):

inferno-chromium:
Thanks a lot!

PR #2328, opened 2021-05-06
Restart emulator

kalder:


src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1491,6 +1491,117 @@ def cleanse_crash(self,
...
+  # pylint: disable=super-init-not-called
+  def __init__(self, executable_path, build_directory, default_args=None):
+    """Inits the AndroidEmulatorLibFuzzerRunner.
+
+    Args:
+      executable_path: Path to the fuzzer executable.
+      build_directory: A MinijailChroot.
+      default_args: Default arguments to always pass to the fuzzer.
+    """
+    self.exec_path = executable_path
```
oliverchang, line 1506: nit: rather than naming these slightly differently, do

```python
self._init_args = {
   'executable_path': executable_path,
  ...
}
```

Then later you can do

```python
super().__init__(**self._init_args)
```

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1491,6 +1491,117 @@ def cleanse_crash(self,
...
+    self.build_dir = build_directory
+    self.def_args = default_args
+    # We don't call super().__init__() here since we need to wait until
```
oliverchang, line 1509: What's the reason why we need to do this? 

Can we refactor the base class instead to avoid doing this?

src/python/bot/fuzzers/libfuzzer.py
```diff
@@ -1491,6 +1491,117 @@ def cleanse_crash(self,
...
+    android.adb.run_as_root()
+    super().__init__(
+        executable_path=self.exec_path,
+        build_directory=self.build_dir,
+        default_args=self.def_args)
+
+  def _stop_emulator(self):
+    self.emu_proc.kill()
+
+  def analyze_dictionary(self,
```
oliverchang, line 1525: Rather than repeating all of these, might be better to do something like (untested):

```python

def wrap_emulator(func):
  @functools.wraps(func)
  def wrapper(self, *args, **kwargs):
    self._start_emulator()
    result = func(self, *args, **kwargs)
    self._stop_emulator()

  return wrapper

analyze_dictionary = wrap_emulator(AndroidLibFuzzerRunner.analyze_dictionary)
fuzz = wrap_emulator(AndroidLibFuzzerRunner.fuzz)
...

```

oliverchang (commented):
Thanks, this looks good overall, but could you please give some context for why this change is need?

src/python/bot/fuzzers/libfuzzer.py
kalder: I reworked this to call the parent `__init__()` function here, but I do still need the `build_dir` to be saved to start the emulator.

src/python/bot/fuzzers/libfuzzer.py
kalder: I've change this to just call the required functionality (copying files to the device) directly.

src/python/bot/fuzzers/libfuzzer.py
kalder: Done. I inlined the _start_emulator() and _stop_emulator() functions as well, since they aren't used anywhere else.

oliverchang (approved):
nice, thanks!

PR #2334, opened 2021-05-11
Wrap another psutil import with a try/except.

oliverchang:


oliverchang:
tbr to fix an exception.

PR #2341, opened 2021-05-18
Add support for Android Cuttlefish flash

Huiziyang:
This change is based on the new design that Clusterfuzz and Cuttlefish will be started in separate docker containers under same network.
To flash the device, Clusterfuzz will fetch the latest Cuttlefish builds from Android build system and launch the cvd on Cuttlefish via **ssh**.


Huiziyang:
@googlebot I fixed it

Huiziyang:
@googlebot I signed it!

Huiziyang:
@googlebot I fixed it.

Huiziyang:
@googlebot I signed it!

src/python/system/environment.py
```diff
@@ -1056,6 +1056,11 @@ def is_android(plt=None):
   return 'ANDROID' in (plt or platform())
 
 
+def is_android_gce(plt=None):
```
oliverchang, line 1059: nit: rename to is_android_cuttlefish to be clearer.

src/python/platforms/android/adb.py
```diff
@@ -424,6 +442,25 @@ def reset_device_connection():
 
   return True
 
+def get_gce_device_ip():
+  try:
+    return socket.gethostbyname('cuttlefish')
+  except socket.gaierror as _:
```
oliverchang, line 448: nit: no need for the "as _" part.

src/python/platforms/android/adb.py
```diff
@@ -424,6 +442,25 @@ def reset_device_connection():
...
+def set_gce_device_serial():
+  device_serial = '%s:6520' % get_gce_device_ip()
+  environment.set_value('ANDROID_SERIAL', device_serial)
+
+
+def get_gce_address():
+  ip = get_gce_device_ip()
+  if ip == '127.0.0.1':
+    return ''
+  return 'vsoc-01@%s' % ip
```
oliverchang, line 462: where does vsoc-01 come from? perhaps put it in a constant somewhere in this file.

src/python/platforms/android/adb.py
```diff
@@ -371,16 +372,33 @@ def recreate_gce_device():
...
-    shell.remove_file(image_file_path)
-
-  # Restore images from backup.
-  backup_image_dir = os.path.join(cvd_dir, 'backup')
-  for image_filename in os.listdir(backup_image_dir):
-    image_src = os.path.join(backup_image_dir, image_filename)
-    image_dest = os.path.join(image_dir, image_filename)
-    shell.copy_file(image_src, image_dest)
+  cvd_address = get_gce_address()
+  rm_cmd = 'rm -rf {cvd_dir}/*'.format(cvd_dir=cvd_dir)
```
oliverchang, line 376: just use f-strings. i.e. f'rm -rf {cvd_dir}/*' 

same for other cases in this file.

src/python/platforms/android/adb.py
```diff
@@ -344,22 +339,28 @@ def start_gce_device():
...
   launch_cvd_path = os.path.join(cvd_bin_dir, 'launch_cvd')
+  cvd_address = get_gce_address()
 
   device_memory_mb = environment.get_value('DEVICE_MEMORY_MB',
                                            DEFAULT_DEVICE_MEMORY_MB)
   launch_cvd_command_line = (
       '{launch_cvd_path} -daemon -memory_mb {device_memory_mb}'.format(
           launch_cvd_path=launch_cvd_path, device_memory_mb=device_memory_mb))
+  if cvd_address:
+    launch_cvd_command_line = 'ssh %s %s' % (cvd_address, launch_cvd_command_line)
```
oliverchang, line 350: Can we quote the `launch_cvd_command_line` part?

src/python/platforms/android/adb.py
```diff
@@ -424,6 +442,25 @@ def reset_device_connection():
...
+def get_gce_device_ip():
+  try:
+    return socket.gethostbyname('cuttlefish')
+  except socket.gaierror as _:
+    logs.log_warn('Failed to get cuttlefish ip address, use localhost instead')
+    return '127.0.0.1'
+
+
+def set_gce_device_serial():
+  device_serial = '%s:6520' % get_gce_device_ip()
```
oliverchang, line 454: Can we put 6520 into a constant somewhere? 

And then use f-strings to build the resulting string.

src/python/platforms/android/adb.py
```diff
@@ -371,16 +372,33 @@ def recreate_gce_device():
...
+  if cvd_address:
+    rm_cmd = 'ssh %s %s' % (cvd_address, rm_cmd)
+  execute_command(rm_cmd, timeout=RECOVERY_CMD_TIMEOUT)
+
+  # Copy and Combine cvd host package and OTA images.
+  image_directory = environment.get_value('IMAGES_DIR')
+  for image_filename in os.listdir(image_directory):
+    image_src = os.path.join(image_directory, image_filename)
+    if cvd_address:
+      if image_src.endswith('.tar.gz'):
```
oliverchang, line 386: There's a lot of duplication here. Let's consolidate this into the one block for both cuttlefish and non-cuttlefish, by making helpers as described above.

src/python/platforms/android/adb.py
```diff
@@ -371,16 +372,33 @@ def recreate_gce_device():
...
-
-  # Restore images from backup.
-  backup_image_dir = os.path.join(cvd_dir, 'backup')
-  for image_filename in os.listdir(backup_image_dir):
-    image_src = os.path.join(backup_image_dir, image_filename)
-    image_dest = os.path.join(image_dir, image_filename)
-    shell.copy_file(image_src, image_dest)
+  cvd_address = get_gce_address()
+  rm_cmd = 'rm -rf {cvd_dir}/*'.format(cvd_dir=cvd_dir)
+  if cvd_address:
```
oliverchang, line 377: This makes it hard to distinguish between a configuration error vs a non-cuttlefish bot. 

We should error out instead, if we are on a cuttlefish bot and the address does not exist.

Also, let's extend execute_command and add a wrapper copy_file. e.g.

```python
def execute_command(..., on_cuttlefish_host=False):
   if on_cuttlefish_host:
        cvd_address = get_gce_address()
        if not cvd_address:
          # raise an exception
    
        cmd = f'ssh {cvd_address} "{cmd}"'
      
   ...


def copy_file(..., on_cuttlefish_host=False):
  # similar
```

This will help consolidate/simplify a lot of the existing code paths.

oliverchang (changes requested):
nice! looks mostly good with some fixes needed.

src/python/system/environment.py
Huiziyang: Done

src/python/platforms/android/adb.py
Huiziyang: Done

src/python/platforms/android/adb.py
Huiziyang: Done

src/python/platforms/android/adb.py
Huiziyang: Done

src/python/platforms/android/adb.py
Huiziyang: Done, not sure if I understand it correctly, copy file has special cases like _tar_  which is has different options on local and remote

src/python/platforms/android/adb.py
Huiziyang: Done

src/python/platforms/android/adb.py
```diff
@@ -62,6 +65,13 @@ def bad_state_reached():
       'Device in bad state.', wait_before_exit=BAD_STATE_WAIT)
 
 
+def connect_to_gce_device():
+  """Connect to Cuttlefish cvd."""
+  device_serial = environment.get_value('ANDROID_SERIAL')
+  connect_cmd = '%s connect %s' % (get_adb_path(), device_serial)
```
oliverchang, line 71: use f-string here. f'{get_adb_path()} connect {device_serial}'.

src/python/platforms/android/adb.py
```diff
@@ -100,8 +110,18 @@ def directory_exists(directory_path):
   return result == expected
 
 
-def execute_command(cmd, timeout=None, log_error=True):
+def execute_command(cmd, timeout=None, on_cuttlefish_host=False, log_error=True):
   """Spawns a subprocess to run the given shell command."""
+  if on_cuttlefish_host:
+    if environment.is_android_cuttlefish():
```
oliverchang, line 116: these two ifs can be merged.

src/python/platforms/android/adb.py
```diff
@@ -147,9 +167,32 @@ def run():
   return bytes_output.strip().decode('utf-8', errors='ignore')
 
 
+def copy_file(filename, src_dir, dest_dir, timeout=None):
```
oliverchang, line 170: change the arguments to 

`copy_file_to_cuttlefish_host(src_path, dest_path, timeout=None)` here (where both are full paths).

Also, we only need to handle the cuttlefish case here right? This is called below from `recreate_gce_device`, which is only called on cuttelfish.

src/python/platforms/android/adb.py
```diff
@@ -147,9 +167,32 @@ def run():
...
+  """Copy file between two locations"""
+  file_src = os.path.join(src_dir, filename)
+  if environment.is_android_cuttlefish():
+    cvd_address = get_gce_address()
+    if not cvd_address:
```
oliverchang, line 175: rather than duplicate these checks, can we just log_fatal inside `get_gce_address` ?

src/python/platforms/android/adb.py
```diff
@@ -147,9 +167,32 @@ def run():
...
+      logs.log_fatal_and_exit(
+          'Unable to get cvd ip address on cuttlefish host')
+
+    if file_src.endswith('.tar.gz'):
+      cmd = f'ssh {cvd_address} tar xzvf - < {file_src}'
```
oliverchang, line 180: We shouldn't be unpacking in a function called "copy_file". If we want to unpack, this should be a separate function

Since this is only called in the cuttlefish context, we just need the one case here: 

```python
def copy_to_cuttlefish(src_path, dest_path, timeout=None)
  """Copy file between to the cuttlefish host"""
  cvd_address = get_gce_address()
  return execute_command(f'scp -R {src_path} {cvd_address}:{dest_path}', timeout=timeout)
```

I think you also need the "-r" to make directory copying work.

As below, can we untar in `download_latest_build` instead and avoid doing it here?

src/python/platforms/android/adb.py
```diff
@@ -371,16 +408,15 @@ def recreate_gce_device():
   stop_gce_device()
 
   # Delete all existing images.
-  image_dir = cvd_dir
-  for image_file_path in glob.glob(os.path.join(image_dir, '*.img')):
-    shell.remove_file(image_file_path)
+  rm_cmd = f'rm -rf {cvd_dir}/*'
```
oliverchang, line 411: can you rename `recreate_gce_device` etc to `recreate_cuttlefish_device` ?

src/python/platforms/android/adb.py
```diff
@@ -424,6 +460,25 @@ def reset_device_connection():
 
   return True
 
+def get_gce_device_ip():
```
oliverchang, line 463: please rename all instances of these `_gce_` functions to `_cuttlefish_`

src/python/platforms/android/flash.py
```diff
@@ -51,6 +55,36 @@
...
+
+    for image_regex in image_regexes:
+      image_file_path = fetch_artifact.get(build_id, target, image_regex,
+                                            image_directory)
+      if not image_file_path:
+        logs.log_error(
+            'Failed to download artifact %s for '
+            'branch %s and target %s.' % (image_file_path, branch, target))
+        return
+      if image_file_path.endswith('.zip'):
```
oliverchang, line 85: why don't we untar here as well instead of in adb.py ?

oliverchang (changes requested):
please also fix the lint and formatting issues.

src/python/platforms/android/adb.py
Huiziyang: My initial intent is to still keep the old logic that cuttlefish runs locally with clusterfuzz on host machine.
But we can add later if do need that case, removed for now.

src/python/platforms/android/flash.py
Huiziyang: I wanted to make the download and copy faster, but agree if we untar here the code is cleaner.

src/python/platforms/android/adb.py
```diff
@@ -32,6 +33,8 @@
 ADB_TIMEOUT = 1200  # Should be lower than |REBOOT_TIMEOUT|.
 BAD_STATE_WAIT = 900
 BOOT_WAIT_INTERVAL = 30
+CUTTLEFISH_USER='vsoc-01'
```
oliverchang, line 36: Space before after and after "="

src/python/platforms/android/adb.py
```diff
@@ -419,11 +435,31 @@ def reset_device_connection():
...
+
+def set_cuttlefish_device_serial():
+  device_serial = '%s:%d' % (get_cuttlefish_device_ip(), CUTTLEFISH_CVD_PORT)
+  environment.set_value('ANDROID_SERIAL', device_serial)
+
+
+def get_cuttlefish_address():
+  ip = get_cuttlefish_device_ip()
+  if ip == '127.0.0.1':
+      logs.log_fatal_and_exit(
```
oliverchang, line 459: If `127.0.0.1` is expected to be an error, then we can just do this in `get_cuttlefish_device_ip` instead, rather than warning and returning it if we're going to end up calling log_fatal anyway.

src/python/platforms/android/adb.py
```diff
@@ -419,11 +435,31 @@ def reset_device_connection():
   state = get_device_state()
   if state != 'device':
     logs.log_warn('Device state is %s, unable to recover using usb reset/'
-                  'gce reconnect.' % str(state))
+                  'cuttlefish reconnect.' % str(state))
     return False
 
   return True
 
+def get_cuttlefish_device_ip():
```
oliverchang, line 443: These all need docstrings to describe these functions.

Please run `python butler.py lint` and `python butler.py format` to resolve many formatting/linting issues in your PR.

src/python/platforms/android/adb.py
```diff
@@ -419,11 +435,31 @@ def reset_device_connection():
...
+    logs.log_warn('Failed to get cuttlefish ip address, use localhost instead')
+    return '127.0.0.1'
+
+
+def set_cuttlefish_device_serial():
+  device_serial = '%s:%d' % (get_cuttlefish_device_ip(), CUTTLEFISH_CVD_PORT)
+  environment.set_value('ANDROID_SERIAL', device_serial)
+
+
+def get_cuttlefish_address():
```
oliverchang, line 456: nit: rename to `get_cuttlefish_ssh_target` to be more specific.

src/python/platforms/android/flash.py
```diff
@@ -52,6 +56,31 @@
...
 
+def download_latest_build(build_info, image_regexes):
+  """Download the latest build artifact for the given branch and target."""
+  # Check if our local build matches the latest build. If not, we will
+  # download it.
+  build_id = build_info['bid']
+  target = build_info['target']
+  image_directory = environment.get_value('IMAGES_DIR')
+  last_build_info = persistent_cache.get_value(constants.LAST_FLASH_BUILD_KEY)
+  if not last_build_info or last_build_info['bid'] != build_id:
```
oliverchang, line 67: nit: invert this condition and early return to avoid the extra indent.

i.e. `if last_build_info and last_build_info['bid'] == build_id:`

src/python/platforms/android/flash.py
```diff
@@ -52,6 +56,31 @@
 FLASH_REBOOT_WAIT = 5 * 60
 
 
+def download_latest_build(build_info, image_regexes):
+  """Download the latest build artifact for the given branch and target."""
+  # Check if our local build matches the latest build. If not, we will
+  # download it.
+  build_id = build_info['bid']
+  target = build_info['target']
+  image_directory = environment.get_value('IMAGES_DIR')
```
oliverchang, line 65: read this env var once from the caller, and have it passed as an argument rather than reading it again here.

oliverchang (approved):
nice! LGTM with minor fixes needed. Please also rebase and fix the conflicts.

src/python/platforms/android/adb.py
```diff
@@ -419,12 +439,32 @@ def reset_device_connection():
...
+                  'cuttlefish reconnect.' % str(state))
     return False
 
   return True
 
 
+def get_cuttlefish_device_ip(ip=None):
+  """Return the ip address of cuttlefish device."""
+  try:
+    ip = socket.gethostbyname('cuttlefish')
```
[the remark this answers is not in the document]
Huiziyang: If I return directly the lint will report "inconsistent-return-statements".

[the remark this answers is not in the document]
oliverchang: Just add a `return None` at the end of the function in that case.

src/python/platforms/android/flash.py
Huiziyang: Done

src/python/platforms/android/adb.py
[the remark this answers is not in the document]
Huiziyang: Done

PR #2342, opened 2021-05-18
[fuchsia] undercoat fixups

eepeep:
Addresses a few outstanding issues that weren't caught by integration tests.

### DIFF
--- a/src/python/platforms/fuchsia/undercoat.py
+++ b/src/python/platforms/fuchsia/undercoat.py
@@ -188,5 +188,8 @@ def get_data(handle, fuzzer, src, dst):
   """Get files from a fuzzer on an instance, via undercoat.
   If src is a directory, it will be copied recursively. Standard globs are
   supported."""
-  return undercoat_instance_command('get_data', handle, '-fuzzer', fuzzer,
-                                    '-src', src, '-dst', dst).output
+  try:
+    return undercoat_instance_command('get_data', handle, '-fuzzer', fuzzer,
+                                      '-src', src, '-dst', dst, abort_on_error=False).output
+  except UndercoatError:
+    return None
--- a/src/python/platforms/fuchsia/undercoat.py
+++ b/src/python/platforms/fuchsia/undercoat.py
@@ -54,6 +54,7 @@ class UndercoatError(Exception):
 
 def undercoat_api_command(*args):
   """Make an API call to the undercoat binary."""
+  logs.log('Running undercoat command %s' % (args,))
   bundle_dir = environment.get_value('FUCHSIA_RESOURCES_DIR')
   undercoat_path = os.path.join(bundle_dir, 'undercoat', 'undercoat')
   undercoat = new_process.ProcessRunner(undercoat_path, args)
@@ -74,15 +75,16 @@ def undercoat_api_command(*args):
   return result
 
 
-def undercoat_instance_command(command, handle, *args):
+def undercoat_instance_command(command, handle, *args, abort_on_error=True):
   """Helper for the subset of undercoat commands that operate on an instance."""
   try:
     return undercoat_api_command(command, '-handle', handle, *args)
   except UndercoatError:
-    # Try to print extra logs and shut down
-    # TODO(eep): Should we be attempting to automatically restart?
-    dump_instance_logs(handle)
-    stop_instance(handle)
+    if abort_on_error:
+      # Try to print extra logs and shut down
+      # TODO(eep): Should we be attempting to automatically restart?
+      dump_instance_logs(handle)
+      stop_instance(handle)
     raise
 
 
@@ -114,8 +116,7 @@ def validate_api_version():
 
 def dump_instance_logs(handle):
   """Dump logs from an undercoat instance."""
-  # Avoids using undercoat_instance_command in order to avoid recursion on error
-  qemu_log = undercoat_api_command('get_logs', '-handle', handle).output
+  qemu_log = undercoat_instance_command('get_logs', handle, abort_on_error=False).output
   logs.log_warn(qemu_log)
 
 
@@ -136,7 +137,7 @@ def stop_all():
   cleanly shut down."""
   for handle in get_running_handles():
     try:
-      undercoat_instance_command('stop_instance', handle)
+      undercoat_instance_command('stop_instance', handle, abort_on_error=False)
     except UndercoatError:
       pass
 
@@ -147,8 +148,7 @@ def stop_all():
 
 def stop_instance(handle):
   """Stop a running undercoat instance."""
-  # Avoids using undercoat_instance_command in order to avoid recursion on error
-  result = undercoat_api_command('stop_instance', '-handle', handle)
+  result = undercoat_instance_command('stop_instance', handle, abort_on_error=False)
 
   # Mark the corresponding handle as having been cleanly shut down
   remove_running_handle(handle)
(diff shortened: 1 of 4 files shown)


eepeep:
The `TMPDIR` solution is probably not final; this is a draft pending discussion of a proper approach with @oliverchang.

src/python/platforms/fuchsia/undercoat.py
```diff
@@ -54,12 +54,14 @@ class UndercoatError(Exception):
...
   """Make an API call to the undercoat binary."""
+  logs.log('Running undercoat command %s' % (args,))
   bundle_dir = environment.get_value('FUCHSIA_RESOURCES_DIR')
   undercoat_path = os.path.join(bundle_dir, 'undercoat', 'undercoat')
   undercoat = new_process.ProcessRunner(undercoat_path, args)
   # The undercoat log is sent to stderr, which we capture to a tempfile
   with tempfile.TemporaryFile() as undercoat_log:
-    result = undercoat.run_and_wait(stderr=undercoat_log)
+    result = undercoat.run_and_wait(
+        stderr=undercoat_log, extra_env={"TMPDIR": "/tmp"})
```
oliverchang, line 64: nit: single quotes for strings.

also, let's use bot.fuzzer.utils.get_temp_dir() (or a subdirectory of it). This is scoped to the task though. What implications would this have?

oliverchang (commented):
Nice! LGTM overall, just a suggestion on the TMPDIR  issue as we discussed.

src/python/platforms/fuchsia/undercoat.py
eepeep: Ah, whoops! I keep missing these because I rely on the linter to catch stuff like this.

Hm, I believe being task-scoped means that if a task gets interrupted there will be no way for the next run to clean up after the old stale qemu process/disk images. 

That said, looking at the `utils.get_temp_dir()` implementation it seems to basically be scope-equivalent to `tempfile.gettempdir()` which just resolves to the `TMPDIR` envvar in this case. This means I'm pretty sure it would still have the issue I was trying to address here, which was having the files get cleared out within a single task. If this seems wrong to you, though, I can try to dig deeper into debugging why that would be happening.

Any other alternatives you can think of? Could they live in the build dir perhaps?

oliverchang: I think we are looking at different functions :)

This is the one I'm talking about: https://github.com/google/clusterfuzz/blob/00cc394d2e0274537f1bc9bc108cad1fd4aa11c2/src/python/bot/fuzzers/utils.py#L130

If this doesn't work though, you'd need to manage your own dir (and make sure you clean it up at the right places to prevent buildup over time).

Something like `os.path.join(environment.get_value('ROOT_DIR'), 'bot', 'undercoat')`. ROOT_DIR/bot already exists, but you'll need to create the undercoat directory if it does not already exist before passing it here.

eepeep: Hm, yeah, that's where I was looking. I'm assuming `FUZZ_INPUTS_DISK` usually isn't set, so it's just a subdir of wherever `tempfile.gettempdir()` points, no? And then https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir lists "1. The directory named by the TMPDIR environment variable." Sorry if I'm totally misunderstanding here.

For the `ROOT_DIR` route, I'll need to think more about what those "right places" are. Maybe it'd be enough to just clear the directory right after running `fuchsia.undercoat.stop_all()` in build_manager? Any handles still left there wouldn't be useful so should be safe to delete them.

oliverchang: FUZZ_INPUTS_DISK is always set for a bot. The fallback is for calling this from outside the context of a bot.

And yeah, `fuchsia.undercoat.stop_all()` seems like a reasonable place. As long as there is something that occasionally clears it, the frequency doesn't have to be too high or fine grained.

eepeep: I couldn't get the `fuzzer.utils` method to work, so I went ahead with the `ROOT_DIR` subdir. LMK if that seems good. Thanks for your help sorting this out!

src/python/platforms/fuchsia/undercoat.py
```diff
@@ -188,5 +208,16 @@ def get_data(handle, fuzzer, src, dst):
...
+        handle,
+        '-fuzzer',
+        fuzzer,
+        '-src',
+        src,
+        '-dst',
+        dst,
+        abort_on_error=False).output
+  except UndercoatError:
+    return None
```
oliverchang, line 223: we don't want to log here?

src/python/platforms/fuchsia/undercoat.py
```diff
@@ -48,18 +49,32 @@ def get_running_handles():
   return persistent_cache.get_value(HANDLE_CACHE_KEY, default_value=[])
 
 
+def get_temp_dir():
+  """Define a tempdir for undercoat to store its data in.
+
+  This tempdir needs to be of a scope that persists across invocations of the
+  bot, to ensure proper cleanup of stale handles/data."""
+  tmpdir = os.path.join(environment.get_value('ROOT_DIR'), 'bot', 'undercoat')
+  if not os.path.exists(tmpdir):
```
oliverchang, line 58: nit: `os.makedirs(tmp_dir, exist_ok=True)` to be more concise.

src/python/platforms/fuchsia/undercoat.py
```diff
@@ -48,18 +49,32 @@ def get_running_handles():
   return persistent_cache.get_value(HANDLE_CACHE_KEY, default_value=[])
 
 
+def get_temp_dir():
+  """Define a tempdir for undercoat to store its data in.
+
+  This tempdir needs to be of a scope that persists across invocations of the
+  bot, to ensure proper cleanup of stale handles/data."""
+  tmpdir = os.path.join(environment.get_value('ROOT_DIR'), 'bot', 'undercoat')
```
oliverchang, line 57: nit: name this temp_dir to be consistent with the function name.

src/python/platforms/fuchsia/undercoat.py
```diff
@@ -48,18 +49,32 @@ def get_running_handles():
...
+  return tmpdir
+
+
 class UndercoatError(Exception):
   """Error for errors while running undercoat."""
 
 
 def undercoat_api_command(*args):
   """Make an API call to the undercoat binary."""
+  logs.log('Running undercoat command %s' % (args,))
```
oliverchang, line 70: Just do `f'Running undercoat command {args}'` here.

oliverchang (approved):
Thanks! LGTM with minor nits.

src/python/platforms/fuchsia/undercoat.py
eepeep: The `logs.log_warn` in the inner `undercoat_api_command` will be triggered by the non-zero exit code, so anything here would be redundant. Sound good enough?

src/python/platforms/fuchsia/undercoat.py
eepeep: Done

src/python/platforms/fuchsia/undercoat.py
oliverchang: Thanks, SGTM.

PR #2345, opened 2021-05-20
    wait for emulator to come up using wait_for_device()

kalder:
Emulators are hanging in production because the one remaining condition in `wait_until_fully_booted()` doesn't happen for emulated devices. Eventually it times out and `bad_state_reached()` is called.
    
This changes the emulator start code to call `wait_for_device()` instead and removes the special-case logic from `wait_until_fully_booted()`.
    
It also adds some additional checks to reset code so that code which doesn't work for emulated devices won't be called.


src/python/platforms/android/emulator.py
```diff
@@ -71,7 +71,7 @@ def run(self):
     environment.set_value('ANDROID_SERIAL', device_serial)
 
     logs.log('Waiting on device')
-    adb.wait_until_fully_booted()
```
oliverchang, line 74: Rather than lose the drive check, can we instead modify wait_until_fully_booted to customize the checks that are done?

e.g. `def wait_until_fully_booted(check_drive=True, check_package_manager=True, etc...)`

And just set the two that don't work to false here.

oliverchang (approved):

PR #2346, opened 2021-05-21
[oss-fuzz] Use flexbox container on OSS-Fuzz home

fmeum:
Using a flexbox container for the projects on the OSS-Fuzz home page allows up to three projects to be shown in a row instead of listing them all beneath each other.

Also fixes an instance of incorrect tag nesting and reindents in separate commits to keep the diff clean.

Before:
before
After:
after

### DIFF
--- a/src/appengine/private/components/oss-fuzz-home/oss-fuzz-home.html
+++ b/src/appengine/private/components/oss-fuzz-home/oss-fuzz-home.html
@@ -92,8 +92,8 @@
                </if-else>
              </template>
             </div>
-          </template>
-        </paper-card>
+          </paper-card>
+        </template>
       </div>
       </div>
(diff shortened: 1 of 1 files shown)


inferno-chromium:
sorry, but we dont want to mix projects on the same row, it is very confusing and most projects (c/c++) have three builds, so they dont see empty space.

fmeum:
> sorry, but we dont want to mix projects on the same row, it is very confusing and most projects (c/c++) have three builds, so they dont see empty space.

I see, that explains why it's only me who sees the empty space. If you want, you can merge https://github.com/google/clusterfuzz/pull/2346/commits/53ace93b137d8574754f8f6a8b84e46f57667397 by itself.

PR #2359, opened 2021-06-17
Replace Omahaproxy to ChromiumDash in build_info

yuanjunh-git:
The new ChromiumDash endpoint is production ready. But it's currently missing the branch_base_position. So migrate to the new endpoint in build_info.py first. After the commit position is ready, I'll revise the revision.py file as well.


yuanjunh-git:
@oliverchang can you help to review? thanks.

src/python/chrome/build_info.py
```diff
@@ -75,14 +86,48 @@ def get_production_builds_info(platform):
...
+  query_url = BUILD_INFO_URL_CD % chromiumdash_platform
+
+  build_info = utils.fetch_url(query_url)
+  if not build_info:
+    logs.log_error('Failed to fetch build info from %s' % query_url)
+    return []
+
+  try:
+    build_info_json = json.loads(build_info)
+    if len(build_info_json) == 0:
```
oliverchang, line 107: nit: just `if not build_info_json`

src/python/chrome/build_info.py
```diff
@@ -75,14 +86,48 @@ def get_production_builds_info(platform):
...
+      logs.log_error('Empty response from %s' % query_url)
+      return []
+  except Exception as e:
+    logs.log_error('Malformed response from %s: %s' % (query_url, str(e)))
```
oliverchang, line 111: No need to include the exception in the string. Our log_error will automatically include it.

src/python/scripts/other-bots/chromium-builder/run.py
```diff
@@ -52,8 +52,16 @@ def main():
   build_dir = environment.get_value('BUILD_DIR')
   wait_time = environment.get_value('WAIT_TIME')
 
-  builds_metadata = build_info.get_production_builds_info(
-      environment.platform())
+  try:
+    builds_metadata = build_info.get_production_builds_info_from_cd(
+        environment.platform())
+  except Exception as e:
+    logs.log_error('Errors when fetching from ChromiumDash: %s' % str(e))
```
oliverchang, line 59: same here, no need to format the exception.

src/python/chrome/build_info.py
```diff
@@ -43,6 +45,15 @@ def _convert_platform_to_omahaproxy_platform(platform):
   return platform_lower
 
 
+def _convert_platform_to_chromiumdash_platform(platform):
+  """Converts platform to Chromium Dash platform.
+  Note that Windows in Chromium Dash is win64 and we only want win32."""
+  platform_lower = platform.lower()
+  if platform_lower in ('windows', 'win'):
```
oliverchang, line 52: why do we need to compare against 'win'? Shouldn't be needed?

src/python/chrome/build_info.py
```diff
@@ -43,6 +45,15 @@ def _convert_platform_to_omahaproxy_platform(platform):
...
+    return 'Win32'
+  return platform_lower.capitalize()
+
+
 def get_production_builds_info(platform):
```
oliverchang, line 57: Since this is no longer referenced after this PR, let's delete this in this PR as well. Same with `_convert_platform_to_omahaproxy_platform`, `BUILD_INFO_PATTERN`/`URL` and anything else.

src/python/chrome/build_info.py
```diff
@@ -22,6 +23,7 @@
                       '[^,]*,[^,]*,[^,]*,[^,]*,[^,]*,'
                       '([0-9a-f]+),.*')
 BUILD_INFO_URL = 'https://omahaproxy.appspot.com/all?csv=1'
+BUILD_INFO_URL_CD = 'https://chromiumdash.appspot.com/fetch_releases?num=1&platform=%s'  # pylint: disable=line-too-long
```
oliverchang, line 26: Can we use a named format for this. i.e. "platform={platform}". Then later `str.format(platform=...)`

Can split up this line by doing

```python
BUILD_INFO_URL_CD = (
   'https://chromiumdash.appspot.com/fetch_releases?'
   'num=1&platform={platform}')
```

src/python/tests/core/chrome/build_info_test.py
```diff
@@ -34,6 +37,18 @@ def setUp(self):
     def _fetch_url(url):
       if url == build_info.BUILD_INFO_URL:
         return open(ALL_CSV, 'r').read()
+
+      match = re.match(
+          r'https://chromiumdash\.appspot\.com/fetch_releases\?num=1&platform=([a-zA-Z0-9]+)',  # pylint: disable=line-too-long
+          url)
+      if match:
```
oliverchang, line 44: nit: `if not match` then `return None` to avoid extra indentaiton.

src/python/tests/core/chrome/build_info_test.py
```diff
@@ -34,6 +37,18 @@ def setUp(self):
     def _fetch_url(url):
       if url == build_info.BUILD_INFO_URL:
         return open(ALL_CSV, 'r').read()
+
+      match = re.match(
+          r'https://chromiumdash\.appspot\.com/fetch_releases\?num=1&platform=([a-zA-Z0-9]+)',  # pylint: disable=line-too-long
```
oliverchang, line 42: Can split up this line over multiple lines just by breaking them up.

```
re.match(
   r'...'
   r'...') 
```

etc

oliverchang (approved):
Nice! Thank you for making these changes and coordinating with chromiumdash. 

LG overall, just some small fixes needed.

src/python/chrome/build_info.py
yuanjunh-git: chatted offline. These are needed in the fallback logic for now.

src/python/chrome/build_info.py
yuanjunh-git: Omahaproxy uses `win` while ChromiumDash use `Win32`. And in `get_release_milestone()`, it passes `win` as the platform. Not sure if anywhere in the code/config files also use this format. So I convert both formats here.

yuanjunh-git:
@oliverchang addressed all comments, PTAL again.

src/python/chrome/build_info.py
oliverchang: I see, that's a bit confusing!

In that case let's fix `get_release_milestone()` to pass 'windows' instead. Everywhere else in ClusterFuzz, we generally use 'windows'. Nowhere else should be passing 'win'.

oliverchang (approved):

src/python/chrome/build_info.py
yuanjunh-git: Done

yuanjunh-git:
@oliverchang changed the `win` to `windows`. Please merge.

oliverchang:
thanks!

PR #2363, opened 2021-06-23
Support Android Cuttlefish kernel fuzzing

Huiziyang:


src/python/platforms/android/adb.py
```diff
@@ -370,6 +370,12 @@ def start_cuttlefish_device():
   launch_cvd_command_line = (
       f'{launch_cvd_path} -daemon -memory_mb {device_memory_mb} '
       '-report_anonymous_usage_stats Y')
+  if use_kernel:
+    kernel_path = os.path.join(cvd_dir, 'bzImage')
+    initramfs_path = os.path.join(cvd_dir, 'initramfs.img')
+    launch_cvd_command_line += (
+        f' -kernel_path={kernel_path} -initramfs_path={initramfs_path}')
```
oliverchang, line 377: where's the default one used if these aren't set?

src/python/build_management/build_manager.py
```diff
@@ -864,6 +864,38 @@ def setup(self):
...
+    result = super().setup()
+    if not result:
+      return result
+
+    environment.set_value('VMLINUX_PATH', self.build_dir)
+
+    cvd_dir = environment.get_value('CVD_DIR')
+    adb.stop_cuttlefish_device()
+
+    for image_filename in ['bzImage', 'initramfs.img']:
```
oliverchang, line 883: nit: put `['bzImage', 'initramfs.img']` as a class constant (and use a tuple instead).

i.e.

```python
class CuttlefishKernelBuild(..):

  _IMAGE_FILES = ('bzImage', 'initramfs.img')
```

src/python/build_management/build_manager.py
```diff
@@ -864,6 +864,38 @@ def setup(self):
...
+      # Delete existing kernel image.
+      rm_cmd = f'rm {cvd_dir}/{image_filename}'
+      adb.execute_command(
+          rm_cmd, timeout=adb.RECOVERY_CMD_TIMEOUT, on_cuttlefish_host=True)
```
oliverchang, line 887: do we need to delete if we're overwriting this anyway on line 892? 

if so, we should also move `image_dest` to before line 885 and use that in the rm command instead.

src/python/bot/fuzzers/engine_common.py
```diff
@@ -312,7 +312,8 @@ def find_fuzzer_path(build_directory, fuzzer_name, is_blackbox=False):
     # Fuchsia targets are not on disk.
     return fuzzer_name
 
-  if environment.platform() == 'ANDROID_KERNEL':
+  if (environment.platform() == 'ANDROID_KERNEL' or
```
oliverchang, line 315: Better to compare to 'ANDROID_KERNEL_X86' instead of `environment.is_kernel_fuzzer_job()` ? 

Should also create a environment.is_android_kernel() which does this and call that here instead.

oliverchang (commented):
nice! some initial comments. did you figure out the srcmap stuff?

src/python/bot/fuzzers/engine_common.py
Huiziyang: Added environment.is_android_kernel() that also check QUEUE_OVERRIDE because there will be a case where os_override=android_x86 but running kernel jobs.

src/python/build_management/build_manager.py
Huiziyang: Not necessary, deleted.

src/python/platforms/android/adb.py
Huiziyang: There is a default one built in Cuttlefish boot img, we need to override it with KASAN build to make kernel fuzzing work.

Huiziyang:
> nice! some initial comments. did you figure out the srcmap stuff?

Yeah I created the srcmaps with url prefix "https://android.googlesource.com" so Clusterfuzz linkify should work

src/python/build_management/build_manager.py
```diff
@@ -864,6 +864,49 @@ def setup(self):
...
+  def setup(self):
+    """Android kernel build setup."""
+    from platforms.android import adb
+
+    result = super().setup()
+    if not result:
+      return result
+
+    # Download syzkaller binary folder.
+    if not environment.get_value('ANDROID_EMULATOR_BUCKET_PATH'):
```
oliverchang, line 881: let's not overload ANDROID_EMULATOR_BUCKET_PATH. Call this SYZKALLER_BUCKET_PATH instead.

src/python/system/environment.py
```diff
@@ -1066,6 +1066,14 @@ def is_android_emulator(plt=None):
   return 'ANDROID_EMULATOR' == (plt or platform())
 
 
+def is_android_kernel(plt=None):
+  """Return true if we are running android kernel jobs."""
+  if not plt:
+    queue_override = get_value('QUEUE_OVERRIDE')
+    plt = queue_override if queue_override else platform()
+  return 'ANDROID_KERNEL' in plt
```
oliverchang, line 1074: Just return `'ANDROID_KERNEL' in platform_group()`. `platform_group` does the same thing you are doing here.

oliverchang (commented):
Can you fix the failing test in https://github.com/google/clusterfuzz/runs/2937924289?check_suite_focus=true ?

oliverchang (approved):
LGTM with a question.

src/python/build_management/build_manager.py
```diff
@@ -1356,6 +1399,8 @@ def setup_regular_build(revision,
     build_class = FuchsiaBuild
   elif environment.is_android_emulator():
     build_class = AndroidEmulatorBuild
+  elif environment.is_kernel_fuzzer_job():
```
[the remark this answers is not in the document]
Huiziyang: Our existing physical syzkaller fuzzing is using custom build so it won't go to setup_regular_build. The logic here is that if syzkaller job doesn't provide custom build, it then uses CuttlefishKernelBuild.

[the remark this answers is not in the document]
oliverchang: I see! Still good to be explicit here and check for cuttlefish, in case this ever changes in the future.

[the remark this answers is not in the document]
Huiziyang: Done.

PR #2368, opened 2021-06-29
Fix timestamp to datetime conversion in bot_run_timed_out.

oliverchang:
This broke as part of the Python 3 migration, which has the following
behaviour:

>>> d = datetime.datetime.utcnow()
>>> ts = d.timestamp()
>>> d2 = datetime.datetime.utcfromtimestamp(ts)
>>> d == d2
False  # ???

utcfromtimestamp() is incorrect and actually converts the datetime to
the local time.

### DIFF
--- a/src/python/datastore/data_handler.py
+++ b/src/python/datastore/data_handler.py
@@ -1101,7 +1101,7 @@ def bot_run_timed_out():
   if not start_time:
     return False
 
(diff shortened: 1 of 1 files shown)


oliverchang:
TBR. other places are most likely broken too, but leaving those for a future PR instead to avoid breaking too many things.

jonathanmetzman:
LGTM

PR #2370, opened 2021-06-30
More datetime fixes.

oliverchang:
Use time.time() instead of datetime.datetime.utcnow().timestamp()
which causes issues depending on the timezone.

This causes issues because the timezone changes after we call
environment.set_bot_environment().


jonathanmetzman (commented):
LGTM
Thanks!

PR #2371, opened 2021-07-01
Fix chromium builder PATH.

oliverchang:
This was causing the chromium builder to fail.


yuanjunh-git (approved):
LTGM. Thanks for the quick fix.

inferno-chromium (commented):
lgtm

jonathanmetzman:
LGTM. Thanks!

PR #2374, opened 2021-07-07
Not set options variable for KASAN.

Huiziyang:


src/python/system/environment.py
```diff
@@ -874,9 +874,10 @@ def reset_current_memory_tool_options(redzone_size=0,
     set_value('UBSAN_OPTIONS', joined_tool_options)
 
   # For Android, we need to set shell property |asan.options|.
-  # For engine-based uzzers, it is not needed as options variable is directly
+  # For engine-based fuzzers, it is not needed as options variable is directly
   # passed to shell.
-  if is_android(bot_platform) and not is_engine_fuzzer_job():
+  if (is_android(bot_platform) and not is_engine_fuzzer_job() and
+      not is_kernel_fuzzer_job()):
```
oliverchang, line 880: can we handle this inside `android.sanitizer` instead? it's better to keep these implementation-specific hacks closer to the place where they're actually used, rather than a general place (here).

oliverchang:
Is this PR still needed?

Huiziyang:
> Is this PR still needed?

Yes, it doesn't affect the functioning so I kind of left it behind. I will address the comments by this week!

oliverchang (approved):

PR #2383, opened 2021-07-23
Restructure ClusterFuzz to be a proper package. 

### LINKED ISSUE #2192: libClusterFuzz accidentally imports modules from dependent projects
To reproduce:
```
pip install clusterfuzz==0.0.1a0
mkdir -p /tmp/myproject
echo "1/0" > /tmp/myproject/metrics.py
cd /tmp/myproject
python -c 'from clusterfuzz import stacktraces'
```

This is the output you get from doing the above:
```
  File "<string>", line 1, in <module>
  File "/tmp/.venv/lib/python3.8/site-packages/clusterfuzz/stacktraces/__init__.py", line 21, in <module>
  File "/tmp/.venv/lib/python3.8/site-packages/clusterfuzz/_internal/base/utils.py", line 32, in <module>
    from base import memoize
  File "/tmp/.venv/lib/python3.8/site-packages/clusterfuzz/_internal/base/memoize.py", line 23, in <module>
    from base import persistent_cache
  File "/tmp/.venv/lib/python3.8/site-packages/clusterfuzz/_internal/base/persistent_cache.py", line 21, in <module>
    from metrics import logs
  File "/tmp/myproject/metrics.py", line 1, in <module>
    1/0
ZeroDivisionError: division by zero
```

oliverchang:
Move most imports into an `clusterfuzz._internal` dir, and
change internal imports to be of the form: `from clusterfuzz._internal.base.* import ` etc.

This avoids a lot of global import namespace pollution, and makes our
libClusterFuzz setup much less hacky.

Actual changes:
- Rename most of "src/python" to "src/clusterfuzz/_internal".
- Keep "src/python/bot/startup" to avoid issues with existing startup
  scripts.
- Move "protos" to "src/clusterfuzz/_internal" as well.
- Convert all imports to be relative from clusterfuzz._internal.
- Use isort for import ordering, remove our own implementation for
  checking of ordering during lint.
- Temporarily add some lint exclusion rules, as there were quite a few
  that were failing.
- Relax line length rules for test files.
- Add a license check exclusion for process_state.proto, since it has a
  difference license from what we expect.

Tests before deploying:
- [x] stage on GCE bot
- [x] deploy to App Engine staging
- [x] test pip package 

Fixes #2192.


oliverchang:
Sorry for the huge CL, most changes are mechanical.

Our previous setup with libClusterFuzz relied on lots of ugly symlink hacks that still lead to a lot of namespace pollution that makes it less easily usable (#2192). This makes ClusterFuzz structured as a proper package to avoid all of those hacks.

.pylintrc
```diff
@@ -34,7 +34,12 @@ disable=
   useless-suppression,
   wrong-import-order,
   wrong-import-position,
-
```
inferno-chromium, line 37: FIXME: in comment

inferno-chromium (approved):
Do we need to call that dir _internal, 
like from clusterfuzz.src.<whatever> reads cleaner than _internal ?

.pylintrc
oliverchang: Done!

oliverchang:
> Do we need to call that dir _internal,
> like from clusterfuzz.src. reads cleaner than _internal ?

Better to be explicit here, since a user installing this package will have access to the full set of imports.

oliverchang:
Deploying to internal first. If things look stable will continue to other deployments.

jonathanmetzman:
I think this caused a lot more errors than normal to hit my inbox.

PR #2385, opened 2021-07-23
Add missing OpenJDK dependency

fmeum:
The library `libXext.so` is required by some Java AWT classes even in headless environments. Without this library, fuzz targets that perform non-trivial image operations fail with an `UnsatisfiedLinkError` (see e.g. https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=33489).


inferno-chromium:
We don't install extra libs needed for specific project in clusterfuzz dockerfile (migration of images is pain), we just recommend adding those in project archive and patchelf-ing them

PR #2387, opened 2021-07-28
Use Windows ES info as Linux ES info

yuanjunh-git:
Extended stable doesn't ship to Linux but the downstream team still needs the impact calculation on Linux extended stable. So use Windows ES info as Linux ES info.


oliverchang (approved):
thanks a lot!

PR #2392, opened 2021-08-02
Add GitHub workflow for publishing libClusterFuzz.

oliverchang:
Triggered whenever we create a release.

### DIFF
--- a/.github/workflows/publish-to-pypi.yaml
+++ b/.github/workflows/publish-to-pypi.yaml
@@ -1,0 +1,34 @@
+name: Publish PyPI package
+on: push
+
+jobs:
+  publish:
+    name: Publish
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: src
+    steps:
+    - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-to-pypi.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - name: Install pypa/build
+      run: >-
+        python -m
+        pip install
+        build
+        --user
+    - name: Build a binary wheel and a source tarball
+      run: ./build.sh
+    - name: Publish distribution to PyPI
+      uses: pypa/gh-action-pypi-publish@master
```
inferno-chromium, line 44: Best to pin these by hashes, same above as well (instead of version or master).

inferno-chromium (approved):

.github/workflows/publish-to-pypi.yaml
oliverchang: done!

jonathanmetzman:
LGTM. Thanks!

PR #2397, opened 2021-08-04
Fix other-bots startup script paths.

oliverchang:


yuanjunh-git (approved):
thanks for this quick fix

Issue #2400, opened 2021-08-05
Allow disabling of strategies etc when fuzzing with libClusterFuzz.


oliverchang:
strategies, two step merge etc add a non-deterministic amount of time, and may not be wanted by all users.

oliverchang:
This is done.

PR #2401, opened 2021-08-06
Disable fuzzing strategies for libClusterFuzz.

oliverchang:
Current strategies can be problematic (e.g. running external mutators).

For: #2400

### DIFF
--- a/src/clusterfuzz/_internal/bot/fuzzers/afl/launcher.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/afl/launcher.py
@@ -291,7 +291,7 @@ class FuzzingStrategies(object):
 
     # If we have already generated a strategy dict, use that in favor of
     # creating a new pool and picking randomly.
-    if strategy_dict:
+    if strategy_dict is not None:
       self.use_corpus_subset = 'corpus_subset' in strategy_dict
       if self.use_corpus_subset:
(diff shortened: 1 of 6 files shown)


src/clusterfuzz/_internal/bot/fuzzers/afl/launcher.py
```diff
@@ -291,7 +291,7 @@ def __init__(self, target_path, strategy_dict=None):
 
     # If we have already generated a strategy dict, use that in favor of
     # creating a new pool and picking randomly.
-    if strategy_dict:
+    if strategy_dict is not None:
```
inferno-chromium, line 294: slightly confused here, since strategies = {} is not None, won't it go inside the loop?

inferno-chromium (approved):

src/clusterfuzz/_internal/bot/fuzzers/afl/launcher.py
oliverchang: That's the intention!

Previously, if we pass "{}", it will go into the "else", which is picking strategies from scratch. 

We don't want this -- we want "{}" to mean "no strategies are picked".

inferno-chromium: Gotcha LGTM!

PR #2405, opened 2021-08-10
Retrieve syzkaller rawcover every interval

kennethkang:
clusterfuzz currently relies on user to manually copy / paste / create a `coverfile` from `/rawcover` output. This PR automates this in a loop for a given interval of time (default 180 seconds)

Rev 1:
* Remove need to pass `--coverfile` arg
* Create cover file if does not exist
* Make and use `LoopingTimer` to periodically retrieve raw cover data from localhost:port
* Cleanup doc strings

### DIFF
--- a/src/clusterfuzz/_internal/bot/fuzzers/syzkaller/engine.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/syzkaller/engine.py
@@ -152,11 +152,6 @@ class SyzkallerEngine(engine.Engine):
 
     args = options.arguments
 
-    # TODO(yanghuiz): Dump coverfile from Syzkaller HTTP endpoint and
-    # remove this.
-    if not environment.is_android_cuttlefish():
-      args += ['--coverfile', runner.get_cover_file_path()]
-
     self.init_corpus(options.corpus_dir, runner.get_work_dir())
     fuzz_result = syzkaller_runner.fuzz(max_time, additional_args=args)
     self.save_corpus(runner.get_work_dir(), options.corpus_dir)
--- a/src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
@@ -43,8 +43,7 @@ def get_work_dir():
   work_dir = os.path.join(
       environment.get_value('FUZZ_INPUTS_DISK'), 'syzkaller')
 
-  if not os.path.exists(work_dir):
-    os.mkdir(work_dir)
+  os.makedirs(work_dir, exist_ok=True)
 
   return work_dir
 
@@ -83,8 +82,6 @@ def get_config():
 def get_cover_file_path():
   """Return location of coverage file for Syzkaller."""
   file_path = os.path.join(get_work_dir(), 'coverfile')
-  if not os.path.isfile(file_path):
-    open(file_path, 'w+').close()
 
   logs.log(f'Rawcover file will be stored in {file_path}')
   return file_path
--- a/src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
@@ -44,7 +44,6 @@ def get_work_dir():
       environment.get_value('FUZZ_INPUTS_DISK'), 'syzkaller')
 
   os.makedirs(work_dir, exist_ok=True)
-
   return work_dir
 
 
@@ -81,19 +80,12 @@ def get_config():
 
 def get_cover_file_path():
   """Return location of coverage file for Syzkaller."""
-  file_path = os.path.join(get_work_dir(), 'coverfile')
-
-  logs.log(f'Rawcover file will be stored in {file_path}')
-  return file_path
+  return os.path.join(get_work_dir(), 'coverfile')
 
 
 def find_syzkaller_port(pid: int) -> int or None:
   """Find localhost port where syzkaller is connected."""
-  try:
-    import psutil  # pylint: disable=g-import-not-at-top
-  except ImportError:
-    logs.log_warn('Runtime environment does not support psutil')
-    return None
+  import psutil  # pylint: disable=g-import-not-at-top
 
   for connection in psutil.net_connections():
     if connection.pid != pid:
@@ -211,8 +203,20 @@ class AndroidSyzkallerRunner(new_process.UnicodeProcessRunner):
       f.write(rawcover)
       logs.log(f'Writing syzkaller rawcover to {file_path}')
 
-  def run_and_wait(self, *args, timeout=None,
+  def run_and_loop(self, *args, timeout=None,
                    **kwargs) -> new_process.ProcessResult:
+    """Adds looping call to run_and_wait method.
+
+    This method adds LoopingTimer() that continuously executes a function
+    that gets / saves rawcover data from Syzkaller.
+
+    Args:
+      *args: args for self.run()
+      timeout: timeout in seconds to stop Syzkaller
+      **kwargs: kwargs for self.run()
+    Returns:
+      new_process.ProcessResult from Syzkaller
+    """
     process = self.run(*args, **kwargs)
     pid = process.popen.pid
     logs.log(f'Syzkaller pid = {pid}')
@@ -233,7 +237,7 @@ class AndroidSyzkallerRunner(new_process.UnicodeProcessRunner):
                                          output,
                                          time.time() - start_time, False)
 
-      result = new_process.wait_process(
+      return new_process.wait_process(
           process,
           timeout=timeout,
           input_data=None,
@@ -242,8 +246,6 @@ class AndroidSyzkallerRunner(new_process.UnicodeProcessRunner):
       )
     finally:
       looping_timer.cancel()
-
-    return result
 
   def fuzz(
       self,
@@ -277,7 +279,7 @@ class AndroidSyzkallerRunner(new_process.UnicodeProcessRunner):
     # Save kernel_bid for later in case the device is down.
     _, kernel_bid = kernel_utils.get_kernel_hash_and_build_id()
 
-    fuzz_result = self.run_and_wait(additional_args, timeout=fuzz_timeout)
+    fuzz_result = self.run_and_loop(additional_args, timeout=fuzz_timeout)
     logs.log('Syzkaller stopped, fuzzing timed out: {}'.format(
         fuzz_result.time_executed))
 
--- a/src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
@@ -20,7 +20,6 @@ import tempfile
 import threading
 import time
 
-import psutil
 import requests
 
 from clusterfuzz._internal.base import utils
@@ -91,26 +90,26 @@ def get_cover_file_path():
   return file_path
 
 
-def find_syzkaller_port(pid: int) -> int:
+def find_syzkaller_port(pid: int) -> int or None:
   """Find localhost port where syzkaller is connected."""
+  try:
+    import psutil  # pylint: disable=g-import-not-at-top
+  except ImportError:
+    logs.log_warn('Runtime environment does not support psutil')
+    return None
 
   for connection in psutil.net_connections():
     if connection.pid != pid:
       continue
 
-    try:
-      local_address = connection.laddr
-
-      if (local_address.ip == LOCAL_HOST and
-          connection.status == psutil.CONN_LISTEN):
-        logs.log(f'Syzkaller port = {local_address.port}')
-        return local_address.port
-
-    except AttributeError:
-      continue
+    local_address = connection.laddr
+    if (local_address.ip == LOCAL_HOST and
+        connection.status == psutil.CONN_LISTEN):
+      logs.log(f'Syzkaller port = {local_address.port}')
+      return local_address.port
 
   # No connection found
-  return 0
+  return None
 
 
 def get_runner(fuzzer_path):
@@ -200,16 +199,12 @@ class AndroidSyzkallerRunner(new_process.UnicodeProcessRunner):
 
   def save_rawcover_output(self, pid: int):
     """Find syzkaller port and write rawcover data to a file."""
-
     port = find_syzkaller_port(pid)
-
-    # port not found
-    if not port:
-      logs.log_warn('Syzkaller port was not found')
+    if port is None:
+      logs.log_warn('Could not find Syzkaller port')
       return
 
     rawcover = requests.get(f'http://localhost:{port}/rawcover').text
-
     if not rawcover:
       logs.log_warn('Syzkaller rawcover not yet loaded')
       return
--- a/src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
@@ -211,16 +211,13 @@ class AndroidSyzkallerRunner(new_process.UnicodeProcessRunner):
       f.write(rawcover)
       logs.log(f'Writing syzkaller rawcover to {file_path}')
 
-  def run_and_wait(self, syzkaller_args=None,
-                   timeout=None) -> new_process.ProcessResult:
-    process = self.run(syzkaller_args)
-
-    start_time = time.time()
-
+  def run_and_wait(self, *args, timeout=None,
+                   **kwargs) -> new_process.ProcessResult:
+    process = self.run(*args, **kwargs)
     pid = process.popen.pid
-
     logs.log(f'Syzkaller pid = {pid}')
 
+    start_time = time.time()
     looping_timer = LoopingTimer(
         RAWCOVER_RETRIEVE_INTERVAL,
         self.save_rawcover_output,
@@ -228,20 +225,23 @@ class AndroidSyzkallerRunner(new_process.UnicodeProcessRunner):
     )
     looping_timer.run()
 
-    if not timeout:
-      output = process.communicate()[0]
-      looping_timer.cancel()
-      return new_process.ProcessResult(process.command, process.poll(), output,
-                                       time.time() - start_time, False)
+    try:
+      if not timeout:
+        output = process.communicate()[0]
+        looping_timer.cancel()
+        return new_process.ProcessResult(process.command, process.poll(),
+                                         output,
+                                         time.time() - start_time, False)
 
-    result = new_process.wait_process(
-        process,
-        timeout=timeout,
-        input_data=None,
-        terminate_before_kill=False,
-        terminate_wait_time=None,
-    )
-    looping_timer.cancel()
+      result = new_process.wait_process(
+          process,
+          timeout=timeout,
+          input_data=None,
+          terminate_before_kill=False,
+          terminate_wait_time=None,
+      )
+    finally:
+      looping_timer.cancel()
 
     return result
 
--- a/src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
@@ -17,6 +17,11 @@ import fnmatch
 import os
 import re
 import tempfile
+import threading
+import time
+
+import psutil
+import requests
 
 from clusterfuzz._internal.base import utils
 from clusterfuzz._internal.bot.fuzzers import utils as fuzzer_utils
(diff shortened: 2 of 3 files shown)


src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
```diff
@@ -145,17 +170,41 @@ def repro(self, repro_timeout, repro_args):
...
                                   result.time_executed, result.output)
 
-  def fuzz(self,
-           fuzz_timeout,
-           additional_args,
-           unused_additional_args=None,
-           unused_extra_env=None):
+  def save_rawcover_output(self, filepath: str):
+    """Find syzkaller port and write rawcover data to a file."""
+    port = new_process.find_localhost_port(get_syzkaller_pid())
```
oliverchang, line 175: Is there a better way to get this? Like parsing some output, or passing it as an option to syzkaller? 

This is relying on a fragile assumption.

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
```diff
@@ -70,14 +73,37 @@ def get_config():
...
 
+def get_syzkaller_pid() -> int:
+  """Get pid of syzkaller.
+
+  Assumes syzkaller is the only child process of clusterfuzz.
+
+  Returns:
+    syzkaller pid. 0 if not found.
+  """
+  children_procs = psutil.Process(os.getpid()).children()
```
oliverchang, line 97: This is a fragile assumption and could easily break in the future. 

We can pipe the pid in a much better way, if we make a simple refactoring  to new_process:

- make `_wait_process` public (i.e. `new_process.wait_process`)
- Expose `pid` in `new_process.ChildProcess`
- Call `ProcessRunner.run` instead of `ProcessRunner.run_and_wait`, and grab the `pid`, start the thread to poll for coverage and then call `new_process.wait_process` on it similar to here: https://github.com/google/clusterfuzz/blob/5b8f9b9efb25e4291024eeacfc254162a34dc320/src/clusterfuzz/_internal/system/new_process.py#L347

oliverchang (changes requested):
nice! some comments on the implementation.

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/engine.py
```diff
@@ -137,7 +137,7 @@ def fuzz(self, target_path, options, unused_reproducers_dir=None, max_time=0):
     Args:
       target_path: Path to the target.
       options: The FuzzOptions object returned by prepare().
-      reproducers_dir: The directory to put reproducers in when crashes
+      unused_reproducers_dir: The directory to put reproducers in when crashes
```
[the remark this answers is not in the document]
kennethkang: argument list doesn't include `reproducers_dir`?

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
```diff
@@ -70,14 +73,37 @@ def get_config():
 
 def get_cover_file_path():
   """Return location of coverage file for Syzkaller."""
-  return os.path.join(get_work_dir(), 'coverfile')
+  file_path = os.path.join(get_work_dir(), 'coverfile')
+  if not os.path.isfile(file_path):
+    open(file_path, 'w+').close()
+
+  logs.info(f'Rawcover file will be stored in {file_path}')
```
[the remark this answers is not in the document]
kennethkang: will change

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
kennethkang: the reason I was not using this is because `process.communicate()` is a blocking call from my understanding. pid will not return until `syzkaller` is finished. We need a process to get `rawcover` while `syzkaller` process is running.

src/clusterfuzz/_internal/system/new_process.py
```diff
@@ -128,6 +182,23 @@ def kill_process_tree(root_pid):
     logs.log_warn('Failed to kill process.')
 
 
+class LoopingTimer(threading.Timer):
```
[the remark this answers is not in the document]
kennethkang: will do

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
kennethkang: best way to get `rawcover` data from already running `syzkaller` process, we were advised by Dmitry, is to get it from `localhost:port/rawcover`

src/clusterfuzz/_internal/system/new_process.py
```diff
@@ -29,6 +29,60 @@
...
+  Args:
+    port (int): port
+    host (str): host
+    timeout (float): timeout for checking port
+    buff_size (int): max buffer size to retrieve
+  Returns:
+    bytes data from socket. None when times out
+  """
+  try:
+    s = socket.create_connection((host, port), timeout=timeout)
```
kennethkang, line 78: I forgot to add `/rawcover`. need to fix

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
oliverchang: I mean getting the port number in a more reliable way, or pre-specifying it somehow rather than trying to figure it out this way at runtime.

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
oliverchang: Yes, please see my refactoring suggestion to resolve this.

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/engine.py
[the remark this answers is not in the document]
oliverchang: That's because it was incorrectly named "unused_reproducers_dir". Please rename it back to "reproducers_dir" (which matches the name in the base class).

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
kennethkang: yeah this is something I considered, but didn't want to make too big of a refactor. I will take your suggestion and try to minimize the refactor size.

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/engine.py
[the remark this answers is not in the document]
kennethkang: ah base class. I see. will revert back.

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
kennethkang: I'm not sure if there is a good way to get a port number. I can reliably get pid from your suggestion and port number should be reliably retrievable from `psutil.Popen(pid).connections()`.

There are a couple of reasons why I am getting port this way opposed to assigning syzkaller a port via config (there is a way to do this):
1. multiple instances of clusterfuzz will be running on one ATP instance. cannot set a constant port
2. pre-assignment of port can lead to race condition: one clusterfuzz spins up a syzkaller with a pre-assigned port. before syzkaller bot is instantiated, another instance of clusterfuzz spins up another syzkaller taking the port.
3. `psutil.Popen(pid)` will reliably give us ports that are only associated with the instance of syzkaller.

kennethkang:
sorry about the force push. in retrospect, I shouldn't have `git commit --amend`ed all of my changes.
Reflected your suggestion and refactored. Tested locally.

kennethkang:
Rev 2:
* Refactor to use existing `wait_process()`
* Get `pid` from `process.run()`
* Move code into `syzkaller` directory
* Create directory if `coverfile` directory does not exist
* Use `/rawcover` endpoint

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
kennethkang: `psutil.Process(pid).connections()` was emitting `AccessDeniedError`. Changed to `psutil.net_connections()` and finding a connection with a matching `pid` and localhost ip with `LISTEN`

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
```diff
@@ -145,17 +198,77 @@ def repro(self, repro_timeout, repro_args):
...
                                   result.time_executed, result.output)
 
-  def fuzz(self,
-           fuzz_timeout,
-           additional_args,
-           unused_additional_args=None,
-           unused_extra_env=None):
+  def save_rawcover_output(self, pid: int):
+    """Find syzkaller port and write rawcover data to a file."""
+
```
oliverchang, line 203: nit: unnecessary newline.

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
```diff
@@ -145,17 +198,77 @@ def repro(self, repro_timeout, repro_args):
...
+    port = find_syzkaller_port(pid)
+
```
oliverchang, line 205: can delete lines 205-206 here (no need for the comment -- the code is obvious enough).

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
```diff
@@ -145,17 +198,77 @@ def repro(self, repro_timeout, repro_args):
...
+    if not rawcover:
+      logs.log_warn('Syzkaller rawcover not yet loaded')
+      return
+
+    file_path = get_cover_file_path()
+    with open(file_path, 'w+') as f:
+      f.write(rawcover)
+      logs.log(f'Writing syzkaller rawcover to {file_path}')
+
+  def run_and_wait(self, syzkaller_args=None,
```
oliverchang, line 222: can you make the signature:

```python
def run_and_wait(self, *args, timeout=None, **kwargs)
```

And then pass it to `self.run(*args, **kwargs)`

You also need to add a docstring here, something along the lines of "Overridden run_and_wait to ... describe why ...."

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
```diff
@@ -70,7 +83,34 @@ def get_config():
...
+  file_path = os.path.join(get_work_dir(), 'coverfile')
+  if not os.path.isfile(file_path):
+    open(file_path, 'w+').close()
+
+  logs.log(f'Rawcover file will be stored in {file_path}')
+  return file_path
+
+
+def find_syzkaller_port(pid: int) -> int:
+  """Find localhost port where syzkaller is connected."""
```
oliverchang, line 95: Can you import psutil here instead? This code may get imported in environments where psutil isn't available.

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
```diff
@@ -70,7 +83,34 @@ def get_config():
...
+      if (local_address.ip == LOCAL_HOST and
+          connection.status == psutil.CONN_LISTEN):
+        logs.log(f'Syzkaller port = {local_address.port}')
+        return local_address.port
+
+    except AttributeError:
+      continue
+
+  # No connection found
+  return 0
```
oliverchang, line 113: return None instead.

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
```diff
@@ -70,7 +83,34 @@ def get_config():
...
+    try:
+      local_address = connection.laddr
+
+      if (local_address.ip == LOCAL_HOST and
+          connection.status == psutil.CONN_LISTEN):
+        logs.log(f'Syzkaller port = {local_address.port}')
+        return local_address.port
+
+    except AttributeError:
+      continue
```
oliverchang, line 110: why do we continue in this case? This could hide real errors.

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
```diff
@@ -70,7 +83,34 @@ def get_config():
 
 def get_cover_file_path():
   """Return location of coverage file for Syzkaller."""
-  return os.path.join(get_work_dir(), 'coverfile')
+  file_path = os.path.join(get_work_dir(), 'coverfile')
+  if not os.path.isfile(file_path):
+    open(file_path, 'w+').close()
```
oliverchang, line 88: why do we need this, when we write to it later anyway?

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
```diff
@@ -29,12 +34,20 @@
...
 
 
 def get_work_dir():
   """Return work directory for Syzkaller."""
-  return os.path.join(environment.get_value('FUZZ_INPUTS_DISK'), 'syzkaller')
+  work_dir = os.path.join(
+      environment.get_value('FUZZ_INPUTS_DISK'), 'syzkaller')
+
+  if not os.path.exists(work_dir):
+    os.mkdir(work_dir)
```
oliverchang, line 48: nit: just `os.makedirs(work_dir, exist_ok=True)`

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
```diff
@@ -145,17 +198,77 @@ def repro(self, repro_timeout, repro_args):
...
+    looping_timer = LoopingTimer(
+        RAWCOVER_RETRIEVE_INTERVAL,
+        self.save_rawcover_output,
+        args=[pid],
+    )
+    looping_timer.run()
+
+    if not timeout:
+      output = process.communicate()[0]
+      looping_timer.cancel()
```
oliverchang, line 241: can you wrap lines 238-252 in a try/finally, and have `looping_timer.cancel()` as part of the `finally` block? 

Otherwise if we exception somewhere here and it's handled somewhere down the call stack, the timer thread will leak.

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
```diff
@@ -145,17 +198,77 @@ def repro(self, repro_timeout, repro_args):
...
+
+    port = find_syzkaller_port(pid)
+
+    # port not found
+    if not port:
+      logs.log_warn('Syzkaller port was not found')
+      return
+
+    rawcover = requests.get(f'http://localhost:{port}/rawcover').text
+
```
oliverchang, line 212: nit: unnecssary newline.

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
```diff
@@ -145,17 +198,77 @@ def repro(self, repro_timeout, repro_args):
...
+      logs.log(f'Writing syzkaller rawcover to {file_path}')
+
+  def run_and_wait(self, syzkaller_args=None,
+                   timeout=None) -> new_process.ProcessResult:
+    process = self.run(syzkaller_args)
+
+    start_time = time.time()
+
+    pid = process.popen.pid
+
```
oliverchang, line 229: can delete this blank line.

oliverchang (commented):
nice! looks good at a glance. just some more comments.

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
oliverchang: Ok, that sounds reasonable. Does syzkaller output anything in its logs to indicate what port it's using.

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
kennethkang: 👍

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
kennethkang: hm I think this was because of the way I implemented it earlier. will fix.

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
kennethkang: will do

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
kennethkang: will change

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
kennethkang: didn't feel right about this `cancel` placement. I like this suggestion.

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
```diff
@@ -70,7 +81,32 @@ def get_config():
...
-  return os.path.join(get_work_dir(), 'coverfile')
+  file_path = os.path.join(get_work_dir(), 'coverfile')
+
+  logs.log(f'Rawcover file will be stored in {file_path}')
+  return file_path
+
+
+def find_syzkaller_port(pid: int) -> int or None:
+  """Find localhost port where syzkaller is connected."""
+  try:
```
oliverchang, line 92: no need for the try/catch in this case, because this will be called in a context where psutil must exist.

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
```diff
@@ -29,12 +33,19 @@
...
 
 
 def get_work_dir():
   """Return work directory for Syzkaller."""
-  return os.path.join(environment.get_value('FUZZ_INPUTS_DISK'), 'syzkaller')
+  work_dir = os.path.join(
+      environment.get_value('FUZZ_INPUTS_DISK'), 'syzkaller')
+
+  os.makedirs(work_dir, exist_ok=True)
+
```
oliverchang, line 47: nit: unnecessary blank line

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
```diff
@@ -145,17 +194,73 @@ def repro(self, repro_timeout, repro_args):
...
+    try:
+      if not timeout:
+        output = process.communicate()[0]
+        looping_timer.cancel()
+        return new_process.ProcessResult(process.command, process.poll(),
+                                         output,
+                                         time.time() - start_time, False)
+
+      result = new_process.wait_process(
```
oliverchang, line 236: you can just return directly here. the "finally" block will still execute.

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
```diff
@@ -70,7 +81,32 @@ def get_config():
 
 def get_cover_file_path():
   """Return location of coverage file for Syzkaller."""
-  return os.path.join(get_work_dir(), 'coverfile')
+  file_path = os.path.join(get_work_dir(), 'coverfile')
+
+  logs.log(f'Rawcover file will be stored in {file_path}')
```
oliverchang, line 86: nit: a better place for this log call would be the caller (i.e. line 210).  Otherwise if in the future this gets called multiple times we'll get redundant logs.

oliverchang (approved):
Thanks! This mostly LGTM with some minor things remaining.

Please fix the lint issues:
https://github.com/google/clusterfuzz/pull/2405/checks?check_run_id=3318973221

You can see lint issues by running:

```
python butler.py lint
```

and also format your code using

```
python butler.py format
```

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
kennethkang: unfortunately `syzkaller` does not log out the port it is using. It just prints what is in `config` which is unfortunately `localhost:0`. Port 0 means auto-select an available port.

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
kennethkang: fixed

kennethkang:
Rev 3:
* Import `psutil` within function
* Remove redundant directory checks
* Cancel looping timer in finally

Rev 4:
* Do not try catch import `psutil`
* In try / finally block, return in try
* Remove redundant logging
* Rename method to `run_and_loop`

kennethkang:
To avoid overwriting with different number of arguments caught by unit test, I changed the method name to `run_and_loop` so as not to overwrite.

kennethkang:
During local testing with appropriate `config.py`,
* `Syzkaller`'s `pid` and `port` were successfully retrieved.
* `localhost:{port}/rawcover` correctly displays `Syzkaller` `rawcover` address list.
* rawcover address list is verified to be saved in a cover file.

PR #2406, opened 2021-08-10
Reland "Support interactive output for fuzzing"

oliverchang:


src/clusterfuzz/__init__.py
```diff
@@ -12,3 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """libClusterfuzz package."""
+
+import os
+
+if not os.getenv('ROOT_DIR') and not os.getenv('GAE_ENV'):
```
oliverchang, line 18: This is the fix (GAE_ENV check)

inferno-chromium (approved):

PR #2410, opened 2021-08-12
Allow fuzzing builds from multiple branches

MHA3:
Pick fuzz targets from the same branch as the OS build for Android devices.


oliverchang (commented):
thanks for this PR! Some questions + suggestions, but I'm not sure that this is the best way to solve this use case. I'll think more about this as well and see if we can do this in a better away. 

This alone also doesn't prevent us trying to reproduce testcases on a different family (seahawk 11 vs 12). What may still happen is that we find a crash on seahawk 11, and then it ends up getting reproduced on a seahawk 12 and incorrectly gets marked as fixed. The way to prevent this is separating jobs by platform.

Huiziyang:
Hi Oliver,
Correct me if I'm wrong, today we are still distinguishing reproduction job based on platform ID right? If yes, platform ID should include branch name or we can add to it so it won't reproduce on different family

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -1232,6 +1232,11 @@ def get_primary_bucket_path():
 
   fuzz_target_build_bucket_path = environment.get_value(
       'FUZZ_TARGET_BUILD_BUCKET_PATH')
+  if environment.is_android():
+    fuzz_target_build_bucket_path = \
+      android.settings.check_and_apply_bucket_overrides(
```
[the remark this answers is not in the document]
MHA3: That does sound like a better idea. Let me work on it...

Let me discuss the testcases part with Huizi

src/clusterfuzz/_internal/platforms/android/constants.py
```diff
@@ -48,3 +48,10 @@
     'muskie': 'wahoo',
     'taimen': 'wahoo',
 }
+
+AUTO_FINGERPRINT_TO_BUILD_BRANCH = {
```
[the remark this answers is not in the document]
MHA3: I discussed it with the auto team. There isn't a public source rn, but two ideas we discussed were:
- to create a config on our gcs bucket root to pick up at runtime
- to try and get these configs from tradefed to clusterfuzz
what do you think of these approaches?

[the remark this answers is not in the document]
oliverchang: As discussed having a JSON file in the GCS bucket root would be much nicer. We can also have that store the full GCS path to replace this with. That way, we don't need to perform potentially fragile URL parsing/replacement. 

We can also make this much more general (and not specific to Android). In `get_bucket_path`, if the bucket path has "platform_id_urls.json" as the last component (i.e. `gs://bucket/blah/platform_id_urls.json`) , we'll download the JSON and map it from `environment.platform_id()`. to the actual URL.

oliverchang (commented):
> Hi Oliver,
Correct me if I'm wrong, today we are still distinguishing reproduction job based on platform ID right? If yes, platform ID should include branch name or we can add to it so it won't reproduce on different family

Indeed, today this will work. I had misunderstood the intention of this PR.

In the future though, if we build a better replacement/implementation of handling platform IDs, this PR may no longer be necessary. This is fine for now in the meantime.

src/clusterfuzz/_internal/base/data_import.py
```diff
@@ -0,0 +1,56 @@
...
+"""Data retrieving helper functions."""
+
+import ast
+import base64
+
+from clusterfuzz._internal.base import utils
+from clusterfuzz._internal.google_cloud_utils import storage
+
+
+def get_url_content(url):
```
oliverchang, line 23: Please move these functions back into revisions, as it's very specific to that.

Other callers can call `storage.read_data` directly.

src/clusterfuzz/_internal/tests/core/platforms/android/settings_data/test_config.json
```diff
@@ -0,0 +1,10 @@
+{
+	"auto_build_version_to_fuzz_path": {
+		"r": "gs://auto/git_r/something/something/%something%/([something]+).zip",
```
oliverchang, line 3: Can this match environment.get_platform_id() completely `android:{codename}_{sanitizer}:{build_version}` instead.

src/clusterfuzz/_internal/tests/core/platforms/android/settings_data/test_config.json
```diff
@@ -0,0 +1,10 @@
+{
+	"auto_build_version_to_fuzz_path": {
```
oliverchang, line 2: Let's make this more generic. Like "build_paths".

src/clusterfuzz/_internal/tests/core/platforms/android/settings_data/test_config.json
```diff
@@ -0,0 +1,10 @@
...
+		"r": "gs://auto/git_r/something/something/%something%/([something]+).zip",
+		"s": "gs://auto/git_s/something/%something%/([something]+).zip"
+	},
+	"auto_build_version_to_revisions_path": {
```
oliverchang, line 6: Let's make this more generic, like "revisions_paths".

That said, do these actually need to be different? Can we keep things simpler and have a single srcmap that represents all revision num -> git hash mappings?

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -1671,3 +1673,11 @@ def check_app_path(app_path='APP_PATH'):
   # APP_PATH.
   return (not environment.get_value('APP_NAME') or
           environment.get_value(app_path))
+
+
+def get_bucket_path(name):
+  """Return build bucket path, applying any set overrides."""
+  bucket_path = environment.get_value(name)
+  if environment.is_android():
```
oliverchang, line 1681: Let's do this in a more generic way (not specific to android), and do something like (as described in https://github.com/google/clusterfuzz/pull/2410/files#discussion_r690874921):

```python
if os.path.basename(bucket_path) == PLATFORM_ID_URLS_FILENAME:
   _apply_platform_id_overrides(environment.get_platform_id(), bucket_path):

def _apply_platform_id_overrides(...):
  # read the `bucket_path`, parse as JSON, and map based on platform_id. 
```

oliverchang (commented):
thanks! some suggestions.

Another higher level question: how often do the branches change? 

This code does add a fair bit of complexity -- and the benefit here seems to only be the ability to avoid manually editing jobs. If this is very infrequent, perhaps it's better to have separate jobs instead (manually maintained for now). That also fits better into something that would segregate platform IDs a bit better, where we could e.g. auto-create/update jobs as necessary.

src/clusterfuzz/_internal/base/data_import.py
MHA3: as i read through the function, it seemed to work really for us to try and import the `.json` config we added. It seems to designed for the `.json` format more generally. Therefore, to avoid repeating a lot of this code in `android.settings`, I thought it'd be better to make it available outside of `revisions.py`. But I think I see your point, we might want to change the way configs are written, in which case, we'd have to stop relying on this function anyway. In that case, it'd just be left hanging outside. Do I understand it correctly?

src/clusterfuzz/_internal/tests/core/platforms/android/settings_data/test_config.json
MHA3: Yes, that is probably better, and easy to generalize for the auto, phone-kernelspace and phone-userspace as well. Will get to work...

src/clusterfuzz/_internal/tests/core/platforms/android/settings_data/test_config.json
MHA3: agreed. On it.

src/clusterfuzz/_internal/tests/core/platforms/android/settings_data/test_config.json
MHA3: I am not sure if the two branches can have overlapping revision numbers. If not, then i think this could work. If they can, i think it's better to have them in their respective directories.

src/clusterfuzz/_internal/build_management/build_manager.py
MHA3: agreed. This seems like a better idea.

MHA3: What do you think of this strategy:
set the bucket path to the config.json file url.
then, based on your suggestion above, at run time, if os.path.basename is the config file, try to get the actual path based on the device platform id.

src/clusterfuzz/_internal/build_management/overrides.py
```diff
@@ -0,0 +1,76 @@
...
+  return new_path
+
+
+def _get_config_dict(url):
+  """Read configs from a json and return them as a dict"""
+  config_dict = {}
+  url_data = storage.read_data(url)
+  if not url_data:
+    return config_dict
+  if not isinstance(url_data, str):
```
oliverchang, line 54: This should always return bytes. No need for this condition. 

No need for the decode() either, just pass to `json.loads()`, which accepts bytes.

src/clusterfuzz/_internal/build_management/overrides.py
```diff
@@ -0,0 +1,76 @@
...
+    url_data = url_data.decode('utf-8')
+
+  if url_data:
+    try:
+      result = ast.literal_eval(url_data)
```
oliverchang, line 59: Don't use literal_eval here. Just use `return json.loads(url_data)`. Don't catch any exceptions.

src/clusterfuzz/_internal/build_management/overrides.py
```diff
@@ -0,0 +1,76 @@
...
+  if found:
+    new_path = path
+  return new_path
+
+
+def _get_config_dict(url):
+  """Read configs from a json and return them as a dict"""
+  config_dict = {}
+  url_data = storage.read_data(url)
+  if not url_data:
```
oliverchang, line 52: If we get to this point we should just exception out if we fail to read any data.

src/clusterfuzz/_internal/build_management/overrides.py
```diff
@@ -0,0 +1,76 @@
...
+    return config_dict
+  if not isinstance(url_data, str):
+    url_data = url_data.decode('utf-8')
+
+  if url_data:
```
oliverchang, line 57: no need for this check (see comment above)

src/clusterfuzz/_internal/build_management/overrides.py
```diff
@@ -0,0 +1,76 @@
...
+"""Helper functions to apply build and revisions path overrides"""
+
+import ast
+import os
+
+from clusterfuzz._internal.google_cloud_utils import storage
+from clusterfuzz._internal.system import environment
+
+PLATFORM_ID_URLS_FILENAME = 'config.json'
+PLATFORM_TO_BUILD_PATH_KEY = 'build_paths'
```
oliverchang, line 23: nit: s/PLATFORM/PLATFORM_ID/

src/clusterfuzz/_internal/build_management/overrides.py
```diff
@@ -0,0 +1,76 @@
...
+        config_dict = result
+
+    except (ValueError, TypeError):
+      pass
+  return config_dict
+
+
+def _get_path_from_config(config_dict, config_key, platform_id):
+  """Return True if a path override is present and return the override."""
+  config_sub_dict = config_dict.get(config_key)
```
oliverchang, line 70: Just return a single value. We can use `None` to mean no path as found. This can be much simpler:

```python
if config_key not in config_dict:
  return None

return config_dict[config_key].get(platform_id)
```

src/clusterfuzz/_internal/build_management/overrides.py
```diff
@@ -0,0 +1,76 @@
...
+    curr_path = _apply_platform_id_overrides(environment.get_platform_id(),
+                                             curr_path, config_key)
+  return curr_path
+
+
+def _apply_platform_id_overrides(platform_id, config_url, config_key):
+  """read the `bucket_path`, parse as JSON, and map based on platform_id."""
+  new_path = config_url
+  config_dict = _get_config_dict(config_url)
+  found, path = _get_path_from_config(config_dict, config_key, platform_id)
```
oliverchang, line 42: If we get to this point and there's no path found, that seems like a fatal error. Just exception out here. 

Define our own exception type:

```
class BuildOverrideError(Exception):
  """Build override error."""


raise BuildOverrideError('Could not find correct build override path.')
```

oliverchang (commented):
thanks! Just some more comments for cleanup. Otherwise at the high level this LG.

src/clusterfuzz/_internal/build_management/build_manager.py
oliverchang: Yes, this is exactly what I'm proposing, thanks!

src/clusterfuzz/_internal/build_management/overrides.py
MHA3: done.

oliverchang:
(Please also rebase your PR to fix the lint issues).

PR #2412, opened 2021-08-13
Update image

jonathanmetzman:


jonathanmetzman:
TODO:
1. Add back python3.7 installation from source. No point in making our life harder with a python upgrade as well.
2. Pin base for Chromium images so we can do this upgrade incrementally.

docker/ml-with-gpu/Dockerfile
```diff
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-FROM gcr.io/clusterfuzz-images/base
+FROM gcr.io/clusterfuzz-images/base@sha256:1110514de5bb678562b75e5cf65130923bd22bb9371dcbc9e99483fd277ac135
```
jonathanmetzman, line 14: Pinning this image for now. Upgrading this will be a very large effort (and for unknown benefit :\

jonathanmetzman:
Related: https://github.com/google/oss-fuzz/issues/6180

oliverchang (approved):
Thanks a lot! I'll test this a bit locally first before merging and deploying on a small number of real bots.

docker/base/Dockerfile
```diff
@@ -40,10 +40,9 @@ RUN apt-get update && \
         lsb-release \
         net-tools \
         psmisc \
-        python \
-        python-dbg \
-        python-dev \
-        python-pip \
+        python3-dbg \
```
oliverchang, line 43: Do we still need all of these when we compile our own python3 ?

The previous packages where for python2, which was unsupported/unused anyway.

jonathanmetzman: Probably not but I didn't want to take the risk of removing it.
How about we leave a TODO to get rid of these or the building python from source (it probably makes more sense to install python from source for us right?)

oliverchang: Having multiple python3 versions probably adds more possibility for subtle issues than not. Let's remove here and I can see if this breaks things.

jonathanmetzman: Done.

PR #2415, opened 2021-08-17
Pin Fuchsia and Chromium images to specific base images.

jonathanmetzman:
This is so that we don't have to these images when
we upgrade the OSS-Fuzz images.

This is related to https://github.com/google/oss-fuzz/issues/6180


inferno-chromium (approved):

oliverchang (approved):
LGTM

PR #2418, opened 2021-08-19
Add googlefuzztest to list of engines for get_engine_for_job.

oliverchang:
This is difficult to generalize today, because this code runs on App
Engine as well, where no engine code is deployed.


oliverchang:
TBR. This is rather ugly, but will require a larger refactoring to figure out a cleaner way to propagate. /gcbrun

inferno-chromium (approved):

jonathanmetzman:
LGTM

PR #2421, opened 2021-08-19
Switch to refs/heads/main for chromium

inferno-chromium:


inferno-chromium:
fyi old revisions wont work after this, we might have to find some other way or repeat the calls.

inferno-chromium:
TBR

oliverchang (commented):
Thanks! Adding another call for "master" SGTM. Can add that.

PR #2429, opened 2021-08-24
Set PYTHONUNBUFFERED in run.py

oliverchang:
Otherwise exception logs from run_bot.py can be lost.


inferno-chromium (approved):

jonathanmetzman (approved):
LGTM

PR #2440, opened 2021-09-02
Various libClusterFuzz improvements.

oliverchang:
- Add a context manager for setting necessary env vars temporarily.
- Set exit codes for fuzz and reproduce.
- Allow additional args to be passed to fuzzing.
- Create prune output directory if needed.

For https://github.com/google/oss-fuzz/pull/6375

### DIFF
--- a/src/clusterfuzz/environment/__init__.py
+++ b/src/clusterfuzz/environment/__init__.py
@@ -1,0 +1,68 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Environment helpers."""
+
+import contextlib
+import os
+
+from clusterfuzz._internal.system import environment
+
+# Inverted SANITIZER_NAME_MAP.
+SANITIZER_MAP = {v: k for k, v in environment.SANITIZER_NAME_MAP.items()}
+
+
+class Environment:
+  """Environment helpers."""
+
+  def __init__(self, engine, sanitizer, target_path, interactive=False):
+    self._previous_values = {}
+    self.engine = engine
+    self.sanitizer = sanitizer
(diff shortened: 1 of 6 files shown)


src/clusterfuzz/environment/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+# limitations under the License.
+"""Environment helpers."""
+
+import contextlib
+import os
+
+from clusterfuzz._internal.system import environment
+
+# Inverted SANITIZER_NAME_MAP.
+SANITIZER_MAP = {v: k for k, v in environment.SANITIZER_NAME_MAP.items()}
```
jonathanmetzman, line 22: Should we assert that each value in SANITIZER_NAME_MAP is unique?

src/clusterfuzz/fuzz/__main__.py
```diff
@@ -47,28 +46,28 @@ def main():
...
-    options.merge_back_new_testcases = False
-    options.analyze_dictionary = False
+    result = engine_impl.fuzz(args.target, options, args.output,
+                              args.max_duration)
 
-  result = engine_impl.fuzz(args.target, options, args.output,
-                            args.max_duration)
   print('Command: ', ' '.join([shlex.quote(part) for part in result.command]))
+  if result.crashes:
+    sys.exit(1)
```
jonathanmetzman, line 68: I'd prefer we returned these values and that `sys.exit(main())` was called on line 74

src/clusterfuzz/environment/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+    if self.interactive:
+      self.set_value('CF_INTERACTIVE', 'True')
+
+    return self
+
+  def __exit__(self, *exc):
+    """Exit environment."""
+    del exc
+
+    for key, value in self._previous_values.items():
```
jonathanmetzman, line 56: We're sure that none of these will get deleted?
Maybe check for key presence instead just to be safe?

jonathanmetzman (approved):
LGTM

src/clusterfuzz/environment/__init__.py
oliverchang: Good point, added!

src/clusterfuzz/fuzz/__main__.py
oliverchang: That's a much bigger change, and not possible today without a lot of refactoring.

src/clusterfuzz/environment/__init__.py
oliverchang: Could you clarify what you mean here? 

We only delete an env var if we previously got "None" when we saved it.

PR #2441, opened 2021-09-02
libFuzzer: Don't subtract buffer times from the provided timeout.

oliverchang:
Previously we were subtracting some buffer timeouts from the provided timeout and
passing that to libFuzzer's `-max_total_time`. This was a relic from our launcher script
days when timeouts were much more strict. 

Now, instead leave `-max_total_time` to be exactly what is passed.
Add the buffer timeouts in the actual process timeout instead.
A difference of up to extra 20 seconds should be inconsequential.

This makes libClusterFuzz much more usable. 

Also for https://github.com/google/oss-fuzz/pull/6375

### DIFF
--- a/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py
@@ -316,12 +316,11 @@ class LibFuzzerEngine(engine.Engine):
     timeout_limit = fuzzer_utils.extract_argument(
         options.arguments, constants.TIMEOUT_FLAG, remove=False)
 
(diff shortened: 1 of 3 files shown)


jonathanmetzman (approved):
LGTM

jonathanmetzman:
Not such a fan of each fuzzer possibly taking 20 seconds longer than expected, it seems excessive to wait for that long IMO. But probably fine.

oliverchang:
> Not such a fan of each fuzzer possibly taking 20 seconds longer than expected, it seems excessive to wait for that long IMO. But probably fine.

That should be very rare in practice. In most cases I expect it will be way less than 10 seconds.

PR #2442, opened 2021-09-08
Make KASAN crash format more flexible

kennethkang:
Rev1:
* Add a new use-after-free crash case
* Amend KASAN crash regex
* Add unit test


kennethkang:
Need to fix `_filter_log`

inferno-chromium:
I don't think this is right, we should already be stripping the prefixes. That is how all KASan crashes worked so far, otherwise nothing would be working.

kennethkang:
@inferno-chromium agreed. My comment about `_filter_log` was to address this issue via stripping prefixes. Will push another revision.

inferno-chromium:
> @inferno-chromium agreed. My comment about `_filter_log` was to address this issue via stripping prefixes. Will push another revision.

yes so some stripping should be already be there, there can be a bug in stripping, but some should be existing as otherwise we wont be parsing any kasan stack.

kennethkang:
@oliverchang hmm where to place a unit test is quite confusing for for me for this package. May I suggest down the road to place unit test with the same directory structure?

e.g
`src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runners.py` vs. `tst/clusterfuzz/_internal/bot/fuzzers/syzkaller/runners_test.py`

kennethkang:
rev 2:
* filter log time stamp instead of regex match
* revert regex change
* update unit test case

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
```diff
@@ -279,13 +279,25 @@ def fuzz(
...
+      [  565.723853] c4   8262 BUG: KASAN: ...
+      vs.
+      BUG: KASAN: ...
+
+      Args:
+        content (str): log content
+      Returns:
+        filtered log with new lines (str)
+      """
+
```
oliverchang, line 295: nit: remove the blank lines on 295 and 297.

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
```diff
@@ -279,13 +279,25 @@ def fuzz(
...
+      strip_regex = re.compile(r'^(\[.*\]\s+)?c\d+\s+\d+\s')
```
oliverchang, line 296: nit: it might be better to use `.*?` instead (non greedy) so that the matching stops at the first ']'.

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
```diff
@@ -279,13 +279,25 @@ def fuzz(
...
-      result = ''
-      strip_regex = re.compile(r'^c\d+\s+\d+\s')
-      for line in content.splitlines():
-        result += strip_regex.sub('', line) + '\n'
-      return result
+    def _filter_log(content: str) -> str:
+      """Remove unnecessary prefix from each line of log.
+
+      e.g
+      [  565.723853] c4   8262 BUG: KASAN: ...
```
oliverchang, line 286: Just curious: what does "c4  8262" mean?

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
```diff
@@ -279,13 +279,25 @@ def fuzz(
...
     """
 
-    def _filter_log(content):
-      """Filter unneeded content from log."""
-      result = ''
-      strip_regex = re.compile(r'^c\d+\s+\d+\s')
-      for line in content.splitlines():
-        result += strip_regex.sub('', line) + '\n'
-      return result
+    def _filter_log(content: str) -> str:
```
oliverchang, line 282: Could we add a test for this filtering functionality?

oliverchang (commented):
> hmm where to place a unit test is quite confusing for for me for this package. May I suggest down the road to place unit test with the same directory structure?

Yeah, the current structure is due to legacy reasons. We'd need some spare cycles to change this. Happy to accept contributions here :) 

For now `src/clusterfuzz/_internal/tests/core/bot/fuzzers/syzkaller/runners_test.py` would be the right place for now.

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
kennethkang: yes we can definitely put a unit test here. I'll place it in the directory you've mentioned :)
`src/clusterfuzz/_internal/tests/core/bot/fuzzers/syzkaller/runners_test.py`

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
kennethkang: I see. I thought regex would match only once? so `\]` would always match the first `]`? Is this accounting for a case with multiple closing brackets?

oliverchang: Yes, .* is greedy by default and will as much as it can.

```
>>> re.match(r'(\[.*\])', '[foo] [bar]').group(1)
'[foo] [bar]'

>>> re.match(r'(\[.*?\])', '[foo] [bar]').group(1)
'[foo]'
```

kennethkang: thanks for the feedback. this is fixed.

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
kennethkang: this is fixed

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
kennethkang: unit test added

kennethkang:
Rev 3:
* Unit test `_filter_log`
* disable `protected-access` pylint for unit test
* fix regex error
* encode log output to `utf-8`

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
kennethkang: we are not entirely sure what it is. possibly pid. We do not need this prefix either way.

inferno-chromium (approved):

PR #2444, opened 2021-09-10
Truncate stacktrace in testcase details if it's too big.

oliverchang:
Limit the length to 500000 bytes. Take the first 250000 and the last
250000 of the stacktrace and show that instead.


inferno-chromium (approved):
Thanks a lot!

jonathanmetzman:
lgtm

PR #2445, opened 2021-09-12
File kasan crash with address range

kennethkang:
Rev 1:
* make `c\d+\s+\d+\s` optional for `_filter_log`
* add observed kernel crash to a test case
* add new match for kernel crash with address range
* add unit tests

### DIFF
--- a/src/clusterfuzz/stacktraces/__init__.py
+++ b/src/clusterfuzz/stacktraces/__init__.py
@@ -146,9 +146,9 @@ class StackParser:
     return False
 
   def update_state_on_match(self,
-                            compiled_regex,
-                            line,
-                            state,
+                            compiled_regex: re.Pattern,
+                            line: str,
+                            state: CrashInfo,
                             new_type=None,
                             new_state=None,
                             new_frame_count=None,
@@ -158,7 +158,7 @@ class StackParser:
                             state_from_group=None,
                             address_filter=lambda s: s,
                             type_filter=lambda s: s,
-                            reset=False):
+                            reset=False) -> re.Match or None:
     """Update the specified parts of the state if we have a match."""
 
     match = compiled_regex.match(line)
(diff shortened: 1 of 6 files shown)


src/clusterfuzz/stacktraces/__init__.py
```diff
@@ -759,15 +759,29 @@ def parse(self, stacktrace: str) -> CrashInfo:
             type_filter=get_fault_description_for_android_kernel)
 
       # Generic KASan errors.
-      if self.update_state_on_match(
+      address_match = self.update_state_on_match(
```
oliverchang, line 762: Nit: no need for these variables. Follow the existing style, adding comments where necessary.

src/clusterfuzz/stacktraces/__init__.py
```diff
@@ -1131,6 +1145,18 @@ def parse(self, stacktrace: str) -> CrashInfo:
           state.crash_address = '0x%s' % android_kernel_match.group(1)
         continue
 
+      # Android kernel stack frame without address
+      android_kernel_match = self.add_frame_on_match(
```
oliverchang, line 1149: same here.

oliverchang (changes requested):

src/clusterfuzz/stacktraces/__init__.py
kennethkang: 1. first and second cases do slightly different things: namely the first adds `0x` to the crash_address. Second case does not.
2. When I was following the existing style, linter kept it quite ugly with 10+ lines crammed in `if()` statement. Not easily readable.

oliverchang: Re 1) Yes they are two cases so I'm suggesting that we have two ifs without these vars. 

Re 2) It's much better for overall readability to be consistent here with the rest of the code in this function/file.

kennethkang: Oh I get what you mean. Thanks for clarifying

src/clusterfuzz/stacktraces/__init__.py
kennethkang: fixed

src/clusterfuzz/stacktraces/__init__.py
kennethkang: fixed

kennethkang:
Rev 2:
* Remove unnecessary variables

oliverchang (approved):

oliverchang (approved):

PR #2452, opened 2021-09-15
libClusterFuzz: Use utils.decode_to_unicode() instead of str.decode().

oliverchang:


oliverchang:
TBR for a trivial fix.

jonathanmetzman:
LGTM. Thanks for fixing!
A test would have been good, but that's fine.

PR #2453, opened 2021-09-15
Do not serialize set

kennethkang:
Rev 1:
* Fix `TypeError` from attempting to serialize a set


src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -161,7 +161,7 @@ def format_record(record):
   }
 
   entry['location'] = getattr(record, 'location', {'error': True})
-  entry['extras'] = getattr(record, 'extras', {})
+  entry['extras'] = getattr(record, 'extras', None)
```
oliverchang, line 164: This isn't correct.

{} on its own is a dict, not a set.

kennethkang: you're right. seems like it's either a set or a bytes type that is being passed that is causing the error. `json.dumps()` seems to have `default` function that can handle this. updating this PR.

oliverchang (approved):
thanks! i'll merge this next week.

PR #2460, opened 2021-09-23
Fix Android kernel job condition.

Huiziyang:
Cuttlefish kernel job doesn't run on "ANDROID_KERNEL" platform so the injected condition in fuzz_task won't work, replace it with environment condition.


oliverchang:
Please merge/rebase from master to fix the CI checks, thanks!

PR #2467, opened 2021-09-30
Don't clear potentially_flaky bit in minimize_task.

oliverchang:
This can otherwise get into a bad state with minimize_task repeatedly
running and unsetting/setting this flag for certain flaky crashes.

minimize_task also spawns a lot of other tasks (e.g. variant) that will
quickly fill up the job queue.


oliverchang:
Example: https://clusterfuzz.com/testcase-detail/5326955048927232

PR #2469, opened 2021-10-05
Add metadata for Testcases that still crash on an unexpected state.

oliverchang:
Set this in progression_task.


jonathanmetzman (approved):
Why are we making this change?

oliverchang:
> Why are we making this change?

This is for some internal triage processes (and it may also make debugging easier).

PR #2471, opened 2021-10-05
Dont consider it a crash when CF stops libFuzzer.

jonathanmetzman:
LibFuzzer sometimes needs to be stopped on its own by CF. CF
does this by sending libFuzzer a SIGTERM. Ordinarily, this causes
libFuzzer to return 72.
(see https://github.com/llvm/llvm-project/blob/1f161919065fbfa2b39b8f373553a64b89f826f8/compiler-rt/lib/fuzzer/FuzzerOptions.h#L25)

Don't consider 72 to be a crashing return code.
Fixes: https://github.com/google/oss-fuzz/issues/6557

### DIFF
--- a/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/constants.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/constants.py
@@ -79,6 +79,7 @@ TARGET_ERROR_EXITCODE = 77
 
 NONCRASH_RETURN_CODES = {
     0,
+    # pylint: disable=line-too-long
     # Code when we interrupt libFuzzer (https://github.com/llvm/llvm-project/blob/1f161919065fbfa2b39b8f373553a64b89f826f8/compiler-rt/lib/fuzzer/FuzzerOptions.h#L25)
     72,
 }


oliverchang:
Thanks! I think you need to rebase to get the CI to pass.

src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -300,7 +300,8 @@ def fuzz(self, target_path, options, reproducers_dir, max_time):
     # If we exited with a non-zero return code with no crash file in output from
     # libFuzzer, this is most likely a startup crash. Use an empty testcase to
     # to store it as a crash.
-    if not crash_testcase_file_path and fuzz_result.return_code:
+    if (not crash_testcase_file_path and
+        fuzz_result.return_code not in constants.NONCRASH_RETURN_CODES):
       crash_testcase_file_path = self._create_empty_testcase_file(
```
oliverchang, line 305: Let's also just change this to create file with less restrictive permissions. There's no reason to lock it down.

jonathanmetzman:
> Thanks! It would be great to get this PR merged. Without it CIFuzz always fails when `report-unreproducible-crashes` is set to `true`

Trying to do this today.

jonathanmetzman:
@oliverchang Is the pip package updated automatically? How can I release a new version to include this change.

oliverchang:
> @oliverchang Is the pip package updated automatically? How can I release a new version to include this change.

We need to make a new release. I'll do it.

PR #2490, opened 2021-11-10
Handle deleted targets from split target builds.

oliverchang:
Previously, if a target was removed from targets.list, existing
testcases will stay open indefinitely.

This change raises an exception during build setup if the expected
target no longer exists in the targets.list, which will lead to
progression_task marking the testcase as invalid.


inferno-chromium (approved):

jonathanmetzman:
LGTM

PR #2496, opened 2021-11-17
return earlier if symbols exist locally

Huiziyang:
We delete symbols archive after unpacking it, but next time `download_system_symbols_if_needed` still tries to find the archive and unpack even though it's not downloaded by `download_artifact_if_needed`.

Move the "check symbols already in cache" out of `download_artifact_if_needed` so `download_system_symbols_if_needed` can return earlier without alerting incorrectly

### DIFF
--- a/src/clusterfuzz/_internal/platforms/android/symbols_downloader.py
+++ b/src/clusterfuzz/_internal/platforms/android/symbols_downloader.py
@@ -61,6 +61,15 @@ def download_artifact_if_needed(
         break
 
 
+def check_symbols_cached(symbols_directory, build_params):
+  build_params_check_path = os.path.join(symbols_directory,
+                                         '.cached_build_params')
+  # Check if we already have the symbols locally.
+  cached_build_params = utils.read_data_from_file(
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/platforms/android/symbols_downloader.py
```diff
@@ -141,6 +137,13 @@ def download_system_symbols_if_needed(symbols_directory):
 
   build_params_check_path = os.path.join(symbols_directory,
```
oliverchang, line 138: Can you add a helper for these rather than duplicate code?

i.e. roughly:

```python
def check_symbols_cached(symbols_directory, build_params):
  build_params_check_path = os.path.join(symbols_directory, ...)
  cached_build_params = ...
  return cached_build_params and cached_build_params == build_params)
```

oliverchang (approved):

src/clusterfuzz/_internal/platforms/android/symbols_downloader.py
Huiziyang: Done

PR #2501, opened 2021-11-30
Add platform override for iOS jobs when filing bugs.

oliverchang:


src/clusterfuzz/_internal/system/environment.py
```diff
@@ -624,6 +624,11 @@ def is_afl_job(job_name=None):
   return get_engine_for_job(job_name) == 'afl'
 
 
+def is_ios_job(job_name=None):
+  """Return True if the current job is for iOS."""
+  return _job_substring_match('ios_', job_name)
```
jonathanmetzman, line 629: I think we really should be explicit instead of doing this pattern.
Since we already do it, I guess it's fine for now.

jonathanmetzman (approved):

src/clusterfuzz/_internal/system/environment.py
oliverchang: Yeah, it's hacky but our convention has been to start the job with the platform. "ios" on its own seems to like it could be too general.

PR #2503, opened 2021-12-01
Add support for the cases where issue.save fails

### LINKED ISSUE #2475: Support fallback components
If a recoverable exception occurs during issue filing, we should support the concept of a fallback component to file into.

Navidem:
Here I added support for fallback components to re-try saving an issue when it exception
Resolves #2475

### DIFF
--- a/src/appengine/libs/issue_management/issue_filer.py
+++ b/src/appengine/libs/issue_management/issue_filer.py
@@ -426,8 +426,8 @@ def file_issue(testcase,
       issue.components.clear()
       issue.components.add(policy.fallback_component)
       if policy.fallback_policy_message:
-        message = policy.fallback_policy_message.format(
-            ' '.join(metadata_components))
+        message = policy.fallback_policy_message.replace(
+            '%COMPONENTS%', ' '.join(metadata_components))
         issue.body += '\n\n' + message
       issue.save()
 
--- a/src/appengine/libs/issue_management/issue_filer.py
+++ b/src/appengine/libs/issue_management/issue_filer.py
@@ -418,8 +418,15 @@ def file_issue(testcase,
     issue.components.add(policy.unreproducible_component)
 
   issue.reporter = user_email
-  issue.save()
 
+  try:
+    issue.save()
+  except:
+    if policy.fallback_component:
+      issue.components.clear()
+      issue.components.add(policy.fallback_component)
+      issue.save()
+     
   # Update the testcase with this newly created issue.
   testcase.bug_information = str(issue.id)
   testcase.put()
--- a/src/appengine/libs/issue_management/issue_filer.py
+++ b/src/appengine/libs/issue_management/issue_filer.py
@@ -418,22 +418,18 @@ def file_issue(testcase,
     issue.components.add(policy.unreproducible_component)
 
   issue.reporter = user_email
-  print("\nBEFORE\n", issue.components.added, issue.components.removed)
-  for v in issue.components.__iter__():
-    print(v)
+
   try:
     issue.save()
   except:
     if policy.fallback_component:
-      issue.components.clear() 
+      issue.components.clear()
       issue.components.add(policy.fallback_component)
       if policy.fallback_policy_message:
-        message = policy.fallback_policy_message.format(' '.join(metadata_components))
+        message = policy.fallback_policy_message.format(
+            ' '.join(metadata_components))
         issue.body += '\n\n' + message
       issue.save()
-      print("AFTER\n", issue.components.added, issue.components.removed)
-      for v in issue.components.__iter__():
-        print(v)
 
   # Update the testcase with this newly created issue.
   testcase.bug_information = str(issue.id)
--- a/src/appengine/libs/issue_management/issue_filer.py
+++ b/src/appengine/libs/issue_management/issue_filer.py
@@ -418,16 +418,22 @@ def file_issue(testcase,
     issue.components.add(policy.unreproducible_component)
 
   issue.reporter = user_email
-
+  print("\nBEFORE\n", issue.components.added, issue.components.removed)
+  for v in issue.components.__iter__():
+    print(v)
   try:
     issue.save()
   except:
     if policy.fallback_component:
-      issue.components.clear()
+      issue.components.clear() 
       issue.components.add(policy.fallback_component)
       if policy.fallback_policy_message:
-        issue.body += '\n\n' + policy.fallback_policy_message
+        message = policy.fallback_policy_message.format(' '.join(metadata_components))
+        issue.body += '\n\n' + message
       issue.save()
+      print("AFTER\n", issue.components.added, issue.components.removed)
+      for v in issue.components.__iter__():
+        print(v)
 
   # Update the testcase with this newly created issue.
   testcase.bug_information = str(issue.id)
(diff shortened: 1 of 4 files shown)


src/appengine/libs/issue_management/issue_filer.py
```diff
@@ -418,8 +418,15 @@ def file_issue(testcase,
...
 
   issue.reporter = user_email
-  issue.save()
 
+  try:
+    issue.save()
+  except:
+    if policy.fallback_component:
+      issue.components.clear()
+      issue.components.add(policy.fallback_component)
```
oliverchang, line 427: Awesome start! Let's add one more thing here. 

Let's append this string to the issue body as well:

```
'\n\n**NOTE**: This bug was filed into this component due to permission or configuration issues with the specified component(s) <components>'
```

Put this in a global _CONSTANT and format in the components from before we cleared.

src/clusterfuzz/_internal/tests/appengine/libs/issue_filer_test.py
```diff
@@ -366,6 +408,36 @@ def test_testcase_metadata_invalid(self):
...
         issue_tracker._itm.last_issue.labels)
 
+  def test_testcase_save_exception(self):
+    """Tests issue filing when issue.save exception"""
+    self.mock.get.return_value = CHROMIUM_POLICY_FALLBACK
+    original_save = monorail.issue.Issue.save
+    helpers.patch(self, ['libs.issue_management.monorail.issue.Issue.save'])
+    
+    def my_save(*args, **kwargs):
+        if getattr(my_save, 'count', 1):
```
oliverchang, line 418: This is a little confusing to read. Why would `'count'` be 1 by default?

Let's rename 'count' to 'raise_exception' and write like so:

```python
if getattr(my_save, 'raise_exception', None):
  raise Exception('Boom!')
else:
  setattr(my_save, 'raise_exception', True)
  return original_save(...)
```

oliverchang (changes requested):

src/clusterfuzz/_internal/tests/appengine/libs/issue_filer_test.py
Navidem: We want it raise exception at the first call to save, and the second call to be normal. So I believe something like this works:

```python
        if getattr(my_save, 'raise_exception', True):
            setattr(my_save, 'raise_exception', False)
            raise Exception("Boom!")
        else:
            return original_save(...)
```

src/appengine/libs/issue_management/issue_filer.py
Navidem: > Put this in a global _CONSTANT and format in the components from before we cleared.

Sorry I'm not sure I follow. Do you mean to use something in the policy and then put it in issue.body like the following?

```python
    issue.body += '\n\n' + policy.deadline_policy_message
```

oliverchang: Yep exactly :)

src/clusterfuzz/_internal/tests/appengine/libs/issue_filer_test.py
oliverchang: Ah yep, that's the correct one :)

src/appengine/libs/issue_management/issue_filer.py
Navidem: Done.

src/clusterfuzz/_internal/tests/appengine/libs/issue_filer_test.py
Navidem: applied :)

src/appengine/libs/issue_management/issue_filer.py
oliverchang: Please modify the tests to account for this.

Also we need to substitute in the components into the message. Let's make the 'fallback_policy_message' support a "%COMPONENTS%" substitution.

i.e.

Given "Blah blah %COMPONENTS" as "fallback_policy_message", we expand that into "Blah blah Component>Blah, Component2>Blah"

Navidem: Done.

src/appengine/libs/issue_management/issue_filer.py
```diff
@@ -418,7 +418,18 @@ def file_issue(testcase,
...
+
+  try:
+    issue.save()
+  except:
+    if policy.fallback_component:
+      issue.components.clear()
+      issue.components.add(policy.fallback_component)
+      if policy.fallback_policy_message:
+        message = policy.fallback_policy_message.format(
+            ' '.join(metadata_components))
```
oliverchang, line 430: Rather than using .format(), can you replace `"%COMPONENTS%"` instead? '{}' is not descriptive and not extensible if we want to add more substitutions.

Navidem: Got it! Done.

src/clusterfuzz/_internal/tests/appengine/libs/issue_filer_test.py
```diff
@@ -366,6 +411,45 @@ def test_testcase_metadata_invalid(self):
...
+    self.mock.save.side_effect = my_save
+
+    issue_tracker = monorail.IssueTracker(IssueTrackerManager('chromium'))
+    issue_filer.file_issue(self.testcase5, issue_tracker)
+
+    six.assertCountEqual(self,
+                         ['fallback>component', '-component1', '-component2'],
+                         issue_tracker._itm.last_issue.components)
+    self.assertTrue(
+        issue_tracker._itm.last_issue.body.find('**NOTE**: This bug was') != -1)
```
oliverchang, line 436: nit: use `self.assertIn`. 

Also let's test that the substitution worked as well.

oliverchang (approved):
LGTM with one last comment!

src/clusterfuzz/_internal/tests/appengine/libs/issue_filer_test.py
Navidem: > Also let's test that the substitution worked as well.

This is testing it:

```python
    self.assertTrue(
        issue_tracker._itm.last_issue.body.endswith(
            ' '.join(metadata_components)))
```

Do you mean something like this instead?

```python
    self.assertTrue(
        issue_tracker._itm.last_issue.body.endswith(
            'component1 component2')))
```

oliverchang: Ah I missed that. What I meant was:

```python
self.assertIn('**NOTE**: This bug was .. (the full message)...  component1 component2)', last_issue.body)
```

To be more explicit.

oliverchang:
Also please fix the lint failure in https://github.com/google/clusterfuzz/runs/4401411018?check_suite_focus=true

Navidem:
> Also please fix the lint failure in https://github.com/google/clusterfuzz/runs/4401411018?check_suite_focus=true

oops, just saw this!

src/clusterfuzz/_internal/tests/appengine/libs/issue_filer_test.py
Navidem: Done.

PR #2504, opened 2021-12-02
Jira integration fixes

quanyangsc:
This fixes two Jira integration issues:
1. Jira objects are not JSON encodable, so this fixes it by returning the raw representation of the Jira object which happens to be in JSON.
2. The format string in the find_issues method wasn't assigned back to the variable and so it wasn't working as intended.

Apart from that, this also implements the `find_issues_url` method by reusing the code from `find_issues`.

### DIFF
--- a/src/appengine/handlers/base_handler.py
+++ b/src/appengine/handlers/base_handler.py
@@ -56,24 +56,24 @@ class JsonEncoder(json.JSONEncoder):
   """Json encoder."""
   _EPOCH = datetime.datetime.utcfromtimestamp(0)
 
-  def default(self, obj):  # pylint: disable=arguments-differ,method-hidden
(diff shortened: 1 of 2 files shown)


oliverchang (approved):
LGTM, thanks!

quanyangsc:
The latest lint error doesn't seem to be relevant to my changes, I could rename the variable back to `o` as the linter is asking.
@inferno-chromium @oliverchang wdyt?

oliverchang:
> The latest lint error doesn't seem to be relevant to my changes, I could rename the variable back to `o` as the linter is asking. @inferno-chromium @oliverchang wdyt?

Sorry about that. I just edited this in your PR to fix that issue.

PR #2505, opened 2021-12-03
Include recovered issue filing exceptions in triage message.

oliverchang:
For issues filed to a fallback component, surface the exception in the
testcase triage message.

Also fix incorrect behaviour with failed issue filings not raising any
exception.

### DIFF
--- a/src/clusterfuzz/_internal/tests/appengine/handlers/cron/triage_test.py
+++ b/src/clusterfuzz/_internal/tests/appengine/handlers/cron/triage_test.py
@@ -354,7 +354,7 @@ class CheckAndUpdateSimilarBug(unittest.TestCase):
 
 
 @test_utils.with_cloud_emulators('datastore')
-class FileIssue(unittest.TestCase):
+class FileIssueTest(unittest.TestCase):
   """Tests for _file_issue."""
 
   def setUp(self):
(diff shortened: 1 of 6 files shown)


src/clusterfuzz/_internal/tests/appengine/libs/issue_filer_test.py
```diff
@@ -453,7 +455,8 @@ def my_save(*args, **kwargs):
     setattr(my_save, 'raise_exception', True)
     issue_tracker = monorail.IssueTracker(IssueTrackerManager('chromium'))
     self.mock.get.return_value = CHROMIUM_POLICY
-    issue_filer.file_issue(self.testcase1, issue_tracker)
+    with self.assertRaises(Exception):
+      issue_filer.file_issue(self.testcase1, issue_tracker)
```
oliverchang, line 459: @Navidem I missed this in the last review, but the correct behaviour here should've been to raise the exception if we fail on the first save() without a fallback.

src/clusterfuzz/_internal/tests/appengine/handlers/cron/triage_test.py
```diff
@@ -351,3 +351,40 @@ def test_similar_testcase_with_issue_recently_closed(self):
     self.assertEqual(
         'Delaying filing a bug since similar testcase (2) in issue (1) '
         'was just fixed.', testcase.get_metadata(triage.TRIAGE_MESSAGE_KEY))
+
+
+@test_utils.with_cloud_emulators('datastore')
+class FileIssue(unittest.TestCase):
```
jonathanmetzman, line 357: I usually end these classes with `Test`, Eg `FileIssueTest`

jonathanmetzman (approved):
lgtm

src/clusterfuzz/_internal/tests/appengine/handlers/cron/triage_test.py
oliverchang: Done!

PR #2516, opened 2021-12-16
Ignore lost connection crash during repro

kennethkang:
Revision 1:
* clean up existing repro logic
* add ignore logic for repro crash

### DIFF
--- a/src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
@@ -19,6 +19,7 @@ import re
 import tempfile
 import threading
 import time
+from typing import List
 
 import requests
 
@@ -35,7 +36,7 @@ from clusterfuzz.fuzz import engine
 
 LOCAL_HOST = '127.0.0.1'
 RAWCOVER_RETRIEVE_INTERVAL = 180  # retrieve rawcover every 180 seconds
-REPRODUCE_REGEX = re.compile(r'reproduced (\d+) crashes')
+REPRODUCE_REGEX = re.compile(r'.*all done. reproduced (\d+) crashes.')
 
 
 def get_work_dir():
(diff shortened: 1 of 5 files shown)


kennethkang:
@oliverchang wanted to get your feedback. I will add a uni test case for this new ignore crash logic.

kennethkang:
sorry about force push. I keep `git commit --amend` from old habit

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
```diff
@@ -163,15 +164,17 @@ def _create_empty_testcase_file(self):
     _, path = tempfile.mkstemp(dir=fuzzer_utils.get_temp_dir())
     return path
 
-  def _crash_was_reproducible(self, output):
+  def _crash_was_reproducible(self, output: str) -> bool:
```
kennethkang, line 167: this is just a cleanup but the same logic

src/clusterfuzz/_internal/bot/fuzzers/syzkaller/runner.py
```diff
@@ -181,7 +184,7 @@ def repro(self, repro_timeout, repro_args):
     logs.log('Running Syzkaller testcase.')
```
oliverchang, line 184: Can we add the retry here instead? 

i.e. if `lost connection to test machine` is in the logs, retry until we hit the retry limit or it's no longer in the logs. 

The `IGNORE_REPRODUCED_CRASH_TYPE` logic is very syzkaller/android specific, and it's better to keep it closer to here.

kennethkang: >better to keep it closer to here
I agree that this issue is very syzkaller specific. What I'm afraid is changes here won't actually resolve the reliably reproducible flag issue. Seems like this logic of marking whether the crash reproduces is independent of the `reliably reproducible` flag that gets set by a job running `test_reproduce_reliability`?

kennethkang: `runner.repro()` seems to currently only check whether the repro job is done and whether the crash happened or not (even if the crash may be of a wrong type)

oliverchang: > Seems like this logic of marking whether the crash reproduces is independent of the reliably reproducible flag that gets set by a job running test_reproduce_reliability?

It should end up with the same outcome if we do it here, unless I'm misunderstanding your question. `test_reproduce_reliability` etc will all end up calling this code. It's also better to fix this here so that it generalises to anywhere else that tries to reproduce a syzkaller testcase. 

What I'm suggesting is to add logic to `runner.repro` to retry if the "lost connection" message is in the log somewhere.

kennethkang: ohh I think I understand where the gap is.

I don't know about `libFuzzer` or other fuzzers, but for Syzkaller it has its own tool called `syz-crush` for reproducing. `runner.repro()` uses `syz-crush` which outputs `syz-crush` log which is different than the log `syzkaller` originally produced. I am curious how clusterfuzz reads crash state from syz-crush output because `syz-crush` log does not provide any call trace.

kennethkang: I found where `syz-crush` stashes its logs. It does produce an output log like the original one including call trace (crash state). This current implementation doesn't actually open this log and uses `syz-crush` output. I'll make the appropriate change.

kennethkang: opening a new PR.

PR #2518, opened 2021-12-23
correctly assign ccs from issue trackers config.yaml

boni-sha:
Currently if ccs are defined in issue_trackers/config.yaml then these are incorrectly assigned as labels instead of ccs/watchers.
This PR rectifies the variable where ccs should be assigned.


oliverchang (approved):
thanks a lot! (and sorry for the delay!)

PR #2520, opened 2022-01-05
Increase OSS-Fuzz build failure threshold by one.

oliverchang:
AFL++ breakages are far more common, and there's no easy way to
distinguish AFL++ specific failures vs other engine failures currently.

### DIFF
--- a/src/appengine/handlers/cron/oss_fuzz_build_status.py
+++ b/src/appengine/handlers/cron/oss_fuzz_build_status.py
@@ -49,7 +49,7 @@ TIMESTAMP_FORMAT = '%Y-%m-%dT%H:%M:%S'
 NO_BUILDS_THRESHOLD = datetime.timedelta(days=2)
 
 # Minimum number of consecutive build failures before filing a bug.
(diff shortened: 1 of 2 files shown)


jonathanmetzman (approved):
lgtm

jonathanmetzman:
Keep in mind this can have some unintended consequences.
I think the longer we take to notify a dev of an issue (i.e. one breaking their build) the less likely they are to fix it.

MarcoFalke:
No objection, but it might help if there was a way to keep the count for specific projects (opt-in) at `2` (or maybe at `1`, if they don't care about intermittent git clone network errors)

oliverchang:
> Keep in mind this can have some unintended consequences. I think the longer we take to notify a dev of an issue (e.g. one breaking their build) the less likely they are to fix it.

Yeah, but just a single extra day shouldn't hurt too much I hope.

> No objection, but it might help if there was a way to keep the count for specific projects (opt-in) at 2 (or maybe at 1, if they don't care about intermittent git clone network errors)

Created https://github.com/google/clusterfuzz/issues/2521 for this.

PR #2523, opened 2022-01-06
Minimize syzkaller testcases

kennethkang:
Rev 1:
* run `minimize` when test case reproduced successfully
* implement minimization
* unit test

### DIFF
--- a/src/clusterfuzz/_internal/bot/fuzzers/syzkaller/engine.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/syzkaller/engine.py
@@ -164,7 +164,8 @@ class Engine(engine.Engine):
       target_path: str,  # pylint: disable=unused-argument
       input_path: str,
       arguments: List[str],  # pylint: disable=unused-argument
-      max_time: int) -> engine.ReproduceResult:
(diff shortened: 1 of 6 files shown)


src/clusterfuzz/_internal/bot/fuzzers/syzkaller/engine.py
```diff
@@ -179,12 +180,24 @@ def reproduce(
...
     binary_dir = self.prepare_binary_path()
-    syzkaller_runner = runner.get_runner(
+    syzcrush_runner = runner.get_runner(
         os.path.join(binary_dir, constants.SYZ_CRUSH))
-    repro_args = runner.get_config()
-    repro_args.extend(
-        ['-infinite=false', '-restart_time={}s'.format(REPRO_TIME), input_path])
-    return syzkaller_runner.repro(max_time, repro_args=repro_args)
+    syzcrush_args = (
+        runner.get_config() +
```
kennethkang, line 186: if `runner.get_config()` in the future gets cached and returns the same object for multiple functions, mutation via `extend()` causes side effect we do not want.

PR #2526, opened 2022-01-09
Change actions and merged_into into a property

quanyangsc:
The jira integration was sending requests like `/rest/api/2/issue/<bound method Issue.merged_into of <libs.issue_management.jira.Issue object at 0x3e14ae683510>>`, and that was due to the merged_into/actions method not being a property.


oliverchang (approved):
thanks a lot for catching and fixing this!

PR #2529, opened 2022-01-13
allow badly setup projects for afl++ still working

vanhauser-thc:
For project that dlopen() instrumented libraries this is an issue for afl++ collision free coverage as the map size must be known when the forkserver is run.
So the current implementation bails when this is detected as this means the fuzzing setup is inefficient.
The solution is to AFL_PRELOAD such dlopen'ed libs.
For the time being though lets allow such setup to still run (less efficient) instead of aborting.


oliverchang (approved):
LGTM, thank! I'll let @jonathanmetzman take a look as well in case he has any concerns.

jonathanmetzman (approved):
LGTM

jonathanmetzman:
Will merge on monday

oliverchang:
@jonathanmetzman is this ready to merge?

vanhauser-thc:
it should be merged asap to ensure that all projects are able to run.

Issue #2533, opened 2022-01-17
Throttle / limit number of times we file an unreproducible crash


oliverchang:
This can lead to spam: https://bugs.chromium.org/p/oss-fuzz/issues/list?q=%22Sanitizer%20CHECK%20failure%22%20curl&can=1 and developer confusion: https://github.com/google/oss-fuzz/issues/6879

DavidKorczynski:
Another spam-related issue from OSS-Fuzz is this one https://github.com/google/oss-fuzz/issues/7040

Monorail query showing 100+ reports for same issue: https://bugs.chromium.org/p/oss-fuzz/issues/list?q=%22ASSERT%3A%20byte%20index%22%20json5format&can=1

oliverchang:
Thanks @DavidKorczynski That looks like a separate issue with our grouping. All these testcases do end up getting grouped into the [one group](https://oss-fuzz.com/testcases?q=group%3A5421756379693056), but it happens too late after the bugs are filed.

MarcoFalke:
An alternative to creating a new issue to re-open the existing one. See also https://github.com/google/oss-fuzz/issues/6791

oliverchang:
> An alternative to creating a new issue to re-open the existing one. See also [google/oss-fuzz#6791](https://github.com/google/oss-fuzz/issues/6791)

Re-opening is something that doesn't fit well into our current bug triage workflow. For the unreproducible issues, we now only ever file them once.

PR #2536, opened 2022-01-19
Only file unreproducible issue for a crash type / group once.

oliverchang:
If a previous similar unreproducible issue was already filed and closed,
don't file it again to avoid spamming the issue tracker.

Part of #2533.


oliverchang:
We could also also set a long interval instead: i.e. file at most one unreproducible crash of a given crash type/group per few months. This will still result in some spam though and contribute to developer fatigue, so I went with this more aggressive approach instead.

inferno-chromium (approved):

jonathanmetzman (approved):
lgtm

PR #2539, opened 2022-01-26
[fuchsia] Add Fuchsia-specific list of stale processes to kill

eepeep:
The default behavior matched processes with cmdlines including the
builds directory. This ended up inadvertently killing the qemu
instance, which is intended to be long-running.

### DIFF
--- a/src/clusterfuzz/_internal/system/process_handler.py
+++ b/src/clusterfuzz/_internal/system/process_handler.py
@@ -565,6 +565,13 @@ def terminate_stale_application_instances():
     # Artifical sleep to let the processes get terminated.
     time.sleep(1)
 
+  elif platform == 'FUCHSIA':
+    processes_to_kill += [
+        'undercoat',
+        llvm_symbolizer_filename,
+    ]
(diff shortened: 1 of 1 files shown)


eepeep:
cc @oliverchang

src/clusterfuzz/_internal/system/process_handler.py
```diff
@@ -565,6 +565,16 @@ def terminate_stale_application_instances():
...
 
+  elif platform == 'FUCHSIA':
+    processes_to_kill += [
+        'undercoat',
+        llvm_symbolizer_filename,
+    ]
+    terminate_processes_matching_names(processes_to_kill, kill=True)
+
+    # Artifical sleep to let the processes get terminated.
+    time.sleep(1)
```
oliverchang, line 576: We can probably skip this. This hack looks Windows-specific.

oliverchang (approved):
LGTM!

src/clusterfuzz/_internal/system/process_handler.py
eepeep: Ok; removed!

PR #2542, opened 2022-01-28
Fetch Fuzzer entities more frequently in project_setup.

oliverchang:
project_setup is long running and currently only fetches the Fuzzer
entities once at the beginning. This makes manual Fuzzer updates by humans
prone to getting overwritten.


oliverchang:
TBR

PR #2544, opened 2022-02-02
Update URL for cloud storage browser.

eepeep:
Clicking on, for example, a fuzzer corpus link would result in a "URL
not found" error. This updates the URL to a working one.


oliverchang (approved):
thanks!

PR #2550, opened 2022-02-07
Jira tracker returning tz-aware datetime

quanyangsc:
This modifies the Jira Issue class to return tz-naive datetime instead as there are areas in ClusterFuzz where tz-naive datetime is expected but the jira integration is returning tz-aware datetime.

Example:
https://github.com/google/clusterfuzz/blob/4f8020c4c7ce73c1da0e68f04943af30bb5f0b32/src/clusterfuzz/_internal/base/dates.py#L43


oliverchang (approved):
thanks!

PR #2554, opened 2022-02-08
[fuchsia] Remove old, pre-undercoat code

eepeep:
Now that undercoat is the sole interface between ClusterFuzz and
Fuchsia, support for pre-undercoat builds can be removed. This will
potentially cause failures during regression testing against older
builds, but this should be uncommon and not block any other tasks.


eepeep:
cc @oliverchang

We don't need to land this immediately, but just wanted to get something up for now.  Do let me know if you see anything I'm missing.

oliverchang (approved):
nice!

oliverchang:
Could you please investigate the test failure? The error message is not very helpful, but I'm guessing that it's a side effect of some test module not being able to be imported properly.

eepeep:
Sorry, classic mistake of an uncommitted local change... looks good now.

oliverchang:
Thanks!

PR #2555, opened 2022-02-08
Allowing multiple trials of each ClusterFuzz reproduction request

Alan32Liu:
Allowing multiple trials of each ClusterFuzz reproduction request:
1. A new parameter named `numTrials` in messages to topic needs-update;
2. Accept a JSON array of `stacktrace`s in messages from topic bug-update, with backward compatibility of the old format;
3. Three new test cases for the new message format.


Alan32Liu:
I did not use more elegant/pythonic functions (e.g `any()` or `all()`) [at this block](https://github.com/google/clusterfuzz/blob/805c40eab5803c3e18738430ed9a61a94359d3c7/src/appengine/handlers/external_update.py#L85) and [this block](https://github.com/google/clusterfuzz/blob/805c40eab5803c3e18738430ed9a61a94359d3c7/src/appengine/handlers/external_update.py#L103) to avoid the overhead of repeating the similarity/security_issue comparisons later [at here](https://github.com/google/clusterfuzz/blob/805c40eab5803c3e18738430ed9a61a94359d3c7/src/appengine/handlers/external_update.py#L122). But please let me know if you'd prefer me to use `any()` or `all()` instead.

oliverchang (commented):
nice! could you please run `python butler.py format` and fix any other lint issues reported by `python butler.py lint` ?

Alan32Liu:
Commit [9c8ff17](https://github.com/google/clusterfuzz/pull/2555/commits/9c8ff1758e58c6a9e072a30c0b04b7241c560800) - [628df5e](https://github.com/google/clusterfuzz/pull/2555/commits/628df5e99fb64bf163b62a696994f416013ea614) are added by `--rebase`, which I had no contribution at all. Should I remove them from my branch later?

src/appengine/handlers/external_update.py
```diff
@@ -23,6 +24,9 @@
 from libs import handler
 from libs import helpers
 
+OLD_PROTOCOL = "1"  # Old message format: Data is a stacktrace.
```
oliverchang, line 27: nit: use single quotes for all strings.

Alan32Liu: Fixed in [9ef80a7](https://github.com/google/clusterfuzz/pull/2555/commits/9ef80a7a864ff6bd1731b0da3c80cd6901b3dacf).

src/appengine/handlers/external_update.py
```diff
@@ -70,31 +86,57 @@ def handle_update(testcase, revision, stacktrace, error):
...
-
-  state = stack_analyzer.get_crash_data(
-      stacktrace,
-      fuzz_target=fuzz_target_name,
-      symbolize_flag=False,
-      already_symbolized=True,
-      detect_ooms_and_hangs=True)
-  crash_comparer = CrashComparer(state.crash_state, testcase.crash_state)
-  if not crash_comparer.is_similar():
+  states = [
```
oliverchang, line 89: There's a bit of complexity in the way you are handling these multiple values. One cleaner way to do this is to just factor out the existing code for validating a single stacktrace into its own function, e.g. `is_still_crashing`, and just call this in a loop in this function. We can also early exit on the first true result, and avoid having to do crash comparisons and other analysis all up front.

i.e.

```python
...
data_handler.record_fuzz_target(fuzz_target.engine, fuzz_target.binary,
                                  testcase.job_type)
for stacktrace in stacktraces:
  if is_still_crashing(testcase, stacktrace):
    logs.log(f'{testcase.key.id()} still crashes.')
    return

# If we get here, it means that every single trial resulted in a non-crash. Close the testcase.
logs.log(f'No matching crash detected in {len(stacktraces)} trials, marking as fixed.')
_mark_as_fixed(testcase, revision)
...
```

oliverchang (commented):
nice! Have a suggestion to make this much simpler.

src/appengine/handlers/external_update.py
```diff
@@ -45,9 +49,41 @@ def _mark_errored(testcase, revision, error):
       testcase, revision, message=message)
 
 
-def handle_update(testcase, revision, stacktrace, error):
+def handle_update(testcase, revision, stacktraces, error, protocol_version):
   """Handle update."""
-  logs.log('Got external update for testcase.', testcase_id=testcase.key.id())
+
+  def is_still_crashing(st_index, stacktrace):
+    """
```
oliverchang, line 56: Nit: we generally start the doctring on the same line as """.

e.g.

```
"""Check if the given stacktrace......"""
```

src/clusterfuzz/_internal/tests/appengine/handlers/external_update_test.py
```diff
@@ -118,6 +121,7 @@ def test_update_changed_security(self):
     self.testcase.put()
 
     stacktrace = self._read_test_data('asan_uaf.txt')
+    print(DATA_DIRECTORY)
```
oliverchang, line 124: Delete?

oliverchang (approved):
LGTM with minor nits!

src/clusterfuzz/_internal/tests/appengine/handlers/external_update_test.py
Alan32Liu: Forgot to hide my curiosity 😆

src/appengine/handlers/external_update.py
Alan32Liu: Fixed in [4febc0c](https://github.com/google/clusterfuzz/pull/2555/commits/4febc0c67d15b22ddaa8680818321a3e535470c3).

src/clusterfuzz/_internal/tests/appengine/handlers/external_update_test.py
Alan32Liu: Fixed in [e49f843](https://github.com/google/clusterfuzz/pull/2555/commits/e49f8435ca87a5f640c0abce42fbd980095bf365).

Alan32Liu:
The following assertion fails every time when running the unit tests locally, but always passes Github's automatic tests.

```python
FAIL: test_lease_reach_latest (tests.appengine.common.tasks_test.LeaseTaskTest)
Test leasing a task and reaching the maximum lease time.
  File "/usr/local/google/home/donggeliu/Code/clusterfuzz/src/clusterfuzz/_internal/tests/appengine/common/tasks_test.py", line 213, in test_lease_reach_latest
    self.assertEqual(3, message.modify_ack_deadline.call_count)
AssertionError: 3 != 2
```

oliverchang:
I can't reproduce this locally, even with your change applied. Could you file an issue for this so we can investigate later?

PR #2556, opened 2022-02-09
Fix jira integration error where issue id is expected to be integer

quanyangsc:
**Context**
Clusterfuzz expects issue.id to be integer in various parts of the code, however, the current implementation of the jira integration returns a string value that is not int-parseable (e.g. fuzz-123), this results in exceptions thrown during cronjobs that fails.

This PR fixes that by returning the id of an issue instead of the key, however, doing this requires making a call to the Jira instance to fetch the key value in order to craft the issue URL (which we memoize for 24 hours after).

Additionally this PR also fixes an issue that I introduced in a previous PR due to an invalid module path: https://github.com/google/clusterfuzz/blob/master/src/appengine/libs/issue_management/jira/issue_tracker_manager.py#L90 (also added a unit test to catch this)

### DIFF
--- a/src/appengine/libs/issue_management/jira/__init__.py
+++ b/src/appengine/libs/issue_management/jira/__init__.py
@@ -43,6 +43,11 @@ class Issue(issue_tracker.Issue):
   @property
   def id(self):
     """The issue identifier."""
+    return int(self.jira_issue.id)
+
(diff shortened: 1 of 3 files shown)


oliverchang (approved):
thanks!

PR #2560, opened 2022-02-10
[fuchsia] Remove old, pre-undercoat code (re-land)

eepeep:
Now that undercoat is the sole interface between ClusterFuzz and
Fuchsia, support for pre-undercoat builds can be removed. This will
potentially cause failures during regression testing against older
builds, but this should be uncommon and not block any other tasks.


eepeep:
cc @oliverchang 

This is a followup to #2559, removing the offending reference (since it was a no-op function anyways.)  It's a separate PR because I couldn't figure out how to push to #2554 now that it's closed.

Apologies for missing the stray reference!

oliverchang (approved):
Thanks! I'll wait until Monday to merge :)

PR #2565, opened 2022-02-18
Fix App Engine dependency breakage.

oliverchang:
Pin markupsafe, itsdangerous to older versions.

All dependencies will be pinned in a later PR.


oliverchang:
tbr

jonathanmetzman (commented):
lgtm

PR #2568, opened 2022-02-23
Add DISABLE_VARIANT option for jobs.

eepeep:
If set to True, variant tasks will not be created for that job. This is
useful in cases where a project's jobs don't contain the same fuzzers.


oliverchang (approved):
LGTM, thanks!

PR #2571, opened 2022-03-03
Add customize-able project labels applying to all fuzz targets in a Project

asraa:
Signed-off-by: Asra Ali <asraa@google.com>

Part of https://github.com/google/oss-fuzz/issues/7318

In Project setup, checks for a wildcard label defined in the `project.yaml`'s `labels` field to apply project-wide labels when filing issues. (Previously only per-fuzz target labels were supported). See issue for use-case.

### DIFF
--- a/src/appengine/handlers/cron/project_setup.py
+++ b/src/appengine/handlers/cron/project_setup.py
@@ -790,11 +790,14 @@ class ProjectSetup(object):
             backup_bucket=backup_bucket_name)
 
       if self._add_info_labels:
-        job.environment_string += (
-            'AUTOMATIC_LABELS = Proj-{project},Engine-{engine}\n'.format(
-                project=project,
-                engine=template.engine,
-            ))
+        automatic_labels = 'Proj-{project},Engine-{engine}'.format(
+            project=project,
+            engine=template.engine,
+        )
+        labels = info.get('labels')
+        if labels and '*' in labels:
+          automatic_labels += ',%s' % ','.join(labels['*'])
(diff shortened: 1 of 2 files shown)


asraa:
cc @krajshiva

src/appengine/handlers/cron/project_setup.py
```diff
@@ -790,11 +790,14 @@ def _sync_job(self, project, info, corpus_bucket_name, quarantine_bucket_name,
...
-                engine=template.engine,
-            ))
+        automatic_labels = 'Proj-{project},Engine-{engine}'.format(
+            project=project,
+            engine=template.engine,
+        )
+        labels = info.get('labels')
+        if labels and '*' in labels:
+          automatic_labels += ',%s' % ','.join(labels['*'])
+        job.environment_string += ('AUTOMATIC_LABELS = %s\n' % automatic_labels)
```
oliverchang, line 800: nit: can we clean this up a bit by doing:

```python
automatic_labels = [
  f 'Project-{project}',
  f 'Engine-{engine}',
]

labels = info.get('labels')
if labels and '*' in labels:
  automatic_labels.extend(labels['*'])

automatic_labels = ','.join(automatic_labels)
job.environment_string += f'AUTOMATIC_LABELS = {automatic_labels}\n'
```
```

oliverchang (approved):
Thanks a lot for implementing this @asraa !!!

LGTM -- Just one comment on making this a bit cleaner.

src/appengine/handlers/cron/project_setup.py
asraa: nice, done! thanks

PR #2572, opened 2022-03-03
Don't include libFuzzer output in log line on failure.

jonathanmetzman:
Exceptions in this file have been including libFuzzer's output in
the log line on failure.
This results in some very ugly logs showing up in stackdriver.

Fixes: #2570


jonathanmetzman:
There's a minor issue with this solution.
It still raises an exception even though we've called log error, so the same error will get logged twice, once locally where we know how to log properly, and once somewhere further up when the task fails.
I don't really know what to do about this, other than maybe refactoring the code to not exception on these errors.

oliverchang (approved):

PR #2573, opened 2022-03-04
Pin all vendored dependencies.

oliverchang:
Use a Pipfile to generate the requirements.txt to vendor.

For App Engine only dependencies, we use "dev-packages" as a hack to
make sure that version resolution always works and doesn't introduce
conflicts.


jonathanmetzman (approved):
LGTM

Issue #2564, opened 2022-02-18
Pass scorecard checks


oliverchang:

jonathanmetzman:
I don't think this is a priority anymore.

PR #2581, opened 2022-03-23
Add dependabot config.

oliverchang:


jonathanmetzman (approved):
LGTM

Thanks!

jonathanmetzman:
Related: https://github.com/google/clusterfuzz/issues/2564

Alan32Liu (approved):
LGTM!

jonathanmetzman:
Is this missing the docs/ folder which has ruby packages? Or do we not care since it isn't production code?

jonathanmetzman:
Also what about the dependencies of our front end code? Bower etc.

oliverchang:
docs/ isn't as high priority as you mention, but we can certainly add it. 

Our bower stuff is also in a weird state since Bower is kind of dead and we should migrate to npm/package.json instead. We can address that then.

PR #2595, opened 2022-03-30
Fix typos in attribute names

Alan32Liu:


src/appengine/libs/issue_management/issue_filer.py
```diff
@@ -293,7 +293,7 @@ def notify_issue_update(testcase, status):
 
   if status in ('verified', 'wontfix'):
     logs.log(f'Closing issue {testcase.github_issue_num} '
-             f'in GitHub repo {testcase.github_id}: '
+             f'in GitHub repo {testcase.github_repo_id}: '
```
jonathanmetzman, line 296: Very odd that this wasn't caught by linting.

Alan32Liu: Oliver and I decided to add some specific mock tests that will execute these functions later : )

PR #2597, opened 2022-03-30
Create codeql-analysis.yml

oliverchang:


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,68 @@
+# For most projects, this workflow file will not need changing; you simply need
```
jonathanmetzman, line 1: Looks like there's quite a few complaints about some go code.

jonathanmetzman (approved):
lgtm

.github/workflows/codeql-analysis.yml
oliverchang: I marked all of these as WontFix, since it's meant to run in a trusted context.

PR #2605, opened 2022-04-01
Update container image tag

dbaker-rh:
The original container image tags (`ce67e3c-202004060711`) pull down Ubuntu 16.04 UTS.  The replacement (`a2f4dd6-202202070654`) is 20.04 LTS.

I notice the sample / proof of concept Heartbleed test case works with both.

However, the real life example binaries I used to test the clusterfuzz instance are more particular in other libraries between the two LTS releases and failed to run without updating to a newer base image.

I expect this small patch will help others out as they get their instance operational.

### DIFF
--- a/configs/test/gce/clusters.yaml
+++ b/configs/test/gce/clusters.yaml
@@ -88,7 +88,7 @@ test-clusterfuzz:
         metadata:
           items:
             - key: docker-image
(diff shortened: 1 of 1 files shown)


oliverchang (approved):
Thanks a lot for this!

PR #2621, opened 2022-04-20
Upgrade all Docker images to latest.

oliverchang:
Un-pin from the Xenial revision.


Alan32Liu (approved):
That is much simpler than I thought : )

oliverchang:
> That is much simpler than I thought : )

This is only the first part :) Once they're built, we need to upgrade our configs to roll it out and make sure nothing breaks.

jonathanmetzman (approved):
LGTM

PR #2623, opened 2022-04-21
strip the .git suffix when necessary

Alan32Liu:
Some project's `main_repo` URL ends with a `.git` suffix.
Remove that suffix when necessary to avoid confusion when filing issues.

### DIFF
--- a/src/appengine/libs/issue_management/oss_fuzz_github.py
+++ b/src/appengine/libs/issue_management/oss_fuzz_github.py
@@ -21,6 +21,7 @@ from clusterfuzz._internal.datastore import data_handler
 from clusterfuzz._internal.metrics import logs
 
 GITHUB_PREFIX = 'https://github.com/'
+GITHUB_SUFFIX = '.git'
 
 TESTCASE_REPORT_URL = 'https://{domain}/testcase?key={testcase_id}'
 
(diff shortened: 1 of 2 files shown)


src/appengine/libs/issue_management/oss_fuzz_github.py
```diff
@@ -21,6 +21,7 @@
 from clusterfuzz._internal.metrics import logs
 
 GITHUB_PREFIX = 'https://github.com/'
+GITHUB_SUFFIX = '.git'
```
oliverchang, line 24: nit: this is not necessarily a "GitHub" specific suffix, so let's name this GIT_SUFFIX.

src/appengine/libs/issue_management/oss_fuzz_github.py
```diff
@@ -94,6 +95,8 @@ def _get_repo(testcase, access):
     logs.log(f'MAIN REPO is not a GitHub url: {repo_url}.')
     return None
   repo_name = repo_url[len(GITHUB_PREFIX):]
+  if repo_url[-len(GITHUB_SUFFIX):] == GITHUB_SUFFIX:
```
oliverchang, line 98: nit: just use repo_url.endswith(GIT_SUFFIX)

oliverchang (changes requested):

oliverchang (approved):
nice!!

PR #2632, opened 2022-04-28
Add introspector report links to oss-fuzz-home

Navidem:


oliverchang:
In addition to testing this on staging, please also fix the test failures.

Navidem:
> 



> In addition to testing this on staging, please also fix the test failures.

Done.

Navidem:
It's weird that the `isort` test passes on my local machine but fails here!

src/appengine/handlers/home.py
```diff
@@ -24,8 +28,12 @@
 from libs import handler
 from libs import helpers
 
+INTROSPECTOR_BUCKET = 'oss-fuzz-introspector'
+INTROSPECTOR_INDEX_JSON = 'build_status.json'
+
 MEMCACHE_TTL_IN_SECONDS = 30 * 60
 
+logging.basicConfig(level=logging.INFO)
```
oliverchang, line 36: Let's delete this line.

src/appengine/handlers/home.py
```diff
@@ -60,9 +68,22 @@ def get_single_fuzz_target_or_none(project, engine_name):
...
 
+def get_introspector_index():
+  """Return introspector projects status"""
+  introspector_bucket = storage.Client().get_bucket(INTROSPECTOR_BUCKET)
+  index_blob = introspector_bucket.blob(INTROSPECTOR_INDEX_JSON)
+  if index_blob.exists():
+    introspector_index = json.loads(index_blob.download_as_string())
+  else:
+    introspector_index = {}
+  logging.info('Loaded introspector status %d', len(introspector_index))
```
oliverchang, line 79: Can you use logs.log instead?

oliverchang (approved):
LGTM with some nits!

src/appengine/handlers/home.py
```diff
@@ -60,9 +68,22 @@ def get_single_fuzz_target_or_none(project, engine_name):
   return fuzz_target_name
 
 
+def get_introspector_index():
+  """Return introspector projects status"""
+  introspector_bucket = storage.Client().get_bucket(INTROSPECTOR_BUCKET)
```
oliverchang, line 73: Missed this: we should be using google_cloud_utils.storage.read_data for this instead. 

(grep for examples for how to call it)

oliverchang (changes requested):

src/appengine/handlers/home.py
Navidem: Done.

src/appengine/handlers/home.py
Navidem: Done.

src/appengine/handlers/home.py
Navidem: Done.

Navidem:
The basic test is failing with `google.auth.exceptions.DefaultCredentialsError: Could not automatically determine credentials. Please set GOOGLE_APPLICATION_CREDENTIALS or explicitly create credentials and re-run the application.`

oliverchang (commented):
LGTM after test is fixed.

We just need to add mock storage.read_data here: https://github.com/google/clusterfuzz/blob/4f8020c4c7ce73c1da0e68f04943af30bb5f0b32/src/clusterfuzz/_internal/tests/appengine/handlers/home_test.py#L32

and have it return a dummy result.

src/clusterfuzz/_internal/tests/appengine/handlers/home_test.py
```diff
@@ -33,8 +33,10 @@ def setUp(self):
         'libs.access.get_access',
         'clusterfuzz._internal.base.external_users.allowed_jobs_for_user',
         'libs.helpers.get_user_email',
+        'clusterfuzz._internal.google_cloud_utils.storage.read_data'
     ])
 
+    self.mock.read_data.return_value = ""
```
oliverchang: nit: please use single quotes.

src/clusterfuzz/_internal/tests/appengine/handlers/home_test.py
```diff
@@ -117,6 +119,8 @@ def test_results_external(self):
                         u'lib',
                     'coverage_job':
                         'libfuzzer_asan_lib',
+                    'has_introspector':
+                        False,
```
oliverchang, line 123: Let's add a testcase such that one of these returns True? (see above)

src/clusterfuzz/_internal/tests/appengine/handlers/home_test.py
```diff
@@ -33,10 +33,12 @@ def setUp(self):
...
         'clusterfuzz._internal.base.external_users.allowed_jobs_for_user',
         'libs.helpers.get_user_email',
-        'clusterfuzz._internal.google_cloud_utils.storage.read_data'
+        'clusterfuzz._internal.google_cloud_utils.storage.read_data',
+        'clusterfuzz._internal.google_cloud_utils.storage.exists'
     ])
 
-    self.mock.read_data.return_value = ""
+    self.mock.read_data.return_value = ''
+    self.mock.exists.return_value = False
```
oliverchang, line 41: Set this to `True` in `tests_results_external` so that for that case we have some tests of the code path were introspector reports does exist. `test_results_internal` can continue to test this with this returning False.
 
We'll also need to set a `read_data` with some fake data too in that test.

Navidem: Done.

oliverchang (approved):
nice!

PR #2634, opened 2022-04-29
Only add to sources.list.d file a single time

dbaker-rh:
The install_deps_bash.sh currently duplicates the lines it adds to /etc/apt/sources.list.d every time it runs.  This patch corrects that.


oliverchang (approved):
thanks!

PR #2636, opened 2022-05-03
Health check responder

Alan32Liu:
1. Add a responser to health checks
2. Include the health check reponser into heartbeat
3. Add testcases for health check responser

### DIFF
--- a/src/clusterfuzz/_internal/system/process_handler.py
+++ b/src/clusterfuzz/_internal/system/process_handler.py
@@ -649,3 +649,19 @@ def terminate_processes_matching_cmd_line(match_strings,
     if any(x in process_path for x in match_strings):
       if not any([x in process_path for x in exclude_strings]):
         terminate_process(process_info['pid'], kill)
+
+
+def processes_are_healthy(expected_processes):
+  """Check if all target processes are all running as expected."""
+  # Note: run_bot.py is expected to go down during source updates
+  #   (which can take a few minutes)
+  # Health checks should be resilient to this
+  # and set a threshold / check interval to account for this.
+  processes_left = expected_processes.copy()
+  for process in psutil.process_iter():
+    for expected_process in processes_left:
+      if any(expected_process in cmdline for cmdline in process.cmdline()):
+        processes_left.remove(expected_process)
+        if not processes_left:
+          return True
+  return False
--- a/src/clusterfuzz/_internal/system/process_handler.py
+++ b/src/clusterfuzz/_internal/system/process_handler.py
@@ -651,17 +651,14 @@ def terminate_processes_matching_cmd_line(match_strings,
         terminate_process(process_info['pid'], kill)
 
 
-def processes_are_healthy(expected_processes):
-  """Check if all target processes are all running as expected."""
-  # Note: run_bot.py is expected to go down during source updates
-  #   (which can take a few minutes)
-  # Health checks should be resilient to this
-  # and set a threshold / check interval to account for this.
-  processes_left = expected_processes.copy()
+def scripts_are_running(expected_scripts):
+  """Check if all target scripts are running as expected."""
+  scripts_left = expected_scripts.copy()
   for process in psutil.process_iter():
-    for expected_process in processes_left:
-      if any(expected_process in cmdline for cmdline in process.cmdline()):
-        processes_left.remove(expected_process)
-        if not processes_left:
+    for expected_script in scripts_left:
+      if any(expected_script == os.path.basename(cmdline)
+             for cmdline in process.cmdline()):
+        scripts_left.remove(expected_script)
+        if not scripts_left:
           return True
   return False
--- a/src/python/bot/startup/health_check_responder.py
+++ b/src/python/bot/startup/health_check_responder.py
@@ -11,35 +11,36 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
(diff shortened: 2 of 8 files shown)


src/python/bot/startup/health_check_responser.py
```diff
@@ -0,0 +1,50 @@
...
+IS_HEALTHY = 200
+NOT_HEALTHY = 500
+
+
+class RequestHandler(BaseHTTPRequestHandler):
+  """Handler for GET request form the health checker"""
+
+  def _is_healthy(self):
+    """Check if all processes are running as expected"""
+    running_processes = os.popen("ps -eo command").read()
```
oliverchang, line 34: Let's use psutil here, which is cross platform. See https://github.com/google/clusterfuzz/blob/26ccc7da38a20a5cf43423b2ad98e58344d50577/src/clusterfuzz/_internal/system/process_handler.py#L639 for examples. 

We can just add a helper function into `process_handler` and use it here.

src/python/bot/startup/heartbeat.py
```diff
@@ -25,6 +25,8 @@
 import sys
 import time
 
+from health_responser import run_server as run_health_check_responser_server
```
oliverchang, line 28: nit: just `run_health_check_server` is shorter and just as descriptive.

src/python/bot/startup/health_check_responser.py
```diff
@@ -0,0 +1,50 @@
...
+    for process in EXPECTED_PROCESSES:
+      if process not in running_processes:
```
oliverchang, line 36: Can you also add a comment here:

```
Note: run_bot.py is expected to go down during source updates (which can take a few minutes) Health checks should be resilient to this and set a threshold / check interval to account for this.
```

oliverchang (commented):
nice!! 

Can you add this line:


```
- iptables -w -A INPUT -p tcp --dport 7123 -j ACCEPT # health check port
```

to after "swapon"  in

https://github.com/google/clusterfuzz/blob/481c78b28fb2d9a401a6f3fe302746a8f6110954/configs/test/gce/linux-init.yaml#L33 and https://github.com/google/clusterfuzz/blob/481c78b28fb2d9a401a6f3fe302746a8f6110954/configs/test/gce/linux-init-ml-with-gpu.yaml.

We'll also need to investigate if any changes are required for our Windows bots in https://github.com/google/clusterfuzz/blob/481c78b28fb2d9a401a6f3fe302746a8f6110954/configs/test/gce/windows-init.ps1. Can do this later.

src/python/bot/startup/heartbeat.py
```diff
@@ -113,6 +115,8 @@ def main():
   except Exception:
     logs.log_error('Failed to beat.')
 
+  run_health_check_responser_server()
```
oliverchang, line 118: nit: s/responser/responder/ makes a bit more sense :)

src/python/bot/startup/heartbeat.py
Alan32Liu: Changed to `run_health_responder_server` as I presume we are the responder to the health checker?

src/clusterfuzz/_internal/system/process_handler.py
```diff
@@ -649,3 +649,19 @@ def terminate_processes_matching_cmd_line(match_strings,
     if any(x in process_path for x in match_strings):
       if not any([x in process_path for x in exclude_strings]):
         terminate_process(process_info['pid'], kill)
+
+
+def processes_are_healthy(expected_processes):
```
oliverchang, line 654: Sorry I meant: 

We can add a helper to `process_handler` that checks if a set of processes are running.

e.g.

```python
def processes_are_running(expected_processes):
 ...
```

and we can call that in `health_check_responder.py`.

src/python/bot/startup/health_check_responder.py
```diff
@@ -0,0 +1,45 @@
...
+IS_HEALTHY = 200
+NOT_HEALTHY = 500
+
+
+class RequestHandler(BaseHTTPRequestHandler):
+  """Handler for GET request form the health checker."""
+
+  def do_GET(self):  # pylint: disable=invalid-name
+    """Handle a GET request."""
+    response_code = IS_HEALTHY if process_handler.processes_are_healthy(
```
oliverchang, line 35: Let's just use http.HTTPStatus.OK and http.HTTPStatus.INTERNAL_SERVER_ERROR directly here.

oliverchang: Also this might be all clearer on separate lines:

```python
if process_handler.processes_are_running(EXPECTED_PROCESSES):
 response_code = ...
else:
  response_code = ..
```

oliverchang (approved):

src/clusterfuzz/_internal/system/process_handler.py
```diff
@@ -649,3 +649,19 @@ def terminate_processes_matching_cmd_line(match_strings,
...
+  """Check if all target processes are all running as expected."""
+  # Note: run_bot.py is expected to go down during source updates
+  #   (which can take a few minutes)
+  # Health checks should be resilient to this
+  # and set a threshold / check interval to account for this.
+  processes_left = expected_processes.copy()
+  for process in psutil.process_iter():
+    for expected_process in processes_left:
+      if any(expected_process in cmdline for cmdline in process.cmdline()):
```
oliverchang, line 663: nit: we should just do a direct comparison (==) to avoid potential false positives.

oliverchang (approved):

oliverchang (approved):

PR #2645, opened 2022-05-11
OSS-Fuzz: Don't disclose bugs if only CCs are vendors.

oliverchang:
Only set the "Reported-" label if at least `primary_contact` or
`auto_ccs` is set.


Alan32Liu (approved):
LGTM

PR #2648, opened 2022-05-12
Add RHEL8, FEDORA36 alternate variants for fuzz bots

dbaker-arch:
Submitting this for consideration.

We have done some initial work to support RHEL (and Fedora) based fuzzing container images to address the base library differences between ubuntu and the other variants.

Attached is a very small patch to add queue definitions, a patch for new pubsub definitions, and sample Dockerfile content for RHEL (using UBI so it's more redistributable) and also F36.

This is working well in our initial tests.  I'd be very happy to discuss options to promote other linux distros in a more flexible manner if this is seen as valuable to the project as a whole.


jonathanmetzman (commented):
Could you explain the usecase more? Is this for testing code that doesn't run on debian distros and only works on RHEL?

dbaker-rh:
Either for code that doesn't run on debian based distros (which, arguably, is going to be few and far between) or more commonly for code that is just problematic for whatever reason to run on them.  Maybe build time complications, library compatibility issues, dependencies on other elements of the base system and things of that nature.

We started down this path as an alternative to the existing container image that (using the xenial build target - https://github.com/google/clusterfuzz/blob/master/docker/base/Dockerfile#L15-L24) takes a set of specific 16.04 libraries and drops them into place in the 20.04 final image.

If we want to fuzz a specific binary in both (e.g.) RHEL7 and RHEL8 varieties, we can have a rhel7 and rhel8 container, each populated with the equivalent base packages, and while the 16.04 + 20.04 library set is a fine stop-gap for that specific need I don't think it scales well enough to cover different distributions.

The other consideration we had was to just ensure every binary was statically linked, but this led to more problems than it avoided by complicating the build step.

oliverchang (commented):
Thanks for contributing this! (And sorry for the delay!).

This is awesome, but given that we don't run these configurations ourselves, it won't be maintainable by us.

Would it be feasible at all to maintain these configs on your end as part of the yaml configs/startup scripts, rather than pushing them upstream here?

Are there any friction points to doing this that we can help resolve to support more general extensibility of platforms?

dbaker-rh:
Absolutely! But, if we can have a general hook for extra queues, and
perhaps a "examples" directory for the Dockerfiles for a public reference
we can easily create a building block for others to use without reinventing
the wheel.

I'm out of the office for a while but will recreate my PR when I'm back to
try to capture this as a non maintained option, along with some extra docs.

Friction points are mainly that every distro changes over time - keeping
the buk on ubuntu (maintained by you) and just spinning the -pre- fuzz only
bots on different distros can help a lot, I think.

PR #2650, opened 2022-05-13
[fuchsia] Ignore more Rust stack frames.

eepeep:
Panicking Rust fuzzers on Fuchsia present a different set of
intermediate stack frames that we wish to ignore.

Reasons for these differences include:
  - The use of `panic=abort` instead of the special aborting panic hook
    installed by libfuzzer-sys.
  - On panic, a full backtrace is dumped by `std::sys_common::backtrace` 
    in addition to the one subsequently generated by libFuzzer/ASAN.


eepeep:
cc @oliverchang 

Note that it obviously doesn't make sense to have two backtraces like this, and I expect we will disable the first one in favor of keeping only the ASAN one. This is not an instant fix to make on our end, however, due to the fact that backtraces are hard-coded to on for Fuchsia (https://github.com/rust-lang/rust/blob/b3800860e123443ffada615538926beed6bc4f85/library/std/src/panic.rs#L300). 

I figured it would make sense to just cover both cases for now, but if you'd prefer we can hold off on merging until we clean up the double-backtrace situation on our end.

oliverchang (approved):
LGTM. thanks for explaining the reason there are two different stacktraces!

oliverchang:
Could you please fix the lint issue pointed out in https://github.com/google/clusterfuzz/runs/6417856514?check_suite_focus=true#step:5:1720 (it's not caused by our PR but it's still getting picked up).

eepeep:
Whoops, looks like I had to re-run `install_deps` to update my pylint config. Done.

PR #2654, opened 2022-05-17
Set GitHub attributes in Testcase during backfilling

Alan32Liu:
Bug fix:
  Upload GitHub info to cloud storage during backfiling.


oliverchang (approved):
Thanks! FYI I fixed up the title of this PR to be more accurate. We also use Datastore (rather than cloud storage) for our DB.

PR #2657, opened 2022-05-17
GitHub issue metadata backfiller

Alan32Liu:


src/appengine/libs/issue_management/oss_fuzz_github.py
```diff
@@ -212,3 +218,20 @@ def close_issue(testcase):
...
            f'in GitHub repository {testcase.github_repo_id}.')
+
+
+def get_my_issues():
+  """Get all issues filed by oss-fuzz-robot."""
+  access_token = _get_access_token()
+  return access_token.search_issues('author:oss-fuzz-robot')
+
+
+def get_bug_information(issue):
```
oliverchang, line 229: This is pretty specific to the butler script and unlikely to be used elsewhere. Let's move it there.

src/local/butler/scripts/backtracker.py
```diff
@@ -0,0 +1,90 @@
...
+  return False
+
+
+def execute(args):
+  """Backtrack GitHub issues filed in the past,
+    update their information in gcloud, and close them when necessary."""
+  for issue in oss_fuzz_github.get_my_issues():
+    # Track testcase.
+    bug_information = oss_fuzz_github.get_bug_information(issue)
+    if bug_information is None:
```
oliverchang, line 65: nit: do `if not bug_information`. 

See https://google.github.io/styleguide/pyguide.html#214-truefalse-evaluations

src/local/butler/scripts/backtracker.py
```diff
@@ -0,0 +1,90 @@
...
+    print(f'Updating testcase (bug information: {testcase.bug_information}):\n'
+          f'  Issue number  {issue.number}\n'
+          f'  Repository ID {issue.repository.id}\n')
+    if args.non_dry_run:
+      oss_fuzz_github.update_testcase_properties(testcase, issue.repository,
+                                                 issue)
+      testcase.put()
+
+    # Backclose issues.
+    if testcase.open:
```
oliverchang, line 84: We should do:

```python
# do this once
issue_tracker = issue_tracker_utils.get_issue_tracker('oss-fuzz')

... 
monorail_issue = issue_tracker.get_original_issue(t.bug_information) 
if monorail_issue.is_open:
  continue
```

instead of this. A testcase can be closed here but the issue might not be closed.

oliverchang, line 84: We should move this check above line 73 (i.e. _testcase_information_verified).

This way, we also auto close any issues that should've been closed even if the properties are there. This can happen because of
- The window where our account was flagged
- If this script gets interrupted.

oliverchang (changes requested):
nice, thanks!! some comments.

src/local/butler/scripts/github_metadata_backfiller.py
```diff
@@ -0,0 +1,102 @@
...
+    testcase = _get_testcase(bug_information)
+
+    # Backclose issues.
+    monorail_issue = issue_tracker.get_original_issue(testcase.bug_information)
+    if not monorail_issue.is_open:
+      print(f'Closing testcase (bug information: {testcase.bug_information}):\n'
+            f'  Issue number  {issue.number}\n'
+            f'  Repository ID {issue.repository.id}\n')
+      if args.non_dry_run:
+        oss_fuzz_github.close_issue(testcase)
```
oliverchang, line 91: Looking at the implementation of `close_issue`, it looks like we early exit if there's no issue info set:

```
  if not _issue_recorded(testcase):                                                                                           
    return  
```

It might be better after all to have your old order after all, and instead of continuing when `_testcase_information_verified`, we do the testcase.put() if necessary.

i.e.

```python
    if not _testcase_information_verified(testcase, issue):
      # Instead of continuing
      print(f'Updating testcase (bug information: {testcase.bug_information}):\n'
            f'  Issue number  {issue.number}\n'
            f'  Repository ID {issue.repository.id}\n')
      if args.non_dry_run:
        oss_fuzz_github.update_testcase_properties(testcase, issue.repository,
        testcase.put()

    monorail_issue = issue_tracker.get_original_issue(testcase.bug_information)
    if not monorail_issue.is_open:
       ...
```

Also, can we check if the GitHub issue is already closed here, to avoid `oss_fuzz_github.close_issue` retrieving the issue unnecessarily if we already know we don't need to do antyhing?

src/local/butler/scripts/github_metadata_backfiller.py
```diff
@@ -0,0 +1,105 @@
...
+          f'Updating testcase (bug information: {testcase.bug_information}):\n'
+          f'  Issue number  {issue.number}\n'
+          f'  Repository ID {issue.repository.id}\n')
+      if args.non_dry_run:
+        oss_fuzz_github.update_testcase_properties(testcase, issue.repository,
+                                                   issue)
+        testcase.put()
+
+    # Backclose issues.
+    if issue.state == 'closed':
```
Alan32Liu, line 96: We can also filter out closed issues when [querying GitHub](https://github.com/google/clusterfuzz/blob/52737a91285799406726f8bfc64f0d5695b50519/src/appengine/libs/issue_management/oss_fuzz_github.py#L226), each method have its pros and cons.
Doing it here is safer. Although we cannot close an issue if we don't have its info in the cloud, the repo owners might close it manually. I suppose this is trivial because we don't really need an issue's info once the issue is closed.

Doing it in the search query can lower the chance of reaching our [rate limit](https://docs.github.com/en/rest/search#rate-limit). When a query reaches the limit, it returns nothing and this script won't work.

I choose to do it here because:
1. We might need the github info of those issues later.
2. My local experiments never reached our rate-limit.

oliverchang (approved):
thanks!!

PR #2670, opened 2022-05-24
add missing GCP API: cloudscheduler.googleapis.com

bookholt:
Adds Cloud Scheduler to list of required GCP APIs. Installation now fails without enabling this API.


oliverchang (approved):
thanks a lot!

PR #2685, opened 2022-05-28
build_manager: Add logging

AdamKorcz:
The initial idea of this commit was to log an error instead of throwing a `list index out of range` in case `len(build_urls)==0`. I then added a few more log statements to track the bucket paths earlier in the build manager.

### DIFF
--- a/src/clusterfuzz/_internal/build_management/build_manager.py
+++ b/src/clusterfuzz/_internal/build_management/build_manager.py
@@ -1263,6 +1263,11 @@ def _get_latest_revision(bucket_paths):
 
     build_urls.append(BuildUrls(bucket_path=bucket_path, urls_list=urls_list))
 
+  if len(build_urls) == 0:
+    logs.log_error(
+        'Attempted to get latest revision, but no build urls were found.')
+    return None
+
   main_build_urls = build_urls[0]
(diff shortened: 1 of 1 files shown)


AdamKorcz:
It looks like the linting errors are unrelated to this PR. I can fix those in a separate PR.

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -1263,6 +1263,11 @@ def _get_latest_revision(bucket_paths):
 
     build_urls.append(BuildUrls(bucket_path=bucket_path, urls_list=urls_list))
 
+  if len(build_urls) == 0:
+    logs.log_error("""Attempted to get latest revision,
```
oliverchang, line 1267: nit: Please use single quotes for strings. 

You can rely on implicit string continuation to break this over multiple lines.

(I haven't checked but it's possible you don't even need to break this over multiple lines if you just put the string
on the next line).

i.e.

````python
logs.log_error(
    'Attempted to get lateset revision,'
   'but not build urls were found.')
```

oliverchang (commented):
thanks!!

oliverchang (approved):

PR #2687, opened 2022-05-30
Don't set "-a public-read" argument in deployment.

### LINKED ISSUE #2683: upgrade process fails due to upload of clusterfuzz-source.manifest file
So I tired to deploy the latest version for 2.5.7 and I kept getting this error that it could not upload the clusterfuzz-source.manifest file.  The command look like this.

gsutil cp -a public-read src/appengine/resources/clusterfuzz-source.manifest gs://deployment.clusterfuzz9-dev.appspot.com/clusterfuzz-source.manifest.3

which generated the following error.

BadRequestException: 400 Cannot insert legacy ACL for an object when uniform bucket-level access is enabled. 

I fixed this by removing the "-a public-read" from the python script in the following file.

clusterfuzz/src/local/butler/deploy.py

After removing that -a ACL setting in the butler deploy script that seem to fix this issue and was able to update my clusterfuzz instance.  So we have all our GCS buckets with uniform ACL access.  Is this setup changed since i first deployed this clusterfuzz?  Does the manifest file need to be a public-read ACL?  Is it a problem that that file is not publicly viewable?  Honestly I would prefer not to set any of the buckets with non-uniform ACL and especially not expose any buckets or objects to be publicly viewable.

oliverchang:
Fixes #2683

### DIFF
--- a/src/clusterfuzz/_internal/tests/core/local/butler/deploy_test.py
+++ b/src/clusterfuzz/_internal/tests/core/local/butler/deploy_test.py
@@ -184,7 +184,7 @@ class DeployTest(fake_filesystem_unittest.TestCase):
                   'mac.zip'),
         mock.call('gsutil cp /linux.zip gs://test-deployment-bucket/'
                   'linux.zip'),
-        mock.call('gsutil cp -a public-read src/appengine/resources/'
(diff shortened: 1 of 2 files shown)


Alan32Liu (approved):
LGTM

PR #2695, opened 2022-06-03
Strip reserved characters when generating JQL for jira search

quanyangsc:
Jira JQL doesn't allow for certain reserved special characters and having them can cause errors, this change solves that by replacing the characters with white space.


oliverchang (approved):

oliverchang:
Thanks! This looks good -- could you please just format the code per the lint issue?

PR #2701, opened 2022-06-13
Fix incorrect logging call.

eepeep:
PR #2685 accidentally added a logging statement that would cause
"TypeError: level must be an integer" to be thrown when reached. This
fixes the syntax to remove the error.


oliverchang (approved):

AdamKorcz:
@eepeep Thank you for fixing that!

PR #2719, opened 2022-07-05
Fix stacktrace parsing for command injection.

oliverchang:
- Don't replace the crash type with the sanitizer signal crash type.
- Explicitly set 'Command injection' as a security bug.


Alan32Liu (approved):
LGTM!

PR #2702, opened 2022-06-16
Execsan integration 1

Alan32Liu:
Part 1 of integrating `execSan` into `clusterfuzz`:
1. Refactor `UnshareProcessRunner` into `ModifierProcessRunner`
2. Prepend fuzzing commands with the sanitiser.

I was using the name `injSan` instead of `execSan` as we may need to add other injection detection later.

TODOs:
1. Build sanitiser in `OSS-FUZZ`
2. Capture error messages from the sanitiser
3. Roll out gradually.

PR #2720, opened 2022-07-06
experimental: Add extra sanitizers strategy.

oliverchang:
This will be turned on in 10% of all libFuzzer runs.

Note that this flag does not propagate to any testcases yet, so any
found bugs will be marked as unreproducible for now.


Alan32Liu (approved):
LGTM

jonathanmetzman:
Is there something missing here? although the environment var is set, are wrapping the fuzzing process with the sanitizer already? Where is this going to be done.

oliverchang:
> Is there something missing here? although the environment var is set, are wrapping the fuzzing process with the sanitizer already? Where is this going to be done.

This was done by @Alan32Liu already in https://github.com/google/clusterfuzz/pull/2702.

See https://oss-fuzz.com/testcase-detail/4824786613698560

PR #2724, opened 2022-07-11
Properly disable LSan when extra sanitizers are used.

oliverchang:
Extra sanitizers are decided later on after the sanitizer options are
initially set.


Alan32Liu (approved):
LGTM!

oliverchang:
> lgtm. I think this WAI but I don't really understand why LSAN and execsan are incompatible

ptrace doesn't work with LSan. This might make the case for using something like eBPF instead etc.

PR #2730, opened 2022-07-14
Add the functionality to recognize flag contradictions for trials

almuthanna-athamina:
This PR allows adding "contradicts" property as a list to Clusterfuzz trials config JSON file which later would be parsed and not allow flags to be added that contradict already added flags.


almuthanna-athamina:
It seems that I have made a mistake while rebasing, I have created a new PR here https://github.com/google/clusterfuzz/pull/2731

PR #2731, opened 2022-07-14
Add the functionality to recognize flag contradictions for trials

almuthanna-athamina:
This PR allows adding "contradicts" property as a list to Clusterfuzz trials config JSON file which later would be parsed and not allow flags to be added that contradict already added flags.


src/clusterfuzz/_internal/bot/tasks/trials.py
```diff
@@ -43,9 +43,10 @@ def __init__(self):
     for extension in extensions_to_strip:
       app_name = utils.strip_from_right(app_name, extension)
 
-    self.trials = {}
     for trial in data_types.Trial.query(data_types.Trial.app_name == app_name):
-      self.trials[trial.app_args] = trial.probability
+      self.trials[trial.app_args] = {'probability': trial.probability}
+      if hasattr(trial, 'contradicts'):
```
oliverchang, line 48: Rather than do this, can you modify https://github.com/google/clusterfuzz/blob/770bee181da6f309c05b424053e1b7bb4b4b340b/src/clusterfuzz/_internal/datastore/data_types.py#L1382 to add this property to the Model?

src/clusterfuzz/_internal/bot/tasks/trials.py
```diff
@@ -61,17 +62,24 @@ def __init__(self):
...
   def setup_additional_args_for_app(self):
     """Select additional args for the specified app at random."""
-    trial_args = [
-        app_args for app_args, probability in self.trials.items()
-        if random.random() < probability
-    ]
+    trial_args = []
+    contradicts = []
+    for app_args, flag_data in self.trials.items():
+      if random.random(
```
oliverchang, line 77: Can you add reformat this so this is ealier to read by adding ( ) surrounding the entire if condition.

i.e. something like

```python
if (random.random() < flag_data[...] and
    app_args not in contradicts):
  ...

```

src/clusterfuzz/_internal/bot/tasks/trials.py
```diff
@@ -61,17 +62,24 @@ def __init__(self):
...
+      ) < flag_data['probability'] and app_args not in contradicts:
+        trial_args.append(app_args)
+        if 'contradicts' in flag_data:
+          for contradiction in flag_data['contradicts']:
```
oliverchang, line 81: just do

```python
for contradiction in flag_data.get('contradicts', []):
   ...
```

(no need for the if on line 80).

src/clusterfuzz/_internal/bot/tasks/trials.py
```diff
@@ -61,17 +62,24 @@ def __init__(self):
       for config in trials_config:
         if config['app_name'] != app_name:
           continue
-        self.trials[config['app_args']] = config['probability']
+        self.trials[config['app_args']] = {'probability': config['probability']}
```
oliverchang, line 65: You can just simplify this with

```python
self.trials[config['app_args']] = {
  'probability': config['probability'],
  'contradicts': config.get('contradicts', [])
}
```

src/clusterfuzz/_internal/bot/tasks/trials.py
```diff
@@ -43,9 +43,10 @@ def __init__(self):
     for extension in extensions_to_strip:
       app_name = utils.strip_from_right(app_name, extension)
 
-    self.trials = {}
     for trial in data_types.Trial.query(data_types.Trial.app_name == app_name):
-      self.trials[trial.app_args] = trial.probability
+      self.trials[trial.app_args] = {'probability': trial.probability}
```
oliverchang, line 47: Can you created a namedtuple or class for this containing the two properties (probability and contradicts) ?

src/clusterfuzz/_internal/bot/tasks/trials.py
mi-ac: We can do this, but note that V8 isn't going to use the in-DB trials anymore. They are legacy from our viewpoint and only have few chrome entries. Not sure who maintains them. From our viewpoint we could keep any new features out of the model, but the code might indeed be cleaner by supporting it without much cost...

mi-ac: Note also, if we don't add it to the model, these two lines are dead code.

src/clusterfuzz/_internal/bot/tasks/trials.py
mi-ac: +1, this is actually required as in the previous code the 'contradicts' field stays uninitialized.

src/clusterfuzz/_internal/tests/core/bot/tasks/trials_test.py
```diff
@@ -206,3 +206,17 @@ def test_corrupted_config_file_is_ignored(self):
...
     config_file_content = '[{"app_args": "--a1", "app_name": "app_1", "probability": 0.8]'
     self.source_side_test(config_file_content, 0.7, 'app_1', '-x', None)
+
+  def test_trial_ignored_if_there_is_contradiction(self):
+    """Ensure that a flag is not added if it contradicts an already added flag."""
+    config_file_content = [{
+        "app_args": "--c4",
+        "app_name": "app_4",
+        "probability": 1.0,
+        "contradicts": ["--c5"]
```
mi-ac, line 216: See my comment above about symmetrical contradictions. This test case is an example for an asymmetric contradiction. If you turn it around, it won't get detected I think. Maybe add another test where they are turned around and enhance the logic as described in my other comment?

src/clusterfuzz/_internal/bot/tasks/trials.py
mi-ac: Re on simple: I wouldn't want to add another table to get a list of flags. It'd do to create a string property and just space separate flags. We sort-of have the convention anyways to not have spaces within one flag.

almuthanna-athamina: done

src/clusterfuzz/_internal/bot/tasks/trials.py
almuthanna-athamina: done

src/clusterfuzz/_internal/bot/tasks/trials.py
almuthanna-athamina: done, created class called AppArgs

src/clusterfuzz/_internal/tests/core/bot/tasks/trials_test.py
almuthanna-athamina: done, added another test case

src/clusterfuzz/_internal/bot/tasks/trials.py
```diff
@@ -61,17 +62,24 @@ def __init__(self):
...
 
   def setup_additional_args_for_app(self):
     """Select additional args for the specified app at random."""
-    trial_args = [
-        app_args for app_args, probability in self.trials.items()
-        if random.random() < probability
-    ]
+    trial_args = []
+    contradicts = []
+    for app_args, flag_data in self.trials.items():
```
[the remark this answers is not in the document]
almuthanna-athamina: done, now we shuffle trials when we loop through them, and since we are checking asymmetrical contradictions too the tests should not be brittle

src/clusterfuzz/_internal/bot/tasks/trials.py
```diff
@@ -61,17 +62,24 @@ def __init__(self):
...
+      if random.random(
+      ) < flag_data['probability'] and app_args not in contradicts:
```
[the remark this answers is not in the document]
almuthanna-athamina: done, we use a set to collect contradictions

src/clusterfuzz/_internal/bot/tasks/trials.py
```diff
@@ -25,11 +25,25 @@
...
 
 
+class AppArgs:
+
+  def __init__(self, probability, contradicts):
+    self.probability = probability
+    self.contradicts = contradicts
+
+
+class AppArgs:
```
mi-ac, line 35: Twice?

src/clusterfuzz/_internal/bot/tasks/trials.py
```diff
@@ -61,17 +75,25 @@ def __init__(self):
...
     """Select additional args for the specified app at random."""
-    trial_args = [
-        app_args for app_args, probability in self.trials.items()
-        if random.random() < probability
-    ]
+    trial_args = []
+    contradicts = set()
+
+    for app_args, flag_data in sorted(
+        self.trials.items(), key=lambda x: random.random()):
```
mi-ac, line 90: Not sure I follow here. Do you want to do random sorting? I thought the key should be stable...

How about:
random.sample(self.trials.items(), k=len(self.trials))

Note from the two ideas I had about how to resolve the probability bias, this is probably the more costly.

src/clusterfuzz/_internal/bot/tasks/trials.py
```diff
@@ -61,17 +75,25 @@ def __init__(self):
...
+      if (random.random() < flag_data.probability and
+          app_args not in contradicts and
+          not any(flag in flag_data.contradicts for flag in trial_args)):
```
mi-ac, line 93: Not sure if I'm maybe micro-optimizing here, but this check makes the algorithm quadratic (for cases where above expressions are true). I'm not sure if Python will build an optimization for the case where flag_data.contradicts is empty.

A simple guard for this would be:
(not flag_data.contradicts or the rest of your line)

Since this gets a bit convoluted, maybe extract it into a function?

src/clusterfuzz/_internal/bot/tasks/trials.py
almuthanna-athamina: this was a rebasing mistake, now it is fixed

src/clusterfuzz/_internal/bot/tasks/trials.py
```diff
@@ -25,11 +25,25 @@
...
+class AppArgs:
+
+  def __init__(self, probability, contradicts):
+    self.probability = probability
+    self.contradicts = contradicts
+
+
+class AppArgs:
+
+  def __init__(self, probability, contradicts=[]):
```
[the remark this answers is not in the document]
almuthanna-athamina: done

src/clusterfuzz/_internal/bot/tasks/trials.py
almuthanna-athamina: done, we shuffle the keys now

src/clusterfuzz/_internal/bot/tasks/trials.py
almuthanna-athamina: done

src/clusterfuzz/_internal/datastore/data_types.py
```diff
@@ -1390,6 +1390,8 @@ class Trial(Model):
   # Additional arguments to apply if selected.
   app_args = ndb.TextProperty()
 
+  # flags that contradict the app args (separated by spaces).
```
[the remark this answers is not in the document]
almuthanna-athamina: done

src/clusterfuzz/_internal/bot/tasks/trials.py
```diff
@@ -43,9 +57,9 @@ def __init__(self):
     for extension in extensions_to_strip:
       app_name = utils.strip_from_right(app_name, extension)
 
-    self.trials = {}
     for trial in data_types.Trial.query(data_types.Trial.app_name == app_name):
-      self.trials[trial.app_args] = trial.probability
+      self.trials[trial.app_args] = AppArgs(trial.probability,
+                                            trial.contradicts.split())
```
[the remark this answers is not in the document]
almuthanna-athamina: done, we now assume the contradicts value has one flag

src/clusterfuzz/_internal/bot/tasks/trials.py
```diff
@@ -61,17 +68,36 @@ def __init__(self):
...
       logs.log_warn('Unable to parse config file: %s' % str(e))
       return
 
-  def setup_additional_args_for_app(self):
+  def doesnt_asymmetrically_contradicts(self, app_args, existing_contradictions,
+                                        trial_args):
+    return (app_args not in existing_contradictions and
+            (not self.trials[app_args].contradicts or
+             not any(flag in self.trials[app_args].contradicts
+                     for flag in trial_args)))
```
mi-ac, line 82: I think it'd be more readable to only extract lines 80-82 in their own function and give it a simpler name, e.g. new_arg_contradicts, so that you can write
not new_arg_contradicts(new_args, existing_args)

Another way to make this more readable is inverting the logic in each expression below and use guard clauses:
if (reason one to not select this flag):
 continue
if (reason two to not select this flag):
 continue
...

If you structure it like that, you probably don't need to extract a function. Each if-statement could get a comment that helps understanding.

src/clusterfuzz/_internal/bot/tasks/trials.py
```diff
@@ -43,9 +50,9 @@ def __init__(self):
     for extension in extensions_to_strip:
       app_name = utils.strip_from_right(app_name, extension)
 
-    self.trials = {}
     for trial in data_types.Trial.query(data_types.Trial.app_name == app_name):
-      self.trials[trial.app_args] = trial.probability
+      self.trials[trial.app_args] = AppArgs(trial.probability,
+                                            [trial.contradicts])
```
mi-ac, line 55: Where did the split go? We'd still want to support multiple possible contradictions that could be listed in DB by space separation... that's also what the comment at the data type says. split on empty string returns an empty list, which would fit.

mi-ac (commented):
I think the new logic that we want to use works now. Some comments to also make DB work and some readability nit.

src/clusterfuzz/_internal/datastore/data_types.py
```diff
@@ -1390,6 +1390,9 @@ class Trial(Model):
   # Additional arguments to apply if selected.
   app_args = ndb.TextProperty()
 
+  # Flags that contradict the app args (separated by spaces).
+  contradicts = ndb.TextProperty(default='')
```
oliverchang, line 1394: Should we make this a `repeated=True` property instead?

src/clusterfuzz/_internal/bot/tasks/trials.py
oliverchang: (See comment below)

How about we make this a repeated property so it's automatically a list (and stored as such) instead?

src/clusterfuzz/_internal/bot/tasks/trials.py
```diff
@@ -61,17 +68,36 @@ def __init__(self):
...
-        self.trials[config['app_args']] = config['probability']
+        self.trials[config['app_args']] = AppArgs(config['probability'],
+                                                  config.get('contradicts', []))
     except Exception as e:
       logs.log_warn('Unable to parse config file: %s' % str(e))
       return
 
-  def setup_additional_args_for_app(self):
+  def doesnt_asymmetrically_contradicts(self, app_args, existing_contradictions,
+                                        trial_args):
```
oliverchang, line 78: Please add a docstring for this function.

src/clusterfuzz/_internal/bot/tasks/trials.py
almuthanna-athamina: done, I have inverted the logic and moved it to the loop with comments

almuthanna-athamina: I have removed the `not self.trials[app_args].contradicts` condition since I do not think it is necessary anymore

src/clusterfuzz/_internal/bot/tasks/trials.py
almuthanna-athamina: done

src/clusterfuzz/_internal/datastore/data_types.py
almuthanna-athamina: done

src/clusterfuzz/_internal/bot/tasks/trials.py
almuthanna-athamina: I have moved the conditions into the loop with comments instead

src/clusterfuzz/_internal/bot/tasks/trials.py
```diff
@@ -92,9 +85,14 @@ def setup_additional_args_for_app(self, shuffle=True):
...
-      if (random.random() < self.trials[app_args].probability and
-          self.doesnt_asymmetrically_contradicts(app_args, contradicts,
-                                                 trial_args)):
+      if random.random() < self.trials[app_args].probability:
+        # Check if the flag is contradicted by an already added flag
+        if app_args in contradicts:
+          continue
+        # Check if the flag contradicts an already added flag
+        if any(
+            flag in self.trials[app_args].contradicts for flag in trial_args):
```
mi-ac, line 94: The optimization you added before vanished now?

if self.trials[app_args].contradicts and any(...

src/clusterfuzz/_internal/bot/tasks/trials.py
```diff
@@ -92,9 +85,14 @@ def setup_additional_args_for_app(self, shuffle=True):
...
+          continue
         trial_args.append(app_args)
         for contradiction in self.trials[app_args].contradicts:
           contradicts.add(contradiction)
```
mi-ac, line 98: contradicts.update(self.trials[app_args].contradicts)

src/clusterfuzz/_internal/bot/tasks/trials.py
almuthanna-athamina: done

src/clusterfuzz/_internal/bot/tasks/trials.py
almuthanna-athamina: done

src/clusterfuzz/_internal/datastore/data_types.py
```diff
@@ -1390,6 +1390,9 @@ class Trial(Model):
   # Additional arguments to apply if selected.
   app_args = ndb.TextProperty()
 
+  # Flags that contradict the app args (separated by spaces).
```
mi-ac, line 1393: nit: drop "(separated by spaces)" since now it's repeated

mi-ac (approved):
LGTM with one nit left.

Oliver, please take another look.

src/clusterfuzz/_internal/datastore/data_types.py
almuthanna-athamina: done

mi-ac:
Guess auto-merge doesn't work here? Oliver could you merge this manually?

Issue #2728, opened 2022-07-14
project_setup: Support a "none" sanitizer. 


oliverchang:

jonathanmetzman:
Why do we want this btw? How will we get stacktraces?

jonathanmetzman:
(I guess libfuzzer has a builtin signal handler that can do this)

oliverchang:
Yeah, this is for e.g. Jazzer fuzzing without a sanitizer.

PR #2747, opened 2022-08-02
Add support for none sanitizer

Navidem:
Fixes https://github.com/google/clusterfuzz/issues/2728


src/appengine/handlers/cron/project_setup.py
```diff
@@ -137,15 +137,21 @@ def job_name(self, project_name, config_suffix):
 GFT_UBSAN_JOB = JobInfo('googlefuzztest_ubsan_', 'googlefuzztest', 'undefined',
                         ['googlefuzztest', 'engine_ubsan'])
 
+LIBFUZZER_NONE_JOB = JobInfo("libfuzzer_none_", "libfuzzer", "none", [])
```
oliverchang, line 140: nit: let's `s/none/nosanitizer/` to make it clearer what `none` means to users.

Navidem: Done.

oliverchang (approved):
nice!

PR #2751, opened 2022-08-04
Skip over V8 API error-reporting frames

backes:
The bug is in the caller that used the API in a wrong way (in the given
example by not checking if a MaybeLocal actually contains a value).


backes:
@oliverchang PTAL; context: https://crbug.com/1348818

oliverchang (approved):

PR #2753, opened 2022-08-05
Report V8 API errors correctly

backes:
Previously they were reported as 'breakpoint', and the actual error
message was missing.
We now handle them separately, similar to chrome's or v8's check
failures.

This is a follow-up to https://github.com/google/clusterfuzz/pull/2751.


backes:
@oliverchang this is a follow-up to https://github.com/google/clusterfuzz/pull/2751.
Is it OK to introduce a new crash type, or would that need more changes in other part? We could also reuse one of the existing types.

oliverchang (approved):
LGTM, thanks!

Adding a new crash type this way is completely OK.

backes:
Thanks, Oliver!

Issue #2750, opened 2022-08-03
Add support for parsing arbitrary file open


oliverchang:
Implemented in https://github.com/google/oss-fuzz/pull/8009

catenacyber:
So, waiting for this to see bugs pop ?

oliverchang:
@Alan32Liu mind doing this when you get the chance?

Alan32Liu:
Happy to!
I will work on it this afternoon.

catenacyber:
@Alan32Liu should have bugs popped by now ? I did not spot any in my projects

Alan32Liu:
> @Alan32Liu should have bugs popped by now? I did not spot any in my projects

Yep, I just checked and can see there are many.
But we might need to look deeper into them to filter out false positives.
I will try to get back to you later this week if we find anything interesting : )

catenacyber:
Ok, I see none with my creds on https://bugs.chromium.org/p/oss-fuzz/issues/list?q=Arbitrary%20file%20open&can=2
Let me know what is next

PR #2759, opened 2022-08-11
Support parsing arbitrary file open

Alan32Liu:
Address #2750 for https://github.com/google/oss-fuzz/pull/8009:
* Parse the bug message of AFO
* Test parsing AFO


oliverchang (commented):
thanks Dongge! Could you please also add this new crash type to https://github.com/google/clusterfuzz/blob/e922a5eeff598e4a789f568679d5b6458c46850a/src/clusterfuzz/_internal/crash_analysis/crash_analyzer.py#L105

oliverchang (approved):

PR #2762, opened 2022-08-15
Init tool_options to fix NOSANITIZER runtime error

Navidem:


src/clusterfuzz/_internal/system/environment.py
```diff
@@ -862,7 +862,7 @@ def reset_current_memory_tool_options(redzone_size=0,
   bot_platform = platform()
 
   # Default options for memory debuggin tool used.
-  if tool_name in ['ASAN', 'HWASAN']:
+  if tool_name in ['ASAN', 'HWASAN', 'NOSANITIZER']:
```
oliverchang, line 865: Could we just separate this out into its own branch and set `tool_options` to be `{}` instead?

Or alternatively, we just set `tool_options = {}` before line 865 and then we don't need the explicit branch.

oliverchang (approved):

Navidem:
weird that CI passes locally.

oliverchang:
Yes the test failure seems to only happen on GCB. I'll merge this for now and debug.

PR #2771, opened 2022-09-01
Create FuzzTarget from Testcase when it does not exist.

### LINKED ISSUE #2769: Handle reproducing testcases where the FuzzTarget has been garbage collected
This: https://github.com/google/clusterfuzz/blob/63a987ca1f37eaae31b0aa4ba9cab55f4d35e588/src/clusterfuzz/_internal/datastore/data_types.py#L720

will lead to an exception otherwise. We should synthesize a FuzzTarget object here if it does not exist so we can continue.

oliverchang:
This is an edge case that can happen when a fuzz target hasn't run for a
very long time, but some other non-preemptible task is still getting
scheduled to run.

Fixes #2769


Alan32Liu (approved):
Thanks!

PR #2787, opened 2022-09-16
Add support for ClusterFuzz-Good.

### LINKED ISSUE #2786: cleanup: Support a "ClusterFuzz-Good" label
Currently if CF attempted to close a bug in the past, and we re-opened it, we won't do it again. This is guarded by this logic here:

https://github.com/google/clusterfuzz/blob/7dc72bc4feceff5effb6edf419ed4f246657db93/src/appengine/handlers/cron/cleanup.py#L447

In some rare cases we need to re-enable automatic testcase closing even if we had previously undid this in the past. We need to support a way to override this.

oliverchang:
Fixes #2786.


Alan32Liu (approved):
LGTM

PR #2791, opened 2022-09-20
Use variants similarity to group testcases

Navidem:


src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -1535,6 +1535,14 @@ def get_testcase_variant(testcase_id, job_type):
   return variant
 
 
+def get_all_testcase_variants(testcase_id):
+  """Get all testcase variant entities based on testcase id."""
+  variants = data_types.TestcaseVariant.query(
+      data_types.TestcaseVariant.testcase_id == testcase_id).get()
```
oliverchang, line 1541: Don't call get(). This only gets a single entity from the entire query. 

Just return the query as is. We can iterate over this query to get all results.

Navidem: Done.

src/appengine/handlers/cron/grouper.py
```diff
@@ -86,6 +86,50 @@ def _get_new_group_id():
...
+
+      # Rule: If both testcase have the same group id, then no work to do.
+      if testcase_1.group_id == testcase_2.group_id and testcase_1.group_id:
+        continue
+
+      # Rule: Check both testcase are under the same project.
+      if testcase_1.project_name != testcase_2.project_name:
+        continue
+
+      # Rule: Group testcase with similar variants
```
jonathanmetzman, line 115: nit: end with period.

src/appengine/handlers/cron/grouper.py
```diff
@@ -86,6 +86,50 @@ def _get_new_group_id():
...
+  """Checks for the testcase variants equality"""
+  if (variant1.crash_type == variant2.crash_type and
+      variant1.crash_state == variant2.crash_state and
+      variant1.security_flag == variant2.security_flag):
+    return True
+  return False
+
+
+def _group_testcases_based_on_variants(testcase_map):
+  """Group testcases that are associated based on variant analysis"""
```
jonathanmetzman, line 99: nit: end with period.

src/appengine/handlers/cron/grouper.py
```diff
@@ -86,6 +86,50 @@ def _get_new_group_id():
...
+  for testcase_1_id, testcase_1 in six.iteritems(testcase_map):
```
jonathanmetzman, line 100: can you just do `testcase_map.items()`? I feel like we don't need `six` anymore, we don't care about python2.

src/appengine/handlers/cron/grouper.py
```diff
@@ -86,6 +86,50 @@ def _get_new_group_id():
   return new_group.key.id()
 
 
+def is_same_variant(variant1, variant2):
+  """Checks for the testcase variants equality"""
+  if (variant1.crash_type == variant2.crash_type and
+      variant1.crash_state == variant2.crash_state and
+      variant1.security_flag == variant2.security_flag):
+    return True
```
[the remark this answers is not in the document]
Navidem: There is no `else return False` :)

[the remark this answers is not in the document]
jonathanmetzman: It's equivalent to else return False right?
You could rewrite this function as ```return (variant1.crash_type == variant2.crash_type and
      variant1.crash_state == variant2.crash_state and
      variant1.security_flag == variant2.security_flag)``` right?

src/appengine/handlers/cron/grouper.py
Navidem: That makes sense, but I think we may leave it to another PR for a complete refactor as `six` is very commonly used in this code.

Navidem: Or I can go ahead avoid using it from my code. Both works for me.

jonathanmetzman: I would lean towards not using six more but it's reasonable to do so if the rest of the file does. I agree that the rest of the usages don't need to be changed in this commit.

src/appengine/handlers/cron/grouper.py
```diff
@@ -86,6 +86,50 @@ def _get_new_group_id():
...
+      for candidate1 in testcase_1_variants + [testcase_1]:
+        for candidate2 in testcase_2_variants + [testcase_2]:
+          if is_same_variant(candidate1, candidate2):
+            has_similar_variants = True
+            break
+
+      if not has_similar_variants:
+        continue
+
+      combine_testcases_into_group(testcase_1, testcase_2, testcase_map)
```
oliverchang, line 130: As a starting point, can we log instead of actually combining this? This will give us a chance to evaluate what we're doing and fix any potential issues before we actually group them.

src/appengine/handlers/cron/grouper.py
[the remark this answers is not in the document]
Navidem: No I see your point! Done :)

src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -1535,6 +1535,15 @@ def get_testcase_variant(testcase_id, job_type):
   return variant
 
 
+def get_all_testcase_variants(testcase_id):
+  """Get all testcase variant entities based on testcase id."""
+  variants_query = data_types.TestcaseVariant.query(
+      data_types.TestcaseVariant.testcase_id == testcase_id)
+  if not variants_query:
```
oliverchang, line 1542: I think this will never be the case. can omit.

src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -1535,6 +1535,15 @@ def get_testcase_variant(testcase_id, job_type):
...
+    return []
+  return  list(variants_query.iter())
```
oliverchang, line 1544: No need to turn this into an eager list. We can just return the query as-is. Users can iterate over it to get all results.

src/appengine/handlers/cron/grouper.py
Navidem: can you elaborate on how to log? like in a file or stdout or etc?

src/appengine/handlers/cron/group_leader.py
```diff
@@ -62,14 +66,35 @@ def _key_func(testcase):
...
+    security_index = None
+    i386_indexes = []
+    for idx, item in enumerate(items):
       item.is_leader = False
-
-    leader_index = find_index(items, is_reproducible_and_has_issue)
+      if not security_index and item.security_flag:
+        security_index = idx
+      try:
+        testcase = data_handler.get_testcase_by_id(item.id)
```
oliverchang, line 77: Can we avoid the `get_testcase_by_id` here? That may slow things down. If needed, we can add the needed attribute (`job_type`) to `FORWARDED_ATTRIBUTES` in `grouper` instead to avoid additional queries.

src/appengine/handlers/cron/grouper.py
```diff
@@ -86,6 +86,48 @@ def _get_new_group_id():
...
+      # Rule: If both testcase have the same group id, then no work to do.
+      if testcase_1.group_id == testcase_2.group_id and testcase_1.group_id:
+        continue
+
+      # Rule: Check both testcase are under the same project.
+      if testcase_1.project_name != testcase_2.project_name:
+        continue
+
+      # Rule: Group testcase with similar variants.
+      testcase_1_variants = data_handler.get_all_testcase_variants(
```
oliverchang, line 114: Thinking more, this may end up being quite slow. 

We can optimize this here, by **only** querying for a single TestcaseVariant with the testcase_2's job_type. 

i.e. 

for `testcase_1`, just check if there is a testcase variant with the job type for `testcase_2`. 

I think this should be sufficient, because this relationship should be symmetric. I.e. it should be redundant to query for a testcase variant with id=`testcase_2` and job type for `testcase_1`.

src/appengine/handlers/cron/grouper.py
oliverchang: Just logs.log

src/appengine/handlers/cron/group_leader.py
Navidem: yes that makes things easier!

Navidem: will pick it up tmrw.

src/appengine/handlers/cron/grouper.py
```diff
@@ -86,6 +86,48 @@ def _get_new_group_id():
...
+      # comparisons.
+      if testcase_1_id <= testcase_2_id:
+        continue
+
+      # Rule: If both testcase have the same group id, then no work to do.
+      if testcase_1.group_id == testcase_2.group_id and testcase_1.group_id:
+        continue
+
+      # Rule: Check both testcase are under the same project.
+      if testcase_1.project_name != testcase_2.project_name:
```
oliverchang, line 110: Once you add the `job_type` property, also check if they're equal here. If they are, we can bail because variant analysis is irrelevant.

src/appengine/handlers/cron/grouper.py
Navidem: This may miss some similar cases: assume `testcase1` has a `variant1` of job_type `vt1` and `testcase2` has a `variant2` of job_type `vt2` where `vt1 != vt2`.
If these two variants are similar based on `is_same_variant` we have to merge them, no?

src/appengine/handlers/cron/grouper.py
Navidem: Done.

src/appengine/handlers/cron/group_leader.py
Navidem: Done.

src/clusterfuzz/_internal/datastore/data_handler.py
Navidem: Done.

src/clusterfuzz/_internal/datastore/data_handler.py
Navidem: Done.

src/appengine/handlers/cron/grouper.py
Navidem: Done.

src/appengine/handlers/cron/grouper.py
oliverchang: In practice, this case should not happen. Our variant analysis gets scheduled for all eligible jobs within the project. i.e. for `testcase1`, we will run analysis jobs for `vt2` in all cases. For `testcase2`, we will run analysis jobs for `vt1` in all cases.

So there should never be a case where a variant does not exist for `testcase1` with `vt2` and `testcase2` and `vt1`, assuming that they are actually the same.

Navidem: I see! So basically for each `testcase` the analysis is run for all possible `job_type` and if a crash is produced, then we can check for equivalence.

src/appengine/handlers/cron/group_leader.py
```diff
@@ -62,14 +64,30 @@ def _key_func(testcase):
...
+        security_index = idx
+      if not asan_index and item.job_type and '_asan_' in item.job_type:
+        asan_index = idx
+      if item.job_type and environment.is_i386(item.job_type):
+        i386_indexes.append(idx)
+
+    leader_index = security_index
+    leader_index = asan_index
+    if leader_index is None:
+      leader_index = find_index(items, is_reproducible_and_has_issue)
```
oliverchang, line 82: We still need to keep these prioritisation steps wrt reproducibility and issues. We would also prefer to file a reproducible non-security bug than a non-reproducible security bug. 

One way to keep this is to just sort the `items` list multiple times (`sort()` in python should be stable), first by timestamp, then by a score that we give based on if it's asan/security and its architecture (e.g. asan -> +1, security -> +1, non-x86 -> +1). 

Please also add tests that would've caught this.

oliverchang (changes requested):
some more changes needed with leader picking. Please address and I'll merge/deploy this tmrw to test.

src/appengine/handlers/cron/grouper.py
```diff
@@ -36,10 +36,12 @@ class TestcaseAttributes(object):
 
   __slots__ = ('id', 'is_leader', 'issue_id') + FORWARDED_ATTRIBUTES
 
-  def __init__(self, testcase_id):
+  def __init__(self, testcase_id, job_type=None, sec_flag=None):
```
oliverchang, line 39: Rather than add these properties in the constructor, can we just set them later when we need them? 

No other properties get this, and there's no reason these two ones we added get this special treatment :)

Navidem: Done.

src/appengine/handlers/cron/group_leader.py
Navidem: Fixed the leader selection.

src/appengine/handlers/cron/group_leader.py
```diff
@@ -62,14 +64,30 @@ def _key_func(testcase):
...
-
-    leader_index = find_index(items, is_reproducible_and_has_issue)
+      if not security_index and item.security_flag:
+        security_index = idx
+      if not asan_index and item.job_type and '_asan_' in item.job_type:
+        asan_index = idx
+      if item.job_type and environment.is_i386(item.job_type):
+        i386_indexes.append(idx)
+
+    leader_index = security_index
```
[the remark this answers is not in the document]
Navidem: Fixed.

src/appengine/handlers/cron/group_leader.py
```diff
@@ -62,14 +64,30 @@ def _key_func(testcase):
...
 
-    for item in items:
+    asan_index = None
+    security_index = None
+    i386_indexes = []
+    for idx, item in enumerate(items):
       item.is_leader = False
-
-    leader_index = find_index(items, is_reproducible_and_has_issue)
+      if not security_index and item.security_flag:
```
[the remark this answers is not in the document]
Navidem: Fixed.

src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -1535,6 +1535,12 @@ def get_testcase_variant(testcase_id, job_type):
   return variant
 
 
+def get_all_testcase_variants(testcase_id):
```
oliverchang, line 1538: No longer needed. Let's delete this in another PR.

oliverchang (approved):

PR #2794, opened 2022-09-20
Support downloading multiple builds for a single job.

oliverchang:
This is necessary for Centipede, where the main binary is not sanitized and additional binaries are required to support sanitizers.

Add support for EXTRA_BUILD_BUCKET_PATH, which will unpack additional binaries into /build/dir/revisions/__extra_build.

### DIFF
--- a/src/clusterfuzz/_internal/build_management/build_manager.py
+++ b/src/clusterfuzz/_internal/build_management/build_manager.py
@@ -76,6 +76,7 @@ FUZZ_TARGET_ALLOWLISTED_PREFIXES = [
     'afl-fuzz',
     'afl-showmap',
     'afl-tmin',
+    'centipede',
     'honggfuzz',
     'jazzer_agent_deploy.jar',
     'jazzer_driver',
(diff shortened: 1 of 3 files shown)


Alan32Liu (approved):
Thanks!

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -1341,15 +1346,41 @@ def setup_regular_build(revision,
...
+  else:
+    return None
+
+  # Additional binaries to pull (for fuzzing engines such as Centipede).
+  extra_bucket_path = get_bucket_path('EXTRA_BUILD_BUCKET_PATH')
+  if extra_bucket_path:
+    # Import here as this path is not available in App Engine context.
+    from clusterfuzz._internal.bot.fuzzers import utils as fuzzer_utils
+    extra_build_urls = get_build_urls_list(extra_bucket_path)
+    extra_build_url = revisions.find_build_url(extra_bucket_path,
```
jonathanmetzman, line 1367: I'm a bit unsure how something like this will work for chrome (and I assume google3).
In chrome different sanitizer builds are done by different builders that do not communicate, and there may not be a build for each sanitizer for a specific commit . How are builds from the same commit supposed to have the same revision?

oliverchang: We'll need to get them to support it. I think the disk space benefit of our approach here is too much to pass up. 

The disk space benefit here is given two builds:

ASan: 5GB
NoSanitizer: 4GB

Rather than consuming 5GB+4GB to download a single zip and then unpacking, this can consume 5GB at most for the zips.

1. Download ASan (5GB)
2. Unpack (very minimal disk usage from selective unpack). The zip is deleted. 
3. Download NoSanitizer build (4GB.)
4. Unpack (again, very minimal disk usage from selective unpack). The zip is deleted.

oliverchang: Also note that this implementation doesn't stop someone from having a single big zip. They just need to follow our expected directory format, and it'll still work.

PR #2797, opened 2022-09-21
[Jira bug] Use jira issue object to fix ticket creation bug

boni-sha:
Current implementation:
Whenever we want to create a new jira ticket we first create a blank ticket(with dummy values) and then update that ticket with desired fields.

Issues:
If the update to these tickets fail (eg: wrong field/value) Then we end up create a lot of dummy tickets because of retries.

Solution:
Eliminate the step to create a ticket with dummy values. Looking at how things are implemented for monorail the right approach here would be to use jira issue objects locally to store all the information and create the ticket with all the desired values with the help of the object. In case if ticket creation fails then this does not create garbage tickets


src/appengine/libs/issue_management/jira/issue_tracker_manager.py
```diff
@@ -47,73 +47,94 @@ def _create_client(self):
...
-        'summary': 'Default summary',
-        'description': 'Default description',
-        'issuetype': {
-            'name': 'Bug'
-        }
-    }
-
-    jira_issue = self.client.create_issue(fields=default_fields)
+    """Create an issue object locally."""
+    raw_fields = {"id": "-1", "fields": {"components": [], "labels": []}}
```
oliverchang, line 56: nit: please use single quotes for all strings.

oliverchang (approved):

src/appengine/libs/issue_management/jira/issue_tracker_manager.py
boni-sha: done [b631f0a](https://github.com/google/clusterfuzz/pull/2797/commits/b631f0addb6aa6d0929151df53bc31a992ac6cf8)

oliverchang (approved):

PR #2798, opened 2022-09-22
[fuchsia] Enforce timeouts on undercoat calls

eepeep:
The Fuchsia libFuzzer runner was ignoring the timeouts passed to it, meaning that a bot could get stuck for a long time if the fuzzing run hangs for any reason. As a fallback mechanism in this sort of situation, this plumbs the timeout through to the call to the subprocess, adding a significant amount of padding along the way (for reasons described in the comments).


eepeep:
cc @nopsledder @oliverchang

It looks like there is no straightforward way to distinguish in the logs a subprocess that was killed by timeout vs some other failure mode, unless you look at the timestamps and notice that undercoat didn't return any error on stderr itself, but this should hopefully at least help things recover if anything goes awry.

oliverchang (approved):

PR #2809, opened 2022-09-30
Delete reproduce_tool.

oliverchang:
This is deprecated, unmaintained, and unused.


Alan32Liu (approved):

src/appengine/handlers/testcase_detail/show.py
```diff
@@ -508,7 +508,7 @@ def get_testcase_detail(testcase):
       'crash_address':
           crash_address,
       'crash_state':
-          crash_state,  # Used by reproduce tool.
+          crash_state,
```
jonathanmetzman, line 511: If reproduce tool was the only use, should we get rid of this. or leave a todo?

src/local/butler/bootstrap.py
```diff
@@ -18,14 +18,10 @@
 from local.butler import common
 
 
-def execute(args):
+def execute(_):
```
jonathanmetzman, line 21: I think you've said that `del X` is the preferred style these days.

jonathanmetzman (approved):
lgtm

src/appengine/handlers/testcase_detail/show.py
oliverchang: I think this is useful enough to keep.

src/local/butler/bootstrap.py
oliverchang: good call. fixed!

josepharhar:
@oliverchang I used reproduce.sh for reproducing all of my clusterfuzz bugs. Is there an alternative program/command that I can just give a clusterfuzz.com/testcase-detail URL that will reproduce crashes for me?

PR #2813, opened 2022-10-03
Update `yapf`

DonggeLiu:
Not sure if I misinterpreted this, but I reckon [`chromium` has been changed to `pep8`](https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/python/python.md#yapf)?
Alternatively, it can be [`google` or `yapf`](https://github.com/google/yapf/blob/main/README.rst#formatting-style).
`yapf` has an error, though.

### DIFF
--- a/.style.yapf
+++ b/.style.yapf
@@ -1,5 +1,5 @@
 [style]
-based_on_style = chromium
+based_on_style = google
 column_limit = 80
 indent_width = 2
 split_before_named_assigns = true


.style.yapf
```diff
@@ -1,5 +1,5 @@
 [style]
-based_on_style = chromium
+based_on_style = pep8
```
jonathanmetzman, line 2: Should probably do "google" instead.

Alan32Liu: Done : )

jonathanmetzman:
@oliverchang should we run yapf with the new settings on everything? Or wait to do this file-by-file?

Alan32Liu:
> > @oliverchang should we run yapf with the new settings on everything? Or wait to do this file-by-file?
> 
> Lacking some context here, did something break with the `chromium` setting here?

Nothing is broken or urgent. I just noticed that `chromium` might be [outdated](https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/python/python.md#yapf) when I was testing an automatic formatter that uses [.style.yapf](https://github.com/google/clusterfuzz/pull/2813/files#diff-afbcca4a5d095c51ca7a560184b48a4702f2588bb44ae83fc91081a1db04fada).

oliverchang:
> > > @oliverchang should we run yapf with the new settings on everything? Or wait to do this file-by-file?
> > 
> > 
> > Lacking some context here, did something break with the `chromium` setting here?
> 
> Nothing is broken or urgent. I just noticed that `chromium` might be [outdated](https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/python/python.md#yapf) when I was testing an automatic formatter that uses [.style.yapf](https://github.com/google/clusterfuzz/pull/2813/files#diff-afbcca4a5d095c51ca7a560184b48a4702f2588bb44ae83fc91081a1db04fada).

Ack, as @jonathanmetzman suggested, if we change this, we need to ensure that it doesn't break linting for all of our current files before merging this change. This seems lower priority if there's a lot to fix and there's no urgency.

jonathanmetzman:
> > > > @oliverchang should we run yapf with the new settings on everything? Or wait to do this file-by-file?
> > > 
> > > 
> > > Lacking some context here, did something break with the `chromium` setting here?
> > 
> > 
> > Nothing is broken or urgent. I just noticed that `chromium` might be [outdated](https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/python/python.md#yapf) when I was testing an automatic formatter that uses [.style.yapf](https://github.com/google/clusterfuzz/pull/2813/files#diff-afbcca4a5d095c51ca7a560184b48a4702f2588bb44ae83fc91081a1db04fada).
> 
> Ack, as @jonathanmetzman suggested, if we change this, we need to ensure that it doesn't break linting for all of our current files before merging this change. This seems lower priority if there's a lot to fix and there's no urgency.

It was really a question, because there are a lot of benefits to either.
Not fixing files:
Doesn't absolutely screw up history by modifying every file.

Fixing files:
Doesn't annoy future committers by forcing them to format files as well.

Alan32Liu:
> > > > > @oliverchang should we run yapf with the new settings on everything? Or wait to do this file-by-file?
> > > > 
> > > > 
> > > > Lacking some context here, did something break with the `chromium` setting here?
> > > 
> > > 
> > > Nothing is broken or urgent. I just noticed that `chromium` might be [outdated](https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/python/python.md#yapf) when I was testing an automatic formatter that uses [.style.yapf](https://github.com/google/clusterfuzz/pull/2813/files#diff-afbcca4a5d095c51ca7a560184b48a4702f2588bb44ae83fc91081a1db04fada).
> > 
> > 
> > Ack, as @jonathanmetzman suggested, if we change this, we need to ensure that it doesn't break linting for all of our current files before merging this change. This seems lower priority if there's a lot to fix and there's no urgency.
> 
> It was really a question, because there are a lot of benefits to either. Not fixing files: Doesn't absolutely screw up history by modifying every file.
> 
> Fixing files: Doesn't annoy future committers by forcing them to format files as well.

Although we probably won't merge this soon: When it is time to merge this, would it be a good idea if I use this PR to fix all files? In this way, we can keep the lint/format consistent with the `.style.yapf` file as well.
This will also fix the `f-string` and `u-string` issues in existing files, which we have to do in different new PRs anyway.

jonathanmetzman:
> > > > > > @oliverchang should we run yapf with the new settings on everything? Or wait to do this file-by-file?
> > > > > 
> > > > > 
> > > > > Lacking some context here, did something break with the `chromium` setting here?
> > > > 
> > > > 
> > > > Nothing is broken or urgent. I just noticed that `chromium` might be [outdated](https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/python/python.md#yapf) when I was testing an automatic formatter that uses [.style.yapf](https://github.com/google/clusterfuzz/pull/2813/files#diff-afbcca4a5d095c51ca7a560184b48a4702f2588bb44ae83fc91081a1db04fada).
> > > 
> > > 
> > > Ack, as @jonathanmetzman suggested, if we change this, we need to ensure that it doesn't break linting for all of our current files before merging this change. This seems lower priority if there's a lot to fix and there's no urgency.
> > 
> > 
> > It was really a question, because there are a lot of benefits to either. Not fixing files: Doesn't absolutely screw up history by modifying every file.
> > Fixing files: Doesn't annoy future committers by forcing them to format files as well.
> 
> Although we probably won't merge this soon: When it is time to merge this, would it be a good idea if I use this PR to fix all files? In this way, we can keep the lint/format consistent with the `.style.yapf` file as well. This will also fix the `f-string` and `u-string` issues in existing files, which we have to do in different new PRs anyway.

I'm generally against mass formatting changes so I actually think maybe let files break in future lints.
That for example is the policy of LLVM (can't find references)
There's a discussion here about ignoring it in git blame: https://stackoverflow.com/questions/53502654/how-do-i-run-a-code-formatter-over-my-source-without-modifying-git-history maybe we format everything and go with that?

jonathanmetzman:
@oliverchang I was thinking about this since I'm about to do a big refactor. Maybe we should land the upgrade without formatting everything and format files as they are modified.

PR #2820, opened 2022-10-04
Cast `reproducer_path` as a `str` from `PosixPath`

Alan32Liu:
`engine.Crash()` expects the `reproducer_path` to be a `str` instead of `Path`.
This was not captured by the test cases but was discovered via this error after deployment:
```
TypeError: PosixPath('/path/to/inputs/fuzzer-testcases/1489f923c4dca729178b3e3233458550d8d has type PosixPath, but expected one of: bytes, unicode
```


oliverchang (approved):
LGTM, but can we also add a testcase later that would have caught this?

PR #2834, opened 2022-10-28
Fix BQ load scalability issues

### LINKED ISSUE #2832: libFuzzer stats on oss-fuzz not appearing after October 20th
https://oss-fuzz.com/fuzzer-stats?group_by=by-day&date_start=2022-10-12&date_end=2022-10-25&fuzzer=libFuzzer&job=libfuzzer_asan_skia&project=skia

oliverchang:
- Shard loads by filename prefix. Filenames are all named with <hash>.json, so we can enumerate hex digit prefixes. 
- Poll for completion of load jobs. 
- Add threading to parallelize loads by fuzzer. This mitigates some of the slowdown of the polling. 

Fixes #2832

### DIFF
--- a/src/appengine/handlers/cron/load_bigquery_stats.py
+++ b/src/appengine/handlers/cron/load_bigquery_stats.py
@@ -13,8 +13,10 @@
 # limitations under the License.
 """Handler used for loading bigquery data."""
 
+from concurrent.futures import ThreadPoolExecutor
 import datetime
 import random
+import string
 import time
 
 from googleapiclient.errors import HttpError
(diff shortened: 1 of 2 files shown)


Alan32Liu (approved):
LGTM.
Would it be a good idea to set number of threads dynamically? E.g., with [`multiprocessing.cpu_count()`](https://docs.python.org/3.10/library/multiprocessing.html?highlight=cpu_count#multiprocessing.cpu_count).

oliverchang:
> LGTM. Would it be a good idea to set number of threads dynamically? E.g., with [`multiprocessing.cpu_count()`](https://docs.python.org/3.10/library/multiprocessing.html?highlight=cpu_count#multiprocessing.cpu_count).

This is mostly io-bound, so probably not needed here.

PR #2836, opened 2022-11-02
Two phase variant based deduplication

Navidem:


src/appengine/handlers/cron/grouper.py
```diff
@@ -135,9 +147,58 @@ def _group_testcases_based_on_variants(testcase_map):
           not is_same_variant(candidate_variant, testcase_2)):
         continue
 
+      if current_project in grouping_candidates:
+        grouping_candidates[current_project].append((testcase_1_id,
```
oliverchang, line 151: you can simplify this and avoid the if statement with: 

```python
grouping_candidates.setdefault(current_project, []).append((testcase_1_id, testcase_2_id))
```

src/appengine/handlers/cron/grouper.py
```diff
@@ -31,6 +31,7 @@
 GROUP_MAX_TESTCASE_LIMIT = 25
 
 VARIANT_CRASHES_IGNORE = ['Out-of-memory', 'Timeout']
+VARIANT_THRESHOLD_PERCENTAGE = 0.05
```
oliverchang, line 34: Let's make this higher. 5% feels a bit low. Maybe 20%?  

The case we saw was of a startup crash in a bad job type, where in theory **every** single testcase will result in the same result in that bad job type. 20% is still fairly conservative here.

oliverchang (commented):
nice! can you please also add a small test to make sure this works?

src/appengine/handlers/cron/grouper.py
Navidem: Thanks for that!
I usually used `collections.defaultdict`, was not sure if we use it in our code?

oliverchang: yes we can definitely use that too!

Navidem: Great! Done.

src/appengine/handlers/cron/grouper.py
```diff
@@ -135,9 +147,58 @@ def _group_testcases_based_on_variants(testcase_map):
...
+        to_ignore_testcases.add(testcase_id)
+
+    for grouping_candidates in candidate_list:
+      ingnored = False
+      for testcase in grouping_candidates:
+        if testcase in to_ignore_testcases:
+          logs.log('VARIANT ANALYSIS (Pruning): Anomalous testcase: (id=%s, '
+                   'matched_count=%d >= threshold=%.2f).' %
+                   (testcase, counter[testcase], threshold))
+          ingnored = True
```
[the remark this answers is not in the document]
Navidem: Done.

src/appengine/handlers/cron/grouper.py
```diff
@@ -135,9 +147,58 @@ def _group_testcases_based_on_variants(testcase_map):
...
+    # Determine anomalous candidates.
+    threshold = (
+        VARIANT_THRESHOLD_PERCENTAGE * project_testcases_counter[project])
+    for testcase_id, count in counter.items():
+      if count >= threshold:
+        to_ignore_testcases.add(testcase_id)
+
+    for grouping_candidates in candidate_list:
+      ingnored = False
+      for testcase in grouping_candidates:
```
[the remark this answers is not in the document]
Navidem: Yes, but I wanted to log the exact anomalous testcase.
Made the change to have less indentation and got rid of `ignored`. The `TODO` mentioned that the two checks should be combined as you suggested once we are done with evaluation.

src/appengine/handlers/cron/grouper.py
```diff
@@ -98,20 +99,31 @@ def is_same_variant(variant1, variant2):
...
+          project_testcases_counter[testcase.project_name] = 1
+
       # Rule: If both testcase have the same group id, then no work to do.
       if testcase_1.group_id == testcase_2.group_id and testcase_1.group_id:
         continue
 
       # Rule: Check both testcase are under the same project.
       if testcase_1.project_name != testcase_2.project_name:
         continue
+      current_project = testcase_1.project_name
```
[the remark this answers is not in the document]
Navidem: Done.

src/appengine/handlers/cron/grouper.py
```diff
@@ -98,20 +99,31 @@ def is_same_variant(variant1, variant2):
 def _group_testcases_based_on_variants(testcase_map):
   """Group testcases that are associated based on variant analysis."""
   logs.log('Grouping based on variant analysis.')
+  grouping_candidates = {}
+  project_testcases_counter = {}
```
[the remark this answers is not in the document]
Navidem: Done.

oliverchang (approved):
thanks!

PR #2837, opened 2022-11-03
Minor BQ loader fixes.

oliverchang:
- Reduce number of threads to avoid OOM.
- Account for rare edge case with the first load failing and subsequent loads incorrectly using WRITE_APPEND instead of WRITE_TRUNCATE.
- Don't bother loading if no schema is defined.


jonathanmetzman (approved):
lgtm

PR #2849, opened 2022-11-14
Add consistency to illegal-instruction handling

pgrace-google:


src/clusterfuzz/stacktraces/__init__.py
```diff
@@ -416,6 +416,15 @@ def parse(self, stacktrace: str) -> CrashInfo:
         if state.crash_type in IGNORE_CRASH_TYPES_FOR_ABRT_BREAKPOINT_AND_ILLS:
           continue
 
+      # Windows formats and surfaces ILL differently from Linux/Posix
+      # Update crash type for consistency.
+      self.update_state_on_match(
```
oliverchang, line 421: Can you please add a test? 

See https://github.com/google/clusterfuzz/commit/da388a2b7e153c05fc8d4fa93aaf411b107c99f3 for an example.

src/clusterfuzz/_internal/crash_analysis/crash_analyzer.py
```diff
@@ -30,6 +30,21 @@
...
+  'Floating-point-exception',
+  'RUNTIME_ASSERT', # RUNTIME_ASSERT in V8 (not a crash, but is a sign of an error
+  'V8 correctness failure',
+  'Timeout',
+  'Out-of-memory',
+  'Unexpected-exit',
+  'Security CHECK failure', # Release SECURITY_CHECK in Blink shouldn't be marked as a security bug
+]
 SIGNAL_SIGNATURES_NOT_SECURITY = [
     'Sanitizer: ABRT',
```
oliverchang, line 49: You probably still want to add the Illegal-signature one here. The reason why we have this separate check is there can be multiple stacktraces in the one output. We want to mark crashes in these cases as non security even if the crash type we end up getting is not one of these signal ones.

pgrace-google: by Illegal-signature you mean illegal-instruction, right? if so, added!
if not, where was the illegal-signature supposed to be in the stacktrace? i totally missed that

src/clusterfuzz/stacktraces/__init__.py
pgrace-google: yessss thank you for the example! super helpful!

pgrace-google: (also I overwrote the old commits with the good intention of consolidating/cleaning the commits.. unsure if thats a git taboo - let me know if it is!! wont do it again c: )

src/clusterfuzz/_internal/crash_analysis/crash_analyzer.py
oliverchang: Yep that was a typo, thanks!

src/clusterfuzz/stacktraces/__init__.py
oliverchang: No need to overwrite older commits. When we merge everything will get squashed in the end :)

oliverchang (approved):
Thanks! LGTM

PR #2850, opened 2022-11-14
Enable Variant-based Grouping

Navidem:


oliverchang (approved):
LGTM, but I noticed another bad case we need to fix first related to stacktrace parsing. I added that example in the doc.

We need to fix that before we merge/deploy this.

oliverchang:
fyi I added one more case to ignore here: https://github.com/google/clusterfuzz/pull/2850/commits/939c02ea2bcea56a2a70fb6ecf2996c26ab7d7ff

PR #2852, opened 2022-11-15
Add additional logging for test_reproduce_reliability.

flowerhack:
We're trying to better understand the performance of the reproduction tasks, and this additional logging will make it simpler to filter/understand the relevant data.


src/clusterfuzz/_internal/bot/testcase_manager.py
```diff
@@ -750,10 +751,12 @@ def test_reproduce_reliability(self, retries, expected_state,
       crash_count += 1
       if crash_count >= reproducible_crash_target_count:
         logs.log('Crash is reproducible.')
+        logs.log('Concluded a reproducibility test.')
```
oliverchang, line 754: Are these logs necessary, given that they are preceded by consistent logs?

flowerhack: Fair point!  Updated CL to just have the start-of-repro logs.

oliverchang (approved):

PR #2859, opened 2022-11-21
Add GPU Failure type to variant ignore list.

oliverchang:
These are almost always issues in the environment rather than real issues.


oliverchang:
TBR

Navidem (commented):
LGTM

PR #2872, opened 2022-12-01
Add logging to fetch_artifact.py to improve debuggability

MHA3:


src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
```diff
@@ -40,8 +40,8 @@ def execute_request_with_retries(request):
     try:
       result = request.execute()
       break
-    except Exception as e:
-      logs.log_error(f'Error calling endpoint {request.uri}: Error {e}')
+    except Exception:
+      pass
```
oliverchang, line 44: Why are we removing this log message?

src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
```diff
@@ -100,16 +100,24 @@ def get_artifacts_for_build(client, bid, target, attempt_id='latest'):
   request = client.buildartifact().list(
       buildId=bid, target=target, attemptId=attempt_id)
 
+  request_str = '{}, {}, {}, {}'.format(request.uri, request.method,
```
oliverchang, line 103: Can you define this closer to where it's used? i.e. right before line 118. 

We can also simplify this by doing a `', '.join([request.uri, ...])` instead.

src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
```diff
@@ -148,10 +156,13 @@ def get_latest_artifact_info(branch, target, signed=False):
       successful=True,
       maxResults=1,
       signed=signed)
+  request_str = '{}, {}, {}, {}'.format(request.uri, request.method,
```
oliverchang, line 159: Same comment here re using `', '.join` instead. ,

src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
MHA3: That was a mistake on my part. I thought I had merged the upstream changes. Let me correct this.

src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
MHA3: I tried testing it in a python shell, and it seems like to be able to use `.join`, all the elements of the list should be strings.
`", ".join(['abc', [1, 2, 3], {'a':1, 'b':2}])` fails, but `"{}, {}, {}".format('abc', [1, 2, 3], {'a':1, 'b':2})` works.
I am not certain if all the items would be strings in this case.

src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
MHA3: Same reason as above.

src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
MHA3: I had to do it before the `request` was used up. If I understand the existing code correctly, it is being used as an iterable. Therefore, if I used it closer to the logging line, that we may not get the full `request` object.

src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
jonathanmetzman: Just use an f-string instead of calling format.

src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
jonathanmetzman: Use an f-string instead of format.

jonathanmetzman (changes requested):

src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
```diff
@@ -100,16 +100,25 @@ def get_artifacts_for_build(client, bid, target, attempt_id='latest'):
   request = client.buildartifact().list(
       buildId=bid, target=target, attemptId=attempt_id)
 
+  request_str = f'{request.uri}, {request.method},\
```
jonathanmetzman, line 103: Please use paren to seperate these strings instead of backslashes: https://google.github.io/styleguide/pyguide.html#32-line-length

jonathanmetzman (approved):
LGTM

PR #2884, opened 2022-12-07
Add parsing for extra sanitizers DNS crash type.

oliverchang:


oliverchang:
TBR since I've deployed the extra sanitizers and want to avoid this filing bugs.

oliverchang:
Actually I think because of unshare (that removes network access), we won't find anything in our production infra. Merging this anyway just to be safe.

oliverchang:
> LGTM. Did you deploy this?

Yes, but see https://github.com/google/clusterfuzz/pull/2888 which is also needed.

PR #2888, opened 2022-12-08
Temporarily disable unshare when using extra sanitizers.

oliverchang:
Otherwise the DNS sanitizer will never work.

### DIFF
--- a/src/clusterfuzz/_internal/system/new_process.py
+++ b/src/clusterfuzz/_internal/system/new_process.py
@@ -456,8 +456,13 @@ class ModifierProcessRunnerMixin(object):
     if additional_args:
       command.extend(additional_args)
 
-    return self.tool_prefix('unshare') + self.tool_prefix(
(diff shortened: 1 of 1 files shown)


oliverchang:
@jonathanmetzman the sanitizer may still have a hard time working due to this requirement  (and because the vuln.dict isn't propagated right now)

https://github.com/google/oss-fuzz/blob/8371f162ca25ff980d1e613f6779534634f359f0/infra/experimental/SystemSan/inspect_dns.cpp#L163

We might need to look at relaxing this.

Alan32Liu (approved):
LGTM

jonathanmetzman (approved):
LGTM

jonathanmetzman:
It's interesting that I don't think we are using unshare for honggfuzz builds https://github.com/search?q=repo%3Agoogle%2Fclusterfuzz%20ModifierProcessRunner&type=code

PR #2892, opened 2022-12-08
Create a simplified testcase upload form

pgrace-google:
I could not - for the love of my work life balance - figure out how to dynamically update (not append) the text that is nested inside a template, so I created the monstrosity that is "thisIsNotTheRightWayButIKnowNoBetter" hoping that you have a way I could not find in my top 5 google search pages


src/appengine/private/components/upload-testcase/upload-form-simplified.html
```diff
@@ -0,0 +1,414 @@
...
+            class="dropdown-filter"
+            no-animations="true"
+            label="Platform"
+            title="Linux, Windows, Mac, or Android"
+            noink>
+          <paper-listbox
+              slot="dropdown-content"
+              selected="{{uploadParams.platform}}"
+              attr-for-selected="val">
+            <paper-item on-click="submit" val="Linux">Linux</paper-item>
```
oliverchang, line 163: did you mean to have on-click="submit" here?

src/appengine/private/components/upload-testcase/upload-testcase.html
```diff
@@ -261,7 +264,10 @@
     <div id="mainContent">
       <div class="toolbar">
         <div class="permission">[[getPermissionText(params)]]</div>
-        <div class="buttons"><paper-button raised toggles active="{{uploadFormOpened}}">Upload</paper-button></div>
+        <div class="buttons">
+          <paper-button raised toggles active="{{uploadFormOpened}}">Upload</paper-button>
+          <paper-button raised toggles active="{{uploadFormSimplifiedOpened}}">Quick Upload</paper-button>
```
oliverchang, line 269: We need to make this conditionally show up only being a chromium deployment. You can use this: https://github.com/google/clusterfuzz/blob/910f08b9316a729c4c6b05ed260f97d1d03c3d88/src/appengine/handlers/upload_testcase.py#L241

src/appengine/private/components/upload-testcase/upload-form-simplified.html
```diff
@@ -0,0 +1,414 @@
...
+              no-animations="true"
+              label="Teset case file type"
+              title="html, js, or pdf"
+              noink>
+            <paper-listbox
+                slot="dropdown-content"
+                selected="{{uploadParams.fileType}}"
+                fallback-selection="html"
+                attr-for-selected="val">
+              <paper-item on-click="submit" val="html">html</paper-item>
```
oliverchang, line 136: Do you mean to have on-click="submit" here?

oliverchang (commented):
Sorry for the slow review!

src/appengine/private/components/upload-testcase/upload-form-simplified.html
```diff
@@ -0,0 +1,414 @@
...
+              selected="{{uploadParams.platform}}"
+              attr-for-selected="val">
+            <paper-item on-click="submit" val="Linux">Linux</paper-item>
+            <paper-item on-click="submit" val="Windows">Windows</paper-item>
+            <paper-item on-click="submit" val="Mac">Mac</paper-item>
+            <paper-item on-click="submit" val="Android">Android</paper-item>
+          </paper-listbox>
+        </paper-dropdown-menu>
+      </div>
+      <template is="dom-if" if="[[thisIsNotTheRightWayButIKnowNoBetter(uploadParams.file, uploadParams.fileType, uploadParams.platform, 'linux_asan_chrome_mp')]]">
```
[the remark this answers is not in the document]
pgrace-google: Yeah it doesn't work :(
I'm trying to show the job we will be running, but once the params are changed, the job variable itself is changed, but the displayed text of the job name doesnt change!

src/appengine/private/components/upload-testcase/upload-testcase.html
pgrace-google: oh yesss will do!

src/appengine/private/components/upload-testcase/upload-form-simplified.html
```diff
@@ -0,0 +1,414 @@
...
+          <p style="background-color: #AFE1AF; padding: 2px">Running linux_asan_chrome_mp</p>
```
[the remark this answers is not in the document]
pgrace-google: done!

src/appengine/private/components/upload-testcase/upload-form-simplified.html
pgrace-google: SO, no. and i could have sworn that removing it made it behave weirdly, but ive just tested TWICE and it works great c:

src/appengine/private/components/upload-testcase/upload-form-simplified.html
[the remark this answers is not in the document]
oliverchang: Could you please try again? 

`getJobType(uploadParams)` as in my example likely won't work without explicit `this.set` calls to notify the framework that a containing variable has changed (e.g. https://github.com/google/clusterfuzz/blob/910f08b9316a729c4c6b05ed260f97d1d03c3d88/src/appengine/private/components/crash-stats/crash-stats.html#L453)

If you use specific variables, e.g. `uploadParams.file`, `uploadParams.fileType` etc, that should work AFAICT.

src/appengine/private/components/upload-testcase/upload-form-simplified.html
```diff
@@ -0,0 +1,391 @@
...
+      handleGcsResponseAndError(ev) {
+        // Finished uploading to GCS, post the remainder of the data to
+        // ClusterFuzz.
+        // Errors are ignored here also. Because the upload is cross-site, we
+        // defer to the server to check that upload succeeded.
+        this.$.ajaxFinish.generateRequest();
+      }
+
+      handleResponse(ev) {
+        this.set('uploadParams.job', uploadParams.job);
```
oliverchang, line 339: The flow of data is a little confusing, and it seems like having this here fixes it inadvertently, but it's not the right place for it.

This should instead be called whenever the `job` variable itself is set (i.e. line 288 in your current revision), but following the thread a little, it's a bit confusing why `updateJobFromPlatformAndFile` is only called from `showWarning`. 

The way this template framework works is more declarative and event driven -- it seems like you're trying to make it write things in an imperative way which leads to all of these issues you're encountering.  You can think of the arguments to these functions here as "dependencies", rather than inputs -- if one of the dependencies changes (i.e. a parameter in a textbox), then the function will be re-called to recalculate the output. 

I'd suggest moving `updateJobFromPlatformAndFile` into `getJobType` inline itself, move the `this.set` call there, and change `showWarning` to accept a job type name instead. 

Then your showWarning call will look something like:

```
<template is="dom-if" if="[[showWarning(uploadParams.job)]]">
```

This will get updated when the frontend framework detects that `uploadParams.job` has changed (via the `this.set` call in `getJobType` which is needed because `uploadParams` is Javascript object). `getJobType` itself has dependencies on `uploadParams.file, uploadParams.fileType, uploadParams.platform`, which may or may not need `this.set` calls in the right places to trigger. 

Does this make sense?

pgrace-google: yesss totally agreed!! thank you!!

oliverchang:
Hey @pgrace-google , sorry for the slowness on getting back to this one after the holiday period. In the future please hit the refresh icon next to my username in the reviewers list to explicitly request another review :)

I deployed this to https://staging3-dot-cluster-fuzz.appspot.com/upload-testcase and some comments:

- The upload does not seem to work. When I click "create" I see: 
```
Uncaught TypeError: Cannot read properties of undefined (reading '$')
    at UploadFormSimplified.computeGcsBody (upload-testcase:32834:23)
    at AjaxDialog.handler (upload-testcase:2568:28)
    at AjaxDialog.submit (upload-testcase:27593:14)
    at PolymerGenerated.handler (upload-testcase:2568:28)
    at Object._fire (upload-testcase:7624:14)
    at Object.forward (upload-testcase:8019:20)
    at Object.click (upload-testcase:7982:14)
    at PolymerGenerated._handleNative (upload-testcase:7402:18)
```

- If a file with an unsupported file extension is selected, no error is generated and the "Job to be used for this testcase:" part remains empty. We should make the error clearer in this case.

pgrace-google:
Good heavens, sorry I didn't test actually clicking upload since I didn't think it would work anyway!! but fixed the $ issue, and added an error message for uncommon file types (only listed a few ones I expect to see), and tested creating different combinations of file and platform!

oliverchang (approved):

PR #2894, opened 2022-12-09
Fix remove file on hang afl device

TheCoryBarker:
When AFL++ fuzz run is initializing and running through the corpus inputs, if one of those corpus files hangs then ClusterFuzz needs to remove that file from the input. It tries to remove this file from the host machine for Android, but needs to remove the file from the Android device instead. This PR is to update that.


oliverchang:
@jonathanmetzman can you please take a look?

jonathanmetzman (approved):
Ah nevermind. I see it's an overridden method.
LGTM

PR #2895, opened 2022-12-12
Unpack all downloading regex file

TheCoryBarker:
When downloading from Android Build based on regex, we need to return a list of all files downloaded, and unpack all of them

### DIFF
--- a/src/clusterfuzz/_internal/platforms/android/flash.py
+++ b/src/clusterfuzz/_internal/platforms/android/flash.py
@@ -71,7 +71,7 @@ def download_latest_build(build_info, image_regexes, image_directory):
   shell.remove_directory(image_directory, recreate=True)
   for image_regex in image_regexes:
     image_file_paths = fetch_artifact.get(build_id, target, image_regex,
-                                         image_directory)
+                                          image_directory)
 
     if not image_file_paths:
       logs.log_error('Failed to download artifact %s for '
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/platforms/android/flash.py
```diff
@@ -70,15 +70,18 @@ def download_latest_build(build_info, image_regexes, image_directory):
...
                      'branch %s and target %s.' %
-                     (image_file_path, build_info['branch'], target))
+                     (image_file_paths, build_info['branch'], target))
       return
-    if image_file_path.endswith('.zip') or image_file_path.endswith('.tar.gz'):
-      archive.unpack(image_file_path, image_directory)
+
+    for file_path in image_file_paths:
+      if file_path.endswith('.zip') or file_path.endswith('.tar.gz'):
+        archive.unpack(file_path, image_directory)
```
oliverchang, line 84: Would these all be part of the same logical "build" ? i.e. is it ever possible these unpacks will clash/overwrite each other in a weird way?

oliverchang (approved):
LGTM with a question

oliverchang:
Also, you need to fix format/lint

src/clusterfuzz/_internal/platforms/android/flash.py
TheCoryBarker: Yes these would all be part of the same logical build (i.e. Coverage enabled Cuttlefish build for x86 or HWASAN cuttlefish build for ARM). I would be surprised if there was clashing of files like boot.img but it seems possible

PR #2907, opened 2022-12-28
Fix local `clusterfuzz` server error

### LINKED ISSUE #2835: Local clusterfuzz failed to install dependencies caused by version of GLIBC on Ubuntu 18.04
## Faulty behavior
When installing dependencies `local/install_deps.bash` following [Prerequisites](https://google.github.io/clusterfuzz/getting-started/prerequisites/), the command `npm install --unsafe-perm -g bower polymer-bundler` result in an error.
Screenshot:
image

## Error
```
node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by node)
```

## Related
#2644 

## Reproduction
- Host: Ubuntu 18.04 on WSL 2.0
- Clusterfuzz: `git clone https://github.com/google/clusterfuzz && git checkout tags/v2.5.7`
- Python: 3.7.7 from source from official repo with following step:
  ```
  RUN apt-get install -y apt-transport-https software-properties-common build-essential git curl \
      libssl-dev zlib1g-dev libncurses5-dev libncursesw5-dev libreadline-dev libsqlite3-dev \
      libgdbm-dev libdb5.3-dev libbz2-dev libexpat1-dev liblzma-dev libffi-dev uuid-dev
  RUN curl -sS https://www.python.org/ftp/python/3.7.7/Python-3.7.7.tgz | tar -C /tmp -xzv && \
      cd /tmp/Python-3.7.7 && \
      ./configure --enable-optimizations && sudo make altinstall && \
      sudo rm -rf /tmp/Python-3.7.7
  ```
- Go: v1.19.2, install as described in [docs](https://go.dev/doc/install)

Next we run `local/install_deps.bash`. This runs result in an error as shown below:
### Version 'GLIBC 2.28' Not Found
```
node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by node)
```

## Expected behavior
All dependencies should be correctly installed with no errors.

## Scope
The host OS Ubuntu 18.04 which is described as a supported system in [docs](https://google.github.io/clusterfuzz/getting-started/prerequisites/#other-dependencies). But with a default Ubuntu 18.04, all dependencies cannot be installed correctly.

Alan32Liu:
Fix #2835 by upgrading `openjdk-8-jdk` to `openjdk-11-jdk`.

### DIFF
--- a/local/install_deps_linux.bash
+++ b/local/install_deps_linux.bash
@@ -111,7 +111,7 @@ sudo apt-get update
 sudo apt-get install -y \
     docker-ce \
     google-cloud-sdk \
-    openjdk-8-jdk \
+    openjdk-11-jdk \
     liblzma-dev
 
 # Install patchelf - latest version not available on some older distros so we


Alan32Liu:
Can confirm this works locally in a clean `Ubuntu:20.04` docker container with `python3.8`.

jonathanmetzman (approved):
LGTM

It's probably not a problem either way since this doesn't affect prod. But we may want to avoid landing this if the issue has existed for two months already. It doesn't prevent me from developing, not sure about you

Alan32Liu:
> LGTM
> 
> It's probably not a problem either way since this doesn't affect prod. But we may want to avoid landing this if the issue has existed for two months already. It doesn't prevent me from developing, not sure about you

It does not affect me either. I reckon this will only affect people who are running their own `clusterfuzz` server locally, IIUC?

oliverchang (approved):

PR #2909, opened 2023-01-03
[libclusterfuzz] Record when a fuzz process times out.

jonathanmetzman:


jonathanmetzman:
If we like this change I guess I will need to implement it for all fuzzers. Or I can leave None as the default.

jonathanmetzman:
Related: https://github.com/google/oss-fuzz/issues/9318

src/clusterfuzz/fuzz/engine.py
```diff
@@ -44,12 +44,14 @@ class FuzzResult(object):
...
 
-  def __init__(self, logs, command, crashes, stats, time_executed):
+  def __init__(self, logs, command, crashes, stats, time_executed,
+               process_timed_out=None):
     self.logs = logs
     self.command = command
     self.crashes = crashes
     self.stats = stats
     self.time_executed = time_executed
+    self.process_timed_out = process_timed_out
```
oliverchang, line 54: nit: should we just name this `timed_out` ? This is more consistent with the other flags (e.g. `time_executed` instead of `process_time_executed`).

oliverchang (approved):

src/clusterfuzz/fuzz/engine.py
jonathanmetzman: Probably, i just think this is more clearly distinguished from e.g. libFuzzer's timeouts

oliverchang: It seems a bit inconsistent with the other flags here which are similarly talking about the process. Let's just rename it, and get this merged.

jonathanmetzman: OK

Issue #2912, opened 2023-01-11
Use renovatebot 


oliverchang:
The current dependabot PRs are too numerous and difficult to merge. We should migrate to renovatebot.

PR #2913, opened 2023-01-11
Disable dependabot.

oliverchang:
To be replaced with https://github.com/google/clusterfuzz/issues/2912 which will allow us to do batched dependency updates much more easily.


Alan32Liu (approved):

jonathanmetzman (approved):
lgtm

PR #2924, opened 2023-01-12
Add timeout for launch_cvd process

MHA3:
This process can sometimes hang, thereby preventing Clusterfuzz from proceeding to fuzz. We would like to kill the process so that Clusterfuzz can have another go at it.

### DIFF
--- a/src/clusterfuzz/_internal/platforms/android/adb.py
+++ b/src/clusterfuzz/_internal/platforms/android/adb.py
@@ -49,6 +49,7 @@ WAIT_FOR_DEVICE_TIMEOUT = 600
 REBOOT_TIMEOUT = 3600
 RECOVERY_CMD_TIMEOUT = 60
 STOP_CVD_WAIT = 20
+LAUNCH_CVD_TIMEOUT = 2700
 
 # Output patterns to parse "lsusb" output.
 LSUSB_BUS_RE = re.compile(r'Bus\s+(\d+)\s+Device\s+(\d+):.*')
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/platforms/android/adb.py
```diff
@@ -378,7 +379,10 @@ def start_cuttlefish_device(use_kernel=False):
     launch_cvd_command_line += (
         f' -kernel_path={kernel_path} -initramfs_path={initramfs_path}')
 
-  execute_command(launch_cvd_command_line, on_cuttlefish_host=True)
+  execute_command(
+      launch_cvd_command_line,
+      timeout=LAUNCH_CVD_TIMEOUT,
```
oliverchang, line 384: This just causes the `execute_command` to return None (https://github.com/google/clusterfuzz/blob/148f3e3a17e7f5223c5957137c1f48e557bbaedc/src/clusterfuzz/_internal/platforms/android/adb.py#L115), without adding any explicit retry. 

Should we be explicitly retrying instead here?

MHA3: I discussed it with Cory. We have manually killed this process in the past which let Clusterfuzz proceed normally. So, just plain killing this process should be okay for us.

MHA3: Clusterfuzz tries to reboot the device once the process is manually killed, and the second time it works.

oliverchang (approved):

PR #2926, opened 2023-01-18
Fix exception with additional environment variables.

oliverchang:


oliverchang:
@catenacyber FYI.

catenacyber:
Thank you very much

catenacyber:
It now works :-)

PR #2928, opened 2023-01-24
Use RegEx to split unintended concatenated lines in stacktrace

Alan32Liu:
Missing `\n` chars in stacktrace will concatenate adjacent lines into one, which sometimes misleads the stracktrace parser to ignore important error messages.
This PR captures one such case and split the lines with RegEx.
It also comes with an obfuscated testcase for this.

### DIFF
--- a/src/clusterfuzz/stacktraces/__init__.py
+++ b/src/clusterfuzz/stacktraces/__init__.py
@@ -382,7 +382,6 @@ class StackParser:
     """Split stacktrace by line, and handle special cases with regex."""
     stacktrace = re.sub(CONCATENATED_SAN_DEADLYSIGNAL_REGEX,
                         SPLIT_CONCATENATED_SAN_DEADLYSIGNAL_REGEX, stacktrace)
-    print(stacktrace)
     return stacktrace.splitlines()
 
   def parse(self, stacktrace: str) -> CrashInfo:
--- a/src/clusterfuzz/stacktraces/__init__.py
+++ b/src/clusterfuzz/stacktraces/__init__.py
@@ -377,6 +377,14 @@ class StackParser:
           state.lkl_kernel_build_id = match.group(2)
     return result
 
+  @staticmethod
+  def split_stacktrace(stacktrace: str):
+    """Split stacktrace by line, and handle special cases with regex."""
+    stacktrace = re.sub(CONCATENATED_SAN_DEADLYSIGNAL_REGEX,
+                        SPLIT_CONCATENATED_SAN_DEADLYSIGNAL_REGEX, stacktrace)
+    print(stacktrace)
+    return stacktrace.splitlines()
+
   def parse(self, stacktrace: str) -> CrashInfo:
     """Parse a stacktrace."""
     state = CrashInfo()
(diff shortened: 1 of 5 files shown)


src/clusterfuzz/stacktraces/__init__.py
```diff
@@ -377,6 +377,14 @@ def remove_lkl_kernel_times_and_set_params(self, state, crash_data):
           state.lkl_kernel_build_id = match.group(2)
     return result
 
+  @staticmethod
+  def split_stacktrace(stacktrace: str):
+    """Split stacktrace by line, and handle special cases with regex."""
+    stacktrace = re.sub(CONCATENATED_SAN_DEADLYSIGNAL_REGEX,
+                        SPLIT_CONCATENATED_SAN_DEADLYSIGNAL_REGEX, stacktrace)
+    print(stacktrace)
```
oliverchang, line 385: nit: remove print.

src/clusterfuzz/stacktraces/__init__.py
```diff
@@ -377,6 +377,14 @@ def remove_lkl_kernel_times_and_set_params(self, state, crash_data):
           state.lkl_kernel_build_id = match.group(2)
     return result
 
+  @staticmethod
+  def split_stacktrace(stacktrace: str):
+    """Split stacktrace by line, and handle special cases with regex."""
+    stacktrace = re.sub(CONCATENATED_SAN_DEADLYSIGNAL_REGEX,
```
oliverchang, line 383: Add a comment here explaining what this is doing. i.e. handling malformed data where there is a missing newline between the actual crash info and this line.

src/clusterfuzz/stacktraces/constants.py
```diff
@@ -185,7 +185,11 @@
     r'.*Sanitizer CHECK failed[:]\s*[^ ]*\s*(.*)')
 SAN_CRASH_TYPE_ADDRESS_REGEX = re.compile(
     r'[ ]*([^ ]*|Atomic [^ ]*) of size ([^ ]*) at ([^ ]*)')
-SAN_DEADLYSIGNAL_REGEX = re.compile(r'.*:DEADLYSIGNAL')
+SAN_DEADLYSIGNAL_REGEX = re.compile(
+    r'(Address|Leak|Memory|UndefinedBehavior|Thread)Sanitizer:DEADLYSIGNAL')
+CONCATENATED_SAN_DEADLYSIGNAL_REGEX = re.compile(
+    r'\n(.+)(' + SAN_DEADLYSIGNAL_REGEX.pattern + r')\n')
```
oliverchang, line 191: To make this more robust, can we match on any non-space characters? i.e. `[^\s]` instad of `.`.

oliverchang (approved):
nice, thanks!

src/clusterfuzz/stacktraces/constants.py
Alan32Liu: Good point, thanks!
I am using `[^\n]*\S[^\n]*` now to ensure there is a least one non-space char and none of the rest is a newline char.
It is a few seconds slower on testcase `test_syzkaller_kasan` (which has a large test data file) but I hope that is acceptable?

oliverchang (approved):

PR #2933, opened 2023-01-25
Add rule to capture suffixed glibc-like assertion failure

Alan32Liu:
1. Add a new rule to capture the assertion error 'optional operator called on a disengaged value'
2. Add a testcase for this new rule.

### DIFF
--- a/src/clusterfuzz/_internal/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
+++ b/src/clusterfuzz/_internal/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
@@ -628,8 +628,8 @@ class StackAnalyzerTestcase(unittest.TestCase):
                                   expected_state, expected_stacktrace,
                                   expected_security_flag)
 
-  def test_assert_disengaged_value(self):
-    """Test the disengaged value assertion failure format."""
+  def test_assert_glibc_suffixed(self):
+    """Test the glibc-like assertion failure format but with suffix."""
     environment.set_value('ASSERTS_HAVE_SECURITY_IMPLICATION', False)
 
     data = self._read_test_data('erroneous_stacktrace.txt')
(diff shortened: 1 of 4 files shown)


src/clusterfuzz/stacktraces/constants.py
```diff
@@ -63,6 +63,10 @@
                                  r'.*assertion failed at\s.*\sin\s*.*: (.*)')
 ASSERT_REGEX_GLIBC = re.compile(
     r'.*:\s*assertion [`\'"]?(.*?)[`\'"]? failed\.?$', re.IGNORECASE)
+ASSERT_DISENGAGED_VALUE = re.compile(
+    r'.*\S.*\/.*:\d+:\s*assertion .* failed:\s*'
+    r'(optional operator.* called on a disengaged value)'
```
oliverchang, line 68: This part is a little too specific. Can we make this match on anything that comes after the "assertion failed" part?

oliverchang (commented):
Thanks! Can you also edit the PR title to make it more clear what this is doing?

src/clusterfuzz/stacktraces/constants.py
Alan32Liu: Ah sure!
I was about to ask you about that as I noticed there are similar errors like this but with a different message.
Fixing that now.

oliverchang: Yes, we should be able to catch any kind of asserts that follow this format. Not individual ones part of a particular project/code base :)

Alan32Liu: I will change the variable/constant names to be less specific about disengaged value, too.

src/clusterfuzz/_internal/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -628,8 +628,8 @@ def test_assert(self):
                                   expected_state, expected_stacktrace,
                                   expected_security_flag)
 
-  def test_assert_disengaged_value(self):
-    """Test the disengaged value assertion failure format."""
+  def test_assert_glibc_suffixed(self):
+    """Test the glibc-like assertion failure format but with suffix."""
```
oliverchang, line 632: What does the "suffix" mean in this case? Can you give an example in a comment next to the regex?

oliverchang (approved):
thanks!!!

PR #2935, opened 2023-01-25
Do not start health_check_responder in Android

marktefftech:
Androids's multiple instances per host model isn't supported yet.

### DIFF
--- a/src/clusterfuzz/_internal/system/environment.py
+++ b/src/clusterfuzz/_internal/system/environment.py
@@ -1096,7 +1096,7 @@ def is_android_kernel(plt=None):
 
 def is_android_real_device():
   """Return True if we are on a real android device."""
-  return platorm() == 'ANDROID'
+  return platform() == 'ANDROID'
 
 
 def is_lib():
--- a/src/clusterfuzz/_internal/system/environment.py
+++ b/src/clusterfuzz/_internal/system/environment.py
@@ -1093,9 +1093,11 @@ def is_android_kernel(plt=None):
   """Return True if we are on android kernel platform groups."""
   return 'ANDROID_KERNEL' in (plt or get_platform_group())
 
+
 def is_android_real_device():
   """Return True if we are on a real android device."""
   return platorm() == 'ANDROID'
+
 
 def is_lib():
   """Whether or not we're in libClusterFuzz."""
--- a/src/clusterfuzz/_internal/system/environment.py
+++ b/src/clusterfuzz/_internal/system/environment.py
@@ -1093,6 +1093,9 @@ def is_android_kernel(plt=None):
   """Return True if we are on android kernel platform groups."""
   return 'ANDROID_KERNEL' in (plt or get_platform_group())
 
+def is_android_real_device():
+  """Return True if we are on a real android device."""
+  return platorm() == 'ANDROID'
 
 def is_lib():
   """Whether or not we're in libClusterFuzz."""
--- a/src/python/bot/startup/health_check_responder.py
+++ b/src/python/bot/startup/health_check_responder.py
@@ -20,7 +20,8 @@ from http.server import HTTPServer
 import threading
 
 from clusterfuzz._internal.base import utils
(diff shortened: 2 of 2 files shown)


src/python/bot/startup/health_check_responder.py
```diff
@@ -46,7 +46,7 @@ def do_GET(self):  # pylint: disable=invalid-name
 
 def run_server():
   """Start a HTTP server to respond to the health checker."""
-  if utils.is_oss_fuzz():
+  if utils.is_oss_fuzz() or environment.is_android():
     # OSS-Fuzz's multiple instances per host model isn't supported yet.
```
jonathanmetzman, line 50: Update the comment as well please.

marktefftech:
`--force-single-line-imports` caused failed `Run basic tests`

marktefftech:
`Run basic tests` fails due to `--force-single-line-imports`

src/python/bot/startup/health_check_responder.py
```diff
@@ -20,7 +20,7 @@
 import threading
 
 from clusterfuzz._internal.base import utils
-from clusterfuzz._internal.system import process_handler
+from clusterfuzz._internal.system import process_handler, environment
```
jonathanmetzman, line 23: Change to
```
from clusterfuzz._internal.system import process_handler
```

src/python/bot/startup/health_check_responder.py
```diff
@@ -46,8 +47,9 @@ def do_GET(self):  # pylint: disable=invalid-name
 
 def run_server():
   """Start a HTTP server to respond to the health checker."""
-  if utils.is_oss_fuzz():
-    # OSS-Fuzz's multiple instances per host model isn't supported yet.
+  if utils.is_oss_fuzz() or environment.is_android():
```
oliverchang, line 50: This may be a little too broad, and capture cases (e.g. Cuttlefish) where there is a single host per "device". 

Would it be possible to have this be narrower here?

marktefftech: Hi @oliverchang thanks for the response..

`environment.is_android()` returns the env variable set for `OS_OVERRIDE`.

In the case of cuttlefish this is set to `ANDROID_X86`:

from `_internal/system/environment.py`:

```
def is_android(plt=None):
  """Return True if we are on android platform."""
  return 'ANDROID' in (plt or platform())


def is_android_cuttlefish(plt=None):
  """Return True if we are on android cuttlefish platform."""
  return 'ANDROID_X86' in (plt or platform())


def is_android_kernel(plt=None):
  """Return True if we are on android kernel platform groups."""
  return 'ANDROID_KERNEL' in (plt or get_platform_group())
```

I believe this means only "Real" devices are detected by `is_android`.

Would changing `is_android()` to `is_android_real_device()` address this concern?

oliverchang: the `in` operator would cause the `'ANDROID' in 'ANDROID_X86'` to match, so `is_android` would return True for cuttlefish devices too. 

A new `is_android_real_device()` which explicitly excludes cuttlefish SGTM.

marktefftech: Done

oliverchang (approved):

src/clusterfuzz/_internal/system/environment.py
```diff
@@ -1094,6 +1094,11 @@ def is_android_kernel(plt=None):
   return 'ANDROID_KERNEL' in (plt or get_platform_group())
 
 
+def is_android_real_device():
+  """Return True if we are on a real android device."""
+  return platorm() == 'ANDROID'
```
oliverchang, line 1099: Please fix the typo/tests, thanks!

oliverchang (changes requested):

src/clusterfuzz/_internal/system/environment.py
oliverchang: Actually I just did this.

PR #2947, opened 2023-02-13
Install recent Node.js in ClusterFuzz images.

oliverchang:
This is needed for Jazzer.js, and matches our OSS-Fuzz images.

### DIFF
--- a/docker/base/setup_19.x
+++ b/docker/base/setup_19.x
@@ -1,0 +1,368 @@
+#!/bin/bash
+
+# Discussion, issues and change requests at:
+#   https://github.com/nodesource/distributions
+#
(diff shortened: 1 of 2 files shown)


docker/base/setup_19.x
```diff
@@ -0,0 +1,368 @@
+#!/bin/bash
```
oliverchang, line 1: Included this file here to avoid the antipattern of piping curl output to bash.

Alan32Liu (approved):

docker/base/setup_19.x
jonathanmetzman: Does this need to be in a third_party directory?

jonathanmetzman (commented):
lgtm

PR #2956, opened 2023-02-28
Support mapping crash types to labels

alpire:
This PR adds the ability to tag crashes with a labels based on the crash type. This will enable users to categorize crashes better if they so choose. For instance, they might want to apply the same label to all arithmetic overflow crashes.


oliverchang (approved):
LGTM, thanks!

PR #2962, opened 2023-02-28
Added device resets for adb recovery state in android emulator platform.

davfu1:
The android_emulator platform used for Trusty device fuzzing frequently sends devices into adb recovery state. Clusterfuzz does not currently detect or reset devices in adb recovery state so this code adds device reset logic for android_emulator.

### DIFF
--- a/src/clusterfuzz/_internal/platforms/android/adb.py
+++ b/src/clusterfuzz/_internal/platforms/android/adb.py
@@ -338,6 +338,18 @@ def hard_reset():
   soft_reset_cmd = get_adb_command_line('reboot')
   execute_command(soft_reset_cmd, timeout=RECOVERY_CMD_TIMEOUT)
 
+  if environment.is_android_emulator():
+    #For recovery state
+    logs.log('Platform ANDROID_EMULATOR detected.')
+    restart_adb()
+    state = get_device_state()
+
+    logs.log('Device state is: %s' % state)
+    if state == 'recovery':
+      logs.log('Rebooting recovery state device with --wipe_data.')
+      run_command('root')
+      run_shell_command('recovery --wipe_data')
+
 
 def kill_processes_and_children_matching_name(process_name):
   """Kills process along with children matching names."""
@@ -434,6 +446,11 @@ def remount():
   run_command('remount')
   wait_for_device()
   run_as_root()
+
+
+def restart_adb():
+  run_command('kill-server')
+  run_command('start-server')
 
 
 def remove_directory(device_directory, recreate=False):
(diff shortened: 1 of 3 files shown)


src/clusterfuzz/_internal/platforms/android/adb.py
```diff
@@ -338,6 +338,18 @@ def hard_reset():
   soft_reset_cmd = get_adb_command_line('reboot')
   execute_command(soft_reset_cmd, timeout=RECOVERY_CMD_TIMEOUT)
 
+  if environment.is_android_emulator():
+    #For recovery state
```
oliverchang, line 342: Can you please format this comment a bit better by adding a space after the "#" and making it a more complete sentence? e.g. `# Handle recovery states for emulator by doing a complete wipe`.

davfu1: Would deleting the comment altogether also work? I was thinking the logging messages and is_android_emulator() call might adequately explain to a new reader what the submitted lines are doing.

oliverchang: sounds good. the log messages seem self explanatory enough.

src/clusterfuzz/_internal/system/environment.py
```diff
@@ -1089,6 +1089,11 @@ def is_android_cuttlefish(plt=None):
   return 'ANDROID_X86' in (plt or platform())
 
 
+def is_android_emulator(plt=None):
+  """Return True if we are on android emulator platform."""
+  return 'ANDROID_EMULATOR' in (plt or get_platform_group())
```
jonathanmetzman, line 1094: I think there's a footgun in here:
```
>>> 'x' in ('y' or 'x')
False
>>> ('y' or 'x')
'y'
```
I think it works if you only want to use get_platform_group() when plt is not passed, but I don't like this kind of magic.
 Let's get rid of the plt behavior since it's not even used anywhere.

davfu1: Wow, great catch - it sounds like this should be changed to just `'ANDROID_EMULATOR' in get_platform_group()` then. Does this mean line 1094 on Android Kernel should be changed also, as well as other environments on 1084 and 1089 using `(plt or platform())`?

jonathanmetzman: To clarify my question: I think this pattern is too clever and should be gotten rid of. But I'm not sure if the original code is correct or not.

oliverchang: This was indeed a convention. The intention of `plt` is for it to be an override that's considered if it's set at all. 

That said, +1 to removing that for this function since it's not used.

oliverchang (approved):

PR #2963, opened 2023-03-01
Fix: enabling jira token auth

andrei-near:
Jira Cloud [no longer supports](https://jira.readthedocs.io/examples.html#authentication) username:password authentification.
This is to enable the use of auth token in issue tracker manager


andrei-near:
I'm at bug #3 in jira issue tracker now. Will publish another PR

PR #2966, opened 2023-03-03
Fix: Jira Issue Tracker fixes

andrei-near:
Jira module version bump to latest supported with python3.7
Jira Token auth instead of user/password
Jira issue tracker example configuration in docs
Fixes in Jira Issue tracker code
Additional AppEngine logs to make debugging easier
Cosmetic change in testcase details page to show Jira issue key instead of ID


src/appengine/libs/issue_management/issue_filer.py
```diff
@@ -334,6 +334,8 @@ def file_issue(testcase,
       is_security=testcase.security_flag, is_crash=is_crash)
 
   issue = issue_tracker.new_issue()
+  logs.log(f"Instantiated a new issue object with ID: {issue.id}")
```
oliverchang, line 337: Please use single quotes for strings.

src/appengine/libs/issue_management/issue_filer.py
```diff
@@ -465,6 +467,7 @@ def file_issue(testcase,
   issue.reporter = user_email
 
   recovered_exception = None
+
```
oliverchang, line 470: Can you please limit the scope of the changes and exclude whitespace changes like these?

configs/test/issue_trackers/config.yaml
```diff
@@ -19,7 +19,102 @@
 #   - %SANITIZER%: The sanitizers associated with the testcase.
 #   - %SEVERITY%: The security severity of the testcase.
 #   - %PLATFORM%: The platform the testcase was found on.
-test-project:
+test-project-jira:
```
oliverchang, line 22: Are these changes required? 

I don't see any test changes that require this, and we typically wouldn't include them here either -- we prefer to inject these/mock this in the test setup instead.

src/appengine/private/components/testcase-detail/testcase-detail.html
```diff
@@ -47,7 +47,8 @@
         max-width: 200px;
```
oliverchang, line 47: Are the changes in this file required?

oliverchang (changes requested):
hi, thanks for your contribution, and sorry for the delay in reviewing.  

please see my comments below

configs/test/issue_trackers/config.yaml
andrei-near: Added this as an example for other users. Renaming back test-project-monorail to test-project to avoid breaking any tests

src/appengine/libs/issue_management/issue_filer.py
andrei-near: Got it

src/appengine/libs/issue_management/issue_filer.py
andrei-near: it my fmt, removing

src/appengine/private/components/testcase-detail/testcase-detail.html
andrei-near: formatter again, clearing

andrei-near:
Hi @oliverchang . Thanks for review. Addressed the comments.

andrei-near:
@oliverchang  can you review this please?

PR #2974, opened 2023-03-20
Added prepending of Trusted App (TA) stacktrace for Trusty fuzzing

davfu1:
Crashes from Trusty fuzzing sometimes produce a highly relevant Trusted App (TA) stacktrace in the logcat. This PR finds and adds the relevant logcat entry to the beginning of engine output to distinguish the TA stacktrace in testcase reporting. Additionally, upcoming PRs adding TA stacktrace analysis may refer to this entry to help identify relevant frames.


src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py
```diff
@@ -1079,6 +1100,8 @@ def run_single_testcase(self,
       result = LibFuzzerCommon.run_single_testcase(self, device_testcase_path,
                                                    timeout, additional_args)
       result.output = self._append_logcat_output_if_needed(result.output)
+      if environment.is_android_emulator():
```
oliverchang, line 1103: nit: for consistency with `_append_logcat_output_if_needed`, can we move the if into the function and call it `_prepend_trusty_stacktrace_if_needed`. 

alternatively, would it make sense to instead generalise  `_append_logcat_output_if_needed` to something like `_add_logcat_output_if_needed` to both do the prepend and append?

src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py
```diff
@@ -941,6 +941,24 @@ def _append_logcat_output_if_needed(self, output):
     return '{output}\n\nLogcat:\n{logcat_output}'.format(
         output=output, logcat_output=android.logger.log_output())
 
+  def _prepend_trusty_stacktrace(self, output):
+    """Add trusty stacktrace to beginning of output if found in logcat."""
+    logcat = android.logger.log_output()
+    begin, end = '---------', 'Built:'
```
oliverchang, line 947: Do we expect these markers to be very stable?

oliverchang (approved):
LGTM generally with some minor suggestions.

PR #2976, opened 2023-03-21
Added prepending of Trusted App (TA) stacktrace for Trusty fuzzing

davfu1:
Crashes from Trusty fuzzing sometimes produce a highly relevant Trusted App (TA) stacktrace in the logcat. This PR finds and adds the relevant logcat entry to the beginning of engine output to distinguish the TA stacktrace in testcase reporting. Additionally, upcoming PRs adding TA stacktrace analysis may refer to this entry to help identify relevant frames.

PR #2979, opened 2023-03-27
Enabled Trusty binaries download and stacktrace symbolization

davfu1:
Trusty App (TA) fuzzing sometimes produces a separate, highly-relevant stacktrace in the logcat. This TA stacktrace refers to addresses in two different binaries (kernel and userspace). This PR downloads these two binaries and uses addr2line to generate debug line references for use by crash analyzers and the stacktrace prepended by https://github.com/google/clusterfuzz/pull/2976.


src/clusterfuzz/_internal/crash_analysis/stack_parsing/stack_symbolizer.py
```diff
@@ -525,6 +528,61 @@ def _lkl_line_parser(self, line):
...
+  def _trusty_line_parser(self, line):
+    """Parses line for memory_space, offset, and addr"""
+    match = STACK_TRACE_LINE_TRUSTY.match(line)
+    if match:
+      memory_space, offset, addr = match.groups()
+      return memory_space, offset, addr
+
+    return None, None, None
+
+  def _extract_trusty_app_name(self, stacktrace):
```
oliverchang, line 540: Nit: please add docstrings for all of these methods.

src/clusterfuzz/_internal/crash_analysis/stack_parsing/stack_symbolizer.py
```diff
@@ -525,6 +528,61 @@ def _lkl_line_parser(self, line):
...
+
+    return ''
+
+  def _close_pipes(self):
+    for pipe in pipes:
+      pipe.stdin.close()
+      pipe.stdout.close()
+      pipe.kill()
+
+  def process_trusty_stacktrace(self, unsymbolized_crash_stacktrace):
```
oliverchang, line 554: Are there any tests you can write for this?

src/clusterfuzz/_internal/platforms/android/symbols_downloader.py
```diff
@@ -179,6 +180,35 @@ def download_system_symbols_if_needed(symbols_directory):
   utils.write_data_to_file(build_params, build_params_check_path)
 
 
+def download_trusty_symbols_if_needed(symbols_directory, app_name):
+  """Downloads and extracts Trusted App ELF files"""
+  ab_target = ''
+  device = settings.get_build_parameters().get('target')
+  if device in ['cheetah', 'panther']:
```
oliverchang, line 187: As a future improvement: would it make sense to encode these details as part of the build rather than hardcode things thing?

src/clusterfuzz/_internal/crash_analysis/stack_parsing/stack_symbolizer.py
davfu1: Done - the docstrings should be added now.

src/clusterfuzz/_internal/crash_analysis/stack_parsing/stack_symbolizer.py
davfu1: Yes, thanks for the suggestion! I've added a test with some sample stacktrace data now.

src/clusterfuzz/_internal/platforms/android/symbols_downloader.py
davfu1: Yes, I definitely agree! - encoding these details would allow this code to continue working as device types and branch names are added or changed. I'm not sure I can personally make that change, but I can bring this up with Trusty team.

src/clusterfuzz/_internal/crash_analysis/stack_parsing/stack_symbolizer.py
davfu1: Hi @oliverchang - Following up on our discussion, I removed the unit test for now (presubmits are OK without that unit test), with the intent to submit the test in a separate PR to further debug why it could be failing /gcbrun.

oliverchang (approved):

PR #2984, opened 2023-03-30
Added Android device ramdump detection and reset

davfu1:
Crashes from Trusty fuzzers often send Android devices into a ramdump state that prevents any further fuzzing. In such ramdump crashes, the stacktrace is left in the ramdump files and is not currently detected by Clusterfuzz. 

This PR detects and resets devices in ramdump state and is intended act as an entry point for future changes to extract ramdump data to generate a new crash object.


davfu1:
(closing this pull request for now because I realized ramdump extraction might best done after AndroidLibfuzzerRunner times out during fuzz(); and if so the device should stay in OEM ramdump mode until that occurs)

Issue #2973, opened 2023-03-19
Detect implicit conversion errors


nwellnhof:
We now fuzz libxml2 with `-fsanitize=integer` which also enables `implicit-conversion` checks. This group contains the following checks:

- implicit-signed-integer-truncation
- implicit-unsigned-integer-truncation
- implicit-integer-sign-change

Implicit conversion errors are shown as "UNKNOWN" in OSS-Fuzz. It seems that a regex has to be added to detect them. A typical error message looks like

    numbers.c:186:21: runtime error: implicit conversion from type 'int' of value -2147483600 (32-bit, signed) to type 'xmlChar' (aka 'unsigned char') changed the value to 48 (8-bit, unsigned)

or

    HTMLparser.c:483:17: runtime error: implicit conversion from type 'long' of value -1 (64-bit, signed) to type 'size_t' (aka 'unsigned long') changed the value to 18446744073709551615 (64-bit, unsigned)

Both examples are `implicit-integer-sign-change` errors, but I think truncation errors have the same format.

PR #2985, opened 2023-04-03
Add implicit conversion crash

hogo6002:
Resolve issue [#2973 ](https://github.com/google/clusterfuzz/issues/2973)


src/clusterfuzz/_internal/crash_analysis/crash_analyzer.py
```diff
@@ -101,6 +101,7 @@
     # math of crpto libraries.
     'Unsigned-integer-overflow',
     'Unreachable code',
+    'Implicit-conversion',
```
Alan32Liu, line 104: Do we need to maintain an alphabetical order of this list?

oliverchang: Yes good catch! let's maintain it to enable easier searching by humans.

Alan32Liu (approved):

oliverchang (approved):

PR #2987, opened 2023-04-03
Added Trusty ramdump logcat extraction

davfu1:
Trusty fuzzing sometimes causes devices to crash into ramdump state. This PR enables Clusterfuzz to download `kernel.log` from ramdumps and to desymbolize it into a logcat using the `ramoops_reader.py` script. The desymbolized logcat will contain a Kernel Panic stacktrace. This PR adds the KP stacktrace and logcat to the crash output, and resets the device.


oliverchang (approved):
LGTM. Have you had a chance to dig into the tests issue you were runnign into btw?

davfu1:
Thanks for the review Oliver! On that particular unit test for https://github.com/google/clusterfuzz/pull/2979, I didn't have the chance to make a further investigation beyond what I had tried for that PR yet. I have one more Trusty stacktrace-related change (crash state parsing, which should have its own unit test in `stack_analyzer_test.py`) and was hoping to dig in after that one.

PR #2992, opened 2023-04-06
Keep Centipede error matching consistent with its test cases

Alan32Liu:
1. Update `Centipede`'s error RegEx.
2. More thorough unittests for `Centipede`
3. Keep `Centipede`'s unittests consistent with its own testcases.
4. Remove redundant code in `Centipede`'s unitests.


src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
```diff
@@ -26,20 +26,18 @@
...
 _DEFAULT_ARGUMENTS = [
     '--exit_on_crash=1',
-    f'--timeout={_TIMEOUT}',
     f'--fork_server={_SERVER_COUNT}',
     f'--rss_limit_mb={_RSS_LIMIT}',
     f'--address_space_limit_mb={_ADDRESS_SPACE_LIMIT}',
 ]
 
-_CRASH_REGEX = re.compile(r'[sS]aving input to:? [\n]?(.*)')
+_CRASH_REGEX = re.compile(r'Saving input to: (.*)')
```
oliverchang, line 40: Should we keep support for the old format to ensure more compatibility over older builds?

oliverchang (approved):
thanks! just one comment

oliverchang:
actually, looks like there are some failing tests. can you please take a look? For these integration tests ones, make sure you're building the binaries using the oss-fuzz builder images.

Alan32Liu:
> actually, looks like there are some failing tests. can you please take a look? For these integration tests ones, make sure you're building the binaries using the oss-fuzz builder images.

Sure, the error from basic test is caused by the old stacktrace sample text, I am working on that now.
I will also build and use the binaries from OSS-Fuzz.

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
Alan32Liu: I can certainly make it more backward-compatible.
In which case would older builds be used?

oliverchang: e.g. when we do our regression (bisection) task. We run older builds to determine which commit range introduced the vulnerability.

Alan32Liu: Ah, I see, thanks!
I will try to make it compatible with all previous build, then : )

PR #2993, opened 2023-04-06
Added stacktrace regex for Trusty crashes

davfu1:
Stacktraces from Trusty fuzzing have unique markers not recognized by the current regex in `/stacktraces/`. This PR enables StackParser to recognize key post-symbolized lines from Trusty stacktraces and to create a crash state from function names for crash deduplication and bucketing.


src/clusterfuzz/stacktraces/__init__.py
```diff
@@ -646,6 +648,10 @@ def parse(self, stacktrace: str) -> CrashInfo:
           state.frame_count = 0
           continue
 
+      #Trusty stacktraces
```
oliverchang, line 651: nit: this comment is not very useful. let's just remove it.

oliverchang (approved):
LGTM with a nit!

PR #2999, opened 2023-04-12
More tests for centipede backcomp

Alan32Liu:
1. More resilient RegEx to capture crash reproducers.
2. Directly run and test `Centipede`'s old binary with the same unittests as the latest binary, ensuring the engine is compatible with both versions.


src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
```diff
@@ -30,15 +30,15 @@
...
+_TIMEOUT_PER_INPUT = 60
 _DEFAULT_ARGUMENTS = [
     '--exit_on_crash=1',
     f'--fork_server={_SERVER_COUNT}',
     f'--rss_limit_mb={_RSS_LIMIT}',
     f'--address_space_limit_mb={_ADDRESS_SPACE_LIMIT}',
 ]
 
-_CRASH_REGEX = re.compile(r'[sS]aving input to:? [\n]?(.*)')
+CRASH_REGEX = re.compile(r'[sS]aving input to:?\s*[\n]?(.*)')
```
oliverchang, line 41: As discussed, let's remove this `\n` bit since it's redundant.

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
```diff
@@ -114,8 +114,6 @@ def prepare(self, corpus_dir, target_path, build_dir):
     else:
       logs.log_warn('Unable to find sanitized target binary.')
 
-    arguments.append(f'--timeout_per_input={_TIMEOUT_PER_INPUT}')
```
oliverchang, line 117: As discussed, let's bring this back, with the different timeouts for fuzzing vs reproduction.

oliverchang (commented):
thanks!

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
Alan32Liu: Yep, I will also change the `centipede-old` binary to the second most recent version to adapt to this change.

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
oliverchang: Did you end up making this change? Or doing this in a future PR?

oliverchang (approved):

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
Alan32Liu: Just pushed it.
I thought I pushed it earlier, but I was apparently wrong.

Thanks for the reminder!

Issue #2975, opened 2023-03-21
Implement proxy for firebase auth 


oliverchang:
https://firebase.google.com/docs/auth/web/redirect-best-practices#proxy-requests

This should solve #2818

MohitBansal321:
Hey can you provide more context on it

Issue #2818, opened 2022-09-13
Can't log in to oss-fuzz.com with Safari/macOS


nwellnhof:
When trying to log in to oss-fuzz.com with my Google account using Safari on macOS, I'm only seeing a box with a loading animation and nothing happens.

Safari Version 16.0 (17614.1.25.9.10, 17614)
macOS Monterey 12.6

oliverchang:
@Alan32Liu or @Navidem can one of you help with this some time soon? 

Per https://github.com/google/clusterfuzz/issues/2818#issuecomment-1266550824 this is likely related to a missing CSP rule which we set up here: https://github.com/google/clusterfuzz/blob/1be95bd84e0868495a2b3c835b18577b2ae3e0e2/src/appengine/libs/csp.py

I don't have a macOS machine to debug this sadly.

oliverchang:
@Navidem can you please help with this? I'm not sure why this is only an issue on Safari.

Navidem:
> @Navidem can you please help with this? I'm not sure why this is only an issue on Safari.

Sure, I will.

Navidem:
I can confirm adding `builder.add('img-src', 'data:')` makes the CSP error disappear, but login issue persists.

oliverchang:
@Navidem can you check if the workaround in https://github.com/firebase/firebaseui-web/issues/977#issuecomment-1291567278 works?

Navidem:
> firebase/firebaseui-web#977 (comment)

The solution did not help: at the time of login, a popup appears to select account but the console log show following error and UI returns back to the login page:

```
[Error] Unhandled Promise Rejection: TypeError: Right side of assignment cannot be destructured
	(anonymous function) (firebase-auth-compat.js:1:86987)
	asyncFunctionResume
	sendToConsumer (firebase-auth-compat.js:1:92175)
	(anonymous function) (firebase-auth-compat.js:1:91705)
	forEach
	onEvent (firebase-auth-compat.js:1:91645)
	(anonymous function) (firebase-auth-compat.js:1:98111)
	mm (cb=gapi.loaded_0:269:97)
	(anonymous function) (cb=gapi.loaded_0:269:268)
	Ek (cb=gapi.loaded_0:246:313)
	Nk (cb=gapi.loaded_0:246:475)
	hk (cb=gapi.loaded_0:247)
	(anonymous function) (cb=gapi.loaded_0:240:374)
	(anonymous function)
```

[...]

PR #3000, opened 2023-04-13
Implement proxy for firebase auth

hogo6002:
Resolve issue [Can't log in to oss-fuzz.com with Safari/macOS](https://github.com/google/clusterfuzz/issues/2818) and [#2975](https://github.com/google/clusterfuzz/issues/2975)


src/appengine/server.py
```diff
@@ -159,6 +160,11 @@ def register_routes(flask_app, routes):
 
 handlers = [
     ('/', home.Handler if _is_oss_fuzz else testcase_list.Handler),
+    ('/__/auth/experiments.js', auth.Handler),
```
oliverchang, line 163: Instead of repeating all of these, can we define a single catchall route?

src/appengine/libs/auth.py
```diff
@@ -45,6 +45,11 @@ def auth_domain():
   return utils.get_application_id() + '.firebaseapp.com'
```
oliverchang, line 45: Let's replace this to raise an exception instead.

src/appengine/handlers/auth.py
```diff
@@ -0,0 +1,44 @@
...
+
+class Handler(base_handler.Handler):
+  """Auth page."""
+
+  @handler.get(handler.HTML)
+  def get(self):
+    """Handle a get request."""
+    target_url = request.url.replace(auth.auth_domain(),
+                                     auth.real_auth_domain(), 1)
+    logging.info("Forwarding auth request to: %s", target_url)
```
oliverchang, line 36: Please use single quotes for strings.

oliverchang (changes requested):
nice! some comments

src/appengine/handlers/auth.py
```diff
@@ -0,0 +1,45 @@
...
+from libs import handler
+
+
+class Handler(base_handler.Handler):
+  """Auth page."""
+
+  # pylint: disable=unused-argument
+  @handler.get(handler.HTML)
+  def get(self, extra=None):
+    """Handle a get request."""
```
oliverchang, line 34:
```python
# We use `request.url` which is already the full URL. 
del  extra
```

See https://google.github.io/styleguide/pyguide.html#214-decision

src/appengine/handlers/auth.py
```diff
@@ -0,0 +1,46 @@
...
+    # We use `request.url` which is already the full URL.
+    del extra
+    target_url = request.url.replace(auth.auth_domain(),
+                                     auth.real_auth_domain(), 1)
+    logging.info('Forwarding auth request to: %s', target_url)
+    response = requests.get(target_url)
+    gzip_response = gzip.compress(response.text.encode('utf-8'))
```
oliverchang, line 40: nit: can we make sure to always set the correct gzip content-encoding header in our own response?

This is to make it a bit more resilient against potential upstream changes. e.g. if Firebase changes their encoding, our endpoint would break beacuse we still expect gzip from upstream.

oliverchang (approved):
nice!! just one final comment to improve this.

PR #3001, opened 2023-04-13
Fixed Crash object creation and device state timeout for Trusty fuzzing

davfu1:
libfuzzer/engine.py requires a crash_testcase_file_path to generate a Crash object. Libfuzzer does not record a filepath in the case of Trusty kernel panics, so so this PR creates an empty filepath to ensure the Crash object is still generated.


src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -312,7 +318,8 @@ def fuzz(self, target_path, options, reproducers_dir, max_time):
     # libFuzzer, this is most likely a startup crash. Use an empty testcase to
     # to store it as a crash.
     if (not crash_testcase_file_path and
-        fuzz_result.return_code not in constants.NONCRASH_RETURN_CODES):
+        fuzz_result.return_code not in constants.NONCRASH_RETURN_CODES
+       ) or self._fuzz_output_contains_trusty_kernel_panic(log_lines):
```
oliverchang, line 322: Can you please add a comment to explain this case? (And to distinguish it from the previous case this was meant for)?

davfu1: Thanks Oliver! - I've added the comment.

oliverchang (approved):

PR #3002, opened 2023-04-14
Use staging assignee over automatic assignee for testcases

davfu1:
Haiku is adding a new 'staging' fuzz config field to indicate that fuzzers are under test and not ready to send bugs for triage. Rather, fuzzers under staging will include in a `.issue_metadata` file, `{"issue_assignee_override":"email@email.com"}` to provide an assignee to replace AUTOMATIC_ASSIGNEE.

### DIFF
--- a/src/appengine/libs/issue_management/issue_filer.py
+++ b/src/appengine/libs/issue_management/issue_filer.py
@@ -391,7 +391,7 @@ def file_issue(testcase,
   if fuzzer_metadata and 'assignee' in fuzzer_metadata:
     issue.status = policy.status('assigned')
     issue.assignee = fuzzer_metadata['assignee']
-    logs.log('Fuzzer testcase has assignee metadata %s.' % issue.assignee)
+    logs.log('Testcase has assignee metadata %s' % issue.assignee)
 
   # Add additional ccs from the job definition and fuzzer.
   ccs = data_handler.get_additional_values_for_variable(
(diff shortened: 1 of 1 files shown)


src/appengine/libs/issue_management/issue_filer.py
```diff
@@ -387,6 +387,13 @@ def file_issue(testcase,
   else:
     issue.status = properties.status
 
+  fuzzer_metadata = testcase.get_metadata('issue_metadata')
+  if fuzzer_metadata and 'issue_assignee_override' in fuzzer_metadata:
```
oliverchang, line 391: nit: can we just use "assignee" as the key name instead? No need for the "issue" and the "override" bits.

src/appengine/libs/issue_management/issue_filer.py
```diff
@@ -387,6 +387,13 @@ def file_issue(testcase,
...
+    issue.status = policy.status('assigned')
+    issue.assignee = fuzzer_metadata['issue_assignee_override']
+    logs.log(
+        'Fuzzer is under staging. Assigned testcase to %s' % issue.assignee)
```
oliverchang, line 395: nit: Can you make this message a bit more generic? The "staging" bit is a bit of an implementation detail for Android.

e.g. "Testcase has assignee metadata (%s)".

src/appengine/libs/issue_management/issue_filer.py
davfu1: Thanks Oliver for these suggestions! The latest changes use 'assignee' and improved the log message.

oliverchang (approved):

Issue #3008, opened 2023-04-21
Infra consolidation metabug


jonathanmetzman:
Tasks
- [ ] Fuzz
- [x] Symbolize
- [x] Minimize
- [x] Progression
- [x] Regression
- [x] Corpus Pruning
- [x] Analyze
- [x] Variant

PR #3013, opened 2023-04-25
Stop keeping track of whether testcases are binary

jonathanmetzman:
It can be done on the fly (only purpose is determining whether to skip manual minimization (and I'm not sure this is actually a valid reason). And the db column is never read only written.


jonathanmetzman:

oliverchang (approved):

PR #3015, opened 2023-04-26
[NFC] Fix variable name

jonathanmetzman:


Alan32Liu (approved):
Just out of curiosity: Is there any naming conventions require prefixing function parameters with `_`? I thought that normally indicates private class functions/attributes, but not sure about function parameters.

jonathanmetzman:
> Just out of curiosity: Is there any naming conventions require prefixing function parameters with `_`? I thought that normally indicates private class functions/attributes, but not sure about function parameters.

I'm not aware of one for function params either. pylint seems to dislike it.

PR #3017, opened 2023-04-26
[NFC][six] Get rid of six

jonathanmetzman:
Python2 isn't supported anymore.


Alan32Liu (approved):
LGTM, most of the lint failures are
```
.pylintrc:1:0: E0015: Unrecognized option found: bad-functions, no-space-check (unrecognized-option)
```
which is irrelevant from this PR.
Would it be better to solve it here or leave it for another PR?

PR #3025, opened 2023-05-03
Edit proto file instead of generated

jonathanmetzman:
Also fix some linting issues.

### DIFF
--- a/src/clusterfuzz/_internal/protos/untrusted_runner.proto
+++ b/src/clusterfuzz/_internal/protos/untrusted_runner.proto
@@ -229,9 +229,7 @@ message StatRequest {
 
 message CrossPollinationStats {
   optional string project_qualified_name = 1;
-  optional string method = 2;
   optional string sources = 3;
-  optional string tags = 4;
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
I think in https://github.com/google/clusterfuzz/commit/9877ff3533a20c064c2436a2463810fa6ac22ece I only edited the generated python files, not the proto files.

oliverchang (approved):

src/clusterfuzz/_internal/protos/untrusted_runner.proto
```diff
@@ -229,9 +229,7 @@ message StatRequest {
 
 message CrossPollinationStats {
   optional string project_qualified_name = 1;
-  optional string method = 2;
```
jonathanmetzman, line 232: @oliverchang did i do this correctly?
I reran the generate.sh script after this edit by the way.

oliverchang: LGTM.

PR #3033, opened 2023-05-08
[rearchitecture] Fix run_bot/butler

jonathanmetzman:
Print stacktraces properly.
Propogate errors so that butler command
fails when run_bot and other commands fails.
Related: #3008

### DIFF
--- a/butler.py
+++ b/butler.py
@@ -55,7 +55,7 @@ def _setup_args_for_remote(parser):
       '-i',
       '--instance-name',
       required=True,
-      help=('The instance name (e.g. clusterfuzz-linux-0005).'))
+      help='The instance name (e.g. clusterfuzz-linux-0005).')
   parser.add_argument('--project', help='The Project ID.')
   parser.add_argument('--zone', help='The Project Zone.')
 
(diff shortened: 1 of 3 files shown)


hogo6002 (approved):

butler.py
```diff
@@ -55,7 +55,7 @@ def _setup_args_for_remote(parser):
       '-i',
       '--instance-name',
       required=True,
-      help=('The instance name (e.g. clusterfuzz-linux-0005).'))
+      help='The instance name (e.g. clusterfuzz-linux-0005).')
```
Alan32Liu, line 58: Very trivial nit: 'e.g.' -> 'e.g.,'.
Probably not worth fixing at all : )

src/local/butler/run_bot.py
```diff
@@ -122,5 +123,8 @@ def _stop_handler(*_):
     signal.signal(signal.SIGTERM, _stop_handler)
     common.process_proc_output(proc)
     proc.wait()
+    exit_code = proc.wait()
```
Alan32Liu, line 126: Does this mean to overwrite the previous line, instead of appending to it?

jonathanmetzman: Nope. that was a mistake.

jonathanmetzman: https://github.com/google/clusterfuzz/pull/3037

PR #3034, opened 2023-05-08
Improve centipede's tests

jonathanmetzman:
Previously centipede's tests write to the source repo, hoping that they can undo it after running.
This is an incorrect assumption, as Ctrl-C or exceptions can prevent this from happening, leading to
modifications being made to the repo during development (I noticed this after centipede and centipede-old kept being modified according to git :-)
Replace it with temporary directories that wont replace tracked files or add files to tracked directories (which makes it impossible to deploy).

### DIFF
--- a/src/clusterfuzz/_internal/tests/core/bot/fuzzers/centipede/centipede_engine_test.py
+++ b/src/clusterfuzz/_internal/tests/core/bot/fuzzers/centipede/centipede_engine_test.py
@@ -13,10 +13,13 @@
 # limitations under the License.
 """Tests for centipede engine."""
 
+import collections
+import contextlib
 import os
-from pathlib import Path
+import pathlib
 import re
 import shutil
+import tempfile
 import unittest
 
 from clusterfuzz._internal.bot.fuzzers import engine_common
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
Doing this to make development easier for ##3008

src/clusterfuzz/_internal/tests/core/bot/fuzzers/centipede/centipede_engine_test.py
```diff
@@ -53,63 +70,46 @@
...
   """Integration tests."""
 
-  def setUp(self):
-    self.maxDiff = None  # pylint: disable=invalid-name
+  def run(self, *args, **kwargs):
     test_helpers.patch_environ(self)
+    with get_test_paths() as test_paths:
+      self.test_paths = test_paths
+      os.environ['BUILD_DIR'] = str(self.test_paths.data)
+      super().run(*args, **kwargs)
```
Alan32Liu, line 106: Not sure if I follow here:
Do we move the preparation steps in `run()` to save time?
Thanks!

Alan32Liu (approved):

Alan32Liu (approved):

src/clusterfuzz/_internal/tests/core/bot/fuzzers/centipede/centipede_engine_test.py
jonathanmetzman: No. We do it so that the testcases are run within the ```with_gets_paths() as test_paths:``` block. We can't do that with setup. We don't 100% need to do this, but this trick saved us a ton of repetition.

PR #3035, opened 2023-05-08
Allow Centipede to fuzz without unsanitized binary

Alan32Liu:
1. `Centipede` can run with a sanitized binary, an unsanitized binary, or both. It raises an error when there neither is provided.
2. Fix import.


src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
```diff
@@ -105,12 +105,25 @@ def prepare(self, corpus_dir, target_path, build_dir):
     # Directory corpus_dir saves the corpus files required by ClusterFuzz.
     arguments.append(f'--corpus_dir={corpus_dir}')
 
-    # The unsanitized binary, Centipede requires it to be the main fuzz target.
-    arguments.append(f'--binary={target_path}')
-
+    # Centipede requires at least one of:
+    # 1. The unsanitized target binary.
+    # 2. The sanitized target binary.
```
jonathanmetzman, line 110: nit: centipede requires X or Y.

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
```diff
@@ -105,12 +105,25 @@ def prepare(self, corpus_dir, target_path, build_dir):
...
+    unsanitized_target_path = pathlib.Path(target_path)
+    unsan_bin_exists = unsanitized_target_path.exists()
```
jonathanmetzman, line 112: Instead of switching to calling it san/unsan, let's be consistent and call it sanitized/unsanitized.

jonathanmetzman (approved):
lgtm

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
Alan32Liu: BTW, I may also need to change to config to always download sanitized binaries to the `extra_build` directory, instead of the default directory (which is for unsanitized binary).
Is it possible to do it in job config, or should I modify the code in clusterfuzz?

Alan32Liu:
I am pretty sure the gcloud build tests failed due to incompatible binaries, and will try to fix it when I have time.

Alan32Liu:
Hi @jonathanmetzman, could you please review this again?
I'm sorry about the repeated work; I made many new changes that I did not expect before the last time I asked for your review.
Thanks!

Alan32Liu:
Merge master to here because of [#3072](https://github.com/google/clusterfuzz/pull/3072).

src/clusterfuzz/_internal/tests/core/bot/fuzzers/centipede/centipede_engine_test.py
```diff
@@ -130,6 +137,20 @@ def _test_reproduce(self,
 
     return re.search(regex, result.output)
 
+  def test_reproduce_uaf_without_unsanitized_binary(self):
```
jonathanmetzman, line 140: Do you have an idea how long these tests add (especially because they are basically doubled in time because of unshare)?

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
```diff
@@ -105,40 +106,73 @@ def prepare(self, corpus_dir, target_path, build_dir):
...
+    # |-------------------------------------------------------|
+    # | 1 binary   | sanitized        | -                     |
+    # |-------------------------------------------------------|
+    # | 2 binaries | unsanitized      | sanitized             |
+    # |-------------------------------------------------------|
+
+    main_target_path = pathlib.Path(target_path)
+    auxiliary_target_path = self._get_auxiliary_target_path(target_path)
+
+    Binary = namedtuple('Binary', ['unsanitized', 'sanitized'])
```
jonathanmetzman, line 144: Maybe just define this outside of this function.

jonathanmetzman, line 144: I don't think this is named so well because it represents two binaries and the variable on 147 is  "target_binaries".
Maybe name it TargetBinaries.

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
```diff
@@ -105,40 +106,73 @@ def prepare(self, corpus_dir, target_path, build_dir):
...
+    # |-------------------------------------------------------|
+    # |            | main target path | auxiliary target path |
+    # |-------------------------------------------------------|
+    # | 1 binary   | sanitized        | -                     |
+    # |-------------------------------------------------------|
+    # | 2 binaries | unsanitized      | sanitized             |
+    # |-------------------------------------------------------|
+
+    main_target_path = pathlib.Path(target_path)
+    auxiliary_target_path = self._get_auxiliary_target_path(target_path)
```
jonathanmetzman, line 142: I feel like the term auxiliary_target doesn't really describe what's happening here if it's possible that there's only an auxiliary and not a main. Maybe just call it sanitized and unsanitized.

src/clusterfuzz/_internal/tests/core/bot/fuzzers/centipede/centipede_engine_test.py
Alan32Liu: Sure!
I tested it a few times, and the minimum total time cost for these 26 testcases is 36(+-0.5) seconds. This reduces the time consumption from an original value of 202 seconds, with the following changes:
1. [`MAX_TIME`](https://github.com/google/clusterfuzz/blob/35909c46be1d0c6f4e00540dd5f94b80cea2b827/src/clusterfuzz/_internal/tests/core/bot/fuzzers/centipede/centipede_engine_test.py#L56) (to test `fuzz` and `reproduce`) reduced from 25 to 2 seconds.
2. [`_TIMEOUT_PER_INPUT_TEST`](https://github.com/google/clusterfuzz/blob/35909c46be1d0c6f4e00540dd5f94b80cea2b827/src/clusterfuzz/_internal/tests/core/bot/fuzzers/centipede/centipede_engine_test.py#LL64C1-L64C24) reduced from 5 to 2 seconds.
3. [`_CLEAN_EXIT_SECS`](https://github.com/google/clusterfuzz/blob/35909c46be1d0c6f4e00540dd5f94b80cea2b827/src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py#L30) reduced from 10 to 3 seconds.

Using any values lower than those will produce false-positive test failures, because `Centipede` could not launch and produce the error message in time.

Alan32Liu: I am a bit worried about using a low value of [_CLEAN_EXIT_SECS](https://github.com/google/clusterfuzz/blob/35909c46be1d0c6f4e00540dd5f94b80cea2b827/src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py#L30), because it may affect production.

I am a bit inclined to set it to 5 seconds, to give it extra redundancy in production.

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
Alan32Liu: I am a bit unsure if I understand you correctly: I don't think we will ever have an auxiliary without the main binary.
I intend to call the binary in the `BUILD_DIR` the main binary (which will always exist), and the one in `EXTRA_BUILD_DIR` the auxiliary (which may not exist).
They are not guaranteed to be sanitized or unsanitized:
1. If only the *main* binary exists, then we assume the main is sanitized. 
2. If both the *main* and the *auxiliary* exist, then we assume the main is unsanitized, and the auxiliary is sanitized.

I added this table to clarify this:
```
    3     # Centipede expects one or two target binaries:
    2     # |-------------------------------------------------------|
    1     # |            | main target path | auxiliary target path |
  136     # |-------------------------------------------------------|
    1     # | 1 binary   | sanitized        | -                     |
    2     # |-------------------------------------------------------|
    3     # | 2 binaries | unsanitized      | sanitized             |
    4     # |-------------------------------------------------------|

```

src/clusterfuzz/_internal/tests/core/bot/fuzzers/centipede/centipede_engine_test.py
jonathanmetzman: >202 seconds

I think this is too much time. Thank you for reducing this.

Yeah I would change _CLEAN_EXIT_SECS to 5 seconds. 48 seconds of testing time is an awful lot for a relatively small change. I think in the future we should do more unittesting of centipede changes and less integration testing, since these take very long.

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
jonathanmetzman: Ah I misunderstood. This is fine!

src/clusterfuzz/_internal/tests/core/bot/fuzzers/centipede/centipede_engine_test.py
Alan32Liu: > can you use mock to reduce the time just during testing?

I did not think of a way yet, because the slow tests (e.g., timeout) needs to run Centipede and parse its output.

PR #3043, opened 2023-05-10
[rearchitecture] Support distinguishing between trusted and untrusted tasks

jonathanmetzman:
Right now there are no untrusted tasks (utasks), but I will be moving over analyze_task soon.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/commands.py
+++ b/src/clusterfuzz/_internal/bot/tasks/commands.py
@@ -40,22 +40,73 @@ from clusterfuzz._internal.system import environment
 from clusterfuzz._internal.system import process_handler
 from clusterfuzz._internal.system import shell
 
-COMMAND_MAP = {
-    'analyze': analyze_task,
-    'blame': blame_task,
-    'corpus_pruning': corpus_pruning_task,
-    'fuzz': fuzz_task,
-    'impact': impact_task,
-    'minimize': minimize_task,
-    'progression': progression_task,
-    'regression': regression_task,
-    'symbolize': symbolize_task,
-    'unpack': unpack_task,
-    'upload_reports': upload_reports_task,
-    'variant': variant_task,
-}
-
 TASK_RETRY_WAIT_LIMIT = 5 * 60  # 5 minutes.
+
+
+class BaseTask:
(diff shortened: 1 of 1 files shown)


jonathanmetzman:

jonathanmetzman:
CC @flowerhack

src/clusterfuzz/_internal/bot/tasks/setup.py
```diff
@@ -280,21 +284,42 @@ def _get_testcase_file_and_path(testcase):
...
+  key = testcase.fuzzed_keys
+  archived = bool(testcase.archive_state & data_types.ArchiveStatus.FUZZED)
+  return key, archived
+
+
+def _is_testcase_minimized(testcase):
+  return testcase.minimized_keys and testcase.minimized_keys != 'NA'
+
+
+def download_testcase(key, testcase_download_url, dst):
```
Alan32Liu, line 309: Are we required to write doc-strings for small functions (like this one and the one above)?
I feel they are straightforward enough and do not need any notes, but I am unsure if doc-string is mandatory.

src/clusterfuzz/_internal/bot/tasks/setup.py
```diff
@@ -303,7 +328,7 @@ def unpack_testcase(testcase):
   else:
     temp_filename = testcase_file_path
```
Alan32Liu, line 329: Not sure what the convention is, but would it be better to get rid of the second layer with:
```
if archived and _is_testcase_minimized(testcase):
  ...
elif archived:
  ...
```

Alan32Liu: (nit)

Alan32Liu (approved):
LGTM with nits.
I might miss something because I was not familiar with the big picture.

src/clusterfuzz/_internal/bot/tasks/setup.py
jonathanmetzman: I think it's only really mandatory when linting requires it.
otherwise I would document if it's not obvious and/or widely used. This is so short and not-widely used that it's easier to read the code.

PR #3052, opened 2023-05-11
Make analyze_task untrusted

jonathanmetzman:
This will be the main PR in the process.
While this PR doesn't complete this work, it makes analyze_task use the untrusted_task API and it will make analyze_task
execute using the in-memory untrusted task executor.
Related: #3008

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
@@ -367,7 +367,6 @@ def utask_handle_errors(output):
     # Unclear if this state is ever actually reached.
     data_handler.update_testcase_comment(
         output.testcase, data_types.TaskState.ERROR, 'Testcase setup failed')
-    pass
 
 
 def utask_postprocess(output):
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
@@ -326,17 +326,19 @@ def utask_main(testcase, testcase_download_url, job_type, metadata):
         testcase,
         metadata=metadata,
         error=ErrorType.NO_CRASH,
-        crash_time=test_timeout,
+        test_timeout=test_timeout,
         job_type=job_type)
   # Update testcase crash parameters.
   update_testcase_after_crash(testcase, state, job_type, http_flag)
 
+  # See if we have to ignore this crash.
+  if crash_analyzer.ignore_stacktrace(state.crash_stacktrace):
+    return uworker_io.UworkerOutput(
+        testcase=testcase, metadata=metadata, error=ErrorType.IGNORE_STACK)
+
   test_for_reproducibility(testcase, testcase_file_path, state, test_timeout)
   return uworker_io.UworkerOutput(
-      metadata=metadata,
-      testcase=testcase,
-      crash_stacktrace=state.crash_stacktrace,
-      crash_time=crash_time)
+      metadata=metadata, testcase=testcase, crash_time=crash_time)
 
 
 def test_for_reproducibility(testcase, testcase_file_path, state, test_timeout):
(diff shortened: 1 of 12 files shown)


jonathanmetzman:
CC @flowerhack

src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
```diff
@@ -269,28 +267,37 @@ def execute_task(testcase_id, job_type):
...
   prepare_env_for_main(metadata)
 
   is_lsan_enabled = environment.get_value('LSAN')
   if is_lsan_enabled:
     # Creates empty local blacklist so all leaks will be visible to uploader.
     leak_blacklist.create_empty_local_blacklist()
 
-  testcase_file_path = setup_testcase_and_build(testcase, metadata, job_type,
-                                                testcase_id)
+  testcase_file_path, output = setup_testcase_and_build(
```
flowerhack, line 290: (nit, probably better addressed in a different CL): would a better variable name here be `results` (and perhaps `UworkerOutput`s should be `UworkerResult`s?  A `result` sounds like something that could contain output + error message, whereas `output` just sounds like it would be "something returned by a successful call", so I did a slight double-take trying to figure out why we'd return `output` when there's no `testcase_file_path`

src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
```diff
@@ -246,8 +244,8 @@ def update_testcase_after_crash(testcase, state, job_type, http_flag):
         bool(testcase.gestures))
 
 
-def execute_task(testcase_id, job_type):
-  """Runs analyze task."""
+def utask_preprocess(testcase_id, job_type, uworker_env):
```
flowerhack, line 247: is `utask_preprocess` called anywhere? or is this just establishing a function to be used in a future PR?

src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
```diff
@@ -315,30 +322,79 @@ def execute_task(testcase_id, job_type):
...
+        output.testcase, data_types.TaskState.ERROR, 'Testcase setup failed')
+  elif output.error == ErrorType.IGNORE_STACK:
+    data_handler.close_invalid_uploaded_testcase(output.testcase,
+                                                 output.metadata, 'Irrelavant')
+
+
+def utask_postprocess(output):
+  """Trusted: Cleans up after a uworker execute_task, writing anything needed to
+  the db."""
+  # TODO(metzman): Get rid of this after main migration PR.
```
flowerhack, line 386: Maybe be *slightly* more descriptive here?  ("Get rid of this after we have enabled untrusted variants for all Clusterfuzz tasks?")  (Just so that, if someone encounters this TODO years from now, they'll have a good idea of whether it can be deleted with confidence or not)

src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
```diff
@@ -269,28 +267,37 @@ def execute_task(testcase_id, job_type):
   metadata.timestamp = datetime.datetime.utcnow()
   metadata.put()
 
+  initialize_testcase_for_main(testcase, job_type)
+
+  testcase_download_url = setup.get_signed_testcase_download_url(testcase)
+  return {
```
flowerhack, line 273: is a dictionary what we want to return from here (rather than e.g. some kind of class/type)?  in general i'd have a slight preference for the latter, since i think it makes it easier to catch errors in advance rather than at runtime... but i don't have a fully formed opinion since i can't seem to find the context where this is called, and maybe a dictionary is in fact better for this use case

flowerhack (approved):
LGTM, added some nits/thoughts that might be helpful

🦜 🌴 🍎

src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
jonathanmetzman: It is called in utasks/__init__

src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
jonathanmetzman: It's just a note to myself to get rid of the testcase = output.testcase junk I did to avoid making this diff larger. I tried to make this diff small by leaving out any nonfunctional changes if possible.

src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
jonathanmetzman: I'll probably switch to classes but they will essentially be used as dictionaries (other than their init function).
I'm going to change soon now that i have a clearer idea of the API these tasks should conform to.

PR #3060, opened 2023-05-15
Fix issue signing URLs where wrong email was used.

jonathanmetzman:


jonathanmetzman:

PR #3065, opened 2023-05-15
Improve docker images to make CI faster.

jonathanmetzman:
1. Don't waste space so we don't have to spend time pulling.
2. Preinstall python deps.


oliverchang:
LGTM

PR #3066, opened 2023-05-16
Set the revision when it's available

jonathanmetzman:


jonathanmetzman:
I really should fix this code instead since it relies on side effects, but I don't have the ability to spend the effort required.

hogo6002 (approved):

PR #3072, opened 2023-05-18
[unittest] Have centipede_engine_test setup and teardown properly.

flowerhack:
When running unit tests, if the centipede_engine_test is skipped (e.g. when running locally), the call to `run` will still happen, but the calls to setUp will not.

This caused a problem with spurious failures in other unit tests, because `run` patched in new environment variables, and when subsequent tests made assertions about them (e.g. calls to `_assert_env_vars` in build_manager_test.py), it caused those tests to fail.

Arguably tests shouldn't assume that sort of thing, but Centipede should probably tear down properly as well. Moving this functionality into setUp will enforce this.

Additionally, tests using assertIsNone to check environment variables ought to check using assertIsFale instead. Due to a quirk of how os.environ is implemented, it's impossible to change an environment variable back to None once it's been set, only to the empty string. I updated build_manager_test.py to handle this appropriately.

### DIFF
--- a/src/clusterfuzz/_internal/tests/core/bot/fuzzers/centipede/centipede_engine_test.py
+++ b/src/clusterfuzz/_internal/tests/core/bot/fuzzers/centipede/centipede_engine_test.py
@@ -53,7 +53,7 @@ def get_test_paths():
 
 
 TEST_PATH = pathlib.Path(__file__).parent
-MAX_TIME = 25
+MAX_TIME = 5
 
 # Centipede's runtime args for testing.
 _SERVER_COUNT = 1
@@ -61,7 +61,7 @@ _RSS_LIMIT = 4096
 _ADDRESS_SPACE_LIMIT = 4096
 _TIMEOUT_PER_INPUT = 25
 _RSS_LIMIT_TEST = 2
-_TIMEOUT_PER_INPUT_TEST = 5  # For testing timeout only.
+_TIMEOUT_PER_INPUT_TEST = 2  # For testing timeout only.
 _DEFAULT_ARGUMENTS = [
     '--exit_on_crash=1',
     f'--fork_server={_SERVER_COUNT}',
@@ -99,16 +99,16 @@ class IntegrationTest(unittest.TestCase):
   """Integration tests."""
 
   def run(self, *args, **kwargs):
-    test_helpers.patch_environ(self)
     with get_test_paths() as test_paths:
       self.test_paths = test_paths
-      os.environ['BUILD_DIR'] = str(self.test_paths.data)
       super().run(*args, **kwargs)
 
   def setUp(self):
     self.maxDiff = None  # pylint: disable=invalid-name
     test_helpers.patch(self, ['os.getpid'])
     self.mock.getpid.return_value = 1337
+    test_helpers.patch_environ(self)
+    os.environ['BUILD_DIR'] = str(self.test_paths.data)
 
   def compare_arguments(self, expected, actual):
     """Compares expected arguments."""


flowerhack:
@jonathanmetzman

src/clusterfuzz/_internal/tests/core/bot/fuzzers/centipede/centipede_engine_test.py
```diff
@@ -99,16 +99,17 @@ class IntegrationTest(unittest.TestCase):
   """Integration tests."""
 
   def run(self, *args, **kwargs):
-    test_helpers.patch_environ(self)
     with get_test_paths() as test_paths:
       self.test_paths = test_paths
       os.environ['BUILD_DIR'] = str(self.test_paths.data)
       super().run(*args, **kwargs)
+      os.environ['BUILD_DIR'] = ''
```
jonathanmetzman, line 106: You could just move `os.environ['BUILD_DIR'] = str(self.test_paths.data)` into setUp and leave a comment as to why.

src/clusterfuzz/_internal/tests/core/build_management/build_manager_test.py
```diff
@@ -1238,11 +1238,11 @@ def setUp(self):
   def _assert_env_vars(self):
     """An auxiliary build show not overwrite the common env variables, but
     should set up some variables with a prefix in their names."""
-    self.assertIsNone(os.environ.get('APP_PATH'))
-    self.assertIsNone(os.environ.get('APP_REVISION'))
-    self.assertIsNone(os.environ.get('BUILD_DIR'))
-    self.assertIsNone(os.environ.get('BUILD_URL'))
-    self.assertIsNone(os.environ.get('GN_ARGS_PATH'))
+    self.assertFalse(os.environ.get('APP_PATH'))
```
jonathanmetzman, line 1241: Can we go back to assertIsNone after my suggestion?

jonathanmetzman:
Thanks for this!

src/clusterfuzz/_internal/tests/core/bot/fuzzers/centipede/centipede_engine_test.py
flowerhack: I'm not sure that entirely works—the way the code is structured now, we're relying on that `with get_test_paths() as foo` bit to automagically destroy the temp directories as soon as it leaves that scope, so it's got to be called inside `run`, which means that variable needs to get inside `run`.

What *could* work, I think, is adding "destroy this tmppath" to tearDown; woudl that be preferable to you?

jonathanmetzman: I think moving os.environ['BUILD_DIR'] = str(self.test_paths.data) and patch_environ works. I just tested it.
I'm pretty sure this will get rid of the need to do `os.environ['BUILD_DIR'] = ''`
Otherwise, if you want to do something only involving setUp and tearDown and eliminating run entirely, that's fine with me.

jonathanmetzman:
Since youre already modifying this file, would you mind making this change:
```
-MAX_TIME = 25
+MAX_TIME = 5
 
 # Centipede's runtime args for testing.
 _SERVER_COUNT = 1
@@ -61,7 +61,7 @@ _RSS_LIMIT = 4096
 _ADDRESS_SPACE_LIMIT = 4096
 _TIMEOUT_PER_INPUT = 25
 _RSS_LIMIT_TEST = 2
-_TIMEOUT_PER_INPUT_TEST = 5  # For testing timeout only.
+_TIMEOUT_PER_INPUT_TEST = 2  # For testing timeout only
```

Currently the tests take 2 minutes to run, this reduces that time by a minute.

flowerhack:
> Since youre already modifying this file, would you mind making this change:

Sure thing!  Got it in the most recent set of changes.

src/clusterfuzz/_internal/tests/core/bot/fuzzers/centipede/centipede_engine_test.py
flowerhack: Huh, you're right.  I must've been confused about the ordering of run vs setup.  Fixed as asked!

Alan32Liu (approved):
Thanks!

jonathanmetzman (approved):
lgtm

jonathanmetzman (approved):
lgtm

PR #3075, opened 2023-05-18
[unittest] Fix minijail unit test by upgrading the pyfakefs version.

flowerhack:
When running the unit tests locally, test_minijail_env_vars fails with an IsADirectoryError (Is a directory in the fake filesystem: '/tmp').

It looks as if this is an issue with pyfakefs misbehaving on certain platforms in newer versions of Python: https://github.com/pytest-dev/pyfakefs/issues/686

By updating the pipfile to a newer version of pyfakefs, the unit test now passes.


flowerhack:
@jonathanmetzman 

So, it occurs to me, after I finished looking into this, that I think i was only seeing failures because I was running `python3 butler.py py_unittest...` instead of `python butler.py py_unittest...`.  I'm assuming based on the docs that python3 actually isn't supported/intended?

This commit doesn't seem to *hurt* anything (I tested running both ways), but up to you whether this fix is relevant given that.

jonathanmetzman:
> @jonathanmetzman
> 
> So, it occurs to me, after I finished looking into this, that I think i was only seeing failures because I was running `python3 butler.py py_unittest...` instead of `python butler.py py_unittest...`. I'm assuming based on the docs that python3 actually isn't supported/intended?
> 
> This commit doesn't seem to _hurt_ anything (I tested running both ways), but up to you whether this fix is relevant given that.

You have to do python3 -m pipenv shell after doing ./local/install_dependencies.sh

jonathanmetzman:
Hmmm...it does seem to break in CI. Maybe if youre not seeing issues anymore we can put this on the backburner

flowerhack:
> Hmmm...it does seem to break in CI. Maybe if youre not seeing issues anymore we can put this on the backburner

Yep, that seems fair.  Closing

PR #3076, opened 2023-05-18
Fix issue with error handling

jonathanmetzman:


jonathanmetzman:
CC @flowerhack

jonathanmetzman:
Completely absurd that pylint doesn't catch this. I think it's because this type of analysis gets broken by the way we use pylint (not on the entire project at once).

flowerhack:
@jonathanmetzman (Out of curiosity, why don't we run pylint on the entire project at once? Would be happy to look into changing that unless it's doomed)

jonathanmetzman:
I'm changing things to do that as we speak. I don't know the answer but I think it's bad that we don't do this, and agree we should be running it on the entire project.
no-member has a lot of false positives but better to manually disable them one by one.

Alan32Liu (commented):
LGTM

Issue #3077, opened 2023-05-18
Improve linting


jonathanmetzman:
We don't lint properly and miss out on easily detectable errors like using something not present in a module.
We need to clean up the project so we can lint everything at once (as is supposed to be done).

jonathanmetzman:
This is done.

PR #3079, opened 2023-05-18
Lint properly

jonathanmetzman:
Lint everything at once so we can catch errors like using a nonexistent function from a different module.
Also make all the code lint clean.


jonathanmetzman:
CC @flowerhack

jonathanmetzman:
Fixes: https://github.com/google/clusterfuzz/issues/3077

Alan32Liu (approved):

PR #3086, opened 2023-05-21
[analyze] Ensure that the correct object is passed to error handler. 

jonathanmetzman:
Previously the error object was wrapped, leading to the handler
trying to access fields that exist on the error object but not
on the output object.


jonathanmetzman:
TBR to fix bug

Alan32Liu (commented):
LGTM

PR #3087, opened 2023-05-21
[consolidation] Securely preserve input for postprocess

jonathanmetzman:
The postprocess task needs to know about the input for the uworker.
For example, if the uworker is supposed to be analyzing testcase X, the postprocess cannot allow it to modify testcase Y.
To deal with this problem, make sure the postprocess task downloads the actual input and does not rely on it being
untampered with in the uworker output.


jonathanmetzman:
CC @flowerhack

hogo6002 (approved):

Alan32Liu (approved):

PR #3089, opened 2023-05-22
[consolidation] Use proto for communication

jonathanmetzman:
This replaces the current JSON-based system.
This provides the following benefits:
1. Deserialization is improved/simplfied. We reuse ndb's entity->proto code instead of implementing our own hacky deserialization for entities (which didn't support e.g. datetimes) 
2. Type safety. 
3. Typos in parameter names will crash closer to where the mistake is made
4. Speed


src/clusterfuzz/_internal/protos/uworker_msg_pb2.py
```diff
@@ -0,0 +1,505 @@
+# Copyright 2020 Google LLC
```
jonathanmetzman, line 1: This file is autogenerated and does not need review.

Alan32Liu:
It seems we have many conflits : )

jonathanmetzman:
> It seems we have many conflicts : )

Heh, they were actually quite minor (using uworker_errors.Type.UNHANDLED instead of uworker_msg_pb2.ErrorType.UNHANDLED

jonathanmetzman:
TBR

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
```diff
@@ -99,28 +84,31 @@ def get_entity_with_properties(ndb_key: ndb.Key, properties) -> ndb.Model:
...
-  serializable = {}
-  ndb_entities = {}
+  uworker_input = uworker_input.copy()
   for key, value in uworker_input.items():
-    if not isinstance(value, ndb.Model):
-      serializable[key] = value
-      continue
-    ndb_entities[key] = make_ndb_entity_input_obj_serializable(value)
+    if isinstance(value, ndb.Model):
+      uworker_input[key] = model._entity_to_protobuf(value)  # pylint: disable=protected-access
```
Alan32Liu, line 105: Would it be better to make the function public, given it has been used publicly in different places?

PR #3095, opened 2023-05-24
Fix serialization into proto

jonathanmetzman:
Don't have any default fields on UworkerOutput, it can trick deserialization into thinking we want to set None on the proto.


jonathanmetzman:
TBR

Alan32Liu (commented):
LGTM

PR #3096, opened 2023-05-24
Ensure that all entities used as input are wrapped

jonathanmetzman:
Otherwise serialization breaks and we can't track changes.


jonathanmetzman:
TBR

Alan32Liu (commented):
LGTM

PR #3098, opened 2023-05-25
[uworker_io] Make types match db

jonathanmetzman:


jonathanmetzman:
TBR to fix another bug.

jonathanmetzman:
I'm actually not sure it's so great to have a part of the code that fails at runtime when types are wrong in a codebase that is otherwise dynamicall typed.

Alan32Liu (approved):

PR #3104, opened 2023-05-31
[consolidation] Ensure uworker_input is always included in uworker_output

jonathanmetzman:
Right now, callers to setup_testcase are all using its error handler which assumes the uworker module adds
uworker_input to the output.
Temporarily include this for callers that are not migrated to utasks.


Alan32Liu (commented):
LGTM

PR #3110, opened 2023-06-01
Add more tests for bug throttler.

hogo6002:
The previous code did not test the situation where the `max_bugs_per_project_per_24hrs` variable is not defined, which caused invalid config value of 'max_bugs_per_project_per_24hrs' error. Adding more tests to test the default limit and adding more details to the logs.


oliverchang (approved):
LGTM, but please add more context to the PR description for this change. e.g. which previous breakage prompted this.

PR #3115, opened 2023-06-01
Remove remaining redundant u-string prefixes

jonathanmetzman:


oliverchang (approved):

Alan32Liu (approved):
Thanks!
I also got pages of errors from `pylint` in the morning, thanks so much for fixing it!

PR #3122, opened 2023-06-02
Make Exception naming style-conformant, modernize

jonathanmetzman:


jonathanmetzman:
@oliverchang Do we want something like this? I went down a rabbit hole after renaming some exceptions. Most of the transformations were automated so it's not a lot of wasted effort.

jonathanmetzman:
I think the f-string transform is the most dubious value change. The others like getting rid of six and making exceptions end in "Error" instead of "Exception" seem more worthwhile, since they are solvable problems (we will probably never entirely get rid of .format)

oliverchang:
definitely seems worthwhile as a general code cleanup to me! 

+1 to f-string change being a little dubious. It's also the most riskiest of all these changes, so we could just omit that for now.

jonathanmetzman:
Done.

oliverchang (approved):
LGTM. Spot checked various places but obviously couldn't comprehensively review this whol PR :) Please monitor carefully.

Alan32Liu (approved):

PR #3126, opened 2023-06-05
Separate out slow deps installation

jonathanmetzman:
It makes sense to rerun the second half of install deps (e.g. installing python packages) often.
But the first half takes a long time and probably only needs to be run once.
Separate them out to make development more rapid.


oliverchang (commented):
LGTM. We typically don't need to re-run install_deps_linux.bash though, and only need `python butler.py bootstrap`. Is this not the case for you?

PR #3127, opened 2023-06-05
Upgrade to Python3.11

jonathanmetzman:


oliverchang:
You'll also need to update https://github.com/google/clusterfuzz/blob/8d52cf4d3ebf22ec92e6398b89670765dd0fb4f8/src/local/butler/constants.py#L52. One challenge is making sure our versions of dependencies have the pre-builts for the Python 3.11 ABI.

jonathanmetzman:
Abandoning this and going with https://github.com/google/clusterfuzz/pull/3229 instead.

PR #3128, opened 2023-06-05
Don't test for old versions of libFuzzer.

jonathanmetzman:
This version is 4 years old.
The tests take 25 seconds to execute.


oliverchang (approved):
Do we need to delete the old binary here too?

jonathanmetzman:
> Do we need to delete the old binary here too?

I think it's still used in tests for mutator plugins (another feature I should delete probably)

PR #3135, opened 2023-06-06
[afl] Make integration tests parallel friendly.

jonathanmetzman:
This simplifies the testing process, the AFL testing code, and speeds things up since the tests can not only be run in parallel with each other, but with all other tests.


jonathanmetzman:
TBR to make my life easier deploying.

PR #3141, opened 2023-06-07
Fix exception with variant task.

oliverchang:
This is likely due to issues with the input serialization not serializing None values. 

There may also be issues at the deserialization end.


oliverchang:
TBR to fix exceptions.

Alan32Liu (approved):

PR #3142, opened 2023-06-07
[consolidation] Set output proto eagerly

jonathanmetzman:
Previously we did it lazily, making errors harder to debug.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
@@ -267,35 +267,33 @@ class UworkerOutput:
   """Convenience class for results from uworker_main. This is useful for
   ensuring we are returning values for fields expected by utask_postprocess."""
 
-  def __init__(self, serializable=True, **kwargs):
-    self.serializable = serializable
+  def __init__(self, **kwargs):
     self.proto = uworker_msg_pb2.Output()
     for key, value in kwargs.items():
       setattr(self, key, value)
 
   def __getattr__(self, attribute):
-    if attribute in ['proto', 'serializable'] or not self.serializable:
+    if attribute in ['proto']:
       # Allow setting and changing proto. Stack overflow in __init__ otherwise.
       return super().__getattr__(attribute)  # pylint: disable=no-member
     return getattr(self.proto, attribute)
 
   def __setattr__(self, attribute, value):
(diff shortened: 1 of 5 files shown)


src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
```diff
@@ -279,31 +265,42 @@ class UworkerOutput:
...
 
-  def __init__(self, **kwargs):
-    self._set_attrs = set()
-    # Reset _set_attrs so we don't consider these set by the user unless they
-    # explictly set them.
-    self.testcase = None
-    self.error = None
-    self._set_attrs = set()
-
+  def __init__(self, serializable=True, **kwargs):
```
oliverchang, line 268: When would serializable be false?

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
```diff
@@ -279,31 +265,42 @@ class UworkerOutput:
   """Convenience class for results from uworker_main. This is useful for
```
oliverchang, line 265: Question for the purpose of this class. We're wrapping a proto, which already is a Python object. Could we instead just use that proto object directly and avoid this setattr hackery?

oliverchang (approved):

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
jonathanmetzman: Never. Sorry, left this in by accident.

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
jonathanmetzman: We can't. There are fields in the proto that are serialized db entities but are not db entities. To use those entities we need to deserialize them from the proto format into a real python object. But we can't simply reassign the field to the deserialized version of the field because proto enforces types.
Did I explain this well? Should I leave this in a comment.

oliverchang: Maybe another ignorant question but can we make the proto type more flexible in that case? 

Otherwise yes, please add a code comment! that would be super helpful in understanding this.

PR #3145, opened 2023-06-07
Fix flaky honggfuzz netdriver test

jonathanmetzman:
The test assumed that inputs are less than 1024 bytes, causing it to frequently break deploys.


oliverchang (approved):

Alan32Liu (commented):
Thanks!

PR #3146, opened 2023-06-07
[storage] Consolidate error handling

jonathanmetzman:
Retryable functions can experience the same transient errors, therefore sharing the exception list makes oversights less likely


oliverchang (approved):
thanks!

PR #3150, opened 2023-06-09
[analyze] Set metadata properly after setting up build

jonathanmetzman:
Also add test for this behavior.
Fixes crbug.com/1450594
Also fix testcase in fuzz_task.py that accidentally inherited from object instead of unittest.TestCase

### DIFF
--- a/src/clusterfuzz/_internal/base/utils.py
+++ b/src/clusterfuzz/_internal/base/utils.py
@@ -562,9 +562,9 @@ def read_data_from_file(file_path, eval_data=True, default=None):
   if not os.path.exists(file_path):
     return default
 
-  failure_wait_interval = environment.get_value('FAIL_WAIT')
+  failure_wait_interval = environment.get_value('FAIL_WAIT', 0)
   file_content = None
-  retry_limit = environment.get_value('FAIL_RETRIES')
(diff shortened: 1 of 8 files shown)


src/clusterfuzz/_internal/base/utils.py
```diff
@@ -562,9 +562,9 @@ def read_data_from_file(file_path, eval_data=True, default=None):
   if not os.path.exists(file_path):
     return default
 
-  failure_wait_interval = environment.get_value('FAIL_WAIT')
+  failure_wait_interval = environment.get_value('FAIL_WAIT', 0)
   file_content = None
-  retry_limit = environment.get_value('FAIL_RETRIES')
+  retry_limit = environment.get_value('FAIL_RETRIES', 1)
```
jonathanmetzman, line 567: I could set these in my test, but it seems like it's better for this function to not depend on these variables being set.

jonathanmetzman:
@oliverchang Any idea why the test fails here: https://github.com/google/clusterfuzz/pull/3150/files#diff-1e18e69d816a44a2f3113f571262103212a81b11a83611fd7971c1d409c16c17R155

I know the metadata is being set. Maybe we don't handle metadata properly in tests.

jonathanmetzman:
I'm just going to delete my test for now.

PR #2955, opened 2023-02-27
Remove MLRNN

jonathanmetzman:
It's no longer used.

PR #3153, opened 2023-06-09
Remove ML cronjob

jonathanmetzman:
Follow up to #2955
It's failing now, because the endpoint was removed.


Alan32Liu (approved):
LGTM, may I ask what was that RNN doing?

PR #3161, opened 2023-06-15
Add integration test for analyze_task

jonathanmetzman:
Also add framework for integration tests for utasks.


jonathanmetzman:
I think this is failing because of some kind of interference with afl and centipede integration tests.

jonathanmetzman:
I'm having trouble running this locally so it's extremely hard to test this.

jonathanmetzman:
So weird, there's "no output" from running the fuzzer. But I think this fuzzer is run in other intergration tests so it shouldn't be broken

jonathanmetzman:
I might just change the test to handle cases when test cases dont repro, since running the fuzzer isn't really the point. What do you think Oliver?

oliverchang:
> I might just change the test to handle cases when test cases dont repro, since running the fuzzer isn't really the point. What do you think Oliver?

That seems fine, but it is strange that the fuzzer isn't working for you locally.

jonathanmetzman:
> > I might just change the test to handle cases when test cases dont repro, since running the fuzzer isn't really the point. What do you think Oliver?
> 
> That seems fine, but it is strange that the fuzzer isn't working for you locally.

Everything is working for me locally now. In CI it's a different story

PR #3174, opened 2023-06-21
[fuchsia] Handle expected cases of missing fuzzers

eepeep:
In some cases, such as regression tasks, it is expected that a given build may not contain the desired target fuzzer. Instead of failing the task due to an underlying undercoat error, we now explicitly check for this situation and raise the expected error type.

Fix: https://fxbug.dev/100832

### DIFF
--- a/src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py
@@ -29,6 +29,7 @@ from clusterfuzz._internal.bot.fuzzers import mutator_plugin
 from clusterfuzz._internal.bot.fuzzers import utils as fuzzer_utils
 from clusterfuzz._internal.bot.fuzzers.libFuzzer import constants
 from clusterfuzz._internal.bot.fuzzers.libFuzzer.peach import pits
+from clusterfuzz._internal.bot.testcase_manager import TargetNotFoundError
 from clusterfuzz._internal.fuzzing import strategy
 from clusterfuzz._internal.metrics import logs
 from clusterfuzz._internal.platforms import android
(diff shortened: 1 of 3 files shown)


eepeep:
This is a little awkward due to the differences in the Fuchsia implementation (e.g. `find_fuzzer_path` is used elsewhere for fuzzer presence check but is meaningless for Fuchsia) . I think it's relatively straightforward, but let me know if you think there's a more appropriate approach here.  Also, do you have a suggestion for a good place to put an integration test for this?

eepeep:
cc @oliverchang I'm not sure if you're still the best person to tag for review?

oliverchang:
> cc @oliverchang I'm not sure if you're still the best person to tag for review?

I still am :) But also adding @jonathanmetzman

oliverchang (approved):

oliverchang:
> This is a little awkward due to the differences in the Fuchsia implementation (e.g. `find_fuzzer_path` is used elsewhere for fuzzer presence check but is meaningless for Fuchsia) . I think it's relatively straightforward, but let me know if you think there's a more appropriate approach here. Also, do you have a suggestion for a good place to put an integration test for this?

Re integration test, we could probably just add one here https://github.com/google/clusterfuzz/blob/390daadac83a88e3e4221240d1b247cbcaa29c95/src/clusterfuzz/_internal/tests/core/bot/fuzzers/libFuzzer/engine_test.py#L920 with a bogus fuzzer name?

eepeep:
Thanks for the quick review!

Ah, yeah, I was thinking about a test at the level of the regression task but I think your suggestion is simpler and good enough. Test added.

/gcbrun

oliverchang (approved):

src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py
```diff
@@ -428,6 +429,24 @@ def _clear_all_target_corpora(self):
     # prepare_fuzzer resets the data/ directory
     undercoat.prepare_fuzzer(self.handle, self.executable_path)
 
+  def _ensure_target_exists(self):
+    """ Check that the target fuzzer exists, raising an error if it does not.
```
jonathanmetzman, line 433: nit: extra space at the beginning.

jonathanmetzman (commented):
lgtm

src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py
eepeep: Oops, thanks :) I was mirroring the format of the functions above, so I've clearly done this before-- will try to remember next time!

PR #3182, opened 2023-06-28
Convert Java Stackframes to a C-like stackframe

jonathanmetzman:
Still left to do is infer where the beginning of this source path is. Just because we know the bug is in apache/lucene/util/ArrayUtil.java, doesn't mean we actually know where that is in the repo.
Gate this on a project config for now.


jonathanmetzman:
Note that I don't turn this on by default because it won't get any path correct without serious changes to src mapping (to figure out where the root of the path is located)

oliverchang (approved):
thanks! +1 to just hardcoding this for the internal case.

src/clusterfuzz/_internal/build_management/source_mapper.py
```diff
@@ -183,8 +187,29 @@ def get_vcs_viewer_for_url(url):
   return None
 
 
+def should_linkify_java_stack_frames():
+  return local_config.Config(local_config.PROJECT_PATH).get('linkify_java')
```
oliverchang, line 191: This may be causing some performance regressions if this is reading in the config from a file for every single line.

PR #3187, opened 2023-06-28
Allow deduplication within fuzz targets

jonathanmetzman:


src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -120,20 +120,48 @@ def get_testcase_by_id(testcase_id):
   return testcase
 
 
+@memoize.wrap(memoize.Memcache(MEMCACHE_TTL_IN_SECONDS))
+def _get_builtin_fuzzers():
+  return data_types.Fuzzer.query(ndb_utils.is_true(data_types.Fuzzer.builtin))
```
oliverchang, line 125: We should just use this instead? https://github.com/google/clusterfuzz/blob/c2c875fa8378a732e88e1c155e12b7695672442f/src/clusterfuzz/_internal/fuzzing/__init__.py#L16

src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -120,20 +120,48 @@ def get_testcase_by_id(testcase_id):
...
+  query_args = [
       data_types.Testcase.project_name == project_name,
       data_types.Testcase.crash_type == crash_type,
       data_types.Testcase.crash_state == crash_state,
       data_types.Testcase.security_flag == security_flag,
       data_types.Testcase.status == 'Processed',
-      ndb_utils.is_true(data_types.Testcase.open))
+      ndb_utils.is_true(data_types.Testcase.open)
+  ]
+  if fuzz_target and environment.get_value('DEDUP_ONLY_SAME_TARGET'):
```
oliverchang, line 144: Can we key this on a local_config property (keyed on the issue filing project) instead?

oliverchang (commented):
can we also add some tests?

src/clusterfuzz/_internal/datastore/data_handler.py
jonathanmetzman: Is that different for android than for google 3?

oliverchang: No, but we could break it out into its own. Env vars are not as reliable and not available consistently in all environments (e.g. App Engine)

jonathanmetzman: I'll land this and try it out with one job and then switch to project (an overloaded term :-() based config.

oliverchang: SGTM. It's good to keep these in a consistent place and reduce our reliance on env vars going forward. If we ever need this in App Engine in the future, this could get confusing.

jonathanmetzman:
> can we also add some tests?

Ah sorry, I meant to do this.

src/clusterfuzz/_internal/datastore/data_handler.py
jonathanmetzman: Yes, I'll switch to a config before using this env var in more jobs then.

PR #3189, opened 2023-06-28
Do batch

jonathanmetzman:


jonathanmetzman:
@oliverchang I want postprocess to run on nonpreemptible for now, is writing to the "high end" queue a guarantee of this?

jonathanmetzman:
@oliverchang Another more design-like consideration I want your thoughts on is whether uworker inputs and outputs should be uploaded to the same bucket. Currently they are, which means that every time the uworker input is uploaded a bot needs to process the message and acknowledge it.

oliverchang:
> @oliverchang Another more design-like consideration I want your thoughts on is whether uworker inputs and outputs should be uploaded to the same bucket. Currently they are, which means that every time the uworker input is uploaded a bot needs to process the message and acknowledge it.

Creating another bucket seems pretty cheap to keep things cleaner here, or are there other considerations here?

jonathanmetzman:
> > @oliverchang Another more design-like consideration I want your thoughts on is whether uworker inputs and outputs should be uploaded to the same bucket. Currently they are, which means that every time the uworker input is uploaded a bot needs to process the message and acknowledge it.
> 
> Creating another bucket seems pretty cheap to keep things cleaner here, or are there other considerations here?

It's definitely cheaper, just a little more complex IMO. OK I'll switch to that model.

jonathanmetzman:
> @oliverchang I want postprocess to run on nonpreemptible for now, is writing to the "high end" queue a guarantee of this?

Ping on this question.

src/clusterfuzz/_internal/protos/uworker_msg.proto
```diff
@@ -39,6 +39,7 @@ message Input {
   optional google.datastore.v1.Entity variant = 8;
   optional string original_job_type = 9;
   optional string fuzzer_name = 10;
+  optional string module_name = 11;
```
oliverchang, line 42: nit: maybe rename this "task_name" instead to tie it less to Python.

PR #3191, opened 2023-06-30
Don't sleep for too long during tests

jonathanmetzman:


oliverchang (approved):

src/clusterfuzz/_internal/base/retry.py
```diff
@@ -31,7 +32,16 @@ def sleep(seconds):
 
 def get_delay(num_try, delay, backoff):
   """Compute backoff delay."""
-  return delay * (backoff**(num_try - 1))
+  delay = delay * (backoff**(num_try - 1))
+  if (environment.get_value('PY_UNITTESTS') or
```
oliverchang, line 36: I wonder if we should move these into a common helper instead.

PR #3196, opened 2023-07-06
Do partial migration of corpora to zipcorpora.

jonathanmetzman:
This is necessary to convert corpus_pruning and fuzz_task to untrusted.

### DIFF
--- a/src/clusterfuzz/_internal/fuzzing/corpus_manager.py
+++ b/src/clusterfuzz/_internal/fuzzing/corpus_manager.py
@@ -250,7 +250,7 @@ class GcsCorpus:
 
     return url
 
-  def get_zipcorpus_gcs_url(self, suffix=''):
+  def get_zipcorpus_gcs_dir_url(self, suffix=''):
     """Build zipcorpus GCS URL for gsutil.
     Returns:
       A string giving the GCS URL.
@@ -286,15 +286,14 @@ class GcsCorpus:
     # TODO(metzman): Get rid of the rest of this function when migration is
     # complete.
     filenames = shell.get_files_list(directory)
-    self._upload_to_zipcorpus(
-        filenames, CORPUS_FILES_SYNC_TIMEOUT, partial=False)
+    self._upload_to_zipcorpus(filenames, partial=False)
 
     # Allow a small number of files to fail to be synced.
     return _handle_rsync_result(result, max_errors=MAX_SYNC_ERRORS)
 
   def get_zipcorpora_gcs_urls(self, max_partial_corpora=float('inf')):
-    yield self.get_zipcorpus_name_and_gcs_url(partial=False)[1]
-    partial_corpora_gcs_url = f'{self.get_zipcorpus_gcs_url()}/partial*'
+    yield self.get_zipcorpus_gcs_url(partial=False)[1]
+    partial_corpora_gcs_url = f'{self.get_zipcorpus_gcs_dir_url()}/partial*'
     partial_corpora = storage.list_blobs(partial_corpora_gcs_url)
     for idx, partial_corpus in enumerate(partial_corpora):
       if idx > max_partial_corpora:
@@ -339,9 +338,7 @@ class GcsCorpus:
     # Allow a small number of files to fail to be synced.
     return _handle_rsync_result(result, max_errors=MAX_SYNC_ERRORS)
 
-  def upload_files(self,
-                   file_paths,
-                   timeout=CORPUS_FILES_SYNC_TIMEOUT):
+  def upload_files(self, file_paths, timeout=CORPUS_FILES_SYNC_TIMEOUT):
     """Upload files to the GCS.
 
     Args:
@@ -364,11 +361,11 @@ class GcsCorpus:
     # Upload zipcorpus.
     # TODO(metzman): Get rid of the rest of this function when migration is
     # complete.
-    self._upload_to_zipcorpus(file_paths, timeout, partial=True)
+    self._upload_to_zipcorpus(file_paths, partial=True)
     return result
 
-  def get_zipcorpus_name_and_gcs_url(self, partial):
-    zipcorpus_url = self.get_zipcorpus_gcs_url()
+  def get_zipcorpus_gcs_url(self, partial):
+    zipcorpus_url = self.get_zipcorpus_gcs_dir_url()
     if partial:
       timestamp = datetime.datetime.now().strftime('%Y-%m-%d-%H%M%S')
       prefix = f'partial-{timestamp}'
@@ -376,16 +373,16 @@ class GcsCorpus:
       prefix = 'base'
 
     filename = f'{prefix}.zip'
-    return filename, zipcorpus_url + filename
+    return zipcorpus_url + filename
 
-  def _upload_to_zipcorpus(self, file_paths, timeout, partial):
-    filename, gcs_url = self.get_zipcorpus_name_and_gcs_url(partial)
-    with zip_files(filename, file_paths) as archive_path:
+  def _upload_to_zipcorpus(self, file_paths, partial):
+    gcs_url = self.get_zipcorpus_gcs_url(partial)
+    with zip_files(file_paths) as archive_path:
       storage.copy_file_to(archive_path, gcs_url)
 
 
 @contextlib.contextmanager
-def zip_files(base_filename, file_paths):
+def zip_files(file_paths):
   with get_temp_zip_filename() as zip_filename:
     with zipfile.ZipFile(zip_filename, 'w') as zip_file:
       for file_path in file_paths:
--- a/src/clusterfuzz/_internal/fuzzing/corpus_manager.py
+++ b/src/clusterfuzz/_internal/fuzzing/corpus_manager.py
@@ -302,6 +302,7 @@ class GcsCorpus:
       yield partial_corpus
 
   def download_zipcorpora(self, dst_dir):
+    """Downloads zipcorpora to |dst_dir|"""
     zipcorpora_urls = self.get_zipcorpora_gcs_urls()
     for zipcorpus_url in zipcorpora_urls:
       # TODO(metzman): Find out what's the tradeoff between writing the file to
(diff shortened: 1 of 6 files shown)


src/clusterfuzz/_internal/fuzzing/corpus_manager.py
```diff
@@ -243,6 +250,19 @@ def get_gcs_url(self, suffix=''):
 
     return url
 
+  def get_zipcorpus_gcs_url(self, suffix=''):
+    """Build zipcorpus GCS URL for gsutil.
+    Returns:
+      A string giving the GCS URL.
+    """
+    url = f'gs://zc/{self.bucket_name}{self.bucket_path}{suffix}'
```
jonathanmetzman, line 258: I need to add something to distinguish from the ordinary corpus. I suppose I could make some kind of change later that drops this while preserving corpora (e.g. we read from both /zc/ and without it).

src/clusterfuzz/_internal/fuzzing/corpus_manager.py
```diff
@@ -243,6 +250,20 @@ def get_gcs_url(self, suffix=''):
 
     return url
 
+  def get_zipcorpus_gcs_dir_url(self):
+    """Build zipcorpus GCS URL for gsutil.
+    Returns:
+      A string giving the GCS URL.
+    """
+    url = storage.get_cloud_storage_file_path(self.bucket_name,
+                                              f'zc{self.bucket_path}')
```
oliverchang, line 259: Rather than prepend 'zc', can we instead use a more descriptive name/path, e.g. "/zipped/" (and put that in a constant in the implementation here)?

src/clusterfuzz/_internal/fuzzing/corpus_manager.py
```diff
@@ -259,12 +280,39 @@ def rsync_from_disk(self,
...
+    # Upload zipcorpus.
+    # TODO(metzman): Get rid of the rest of this function when migration is
+    # complete.
+    filenames = shell.get_files_list(directory)
+    self._upload_to_zipcorpus(filenames, partial=False)
 
     # Allow a small number of files to fail to be synced.
     return _handle_rsync_result(result, max_errors=MAX_SYNC_ERRORS)
 
+  def get_zipcorpora_gcs_urls(self, max_partial_corpora=float('inf')):
```
oliverchang, line 295: nit: docstring.

oliverchang: Also add more comments about what this is doing: This is getting the N (latest?/random?) partial corpora.

src/clusterfuzz/_internal/fuzzing/corpus_manager.py
```diff
@@ -259,12 +280,39 @@ def rsync_from_disk(self,
...
+    yield self.get_zipcorpus_gcs_url(partial=False)
+    partial_corpora_gcs_url = f'{self.get_zipcorpus_gcs_dir_url()}/partial*'
+    partial_corpora = storage.list_blobs(partial_corpora_gcs_url)
+    for idx, partial_corpus in enumerate(partial_corpora):
+      if idx > max_partial_corpora:
+        break
+      yield partial_corpus
+
+  def download_zipcorpora(self, dst_dir):
+    zipcorpora_urls = self.get_zipcorpora_gcs_urls()
```
oliverchang, line 305: nit: no need for the intermediate variable.

src/clusterfuzz/_internal/fuzzing/corpus_manager.py
```diff
@@ -259,12 +280,39 @@ def rsync_from_disk(self,
...
+  def get_zipcorpora_gcs_urls(self, max_partial_corpora=float('inf')):
+    yield self.get_zipcorpus_gcs_url(partial=False)
+    partial_corpora_gcs_url = f'{self.get_zipcorpus_gcs_dir_url()}/partial*'
+    partial_corpora = storage.list_blobs(partial_corpora_gcs_url)
+    for idx, partial_corpus in enumerate(partial_corpora):
+      if idx > max_partial_corpora:
+        break
+      yield partial_corpus
+
+  def download_zipcorpora(self, dst_dir):
```
oliverchang, line 304: nit: docstring.

src/clusterfuzz/_internal/fuzzing/corpus_manager.py
```diff
@@ -259,12 +280,39 @@ def rsync_from_disk(self,
...
+    zipcorpora_urls = self.get_zipcorpora_gcs_urls()
+    for zipcorpus_url in zipcorpora_urls:
+      # TODO(metzman): Find out what's the tradeoff between writing the file to
+      # disk first or unpacking it in-memory.
+      with get_temp_zip_filename() as temp_zip_filename:
+        if not storage.exists(zipcorpus_url):
```
oliverchang, line 310: Should we log a warning/error? When would this happen? Race condition from corpus pruning?

src/clusterfuzz/_internal/fuzzing/corpus_manager.py
```diff
@@ -297,16 +350,49 @@ def upload_files(self, file_paths, timeout=CORPUS_FILES_SYNC_TIMEOUT):
...
+    result = self._gsutil_runner.upload_files_to_url(
         file_paths, gcs_url, timeout=timeout)
 
+    # Upload zipcorpus.
+    # TODO(metzman): Get rid of the rest of this function when migration is
+    # complete.
+    self._upload_to_zipcorpus(file_paths, partial=True)
+    return result
+
+  def get_zipcorpus_gcs_url(self, partial):
```
oliverchang, line 370: nit: docstring here and everywhere else.

oliverchang, line 370: nit: move this together with `get_zipcorpus_gcs_dir_url`

src/clusterfuzz/_internal/fuzzing/corpus_manager.py
jonathanmetzman: I've made it longer.

src/clusterfuzz/_internal/fuzzing/corpus_manager.py
jonathanmetzman: 1. When a fuzzer is new. This is fine and I can't imagine special handling.
2. The pruning race. I think this is pretty rare so I'm going to not handle it specially (one could imagine getting a new list).
I'm going to add logging to keep track of whether 2. is indeed common.

src/clusterfuzz/_internal/fuzzing/corpus_manager.py
jonathanmetzman: That function is used by another caller so I can't.

src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/corpus_pruning_task_test.py
```diff
@@ -364,10 +361,10 @@ def setUp(self):
                             f'QUARANTINE_BUCKET = {self.quarantine_bucket}\n'
                             f'BACKUP_BUCKET={self.backup_bucket}\n'
                             'RELEASE_BUILD_BUCKET_PATH = '
-                            'gs://clusterfuzz-test-data/test_libfuzzer_builds/'
+                            'gs://clusterfuzz-test-data/test_libfuzzer_builds2/'
```
jonathanmetzman, line 364: test_libfuzzer_builds has a fuzzer with a crazy bug where ASAN crashes before fuzzing starts if the path name is too long (Really true!). It's not clear to me how this test was passing this whole time. I rebuilt it a functioning version in test_libfuzzer_builds2 but this one does crash on every input (which I'm not sure was the intent).
I also had another issue when I compiled this with optimizations since inlining broke some other assumptions the test makes about the result.

src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/corpus_pruning_task_test.py
```diff
@@ -395,7 +392,6 @@ def setUp(self):
         job='libfuzzer_asan_job2',
         last_run=datetime.datetime.now()).put()
 
-    environment.set_value('USE_MINIJAIL', True)
```
jonathanmetzman, line 398: I assume minijail was used because untrusted used to be used with it all the time. Since this is not the case anymore, I've removed it.

src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/corpus_pruning_task_test.py
```diff
@@ -439,30 +435,20 @@ def tearDown(self):
   def test_prune(self):
     """Test pruning."""
     self._setup_env(job_type='libfuzzer_asan_job')
-    self.mock._record_cross_pollination_stats.side_effect = (
```
jonathanmetzman, line 442: I'm not sure if this check ever passed, it's failures are silent because after they happen a `catch Exception` in corpus_pruning_task hides the error.

src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/corpus_pruning_task_test.py
```diff
@@ -439,30 +435,20 @@ def tearDown(self):
...
     uworker_input = uworker_io.DeserializedUworkerMsg(
         job_type='libfuzzer_asan_job', fuzzer_name='libFuzzer_test_fuzzer')
+
     corpus_pruning_task.utask_main(uworker_input)
 
     corpus_dir = os.path.join(self.temp_dir, 'corpus')
     os.mkdir(corpus_dir)
-    self.corpus.rsync_to_disk(corpus_dir)
 
+    self.corpus.rsync_to_disk(corpus_dir)
```
jonathanmetzman, line 446: Another thing that made this test nightmarish to debug was it's use of rsync_to_disk and rsync_from_disk, both in setup and here, which changed the results in subtle ways because of how I modified the functions.

src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/corpus_pruning_task_test.py
oliverchang: Wow...

PR #3197, opened 2023-07-06
[rearchitecture] Prepare to migrate fuzz_task to a utask.

jonathanmetzman:


src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
```diff
@@ -23,6 +23,7 @@
 def tworker_preprocess_no_io(utask_module, task_argument, job_type,
                              uworker_env):
   logs.log('Starting utask_preprocess: %s.' % utask_module)
+  set_uworker_env(uworker_env)
```
jonathanmetzman, line 26: @oliverchang not sure if I should do this but it makes reusing existing code (which basically expects the job env at any point during the task) much easier.

PR #3198, opened 2023-07-06
[fuzz][rearchitecture] Move testcase stats uploading to postprocess

jonathanmetzman:


src/clusterfuzz/_internal/protos/uworker_msg.proto
```diff
@@ -39,6 +39,12 @@ message Input {
   optional google.datastore.v1.Entity variant = 8;
   optional string original_job_type = 9;
   optional string fuzzer_name = 10;
+  optional int64 crash_revision = 11;
```
oliverchang, line 42: This is getting a bit long and complicated. Can we make this more structured by defining sub-types?

src/clusterfuzz/_internal/protos/uworker_msg.proto
```diff
@@ -39,6 +39,12 @@ message Input {
...
+  optional int64 job_run_timestamp = 12;
+  optional int64 new_crash_count = 13;
+  optional int64 known_crash_count = 14;
+  optional int64 testcases_executed = 15;
+  optional Json job_run_crashes = 16;
```
oliverchang, line 47: Also, should these be added to the `Output`  instead of `Input` ?

src/clusterfuzz/_internal/protos/uworker_msg.proto
jonathanmetzman: Done

src/clusterfuzz/_internal/protos/uworker_msg.proto
jonathanmetzman: Done

jonathanmetzman:
This may suffer from some problems fixed by https://github.com/google/clusterfuzz/pull/3203 (though definitely not the known problem) since it creates sessions twice.

jonathanmetzman:
This wont work as is, we need to get the fuzz target name some other way.

oliverchang (commented):
nice, the new more structured output looks a bit better.

Let's see if we can apply that to other existing and future fields (for the input also) too.

jonathanmetzman:
> Let's see if we can apply that to other existing and future fields (for the input also) too.

I'm definitely going to apply it to future fields. To existing fields I'll submit another PR.

oliverchang (approved):

PR #3203, opened 2023-07-09
Don't create fuzz sessions multiple times.

jonathanmetzman:
It is probably uneeded and breaks things because of ClusterFuzz's use of global variables, by causing the same flags to be provided multiple times to the target.

Fixes: https://crbug.com/1462793

This reverts commits 2a4bf955c863c4bf387b7ef12cf5c396caf7a6ab and 5ff6ea69e90fefad339387b8f7bfdb559cf89051.


oliverchang (commented):
LGTM

PR #3209, opened 2023-07-11
Apply ASAN options to targets

DonggeLiu:
[The previous bug](https://clusterfuzz.com/testcase-detail/6589725262151680) regarding `v8_script_parser_fuzzer` was fixed, indicating the correct ASAN options were applied.
However, ASAN options seem not applied on a different target [`html_in_process_fuzzer`](https://clusterfuzz.com/testcases?fuzzer=libFuzzer_html_in_process_fuzzer), even though the options on the report were correct.

This PR addresses this error, and removes some redundant code.

### DIFF
--- a/src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
@@ -188,7 +188,7 @@ class Engine(engine.Engine):
       A FuzzResult object.
     """
     runner = _get_runner(target_path)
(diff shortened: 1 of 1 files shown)


oliverchang (approved):

jonathanmetzman:
Is this ready?

jonathanmetzman:
ping

DonggeLiu:
> ping

Sorry that I missed this, @jonathanmetzman.
I don't recall why I discontinued this, but it appears to be ready but has a low priority.
Would it be safer to leave this for now so it won't affect production?

I can go back to this when I have more time.

PR #3212, opened 2023-07-11
Set up filters for utasks on pub/sub

jonathanmetzman:
Start setting up a system to allow linux workers to execute the preprocess and postprocess parts of utasks.
The first and last part of a utask can be executed on any platform since it only involves running the ClusterFuzz code which is cross-platform and not the fuzzers which are native code and platform-specific.
Therefore Linux workers will need to read *specific* task commands from non-linux queues.
This PR helps with that end by setting up filters to allow linux workers to read just preprocess and postprocess commands from non-linux queues. It also will be used by non-linux workers to ignore preprocess and postprocess commands from those same queues.
A notable change is that we need to seperate our definition of the type of each task from the execution (commands.py) of that task. This is because many tasks depend on the task scheduling module, which now needs to know the types of tasks to filter. Thus we seperate the types to avoid circular imports.


src/clusterfuzz/_internal/base/tasks.py
```diff
@@ -207,6 +211,31 @@ def get_regular_task(queue=None):
       return task
 
 
+def get_filters(is_chromium, is_linux):
```
oliverchang, line 214: Where is this being used?

src/clusterfuzz/_internal/bot/tasks/task_types.py
```diff
@@ -0,0 +1,161 @@
...
+    'symbolize': TrustedTask,
+    'unpack': TrustedTask,
+    'uworker_postprocess': PostprocessTask,
+    'upload_reports': TrustedTask,
+    'uworker_main': UworkerMainTask,
+    'variant': UTaskLocalExecutor,
+}
+
+
+def is_trusted_portion_of_utask(task_name):
```
oliverchang, line 148: Can you add docstrings? I'm also a little unclear on what this is doing exactly.

src/clusterfuzz/_internal/bot/tasks/commands.py
```diff
@@ -43,108 +43,38 @@
...
-    'minimize': utask_factory(minimize_task),
-    'progression': utask_factory(progression_task),
-    'regression': utask_factory(regression_task),
-    'symbolize': TrustedTask(symbolize_task),
-    'unpack': TrustedTask(unpack_task),
-    'postprocess': PostprocessTask(),
-    'upload_reports': TrustedTask(upload_reports_task),
-    'variant': utask_factory(variant_task),
-}
+def get_command_map():
```
oliverchang, line 74: Is there any reason to compute this lazily? Can we just do this on module import?

src/clusterfuzz/_internal/bot/tasks/task_types.py
```diff
@@ -0,0 +1,161 @@
...
+  the original ones in ClusterFuzz."""
+
+  def execute(self, task_argument, job_type, uworker_env):
+    # Simple tasks can just use the environment they don't need the uworker env.
+    del uworker_env
+    self.module.execute_task(task_argument, job_type)
+
+
+class UTask(BaseTask):
+  """Represents an untrusted task. Executes the preprocess part on this machine
```
oliverchang, line 43: This is the final state we want to be in for most platforms right?

src/clusterfuzz/_internal/base/tasks.py
jonathanmetzman: It's not being used yet.

src/clusterfuzz/_internal/bot/tasks/task_types.py
jonathanmetzman: Yes, it is the final state for all the tasks we plan to migrate. note that the only difference based on platform will be that things like mac bots will read from a queue to get their jobs instead of being scheduled on batch.

PR #3216, opened 2023-07-13
also catch rust error messages that have been caught and resumed

Ekleog-NEAR:
This, in particular, should make fuzzers built with [cargo-bolero](https://github.com/camshaft/bolero) able to get the proper error message auto-detected, rather than a generic abort.

PS: Hopefully this should be small enough to not require a CLA, as I am contributing on behalf of my employer and the process to get a CLA signed would be long for no reason. If you do need a CLA even for the smallest contributions, please feel free to let me know and I’ll see what I can do. But also don’t hesitate to do the change yourselves as though this were just a bug report, I’m just opening this to make your life simpler.


Ekleog-NEAR:
Turns out the CLA process was easier than I expected, so that’s now done. AFAICT the remaining CI check is just someone from google needing to look at the PR, so I can’t do anything about it.

jonathanmetzman:
We do require a CLA but I think I can probably make this contribution myself.

jonathanmetzman:
Could you add a test?

Ekleog-NEAR:
I just added a test, and ./local/tests/ci_tests.bash seems to pass, so hopefully this will be good to go :)

Ekleog-NEAR:
@jonathanmetzman Not sure whether you saw, but this is I think ready for review round 2 :)

Ekleog-NEAR:
@jonathanmetzman I don’t want to bother you so I’ll try just one last ping, I think this is ready for review :)

Anyway, thank you for the work you do on clusterfuzz, and hope this can help, whenever you manage to get to it!

jonathanmetzman (approved):
LGTM
@adetaylor @oliverchang WDYT of this change?

adetaylor (approved):

Ekleog-NEAR:
Welp seems like I don’t have permission to run this. Anyway, what I can say for sure is, `./local/tests/ci_tests.bash` passes locally :)

Ekleog-NEAR:
Actually, a thought: would it be better to get the line and char out of the summary? I’d just need to end the regex at `:` instead of end-of-line; and it might help with clusterfuzz automatically detecting that it’s the same fuzzer crash. OTOH, if it doesn’t actually help clusterfuzz detect that, then it’s probably better to just leave it, as it makes reading the crash easier.

Issue #3217, opened 2023-07-13
Optimize zipcorpora


jonathanmetzman:
Zip and upload in memory so we don't need to take more disk space.

jonathanmetzman:
Related: https://github.com/google/clusterfuzz/issues/3008

PR #3221, opened 2023-07-14
[zipcorpora] Zip zipcorpora lazily and in-memory 

jonathanmetzman:
Do this so that we don't waste disk space or run out of memory when
uploading corpora.

Related: https://github.com/google/clusterfuzz/issues/3217


oliverchang (approved):
very nice!!

Alan32Liu (approved):

PR #3236, opened 2023-07-17
[rearchitecture] Add functions for determining machine specs.

jonathanmetzman:
This will be used by tworkers to launch batch tasks that execute utask_main. Related: https://github.com/google/clusterfuzz/issues/3008

### DIFF
--- a/src/clusterfuzz/_internal/base/tasks.py
+++ b/src/clusterfuzz/_internal/base/tasks.py
@@ -250,8 +250,8 @@ def get_ttask_commands_queues():
 
 def get_machine_template_for_queue(queue_name):
   """Gets the machine template for the instance used to execute a task from
-  |queue_name|. This will be used by tworkers to schedule the appropriate machine using batch
-  to execute the utask_main part of a utask."""
+  |queue_name|. This will be used by tworkers to schedule the appropriate
+  machine using batch to execute the utask_main part of a utask."""
   initial_queue_name = queue_name
 
   # Handle it being high-end (preemptible) or not.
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
@oliverchang I think there's a conflation between high end tasks in the code. I.e. high end tasks can never be preempted. Non-preemptible tasks (I assume analyze) cannot be preempted (therefore must be high end). 
Is this understanding right?

jonathanmetzman:
TBR

oliverchang:
> 

I don't think so. High end means different things for OSS-fuzz vs others:
- OSS-Fuzz: high end means a set of hosts that have less workers. This means more cpu/ram to dedicate for each, and was needed for blackbox JS fuzzing.
- Others/Chromium: A set of bots with higher specs.

src/clusterfuzz/_internal/base/tasks.py
```diff
@@ -241,6 +242,51 @@ def get_utask_filters(is_chromium, is_linux):
...
+          f'Found machine template for {initial_queue_name}',
+          machine_template=template)
+      return template
+  return None
+
+
+def get_machine_templates():
+  """Returns machine templates."""
+  # TODO(metzman): Cache this.
+  clusters_config = local_config.Config(local_config.GCE_CLUSTERS_PATH).get()
```
oliverchang, line 284: We may want to replace this with a more specialized format later, as I want us to get rid of the whole `manage_vms` bits.

jonathanmetzman:
> > 
> 
> I don't think so. High end means different things for OSS-fuzz vs others:
> 
> * OSS-Fuzz: high end means a set of hosts that have less workers. This means more cpu/ram to dedicate for each, and was needed for blackbox JS fuzzing.
> * Others/Chromium: A set of bots with higher specs.

But this makes it seem like this is conflated: https://github.com/google/clusterfuzz/blob/aab96e6813877a307dfad7b75678fca6790cf041/src/clusterfuzz/_internal/base/tasks.py#L257

PR #3242, opened 2023-07-25
Fix bug in metrics due to upgrade

jonathanmetzman:


oliverchang (commented):
LGTM. Was there an API change to protobuf? I wonder if there are other instances we need to fix.

jonathanmetzman:
> LGTM. Was there an API change to protobuf? I wonder if there are other instances we need to fix.

It's not to protobuf it's to the monitoring package. I checked out their upgrading guide https://cloud.google.com/python/docs/reference/monitoring/latest/google.cloud.monitoring_v3.types.TimeSeries
but it seems like there may be some things they missed. I'm looking for a way to test this all now. For now, this fix (which I've verified) only helps.

PR #3247, opened 2023-07-25
Fix timestamps in metrics flushing.

jonathanmetzman:
Objects are now initialized to None by default, resulting in attribute errors if not explicitly initialized.


jonathanmetzman:
TBR

oliverchang (commented):
LGTM

PR #3249, opened 2023-07-25
Fix bug preventing updating protobuf library.

jonathanmetzman:
The protbuf library writes some write protected files that needed to be removed before written to.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/update_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/update_task.py
@@ -248,7 +248,12 @@ def update_source_code():
                      'version.' % absolute_filepath)
 
     try:
-      extracted_path = zip_archive.extract(filepath, output_directory)
+      try:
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/bot/tasks/update_task.py
```diff
@@ -248,7 +248,12 @@ def update_source_code():
                      'version.' % absolute_filepath)
 
     try:
-      extracted_path = zip_archive.extract(filepath, output_directory)
+      try:
```
jonathanmetzman, line 251: It's a bit ugly, but I can't really think of a better way to deal with this issue.

jonathanmetzman: I could remove every file first, but I don't know if this has any unintended consequences.

oliverchang: Another approach: Could we just chmod/chown every extracted first to ensure they're writable?

PR #3254, opened 2023-07-26
Try different strategy for fixing source updating

jonathanmetzman:
Try to make files writeable on unpack.


jonathanmetzman:
This doesn't work, because on startup the zipfile is unpacked by different means. We need to ensure the zip doesn't contain any readonly files.

oliverchang:
> This doesn't work, because on startup the zipfile is unpacked by different means. We need to ensure the zip doesn't contain any readonly files.

Hmm, sounds like a step we just need to do on the source deployment then?

jonathanmetzman:
Ha yep, PR up now.

jonathanmetzman:
Funny how we thought of each possible option in the same order.

PR #3255, opened 2023-07-27
Make sure we only package writeable files

jonathanmetzman:


jonathanmetzman:
Fingers crossed but I will be very surprised if this problem still exists after this lands.

oliverchang (approved):

PR #3263, opened 2023-07-31
Add Terraform scripts.

hogo6002:
Adding Terraform scripts to create a GKE cluster to migrate corn jobs from GAE to GKE


src/infra/terraform/cron_cluster/deploy.sh
```diff
@@ -0,0 +1,49 @@
...
+
+export TF_VAR_project_id=$project_id
+export TF_VAR_bucket_name=$bucket_name
+export TF_VAR_region=$region
+export TF_VAR_subnet_name=$subnet_name
+export TF_VAR_network_name=$network_name
+export TF_VAR_ip_cidr_range=${6:-'10.128.0.0/16'}
+
+# Checks if the subnet exists
+if ! terraform state show google_compute_subnetwork.subnet; then
```
oliverchang, line 40: Are these intended to temporary as we transition into terraform? Long term we should not need these right?

src/infra/terraform/cron_cluster/deploy.sh
```diff
@@ -0,0 +1,49 @@
...
+if [ $# -lt 2 ]; then
+  echo "Usage: $0 PROJECT_ID BUCKET_NAME [REGION] [subnet_name] [network_name] [ip_cidr_range]"
+  exit 1
+fi
+
+terraform init \
+  -backend-config="bucket=$2" \
+  -backend-config="prefix=terraform/state"
+
+project_id=$1
```
oliverchang, line 26: Should these all be encoded as part of clusterfuzz-config instead? I.e. the terraform vars would be part of the individual configs.

src/infra/terraform/cron_cluster/deploy.sh
```diff
@@ -0,0 +1,49 @@
...
+bucket_name=$2
+region=${3:-us-central1}
+subnet_name=${4:-us-central1}
+network_name=${5:-main}
+
+export TF_VAR_project_id=$project_id
```
Alan32Liu, line 32: nit: Use double-quotes (e.g., `export TF_VAR_project_id="$project_id"` in case there are spaces.

oliverchang:
Also see https://github.com/google/osv.dev/blob/master/deployment/terraform/environments/oss-vdb/main.tf as an example for how OSV separates out different deployments and shares a common set of configs that differ mostly only in some variables.

Alan32Liu (approved):

andrei-near:
There seem to be a gap in both documentation and code on generating the k8s and terraform configuration.
Is there an example somewhere how tf and k8s configuration should look like?

PR #3267, opened 2023-08-04
Fix exception in analyze.

oliverchang:


src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
```diff
@@ -45,6 +45,10 @@ def _add_default_issue_metadata(testcase):
 
   testcase_metadata = testcase.get_metadata()
   for key, default_value in default_metadata.items():
+    # Only string metadata are supported.
+    if not isinstance(default_value, str):
```
jonathanmetzman, line 49: Is this just a bandaid to stop bleeding? Or do i still need to figure out why a dict was passed here in the first place.
Also, it probably doesn't matter but I think parse delimited accepts file handles too.

jonathanmetzman (approved):
LGTM

src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
oliverchang: It's legitimate for a dict to be passed here, which we weren't accounting for. 

That said, I'm not sure if it makes sense for us to be handling these dict updates here in the same way as when they're comma delimited string values.

Alan32Liu (commented):
LGTM

Issue #3257, opened 2023-07-27
Replace download_as_string with download_as_bytes


jonathanmetzman:
download_as_string is deprecated.

MjMoshiri:
Hi @jonathanmetzman, there was only one case and it's now replaced. 🙏🏻

jonathanmetzman:
Actually looks like it's still here https://github.com/google/clusterfuzz/blob/c2de16f9c9ae209a56156f69431e0bc48e608969/src/clusterfuzz/_internal/google_cloud_utils/storage.py#L289

MjMoshiri:
Hello again,
[pull request](https://github.com/google/clusterfuzz/pull/3291)  above changes that line but it's not approved yet.

cdxvy30:
Sorry I am new to ClusterFuzz and trying to introduce it to my team. Does this issue mean that if I deployed a ClusterFuzz production setup with version 2.6.0 tag recently, reading files from Google Cloud Storage will not work?

I just deployed the ClusterFuzz to App Engine and can access to the <project-id>.appspot.com, also built and uploaded OpenSSL fuzzer target to ClusterFuzz and the bot running on the GCP did show the string "fuzz libFuzzer libfuzzer_asan_linux_openssl". However, the ClusterFuzz homepage (ie: the Testcases page) did not show the testcase titled “Heap-buffer-overflow READ{*}” as expected. Seems like it did not fuzz the target I uploaded.

jonathanmetzman:
This issue does not mean that. I'd look at the logs in google cloud to see why it's failed to produce crashes.

PR #3291, opened 2023-08-17
Replace download_as_string with download_as_bytes

MjMoshiri:
Fix issue #3257 by replacing ALLLLLLLLLLLLLLLLLL the download_as_string with download_as_bytes as it's  Deprecated.


src/clusterfuzz/_internal/google_cloud_utils/storage.py
```diff
@@ -286,7 +286,7 @@ def read_data(self, remote_path):
     try:
       bucket = client.bucket(bucket_name)
       blob = bucket.blob(path, chunk_size=self._chunk_size())
-      return blob.download_as_string()
+      return blob.download_as_bytes()
```
jonathanmetzman, line 289: Does decode need to be called on this result to return the same type?

PR #3292, opened 2023-08-17
Update Android code links

TheCoryBarker:
URL change after CS migration

### DIFF
--- a/src/clusterfuzz/_internal/build_management/source_mapper.py
+++ b/src/clusterfuzz/_internal/build_management/source_mapper.py
@@ -121,9 +121,9 @@ class GoogleSourceVCS(VCSViewer):
 
 class GoogleVCS(VCSViewer):
   VCS_URL_REGEX = re.compile(r'^//(.*)$')
-  VCS_REVISION_SUB = r'https://cs.corp.google.com/\1/?rcl={revision}'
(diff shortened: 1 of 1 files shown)


oliverchang (approved):

oliverchang:
This looks good, assuming that drc can be used in all places that rcl could have been used (for non-Android).

Can you please update the tests here too? Should be a simple fix.

TheCoryBarker:
I dont think this is interchangeable, this is dealing with code searches update to deprecate packages that should affect everyone (Chrome included). Is Chrome still able to use the legacy code search packages? The new link in Chrome should look something like this:

https://source.corp.google.com/h/chrome-internal/codesearch/chrome/src/+/main:<package>/<path>/<file>;l=1;bpv=1;bpt=0;drc=<commit>

it's a change to the sourcemaps file too. Here is my change to the sourcemaps file:
http://critique/557948970

TheCoryBarker:
Looks like there is a redirect in code search right now but it just isn't correct for Android. Maybe I'll just update the redirect to fix our link problems instead of updating the links we are providing. What do you think?

jonathanmetzman:
Do you want to fix the tests so we can land this?

jonathanmetzman:
ping

PR #3295, opened 2023-08-22
Add k8s deployment script.

hogo6002:


docker/ci/Dockerfile
```diff
@@ -39,6 +39,15 @@ RUN curl -s https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add
     echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list && \
     apt-get update && apt-get install -y google-chrome-stable
 
+# Install latest kubectl.
+RUN curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl" \
```
jonathanmetzman, line 43: If you do curl -O you can write it to a specific path directly.

jonathanmetzman, line 43: @oliverchang is this risky? I guess it's only in CI.

src/local/butler/common.py
```diff
@@ -378,6 +378,16 @@ def has_file_in_path(filename):
   return False
 
 
+def get_all_files(directory):
+  """Gets all files under a given directory."""
```
jonathanmetzman, line 382: Maybe add "recursively"

src/local/butler/common.py
```diff
@@ -378,6 +378,16 @@ def has_file_in_path(filename):
...
+  files = []
```
jonathanmetzman, line 383: lets make this a new variable `all_files` instead of overwritting the same variable in the loop I'm surprised this works

src/local/butler/deploy.py
```diff
@@ -419,9 +420,35 @@ def _prod_deployment_helper(config_dir,
...
+    _deploy_k8s()
   print('Production deployment finished.')
 
 
+def _deploy_terraform(config_dir):
+  """Deploys GKE cluster via terraform."""
+  terraform_dir = os.path.join(config_dir, 'terraform')
+  terraform = f'terraform -chdir=={terraform_dir}'
+  common.execute(f'{terraform} init')
+  # TODO: Remove terraform plan and set apply to auto-approve after testing.
```
jonathanmetzman, line 435: TODO(holly)

jonathanmetzman (approved):
lgtm

src/local/butler/common.py
hogo6002: Thanks for pointing out!

docker/ci/Dockerfile
oliverchang: We should install this via google cloud SDK: https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-access-for-kubectl# 

We already have the Cloud SDK APT repo set up.

Per the "apt" instructions on https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-access-for-kubectl#apt, we might be able to jsut do `apt-get install -y kubectl`?

Alan32Liu (approved):

oliverchang (approved):

PR #3296, opened 2023-08-23
Add redis memory store on k8s

hogo6002:


src/clusterfuzz/_internal/system/environment.py
```diff
@@ -1051,7 +1056,7 @@ def bot_noop(func):
 
   @functools.wraps(func)
   def wrapper(*args, **kwargs):
-    is_bot = not is_running_on_app_engine()
+    is_bot = not (is_running_on_app_engine() or is_running_on_k8s())
```
hogo6002, line 1059: I will pass 'REDIS_HOST' value to k8s script for testing purpose. After testing, the environment variable will be set during the deploying stage, same with app engine. https://github.com/google/clusterfuzz/blob/52db8878254a2bff7cfe5fce68815ab00bb64824/src/local/butler/deploy.py#L83

oliverchang: Instead of re-using `bot_noop`, we should create a new one:

```python
def if_redis_available(func):
  ...


@if_redis_available
def get(...)
```

`bot_noop` is meant for a different use case (i.e. are we a bot or not). better to separate out the intent of this new chekc.

infra/terraform/main.tf
```diff
@@ -72,3 +72,17 @@ resource "google_container_node_pool" "primary_nodes" {
     }
   }
 }
+
+resource "google_redis_instance" "memorystore_redis_instance" {
+  project        = var.project_id
+  name           = "k8s-redis"
+  tier           = "BASIC"
+  memory_size_gb = var.memory_size_gb
```
oliverchang, line 80: no need to make this configurable. Just put in the number here.

Also, let's use 16GB instead to give more room than 1GB.

oliverchang (approved):
LGTM with just one comment.

oliverchang (approved):

PR #3298, opened 2023-08-24
Fix release 260

cdxvy30:


cdxvy30:
I am trying to use tags/v2.6.0 with the tutorial of the documentation to deploy service in production setup. My environment is Ubuntu 20.04 with default Python 3.8.10. During setup in this environment, I found that the `clouddebugger.googleapis.com` in `/src/local/butler/create_config.py` is depreacated in 2023/05/31 so following the tutorial of the doc may not be able to build up the service.

src/local/butler/deploy.py
```diff
@@ -424,7 +424,7 @@ def _prod_deployment_helper(config_dir,
 
 def execute(args):
   """Deploy Clusterfuzz to Appengine."""
-  if sys.version_info.major != 3 or sys.version_info.minor != 7:
+  if sys.version_info.major != 3:
```
jonathanmetzman, line 427: Why remove the .7 check? I think things will fail without .7?

jonathanmetzman:
Can you please sign the CLA

src/local/butler/deploy.py
cdxvy30: In my workflow, I use Ubuntu 20.04 to setup the environment, and in `src/local/install_deps_linux.bash` it will check python installed or not. In Ubuntu 20.04, it has installed python3.8 by default, so `python3.8` would be exported as the environment variable. Therefore, when running to `src/local/butler/deploy.py`, the version check will fail.
I am not sure if it is the normal workflow.

cdxvy30:
Sure, I have just signed the CLA. Should I do anything else?

jonathanmetzman:
Ugh looks like there's an issue with the CLA check.

src/clusterfuzz/_internal/google_cloud_utils/storage.py
```diff
@@ -271,7 +271,7 @@ def read_data(self, remote_path):
     try:
       bucket = client.bucket(bucket_name)
       blob = bucket.blob(path, chunk_size=self._chunk_size())
-      return blob.download_as_string()
+      return blob.download_as_bytes()
```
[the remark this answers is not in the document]
cdxvy30: Sure, I have undo this change.

cdxvy30:
I have rescan my CLA sign. Thanks.

jonathanmetzman (approved):
lgtm

bkosciarz:
Is there a timeline for this going in and a new release coming out? I'm running into the same issue and it's annoying to patch because of the `is_git_dirty` check [here](https://github.com/google/clusterfuzz/blob/v2.6.0/src/local/butler/create_config.py#L240-L242)

PR #3304, opened 2023-08-30
Add module initializaion code for run_cron.py

hogo6002:


src/python/bot/startup/run_cron.py
```diff
@@ -42,6 +43,18 @@ def main():
...
+                                     'config', 'modules')
+  if os.path.exists(config_modules_path):
+    sys.path.append(config_modules_path)
+
+  try:
+    # Run any module initialization code.
+    import module_init_k8s
+    module_init_k8s.init()
+  except ImportError:
+    pass
```
Alan32Liu, line 56: nit: Would it be useful to print out an error message?

Alan32Liu (approved):

src/python/bot/startup/run_cron.py
hogo6002: It's from the [existing code](https://github.com/google/clusterfuzz/blob/543b83b6fbef34ff16e4fc60c647e900052290d3/src/appengine/main.py#L40).  We only want to import module when it is defined in config repo. For some instances, we don't have extra modules to import, so it fails by design, and no need to print error messages.

oliverchang (approved):

PR #3306, opened 2023-08-30
Add the ability to create queues in project_setup

marktefftech:


src/appengine/handlers/cron/project_setup.py
```diff
@@ -161,10 +165,16 @@ def job_name(self, project_name, config_suffix):
...
             'none': LIBFUZZER_NONE_I386_JOB,
         },
+        'arm': {
+            'hardware': LIBFUZZER_HWASAN_JOB,
+        }
     },
     'afl': {
         'x86_64': {
             'address': AFL_ASAN_JOB,
+        },
```
jonathanmetzman, line 175: nit: In Oss-fuzz we only try to make sure each target is fuzzed with AFL in one configuration (e.g. libfuzzer has x86 and x64 builds but afl only has x86). I wouldn't add more AFL configs otherwise, it can become a headache for devs/maintainers to deal with and has limited benefit.

src/appengine/handlers/cron/project_setup.py
```diff
@@ -547,7 +557,13 @@ def create_project_settings(project, info, service_account):
 
 def create_pubsub_topics(project):
   """Create pubsub topics for tasks."""
-  for platform in PUBSUB_PLATFORMS:
+  platforms = PUBSUB_PLATFORMS
+  if 'android_' in project:
+    platforms = ['android']
+    # Avoid the stutter in queue name
+    project = project.replace('android_', '')
```
jonathanmetzman, line 564: WOuld probably right this as
```
if 'android_' not in project:
  platforms = PUBSUB_PLATFORMS
else:
    platforms = ['android']
    # Avoid the stutter in queue name
    project = project.replace('android_', '')

src/appengine/handlers/cron/project_setup.py
marktefftech: Done.

src/appengine/handlers/cron/project_setup.py
marktefftech: Done.

marktefftech:
Closing in favor of 
https://github.com/google/clusterfuzz/pull/3362

PR #3324, opened 2023-09-08
Update proto generate script to use user path

pgrace-google:
I don't know bash very well so this might be butchering it!! feel free to change it yourself to something more bash-pro-esque


pgrace-google:
(or tell me how to fix it - happy to!!! just suggested the former since i thought you might prefer it)

jonathanmetzman:
> (or tell me how to fix it - happy to!!! just suggested the former since i thought you might prefer it)

Is the point of this change allowing one to call it from clusterfuzz instead of needing to cd into src/clusterfuzz/_internal/protos before calling it?

pgrace-google:
The goal is to replace the hardcoded ldap!

jonathanmetzman:
I've reverted to the state before my accidental change. https://github.com/google/clusterfuzz/pull/3089/files#diff-a09ba0bbef1454f94ab53d1bdde27063e64442fa0bbf8e33acbc35d6c7c52afd

PR #3327, opened 2023-09-11
Progression 1

alhijazi:


alhijazi:
@jonathanmetzman It seems a lot of changes are needed to be done. I've started with a first part here, could you please take a look

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -37,6 +37,117 @@
...
 from clusterfuzz._internal.system import environment
 
+from clusterfuzz._internal.protos import uworker_msg_pb2
+
+
+def handle_progression_timeout(uworker_output: uworker_io.UworkerOutput):
+  testcase_id = uworker_output.uworker_input.testcase_id
+  job_type = uworker_output.uworker_input.job_type
+  testcase = data_handler.get_testcase_by_id(testcase_id)
+  error_message = ('Timed out, current range r%d:r%d' %
```
jonathanmetzman, line 47: nit use an f-string instead.

jonathanmetzman (commented):
Thanks!
I did a first pass without looking at the main change (progression_task.py)

jonathanmetzman:
You should run `python butler.py format` and `python butler.py lint` and push again to fix some small mistakes.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -222,43 +315,53 @@ def _store_testcase_for_regression_testing(testcase, testcase_file_path):
...
+  _, testcase_file_path, error = setup.setup_testcase(testcase, job_type)
+  if error:
+    # TODO: reset progressing_pending in postrpocess in case of errors?
+    return None, error
 
   # Custom binaries are handled as special cases.
   if build_manager.is_custom_binary():
-    _check_fixed_for_custom_binary(testcase, job_type, testcase_file_path)
-    return
+    return _check_fixed_for_custom_binary(uworker_input, testcase, testcase_file_path)
```
jonathanmetzman, line 356: check_fixed_for_custom_binary returns one value but this function returns two?

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -222,43 +315,53 @@ def _store_testcase_for_regression_testing(testcase, testcase_file_path):
...
+  """Attempt to find the revision range where a testcase was fixed.""" 
+  deadline = tasks.get_task_completion_deadline()
+  testcase_id = uworker_input.testcase_id
+  testcase = uworker_input.testcase
+  job_type = uworker_input.job_type
+ 
+  # Setup testcase and its dependencies.
+  _, testcase_file_path, error = setup.setup_testcase(testcase, job_type)
+  if error:
+    # TODO: reset progressing_pending in postrpocess in case of errors?
```
jonathanmetzman, line 351: nit: Please put a name or issue in TODOs.
See https://google.github.io/styleguide/pyguide.html#312-todo-comments

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -285,23 +389,33 @@ def find_fixed_range(testcase_id, job_type):
...
 
   # Check to see if this testcase is still crashing now. If it is, then just
   # bail out.
-  result = _testcase_reproduces_in_revision(
+  result, error = _testcase_reproduces_in_revision(
       testcase,
       testcase_file_path,
       job_type,
       max_revision,
+      # Changed by me to false
```
jonathanmetzman, line 414: What does this comment mean?
Also nit: End comments with a period please.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -312,27 +426,25 @@ def find_fixed_range(testcase_id, job_type):
...
+    testcase.crash_state = state.frames
+    testcase.crash_revision = max_revision
+    error_message = 'still crashes on latest revision r%s' % max_revision
+    return uworker_io.UworkerOutput(
+          testcase=testcase,
+          uworker_input=uworker_input,
+          error_message=error_message,
+          error=uworker_msg_pb2.ErrorType.PROGRESSION_CRASH_ON_LATEST)  
+
+  # TODO: these don't seem to be used anywhere.
```
jonathanmetzman, line 446: @oliverchang I think crashes_on_unexpected_state is not used anywhere. Can we get rid of it?
Ali I guess if we get rid of it, we shoudl get rid of result.unexpected_crash as it isn't used for any other purpose.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -343,28 +455,15 @@ def find_fixed_range(testcase_id, job_type):
...
-
-    data_handler.clear_progression_pending(testcase)
-    error_message = (
-        'Known crash revision %d did not crash' % known_crash_revision)
-    data_handler.update_testcase_comment(testcase, data_types.TaskState.ERROR,
-                                         error_message)
-    task_creation.mark_unreproducible_if_flaky(testcase, True)
-    return
-
+    error_message = ('Known crash revision %d did not crash' % known_crash_revision)
```
jonathanmetzman, line 458: Let's use an f-string.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -343,28 +455,15 @@ def find_fixed_range(testcase_id, job_type):
...
+    testcase.crash_revision = max_revision
+    return uworker_io.UworkerOutput(
+          testcase=testcase,
+          uworker_input=uworker_input,
+          error_message=error_message,
+          error=uworker_msg_pb2.ErrorType.PROGRESSION_NO_CRASH)  
+
+  # TODO: this part is to be moved into a separate task to be scheduled on postprocess.
```
jonathanmetzman, line 466: nit: Capital T for "this" at the beginning of TODO.
Similar to https://google.github.io/styleguide/pyguide.html#312-todo-comments

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -403,83 +498,72 @@ def find_fixed_range(testcase_id, job_type):
...
 
+    result, error = _testcase_reproduces_in_revision(testcase, testcase_file_path,
+                                                job_type, middle_revision)
+    if error is not None:
+      if error.error == uworker_msg_pb2.ErrorType.PROGRESSION_BAD_BUILD:
+        # Skip this revision.
+        del revision_list[middle_index]
+        max_index -= 1
+        continue
+      # only bad build errors are recoverable.
```
jonathanmetzman, line 510: nit: start comments with a capital letter.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -403,83 +498,72 @@ def find_fixed_range(testcase_id, job_type):
...
+def utask_main(uworker_input):
+  """Execute progression task."""
+  return find_fixed_range(uworker_input)
 
 
 def utask_postprocess(output):
-  del output
-
+  if output.error is not None:
+    # TODO(https://github.com/google/clusterfuzz/issues/3008): Change this when
```
jonathanmetzman, line 541: This is the change. The TODO is completed and outdated now. Please remove it :-)

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -403,83 +498,72 @@ def find_fixed_range(testcase_id, job_type):
...
+    # progression is migrated.
+    all_errors = uworker_handle_errors.get_all_handled_errors()
+    uworker_handle_errors.handle(output.error, all_errors)
+    return
+  
+  if output.custom_binary is True:
```
jonathanmetzman, line 547: Not clear to me why we only do this for a custom binary, I think this wasn't considered in the old code. Can you explain why please?

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -403,83 +498,72 @@ def find_fixed_range(testcase_id, job_type):
...
+    # Retry once on another bot to confirm our results and in case this bot is in
+    # a bad state which we didn't catch through our usual means.
+    testcase = data_handler.get_testcase_by_id(output.uworker_input.testcase_id)
```
jonathanmetzman, line 550: Do we modify output.testcase at all in uworker_main? if we do then we probably should use it and `put()` it rather than fetching the stale version.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -403,83 +498,72 @@ def find_fixed_range(testcase_id, job_type):
...
+        testcase, output.testcase.crash_revision, message='fixed on latest custom build')
+    return
 
+  # TODO: This should probably be moved to the end of the newly created progression_bisection task
   # If there is a fine grained bisection service available, request it. Both
   # regression and fixed ranges are requested once. Regression is also requested
   # here as the bisection service may require details that are not yet available
   # (e.g. issue ID) at the time regress_task completes.
-  testcase = data_handler.get_testcase_by_id(uworker_input.testcase_id)
+  testcase = data_handler.get_testcase_by_id(output.testcase.key.id())
```
jonathanmetzman, line 568: lets use output.uworker_input.testcase_id for this.
We don't want to trust the key here as it could allow the untrusted worker to mess with other testcases.

jonathanmetzman:
Can you see this test failure: https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/e3595d19-fe96-4695-b02d-ec62d4549e30;step=4?project=clusterfuzz-testing-public

I think they mean progression task is broken by this change (not clear to me how though).

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
alhijazi: You can ignore this, left here by mistake.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
jonathanmetzman: Ping.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -403,83 +525,93 @@ def find_fixed_range(testcase_id, job_type):
...
 
 def utask_postprocess(output):
-  del output
+  """Trusted: Cleans up after a uworker execute_task, writing anything needed to
+  the db."""
+  if output.error is not None:
+    uworker_handle_errors.handle(output, HANDLED_ERRORS)
+    return
 
+  if output.progression_task_output.crash_on_latest is True:
```
jonathanmetzman, line 586: I usually do `if blah` instead of `if blah is True`

jonathanmetzman (commented):
I think there were some comments that weren't addresses from last round.

alhijazi:
> I think there were some comments that weren't addresses from last round.

Hello, indeed, I did not resolve the comments I didn't address yet. I'm working on them currently.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
alhijazi: Should be okay now.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
alhijazi: These come from the initial implementation of `_check_fixed_for_custom_binary`
You can take a look at: https://github.com/google/clusterfuzz/pull/3327/files#diff-e67a9c5a80de5ba8fcd9fabf0d42884679d4920f321be15b70088c85d0ee75bbR235-R239
Where I moved the parts requiring privileged access to postprocess.
Initially, if we detect that the testcase is fixed on the custom binary, we retry once then mark the testcase as fixed.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
alhijazi: Indeed, I'll fix that.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
alhijazi: I'm going to modify all instances to use `output.uworker_input.testcase_id`

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -37,6 +37,117 @@
...
+  data_handler.update_testcase_comment(testcase, data_types.TaskState.ERROR,
+                                        uworker_output.error_message)
+  
+def handle_revision_list_error(uworker_output: uworker_io.UworkerOutput):
+  testcase = data_handler.get_testcase_by_id(uworker_output.testcase.key.id())    
+  data_handler.close_testcase_with_error(testcase,
+                                          'Failed to fetch revision list')
+  return
+
+def handle_crash_on_latest(uworker_output: uworker_io.UworkerOutput):
```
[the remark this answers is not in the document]
alhijazi: I kept the method but now it's directly called from postprocess.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
alhijazi: Looking at the initial behaviour, It seems we were refetching the testcase after running `testcase_manager.test_for_crash_with_retries`.
The refetch call was here: https://github.com/google/clusterfuzz/pull/3327/files#diff-e67a9c5a80de5ba8fcd9fabf0d42884679d4920f321be15b70088c85d0ee75bbL86
 Was there a particular reason for refetching often?

alhijazi:
> > > I think there were some comments that weren't addresses from last round.
> > 
> > 
> > Hello, indeed, I did not resolve the comments I didn't address yet. I'm working on them currently.
> 
> Great. Just let me know when you are ready for another review

I pushed a small update. I think it's ready for you to take a look. Hoping I didn't miss any of your previous comments!

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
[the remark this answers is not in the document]
jonathanmetzman: Maybe just omit the word "handle_" then. It seems kind of confusing as is.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -139,22 +256,29 @@ def _update_issue_metadata(testcase):
...
 
 
-def _testcase_reproduces_in_revision(testcase,
+def _testcase_reproduces_in_revision(uworker_input,
+                                     testcase,
                                      testcase_file_path,
                                      job_type,
                                      revision,
                                      update_metadata=False):
   """Test to see if a test case reproduces in the specified revision."""
```
jonathanmetzman, line 265: nit: can you document the return values now?

jonathanmetzman (commented):
Looks pretty good now. Will take another look tomorrow AM

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -35,9 +34,143 @@
...
+  data_handler.update_progression_completion_metadata(
+      testcase,
+      uworker_output.progression_task_output.max_revision,
+      message=message)
+
+  # Let the bisection service know about the NA status.
+  bisection.request_bisection(testcase)
+
+
+def handle_no_crash(uworker_output: uworker_io.UworkerOutput):
```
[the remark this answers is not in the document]
alhijazi: If you prefer, we can treat this in a different way than errors. We  still need to set some binary flag in the returned uworker output to signal that an expected crash revision no longer crashes.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -222,43 +345,56 @@ def _store_testcase_for_regression_testing(testcase, testcase_file_path):
...
   build_bucket_path = build_manager.get_primary_bucket_path()
   revision_list = build_manager.get_revisions_list(
       build_bucket_path, testcase=testcase)
   if not revision_list:
-    data_handler.close_testcase_with_error(testcase,
-                                           'Failed to fetch revision list')
-    return
+    return uworker_io.UworkerOutput(
+        testcase=testcase,
+        uworker_input=uworker_input,
```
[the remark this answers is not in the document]
alhijazi: so what you're saying is, I only need to set `uworker_input` when I pass to other non-trusted modules right? This means that for all error handling reporting stuff, we don't need to set `uworker_input`.

The reason I thought it's necessary is that we are creating `uworker_input` manually in `setup_testcase` and returning it to the trusted worker on failure.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -285,86 +421,92 @@ def find_fixed_range(testcase_id, job_type):
...
-    return
+    testcase.last_tested_crash_stacktrace = data_handler.filter_stacktrace(
+        stacktrace)
 
+    error_message = f'Still crashes on latest revision r{max_revision}.'
+    progression_task_output = uworker_io.ProgressionTaskOutput(
+        crash_on_latest=True,
+        crash_type=state.crash_type,
+        crash_address=state.crash_address,
+        # TODO(alhijazi): Check how to properly serialize the frames.
```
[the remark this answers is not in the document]
alhijazi: I had created a field in the ProgressionTaskOutput for `crash_frames`. I chose string as a type but I'm not sure it's the appropriate type. I wanted to check with you before landing this.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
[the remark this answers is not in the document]
jonathanmetzman: > so what you're saying is, I only need to set `uworker_input` when I pass to other non-trusted modules right? 

Maybe, It's only useful when the caller isn't migrated yet.
The correct behavior when you get a uworker_output containing an error is to return from main and let postprocess handle it. In that case we don't need to add uworker_input. Because when we download uworker_output, we also download uworker_input. However, for unmigrated users that aren't yet behaving properly, we need to include uworker_input.

>This means that for all error handling reporting stuff, we don't need to set `uworker_input`.

Yes, if the caller is migrated properly, then we don't need to set it.

> 
> The reason I thought it's necessary is that we are creating `uworker_input` manually in `setup_testcase` and returning it to the trusted worker on failure.

Right, setup_testcase has many uunmigrated callers which is why we need this.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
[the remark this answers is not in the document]
jonathanmetzman: I think these are StackFrameStructure.
1. I think you should convert these to a dictionary (`__dict__` method probably works)
2. make crash_frames a JSON field. 
3. Override ProgressionTaskOutput's save_rich_type method to include this (it's done elsewhere in the file, use inspiration)
```
if isinstance(value, (dict, list)):
      save_json_field(field, value)
      return
```
4. In postprocess, convert back from the dict to StackFrameStructure
5. Test this very thouroghly!!!

jonathanmetzman (commented):
I reviewed this pretty well.
I think once you solve the `StackFrameStructure` issue and add some tests (for other things as well if you can please) I can land this tomorrow then.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
jonathanmetzman: ACK

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -378,112 +525,115 @@ def find_fixed_range(testcase_id, job_type):
...
-                                                job_type, middle_revision)
-    except errors.BadBuildError:
-      # Skip this revision.
-      del revision_list[middle_index]
-      max_index -= 1
-      continue
+    result, error = _testcase_reproduces_in_revision(
+        uworker_input, testcase, testcase_file_path, job_type, middle_revision)
+    if error is not None:
+      if error.error == uworker_msg_pb2.ErrorType.PROGRESSION_BAD_BUILD:
```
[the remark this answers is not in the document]
alhijazi: I don't think that would work.
I think `uworker_output::error` (ErrorType) should be renamed to error_type to avoid confusion.
We need to first check if error (which is a uworker_output) is not null. In that case we need to check error.error whether it is a `PROGRESSION_BAD_BUILD`.
Did I misunderstand something?

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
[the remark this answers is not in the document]
alhijazi: `state.crash_type`, `state.crash_address`, `state.frames` are only used to create `serialized_crash_stack_frames` in `crash_uploader.save_crash_info_if_needed` using `crash_uploader.get_symbolized_stack_bytes`.
So instead of passing these three variables and going through the trouble of serializing/deserializing them. I modified `save_crash_info_if_needed` to take the serialized frames. We now call `get_symbolized_stack_bytes` in the utask and pass the result to the trusted worker using progression_task_output.
`get_symbolized_stack_bytes` returns `process_state.SerializeToString()` (a string).

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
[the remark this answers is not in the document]
alhijazi: `uworker_input` not set anymore in `uworker_output` from utask_main

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -97,12 +223,13 @@ def _check_fixed_for_custom_binary(testcase, job_type, testcase_file_path):
...
     testcase.last_tested_crash_stacktrace = data_handler.filter_stacktrace(
         stacktrace)
-    data_handler.update_progression_completion_metadata(
-        testcase,
-        revision,
-        is_crash=True,
-        message='still crashes on latest custom build')
-    return
+    progression_task_output = uworker_io.ProgressionTaskOutput(
+        custom_binary=True, crash_on_latest=True, crash_revision=revision)
```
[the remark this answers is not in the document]
alhijazi: I was not sure if `build_manager.is_custom_binary()` needed `setup.setup_testcase` to be called before so that the env has all the needed variables. 
Otherwise, we could probably look for the related environment variables in the `uworker_env` directly in preprocess. WDYT?

src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
```diff
@@ -96,6 +96,8 @@ def setup_build(testcase: data_types.Testcase,
 
     revision_index = revisions.find_min_revision_index(revision_list, revision)
     if revision_index is None:
+      # TODO(https://github.com/google/clusterfuzz/issues/3008): Move the update
```
jonathanmetzman, line 99: Have a PR to make this comment unneeded.

src/clusterfuzz/_internal/chrome/crash_uploader.py
```diff
@@ -566,14 +566,12 @@ def get_symbolized_stack_bytes(crash_type, crash_address, symbolized_stack):
     return None
 
 
-def save_crash_info_if_needed(testcase_id, crash_revision, job_type, crash_type,
-                              crash_address, crash_frames):
+def save_crash_info_if_needed(testcase_id, crash_revision, job_type,
```
jonathanmetzman, line 569: Might be better to shelve changes for crash_uploader for now. I'm going to delete it (today probably)

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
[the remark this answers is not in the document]
jonathanmetzman: Youre right it doesn't work.
Feel free to make the error_type change in new PR.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -291,86 +425,86 @@ def find_fixed_range(testcase_id, job_type):
...
-                                             state.crash_address, state.frames)
-    return
+    testcase.last_tested_crash_stacktrace = data_handler.filter_stacktrace(
+        stacktrace)
 
+    error_message = f'Still crashes on latest revision r{max_revision}.'
+    serialized_crash_stack_frames = crash_uploader.get_symbolized_stack_bytes(
+        state.crash_type, state.crash_address, state.frames)
+    progression_task_output = uworker_io.ProgressionTaskOutput(
+        crash_on_latest=True,
```
jonathanmetzman, line 473: Can you make a crash_on_latest_message field and use that instead of (mis)using error message please.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -97,12 +221,12 @@ def _check_fixed_for_custom_binary(testcase, job_type, testcase_file_path):
         command, symbolized_crash_stacktrace, unsymbolized_crash_stacktrace)
     testcase.last_tested_crash_stacktrace = data_handler.filter_stacktrace(
```
jonathanmetzman, line 222: nit: let's pass this field directly instead of through testcase please.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -291,86 +425,86 @@ def find_fixed_range(testcase_id, job_type):
...
-    data_handler.clear_progression_pending(testcase)
+  if result and not result.is_crash():  # pylint: disable=no-member
     error_message = (
-        'Known crash revision %d did not crash' % known_crash_revision)
-    data_handler.update_testcase_comment(testcase, data_types.TaskState.ERROR,
-                                         error_message)
-    task_creation.mark_unreproducible_if_flaky(testcase, True)
-    return
-
+        f'Known crash revision {known_crash_revision} did not crash')
```
jonathanmetzman, line 497: I wouldn't bother with error_message here it doesn't add much, can you just repeat that string twice in the error handler?

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -291,86 +425,86 @@ def find_fixed_range(testcase_id, job_type):
...
+    progression_task_output = uworker_io.ProgressionTaskOutput(
+        crash_revision=max_revision)
+    return uworker_io.UworkerOutput(
+        testcase=testcase,
+        progression_task_output=progression_task_output,
+        error_message=error_message,
+        error=uworker_msg_pb2.ErrorType.PROGRESSION_NO_CRASH)
+
+  # TODO(alhijazi): This part is to be moved into a separate task to be
```
jonathanmetzman, line 506: I don't see why we can't return details about multiple revisions from one main_task in the future.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -52,39 +188,27 @@ def _write_to_bigquery(testcase, progression_range_start,
...
-        testcase, data_types.TaskState.ERROR,
-        'Build setup failed for custom binary')
-    build_fail_wait = environment.get_value('FAIL_WAIT')
-    tasks.add_task(
-        'progression', testcase_id, job_type, wait_time=build_fail_wait)
-    return
+    return uworker_io.UworkerOutput(
+        testcase=testcase,
+        error_message='Build setup failed for custom binary',
+        error=uworker_msg_pb2.ErrorType.PROGRESSION_BUILD_SETUP)
```
jonathanmetzman, line 205: Maybe call this PROGESSION_BUILD_SETUP_ERROR.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -144,17 +260,23 @@ def _testcase_reproduces_in_revision(testcase,
...
 
   if testcase_manager.check_for_bad_build(job_type, revision):
-    log_message = 'Bad build at r%d. Skipping' % revision
-    testcase = data_handler.get_testcase_by_id(testcase.key.id())
-    data_handler.update_testcase_comment(testcase, data_types.TaskState.WIP,
-                                         log_message)
-    raise errors.BadBuildError(revision, job_type)
+    # TODO(alhijazi): This is not logged for recoverable builds.
+    log_message = f'Bad build at r{revision}. Skipping'
```
jonathanmetzman, line 275: nit: error_message not log_message.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -144,17 +260,23 @@ def _testcase_reproduces_in_revision(testcase,
                                      job_type,
                                      revision,
                                      update_metadata=False):
-  """Test to see if a test case reproduces in the specified revision."""
+  """Test to see if a test case reproduces in the specified revision.
```
jonathanmetzman, line 263: nit: "Tests"

jonathanmetzman (commented):
Pretty close to landing!

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
alhijazi: Removed.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
alhijazi: Done.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
alhijazi: I'm not sure I understand what string to repeat and where?

PR #3328, opened 2023-09-12
Fix CF update.

oliverchang:
We were unzipping to the wrong paths after
https://github.com/google/clusterfuzz/commit/572b256007fa666ce54c942a0f35c8199ece3e39.

i.e. files were getting unzipped to clusterfuzz/clusterfuzz rather than clusterfuzz. This meant that bots got into an 
unrecoverable state after the first update.

Fix up the incorrect base paths and revert a variable name change to make things clearer.


hogo6002 (approved):

jonathanmetzman:
LGTM.
Thanks!

PR #3333, opened 2023-09-18
Prioritized 'Timeout' crash_type; Added hotlisting based on fuzzer crash metadata

davfu1:
Change 1:
Enabled ranking of crash types where a crash may be eligible for multiple types; Set 'Timeout' as a higher priority crash_type. 

stacktraces.parse() sequentially applies series of regexes to assign crash_type, such that a crash_type discerned from a line later in the stacktrace will overwrite a crash_type from an earlier line. This CL aims to add crash_type ranking, which allows us to specify which crash_type prevails in the event of multiple crash_types. For example, the following were libFuzzer timeouts filed as 'Crash' or 'CHECK failure': 

https://clusterfuzz.corp.google.com/testcase-detail/5445029307744256
https://clusterfuzz.corp.google.com/testcase-detail/5634783893389312

Ideally these testcases would have been filed as 'Timeout' to best indicate the non-security impact of the crashes. This CL specifically assigns Timeout a higher ranking than types 'UNKNOWN' and 'CHECK failure', such that the 'Timeout' crash_type would be reported in a crash satisfying multiple regexes.

Change 2:
Enabled crash metadata-based labels/hotlists, adding the field 'crash_subtypes' to crash object definitions to carry metadata to issue_filer.py. Added an initial crash_subtype of 'Fuzzer exit' to describe crashes with stacktraces indicating a crash location inside a fuzzer itself, as opposed to a fuzzed library. Example:

https://clusterfuzz.corp.google.com/testcase-detail/5848511005720576

Fuzzer exits have been a good heuristic for 'No security impact' rating due to likely fuzzer design or infrastructure error. We intend to use the hotlist in Android to reroute such bugs from ISE, towards feature teams. This change is intended to accompany a separate change to configs/google/issue_trackers/config.yaml.


jonathanmetzman:
I'm not so comfortable with this change, why is a crash that is ambiguously a real crash or a timeout "non-security".
Can you add a testcase for the case you have in mind.
@oliverchang WDYT?

oliverchang:
> I'm not so comfortable with this change, why is a crash that is ambiguously a real crash or a timeout "non-security". Can you add a testcase for the case you have in mind. @oliverchang WDYT?

This seems like it's just addressing a prioritization issue with our stacktrace parsing? i.e. some timeouts throw a signal, and we report the signal instead of the timeout. If that's the case, that seems fine to me.

+1 to adding a test for this to https://github.com/google/clusterfuzz/blob/961bdc55473a8284ae611ecb07415b7326f1501c/src/clusterfuzz/_internal/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py

davfu1:
Thanks @jonathanmetzman and @oliverchang for the comments - I'd confirm what Oliver was saying, that this CL hopes to address what to do if a crash has multiple signals for crash_type, where one of those signals suggests timeout. Currently Clusterfuzz just selects the crash_type most recently suggested by the latest line in a stacktrace, including using latest lines from Android logcats. I've added the test now, and modified how the CL's __init__.py figures out to prioritize the Timeout crash_type.

src/clusterfuzz/stacktraces/__init__.py
```diff
@@ -146,7 +146,7 @@ def ignore_stack_frame(self, stack_frame):
       return True
 
     return False
-  
+
```
jonathanmetzman, line 149: Is this change adding a tab or something?

src/clusterfuzz/_internal/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
```diff
@@ -3595,15 +3595,16 @@ def test_symbolized_trusty_stacktrace(self):
     self._validate_get_crash_data(data, expected_type, expected_address,
                                   expected_state, expected_stacktrace,
                                   expected_security_flag)
-  
+
```
jonathanmetzman, line 3598: SImilarly, what's happening here.

src/clusterfuzz/stacktraces/__init__.py
davfu1: That's strange Jonathan, I don't see anything in the "Files Changed" diff for lines 149 or 3598.

jonathanmetzman: Ha must've been an old commit.

src/clusterfuzz/stacktraces/__init__.py
```diff
@@ -173,7 +179,8 @@ def update_state_on_match(self,
       state.frame_count = 0
 
     # Direct updates.
-    if new_type is not None:
+    if new_type is not None and self.get_rank(new_type) >= self.get_rank(
```
jonathanmetzman, line 182: Can you comment that we do `>=` because we want later types to be prioritized?

jonathanmetzman (commented):
I'm not sure I'm comfortable with the goal of prioritizing timeouts over non for users other than Android (though I suspect it's fine).
I can mege if oliver think's it's OK with this

src/clusterfuzz/stacktraces/__init__.py
davfu1: Thanks, I added a comment explaining the prioritization.

davfu1:
> I'm not sure I'm comfortable with the goal of prioritizing timeouts over non for users other than Android (though I suspect it's fine). I can mege if oliver think's it's OK with this

Hi Jonathan, thanks for all the attention to this CL so far! Over a Sept 25 e-mail Oliver had said 'The change itself LGTM' in asking a question to us. I would propose that the current priority is almost random (based on which line was latest in the entire testcase report) and so adding a different priority scheme might not significantly impact intended results.

oliverchang (approved):
This LGTM.

@jonathanmetzman do you see any risks here? This seems pretty low risk to me, given that the case this is addressing is when there's 2 stacktraces talking about the same timeout. i.e. a timeout stack given by libFuzzer, but also some other signal handler that caught the signal shutting down the fuzzer and reporting it as SIGABRT or something else.

In such cases, it makes sense to me to report it as a timeout. It seems unlikely there would be some other legitimate and unrelated stacktrace in the same report that we'd mask.

jonathanmetzman:
> This LGTM.
> 
> @jonathanmetzman do you see any risks here? This seems pretty low risk to me, given that the case this is addressing is when there's 2 stacktraces talking about the same timeout. i.e. a timeout stack given by libFuzzer, but also some other signal handler that caught the signal shutting down the fuzzer and reporting it as SIGABRT or something else.

I guess if a program catches its own signals and does some cleanup, maybe it could be an issue (not sure about order of of signal handlers)

> 
> In such cases, it makes sense to me to report it as a timeout. It seems unlikely there would be some other legitimate and unrelated stacktrace in the same report that we'd mask.

davfu1:
Hi @jonathanmetzman, @oliverchang - Please note I've significantly changed this PR to include the discussed hotlisting of crash metadata such as 'Fuzzer exit' (crashes with stacktraces indicating a crash within the fuzzer itself). I couldn't create a separate PR while the Timeout one was in place, due to only one fork being allowed at a time per Github user.

jonathanmetzman:
Can you fix the failing test. I'll try to merge this monday if it's ready by then.

davfu1:
> Can you fix the failing test. I'll try to merge this monday if it's ready by then.

Thanks, Jonathan. I've fixed the test. I do want to highlight that the new changes (see updated description and proto changes) add a new field to the Testcase() stored by NDB - from what I've read this shouldn't be an issue due to NDB being NoSQL, but I haven't worked with Clusterfuzz or Google Datastore before.

src/clusterfuzz/_internal/tests/appengine/libs/issue_filer_test.py
```diff
@@ -826,6 +832,37 @@ def test_crash_type_labels(self):
...
+            'new': 'Untriaged',
+            'wontfix': 'WontFix',
+            'fixed': 'Fixed'
+        },
+        'all': {
+            'status': 'new',
+        },
+        'non_security': {},
+        'labels': {
+            'crash_subtypes': {
```
oliverchang, line 851: How about crash_categories? subtypes means something different to me from what's being intended here.

src/clusterfuzz/_internal/issue_management/issue_filer.py
```diff
@@ -455,6 +455,11 @@ def file_issue(testcase,
 
   for cc in ccs:
     issue.ccs.add(cc)
+  # Add labels from crash metadata.
+  for crash_subtype in testcase.crash_subtypes:
```
oliverchang, line 459: Could we place this closer to the existing crash_type handling on line 370?

src/clusterfuzz/stacktraces/constants.py
```diff
@@ -114,6 +114,8 @@
 FATAL_ERROR_REGEX = re.compile(r'#\s*Fatal error in (.*)')
 FATAL_ERROR_LINE_REGEX = re.compile(r'#\s*Fatal error in (.*), line [0-9]+')
 FATAL_ERROR_UNREACHABLE = re.compile(r'# un(reachable|implemented) code')
+FUZZER_EXIT_REGEX = re.compile(r'^    (?:#0|#1) 0x.*(?:fuzzer|fuzz/|/fuzz)',
```
oliverchang, line 117: This seems a bit hacky. Are there any other alternatives we can be doing here to extract this in a more reliable way?

src/clusterfuzz/_internal/datastore/data_types.py
```diff
@@ -375,6 +375,9 @@ class Testcase(Model):
   # Complete stacktrace.
   crash_stacktrace = ndb.TextProperty(indexed=False)
 
+  # Hotlists associated with the crash.
+  crash_subtypes = ndb.StringProperty(repeated=True)
```
oliverchang, line 379: Rather than a full blown Testcase property (which is useful when we need to add a filter etc, could we instead use ` testcase.set_metadata` ? 

(Also, I think the comment is wrong :) )

CC @jonathanmetzman for thoughts too.

src/clusterfuzz/_internal/tests/appengine/libs/issue_filer_test.py
davfu1: Thanks Oliver, I've changed this variable name.

src/clusterfuzz/_internal/issue_management/issue_filer.py
davfu1: Done.

src/clusterfuzz/_internal/datastore/data_types.py
davfu1: Thanks for the suggestion! I refactored the code to use set_metadata instead of a new Testcase property.

src/clusterfuzz/stacktraces/constants.py
davfu1: Thanks Oliver - were you thinking that this might be under/overinclusive? 

The intention was to identify early stacktrace references to fuzzing or a fuzzer (here, frame 0 or 1). I know it looks a bit strange, as this is a heuristic as opposed to something more definite, but applied across all testcases this regex seem to have hit only the right targets. Our analysis of some several hundred fuzzer design/infra-related Android crashes has shown that this kind of exit has been consistently evaluated by our ISE as having ‘No Security Impact’, often with the reasoning that the crash appeared to be within the fuzzer itself or caused by poor fuzzer design. For example,

https://clusterfuzz.corp.google.com/testcase-detail/5379218936561664
https://clusterfuzz.corp.google.com/testcase-detail/6097477842894848
https://clusterfuzz.corp.google.com/testcase-detail/5162986195779584
https://clusterfuzz.corp.google.com/testcase-detail/6489775183101952
https://clusterfuzz.corp.google.com/testcase-detail/5848511005720576

oliverchang: My main concern was that this seemed a bit fragile and hacky. If you've done a lot of analysis here and this works for you, LGTM.

oliverchang (approved):
LGTM. @jonathanmetzman for final signoff and merge.

davfu1:
Hi @jonathanmetzman - I was hoping to kindly bump this again for your attention, if you have the chance.

src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -785,6 +785,10 @@ def store_testcase(crash, fuzzed_keys, minimized_keys, regression, fixed,
   # Set metadata fields (e.g. build url, build key, platform string, etc).
   set_initial_testcase_metadata(testcase)
 
+  # Set crash metadata.
+  if hasattr(crash, 'crash_categories') and crash.crash_categories:
```
jonathanmetzman, line 789: Do we really need this hasattr check? I think it should always be around now right?

src/clusterfuzz/_internal/issue_management/issue_filer.py
```diff
@@ -372,6 +372,14 @@ def file_issue(testcase,
   if crash_type_label:
     issue.labels.add(crash_type_label)
 
+  # Add labels from crash metadata.
+  crash_categories = testcase.get_metadata('crash_categories')
+  if crash_categories:
+    for crash_category in crash_categories:
```
jonathanmetzman, line 378: Replace this with
```
for crash_category in testcase.get_metadata('crash_categories', []):

File: src/clusterfuzz/stacktraces/constants.py [hunk 10]
```diff
@@ -114,6 +114,8 @@
 FATAL_ERROR_REGEX = re.compile(r'#\s*Fatal error in (.*)')
 FATAL_ERROR_LINE_REGEX = re.compile(r'#\s*Fatal error in (.*), line [0-9]+')
 FATAL_ERROR_UNREACHABLE = re.compile(r'# un(reachable|implemented) code')
+FUZZER_EXIT_REGEX = re.compile(r'^\s*(?:#0|#1) 0x.*(?:fuzzer|fuzz/|/fuzz)',
+                               re.IGNORECASE)
```
jonathanmetzman, line 118: A more robust but more complicated solution would be to determine the file LLVMFuzzerTestOneInput is in (from the stacktrace) and use that to determine functions in the same file. But I'm fine with this approach.

jonathanmetzman:
Mostly LGTM

src/clusterfuzz/_internal/datastore/data_handler.py
davfu1: Ah, this is here because corpus_pruning_task._process_corpus_crashes() calls store_testcase() with crash=CorpusCrash(), and CorpusCrash() does not have a crash_categories field. Earlier iterations of this PR *did* add crash_categories as a field to CorpusCrash(), but the added field required changes to untrusted_runner.proto and Testcase(). I removed the field when Oliver suggested in earlier comments against changing the Testcase() definition, resulting in this current PR and if-statement.

src/clusterfuzz/_internal/issue_management/issue_filer.py
davfu1: Following up on my other reply, the if-statement here is necessary to avoid a "TypeError: NoneType object is not iterable" when a testcase was created without the crash_categories field via the corpus_pruning_task workflow.

src/clusterfuzz/stacktraces/constants.py
davfu1: Thanks Jonathan! This PR is just the first out of a series targeting CF-filed issues arising from fuzzer implementation/infra errors, so I should have opportunity to further investigate the approach.

src/clusterfuzz/_internal/issue_management/issue_filer.py
jonathanmetzman: `testcase.get_metadata('crash_categories', [])` returns `[]` when crash_categories is empty. I'm pretty sure the `if` is not necessary.

src/clusterfuzz/_internal/datastore/data_handler.py
jonathanmetzman: Can you leave a TODO(#THIS COMMENTTHREAD) to get rid of this then please?

src/clusterfuzz/_internal/issue_management/issue_filer.py
davfu1: Ah I'm sorry, I misread your comment and didn't notice the get_metadata() default value! Thanks for explaining, I've made the fix.

src/clusterfuzz/_internal/datastore/data_handler.py
davfu1: Thanks - I've just added a new comment with a link to this thread.

PR #3334, opened 2023-09-19
Reland: Remove temporary abstractions.

jonathanmetzman:


jonathanmetzman:
I removed it last time because of a misunderstanding about a breakage that I think was actually caused by a k8s migration.

PR #3339, opened 2023-09-19
Make get_revisions_list uworker safe

jonathanmetzman:
Split the unsafe part (get_job_bad_builds) into it's own function that can be called in preprocess.
Call it in preprocess in analyze_task.


oliverchang (commented):
LGTM

PR #3351, opened 2023-09-20
Add auto-cc-visibility issue message temporarily

pgrace-google:
crbug/1478930

### DIFF
--- a/src/appengine/private/templates/error-403.html
+++ b/src/appengine/private/templates/error-403.html
@@ -36,7 +36,8 @@
   {% if message %}
     <p>{{message}}</p>
   {% endif %}
-  <p>Please contact {{contact_string}}.</p>
+  <!--TODO(pgrace) remove this comment in Jan 2024 when we have moved to Buganizer -->
+  <p>If you reached here via a crbug link, please confirm that you have been manually cc'ed. Otherwise, Please contact {{contact_string}}.</p>
   <p>
     <a href="{{switch_account_url}}">Switch account</a><br />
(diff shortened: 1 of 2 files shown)


src/appengine/private/templates/error-403.html
```diff
@@ -36,7 +36,8 @@
   {% if message %}
     <p>{{message}}</p>
   {% endif %}
-  <p>Please contact {{contact_string}}.</p>
+  <!--TODO(pgrace) remove this comment in Jan 2024 when we have moved to Buganizer -->
```
jonathanmetzman, line 39: Hmm...This is a little problematic because it assumes the user is from Chrome. Maybe specify if this website is clusterfuzz.com - please confirm you have bene manually CCed.
@oliverchang WDYT?

pgrace-google: ahh
would using is_chromium (https://github.com/google/clusterfuzz/blob/65b5ca95435835aeeb287b723452650a81d742ad/src/clusterfuzz/_internal/base/utils.py#L915) be good enough to make it only show up for chrome?

oliverchang: +1 yep. We can pipe it through the template similar to here: https://github.com/google/clusterfuzz/blob/65b5ca95435835aeeb287b723452650a81d742ad/src/appengine/handlers/upload_testcase.py#L247

pgrace-google: done! thank you both!

jonathanmetzman (approved):
lgtm

PR #3352, opened 2023-09-20
Allow remote postprocessing of analyze task based on config.

jonathanmetzman:
Add support for remote postprocessing.
Allow remote postprocessing for analyze task on an opt-in basis.

### DIFF
--- a/src/clusterfuzz/_internal/base/tasks.py
+++ b/src/clusterfuzz/_internal/base/tasks.py
@@ -16,7 +16,6 @@
 import contextlib
 import datetime
 import random
-import re
 import threading
 import time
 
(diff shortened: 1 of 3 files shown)


src/clusterfuzz/_internal/tests/core/base/tasks_test.py
```diff
@@ -46,33 +46,23 @@ def test_initialize_trusted_task(self):
...
         'command': self.command,
         'argument': self.argument,
         'job': self.job,
     }
     task = tasks.initialize_task([self.message])
     self.assertFalse(isinstance(task, tasks.PubSubTask))
-    self.assertEqual(task.command, 'uworker_postprocess')
-    self.assertEqual(task.argument, '/mybucket/worker.output')
+    self.assertEqual(task.command, 'postprocess')
+    # !!! Is this bad for local development.
```
jonathanmetzman, line 62: @oliverchang Is it a problem if the task is using gs://bucket/data instead of /bucket/data ?
I don't think this will be executed in local development since pub/sub notifications for GCS can't really be done locally.

jonathanmetzman:
CC @pgrace-google @alhijazi

PR #3353, opened 2023-09-20
[minimize] Fix bug caused by serializing a testcase gotten during main.

jonathanmetzman:
Output doesn't support serializing objects, only wrapped ones.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
@@ -403,7 +403,6 @@ def utask_main(uworker_input):
     logs.log_error('Unable to setup build for minimization.')
     build_fail_wait = environment.get_value('FAIL_WAIT')
     return uworker_io.UworkerOutput(
-        testcase=testcase,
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
CC @pgrace-google 
One error to fix.

jonathanmetzman:
TBR

pgrace-google:
oh dang
handle_minimize_setup_error that uses this output didnt need testcase anyway so, nice!

but whats the difference between that testcase and the testcase being passed in at the end of main in minimize, or places in analyze_task? I dont see it being explicitly differently treated

jonathanmetzman:
>but whats the difference between that testcase and the testcase being passed in at the end of main in minimize, or places in analyze_task? I dont see it being explicitly differently treated

The difference is based on a too-clever solution I made where entities we deserialize from input are wrapped with another object that passes through any variable setting/accesses to the underlying db entity. This wrapper tracks the changes and so it makes it somewhat easier to update only the changed fields. But I think this may not be so useful in the end as I'd rather we send the fields we want in Output then rely on magic like this.

oliverchang:
+1 to removing magic if we can.

PR #3354, opened 2023-09-20
[minimize] Get attribute from correct output object.

jonathanmetzman:


jonathanmetzman:
TBR to fix minimize task.

jonathanmetzman:
Grace could you maybe make a really simple test for postprocess that mocks finalize_testcase, this should catch any silly exceptions.

pgrace-google:
yes will do today!!

PR #3356, opened 2023-09-20
Fix another attr error

jonathanmetzman:


jonathanmetzman:
I should've caught this after finding the last one.

jonathanmetzman:
TBR

PR #3357, opened 2023-09-21
[minimize] Fix another rare deserialization bug

jonathanmetzman:
job_type is not a field on UworkerOutput, it is retrieved from uworker_input.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
@@ -403,7 +403,6 @@ def utask_main(uworker_input):
     logs.log_error('Unable to setup build for minimization.')
     build_fail_wait = environment.get_value('FAIL_WAIT')
     return uworker_io.UworkerOutput(
-        job_type=uworker_input.job_type,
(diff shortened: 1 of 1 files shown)


oliverchang (approved):
Can you give more context for how this happened?

Can we address this in a more systematic way?

pgrace-google:
rip I guess we werent using this job_type anyway but
the test is direly needed (and I am on it!)

pgrace-google (approved):
LGTM!

jonathanmetzman:
> Can you give more context for how this happened?
> 
> Can we address this in a more systematic way?

This happened for two reasons:
1. Lack of testing.
2. The linter is unable to figure out what are the actual fields allowed on Output

The best ways to fix this in a systemic way are 1. More tests (not perfect because many of these bugs occur in edge cases, so the testing is not only hard to do but needs to be more thorough than usual). 2. I'm going to pay attention much more to these objects since they seem to frequently be misused.

Another possibility is to use completely flat Output objects. This might lead to some repetition if we want don't want to use one sole Output object.

WDYT?

jonathanmetzman:
I could also change this mistake to a log error instead of an exception.

PR #3358, opened 2023-09-21
Remove chromecrash uploading

jonathanmetzman:
This feature isn't used anymore.
I don't delete the DB fields yet, in case we want to restore this future because someone complains, when that doesn't happen, I will remove them.


oliverchang (approved):
nice!!!

PR #3322, opened 2023-09-08
Delete all migrated cron jobs and update UT.

hogo6002:
1. Deleted all migrated cron jobs from GAE.
2. Removed unused cron jobs from GAE server.py
3. Updated all cron job unit tests.
4. Moved predator_pull code logic from GAE to _internal.cron (Cannot migrate Predator_pull due to some permission issues).
5. Replaced GAE issue_management with clusterfuzz._internal.issue_management.
6. Remove mail.py and request_cache.py from GAE libs. Added request_cache.py to clusterfuzz._internal_issue_management.
7. updated get_issue_tracker to use request_cache for app engine and memoize wrap for GKE. Added UT for get_issue_tracker.
8. Added is_appengine check for get_issue_tracker, get_issue_tracker_for_testcase, get_issue_for_testcase, and get_issue_url.

PR #3362, opened 2023-09-26
Add the ability to create queues in project_setup (replaces #3306)

marktefftech:
A recently merged PR (https://github.com/google/clusterfuzz/pull/3322/) refactored many of the original files on this initial feature (PR#3306)

### DIFF
--- a/src/clusterfuzz/_internal/base/tasks.py
+++ b/src/clusterfuzz/_internal/base/tasks.py
@@ -100,10 +100,10 @@ class InvalidRedoTask(Error):
 
 def queue_suffix_for_platform(platform):
   """Get the queue suffix for a platform."""
-  if not SUBQUEUE_IDENTIFIER in platform:
-    return '-' + platform.lower().replace('_', '-')
+  if SUBQUEUE_IDENTIFIER in platform:
+    return '-' + platform.lower().replace(':', '-')
 
-  return '-' + platform.lower().replace(':', '-')
+  return '-' + platform.lower().replace('_', '-')
 
 
 def default_queue_suffix():
--- a/src/clusterfuzz/_internal/base/tasks.py
+++ b/src/clusterfuzz/_internal/base/tasks.py
@@ -100,9 +100,8 @@ class InvalidRedoTask(Error):
 
 def queue_suffix_for_platform(platform):
   """Get the queue suffix for a platform."""
-  if SUBQUEUE_IDENTIFIER in platform:
-    return '-' + platform.lower().replace(':', '-')
-
+  # Handle the case where a subqueue is used.
+  platform = platform.lower().replace(':', '-')
   return '-' + platform.lower().replace('_', '-')
 
 
--- a/src/clusterfuzz/_internal/base/tasks.py
+++ b/src/clusterfuzz/_internal/base/tasks.py
@@ -101,7 +101,7 @@ class InvalidRedoTask(Error):
 def queue_suffix_for_platform(platform):
   """Get the queue suffix for a platform."""
   # Handle the case where a subqueue is used.
-  platform = platform.lower().replace(':', '-')
+  platform = platform.lower().replace(SUBQUEUE_IDENTIFIER, '-')
   return '-' + platform.lower().replace('_', '-')
 
 
--- a/src/clusterfuzz/_internal/base/tasks.py
+++ b/src/clusterfuzz/_internal/base/tasks.py
@@ -84,6 +84,9 @@ FILTERS_OR = 'OR'
 
 POSTPROCESS_QUEUE = 'postprocess'
 
+# See https://github.com/google/clusterfuzz/issues/3347 for usage
+SUBQUEUE_IDENTIFIER = ':'
+
 
 class Error(Exception):
   """Base exception class."""
@@ -97,7 +100,10 @@ class InvalidRedoTask(Error):
 
 def queue_suffix_for_platform(platform):
   """Get the queue suffix for a platform."""
-  return '-' + platform.lower().replace('_', '-')
+  if not SUBQUEUE_IDENTIFIER in platform:
+    return '-' + platform.lower().replace('_', '-')
+
+  return '-' + platform.lower().replace(':', '-')
 
 
 def default_queue_suffix():
--- a/src/clusterfuzz/_internal/bot/init_scripts/init_runner.py
+++ b/src/clusterfuzz/_internal/bot/init_scripts/init_runner.py
@@ -34,8 +34,7 @@ def _extension(platform):
 def run():
   """Run custom platform specific init scripts."""
   platform = environment.platform().lower()
-  if tasks.SUBQUEUE_IDENTIFIER in platform:
-    platform = environment.base_platform(platform)
+  platform = environment.base_platform(platform)
   script_path = os.path.join(environment.get_config_directory(), SCRIPT_DIR,
                              platform + _extension(platform))
   if not os.path.exists(script_path):
--- a/src/clusterfuzz/_internal/bot/init_scripts/init_runner.py
+++ b/src/clusterfuzz/_internal/bot/init_scripts/init_runner.py
@@ -15,7 +15,6 @@
 
 import os
 
-from clusterfuzz._internal.base import tasks
 from clusterfuzz._internal.metrics import logs
 from clusterfuzz._internal.system import environment
 from clusterfuzz._internal.system import process_handler
--- a/src/clusterfuzz/_internal/bot/init_scripts/init_runner.py
+++ b/src/clusterfuzz/_internal/bot/init_scripts/init_runner.py
@@ -15,6 +15,7 @@
 
 import os
 
+from clusterfuzz._internal.base import tasks
 from clusterfuzz._internal.metrics import logs
 from clusterfuzz._internal.system import environment
 from clusterfuzz._internal.system import process_handler
@@ -33,6 +34,8 @@ def _extension(platform):
 def run():
   """Run custom platform specific init scripts."""
   platform = environment.platform().lower()
+  if tasks.SUBQUEUE_IDENTIFIER in platform:
+    platform = 'android'
   script_path = os.path.join(environment.get_config_directory(), SCRIPT_DIR,
                              platform + _extension(platform))
   if not os.path.exists(script_path):
--- a/src/clusterfuzz/_internal/bot/init_scripts/init_runner.py
+++ b/src/clusterfuzz/_internal/bot/init_scripts/init_runner.py
@@ -35,7 +35,7 @@ def run():
   """Run custom platform specific init scripts."""
   platform = environment.platform().lower()
   if tasks.SUBQUEUE_IDENTIFIER in platform:
-    platform = 'android'
+    platform = environment.base_platform(platform)
   script_path = os.path.join(environment.get_config_directory(), SCRIPT_DIR,
                              platform + _extension(platform))
   if not os.path.exists(script_path):
--- a/src/clusterfuzz/_internal/bot/tasks/commands.py
+++ b/src/clusterfuzz/_internal/bot/tasks/commands.py
@@ -304,10 +304,9 @@ def process_command(task):
         current_platform_id = environment.get_platform_id()
         testcase_platform_id = testcase.platform_id
 
-        # This indicates we are trying to run this job on the wrong platform.
-        # This can happen when you have different type of devices (e.g
-        # android) on the same platform group. In this case, we just recreate
-        # the task.
+        # This indicates we are trying to run this job on the wrong platform
+        # and potentially blocks fuzzing. See the 'subqueues' feature for
+        # more details: https://github.com/google/clusterfuzz/issues/3347
         if (task_name != 'variant' and testcase_platform_id and
             not utils.fields_match(testcase_platform_id, current_platform_id)):
           logs.log(
--- a/src/clusterfuzz/_internal/cron/project_setup.py
+++ b/src/clusterfuzz/_internal/cron/project_setup.py
@@ -39,6 +39,8 @@ from clusterfuzz._internal.system import environment
 
 from . import service_accounts
 
+ANDROID_CUTTLEFISH = 'ANDROID_X86'
+
 BUILD_BUCKET_PATH_TEMPLATE = (
     'gs://%BUCKET%/%PROJECT%/%PROJECT%-%SANITIZER%-([0-9]+).zip')
 
@@ -106,6 +108,8 @@ class JobInfo:
 # the earlier ones. An engine template may override vars set for a sanitizer.
 LIBFUZZER_ASAN_JOB = JobInfo('libfuzzer_asan_', 'libfuzzer', 'address',
                              ['libfuzzer', 'engine_asan', 'prune'])
+LIBFUZZER_HWASAN_JOB = JobInfo('libfuzzer_hwasan_', 'libfuzzer', 'address',
+                               ['libfuzzer', 'engine_asan', 'android'])
 LIBFUZZER_MSAN_JOB = JobInfo('libfuzzer_msan_', 'libfuzzer', 'memory',
                              ['libfuzzer', 'engine_msan'])
 LIBFUZZER_UBSAN_JOB = JobInfo('libfuzzer_ubsan_', 'libfuzzer', 'undefined',
@@ -159,11 +163,14 @@ JOB_MAP = {
             'address': LIBFUZZER_ASAN_I386_JOB,
             'none': LIBFUZZER_NONE_I386_JOB,
         },
+        'arm': {
+            'hardware': LIBFUZZER_HWASAN_JOB,
+        },
     },
     'afl': {
         'x86_64': {
             'address': AFL_ASAN_JOB,
-        }
+        },
     },
     'honggfuzz': {
         'x86_64': {
@@ -180,7 +187,7 @@ JOB_MAP = {
     'none': {
         'x86_64': {
             'address': NO_ENGINE_ASAN_JOB,
-        }
+        },
     },
     'centipede': {
         'x86_64': {
@@ -545,7 +552,15 @@ def create_project_settings(project, info, service_account):
 
 def create_pubsub_topics(project):
   """Create pubsub topics for tasks."""
-  for platform in PUBSUB_PLATFORMS:
+  platforms = PUBSUB_PLATFORMS
+  if 'android_' not in project:
+    platforms = PUBSUB_PLATFORMS
+  else:
+    platforms = [project.replace('android_', '')]
+    # Avoid the stutter in queue name
+    project = 'android'
+
+  for platform in platforms:
     name = untrusted.queue_name(project, platform)
     client = pubsub.PubSubClient()
     application_id = utils.get_application_id()
@@ -767,8 +782,12 @@ class ProjectSetup:
       job.name = job_name
       if self._segregate_projects:
         job.platform = untrusted.platform_name(project, 'linux')
+      elif 'android' in job.name:
+        if 'hwasan' in job.name:
+          job.platform = job_name.split('_')[-1].upper()
+        else:
+          job.platform = ANDROID_CUTTLEFISH
       else:
-        # TODO(ochang): Support other platforms?
         job.platform = 'LINUX'
 
       job.templates = template.cf_job_templates
@@ -863,6 +882,14 @@ class ProjectSetup:
 
       file_github_issue = info.get('file_github_issue', False)
       job.environment_string += f'FILE_GITHUB_ISSUE = {file_github_issue}\n'
+
+      # Android creates device-specific queues during project setup.
+      queue_id = info.get('queue_id', False)
+      if queue_id:
+        if 'android' not in job.templates:
+          job.templates.append('android')
+        job.queue = queue_id
+        create_pubsub_topics(project)
 
       if (template.engine == 'libfuzzer' and
           template.architecture == 'x86_64' and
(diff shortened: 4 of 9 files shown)


src/clusterfuzz/_internal/cron/project_setup.py
```diff
@@ -545,7 +552,15 @@ def create_project_settings(project, info, service_account):
 
 def create_pubsub_topics(project):
   """Create pubsub topics for tasks."""
-  for platform in PUBSUB_PLATFORMS:
+  platforms = PUBSUB_PLATFORMS
+  if 'android_' not in project:
```
oliverchang, line 556: Rather than add this condition, can we refactor things such that there are two functions -- one for `create_pubsub_topics_for_untrusted`, and one for `create_pubsub_topics_for_queue_id` ? That way the logic is a bit cleaner between the two cases, and if there are still significant bits of shared code we can put them into a common helper.

oliverchang (commented):
nice! some initial comments.

src/clusterfuzz/_internal/base/tasks.py
```diff
@@ -97,7 +100,10 @@ def __init__(self, task):
 
 def queue_suffix_for_platform(platform):
   """Get the queue suffix for a platform."""
-  return '-' + platform.lower().replace('_', '-')
+  if not SUBQUEUE_IDENTIFIER in platform:
```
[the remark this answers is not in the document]
marktefftech: Done.

src/clusterfuzz/_internal/bot/init_scripts/init_runner.py
```diff
@@ -33,6 +34,8 @@ def _extension(platform):
 def run():
   """Run custom platform specific init scripts."""
   platform = environment.platform().lower()
+  if tasks.SUBQUEUE_IDENTIFIER in platform:
+    platform = 'android'
```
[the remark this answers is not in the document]
marktefftech: I introduced the new function `environment.base_platform()` to extract the base_platform.

However, the function is very simple (1-line). I do prefer the way that the code reads, although let me know if you prefer the just to extract the base_platform directly instead of introducing the new function.

src/clusterfuzz/_internal/cron/project_setup.py
marktefftech: Refactored into two new functions:
1. `create_pubsub_topics_for_untrusted`
2. `create_pubsub_topics_for_queue_id`

Combine the common logic into the new `_create_pubsub_topic()` helper function.

src/clusterfuzz/_internal/cron/project_setup.py
```diff
@@ -545,7 +552,15 @@ def create_project_settings(project, info, service_account):
...
+    platforms = PUBSUB_PLATFORMS
+  else:
+    platforms = [project.replace('android_', '')]
```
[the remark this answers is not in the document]
marktefftech: Yes, per the suggestion above, this was generalized into the two new functions.

The new function creates a queue name directly based on the `queue_id`, however, the downsteam helper function still extracts `project` & `platform` from the `queue_id`. I prefer this implementation since the queue naming logic is heavily dependent on this existing  project / platform combination:

```
def create_pubsub_topics_for_queue_id(queue_id):
  """Create pubsub topics from project configs for tasks."""
  project, platform = queue_id.split('-') # Although we ingest queue_id, still rely on legacy naming convention
  name = untrusted.queue_name(project, platform)
  _create_pubsub_topic(name)
```

Let me know if this looks alright to you

src/clusterfuzz/_internal/cron/project_setup.py
```diff
@@ -767,8 +782,12 @@ def _sync_job(self, project, info, corpus_bucket_name, quarantine_bucket_name,
       job.name = job_name
       if self._segregate_projects:
         job.platform = untrusted.platform_name(project, 'linux')
+      elif 'android' in job.name:
+        if 'hwasan' in job.name:
+          job.platform = job_name.split('_')[-1].upper()
```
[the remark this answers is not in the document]
marktefftech: This logic was changed to ingest the `template.architecture` as opposed to the special conditions above.

I also moved this logic to the `queue_id` section to modularize the platform setting as a part of the new queue recognition logic:

```
# Android creates device-specific queues during project setup.
      queue_id = info.get('queue_id', False)
      if queue_id:
        if 'android' not in job.templates:
          job.templates.append('android')
        if template.architecture == 'arm': # Set platform based on architecture
          job.platform = queue_id.split('-')[1].upper()
        else:
          job.platform = ANDROID_CUTTLEFISH
        create_pubsub_topics_for_queue_id(queue_id)
```

marktefftech:
@oliverchang friendly bump

src/clusterfuzz/_internal/base/tasks.py
```diff
@@ -97,6 +100,9 @@ def __init__(self, task):
 
 def queue_suffix_for_platform(platform):
   """Get the queue suffix for a platform."""
+  if SUBQUEUE_IDENTIFIER in platform:
+    return '-' + platform.lower().replace(':', '-')
```
oliverchang, line 104: It should be harmless to always do this replacement right? 

We can just chain the replace without the if and add a comment explaining the cases.

src/clusterfuzz/_internal/bot/init_scripts/init_runner.py
```diff
@@ -33,6 +34,8 @@ def _extension(platform):
 def run():
   """Run custom platform specific init scripts."""
   platform = environment.platform().lower()
+  if tasks.SUBQUEUE_IDENTIFIER in platform:
```
oliverchang, line 37: Can we just remove this if and always call base_platform? I'm assuming it's a noop when there's no subqueue.

src/clusterfuzz/_internal/system/environment.py
```diff
@@ -1107,7 +1112,7 @@ def is_android_kernel(plt=None):
 
 def is_android_real_device():
   """Return True if we are on a real android device."""
-  return platform() == 'ANDROID'
+  return platform() == 'ANDROID' or ':' in platform()
```
oliverchang, line 1115: Can we make more specific to Android? i.e. check if the platform().startswith('ANDROID') ?

src/clusterfuzz/_internal/tests/appengine/handlers/cron/project_setup_test.py
```diff
@@ -1721,6 +1721,28 @@ def _mock_read_data(path):
...
+  if 'android' in path:
+    return json.dumps({
+        'projects': [{
+            'build_path': 'gs://bucket-android/%ENGINE%/%SANITIZER%/'
+                          '%TARGET%/([0-9]+).zip',
+            'name': 'android_pixel7',
+            'fuzzing_engines': ['libfuzzer'],
+            'architectures': ['arm'],
+            'sanitizers': ['hardware'],
+            'queue_id': 'android-pixel7'
```
oliverchang, line 1733: Can we instead do:

```
'platform': 'ANDROID'
'queue_id': 'pixel7'
```

?

src/clusterfuzz/_internal/fuzzing/fuzzer_selection.py
```diff
@@ -116,6 +117,11 @@ def get_fuzz_task_payload(platform=None):
     query = data_types.FuzzerJob.query()
     query = query.filter(data_types.FuzzerJobs.platform == platform)
     mappings = list(ndb_utils.get_all_from_query(query))
+  elif tasks.SUBQUEUE_IDENTIFIER in platform:
```
oliverchang, line 120: Can  you explain what this is doing a bit more? It's a bit unclear to me.

src/clusterfuzz/_internal/cron/project_setup.py
```diff
@@ -864,6 +885,17 @@ def _sync_job(self, project, info, corpus_bucket_name, quarantine_bucket_name,
       file_github_issue = info.get('file_github_issue', False)
       job.environment_string += f'FILE_GITHUB_ISSUE = {file_github_issue}\n'
 
+      # Android creates device-specific queues during project setup.
+      queue_id = info.get('queue_id', False)
```
oliverchang, line 889: see my comment below and making this:

```
'platform': 'ANDROID'
'queue_id': 'pixel7'
```

instead. 

Then we can key on 'platform'  == 'ANDROID' for many of thjese android specific pieces, and avoid making assumptions on formatting (i.e. the splitting on -). We can also set 'ANDROID_X86' in the projects.json without adding a special case here.

src/clusterfuzz/_internal/cron/project_setup.py
```diff
@@ -864,6 +885,17 @@ def _sync_job(self, project, info, corpus_bucket_name, quarantine_bucket_name,
...
+      if queue_id:
+        if 'android' not in job.templates:
+          job.templates.append('android')
+        if template.architecture == 'arm':
+          job.platform = queue_id.split('-')[1].upper()
+        else:
+          job.platform = ANDROID_CUTTLEFISH
```
oliverchang, line 896: Can  you please add a comment for why these cases are different?

src/clusterfuzz/_internal/system/environment.py
marktefftech: We can't use `platform().startswith('ANDROID')` because this would create false positive for other devices (ex. emulator `ANDROID_EMULATOR`) would return true.

Now that `base_platform()` is generalized it will work here

src/clusterfuzz/_internal/fuzzing/fuzzer_selection.py
marktefftech: Part of this project is to make it so where a bot can pull fuzzing tasks from two different groups: 
1. Device-specific fuzzing jobs and
2. It's general platform fuzzing jobs

So any bot with`QUEUE_OVERRIDE=<PLATFORM>:<DEVICE` defined, take `ANDROID:PIXEL8` for example, will be able to choose jobs where the platform is either `PIXEL8` (for fuzzers that can only run on pixel8) or `ANDROID` (for generic fuzzers.

However, if I implement your suggestion above, we will not be able to query fuzzing jobs for both platforms, since all jobs will have `ANDROID` or `ANDROID_X86` as the platform:

```
'platform': 'ANDROID'
'queue_id': 'pixel7
```

This was discussed in the "Pulling in Fuzz Tasks" section of the design document. 

Would you prefer I refactor this code to query fuzzing jobs for via something other than platform, or is the current design adequate?

src/clusterfuzz/_internal/base/tasks.py
marktefftech: Done.

src/clusterfuzz/_internal/fuzzing/fuzzer_selection.py
oliverchang: Ah, apologies, it's been a while since I read the relevant doc :) 

`platform` is the filter to use, but I think we can make things a bit if we standardize some conventions here. i.e. we have a single "ANDROID:DEVICE_ID" as both the platform name AND the QUEUE_OVERRIDE.

Re my other suggestion to adopt:

```
'platform': 'ANDROID'
'queue_id': 'pixel7
```

This is also mostly from the perspective of making projects.json clearer and the project_setup more generalized. What project setup could do is construct the relevant platform strings (e.g. "ANDROID:pixel7" ), and for the implementation we base things on `platform` rather than `.startswith('ANDROID')`. 

Then here, if we standardize on `ANDROID:DEVICE_ID` as both the platform and queue override, we could just do something like:

```python
platforms = [platform]
base_platform = platform.split(':')[0]

# Generalized queue for platforms with a base platform (e.g. ANDROID)
if base_platform != platform:
  platforms.append(base_platform)

query = query.filter(.....IN(platforms))
```

marktefftech: This was a really helpful explanation, thank you. I implemented the changes you suggested and included `platform` as a value passed via the config file.

Please take a look and lmk if it looks better to you or can still be enhanced.

p.s. no worries about the design doc, I know it has been a while since we talked about this initially. thx for the comments

src/clusterfuzz/_internal/cron/project_setup.py
marktefftech: Done.

src/clusterfuzz/_internal/cron/project_setup.py
marktefftech: This is no longer necessary as the `job.platform` is being constructed from the variables passed in

src/clusterfuzz/_internal/tests/appengine/handlers/cron/project_setup_test.py
marktefftech: Done.

src/clusterfuzz/_internal/bot/init_scripts/init_runner.py
marktefftech: Done.

src/clusterfuzz/_internal/cron/project_setup.py
```diff
@@ -543,20 +551,32 @@ def create_project_settings(project, info, service_account):
         ccs=ccs).put()
 
 
-def create_pubsub_topics(project):
-  """Create pubsub topics for tasks."""
+def _create_pubsub_topic(name):
+  """Create a pubsub topic and subscription if needed."""
+  client = pubsub.PubSubClient()
```
oliverchang, line 556: Can we pull this client out and have the caller create this once instead?

src/clusterfuzz/_internal/fuzzing/fuzzer_selection.py
```diff
@@ -112,13 +112,18 @@ def get_fuzz_task_payload(platform=None):
     queue_override = environment.get_value('QUEUE_OVERRIDE')
     platform = queue_override if queue_override else environment.platform()
 
+  platforms = [platform]
+  base_platform = platform.split(':')[0]
+  # Generalized queue for platforms with a base platform (e.g. ANDROID)
+  if base_platform != platform:
+    platforms.append(base_platform)
```
oliverchang, line 119: nit: newline after this.

src/clusterfuzz/_internal/fuzzing/fuzzer_selection.py
```diff
@@ -112,13 +112,18 @@ def get_fuzz_task_payload(platform=None):
...
   if environment.is_local_development():
     query = data_types.FuzzerJob.query()
     query = query.filter(data_types.FuzzerJobs.platform == platform)
```
oliverchang, line 122: nit: change this to IN too. 

OR actually, can we restructure the code so that there the query setup is only done once between the two branches?

src/clusterfuzz/_internal/bot/tasks/commands.py
```diff
@@ -304,10 +304,9 @@ def process_command(task):
         current_platform_id = environment.get_platform_id()
         testcase_platform_id = testcase.platform_id
 
-        # This indicates we are trying to run this job on the wrong platform.
-        # This can happen when you have different type of devices (e.g
-        # android) on the same platform group. In this case, we just recreate
-        # the task.
+        # This indicates we are trying to run this job on the wrong platform
+        # and potentially blocks fuzzing. See the 'subqueues' feature for
+        # more details: https://github.com/google/clusterfuzz/issues/3347
```
oliverchang, line 309: Can we completely delete this code after you start using this?

oliverchang (commented):
nice! this is looking pretty good generally. just some final minor comments.

src/clusterfuzz/_internal/cron/project_setup.py
```diff
@@ -864,6 +883,17 @@ def _sync_job(self, project, info, corpus_bucket_name, quarantine_bucket_name,
       file_github_issue = info.get('file_github_issue', False)
       job.environment_string += f'FILE_GITHUB_ISSUE = {file_github_issue}\n'
 
+      # Device-specific queues created during project setup.
+      queue_id = info.get('queue_id', False)
+      if queue_id:
+        platform = info.get('platform',
+                            'ANDROID').upper() + ':' + queue_id.upper()
```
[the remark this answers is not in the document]
marktefftech: No, this is not necessary. This is simply adding `ANDROID` as the default platform if one was not specified.

Alternatively, I could change this to a different platform `None` / `LINUX` or raise an error if platform was not specified. Since Android is the only project currently using it I thought it made sense to implement this way, but definitely willing to modify if you think one of the other 2 routes is more appropriate.

src/clusterfuzz/_internal/base/tasks.py
```diff
@@ -97,6 +100,8 @@ def __init__(self, task):
 
 def queue_suffix_for_platform(platform):
   """Get the queue suffix for a platform."""
+  # Handle the case where a subqueue is used.
+  platform = platform.lower().replace(':', '-')
```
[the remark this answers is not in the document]
marktefftech: Done.

src/clusterfuzz/_internal/tests/appengine/handlers/cron/project_setup_test.py
```diff
@@ -2014,18 +2076,109 @@ def test_execute(self):
     self.assertEqual(None, job.external_updates_subscription)
     self.assertFalse(job.is_external())
 
+    job = data_types.Job.query(
+        data_types.Job.name == 'libfuzzer_hwasan_android_pixel7').get()
+    print(job)
```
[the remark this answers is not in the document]
marktefftech: Done.

src/clusterfuzz/_internal/cron/project_setup.py
```diff
@@ -543,20 +551,32 @@ def create_project_settings(project, info, service_account):
...
-    topic_name = pubsub.topic_name(application_id, name)
-    if client.get_topic(topic_name) is None:
-      client.create_topic(topic_name)
 
-    subscription_name = pubsub.subscription_name(application_id, name)
-    if client.get_subscription(subscription_name) is None:
-      client.create_subscription(subscription_name, topic_name)
+def create_pubsub_topics_for_queue_id(queue_id, platform):
+  """Create pubsub topics from project configs for tasks."""
+  platform = environment.base_platform(platform)
```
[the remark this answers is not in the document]
marktefftech: No, given the new implementation this is no longer necessary.

However, because of how `untrusted.queue_name()` constructs the queue name, we do need to still pass it both the platform and the queue / device model. 

Updated the logic to account for this.

src/clusterfuzz/_internal/fuzzing/fuzzer_selection.py
marktefftech: Done.

src/clusterfuzz/_internal/cron/project_setup.py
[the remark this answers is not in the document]
oliverchang: Can you please change the default to 'LINUX'? That's most likely the common platform on ClusterFuzz, and for Android I'd expect you'd always be setting `platform`.

src/clusterfuzz/_internal/bot/tasks/commands.py
oliverchang: Ack on comment being useful to other teams, but I think that could go somewhere else potentially. 

I was more referring to the existing platform_id mechanism, which we can likely just get rid of after this right? 

Either way, this is just something for the future, so nothing actionable right now.

oliverchang (approved):
LGTM!

src/clusterfuzz/_internal/bot/tasks/commands.py
marktefftech: Hypothetically yes, there will be no need to compare `current_platform_id` to  `testcase_platform_id` after this change as they will always be the same.

src/clusterfuzz/_internal/cron/project_setup.py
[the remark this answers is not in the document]
marktefftech: Done.

src/clusterfuzz/_internal/fuzzing/fuzzer_selection.py
```diff
@@ -120,8 +120,8 @@ def get_fuzz_task_payload(platform=None):
 
   query = data_types.FuzzerJobs.query()
   if environment.is_local_development():
-    query = query.filter(data_types.FuzzerJobs.platform.IN(platforms))[0]
-    mappings = list(ndb_utils.get_all_from_query(query))
+    query = query.filter(data_types.FuzzerJobs.platform.IN(platforms))
+    mappings = list(ndb_utils.get_all_from_query(query))[0]
```
jonathanmetzman, line 124: I think this is still wrong. Maybe do list(ndb_utils.get_all_from_query(query))[:1]

marktefftech: Thanks for your proactivity here Jonathan. I expanded on the `fuzzer_selection_test : test_platform_restriction()` to account for this and it passes successfully, which lead me to believe this implementation worked correctly.

From a strictly pythonic standpoint, aren't the below two lines exactly the same (except in the case of no data found, which would lead to a `KeyError` in scenario 1)?

```
mappings = list(ndb_utils.get_all_from_query(query))[0]
vs.
mappings = list(ndb_utils.get_all_from_query(query))[:1]
```

marktefftech: Regardless, I pushed the suggested change, but would highly prefer to verify it works so as not to break anything. This is a crucial snippet of code for bots.

jonathanmetzman: They arent the same. `L[0]` gets the 0th element in `L`, whereas `L[:1]` gets a list containing everything up to but not including the 1st element in L.
```
L = ['A', 'B']
L[0]   # 'A'
L[:1]  # ['A']
```

PR #3365, opened 2023-09-27
Make Task Outputs extend UworkerOutput

alhijazi:


alhijazi:
@jonathanmetzman  could you please take a look?

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
```diff
@@ -389,7 +374,7 @@ class UworkerOutput(UworkerMsg):
 
   def save_rich_type(self, attribute, value):
     field = getattr(self.proto, attribute)
-    if isinstance(value, dict):
```
jonathanmetzman, line 392: I don't think we should make this change, it's not obvious to me that a list should always be a json field in every case. Maybe keep the override method in fuzz and leave this unchanged.

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
```diff
@@ -469,29 +476,6 @@ class RegressionTaskInput(UworkerInput):
   PROTO_CLS = uworker_msg_pb2.RegressionTaskInput
 
 
-class AnalyzeTaskOutput(UworkerMsg):  # pylint: disable=abstract-method
```
jonathanmetzman, line 472: I guess I already did this with input but I think the model for base inputs and specific inputs are confused. In protos theyre subfields, here they are child objects.

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
alhijazi: which lists should be json fields and which should not?

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
alhijazi: There is confusion indeed. I think that to avoid code duplication, and since normally TaskOutput types should behave similarly, maybe we could create a `TaskOutput` class and make Analyze,Minimize,... extend it. I made this change because I saw we were already duplicating the `save_rich_type` method in `MinimizeTaskOutput` and `AnalyzeTaskOutput`.
Same thing should probably be done for TaskInputs instead of extending `UworkerInput`.
WDYT?

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
```diff
@@ -411,6 +396,40 @@ def save_rich_type(self, attribute, value):
     field.CopyFrom(wrapped_entity_proto)
 
 
+class TaskOutput(UworkerMsg):
+  """ Base class representing a generic TaskOutput message from a task."""
```
jonathanmetzman, line 400: nit: Remove space

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
jonathanmetzman: Eh I think inheriting from UworkerOutput and UworkerInput is fine.

alhijazi:
No longer needed as we're looking at a different approach (using protos directly).

PR #3375, opened 2023-10-02
Remove unneeded packages lxml and twisted.

jonathanmetzman:
Useful for python3.11, we don't need to worry about compatibility of unused packages
Land some of the changes from: #3264


jonathanmetzman:
CC @alhijazi

oliverchang (approved):

PR #3380, opened 2023-10-02
Upgrade Google cloud batch:

jonathanmetzman:


jonathanmetzman:
Actually, I don't think this is needed for Python upgrade.

PR #3386, opened 2023-10-02
Stop re-fetching the testcase within utask_main

alhijazi:


alhijazi:
@jonathanmetzman  Could you please take a look?

jonathanmetzman:
Do you know if this is actually safe to do? Will this change result in any change in behavior with regards to concurrency. I understand how this fixes our current problem, but it may have problems of its own.

src/clusterfuzz/_internal/protos/uworker_msg.proto
```diff
@@ -111,6 +111,11 @@ message ProgressionTaskOutput{
   // Information passed back to trusted worker when we crash on latest.
   optional string crash_revision = 5;
   optional string last_tested_crash_stacktrace = 6;
+  optional string last_progression_min = 7;
+  optional string last_progression_max = 8;
+  // When set to True, we delete the metadata related to: last_progression_min
+  // and last_progression_max from the testcase on postprocess
```
jonathanmetzman, line 117: nit: end with period.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -54,6 +80,7 @@ def handle_progression_build_not_found(
   clears progression_pending testcase metadata"""
   testcase_id = uworker_output.uworker_input.testcase_id
   testcase = data_handler.get_testcase_by_id(testcase_id)
+  _maybe_clear_progression_last_min_max_metadata(testcase, uworker_output)
```
jonathanmetzman, line 83: This is OK for now, but I guess we have to do this before error handling (ugh, bad assumptions).

jonathanmetzman (approved):
lgtm

jonathanmetzman:
Actually let me check this over again

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -37,12 +37,38 @@
...
+  task_output = uworker_output.progression_task_output
+  if task_output is None:
+    return
+
+  if task_output.clear_min_max_metadata:
+    testcase.delete_metadata('last_progression_min', update_testcase=False)
+    testcase.delete_metadata('last_progression_max', update_testcase=False)
+
+
+def _save_current_fixed_range_indices(testcase, uworker_output):
```
jonathanmetzman, line 52: Try not to move functions around like this please, it makes the diff harder to read.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -536,16 +554,19 @@ def find_fixed_range(uworker_input):
...
+          clear_min_max_metadata=clear_min_max_metadata)
       return error
 
     if result.is_crash():
       min_index = middle_index
     else:
       max_index = middle_index
 
-    # TODO(alhijazi): This cannot be done from an untrusted bot.
-    _save_current_fixed_range_indices(testcase_id, revision_list[min_index],
```
jonathanmetzman, line 547: As far as I can tell, this should only happen when there is a timeout. But you also added it to handle_progression_bad_state_min_max
Why?

alhijazi: `_save_current_fixed_range_indices` was saving the changing to the testcase and uploading it directly with `testcase.put()`. We are not doing this anymore, so in order no to lose changes and keep the progress we did so far, we need to save the last_progression_min,last_progression_max values we have in case of error in this loop.

PR #3390, opened 2023-10-04
Temporary fix for type issue caused by progression

jonathanmetzman:


alhijazi (approved):

jonathanmetzman:
Not sure I should do this, may try something else

PR #3391, opened 2023-10-04
Revert progression task rearch

jonathanmetzman:


jonathanmetzman:
I don't think we need this anymore.

PR #3392, opened 2023-10-04
Allow preprocess to return None

jonathanmetzman:
Skip main when this happens.
This behavior was originally intended and is needed to preserved functionality in progression_task.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
@@ -353,7 +353,10 @@ def utask_preprocess(testcase_id, job_type, uworker_env):
     return None
 
   if testcase.fixed:
-    logs.log_error(f'Fixed range is already set as {testcase.fixed}, skip.')
+    error_message = f'Fixed range is already set as {testcase.fixed}, skip.'
+    logs.log_error(error_message)
+    data_handler.update_testcase_comment(testcase, data_types.TaskState.ERROR,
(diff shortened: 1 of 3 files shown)


src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -353,7 +353,10 @@ def utask_preprocess(testcase_id, job_type, uworker_env):
     return None
 
   if testcase.fixed:
-    logs.log_error(f'Fixed range is already set as {testcase.fixed}, skip.')
+    error_message = f'Fixed range is already set as {testcase.fixed}, skip.'
+    logs.log_error(error_message)
+    data_handler.update_testcase_comment(testcase, data_types.TaskState.ERROR,
```
alhijazi, line 358: This wasn't the previous behaviour as far as I can see in:
https://github.com/google/clusterfuzz/commit/65dd6f43a44b6eff009c1c02bd432e9c5f3b6d90#diff-e67a9c5a80de5ba8fcd9fabf0d42884679d4920f321be15b70088c85d0ee75bbL232

jonathanmetzman: Thanks not sure why I thought this was the case.

jonathanmetzman:
I think this is the original behavior, if the testcase does not exist or is fixed, we return gracefully.

alhijazi (approved):

PR #3394, opened 2023-10-04
[python3.11] Upgrade psutil

jonathanmetzman:
Do this so that we don't need to upgrade when upgrading Python3.11 and we can isolate the source of any bugs.

Redo upgrade reverted by https://github.com/google/clusterfuzz/pull/3264


oliverchang (approved):

jonathanmetzman:
I suppose nothing will use this change until we rebuild the docker images right?

oliverchang:
Yep, we need to rebuild Docker images.

PR #3396, opened 2023-10-05
[python3.11] Upgrade firebase.

jonathanmetzman:
Do this so that we don't need to upgrade when upgrading Python3.11 and we can isolate the source of any bugs.

Redo upgrade reverted by #3264


jonathanmetzman:
I think this needs a proto/other cloud upgrade

PR #3400, opened 2023-10-05
Retrieve APP_REVISION for custom_binaries after setup_build is called.

alhijazi:
@jonathanmetzman  Could you please take a look?
This is related to: 
https://pantheon.corp.google.com/errors/detail/CLm6ree1lZ_79QE;filter=%5B%22progression%22%5D?e=-13802955&mods=logs_tg_prod&project=google.com:clusterfuzz

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
@@ -214,9 +214,9 @@ def _log_output(revision, crash_result):
 
 def _check_fixed_for_custom_binary(testcase, testcase_file_path):
   """Simplified fixed check for test cases using custom binaries."""
-  revision = environment.get_value('APP_REVISION')
-
   build_manager.setup_build()
+  # 'APP_REVISION' is set during setup_build().
+  revision = environment.get_value('APP_REVISION')
   if not build_manager.check_app_path():
     return uworker_io.UworkerOutput(
         testcase=testcase,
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -214,9 +214,9 @@ def _log_output(revision, crash_result):
 
 def _check_fixed_for_custom_binary(testcase, testcase_file_path):
   """Simplified fixed check for test cases using custom binaries."""
-  revision = environment.get_value('APP_REVISION')
-
   build_manager.setup_build()
+  # 'APP_REVISION' is set during setup_build().
+  revision = environment.get_value('APP_REVISION')
```
jonathanmetzman, line 219: This doesn't match the old behavior before rearch: https://github.com/google/clusterfuzz/blob/f3a760c6fea10d62d1eb6299e0d1f5d5c04b362b/src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py

I think the bug comes from doing `crash_revision=int(revision)` when revision is None (which is legit) so I think the fix is not setting it at all. WDYT?

jonathanmetzman: Hmm, can you open another PR that does what I suggest.
We can land this one afterwards when things stabilize a bit.

alhijazi: Alright

jonathanmetzman: We decided against this in the end.
In the end, we said we should keep this fix but log when it's none. Can you do this?

alhijazi: yes I'm on it.

jonathanmetzman (approved):
lgtm

PR #3403, opened 2023-10-05
Increase libfuzzer rss_limit_mb limit for bigger fuzzers

paulsemel:


src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/fuzzer.py
```diff
@@ -70,10 +70,10 @@ def get_arguments(fuzzer_path):
         '%s%d' % (constants.RSS_LIMIT_FLAG, constants.DEFAULT_RSS_LIMIT_MB))
   else:
     # Custom rss_limit_mb value shouldn't be greater than the default value.
-    if rss_limit_mb > constants.DEFAULT_RSS_LIMIT_MB:
+    if rss_limit_mb > constants.MAXIMUM_RSS_LIMIT_MB:
```
jonathanmetzman, line 73: Hmm...so I think this creates an issue for us, in OSS-Fuzz we only have 3 GB of RAM on each machine while on Chrome we have 8. So can you make this only happen if is_chromium()?

oliverchang: Or better, can we use https://psutil.readthedocs.io/en/latest/#psutil.virtual_memory and base the maximum on that instead? We just need to specify a hardcoded overhead here. 

i.e. 

```
MEMORY_OVERHEAD = 1024  # 1 GB

max_memory_limit = psutil.virtual_memory().total - MEMORY_OVERHEAD 
```

paulsemel: Good point, I think actually using the limit makes sense here!

oliverchang (approved):
LGTM. @jonathanmetzman WDYT?

jonathanmetzman (approved):
LGTM

PR #3407, opened 2023-10-05
[rearch][fuzz_task] Save targets count in postprocess

jonathanmetzman:


jonathanmetzman:
TBR

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -1928,11 +1927,14 @@ def utask_preprocess(fuzzer_name, job_type, uworker_env):
   do_multiarmed_bandit_strategy_selection(uworker_env)
   environment.set_value('PROJECT_NAME', data_handler.get_project_name(job_type),
                         uworker_env)
+  targets_count = ndb.Key(data_types.FuzzTargetsCount, job_type).get()
+  fuzz_task_input = uworker_io.FuzzTaskInput(targets_count=targets_count)
```
alhijazi, line 1931: I don't see `fuzz_task_input` being used within `utask_main`.
I'd suggest to remove the `FuzzTaskInput.targets_count`. All we need is to return `the new_targets_count` to postprocess and then compare and update if the values are different.

PR #3409, opened 2023-10-05
Fix base image build

jonathanmetzman:


jonathanmetzman:
This was why the bad deploy (leading to psutil not being found everywhere) happened.

jonathanmetzman:
> This was why the bad deploy (leading to psutil not being found everywhere) happened.

The code was initially removed by me then added back but removed again during the 3.11->3.7 revert.

jonathanmetzman:
TBR

oliverchang (commented):
LGTM

PR #3410, opened 2023-10-05
Fix installation of xenial libs

jonathanmetzman:
Without this, building new docker images fails.


oliverchang (approved):

jonathanmetzman:
I think the server was fixed.

PR #3414, opened 2023-10-06
Clarify that get_testcase_by_id() cannot return None

letitz:
Instead, it raises an `InvalidTestcaseError` exception.

Along the way, clean up `get_testcase_by_id()` a bit by relying on `int()` exception handling to  catch bad IDs, and augment `InvalidTestcaseError` to capture the invalid ID for (hopefully) better logging.

### DIFF
--- a/src/clusterfuzz/_internal/base/errors.py
+++ b/src/clusterfuzz/_internal/base/errors.py
@@ -50,8 +50,8 @@ class Error(Exception):
 class InvalidTestcaseError(Error):
   """Error thrown when there is an attempt to access an invalid test case."""
 
-  def __init__(self):
-    super().__init__('Invalid test case.')
+  def __init__(self, testcase_id):
+    super().__init__(f'Invalid test case {testcase_id!r}.')
 
 
 class InvalidFuzzerError(Error):
--- a/src/clusterfuzz/_internal/datastore/data_handler.py
+++ b/src/clusterfuzz/_internal/datastore/data_handler.py
@@ -110,13 +110,17 @@ def get_domain():
 
 
 def get_testcase_by_id(testcase_id):
-  """Return the testcase with the given id, or None if it does not exist."""
-  if not testcase_id or not str(testcase_id).isdigit() or int(testcase_id) == 0:
-    raise errors.InvalidTestcaseError
+  """Return the testcase with the given id.
+  Raises InvalidTestcaseError if no such testcase exists.
+  """
+  try:
+    parsed_id = int(testcase_id)
+  except ValueError:
+    raise errors.InvalidTestcaseError(testcase_id)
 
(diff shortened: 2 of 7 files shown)


letitz:
@jonathanmetzman PTAL.

src/clusterfuzz/_internal/base/errors.py
```diff
@@ -50,8 +50,8 @@ class Error(Exception):
 class InvalidTestcaseError(Error):
   """Error thrown when there is an attempt to access an invalid test case."""
 
-  def __init__(self):
-    super().__init__('Invalid test case.')
+  def __init__(self, testcase_id):
+    super().__init__(f'Invalid test case {testcase_id!r}.')
```
jonathanmetzman, line 54: what is `!r`?

jonathanmetzman (commented):
Thanks for cleaning up some cruft!

src/clusterfuzz/_internal/base/errors.py
letitz: It formats with `repr()`, so that `"invalid-garbage"` would be rendered as `Invalid test case "invalid-garbage".`.

jonathanmetzman: Ah so you can omit the quotes?

jonathanmetzman (approved):
lgtm

src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -110,13 +110,20 @@ def get_domain():
...
-    raise errors.InvalidTestcaseError
+  """Return the testcase with the given id.
+  Raises InvalidTestcaseError if no such testcase exists.
+  """
+  try:
+    parsed_id = int(testcase_id)
+  except ValueError:
+    raise errors.InvalidTestcaseError(testcase_id)
+
+  if parsed_id == 0:
```
jonathanmetzman, line 121: not so sure this is necessary.

src/clusterfuzz/_internal/base/errors.py
letitz: `repr(foo)` basically returns a string that evaluates to `foo` (at least for most basic types): https://docs.python.org/3/library/functions.html#repr. It's useful for debugging, since you can tell the difference between 12 and "12", b"12", Error("12"), etc.

src/clusterfuzz/_internal/datastore/data_handler.py
letitz: Agreed. We might want to check for non-positive values in general, but the ndb lookup should fail anyway if the value is wrong. It seems just a bit faster to check locally first.

jonathanmetzman: Yeah, I probably wouldn't have included this check but I guess since it was already there.

PR #3416, opened 2023-10-06
Make check_for_bad_build uworker safe

alhijazi:
@jonathanmetzman could you please take a look?

### DIFF
--- a/src/clusterfuzz/_internal/bot/testcase_manager.py
+++ b/src/clusterfuzz/_internal/bot/testcase_manager.py
@@ -1171,18 +1171,22 @@ def check_for_bad_build(job_type, crash_revision):
 
   # Any of the conditions below indicate that bot is in a bad state and it is
   # not caused by the build itself. In that case, just exit.
-  build_state = data_handler.get_build_state(job_type, crash_revision)
   if is_bad_build and utils.sub_string_exists_in(BAD_STATE_HINTS, output):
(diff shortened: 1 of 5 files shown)


src/clusterfuzz/_internal/bot/testcase_manager.py
```diff
@@ -1171,18 +1171,22 @@ def check_for_bad_build(job_type, crash_revision):
...
   if is_bad_build and utils.sub_string_exists_in(BAD_STATE_HINTS, output):
     logs.log_fatal_and_exit(
         'Bad bot environment detected, exiting.',
         output=build_run_console_output,
         snapshot=process_handler.get_runtime_snapshot())
 
+  return is_bad_build, crash_result.should_ignore(), build_run_console_output
+
+
+def update_build_metadata(job_type, crash_revision, is_bad_build,
```
jonathanmetzman, line 1183: So is this going to be done on the tworker in postprocess?

alhijazi: Yes, `update_build_metadata` queries the database thru the call to `data_handler.get_build_state` and then conditionally writes back with `data_handler.add_build_metadata`.
The next step would be to transmit the build data to the tworker postprocess and then call it from there.

jonathanmetzman (approved):
lgtm

Issue #1609, opened 2020-03-18
Use google.cloud.logging for bot logging.


oliverchang:
Rather than fluentd. This will simplify the docker environment a lot.

vitaliset:
Solved by #4668.

PR #3422, opened 2023-10-11
Delete fluentd.

oliverchang:
Use google-cloud-logging instead.

#1609

DO NOT MERGE. This requires google-cloud-logging to be upgraded past 3.0 first.


oliverchang:
@vitorguidi this would be a good upgrade to do once the Python upgrade sticks.

vitaliset:
Solved by #4668.

PR #3423, opened 2023-10-11
Revert "Make check_for_bad_build uworker safe (#3416)"

jonathanmetzman:
Causes this error: https://pantheon.corp.google.com/errors/detail/CNzroovjsbCc_QE;time=P30D?project=google.com:clusterfuzz&utm_source=error-reporting-notification&utm_medium=email&utm_content=new-error&rapt=AEjHL4OLAyDr3QmUrhtV3sichgUwbbJvF1LQzbMP9wuIrDZX_A68hxubp4biEPUIEPcK5wQAx-EhTPkxpGQqxlnPtxNrqpvUag&e=-13802955&mods=logs_tg_prod

### DIFF
--- a/src/clusterfuzz/_internal/bot/testcase_manager.py
+++ b/src/clusterfuzz/_internal/bot/testcase_manager.py
@@ -1171,22 +1171,18 @@ def check_for_bad_build(job_type, crash_revision):
 
   # Any of the conditions below indicate that bot is in a bad state and it is
   # not caused by the build itself. In that case, just exit.
+  build_state = data_handler.get_build_state(job_type, crash_revision)
   if is_bad_build and utils.sub_string_exists_in(BAD_STATE_HINTS, output):
     logs.log_fatal_and_exit(
         'Bad bot environment detected, exiting.',
         output=build_run_console_output,
         snapshot=process_handler.get_runtime_snapshot())
 
-  return is_bad_build, crash_result.should_ignore(), build_run_console_output
-
-
-def update_build_metadata(job_type, crash_revision, is_bad_build,
-                          should_ignore_crash_result, build_run_console_output):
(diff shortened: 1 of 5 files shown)


src/clusterfuzz/_internal/bot/testcase_manager.py
```diff
@@ -1113,12 +1113,10 @@ def setup_user_profile_directory_if_needed(user_profile_directory):
 
 
 def check_for_bad_build(job_type, crash_revision):
-  """Returns result of bad build check i.e. crashes on startup. Results include
-  whether the build is bad, whether the crash should be ignored, the
-  build_run_console_output."""
+  """Return true if the build is bad, i.e. crashes on startup."""
   # Check the bad build check flag to see if we want do this.
   if not environment.get_value('BAD_BUILD_CHECK'):
-    return False, False, None
```
alhijazi, line 1121: I think the issue was possibly coming from here, we should return empty string `''` instead of `None`. That was the default value as you can see on line 1142

oliverchang (approved):

src/clusterfuzz/_internal/bot/testcase_manager.py
jonathanmetzman: I think your suggestion fixes the immediate problem, but I'm not so sure it's a proper fix. I feel like update_build_metadata should not be called at all in this case.
Let's reland this tomorrow. I have to go offline so I'll revert rather than leave it broken. I can reland first thing in the

alhijazi: You are right, we need a way to skip calling `update_build_metadata` when `BAD_BUILD_CHECK` is not set and `check_for_bad_build` returns early.

PR #3424, opened 2023-10-12
[build manager] fix unzipping matching callback

paulsemel:
The allowed filenames are matched against the root directory, which happens to be controlled by the packagers, and thus doesn't really need to be matched against.

This fixes Centipede full disk issue.

### DIFF
--- a/src/clusterfuzz/_internal/build_management/build_manager.py
+++ b/src/clusterfuzz/_internal/build_management/build_manager.py
@@ -235,6 +235,13 @@ def _get_file_match_callback():
      with one of the |blocklisted_extensions|.
     """
     path_components = os.path.normpath(filepath).split(os.sep)
+
+    # The first component is the root directory name, which is controlled by
+    # the packaging tool. It doesn't really make sense matching against it, or
+    # otherwise we'd either match nothing, or match all files, which defeats
+    # the purpose of this callback.
+    if len(path_components) > 1:
+      path_components = path_components[1:]
     # Is it an allowlisted binary?
     if any(
         component.startswith(allowlisted_names)


paulsemel:
@jonathanmetzman @oliverchang How does packaging work for other platforms? Is there always a `root` directory for those as well or not?

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -235,6 +235,13 @@ def file_match_callback(filepath):
      with one of the |blocklisted_extensions|.
     """
     path_components = os.path.normpath(filepath).split(os.sep)
+
+    # The first component is the root directory name, which is controlled by
+    # the packaging tool. It doesn't really make sense matching against it, or
+    # otherwise we'd either match nothing, or match all files, which defeats
+    # the purpose of this callback.
```
jonathanmetzman, line 242: Not sure about this. In OSS-Fuzz we don't have a root directory. Maybe Chrome can get rid of theirs to match oss-fuzz?

oliverchang:
Thanks for this change!

> This fixes Centipede full disk issue.

Could you please clarify how this fixes disk spaces issues arising from Centipede usage?

paulsemel:
> Thanks for this change!
> 
> > This fixes Centipede full disk issue.
> 
> Could you please clarify how this fixes disk spaces issues arising from Centipede usage?

Basically, we're matching the list of words against both file names and path components. Unfortunately, the archive names for centipede base fuzzer for chrome are starting with the work `centipede`. This leads to deflating the entire archive instead of only the fuzzing target + the `centipede` binary (which is the reason why `centipede` is in this allow list).

I guess there are two options to fix this issue:
- revise how we're actually matching the path components against the allow list.
- change the way chrome is packaging the fuzzers and remove the `root` directory in the archives.

I already started the second option change here: https://chromium-review.googlesource.com/c/chromium/tools/build/+/4936670, but I am not 100% sure this solution will work correctly with ClusterFuzz since I am unsure about how ClusterFuzz handles the fact that there might or might not be a `root` dir in the archive.

PR #3432, opened 2023-10-17
Add 'collaborators' Field to Issue

micahbales:
This is a first step in creating a security workflow in Google Issue Tracker that auto-assigns a collaborator to the issue.

### DIFF
--- a/src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
+++ b/src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
@@ -74,6 +74,10 @@ class Issue(issue_tracker.Issue):
     ccs = data['issueState'].get('ccs', [])
     self._ccs = issue_tracker.LabelStore(
         [user['emailAddress'] for user in ccs if 'emailAddress' in user])
+    collaborators = data['issueState'].get('collaborators', [])
+    self._collaborators = issue_tracker.LabelStore([
+        user['emailAddress'] for user in collaborators if 'emailAddress' in user
+    ])
     labels = [
         str(hotlist_id)
         for hotlist_id in data['issueState'].get('hotlistIds', [])
@@ -88,6 +92,7 @@ class Issue(issue_tracker.Issue):
     """Resets diff tracking."""
     self._changed.clear()
     self._ccs.reset_tracking()
+    self._collaborators.reset_tracking()
     self._labels.reset_tracking()
     self._components.reset_tracking()
 
@@ -197,6 +202,11 @@ class Issue(issue_tracker.Issue):
     return self._ccs
 
   @property
+  def collaborators(self):
+    """The issue collaborators list."""
+    return self._collaborators
+
+  @property
   def labels(self):
     """The issue labels list."""
     return self._labels
@@ -304,6 +314,8 @@ class Issue(issue_tracker.Issue):
     self._add_update_single(update_body, added, removed, 'title', 'title')
     self._add_update_collection(update_body, added, removed, 'ccs', 'ccs',
                                 _make_users)
+    self._add_update_collection(update_body, added, removed, 'collaborators',
+                                'collaborators', _make_users)
     update_body['addMask'] = ','.join(added)
     update_body['removeMask'] = ','.join(removed)
     if notify:
@@ -383,6 +395,9 @@ class Issue(issue_tracker.Issue):
       ccs = list(self._ccs)
       if ccs:
         self._data['issueState']['ccs'] = _make_users(ccs)
+      collaborators = list(self._collaborators)
+      if collaborators:
+        self._data['issueState']['collaborators'] = _make_users(collaborators)
       self._data['issueState']['hotlistIds'] = [
           int(label) for label in self.labels
       ]
@@ -515,6 +530,17 @@ class Action(issue_tracker.Action):
     return change_list
 
   @property
+  def collaborators(self):
+    """The issue collaborators change list."""
+    removed, added = self._get_field_update_changes('collaborators')
+    change_list = issue_tracker.ChangeList()
+    if added:
+      change_list.added.extend(added)
+    if removed:
+      change_list.removed.extend(removed)
+    return change_list
+
+  @property
   def labels(self):
     """The issue labels change list."""
     # We need to use the snake_case version of the field here, as that's the
@@ -585,6 +611,7 @@ class IssueTracker(issue_tracker.IssueTracker):
         'issueState': {
             'componentId': self._default_component_id,
             'ccs': [],
+            'collaborators': [],
             'hotlistIds': [],
         }
     }
--- a/src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
+++ b/src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
@@ -202,11 +202,6 @@ class Issue(issue_tracker.Issue):
     return self._ccs
 
   @property
-  def collaborators(self):
-    """The issue collaborators list."""
-    return self._collaborators
-
-  @property
   def labels(self):
     """The issue labels list."""
     return self._labels
@@ -314,7 +309,7 @@ class Issue(issue_tracker.Issue):
     self._add_update_single(update_body, added, removed, 'title', 'title')
     self._add_update_collection(update_body, added, removed, 'ccs', 'ccs',
                                 _make_users)
-    self._add_update_collection(update_body, added, removed, 'collaborators',
+    self._add_update_collection(update_body, added, removed, '_collaborators',
                                 'collaborators', _make_users)
     update_body['addMask'] = ','.join(added)
     update_body['removeMask'] = ','.join(removed)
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
```diff
@@ -514,6 +529,17 @@ def ccs(self):
       change_list.removed.extend(removed)
     return change_list
 
+  @property
+  def collaborators(self):
```
oliverchang, line 533: Do we need to add this to the base class as well? i.e. https://github.com/google/clusterfuzz/blob/709fbb62ca1e2a809dcc6b197076d97d75263115/src/clusterfuzz/_internal/issue_management/issue_tracker.py#L108

oliverchang (commented):
nice! is there some context for how this is planned to be used that you can share?

src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
micahbales: When we are done with our changes, collaborators should only be used internally in Issue, and only for Google Issue Tracker's implementation. I've gone ahead and added an update to explicitly treat _collaborators as an internal entity, so there's no confusion about the public API.

micahbales:
> nice! is there some context for how this is planned to be used that you can share?

The context here is that this is the first step in building a save and update flow for issues that are security bugs. Whenever we have a security bug being filed, we want to automatically assign Chrome Security to the issue as a member of collaborators. We also want to set the Issue Access Level to Limited Viewing (LIMIT_VIEW).

The next steps after this will be: 1) Adding IAL as a private value for Issue; 2) adding a step in Issue.save() to check for whether the type of the issue is Security-Bug - and if so, applying the collaborator and IAL setting; and 3) adding similar behavior when updating an issue.

One thing that we want to call out here, is that we are depending on the yaml config to set the label 'Type-Security-Bug' in order for this feature to work. This seems like a workable solution to us, but we want to make sure that you're satisfied with this direction, as well. We prefer this route for ease, but if this presented a problem, another possible strategy that occurs to us is that we could add new fields for the .yaml file in order to introduce specific collaborators and IAL settings for the security rule.

oliverchang (approved):

oliverchang:
> > nice! is there some context for how this is planned to be used that you can share?
> 
> The context here is that this is the first step in building a save and update flow for issues that are security bugs. Whenever we have a security bug being filed, we want to automatically assign Chrome Security to the issue as a member of collaborators. We also want to set the Issue Access Level to Limited Viewing (LIMIT_VIEW).
> 
> The next steps after this will be: 1) Adding IAL as a private value for Issue; 2) adding a step in Issue.save() to check for whether the type of the issue is Security-Bug - and if so, applying the collaborator and IAL setting; and 3) adding similar behavior when updating an issue.
> 
> One thing that we want to call out here, is that we are depending on the yaml config to set the label 'Type-Security-Bug' in order for this feature to work. This seems like a workable solution to us, but we want to make sure that you're satisfied with this direction, as well. We prefer this route for ease, but if this presented a problem, another possible strategy that occurs to us is that we could add new fields for the .yaml file in order to introduce specific collaborators and IAL settings for the security rule.

Hmm, I'd like to avoid introducing chrome hardcodes in the issue tracker implementation here, since I think that's what this would entail. 

I'd instead prefer that we add new configuration fields to the yaml. 

e.g.

```yaml
chromium:
  type: google_issue_tracker
  policies:
    ...
    security:
      # Rules to apply to new security issues only (in addition to "all").
      labels:
        - ...
      ext_collaborators:
        - x@b.com
      ext_issue_access_level:
        - x
``` 

Let's also add a "ext_" prefix to them to make it clear that they're google-issue-tracker specific extensions, and that way we can avoid adding these to the base `Issue` class etc.

oliverchang (approved):

oliverchang:
> > > > nice! is there some context for how this is planned to be used that you can share?
> > > 
> > > 
> > > The context here is that this is the first step in building a save and update flow for issues that are security bugs. Whenever we have a security bug being filed, we want to automatically assign Chrome Security to the issue as a member of collaborators. We also want to set the Issue Access Level to Limited Viewing (LIMIT_VIEW).
> > > The next steps after this will be: 1) Adding IAL as a private value for Issue; 2) adding a step in Issue.save() to check for whether the type of the issue is Security-Bug - and if so, applying the collaborator and IAL setting; and 3) adding similar behavior when updating an issue.
> > > One thing that we want to call out here, is that we are depending on the yaml config to set the label 'Type-Security-Bug' in order for this feature to work. This seems like a workable solution to us, but we want to make sure that you're satisfied with this direction, as well. We prefer this route for ease, but if this presented a problem, another possible strategy that occurs to us is that we could add new fields for the .yaml file in order to introduce specific collaborators and IAL settings for the security rule.
> > 
> > 
> > Hmm, I'd like to avoid introducing chrome hardcodes in the issue tracker implementation here, since I think that's what this would entail.
> > I'd instead prefer that we add new configuration fields to the yaml.
> > e.g.
> > ```yaml
> > chromium:
> >   type: google_issue_tracker
> >   policies:
> >     ...
> >     security:
> >       # Rules to apply to new security issues only (in addition to "all").
> >       labels:
> >         - ...
> >       ext_collaborators:
> >         - x@b.com
> >       ext_issue_access_level:
> >         - x
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Let's also add a "ext_" prefix to them to make it clear that they're google-issue-tracker specific extensions, and that way we can avoid adding these to the base `Issue` class etc.
> 
> Thanks for the comment Oliver!
> 
> We are trying to understand what changes will be required to add new configuration fields. Does something like this look roughly right? - [master...rmistry:clusterfuzz:new-config-fields](https://github.com/google/clusterfuzz/compare/master...rmistry:clusterfuzz:new-config-fields)

Not quite. I'm thinking something like the following to hide the details of these extension fields from the general interface. Something like this:

```python

class Issue:
  # New 
  def apply_extension_fields(extension_fields):
     return 

class NewIssuePolicy:
  # New
  extension_fields: dict


issue.apply_extension_fields(issue_policy.extension_fields)

```

We could auto-fill extension fields based on the `ext_` prefix. WDYT?  

> 
> Also, even with that approach, Micah's 2 PRs should be hopefully good to go?-
> 
> * [Add 'collaborators' Field to Issue #3432](https://github.com/google/clusterfuzz/pull/3432) (adding collaborators)
> * [Add Issue Access Limit to Issue Class as Private Data #3437](https://github.com/google/clusterfuzz/pull/3437) (adding issue_access_level)

Approved these two!

rmistry:
> > > > > nice! is there some context for how this is planned to be used that you can share?
> > > > 
> > > > 
> > > > The context here is that this is the first step in building a save and update flow for issues that are security bugs. Whenever we have a security bug being filed, we want to automatically assign Chrome Security to the issue as a member of collaborators. We also want to set the Issue Access Level to Limited Viewing (LIMIT_VIEW).
> > > > The next steps after this will be: 1) Adding IAL as a private value for Issue; 2) adding a step in Issue.save() to check for whether the type of the issue is Security-Bug - and if so, applying the collaborator and IAL setting; and 3) adding similar behavior when updating an issue.
> > > > One thing that we want to call out here, is that we are depending on the yaml config to set the label 'Type-Security-Bug' in order for this feature to work. This seems like a workable solution to us, but we want to make sure that you're satisfied with this direction, as well. We prefer this route for ease, but if this presented a problem, another possible strategy that occurs to us is that we could add new fields for the .yaml file in order to introduce specific collaborators and IAL settings for the security rule.
> > > 
> > > 
> > > Hmm, I'd like to avoid introducing chrome hardcodes in the issue tracker implementation here, since I think that's what this would entail.
> > > I'd instead prefer that we add new configuration fields to the yaml.
> > > e.g.
> > > ```yaml
> > > chromium:
> > >   type: google_issue_tracker
> > >   policies:
> > >     ...
> > >     security:
> > >       # Rules to apply to new security issues only (in addition to "all").
> > >       labels:
> > >         - ...
> > >       ext_collaborators:
> > >         - x@b.com
> > >       ext_issue_access_level:
> > >         - x
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Let's also add a "ext_" prefix to them to make it clear that they're google-issue-tracker specific extensions, and that way we can avoid adding these to the base `Issue` class etc.
> > 
> > 
> > Thanks for the comment Oliver!
> > We are trying to understand what changes will be required to add new configuration fields. Does something like this look roughly right? - [master...rmistry:clusterfuzz:new-config-fields](https://github.com/google/clusterfuzz/compare/master...rmistry:clusterfuzz:new-config-fields)
> 
> Not quite. I'm thinking something like the following to hide the details of these extension fields from the general interface. Something like this:
> 
> ```python
> class Issue:
>   # New 
>   def apply_extension_fields(extension_fields):
>      return 
> 
> class NewIssuePolicy:
>   # New
>   extension_fields: dict
> 
> 
> issue.apply_extension_fields(issue_policy.extension_fields)
> ```
> 
> We could auto-fill extension fields based on the `ext_` prefix. WDYT?

In https://github.com/google/clusterfuzz/pull/3442
Micah uses the extension fields approach. Does that one look OK? (it does not do the dictionary + auto-filling yet).

(Micah is OOO till Wednesday which is why I have been helping with his PRs)

> 
> > Also, even with that approach, Micah's 2 PRs should be hopefully good to go?-
> > 
> > * [Add 'collaborators' Field to Issue #3432](https://github.com/google/clusterfuzz/pull/3432) (adding collaborators)
> > * [Add Issue Access Limit to Issue Class as Private Data #3437](https://github.com/google/clusterfuzz/pull/3437) (adding issue_access_level)
> 
> Approved these two!

Thanks!

rmistry:
> In #3442 Micah uses the extension fields approach. Does that one look OK? (it does not do the dictionary + auto-filling yet).
> 
> (Micah is OOO till Wednesday which is why I have been helping with his PRs)
> 

Thanks for your comments on that PR! Will respond there.

PR #3433, opened 2023-10-17
Integrate with issuetracker public API + Some cleanup

rmistry:
This PR contains the changes necessary to get ClusterFuzz working with the issuetracker public API- It fixes the discovery URL and completes the `client.build_http()` method.

The integration with public API was verified locally by using a GCP service account and tested against a test component. Example created issue: b/306010501
The steps taken to setup the local environment have been documented in the `issue_tracker.py` file.

Cleanup has also been done-
* The `_override_priority_and_type` method has been removed. It was not used.
* The `config`dict  has been replaced by `component_id`. That was the only field accessed from the `config` dict.

### DIFF
--- a/src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
+++ b/src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
@@ -352,8 +352,6 @@ class Issue(issue_tracker.Issue):
     if self._is_new:
       priority = _extract_label(self.labels, 'Pri-')
       issue_type = _extract_label(self.labels, 'Type-') or 'BUG'
-      issue_type = 'BUG'
-      priority = None
       self._data['issueState']['type'] = issue_type
       if priority:
         self._data['issueState']['priority'] = priority
--- a/src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
+++ b/src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
@@ -347,33 +347,13 @@ class Issue(issue_tracker.Issue):
                                       issueId=str(self.id)))
     return result
 
-  def _override_priority_and_type(self):
-    """Determines whether if we should override the priority and type."""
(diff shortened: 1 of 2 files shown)


rmistry:
@oliverchang PTAL

src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
```diff
@@ -665,6 +643,30 @@ def _get_query(keywords, only_open):
   return query
 
 
-def get(project, config, issue_tracker_client=None):
+def get(project, component_id, issue_tracker_client=None):
```
oliverchang, line 646: This can't be changed and is part of the contract for an issue tracker to be "registered" for ClusterFuzz. 

see e.g. other implementations that do the same thing https://github.com/google/clusterfuzz/blob/709fbb62ca1e2a809dcc6b197076d97d75263115/src/clusterfuzz/_internal/issue_management/monorail/__init__.py#L307

src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
```diff
@@ -544,10 +522,10 @@ def components(self):
 class IssueTracker(issue_tracker.IssueTracker):
   """Google issue tracker implementation."""
 
-  def __init__(self, project, http_client, config):
+  def __init__(self, project, http_client, component_id):
     self._project = project
     self._client = http_client
-    self._default_component_id = config['default_component_id']
+    self._component_id = component_id
```
oliverchang, line 528: Can we perhaps keep the `_default` bit here? it makes it a bit clearer what it's meant for: as a fallback for when there's no more testcase-specific component available.

rmistry: Done (I did not know it could be overridden by testcases but that makes sense)

rmistry:
Back to @oliverchang for review.

oliverchang (approved):

Issue #3425, opened 2023-10-12
Inputs for Progression Task failing because they are over 2 GB


jonathanmetzman:
https://pantheon.corp.google.com/errors/detail/CIjMovecxqqfoQE;time=P7D?e=-13802955&mods=logs_tg_prod&project=google.com:clusterfuzz

jonathanmetzman:
This is fixed.

PR #3435, opened 2023-10-18
Get list of bad revisions instead of bad builds

alhijazi:
@jonathanmetzman Could you please take a look?

### DIFF
--- a/src/clusterfuzz/_internal/build_management/build_manager.py
+++ b/src/clusterfuzz/_internal/build_management/build_manager.py
@@ -1109,7 +1109,8 @@ def get_primary_bucket_path():
 
 def get_revisions_list(bucket_path, bad_revisions, testcase=None):
   """Returns a sorted ascending list of revisions from a bucket path, excluding
-  bad build revisions and testcase crash revision (if any)."""
+  bad build revisions. Testcase crash revision is not excluded from the list
+  even if it appears in the bad_revisions list."""
   revision_pattern = revisions.revision_pattern_from_build_bucket_path(
       bucket_path)
 
--- a/src/clusterfuzz/_internal/build_management/build_manager.py
+++ b/src/clusterfuzz/_internal/build_management/build_manager.py
@@ -1140,12 +1140,13 @@ def get_revisions_list(bucket_path, bad_revisions, testcase=None):
 
 def get_job_bad_revisions():
   job_type = environment.get_value('JOB_NAME')
-  bad_builds = list(
-      ndb_utils.get_all_from_query(
+
+  bad_revisions = list(
+      build.revision for build in ndb_utils.get_all_from_query(
           data_types.BuildMetadata.query(
               ndb_utils.is_true(data_types.BuildMetadata.bad_build),
-              data_types.BuildMetadata.job_type == job_type)))
-  bad_revisions = [build.revision for build in bad_builds]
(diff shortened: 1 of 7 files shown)


alhijazi:
Related: https://github.com/google/clusterfuzz/issues/3425

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -1125,27 +1125,29 @@ def get_revisions_list(bucket_path, bad_builds, testcase=None):
...
+
+  bad_revisions = list(
+      build.revision for build in ndb_utils.get_all_from_query(
           data_types.BuildMetadata.query(
               ndb_utils.is_true(data_types.BuildMetadata.bad_build),
-              data_types.BuildMetadata.job_type == job_type)))
-  return bad_builds
+              data_types.BuildMetadata.job_type == job_type),
+          projection=[data_types.BuildMetadata.revision]))
+  return bad_revisions
```
jonathanmetzman, line 1150: Nit: I would do this:
```
query=data_types.BuildMetadata(...)
return [build.revision for build in query]
```

jonathanmetzman (commented):
LGTM with nits

src/clusterfuzz/_internal/protos/uworker_msg.proto
```diff
@@ -35,7 +35,7 @@ message SetupInput {
 }
 
 message AnalyzeTaskInput {
-  repeated google.datastore.v1.Entity bad_builds = 1;
```
letitz, line 38: This is backwards-incompatible. If there are any protos of this type already stored somewhere where they could be read later, or if messages of this type are already being exchanged over the network between two binaries that will be rolled out separately, then it would be better to mark 1 as reserved and increment the field index to 2:

```
reserved 1;

repeated int64 bad_revisions = 2;
```

If not, then this is fine!

letitz (commented):
A couple of drive-by comments.

src/clusterfuzz/_internal/protos/uworker_msg.proto
alhijazi: I don't think we have to worry about backward compatibility. @jonathanmetzman correct me if I'm wrong.

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -1107,7 +1107,7 @@ def get_primary_bucket_path():
       'needs to be defined.')
 
 
-def get_revisions_list(bucket_path, bad_builds, testcase=None):
+def get_revisions_list(bucket_path, bad_revisions, testcase=None):
   """Returns a sorted ascending list of revisions from a bucket path, excluding
   bad build revisions and testcase crash revision (if any)."""
```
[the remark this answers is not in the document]
alhijazi: Indeed, I'll modify the description to clearly reflect that.

src/clusterfuzz/_internal/protos/uworker_msg.proto
jonathanmetzman: It wouldn't matter for now. The tasks are all executed on one machine so as soon as they pull in the new source, they won't have any old protos.

jonathanmetzman (approved):
LGTM

PR #3441, opened 2023-10-19
Adapt check bad builds

alhijazi:
@jonathanmetzman could you please take a look?
So basically, what I changed is that we're now returning `False, True, None` when `'BAD_BUILD_CHECK'` is not set. This leads `update_build_metadata` to return early.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
@@ -1768,18 +1768,16 @@ class FuzzingSession:
     logs.log('Checking for bad build.')
     crash_revision = environment.get_value('APP_REVISION')
 
-    (is_bad_build, should_ignore_crash_result,
-     build_run_console_output) = testcase_manager.check_for_bad_build(
-         self.job_type, crash_revision)
+    build_data = testcase_manager.check_for_bad_build(self.job_type,
+                                                      crash_revision)
     # TODO(https://github.com/google/clusterfuzz/issues/3008): Move this to
     # postprocess.
-    testcase_manager.update_build_metadata(
-        self.job_type, crash_revision, is_bad_build, should_ignore_crash_result,
-        build_run_console_output)
-    _track_build_run_result(self.job_type, crash_revision, is_bad_build)
-    if is_bad_build:
-      return uworker_io.UworkerOutput(
-          error_type=uworker_msg_pb2.ErrorType.UNHANDLED)
+    testcase_manager.update_build_metadata(self.job_type, crash_revision,
+                                           build_data)
+    _track_build_run_result(self.job_type, crash_revision,
+                            build_data.is_bad_build)
+    if build_data.is_bad_build:
+      return uworker_io.UworkerOutput(error=uworker_msg_pb2.ErrorType.UNHANDLED)
 
     # Data bundle directories can also have testcases which are kept in-place
     # because of dependencies.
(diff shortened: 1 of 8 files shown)


src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -1767,8 +1767,15 @@ def run(self):
     # If yes, bail out.
     logs.log('Checking for bad build.')
     crash_revision = environment.get_value('APP_REVISION')
-    is_bad_build = testcase_manager.check_for_bad_build(self.job_type,
-                                                        crash_revision)
+
+    (is_bad_build, should_ignore_crash_result,
+     build_run_console_output) = testcase_manager.check_for_bad_build(
```
letitz, line 1772: It might be time to introduce a small class to hold these - three return values feels like a lot.

Edit: maybe you intend to define a proto message to hold these? That would make sense to me.

src/clusterfuzz/_internal/bot/testcase_manager.py
```diff
@@ -1113,10 +1113,14 @@ def setup_user_profile_directory_if_needed(user_profile_directory):
 
 
 def check_for_bad_build(job_type, crash_revision):
-  """Return true if the build is bad, i.e. crashes on startup."""
+  """Returns result of bad build check i.e. crashes on startup. Results include
```
letitz, line 1116: This is quite unclear. Can we be more specific? Typically, I've seen Python methods documented with Arguments/Returns/Raises sections as follows:

```py
"""
Checks whether the target binary fails to execute at the given revision.

Arguments:
  job_type (str): The type of job we are executing on.
  crash_revision (int): The revision at which the target was built.

Returns a triple consisting of:
  is_bad_build (bool): Whether the target build is bad. If True, the target cannot
    be used for executing testcases.
  should_ignore_crash (bool): True iff the target crashed, but we should ignore it.
  build_run_console_output (Optional[str]): The crash output, containing stacktraces, or None
    if the target was not executed.
"""
```

Writing this out makes me realize 2 things:

-  `build_run_console_output` could be named `crash_stacktrace_output` to better align with `utils.get_crash_stacktrace_output`
- it could be initialized to the empty string in the early return, to avoid the unnecessary `None` vs `''` distinction

src/clusterfuzz/_internal/bot/testcase_manager.py
```diff
@@ -1171,18 +1175,24 @@ def check_for_bad_build(job_type, crash_revision):
...
   if is_bad_build and utils.sub_string_exists_in(BAD_STATE_HINTS, output):
     logs.log_fatal_and_exit(
         'Bad bot environment detected, exiting.',
         output=build_run_console_output,
         snapshot=process_handler.get_runtime_snapshot())
 
+  return is_bad_build, crash_result.should_ignore(), build_run_console_output
+
+
+def update_build_metadata(job_type, crash_revision, is_bad_build,
```
letitz, line 1187: Similar comment re: documentation.

letitz (commented):
Drive-by review, because I was curious :)

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
jonathanmetzman: +1, let's go with a namedtuple.

jonathanmetzman (commented):
Almost good to land

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
alhijazi: ended up using a dataclass since I wanted to use the type from tests also.

alhijazi:
> Almost good to land

PR updated, the test failure comes from an issue on master.

PR #3450, opened 2023-10-23
Update default build target for flashing Android Cuttlefish

marktefftech:


jonathanmetzman (commented):
Do you know if this will affect chrome's android bots?

marktefftech:
> Do you know if this will affect chrome's android bots?

I don't know much about chrome's android bots. If you're referring to the new one's that Cory is setting up, then I don't think this is an issue.

The real question is what they have their `PLATFORM` set to. If they hit the below piece of code, then the answer is yes, they will also be affected:

https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/tasks/update_task.py#L173

Regardless, if chrome's android bots are still pointed to `git_main` branch and `cf_x86_64_phone-userdebug` target, then they wouldn't be working anyways since this combination doesn't exists.

This is what the current error looks like in the logs:

```
googleapiclient.errors.HttpError: <HttpError 404 when requesting https://androidbuildinternal.googleapis.com/android/internal/build/v2beta1/builds/10988442/cf_x86_64_phone-userdebug/attempts/latest/artifacts?alt=json returned "Build attempt for build_type: SUBMITTED build_id: 10988442 target_name: cf_x86_64_phone-userdebug build_attempt_id: latest not found">
```

jonathanmetzman:
> > Do you know if this will affect chrome's android bots?
> 
> I don't know much about chrome's android bots. If you're referring to the new one's that Cory is setting up, then I don't think this is an issue.

I mostly mean the ones they already have set up.
> 
> The real question is what they have their `PLATFORM` set to. If they hit the below piece of code, then the answer is yes, they will also be affected:
> 
> https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/tasks/update_task.py#L173

I think it is hit.

> 
> Regardless, if chrome's android bots are still pointed to `git_main` branch and `cf_x86_64_phone-userdebug` target, then they wouldn't be working anyways since this combination doesn't exists.

Not sure if it's pinned to this @letitz @oliverchang Do we know?

> 
> This is what the current error looks like in the logs:
> 
> ```
> googleapiclient.errors.HttpError: <HttpError 404 when requesting https://androidbuildinternal.googleapis.com/android/internal/build/v2beta1/builds/10988442/cf_x86_64_phone-userdebug/attempts/latest/artifacts?alt=json returned "Build attempt for build_type: SUBMITTED build_id: 10988442 target_name: cf_x86_64_phone-userdebug build_attempt_id: latest not found">
> ```

src/clusterfuzz/_internal/platforms/android/flash.py
```diff
@@ -121,10 +122,9 @@ def flash_to_latest_build_if_needed():
   branch = environment.get_value('BUILD_BRANCH')
   target = environment.get_value('BUILD_TARGET')
   if not target:
-    # We default to userdebug configuration.
     build_params = settings.get_build_parameters()
     if build_params:
-      target = build_params.get('target') + '-userdebug'
```
jonathanmetzman, line 127: This issue only affects  cuttlefish right? Can you add a check for cuttlefish?

marktefftech: Added. This helps with clarity, but Cuttlefish emulators are the only users of this code. Flashing is disabled (through ClusterFuzz, we handle it in the lab) for physical devices.

ParisMeuleman:
> > The real question is what they have their `PLATFORM` set to. If they hit the below piece of code, then the answer is yes, they will also be affected:
> > https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/tasks/update_task.py#L173
> 
> I think it is hit.
> 
Our platform is either ANDROID or ANDROID_X86.
I am pretty sure this is hit. I am curious how the Haiku jobs are setup if they don't hit this though?

> > Regardless, if chrome's android bots are still pointed to `git_main` branch and `cf_x86_64_phone-userdebug` target, then they wouldn't be working anyways since this combination doesn't exists.
> 
> Not sure if it's pinned to this @letitz @oliverchang Do we know?
we have from puppet ("usr_local_bin_clusterfuzz.sh.erb" script): 
export BUILD_BRANCH="git_pi-release"
export BUILD_TARGET="walleye-userdebug"

I am pretty sure this was overloaded later though on the pixel 6 hosts, as I've seen [this log](https://github.com/google/clusterfuzz/blob/a30612dabc4b65181096df5c00977187bf27e77d/src/clusterfuzz/_internal/platforms/android/flash.py#L78C42-L78C50
)  with something else. I'm currently not able to connect to the hosts though.

In any case, I don't think this will impact our bots, but maybe we can keep the current behavior for non-cuttlefish devices to play it safe, e.g.:
```
    if build_params:
        target = build_params.get('target') + CUTTLEFISH_FLASH_DEFAULT_BUILD_TARGET
      else:
        target = build_params.get('target') + FLASH_DEFAULT_BUILD_TARGET
```

src/clusterfuzz/_internal/platforms/android/flash.py
jonathanmetzman: I still think there's a change here that affects Chrome.
Could you change it to:
```
    if build_params:
        target = build_params.get('target') + FLASH_DEFAULT_BUILD_TARGET
      else:
        target = build_params.get('target') + '-userdebug'
```

marktefftech: Done.

marktefftech:
@ParisMeuleman When the `RUN_TIMEOUT` environment variable is defined, then the device will never be flashed since it is assumed the device was flashed by a harness before testing:

https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/platforms/android/flash.py#L93-L97

ParisMeuleman:
> @ParisMeuleman When the `RUN_TIMEOUT` environment variable is defined, then the device will never be flashed since it is assumed the device was flashed by a harness before testing:
> 
> https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/platforms/android/flash.py#L93-L97

RUN_TIMEOUT is defined so there shouldn't be any call to that then, I thought remembering logs of flashing failed, but couldn't find them on GCP/Logs, at this point I will consider that I am confusing with other things. 
LGTM then.

PR #3452, opened 2023-10-24
Add pyright support.

letitz:
Adds support for type checking python code using pyright.

Also adds a dependency on `mypy-protobuf` in order to generate `.pyi` type stubs for generated protobuf code, without which the type checker gets very confused.

pyright was chosen because it seems like the leading python type checker these days, and because it has broad python3 version support.

This PR also integrates pyright with `butler.py lint`, such that changed files can optionally be type-checked if the `--type-check` command-line argument is set. This allows for local experimentation before we decide on the next steps.


letitz:
cc @alhijazi who was curious about this.

jonathanmetzman:
Why use this instead of pytype? Is it because pyright doesn't fail on 3.11?

Pipfile
```diff
@@ -8,6 +8,8 @@ cryptography = "==3.4.8"
 crcmod = "==1.7"
 protobuf = "==3.20.3"
 psutil = "==5.7.0"
+pyright = "*"
+mypy-protobuf = "*"
```
jonathanmetzman, line 12: Think this needs to go into dev-packages instead.

src/clusterfuzz/_internal/protos/generate.sh
```diff
@@ -18,7 +18,8 @@ SCRIPT_DIR=$( readlink -f $( dirname ${BASH_SOURCE[0]} ) )
 PARENT_DIR=$( dirname $( dirname $( dirname ${SCRIPT_DIR} ) ) )
 echo "PARENT_DIR $PARENT_DIR"
 python -m grpc_tools.protoc --proto_path=$PARENT_DIR/third_party:$PARENT_DIR \
-  --python_out=$PARENT_DIR --grpc_python_out=$PARENT_DIR $SCRIPT_DIR/*.proto
+  --python_out=$PARENT_DIR --grpc_python_out=$PARENT_DIR \
```
jonathanmetzman, line 21: Weird indendation.

src/clusterfuzz/_internal/protos/generate.sh
```diff
@@ -36,7 +37,7 @@ read -r -d '' COPYRIGHT_HEADER <<EOF
 # limitations under the License.
 EOF
 
-for generated in $SCRIPT_DIR/*pb2*.py; do
+for generated in $SCRIPT_DIR/*pb2*.py{,i}; do
```
jonathanmetzman, line 40: What does this addittion do?

Pipfile
jonathanmetzman, line 12: Why is mypy-protobuf needed?

letitz: Good point, will fix.

src/clusterfuzz/_internal/protos/generate.sh
letitz: I'll put each arg on its own line.

src/clusterfuzz/_internal/protos/generate.sh
letitz: mypy-protobuf generates type stubs as `.pyi` files, which need to be copied to the right place as well as the `_pb2.py` files.

letitz:
Right, as I recall one of the problems you ran into with pytype is that it does not support 3.11, whereas this does. It also seems to have some traction generally as a good python type checker, faster and less idiosyncratic than mypy.

Pipfile
letitz: Had to upgrade `nodeenv` as well to make it work, for some reason. Hopefully that's ok.

letitz:
Rebased this PR, and added it to `python butler.py lint` in the way we chatted about last week: only changed files are typechecked. This is still likely to raise a ton of errors on new PRs that touch ~anything, but at least it scopes it down to what's being changed in the PR.

This approach also won't prevent breakages from making it into the repo (like the linter), since the checks do not propagate through dependency relationships: we do not check files that depend on changed files.

Pipfile
jonathanmetzman: Should be.

jonathanmetzman: ACK

Pipfile
```diff
@@ -15,14 +15,16 @@ Fabric = "==1.14.1"
 grpcio-tools = "==1.62.2"
 gunicorn = "*"
 isort = "*"
-nodeenv = "==1.0.0"
+mypy-protobuf = "*"
```
jonathanmetzman, line 18: Can we pin these to a particular version?

src/clusterfuzz/_internal/protos/generate.sh
jonathanmetzman: ACK

Pipfile
letitz: Sure! Done.

letitz:
Latest commit adds a `--type-check` argument to `butler.py lint` that controls whether to run the type checker or not, it defaults to false. I can switch it back to true if you have appetite for aggressively fixing type-checking errors on every PR in the near future. Another approach would be to default it to true, but set it to false on CI. That way PRs aren't blocked from landing but local linting surfaces errors. WDYT?

letitz:
Another option is to set pyright to only output warnings. This annoyingly requires explicitly listing all diagnostic types and setting them to `"warning"` in `pyrightconfig.json`, but it's not terribly hard.

jonathanmetzman:
> Latest commit adds a `--type-check` argument to `butler.py lint` that controls whether to run the type checker or not, it defaults to false. I can switch it back to true if you have appetite for aggressively fixing type-checking errors on every PR in the near future. Another approach would be to default it to true, but set it to false on CI. That way PRs aren't blocked from landing but local linting surfaces errors. WDYT?

I think let's merge this as is, I'll play with it, and determine if it's ok to run this on every PR?

letitz:
> I think let's merge this as is, I'll play with it, and determine if it's ok to run this on every PR?

That seems fine to me! I'll amend the PR description for a better commit message.

letitz:
Tests are now passing, PTAL @jonathanmetzman :)

oliverchang (approved):
let's get this in to avoid things diverging.

PR #3453, opened 2023-10-24
Remove projection causing error in production.

jonathanmetzman:
Follow up to #3435


oliverchang (approved):
LGTM. this needs an index to work, but it's not clear if any optimizations caused by this is worth that.

jonathanmetzman:
Causes https://pantheon.corp.google.com/errors/detail/CL-t96Kh3IzDEg;filter=%5B%22%5C%22_datastore_api.py%5C%22%22%5D?project=google.com:clusterfuzz

PR #3459, opened 2023-10-25
Allow postprocess task to finish

jonathanmetzman:


jonathanmetzman:
TBR

src/clusterfuzz/_internal/base/tasks.py
```diff
@@ -448,6 +446,48 @@ def lease(self, _event=None):  # pylint: disable=arguments-differ
...
+
+  def __init__(self,
+               output_url_argument,
+               pubsub_message,
+               is_command_override=False):
+    command = 'postprocess'
+    job_type = 'none'
+    eta = None
+    high_end = False
+    grandparent_class = super(PubSubTask, self)
```
oliverchang, line 460: Why grandparent_class? Can we restructure the classes a bit instead a bit better to avoid this (or just rely on super() directly? )

src/clusterfuzz/_internal/base/tasks.py
```diff
@@ -448,6 +446,48 @@ def lease(self, _event=None):  # pylint: disable=arguments-differ
...
+    grandparent_class.__init__(command, output_url_argument, job_type, eta,
+                               is_command_override, high_end)
+    self._pubsub_message = pubsub_message
+
+  @contextlib.contextmanager
+  def lease(self, _event=None):  # pylint: disable=arguments-differ
```
oliverchang, line 466: Are there big differences from the other lease()? can we avoid duplicated code here?

PR #3460, opened 2023-10-25
Issue Update Security Flow (Google Issue Tracker)

micahbales:
In https://github.com/google/clusterfuzz/pull/3442 we added a save-issue security flow. This PR tidies up and ensures that this flow also works for updating an issue.


src/clusterfuzz/_internal/tests/appengine/libs/issue_management/google_issue_tracker/google_issue_tracker_test.py
```diff
@@ -429,6 +429,111 @@ def test_update_issue(self):
...
+    issue.assignee = 'assignee2@google.com'
+    issue.ccs.add('cc@google.com')
+    issue.components.add('9001')
+    issue.labels.add('12345')
+    issue.status = 'ASSIGNED'
+    issue.title = 'issue title2'
+    issue.save()
+    self.assertEqual(68828938, issue.id)
+    self.client.issues().modify.assert_has_calls([
+        mock.call(),
```
micahbales, line 495: It is not clear to me why we have an additional (empty) call here.

micahbales:
@oliverchang PTAL

src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
```diff
@@ -108,12 +108,16 @@ def _reset_tracking(self):
...
       for collaborator in extension_fields['_ext_collaborators']:
         self._collaborators.add(collaborator)
-    self._issue_access_limit = extension_fields.get(
-        '_ext_issue_access_limit') or ({
-            'access_level': IssueAccessLevel.LIMIT_NONE
-        })
+
+    if extension_fields.get('_ext_issue_access_limit'):
+      self._changed.add('_issue_access_limit')
+      self._issue_access_limit = extension_fields.get(
```
rmistry, line 117: Do not need the `get(..)` and the `or` here.

This can directly be `self._issue_access_limit = extension_fields['_ext_issue_access_limit']`

Might need an else for `self._issue_access_limit = {'access_level': IssueAccessLevel.LIMIT_NONE}`

micahbales: Ah, yes, thanks!

We shouldn't need an 'else' statement for the default, as we set that default on the issue at initialization.

src/clusterfuzz/_internal/tests/appengine/libs/issue_management/google_issue_tracker/google_issue_tracker_test.py
jonathanmetzman: Maybe use remote-pdb and check who is executing it twice?
https://pypi.org/project/remote-pdb/

jonathanmetzman:
Resolve conflicts pls

src/clusterfuzz/_internal/tests/appengine/libs/issue_management/google_issue_tracker/google_issue_tracker_test.py
rmistry: > It is not clear to me why we have an additional (empty) call here.

Looks like this is expected and is likely due to the mocking framework and happens in other places where we specify a `return_value`.
Eg: ` self.client.issues().issueUpdates().list().execute.mock_calls` also has `mock.calls()` in it's calls list. But we do not assert against it because it is not the call we are interested in.

If we did care about asserting against it then we would be doing a `self.assertEqual(mock.mock_calls, [...])` everywhere instead of `assert_has_calls`.

I went ahead and removed the `mock.call()` from here because:
1. It is not relevant to the assertion
2. To be consistent with the rest of the test file
3. Tests pass as expected without it

This PR should be good to go now.

rmistry:
> Resolve conflicts pls

Done.

rmistry:
Lint failure looks unrelated to this PR:
```
| src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py:434:14: E1101: Module 'clusterfuzz._internal.protos.uworker_msg_pb2' has no 'BuildData' member (no-member)
```

rmistry:
> Lint failure looks unrelated to this PR:
> 
> ```
> | src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py:434:14: E1101: Module 'clusterfuzz._internal.protos.uworker_msg_pb2' has no 'BuildData' member (no-member)
> ```

It is from this PR- https://github.com/google/clusterfuzz/pull/3441

jonathanmetzman:
Pull again. I've fixed it

rmistry:
> Pull again. I've fixed it

Done.

jonathanmetzman (approved):
lgtm

PR #3463, opened 2023-10-26
[Progression] Move update_build_metadata to postprocess

alhijazi:


src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -450,16 +459,23 @@ def find_fixed_range(uworker_input):
...
   result, error = _testcase_reproduces_in_revision(
       testcase,
       testcase_file_path,
       job_type,
       max_revision,
+      build_data_list,
       update_metadata=True,
       clear_min_max_metadata=clear_min_max_metadata)
   if error is not None:
+    error.progression_task_output.build_data_list.extend(build_data_list)
```
jonathanmetzman, line 478: Does assigning to a list instead of extending break things?

alhijazi: yes, and I think the issue comes from here:
https://github.com/google/clusterfuzz/blob/4c9ca5e17c5d95f58bd761a1cf2204f8ceb8d3be/src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py#L438
Where `save_json_field` doesn't seem to be working for lists in the current state.

On the other hand, we can't assign lists of primitive types either but the error message is different.
Example for bad_revisions:
```
  File "/usr/local/google/home/ahijazi/alhijazi-clusterfuzz/clusterfuzz/src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/progression_task_test.py", line 177, in test_preprocess_uworker_output_custom_binary
    str(testcase.key.id()), 'job_type', None)
  File "/usr/local/google/home/ahijazi/alhijazi-clusterfuzz/clusterfuzz/src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py", line 384, in utask_preprocess
    progression_input.bad_revisions=build_manager.get_job_bad_revisions()
  File "/usr/local/google/home/ahijazi/alhijazi-clusterfuzz/clusterfuzz/src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py", line 352, in __setattr__
    setattr(self.proto, attribute, value)
AttributeError: Assignment not allowed to repeated field "bad_revisions" in protocol message object.
```

alhijazi: I'll take a look at how we serialize lists of objects tomorrow and see if a fix is needed somewhere.
It seems for the moment that we properly handle serialization of lists in UworkerInput but only for `ndb.Model` instances.
https://github.com/google/clusterfuzz/blob/4c9ca5e17c5d95f58bd761a1cf2204f8ceb8d3be/src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py#L456

jonathanmetzman:
This isn't ready to merge is it?

alhijazi:
> This isn't ready to merge is it?

I would say it is. Just a tiny modification to assign the list instead of extend when the serialisation fixes land.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -298,9 +308,7 @@ def _testcase_reproduces_in_revision(testcase,
         error_type=uworker_msg_pb2.ErrorType.PROGRESSION_BUILD_SETUP_ERROR)
 
   build_data = testcase_manager.check_for_bad_build(job_type, revision)
-  # TODO(https://github.com/google/clusterfuzz/issues/3008): Move this to
-  # postprocess.
-  testcase_manager.update_build_metadata(job_type, revision, build_data)
+  progression_task_output.build_data_list.append(build_data.proto)
```
alhijazi, line 311: @letitz, @jonathanmetzman 
I unfortunately am not able to simply append build_data object here and have to pass the proto.
I think we should go ahead and use protos directly.  What do you think?

letitz: I generally support moving towards simpler protos with more explicit conversion code where needed.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -429,7 +436,7 @@ def find_fixed_range(uworker_input):
   min_revision = testcase.get_metadata('last_progression_min')
   max_revision = testcase.get_metadata('last_progression_max')
   progression_task_output = uworker_msg_pb2.ProgressionTaskOutput(
-      clear_min_max_metadata=False)
+      clear_min_max_metadata=False, build_data_list=[])
```
jonathanmetzman, line 439: Why set it to an empty list?

jonathanmetzman (approved):
LGTM

PR #3468, opened 2023-10-30
Refactor untrusted worker error handling

letitz:
This breaks an implicit dependency cycle between tasks and error handling code:

- tasks defined error handlers
- which were stored in a global map in `uworker_handle_errors`
- referenced by `uworker_handle_errors.handle()`
- which was called by tasks to handle errors

Because not all tasks could handle all errors, a subset of error types to handle was passed to `uworker_handle_errors.handle()` along with the error itself. This implicitly depended on those errors being mapped to the right handlers in `uworker_handle_errors`.

Instead, this new approach has tasks build their own maps of handlers (called `CompositeErrorHandler` here) out of building blocks provided by `uworker_handle_errors` and other dependencies of theirs (currently, only `setup`). Dependencies like `setup` can provide their own error handlers for composition by consumers.

Hopefully this makes sense :)

A benefit of this approach is that since the error -> handler mapping is defined per task module, error types can be reused across tasks. For example, both progression task and regression task fail when the build manager fails to find any good revisions to use. #3469 introduces a new `REGRESSION_REVISION_LIST_ERROR` error that mirrors `PROGRESSION_REVISION_LIST_ERROR`, just because the keys need to be distinct in `uworker_handle_error`'s global map. Were this refactor merged, we could define a generic `EMPTY_REVISION_LIST` error type that both progression task and regression task could use and map to their respective error handlers.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
@@ -1968,12 +1968,12 @@ def _make_session(uworker_input):
 
 
 _ERROR_HANDLER = uworker_handle_errors.CompositeErrorHandler({
-    uworker_msg_pb2.ErrorType.FUZZ_NO_FUZZER:
-        handle_fuzz_no_fuzzer,
     uworker_msg_pb2.ErrorType.FUZZ_BUILD_SETUP_FAILURE:
         handle_fuzz_build_setup_failure,
     uworker_msg_pb2.ErrorType.FUZZ_DATA_BUNDLE_SETUP_FAILURE:
         handle_fuzz_data_bundle_setup_failure,
+    uworker_msg_pb2.ErrorType.FUZZ_NO_FUZZER:
+        handle_fuzz_no_fuzzer,
 }).compose_with(uworker_handle_errors.UNHANDLED_ERROR_HANDLER)
 
 
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
@@ -1967,12 +1967,17 @@ def _make_session(uworker_input):
   )
 
 
-HANDLED_ERRORS = [
-    uworker_msg_pb2.ErrorType.UNHANDLED,
-    uworker_msg_pb2.ErrorType.FUZZ_BUILD_SETUP_FAILURE,
-    uworker_msg_pb2.ErrorType.FUZZ_DATA_BUNDLE_SETUP_FAILURE,
-    uworker_msg_pb2.ErrorType.FUZZ_NO_FUZZER,
-]
+_ERROR_HANDLER = uworker_handle_errors.CompositeErrorHandler.compose(
+    uworker_handle_errors.CompositeErrorHandler({
+      uworker_msg_pb2.ErrorType.FUZZ_NO_FUZZER:
+          handle_fuzz_no_fuzzer,
+      uworker_msg_pb2.ErrorType.FUZZ_BUILD_SETUP_FAILURE:
+          handle_fuzz_build_setup_failure,
+      uworker_msg_pb2.ErrorType.FUZZ_DATA_BUNDLE_SETUP_FAILURE:
+          handle_fuzz_data_bundle_setup_failure,
+    }),
+    uworker_handle_errors.UNHANDLED_ERROR_HANDLER,
+)
 
 
 def utask_preprocess(fuzzer_name, job_type, uworker_env):
@@ -1998,7 +2003,8 @@ def utask_preprocess(fuzzer_name, job_type, uworker_env):
 
 def utask_postprocess(output):
   if output.error_type != uworker_msg_pb2.ErrorType.NO_ERROR:
-    uworker_handle_errors.handle(output, HANDLED_ERRORS)
+    _ERROR_HANDLER.handle(output)
     return
+
   session = _make_session(output.uworker_input)
   session.postprocess(output)
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
@@ -1967,17 +1967,14 @@ def _make_session(uworker_input):
   )
 
 
-_ERROR_HANDLER = uworker_handle_errors.CompositeErrorHandler.compose(
-    uworker_handle_errors.CompositeErrorHandler({
-        uworker_msg_pb2.ErrorType.FUZZ_NO_FUZZER:
-            handle_fuzz_no_fuzzer,
-        uworker_msg_pb2.ErrorType.FUZZ_BUILD_SETUP_FAILURE:
-            handle_fuzz_build_setup_failure,
-        uworker_msg_pb2.ErrorType.FUZZ_DATA_BUNDLE_SETUP_FAILURE:
-            handle_fuzz_data_bundle_setup_failure,
-    }),
-    uworker_handle_errors.UNHANDLED_ERROR_HANDLER,
-)
+_ERROR_HANDLER = uworker_handle_errors.CompositeErrorHandler({
+    uworker_msg_pb2.ErrorType.FUZZ_NO_FUZZER:
+        handle_fuzz_no_fuzzer,
+    uworker_msg_pb2.ErrorType.FUZZ_BUILD_SETUP_FAILURE:
+        handle_fuzz_build_setup_failure,
+    uworker_msg_pb2.ErrorType.FUZZ_DATA_BUNDLE_SETUP_FAILURE:
+        handle_fuzz_data_bundle_setup_failure,
+}).compose_with(uworker_handle_errors.UNHANDLED_ERROR_HANDLER)
 
 
 def utask_preprocess(fuzzer_name, job_type, uworker_env):
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
@@ -1969,12 +1969,12 @@ def _make_session(uworker_input):
 
 _ERROR_HANDLER = uworker_handle_errors.CompositeErrorHandler.compose(
     uworker_handle_errors.CompositeErrorHandler({
-      uworker_msg_pb2.ErrorType.FUZZ_NO_FUZZER:
-          handle_fuzz_no_fuzzer,
-      uworker_msg_pb2.ErrorType.FUZZ_BUILD_SETUP_FAILURE:
-          handle_fuzz_build_setup_failure,
-      uworker_msg_pb2.ErrorType.FUZZ_DATA_BUNDLE_SETUP_FAILURE:
-          handle_fuzz_data_bundle_setup_failure,
+        uworker_msg_pb2.ErrorType.FUZZ_NO_FUZZER:
+            handle_fuzz_no_fuzzer,
+        uworker_msg_pb2.ErrorType.FUZZ_BUILD_SETUP_FAILURE:
+            handle_fuzz_build_setup_failure,
+        uworker_msg_pb2.ErrorType.FUZZ_DATA_BUNDLE_SETUP_FAILURE:
+            handle_fuzz_data_bundle_setup_failure,
     }),
     uworker_handle_errors.UNHANDLED_ERROR_HANDLER,
 )
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
@@ -611,19 +611,6 @@ def utask_main(uworker_input: uworker_msg_pb2.Input):
   return uworker_msg_pb2.Output(minimize_task_output=minimize_task_output)
 
 
-HANDLED_ERRORS = [
-    uworker_msg_pb2.ErrorType.MINIMIZE_SETUP,
-    uworker_msg_pb2.ErrorType.TESTCASE_SETUP,
-    uworker_msg_pb2.ErrorType.MINIMIZE_UNREPRODUCIBLE_CRASH,
-    uworker_msg_pb2.ErrorType.MINIMIZE_CRASH_TOO_FLAKY,
-    uworker_msg_pb2.ErrorType.MINIMIZE_DEADLINE_EXCEEDED,
-    uworker_msg_pb2.ErrorType.MINIMIZE_DEADLINE_EXCEEDED_IN_MAIN_FILE_PHASE,
-    uworker_msg_pb2.ErrorType.LIBFUZZER_MINIMIZATION_UNREPRODUCIBLE,
-    uworker_msg_pb2.ErrorType.LIBFUZZER_MINIMIZATION_FAILED,
-    uworker_msg_pb2.ErrorType.UNHANDLED
-]
-
-
 def update_testcase(output: uworker_msg_pb2.Output):
   """Updates the tescase using the values passed from utask_main. This is done
   at the beginning of utask_postprocess and before error handling is called."""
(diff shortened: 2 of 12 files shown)


src/clusterfuzz/_internal/bot/tasks/utasks/uworker_handle_errors.py
```diff
@@ -12,73 +12,85 @@
...
 """Module for handling errors in utasks."""
-from clusterfuzz._internal.bot.tasks import setup
-from clusterfuzz._internal.bot.tasks.utasks import analyze_task
-from clusterfuzz._internal.bot.tasks.utasks import fuzz_task
-from clusterfuzz._internal.bot.tasks.utasks import minimize_task
-from clusterfuzz._internal.bot.tasks.utasks import progression_task
-from clusterfuzz._internal.bot.tasks.utasks import variant_task
+
+from typing import Callable
+from typing import Dict
```
jonathanmetzman, line 17: I think we prefer importing thme on the same line, though not sure.

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_handle_errors.py
```diff
@@ -12,73 +12,85 @@
...
 
+# A handler takes an output object with its `error_type` set, and does whatever
+# it wants with it.
+ErrorHandler = Callable[[uworker_io.UworkerOutput], None]
+
+# Must use string type because of protobuf enum shenanigans.
+HandlerDict = Dict["uworker_msg_pb2.ErrorType", ErrorHandler]
+
+
+class CompositeErrorHandler(object):
```
jonathanmetzman, line 30: (object) is not necessary in Python3 iirc. I think it's implicit

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_handle_errors.py
```diff
@@ -12,73 +12,85 @@
...
+
+from clusterfuzz._internal.bot.tasks.utasks import uworker_io
 from clusterfuzz._internal.protos import uworker_msg_pb2
 
+# A handler takes an output object with its `error_type` set, and does whatever
+# it wants with it.
+ErrorHandler = Callable[[uworker_io.UworkerOutput], None]
+
+# Must use string type because of protobuf enum shenanigans.
+HandlerDict = Dict["uworker_msg_pb2.ErrorType", ErrorHandler]
```
jonathanmetzman, line 27: single quotes.

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_handle_errors.py
```diff
@@ -12,73 +12,85 @@
...
+    Raises:
+      ValueError: if more than one handler handles the same error type.
+    """
+    handlers = {}
+
+    # Merge all handler dicts, checking that no keys are duplicated.
+    for composite_handler in args:
+      for error_type, handler in composite_handler.handlers().items():
+        if error_type in handlers:
+          raise ValueError(f"Duplicate handlers for error type {error_type}")
```
jonathanmetzman, line 61: single quotes

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_handle_errors.py
```diff
@@ -12,73 +12,85 @@
...
+
+  def handle(self, output: uworker_io.UworkerOutput):
+    """Handles the given `output`, delegating to underlying handlers.
+
+    Raises:
+      RuntimeError: if `output.error_type` is not handled by this instance.
+    """
+    handler = self._handlers.get(output.error_type)
+    if handler is None:
+      raise RuntimeError(f"Cannot handle error type {output.error_type}")
```
jonathanmetzman, line 83: single quote

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_handle_errors.py
```diff
@@ -12,73 +12,85 @@
...
+    For example:
+
+      CompositeErrorHandler({FOO: handle_foo, BAR: handle_bar})
+
+    Will delegate handling `FOO` errors to `handle_foo`, and `BAR` errors
+    to `handle_bar`.
+    """
+    self._handlers = handlers
+
+  # Mut use string types because `CompositeErrorHandler` is not defined yet.
```
jonathanmetzman, line 45: Must

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_handle_errors.py
```diff
@@ -12,73 +12,85 @@
...
+  @classmethod
+  def compose(cls, *args: "CompositeErrorHandler") -> "CompositeErrorHandler":
```
jonathanmetzman, line 47: single quote

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_handle_errors.py
```diff
@@ -12,73 +12,85 @@
...
+    for composite_handler in args:
+      for error_type, handler in composite_handler.handlers().items():
+        if error_type in handlers:
+          raise ValueError(f"Duplicate handlers for error type {error_type}")
+
+        handlers[error_type] = handler
+
+    return cls(handlers)
+
+  def handlers(self) -> HandlerDict:
```
jonathanmetzman, line 67: should this be a property?
Why a copy? I guess it's safer to prevent callers from mucking with it. I'm generally not this defensive.

src/clusterfuzz/_internal/bot/tasks/symbolize_task.py
```diff
@@ -55,9 +54,9 @@ def execute_task(testcase_id, job_type):
   if error:
     # TODO(metzman): Fix all this, symbolize task needs to become a utask.
     # Because this is trusted, we can trust the error.
-    all_errors = uworker_handle_errors.get_all_handled_errors()
+    assert setup.ERROR_HANDLER.is_handled(error)
```
jonathanmetzman, line 57: Why not let the valueerror happen instead of having an assert failure?

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -594,16 +594,26 @@ def utask_main(uworker_input):
...
-    uworker_msg_pb2.ErrorType.PROGRESSION_NO_CRASH,
-    uworker_msg_pb2.ErrorType.PROGRESSION_BUILD_SETUP_ERROR,
-    uworker_msg_pb2.ErrorType.PROGRESSION_TIMEOUT,
-    uworker_msg_pb2.ErrorType.PROGRESSION_BAD_BUILD,
-    uworker_msg_pb2.ErrorType.PROGRESSION_REVISION_LIST_ERROR,
-    uworker_msg_pb2.ErrorType.PROGRESSION_BUILD_NOT_FOUND,
-    uworker_msg_pb2.ErrorType.PROGRESSION_BAD_STATE_MIN_MAX,
-    uworker_msg_pb2.ErrorType.TESTCASE_SETUP,
-]
+_ERROR_HANDLER = uworker_handle_errors.CompositeErrorHandler.compose(
```
jonathanmetzman, line 597: This API will seem intimidating to other users. heh I wonder if operator overloading will help here (just say no)

jonathanmetzman (commented):
Looks pretty good on first pass.
@oliverchang What do you think about this?

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_handle_errors.py
letitz: The formatter puts them on their own lines, and the linter complains if I manually revert that.

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_handle_errors.py
letitz: Right, python2 habits. Fixed.

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_handle_errors.py
letitz: Fixed.

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_handle_errors.py
letitz: Good catch, fixed.

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_handle_errors.py
letitz: Good point! Turns out the only use of this is in `compose()`, which can just reach into `_handlers`. Removed entirely.

src/clusterfuzz/_internal/bot/tasks/symbolize_task.py
letitz: Good point, and also this is obsolete since the original code has been deleted.

src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
```diff
@@ -626,6 +607,26 @@ def handle_minimize_setup_error(output):
...
+def utask_postprocess(output):
+  """Postprocess in a trusted bot."""
+  if output.error_type is not None:
+    _ERROR_HANDLER.handle(output)
+    return
+
+  finalize_testcase(
+      output.uworker_input.testcase_id,
+      output.minimize_task_output.last_crash_result_dict,
+      flaky_stack=output.minimize_task_output.flaky_stack)
```
letitz, line 627: I don't understand where this came from. Why is my PR suddenly adding it?

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_handle_errors.py
letitz: Fixed.

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_handle_errors.py
letitz: Fixed.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
letitz: How about this API shape? Looks better to me - less nesting.

jonathanmetzman:
@letitz Sorry, major merge request after Ali's PR. I don't know if I can push to your repo so I can't fix this myself. I think it's straightforward to fix though.

letitz:
Rebased once more, and found the source of the previous test error. Can you please review @jonathanmetzman? Would love to finally land this before more merge conflicts make their way in.

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -1967,12 +1967,14 @@ def _make_session(uworker_input):
...
-    uworker_msg_pb2.ErrorType.FUZZ_NO_FUZZER,
-]
+_ERROR_HANDLER = uworker_handle_errors.CompositeErrorHandler({
+    uworker_msg_pb2.ErrorType.FUZZ_NO_FUZZER:
+        handle_fuzz_no_fuzzer,
+    uworker_msg_pb2.ErrorType.FUZZ_BUILD_SETUP_FAILURE:
+        handle_fuzz_build_setup_failure,
+    uworker_msg_pb2.ErrorType.FUZZ_DATA_BUNDLE_SETUP_FAILURE:
+        handle_fuzz_data_bundle_setup_failure,
+}).compose_with(uworker_handle_errors.UNHANDLED_ERROR_HANDLER)
```
jonathanmetzman, line 1977: Do you think UNHANDLED should be special cases so this doesn't need to be repeated everywhere?

src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
```diff
@@ -956,22 +1002,6 @@ def check_deadline_exceeded_and_store_partial_minimized_testcase(
   return time.time() > deadline
 
 
-def handle_minimize_deadline_exceeded(output: uworker_msg_pb2.Output):
```
jonathanmetzman, line 959: Are these functions just moved and not otherwise changed?

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_handle_errors.py
```diff
@@ -12,100 +12,79 @@
...
+    For example:
+
+      CompositeErrorHandler({FOO: handle_foo, BAR: handle_bar})
+
+    Will delegate handling `FOO` errors to `handle_foo`, and `BAR` errors
+    to `handle_bar`.
+    """
+    self._handlers = handlers
+
+  # Must use string types because `CompositeErrorHandler` is not defined yet.
```
jonathanmetzman, line 44: I didn't know how to handle this before. Cool!

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -599,16 +599,25 @@ def utask_main(uworker_input):
...
-    uworker_msg_pb2.ErrorType.PROGRESSION_BUILD_SETUP_ERROR,
-    uworker_msg_pb2.ErrorType.PROGRESSION_TIMEOUT,
-    uworker_msg_pb2.ErrorType.PROGRESSION_BAD_BUILD,
-    uworker_msg_pb2.ErrorType.PROGRESSION_REVISION_LIST_ERROR,
-    uworker_msg_pb2.ErrorType.PROGRESSION_BUILD_NOT_FOUND,
-    uworker_msg_pb2.ErrorType.PROGRESSION_BAD_STATE_MIN_MAX,
-    uworker_msg_pb2.ErrorType.TESTCASE_SETUP,
-]
+_ERROR_HANDLER = uworker_handle_errors.CompositeErrorHandler({
+    uworker_msg_pb2.ErrorType.PROGRESSION_REVISION_LIST_ERROR:
```
jonathanmetzman, line 603: nit: Do you think it would be better to alphabetize the handlers?

src/clusterfuzz/_internal/bot/tasks/utasks/regression_task.py
```diff
@@ -255,21 +255,6 @@ def validate_regression_range(
   return None
 
 
-def handle_low_confidence_in_regression_range(output: uworker_msg_pb2.Output):
```
jonathanmetzman, line 258: Can you confirm these functions are just moved and not modified?

jonathanmetzman (commented):
Looks good to me. 
There's one more conflict sorry. I'll prevent any more conflicts, and merge this when you get it ready again.

src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
letitz: Yes! I just needed to define them above `_ERROR_HANDLER`.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
letitz: Sure, done.

src/clusterfuzz/_internal/bot/tasks/utasks/regression_task.py
letitz: Yes, I confirm! I reordered stuff so all the error handlers would be together.

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_handle_errors.py
letitz: Yup!

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
letitz: It's not used in `regression_task.py`. Should it?

In any case, if all error handlers should do this by default, then happy to add this to the `CompositeErrorHandler` class. In the interest of landing this quickly, I can do that in a follow-up PR?

letitz:
Thanks for the review and for holding up other PRs. I think this is ready now - as long as tests agree :)

PR #3469, opened 2023-10-30
[Rearch] Migrate regression task #2

letitz:
Handle empty revision list errors in `utask_postprocess` instead of `utask_main`.
Introduce a new error type to represent this error.

As I wrote tests, I ran into serialization errors that confused me until I understood that my mental model of types, and the corresponding type annotations I had written, were wrong. Turns out that the arguments to `utask_main` and `utask_postprocess` are not `UworkerInput` and `UworkerOutput` as I has thought, but instead generic `DeserializedUworkerMsg` objects. I find this unfortunate, but I guess the input as created by `utask_preprocess` should not be the same type as the input passed to `utask_main` due to the change detection logic required for `utask_main`'s argument?


jonathanmetzman:
> Handle empty revision list errors in `utask_postprocess` instead of `utask_main`. Introduce a new error type to represent this error.
> 
> As I wrote tests, I ran into serialization errors that confused me until I understood that my mental model of types, and the corresponding type annotations I had written, were wrong. Turns out that the arguments to `utask_main` and `utask_postprocess` are not `UworkerInput` and `UworkerOutput` as I has thought, but instead generic `DeserializedUworkerMsg` objects. I find this unfortunate, but I guess the input as created by `utask_preprocess` should not be the same type as the input passed to `utask_main` due to the change detection logic required for `utask_main`'s argument?

They can't be the protos, because then the attributes would be unserialized and thus weird for users to work with.

jonathanmetzman:
> > Handle empty revision list errors in `utask_postprocess` instead of `utask_main`. Introduce a new error type to represent this error.
> > As I wrote tests, I ran into serialization errors that confused me until I understood that my mental model of types, and the corresponding type annotations I had written, were wrong. Turns out that the arguments to `utask_main` and `utask_postprocess` are not `UworkerInput` and `UworkerOutput` as I has thought, but instead generic `DeserializedUworkerMsg` objects. I find this unfortunate, but I guess the input as created by `utask_preprocess` should not be the same type as the input passed to `utask_main` due to the change detection logic required for `utask_main`'s argument?
> 
> They can't be the protos, because then the attributes would be unserialized and thus weird for users to work with.

I guess they could be if we make deserialization happen lazily. What do you think of this idea?

letitz:
> They can't be the protos, because then the attributes would be unserialized and thus weird for users to work with.

Do you mean that some attributes would be serialized, in particular `Json` message contents?

In general, I'm curious to understand what makes working directly with protos difficult. To be clear, this is a common problem that does not surprise me, but I would just like to clarify the problem before discussing solutions.

AIUI, the pain points of working directly with uworker i/o proto messages are:

1. `Json` messages store their contents serialized
2. `Entity` messages are much less ergonomic than the corresponding `ndb.Model` instances
3. Some `Entity` messages need wrapping for change detection in `utask_main()`
4. Setting nonexistent fields does not throw an error
5. Any others?

If this is the case, then I think we could trade some of the descriptor and attr cleverness for a bit of code duplication, by defining python classes for proto messages containing problematic fields (either `Json` or `Entity`, if the above is correct).

Something like:

```py
# message SetupInput {
#   optional google.datastore.v1.Entity fuzzer = 1;
#   optional string fuzzer_name = 2;
#   repeated google.datastore.v1.Entity data_bundles = 3;
#   optional string fuzzer_log_upload_url = 4;
#   optional string fuzzer_download_url = 5;
#   optional string testcase_download_url = 6;
# }

@dataclass
class SetupInput(object):
  fuzzer: Optional[data_types.Fuzzer] = None
  fuzzer_name: Optional[str] = None
  data_bundles: List[data_types.DataBundle] = []
  fuzzer_log_upload_url: Optional[str] = None
  fuzzer_download_url: Optional[str] = None
  testcase_download_url: Optional[str] = None
```

Conversion to/from protos could be implemented by hand, or with a bit (a lot?) of cleverness I think it can be automated using dataclass field introspection and/or descriptors:

```py
def copy_to_proto(value, message):
   class_fields = dataclass.fields(value)
   # Iterate over fields in `value` using `class_fields`:
   # - Convert subclasses of `ndb.Model` to `Entity` protobufs
   # - Convert dicts to `Json` protobufs
   # - Set the right field in `result`
   return message

def copy_from_proto(message, value):
  class_fields = dataclass.fields(value)
  # Iterate over fields in `message` using `class_fields`:
  # - Convert `Entity` protobufs to the expected `ndb.Model` subclass
  # - Convert `Json` protobufs to dicts
  # - Set the right field in `value`
  return value
```

I'm not sure how best to define the mapping between dataclass and protobuf, and how to generate the best API on dataclasses. I would like to be able to write the following:

```
setup_input = SetupInput.from_proto(setup_input_proto)
setup_input_proto2 = setup_input.to_proto()
```

Options:

- (similar to existing) base class, with a class variable on `SetupInput` defining the corresponding proto type
- class decorator
- metaclass???

Bonus points if it plays well with typing.

WDYT?

PR #3470, opened 2023-10-30
Make symbolize a utask

jonathanmetzman:


jonathanmetzman:
FYI, symbolize needs to be untrusted since it runs testcases.

oliverchang (approved):

PR #3472, opened 2023-10-30
Improve uworker_io serialization.

alhijazi:
All subfields of UworkerOutput now extend it and use the same save_rich_type method.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 """Module for dealing with input and output (I/O) to a uworker."""
 
-import collections
 import json
 import uuid
 
@@ -21,6 +20,7 @@ from google.cloud import ndb
 from google.cloud.datastore_v1.proto import entity_pb2
 from google.cloud.ndb import model
 from google.protobuf import message
+from google.protobuf.descriptor import FieldDescriptor
 
 from clusterfuzz._internal.datastore import data_types
 from clusterfuzz._internal.google_cloud_utils import storage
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
```diff
@@ -349,6 +349,11 @@ def __setattr__(self, attribute, value):
 
     field_descriptor = self.proto.DESCRIPTOR.fields_by_name[attribute]
     if field_descriptor.message_type is None:
+      # Handle the repeated fields of non-composite types.
+      if field_descriptor.label == FieldDescriptor.LABEL_REPEATED:
+        field = getattr(self.proto, attribute)
+        field.extend(value)
```
letitz, line 355: This does not clear the previous field value before extending, which I think will lead to weird behavior:

```py
>>> output.repeated_field = [1, 2]
>>> output.repeated_field = [3, 4]
>>> output.repeated_field
[1, 2, 3, 4]
```

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
```diff
@@ -428,18 +404,24 @@ def save_rich_type(self, attribute, value):
     field.CopyFrom(wrapped_entity_proto)
 
 
-class BuildData(UworkerMsg):
-  """Class representing an unserialized ProgressionTaskOutput message from
+class FuzzTaskOutput(UworkerOutput):
```
letitz, line 407: It's weird to me that fields of uworker output would subclass `UworkerOutput`. The inheritance "is a" relationship does not hold: a `BuildData` object *is not a* `UworkerOutput` object as well. `UworkerMsg` seems to provide the base functionality we would like to reuse here? If not, could we extend it to provide the right functionality?

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
jonathanmetzman: +1

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
jonathanmetzman: +1

alhijazi:
I think this is no longer needed and we should use protos directly.

PR #3475, opened 2023-10-31
Log changes to environment variables.

letitz:
This should help debug ClusterFuzz issues in production.

### DIFF
--- a/src/clusterfuzz/_internal/system/environment.py
+++ b/src/clusterfuzz/_internal/system/environment.py
@@ -24,6 +24,7 @@ import sys
 import yaml
 
 from clusterfuzz._internal import fuzzing
+from clusterfuzz._internal.metrics import logs
 
 # Tools supporting customization of options via ADDITIONAL_{TOOL_NAME}_OPTIONS.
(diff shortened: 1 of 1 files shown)


letitz:
This is a bit of a silly idea, but I thought I would throw it out for consideration. It might result in a crazy amount of log spam, though we could downgrade the log level to `DEBUG` to help with filtering (not sure how that interacts with GCP logging).

Another approach I could see would be to bundle up the current set of environment variables in the log `extras`.

WDYT?

jonathanmetzman:
Both of these seem very spammy to be honest.

letitz:
Would this be more palatable as `DEBUG` logs?

jonathanmetzman:
OK so looking at this particular case. That build 88437 is an empty zip. So maybe we could add logging in setup_regular_build or elsewhere that 1. logs when the build is missing the APP_NAME. 2. Marks the build as a bad build.

letitz:
I've got 1) covered by #3476. I can send another PR to have the setup code return a uworker error in these cases, and then mark the build as bad in that error's handler.

PR #3484, opened 2023-11-02
Fix serialization.

jonathanmetzman:
1. Don't use UworkerMsg's underlying proto as the source of truth before
serializing. To type check, we copy any messages immediately to the proto.
This can lead to errors as the UworkerMsg can still be used and assigned
to, leaving proto's representation of the UworkerMsg stale. To deal with
this: Re-set all values on proto fields before serializing (and do the same
to any submessages so that we properly handle (great) grandchildren).
2. Handle initializing with list values.
3. Handle reassigning list values properly by clearing any existing items and then extending.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
@@ -265,6 +265,10 @@ def deserialize_uworker_output(uworker_output_str):
   return uworker_output
 
 
+def to_json_msg(value):
+  return uworker_msg_pb2.Json(serialized=json.dumps(value))
+
+
 class UworkerEntityWrapper:
   """Wrapper for db entities on the uworker. This wrapper functions the same as
   the entity but also tracks changes made to the entity. This is useful for
@@ -322,8 +326,9 @@ class UworkerEntityWrapper:
 
 
 class UworkerMsg:
-  """Convenience class for results utask_function. This is useful for ensuring
-  we are returning values for fields expected by main or postprocess."""
+  """Convenience class for results from utask functions. This is useful for
+  ensuring we are returning values for fields expected by main or
+  postprocess."""
 
   # Child must implement.
   PROTO_CLS = None
@@ -337,9 +342,6 @@ class UworkerMsg:
     assert self.PROTO_CLS is not None
 
   def __getattr__(self, attribute):
-    if attribute in ['_proto', '_set_fields']:
-      # Allow setting and changing _proto. Stack overflow in __init__ otherwise.
-      return super().__getattr__(attribute)  # pylint: disable=no-member
     return getattr(self._proto, attribute)
 
   def __setattr__(self, attribute, value):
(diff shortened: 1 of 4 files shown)


src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
```diff
@@ -361,7 +365,18 @@ def save_rich_type(self, attribute, value):
     raise NotImplementedError('Child must implement.')
 
   def serialize(self):
-    return self.proto.SerializeToString()
+    """Serializes UworkerMsg into proto wire format."""
+    for attr in self._set_fields:
+      # Do this to make sure _proto fields are up to date.
```
oliverchang, line 370: Can you explain how this can get out of sync? I'm wondering if this wrapper approach should be replaced by a dumber approach of just modelling the input/output classes in Python and serializing them to proto only when we need to send through through the wire. 

There's a lot of black magic happening here.

jonathanmetzman: I can leave a TODO. Titouan is exploring a new approach.

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
```diff
@@ -448,23 +463,29 @@ class UworkerInput(UworkerMsg):
...
+      del field[:]
       value = list(value)
       if len(value) == 0:
         return
-      assert isinstance(value[0], ndb.Model), value[0]
-      field.extend([model._entity_to_protobuf(entity) for entity in value])  # pylint: disable=protected-access
+      if not isinstance(value[0], ndb.Model):
+        field.extend(value)
+      else:
+        # TODO(metzman): This is probably unnecessary and very complext to deal
```
letitz, line 481: Do we hit this case in practice? If not, can we just `raise UnimplementedError()`?

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
```diff
@@ -448,23 +463,29 @@ class UworkerInput(UworkerMsg):
...
+        # with.
+        new_value = [model._entity_to_protobuf(entity) for entity in value]  # pylint: disable=protected-access
+        field.extend(new_value)
       return
 
     if isinstance(value, UworkerMsg):
```
letitz, line 487: Again, can we handle this in `UworkerMsg` itself?

src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/uworker_io_test.py
```diff
@@ -169,7 +169,7 @@ def test_error_and_testcase_behavior(self):
     error_value = 1
     self.output.error_type = error_value
     self.assertEqual(self.output.error_type, error_value)
-    self.assertEqual(self.output.proto.error_type, error_value)
+    self.assertEqual(self.output._proto.error_type, error_value)  # pylint: disable=protected-member
```
letitz, line 172: You could instead:

- just add a `proto()` method to `UworkerMsg`, or even a `proto_copy()` if you really want to avoid exposing internal state
- roundtrip `self.output` and check that `error_type` is still good

src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/uworker_io_test.py
```diff
@@ -439,3 +443,45 @@ def test_submessage_serialization_and_deserialization(self):
...
+    analyze_task_input.bad_revisions = [1]
+    uworker_input = uworker_io.UworkerInput(
+        analyze_task_input=analyze_task_input)
+    wire_format = uworker_io.serialize_uworker_input(uworker_input)
+    deserialized = uworker_io.deserialize_uworker_input(wire_format)
+    self.assertEqual(deserialized.analyze_task_input.bad_revisions, [1])
+
+  def test_submessage_references(self):
+    """Tests that updating a submessage works both when directly reading from
+    uworker_input and from reading from it once it has been serialized and
```
letitz, line 473:
```suggestion
    uworker_input and when reading from it once it has been serialized and
```

jonathanmetzman:
We abandoned this approach in favor of using protos directly.

PR #3485, opened 2023-11-03
Allow Python 3.11 to use bulter

Alan32Liu:
Needed by my workstation.

### DIFF
--- a/src/local/butler/constants.py
+++ b/src/local/butler/constants.py
@@ -51,7 +51,7 @@ elif sys.version_info.major == 3 and sys.version_info.minor == 9:
 elif sys.version_info.major == 3 and sys.version_info.minor == 10:
   ABIS = {'linux': 'cp310', 'windows': 'cp310', 'macos': 'cp310'}
 elif sys.version_info.major == 3 and sys.version_info.minor == 11:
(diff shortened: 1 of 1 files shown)


src/local/butler/constants.py
```diff
@@ -50,6 +50,8 @@
   ABIS = {'linux': 'cp39', 'windows': 'cp39', 'macos': 'cp39'}
 elif sys.version_info.major == 3 and sys.version_info.minor == 10:
   ABIS = {'linux': 'cp310', 'windows': 'cp310', 'macos': 'cp310'}
+elif sys.version_info.major == 3 and sys.version_info.minor == 11:
+  ABIS = {'linux': 'cp310', 'windows': 'cp310', 'macos': 'cp310'}
```
oliverchang, line 54: These don't need to be cp311?

Alan32Liu: 310 works fine on my computer.
Changing to 311 now (which also works).

jonathanmetzman: Is it possible, this "worked" because the ABIs are only used during deploy (so really cp310 would fail if actually used with 3.11)

Alan32Liu: Yes, it could be. I did not deploy yet.

oliverchang (approved):

PR #3488, opened 2023-11-03
Added Fuzzer-crash-state crash_category

davfu1:
The inclusion of frames arising from fuzzing directories in crash_state has been a good heuristic in Android for 'No Security Impact' bugs caused by poor fuzzer implementation or fuzzer error. Examples:
  
  https://clusterfuzz.corp.google.com/testcase?key=6233947671429120
  https://clusterfuzz.corp.google.com/testcase-detail/5752617099984896
  https://clusterfuzz.corp.google.com/testcase-detail/5147091159875584

This change adds a label to CF-filed issues, the label indicating that a majority of the frames used for crash_state had arisen from a fuzzer directory.

This change is an alternative design to using STACK_FRAME_IGNORE_REGEXES to remove fuzzing stack frames from crash_state consideration altogether. Because crash_state is used as a deduplication key, changing crash_state construction would regenerate past testcases when the same fuzzer crashes are seen again, and allow an unknown number of previously-triaged bugs to bypass deduping and be refiled. @jonathanmetzman noticed that this regeneration could be potentially avoided with a script triggering progression tasks, although efforts would be needed to avoid a race condition between progression and fuzzing each generating new crash_states.

### DIFF
--- a/src/clusterfuzz/_internal/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
+++ b/src/clusterfuzz/_internal/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
@@ -419,7 +419,7 @@ class StackAnalyzerTestcase(unittest.TestCase):
     expected_state = 'boom_internal\nanother_boom\nboom\n'
     expected_stacktrace = data
     expected_security_flag = True
-    expected_categories = {'Fuzzer-exit'}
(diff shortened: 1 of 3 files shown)


davfu1:
Hi @oliverchang, I was hoping to gently bump this to your attention if you happen to get a chance to take a look.

oliverchang (approved):

PR #3496, opened 2023-11-07
[Progression] Remove the need to pass back the testcase thru uworker_…

alhijazi:
…output

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/task_types.py
+++ b/src/clusterfuzz/_internal/bot/tasks/task_types.py
@@ -79,6 +79,25 @@ class UTaskLocalExecutor(BaseTask):
     logs.log('Utask local: done.')
 
 
+class ProgressionUTaskLocalExecutor(BaseTask):
+  """Represents an untrusted progression task. Executes it entirely locally and
+  in memory."""
+
+  def execute(self, task_argument, job_type, uworker_env):
+    """Executes a utask locally in-memory."""
+    uworker_input = utasks.progression_tworker_preprocess_no_io(
+        self.module, task_argument, job_type, uworker_env)
+    if uworker_input is None:
+      return
+    uworker_output = utasks.progression_uworker_main_no_io(
+        self.module, uworker_input)
+    if uworker_output is None:
+      return
+    utasks.progression_tworker_postprocess_no_io(self.module, uworker_output,
+                                                 uworker_input)
+    logs.log('Utask local: done.')
+
+
 class UTaskLocalPreprocessAndMain(UTaskLocalExecutor):
   """Represents an untrusted task. Executes the preprocess and main parts on
   this machine and causes postprocess to be executed on on other machines."""
@@ -146,7 +165,7 @@ COMMAND_TYPES = {
     'fuzz': UTaskLocalExecutor,
     'impact': TrustedTask,
     'minimize': UTaskLocalExecutor,
-    'progression': UTaskLocalExecutor,
+    'progression': ProgressionUTaskLocalExecutor,
     'regression': UTaskLocalExecutor,
     'symbolize': UTaskLocalExecutor,
     'unpack': TrustedTask,
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
@@ -16,6 +16,8 @@
 import os
 import time
 
+from google.cloud.ndb import model
+
 from clusterfuzz._internal.base import bisection
 from clusterfuzz._internal.base import tasks
 from clusterfuzz._internal.base import utils
@@ -24,7 +26,6 @@ from clusterfuzz._internal.bot.fuzzers import engine_common
 from clusterfuzz._internal.bot.tasks import setup
 from clusterfuzz._internal.bot.tasks import task_creation
 from clusterfuzz._internal.bot.tasks.utasks import uworker_handle_errors
-from clusterfuzz._internal.bot.tasks.utasks import uworker_io
 from clusterfuzz._internal.build_management import build_manager
 from clusterfuzz._internal.build_management import revisions
 from clusterfuzz._internal.datastore import data_handler
@@ -63,7 +64,7 @@ def _save_current_fixed_range_indices(testcase, uworker_output):
       update_testcase=False)
 
 
-def handle_progression_timeout(uworker_output: uworker_io.UworkerOutput):
+def handle_progression_timeout(uworker_output: uworker_msg_pb2.Output):
   """Job has exceeded the deadline. Recreate the task to pick up where we left
   off."""
   testcase_id = uworker_output.uworker_input.testcase_id
@@ -75,8 +76,7 @@ def handle_progression_timeout(uworker_output: uworker_io.UworkerOutput):
   tasks.add_task('progression', testcase_id, job_type)
 
 
-def handle_progression_build_not_found(
-    uworker_output: uworker_io.UworkerOutput):
+def handle_progression_build_not_found(uworker_output: uworker_msg_pb2.Output):
   """Handles an expected build that no longer exists, we can't continue. Also,
   clears progression_pending testcase metadata"""
   testcase_id = uworker_output.uworker_input.testcase_id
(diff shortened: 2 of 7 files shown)


alhijazi:
Hello @jonathanmetzman, @oliverchang, @letitz
This is a POC on using protos directly instead of going thru uworker_io wrappers in progression_task.
The first commit removes the need to use `UworkerEntityWrapper` within the Output for progression_task only.
The second commit uses the protos directly in progression_task

In order to use this approach for all tasks, we need to first remove the `UworkerEntityWrapper` members of Output proto
If you agree on this approach, I can proceed and start removing the UworkerEntityWrappers.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -355,6 +353,43 @@ def _store_testcase_for_regression_testing(testcase, testcase_file_path):
                    regression_testcase_url)
 
 
+# If this approach of using the protos directly is adopted, these two function
+# would no longer be needed and we'd make setup.preprocess_setup_testcase
+# return a proto directly.
+def setup_input_to_proto(setup_input):
+  """A temporary workaround to limit the changes inside this file."""
+  setup_input_proto = uworker_msg_pb2.SetupInput()
```
letitz, line 361: Can we even just reach inside `SetupInput` and just return `setup_input._proto`?

src/clusterfuzz/_internal/protos/uworker_msg.proto
```diff
@@ -86,6 +86,7 @@ message Input {
   optional RegressionTaskInput regression_task_input = 17;
   optional SymbolizeTaskInput symbolize_task_input = 18;
   optional string module_name = 19;
+  map<string, string> progression_uworker_env = 20;
```
letitz, line 89: This would eventually replace `uworker_env`, right? The naming is just because this PR only touches `progression_task` so far?

letitz (commented):
Looks great to me! The code is not meaningfully harder to read (or write, it seems) and this allows us to remove a lot of complexity.

src/clusterfuzz/_internal/protos/uworker_msg.proto
alhijazi: Indeed. I also think that using protos directly would force us to actively try to simplify the types used and avoid using the user-defined JSON proto msg.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
alhijazi: I think you are right, I should try simply that!

src/clusterfuzz/_internal/protos/uworker_msg.proto
jonathanmetzman: Can you mention that in a comment please?

src/clusterfuzz/_internal/bot/tasks/task_types.py
```diff
@@ -146,7 +165,7 @@ def execute(self, task_argument, job_type, uworker_env):
     'fuzz': UTaskLocalExecutor,
     'impact': TrustedTask,
     'minimize': UTaskLocalExecutor,
-    'progression': UTaskLocalExecutor,
+    'progression': ProgressionUTaskLocalExecutor,
```
jonathanmetzman, line 168: Can even call this stuff V2 instead of Progression, since it will soon be expanded.

src/clusterfuzz/_internal/protos/uworker_msg.proto
jonathanmetzman: I think there's some code in postprocess or error handling that grabs values from uworker_input.uworker_env. We could change that to grab directly from the env (though this will make testing harder).

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -355,6 +353,27 @@ def _store_testcase_for_regression_testing(testcase, testcase_file_path):
                    regression_testcase_url)
 
 
+# If this approach of using the protos directly is adopted, this function
```
jonathanmetzman, line 356: Ouch this is a pain. Can you make this a TODO to get rid of this.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -371,26 +390,27 @@ def utask_preprocess(testcase_id, job_type, uworker_env):
...
   data_handler.update_testcase_comment(testcase, data_types.TaskState.STARTED)
-  progression_input = uworker_io.ProgressionTaskInput()
+  progression_input = uworker_msg_pb2.ProgressionTaskInput()
   progression_input.custom_binary = build_manager.is_custom_binary()
-  progression_input.bad_revisions.extend(build_manager.get_job_bad_revisions())
+  progression_input.bad_revisions.extend(build_manager.get_job_bad_revisions())  # pylint: disable=no-member
   # Setup testcase and its dependencies.
   setup_input = setup.preprocess_setup_testcase(testcase)
-  return uworker_io.UworkerInput(
+  testcase_proto = model._entity_to_protobuf(testcase)  # pylint: disable=protected-access
```
jonathanmetzman, line 398: Are we going to stop using Entities in preprocess/Input? Because if not, I think this is starting to become ugly without my abstractions.

jonathanmetzman (commented):
I think this is an important change, but I'm not so sure I like it. Let's address some of the comments and see what it looks like.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
alhijazi: No, we will continue using entities in preprocess/input, we will have to explicitly call `_entity_to_protobuf`. At least this way, what we're doing will be explicit, instead of some magic behind the scenes. 
For what it's worth, I don't think this will pollute the code a lot.

What I suggest is we put the second commit of this PR aside until I remove the entities, then I can create a PR for the migration of all tasks to use protos and it will be clear if this is a bad idea.
@jonathanmetzman  WDYT?

alhijazi:
Converting this to draft while we land the pre-requisite changes to remove the UworkerEntityWrappers from Output.

src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/progression_task_test.py
```diff
@@ -120,7 +121,10 @@ def setUp(self):
     helpers.patch(
         self,
         ['clusterfuzz._internal.bot.tasks.setup.preprocess_setup_testcase'])
-    self.mock.preprocess_setup_testcase.return_value = uworker_io.SetupInput()
+    setup_input = uworker_io.SetupInput(fuzzer_name='fuzzer_name')
+    # For some reason, setup_input.fuzzer is not None as a result of the above statement alone.
```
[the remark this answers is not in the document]
letitz: go/proto2-versus-proto3#defaults (google-internal) states one cannot set default values on proto3 fields.

I agree this is one of the least ergonomic things about protobuf generated code in Python, but once you know about it you can just call `HasField("fuzzer")` instead?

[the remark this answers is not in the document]
jonathanmetzman: What do you think of accessing fields thru a helper function that handles this properly?
Sort of like using a wrapper, but a little more explicit at the cost of ergnomics.
I think this behavior is pretty error prone.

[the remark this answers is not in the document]
letitz: What do you have in mind? I don't see an option that would make it much more ergonomic:

```py
fuzzer_or_none = get_optional_field(setup_input, "fuzzer")
if fuzzer_or_none is None:
  frobnicate()
else:
  reticulate_splines(fuzzer_or_none)
```

does not strike me as much better than:

```py
if setup_input.HasField("fuzzer"):
  reticulate_splines(setup_input.fuzzer)
else:
  frobnicate()
```

ISTM errors of this kind can be caught well-enough with a combination of:

 - code review and general knowledge of proto generated code APIs
 - unit tests

I'll also plug that type checking would flag `setup_input.fuzzer is None` as a type mismatch :)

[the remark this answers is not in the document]
alhijazi: FYI, I've been working on migrating the other tasks to only use protos also today. I should have something ready early next week. I need to merge my PRs removing the need to pass back entities from utask_main before.
Thanks,

alhijazi:
Closing this as it is no longer needed.

PR #3499, opened 2023-11-07
Add Documentation for Local Development

micahbales:
This commit adds some additional details about Python installation, including a link to a third-party site that we found helpful in getting started.

It also adds guidance for dealing with a number of dependency issues that developers may encounter as they are getting set up.

Finally, it adds a couple of "tips" for ClusterFuzz development. This last little bit felt like it might be its own section (rather than a preqrequisite), but perhaps it's still too small to break out a new section for it. If it got longer, we might want to do that.

### DIFF
--- a/docs/getting-started/prerequisites.md
+++ b/docs/getting-started/prerequisites.md
@@ -161,4 +161,4 @@ You might get stuck in a mode where the python version is the global value, rath
 ## ClusterFuzz Development Tips
 
 - Before committing, run `python butler.py format` to clean up formatting
-- Make sure you lint all updated files. Either with `python butler.py lint` or individually with `pylint --score=no --jobs=0 <name of directory>`
+- Make sure you lint all updated files with `python butler.py lint`. (It's also possible, though not recommended, to lint individual files with `pylint --score=no --jobs=0 <name of directory>`.)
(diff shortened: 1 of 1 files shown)


micahbales:
@oliverchang PTAL

docs/getting-started/prerequisites.md
```diff
@@ -85,21 +98,67 @@ Our supported systems include:
...
+```
+
+#### Refreshing the Python Environment
+
+You might get stuck in a mode where the python version is the global value, rather than the version you set for the pipenv environment - even when apparently leaving the virtual environment through the `exit`command. If this happens, you may still be in a virtual environment (especially if you see `(clusterfuzz)`before the command prompt). To escape this virtual environment, use the command `deactivate`, then run `python3.7 -m pipenv shell` again to start the pipenv shell with the correct python version.
+
+## ClusterFuzz Development Tips
+
+- Before committing, run `python butler.py format` to clean up formatting
+- Make sure you lint all updated files. Either with `python butler.py lint` or individually with `pylint --score=no --jobs=0 <name of directory>`
```
jonathanmetzman, line 164: Better to do `python butler.py lint` since linting all files is necessary when you change something like a function used in another module.

micahbales: Agreed that this is the preferred way. However, my experience was that I ran into problems running `python butler.py lint`in my local environment, and so it was useful to know a workaround.

I updated to make it clearer what the preferred strategy is here. If you'd prefer that I leave out this alternative entirely, I'll go ahead and remove it - but it would have been useful to me when I was getting started!

jonathanmetzman (approved):
lgtm

PR #3503, opened 2023-11-09
Fix extension_fields. It should operate on issue_types not self._data

rmistry:


rmistry:
PTAL @oliverchang 
FYI @micahbales

oliverchang (approved):

PR #3511, opened 2023-11-15
[Analyze] Don't pass the testcase back thru UworkerOutput

alhijazi:
Also stops passing testcase_upload_metadata back through UworkerOuput.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
@@ -437,6 +437,9 @@ HANDLED_ERRORS = [
 
 def _update_testcase(output):
   """Updates the testcase using the info passed from utask_main."""
+  if output.analyze_task_output is None:
+    return
+
   testcase = data_handler.get_testcase_by_id(output.uworker_input.testcase_id)
   analyze_task_output = output.analyze_task_output
   testcase.crash_revision = analyze_task_output.crash_revision
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
@@ -71,8 +71,8 @@ def _add_default_issue_metadata(testcase):
 
 
 def handle_analyze_no_revisions_list_error(output):
-  data_handler.update_testcase_comment(output.testcase,
-                                       data_types.TaskState.ERROR,
+  testcase = data_handler.get_testcase_by_id(output.uworker_input.testcase_id)
+  data_handler.update_testcase_comment(testcase, data_types.TaskState.ERROR,
                                        'Failed to fetch revision list')
   handle_build_setup_error(output)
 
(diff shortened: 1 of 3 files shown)


src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
```diff
@@ -406,29 +420,58 @@ def handle_build_setup_error(output):
...
   """Trusted: Cleans up after a uworker execute_task, writing anything needed to
   the db."""
+  _update_testcase(output)
   if output.error_type is not None:
     uworker_handle_errors.handle(output, HANDLED_ERRORS)
     return
-  testcase = output.testcase
-  testcase_upload_metadata = output.testcase_upload_metadata
+  testcase = data_handler.get_testcase_by_id(output.uworker_input.testcase_id)
+  testcase_upload_metadata = output.uworker_input.testcase_upload_metadata
```
oliverchang, line 474: Can we remove the `testcase_upload_metadata` field from the Output proto?

src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
```diff
@@ -406,29 +420,58 @@ def handle_build_setup_error(output):
...
+    uworker_msg_pb2.ErrorType.ANALYZE_NO_REVISION_INDEX,
     uworker_msg_pb2.ErrorType.UNHANDLED,
 ] + setup.HANDLED_ERRORS
 
 # pylint: enable=no-member
 
 
+def _update_testcase(output):
+  """Updates the testcase using the info passed from utask_main."""
+  testcase = data_handler.get_testcase_by_id(output.uworker_input.testcase_id)
```
oliverchang, line 440: I assume we can trust `uworker_input` here because we always make sure this is the original input we requested right?  @jonathanmetzman

src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
alhijazi: `testcase_upload_metadata` is no longer part of the output here. It is accessed through the `uworker_input` which is added to the output after `utask_main`

oliverchang: Right. Can we remove the proto field completely? I.e. this: https://github.com/google/clusterfuzz/blob/cefc990128d95fa570bf84bed2494fcbc48aba23/src/clusterfuzz/_internal/protos/uworker_msg.proto#L181

src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
jonathanmetzman: Yes. When uworker_output comes from an untrusted source, the trusted receiver derives the input file (which the untrusted party only has read access to) and uses that as input.

src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
jonathanmetzman: I think so.

src/clusterfuzz/_internal/protos/uworker_msg.proto
```diff
@@ -93,7 +93,18 @@ message SymbolizeTaskOutput {
 }
 
 message AnalyzeTaskOutput {
- // TODO(metzman): Fill this out.
+  optional int64 crash_revision = 1;
```
jonathanmetzman, line 96: Are we sure this is everything to save?

alhijazi: To create this list, I looked at all the testcase fields updated within utask_main. This is normally everything (unless I missed some field, hopefully not)

src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
alhijazi: Removed, I'll also remove the testcase and variant UworkerEntityWrapper once the other PR lands.

oliverchang (approved):

src/clusterfuzz/_internal/protos/uworker_msg.proto
jonathanmetzman: Did you look in setup.py as well?

alhijazi: Yes, I also checked setup.py, and as far as I saw, it only reads data from testcase and never updates it.

PR #3514, opened 2023-11-16
Fix detection of 'unreachable code' in V8

backes:
Do not overwrite the already-detected 'unreachable code' state with the ASan abort.
We extend an existing test to include the ASan report (this is the output of a real example which got misclassified as "Abrt · NULL").


backes:
@oliverchang @jonathanmetzman 
Another little fix (and a more complete test case). PTAL.

jonathanmetzman (approved):
lgtm

backes:
Thanks! Feel free to merge when everything is green.

PR #3515, opened 2023-11-16
archive: rework archive classes to make it extendable

paulsemel:
In chrome, we have very specific needs when it comes to unzipping. For instance, every chrome fuzzers are package with a `.runtime_deps` file that lists all the runtime dependencies needed by the target. We could use this to selectively unpack our archives.

Similarly, we could use ripunzip to improve our zipping process.

This rework will enable those new features more easily, but also make the code clearer on what exactly is going on.

### DIFF
--- a/src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py
@@ -1327,7 +1327,7 @@ def use_peach_mutator(extra_env, grammar):
   unzipped = os.path.join(peach_dir, 'mutator')
   source = os.path.join(peach_dir, 'peach_mutator.zip')
 
-  reader = archive.get_archive_reader(source)
+  reader = archive.open(source)
   archive.unpack(reader, unzipped, trusted=True)
 
   # Set LD_PRELOAD.
--- a/src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py
@@ -1327,8 +1327,8 @@ def use_peach_mutator(extra_env, grammar):
   unzipped = os.path.join(peach_dir, 'mutator')
   source = os.path.join(peach_dir, 'peach_mutator.zip')
 
-  reader = archive.open(source)
-  archive.unpack(reader, unzipped, trusted=True)
+  with archive.open(source) as reader:
+    archive.unpack(reader, unzipped, trusted=True)
 
   # Set LD_PRELOAD.
   peach_path = os.path.join(unzipped, 'peach_mutator', 'src', 'peach.so')
--- a/src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py
@@ -1327,7 +1327,9 @@ def use_peach_mutator(extra_env, grammar):
   unzipped = os.path.join(peach_dir, 'mutator')
   source = os.path.join(peach_dir, 'peach_mutator.zip')
 
-  archive.unpack(source, unzipped, trusted=True)
+  reader = archive.get_archive_reader(source)
+  assert reader
+  archive.unpack(reader, unzipped, trusted=True)
 
   # Set LD_PRELOAD.
   peach_path = os.path.join(unzipped, 'peach_mutator', 'src', 'peach.so')
--- a/src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py
@@ -1328,7 +1328,6 @@ def use_peach_mutator(extra_env, grammar):
   source = os.path.join(peach_dir, 'peach_mutator.zip')
 
   reader = archive.get_archive_reader(source)
-  assert reader
   archive.unpack(reader, unzipped, trusted=True)
 
   # Set LD_PRELOAD.
--- a/src/clusterfuzz/_internal/build_management/build_manager.py
+++ b/src/clusterfuzz/_internal/build_management/build_manager.py
@@ -543,7 +543,7 @@ class Build(BaseBuild):
     try:
       reader = archive.open(build_local_archive)
     except:
-      logs.log_error('Unable to open build archive %s.' % build_local_archive)
+      logs.log_error(f'Unable to open build archive {build_local_archive}.')
       return False
 
     if not unpack_everything:
@@ -582,6 +582,7 @@ class Build(BaseBuild):
       logs.log_error('Unable to unpack build archive %s.' % build_local_archive)
       return False
 
+    reader.close()
     if unpack_everything:
       # Set a random fuzz target now that the build has been unpacked, if we
       # didn't set one earlier. For an auxiliary build, fuzz target is already
@@ -845,8 +846,8 @@ class CuttlefishKernelBuild(RegularBuild):
     # Extract syzkaller binary.
     syzkaller_path = os.path.join(self.build_dir, 'syzkaller')
     shell.remove_directory(syzkaller_path)
-    reader = archive.open(archive_dst_path)
-    archive.unpack(reader, syzkaller_path)
+    with archive.open(archive_dst_path) as reader:
+      archive.unpack(reader, syzkaller_path)
     shell.remove_file(archive_dst_path)
 
     environment.set_value('VMLINUX_PATH', self.build_dir)
@@ -990,6 +991,7 @@ class CustomBuild(Build):
       # Remove the archive.
       shell.remove_file(build_local_archive)
 
+    reader.close()
     self._pick_fuzz_target(
         self._get_fuzz_targets_from_dir(self.build_dir), self.target_weights)
     return True
--- a/src/clusterfuzz/_internal/build_management/build_manager.py
+++ b/src/clusterfuzz/_internal/build_management/build_manager.py
@@ -151,13 +151,14 @@ def _make_space(requested_size, current_build_dir=None):
   return result
 
 
-def _make_space_for_build(build_local_archive,
+def _make_space_for_build(archive_reader,
                           current_build_dir,
                           file_match_callback=None):
   """Make space for extracting the build archive by deleting the least recently
   used builds."""
-  extracted_size = archive.extracted_size(
-      build_local_archive, file_match_callback=file_match_callback)
+  extracted_size = 0
+  if archive_reader:
+    extracted_size = archive_reader.extracted_size(file_match_callback)
 
   return _make_space(extracted_size, current_build_dir=current_build_dir)
 
@@ -558,8 +559,8 @@ class Build(BaseBuild):
     file_match_callback = _get_file_match_callback()
     assert not (unpack_everything and file_match_callback is not None)
 
-    if not _make_space_for_build(build_local_archive, base_build_dir,
-                                 file_match_callback):
+    reader = archive.get_archive_reader(build_local_archive)
+    if not _make_space_for_build(reader, base_build_dir, file_match_callback):
       shell.clear_data_directories()
       logs.log_fatal_and_exit(
           'Failed to make space for build. '
@@ -965,7 +966,8 @@ class CustomBuild(Build):
 
     # If custom binary is an archive, then unpack it.
     if archive.is_archive(self.custom_binary_filename):
-      if not _make_space_for_build(build_local_archive, self.base_build_dir):
+      if not _make_space_for_build(
+          archive.get_archive_reader(build_local_archive), self.base_build_dir):
         # Remove downloaded archive to free up space and otherwise, it won't get
         # deleted until next job run.
         shell.remove_file(build_local_archive)
--- a/src/clusterfuzz/_internal/build_management/build_manager.py
+++ b/src/clusterfuzz/_internal/build_management/build_manager.py
@@ -157,8 +157,7 @@ def _make_space_for_build(archive_reader,
   """Make space for extracting the build archive by deleting the least recently
   used builds."""
   extracted_size = 0
-  if archive_reader:
-    extracted_size = archive_reader.extracted_size(file_match_callback)
+  extracted_size = archive_reader.extracted_size(file_match_callback)
 
   return _make_space(extracted_size, current_build_dir=current_build_dir)
 
@@ -612,7 +611,7 @@ class Build(BaseBuild):
     # Import here as this path is not available in App Engine context.
     from clusterfuzz._internal.bot.fuzzers import utils as fuzzer_utils
 
-    for archive_file in reader.list_files():
+    for archive_file in reader.list_members():
       if fuzzer_utils.is_fuzz_target_local(
           archive_file.filename, reader.try_open(archive_file.filename)):
         fuzz_target = _normalize_target_name(archive_file.filename)
--- a/src/clusterfuzz/_internal/build_management/build_manager.py
+++ b/src/clusterfuzz/_internal/build_management/build_manager.py
@@ -156,7 +156,6 @@ def _make_space_for_build(archive_reader,
                           file_match_callback=None):
   """Make space for extracting the build archive by deleting the least recently
   used builds."""
-  extracted_size = 0
   extracted_size = archive_reader.extracted_size(file_match_callback)
 
   return _make_space(extracted_size, current_build_dir=current_build_dir)
@@ -612,9 +611,9 @@ class Build(BaseBuild):
     from clusterfuzz._internal.bot.fuzzers import utils as fuzzer_utils
 
     for archive_file in reader.list_members():
-      if fuzzer_utils.is_fuzz_target_local(
-          archive_file.filename, reader.try_open(archive_file.filename)):
-        fuzz_target = _normalize_target_name(archive_file.filename)
+      if fuzzer_utils.is_fuzz_target_local(archive_file.name,
+                                           reader.try_open(archive_file.name)):
+        fuzz_target = _normalize_target_name(archive_file.name)
         yield fuzz_target
 
   def _get_fuzz_targets_from_dir(self, build_dir):
--- a/src/clusterfuzz/_internal/build_management/build_manager.py
+++ b/src/clusterfuzz/_internal/build_management/build_manager.py
@@ -566,12 +566,15 @@ class Build(BaseBuild):
           'Failed to make space for build. '
           'Cleared all data directories to free up space, exiting.')
 
+    if not reader:
+      logs.log_error('Unable to open archive %s.' % build_local_archive)
+      return False
     # Unpack the local build archive.
     logs.log('Unpacking build archive %s.' % build_local_archive)
     trusted = not utils.is_oss_fuzz()
     try:
       archive.unpack(
-          build_local_archive,
+          reader,
           build_dir,
           trusted=trusted,
           file_match_callback=file_match_callback)
@@ -846,7 +849,9 @@ class CuttlefishKernelBuild(RegularBuild):
     # Extract syzkaller binary.
     syzkaller_path = os.path.join(self.build_dir, 'syzkaller')
     shell.remove_directory(syzkaller_path)
-    archive.unpack(archive_dst_path, syzkaller_path)
+    reader = archive.get_archive_reader(archive_dst_path)
+    assert reader
+    archive.unpack(reader, syzkaller_path)
     shell.remove_file(archive_dst_path)
 
     environment.set_value('VMLINUX_PATH', self.build_dir)
@@ -974,9 +979,14 @@ class CustomBuild(Build):
 
         logs.log_fatal_and_exit('Could not make space for build.')
 
-      try:
-        archive.unpack(build_local_archive, self.build_dir, trusted=True)
-      except:
+      reader = archive.get_archive_reader(build_local_archive)
+      error_occurred = reader is None
+      if reader:
+        try:
+          archive.unpack(reader, self.build_dir, trusted=True)
+        except:
+          error_occurred = True
+      if error_occurred:
         logs.log_error(
             'Unable to unpack build archive %s.' % build_local_archive)
         return False
--- a/src/clusterfuzz/_internal/build_management/build_manager.py
+++ b/src/clusterfuzz/_internal/build_management/build_manager.py
@@ -606,10 +606,14 @@ class Build(BaseBuild):
     # Import here as this path is not available in App Engine context.
     from clusterfuzz._internal.bot.fuzzers import utils as fuzzer_utils
 
-    for archive_file in archive.iterator(archive_path):
-      if fuzzer_utils.is_fuzz_target_local(archive_file.name,
-                                           archive_file.handle):
-        fuzz_target = _normalize_target_name(archive_file.name)
+    reader = archive.get_archive_reader(archive_path)
+    if not reader:
+      return
+
+    for archive_file in reader.list_files():
+      if fuzzer_utils.is_fuzz_target_local(
+          archive_file.filename, reader.try_open(archive_file.filename)):
+        fuzz_target = _normalize_target_name(archive_file.filename)
         yield fuzz_target
 
   def _get_fuzz_targets_from_dir(self, build_dir):
--- a/src/clusterfuzz/_internal/build_management/build_manager.py
+++ b/src/clusterfuzz/_internal/build_management/build_manager.py
@@ -543,7 +543,7 @@ class Build(BaseBuild):
         'UNPACK_ALL_FUZZ_TARGETS_AND_FILES')
 
     try:
-      reader = archive.get_archive_reader(build_local_archive)
+      reader = archive.open(build_local_archive)
     except:
       logs.log_error('Unable to open build archive %s.' % build_local_archive)
       return False
@@ -847,7 +847,7 @@ class CuttlefishKernelBuild(RegularBuild):
     # Extract syzkaller binary.
     syzkaller_path = os.path.join(self.build_dir, 'syzkaller')
     shell.remove_directory(syzkaller_path)
-    reader = archive.get_archive_reader(archive_dst_path)
+    reader = archive.open(archive_dst_path)
     archive.unpack(reader, syzkaller_path)
     shell.remove_file(archive_dst_path)
 
@@ -967,7 +967,7 @@ class CustomBuild(Build):
       return False
 
     try:
-      reader = archive.get_archive_reader(build_local_archive)
+      reader = archive.open(build_local_archive)
     except:
       logs.log_error('Unable to open build archive %s.' % build_local_archive)
       return False
--- a/src/clusterfuzz/_internal/build_management/build_manager.py
+++ b/src/clusterfuzz/_internal/build_management/build_manager.py
@@ -541,6 +541,13 @@ class Build(BaseBuild):
 
     unpack_everything = environment.get_value(
         'UNPACK_ALL_FUZZ_TARGETS_AND_FILES')
+
+    try:
+      reader = archive.get_archive_reader(build_local_archive)
+    except:
+      logs.log_error('Unable to open build archive %s.' % build_local_archive)
+      return False
+
     if not unpack_everything:
       # For fuzzing, pick a random fuzz target so that we only un-archive that
       # particular fuzz target and its dependencies and save disk space.  If we
@@ -551,24 +558,19 @@ class Build(BaseBuild):
       # large builds (such as Chrome OS). Defer setting it until after the build
       # has been unpacked.
       self._pick_fuzz_target(
-          self._get_fuzz_targets_from_archive(build_local_archive),
-          target_weights)
+          self._get_fuzz_targets_from_archive(reader), target_weights)
 
     # Actual list of files to unpack can be smaller if we are only unarchiving
     # a particular fuzz target.
     file_match_callback = _get_file_match_callback()
     assert not (unpack_everything and file_match_callback is not None)
 
-    reader = archive.get_archive_reader(build_local_archive)
     if not _make_space_for_build(reader, base_build_dir, file_match_callback):
       shell.clear_data_directories()
       logs.log_fatal_and_exit(
           'Failed to make space for build. '
           'Cleared all data directories to free up space, exiting.')
 
-    if not reader:
-      logs.log_error('Unable to open archive %s.' % build_local_archive)
-      return False
     # Unpack the local build archive.
     logs.log('Unpacking build archive %s.' % build_local_archive)
     trusted = not utils.is_oss_fuzz()
@@ -605,14 +607,10 @@ class Build(BaseBuild):
 
     return True
 
-  def _get_fuzz_targets_from_archive(self, archive_path):
+  def _get_fuzz_targets_from_archive(self, reader):
     """Get iterator of fuzz targets from archive path."""
     # Import here as this path is not available in App Engine context.
     from clusterfuzz._internal.bot.fuzzers import utils as fuzzer_utils
-
-    reader = archive.get_archive_reader(archive_path)
-    if not reader:
-      return
 
     for archive_file in reader.list_files():
       if fuzzer_utils.is_fuzz_target_local(
@@ -850,7 +848,6 @@ class CuttlefishKernelBuild(RegularBuild):
     syzkaller_path = os.path.join(self.build_dir, 'syzkaller')
     shell.remove_directory(syzkaller_path)
     reader = archive.get_archive_reader(archive_dst_path)
-    assert reader
     archive.unpack(reader, syzkaller_path)
     shell.remove_file(archive_dst_path)
 
@@ -969,24 +966,25 @@ class CustomBuild(Build):
     if not blobs.read_blob_to_disk(self.custom_binary_key, build_local_archive):
       return False
 
+    try:
+      reader = archive.get_archive_reader(build_local_archive)
+    except:
+      logs.log_error('Unable to open build archive %s.' % build_local_archive)
+      return False
+
     # If custom binary is an archive, then unpack it.
     if archive.is_archive(self.custom_binary_filename):
-      if not _make_space_for_build(
-          archive.get_archive_reader(build_local_archive), self.base_build_dir):
+      if not _make_space_for_build(reader, self.base_build_dir):
         # Remove downloaded archive to free up space and otherwise, it won't get
         # deleted until next job run.
+        reader.close()
         shell.remove_file(build_local_archive)
 
         logs.log_fatal_and_exit('Could not make space for build.')
 
-      reader = archive.get_archive_reader(build_local_archive)
-      error_occurred = reader is None
-      if reader:
-        try:
-          archive.unpack(reader, self.build_dir, trusted=True)
-        except:
-          error_occurred = True
-      if error_occurred:
+      try:
+        archive.unpack(reader, self.build_dir, trusted=True)
+      except:
         logs.log_error(
             'Unable to unpack build archive %s.' % build_local_archive)
         return False
--- a/src/clusterfuzz/_internal/system/archive.py
+++ b/src/clusterfuzz/_internal/system/archive.py
@@ -54,11 +54,11 @@ def _is_attempting_path_traversal(archive_name, output_dir, filename) -> bool:
 class ArchiveMemberInfo:
   """ArchiveMemberInfo"""
 
-  def __init__(self, filename, is_dir, file_size, compress_size):
+  def __init__(self, filename, is_dir, file_size, mode):
     self._filename = filename
     self._is_dir = is_dir
     self._file_size = file_size
-    self._compress_size = compress_size
+    self._mode = mode
 
   @property
   def filename(self):
@@ -72,8 +72,8 @@ class ArchiveMemberInfo:
     return self._file_size
 
   @property
-  def compress_size(self):
-    return self._compress_size
+  def mode(self):
+    return self._mode
 
 
 class ArchiveReader:
@@ -87,6 +87,9 @@ class ArchiveReader:
     raise NotImplementedError
 
   def open(self, member):
+    raise NotImplementedError
+
+  def close(self) -> None:
     raise NotImplementedError
 
   def try_open(self, member):
@@ -120,14 +123,15 @@ class TarArchiveReader(ArchiveReader):
   def list_files(self) -> [ArchiveMemberInfo]:
     return [
         ArchiveMemberInfo(
-            filename=f.name,
-            is_dir=f.isdir(),
-            file_size=f.size,
-            compress_size=f.size) for f in self.archive.getmembers()
+            filename=f.name, is_dir=f.isdir(), file_size=f.size, mode=f.mode)
+        for f in self.archive.getmembers()
     ]
 
   def open(self, member):
     return self.archive.extractfile(member)
+
+  def close(self) -> None:
+    self.archive.close()
 
   def extract(self, member, path=None, trusted=False):
     # If the output directory is a symlink, get its actual path since we will be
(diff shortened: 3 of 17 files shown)


paulsemel:
No need to review as for now, this is really just a draft for now.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -34,6 +33,197 @@
...
+
+  if real_file_path != absolute_file_path:
+    logs.log_error('Directory traversal attempted while unpacking archive %s '
+                   '(file path=%s, actual file path=%s). Aborting.' %
+                   (archive_name, absolute_file_path, real_file_path))
+    return True
+  return False
+
+
+class ArchiveMemberInfo:
```
letitz, line 54: Consider using a dataclass for this to automate the boilerplate:

```py
import dataclasses

@dataclasses.dataclass
class ArchiveMemberInfo:
  filename: str
  is_dir: bool
  file_size: int
  compress_size: int
```

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -34,6 +33,197 @@
...
+  """ArchiveMemberInfo"""
+
+  def __init__(self, filename, is_dir, file_size, compress_size):
+    self._filename = filename
+    self._is_dir = is_dir
+    self._file_size = file_size
+    self._compress_size = compress_size
```
letitz, line 61: What unit are these in? It would be best to suffix the member names to clarify: `file_size_bytes`, `file_size_mb`, etc.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -34,6 +33,197 @@
...
+  @property
+  def file_size(self):
+    return self._file_size
+
+  @property
+  def compress_size(self):
+    return self._compress_size
+
+
+class ArchiveReader:
```
letitz, line 79: This seems like an Abstract Base Class, consider using `abc` to ensure subclasses override methods:

```py
import abc

class ArchiveReader(abc.ABC):

  @abc.abstractmethod
  def list_files(self) ...
```

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -34,6 +33,197 @@
...
+  """ArchiveReader"""
+
+  # This should have a few of the handful functions a tool like unzip has
+  def list_files(self) -> [ArchiveMemberInfo]:
```
letitz, line 83: List type annotation should be `List[ArchiveMemberInfo]` in Python 3.7, and you'll need to do `from typing import List`.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -34,6 +33,197 @@
...
+    return sum(f.file_size
+               for f in self.list_files()
+               if not file_match_callback or file_match_callback(f.filename))
+
+
+class TarArchiveReader(ArchiveReader):
+  """TarArchiveReader"""
+
+  def __init__(self, archive_path, file_obj=None) -> None:
+    # we need to know whether it's a lzma file or not.
```
letitz, line 105: Though this bit is ignored with `file_obj` is not None.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -34,6 +33,197 @@
...
+  def open(self, member):
+    return self.archive.extractfile(member)
+
+  def extract(self, member, path=None, trusted=False):
+    # If the output directory is a symlink, get its actual path since we will be
+    # doing directory traversal checks later when unpacking the archive.
+    output_directory = os.path.realpath(path)
+
+    if not trusted:
+      if (_is_attempting_path_traversal(self.archive_path, output_directory,
```
letitz, line 132: This is not your fault, but do we really want to allow path traversal in *any* case? Why gate it on `trusted`?

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -34,6 +33,197 @@
...
+            compress_size=f.compress_size) for f in self.zip_archive.infolist()
+    ]
+
+  def open(self, member):
+    return self.zip_archive.open(name=member)
+
+  def extract(self, member, path=None, trusted=False):
+    # If the output directory is a symlink, get its actual path since we will be
+    # doing directory traversal checks later when unpacking the archive.
+    output_directory = os.path.realpath(path)
```
letitz, line 167: What happens here if `path` is `None`?

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -57,88 +247,38 @@ def iterator(archive_path,
...
+        f for f in reader.list_files()
+        if not file_match_callback or file_match_callback(f.filename)
+    ]
+    error_file_paths = []
+    for file in files:
+      try:
+        yield ArchiveFile(
+            file.filename,
+            file.file_size,
+            handle=None if not should_extract else reader.open(file.filename))
```
letitz, line 264: `reader.open()` is the only operation that can throw here, let's rejigger the code to make this explicit:

```py
handle = None
try:
  handle = reader.open(file.filename)
except KeyError:
  error_file_paths.append(file.filename)

yield ArchiveFile(file.filename, file.file_size, handle)
```

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -57,88 +247,38 @@ def iterator(archive_path,
...
+      except KeyError:
+        error_file_paths.append(file.filename)
+        yield ArchiveFile(file.filename, file.file_size, None)
+    if error_file_paths:
+      logs.log_warn(
+          'Check archive %s for broken links.' % archive_path,
+          error_filepaths=error_file_paths)
+  except:
+    logs.log_error(f'Bad file {archive_path}')
```
letitz, line 273: Let's capture the exception and log it, otherwise this log is nearly useless.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -163,24 +303,26 @@ def has_extension(extensions):
...
-def get_file_list(archive_path, archive_obj=None, file_match_callback=None):
+def get_file_list(archive_path, file_match_callback=None):
   """List all files in an archive."""
+  reader = get_archive_reader(archive_path)
   return [
-      f.name for f in iterator(
-          archive_path, archive_obj, file_match_callback, should_extract=False)
+      f.filename
+      for f in reader.list_files()
+      if not file_match_callback or file_match_callback(f.filename)
```
letitz, line 312: This is the third instance of this pattern yet, how about extracting it to a method on `ArchiveReader`:

```py
class ArchiveReader(abc.ABC):

 # This is overridden by subclasses.
 @abc.abstractmethod
 def list_all_files(self) ...

def list_files(self, file_match_callback=None):
  if file_match_callback is None:
    return self.list_all_files()

  return [f for f in self.list_all_files() if file_match_callback(f.filename))]
```

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -34,6 +33,197 @@
...
+      self.extract(member=member, path=path, trusted=trusted)
+
+
+class ZipArchiveReader(ArchiveReader):
+  """ZipArchiveReader"""
+
+  def __init__(self, file) -> None:
+    self.zip_archive = zipfile.ZipFile(file, mode='r')
+
+  def list_files(self):
```
letitz, line 152: Consider yielding entries instead, since many usecases below don't need the whole list at once.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -163,24 +303,26 @@ def has_extension(extensions):
...
 
 def get_first_file_matching(search_string, archive_obj, archive_path):
   """Returns the first file with a name matching search string in archive."""
-  for current_file in get_file_list(archive_path, archive_obj):
-    # Exclude MAC resouce forks.
-    if current_file.startswith('__MACOSX/'):
+  reader = get_archive_reader(archive_path, archive_obj)
+  for file in reader.list_files():
+    if file.filename.startswith('__MACOSX/'):
+      # Exclude MAC resouce forks.
```
letitz, line 321:
```suggestion
      # Exclude MAC resource forks.
```

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -197,137 +339,28 @@ def unpack(archive_path,
   if not os.path.exists(archive_path):
     logs.log_error('Archive %s not found.' % archive_path)
     return False
+  reader = get_archive_reader(archive_path=archive_path)
+  assert reader is not None
```
letitz, line 343: It seems that every call to `get_archive_reader()` assumes this is true. Let's extract this check into `get_archive_reader()` and add some information to the exception (filename)?

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -197,137 +339,28 @@ def unpack(archive_path,
...
-
-        # Keep heartbeat happy by updating with our progress.
-        archive_file_unpack_count += 1
-        if archive_file_unpack_count % 1000 == 0:
-          logs.log('Unpacked %d/%d.' % (archive_file_unpack_count,
-                                        archive_file_total_count))
-
+  error_occurred = False
+  for file in file_list:
+    error_occurred = reader.extract(
```
letitz, line 358: This should likely be:

```suggestion
    error_occurred |= reader.extract(
```

letitz (commented):
Drive-by review.

src/clusterfuzz/_internal/system/archive.py
paulsemel: I didn't know about this, thanks, that's pretty cool!

src/clusterfuzz/_internal/system/archive.py
paulsemel: Good point, done!

src/clusterfuzz/_internal/system/archive.py
paulsemel: Agreed. For now, I've let the `trusted` part in. Once this gets merged, I'll change all those annoying things in a second time. The idea is to have sort of a "non functional change" for this CL.

src/clusterfuzz/_internal/system/archive.py
paulsemel: Yeah, let's not allow an optional path here.

paulsemel:
**Note to reviewers**:

This PR will be followed by other PRs (at least one) that will improve the way ClusterFuzz unzips fuzzers. This first PR is meant to centralize and make it easier to extend the current functionality. For instance, `build_manager.py` now only uses `ArchiveReader`, and from there it can be easy to:
1. Use `ripunzip` and bundle it into an `ArchiveReader` and things will "just work".
2. Improve the actual intelligence behind unzipping part of the archive. The current callback mechanism remains for now just because I do not want this PR to be too big, but in a subsequent PR it will go away (and it'll be easier to make it go away). It will also make it clearer what exactly is being done.

paulsemel:
Hey @jonathanmetzman @oliverchang, can you PTAL? I think this CL can open the door to unblock us on various issues we currently have (regarding `.runtime_deps` file mostly).

src/clusterfuzz/_internal/system/archive.py
oliverchang: +1, let's just remove this concept completely and disallow path traversal in all cases.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -34,111 +36,313 @@
...
+      archive_name (str): the name of the archive.
+      output_dir (str): the output directory.
+      filename (str): the name of the file being checked.
 
-class ArchiveFile:
-  """File in an archive."""
+  Returns:
+      bool: Whether there is a path traversal attempt
+  """
+  absolute_file_path = os.path.join(output_dir, os.path.normpath(filename))
```
oliverchang, line 50: can we do `output_dir = os.path.realpath(output_dir)` before this?

If `output_dir`  has `..` in it, e.g. `../a` or some other variation (which is trusted), then

```
absolute_file_path == '../a/filename'
real_file_path = '<some other dir>/a/filename`
```

Then this will fail, even though there's no directory traversal.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -34,111 +36,313 @@
...
+    for member in to_extract:
+      self.extract(member=member, path=path, trusted=trusted)
+
+
+class ZipArchiveReader(ArchiveReader):
+  """A zip archive reader.
+  """
+
+  def __init__(self, file) -> None:
+    self.zip_archive = zipfile.ZipFile(file, mode='r')
```
oliverchang, line 239: Can we make these members private? i.e. `self._zip_archive`

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -34,111 +36,313 @@
...
+               if not file_match_callback or file_match_callback(f.filename))
+
+
+class TarArchiveReader(ArchiveReader):
+  """A tar archive reader. Currently supports TAR and TAR_LZMA archives.
+  """
+
+  def __init__(self, archive_path, file_obj=None) -> None:
+    if file_obj:
+      self.archive = tarfile.open(fileobj=file_obj)
```
oliverchang, line 192: same comment here around making members private.

oliverchang (commented):
this looks great! thanks for refactoring this.

src/clusterfuzz/_internal/system/archive.py
paulsemel: Oh, nice, good point, I didn't think about this!

src/clusterfuzz/_internal/system/archive.py
paulsemel: Totally agree on this. I just would like to do it separately, because zip also does things with symlinks when the archive is trusted, and I am unsure what exactly uses this. I am pretty confident with the current state of this PR, so the idea is to first land this "no functional change" PR, and then land much smaller ones that will modify the behavior slightly to clean things up.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -34,111 +36,316 @@
...
 
-  if archive_type == ArchiveType.ZIP:
-    try:
-      with zipfile.ZipFile(archive_obj or archive_path) as zip_file:
-        for info in zip_file.infolist():
-          if not file_match_callback(info.filename):
-            continue
+  @abc.abstractmethod
+  def extract(self, member, path, trusted=False) -> str:
+    """Extracts `member` out of the archive to the provided path.
```
letitz, line 93: What happens if the member is a directory? I assume contents are extracted too, can we document this?

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -34,111 +36,316 @@
...
+    """Tries to open the archive. Does not throw.
 
-    except (zipfile.BadZipfile, zipfile.LargeZipFile):
-      logs.log_error('Bad zip file %s.' % archive_path)
+    Args:
+        member (str): the member name
 
-  elif archive_type == ArchiveType.TAR:
+    Returns:
+        [IO]: a file-like object that can be `read`.
```
letitz, line 145: Or `None`. The function can be annotated as `-> Optional[IO]` (with `from typing import Optional`).

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -34,111 +36,316 @@
...
+    raise NotImplementedError
+
+  @abc.abstractmethod
+  def close(self) -> None:
+    """Closes the archive.
+    """
+    raise NotImplementedError
+
+  @abc.abstractmethod
+  def extractall(self, path, members=None, trusted=False) -> None:
```
letitz, line 126: Maybe this is modelled after the `zip` module API in which case it can stay, but for consistency it would be nice to keep this `snake_case` and name it `extract_all()`.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -34,111 +36,316 @@
...
+      if file.filename.startswith('__MACOSX/'):
+        # Exclude MAC resource forks.
+        continue
+
+      if search_string in file.filename:
+        return file.filename
+    return None
+
+  def extracted_size(self, file_match_callback=None):
+    """Gets the total extracted size of the file matched by
```
letitz, line 171:
```suggestion
    """Gets the total extracted size of the files matched by
```

Multiple files can match.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -34,111 +36,316 @@
...
+    file_match_callback. If file_match_callback is None, gets the extracted
+    size of the whole archive.
+
+    Args:
+        file_match_callback (optional): the file matching callback. Defaults
+        to None.
+
+    Returns:
+        int: the extracted size.
```
letitz, line 180:
```suggestion
        int: the extracted size in bytes.
```

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -34,111 +36,316 @@
...
+  def __init__(self, archive_path, file_obj=None) -> None:
+    if file_obj:
+      self._archive = tarfile.open(fileobj=file_obj)
+    else:
+      archive_type = get_archive_type(archive_path)
+      mode = 'r' if archive_type == ArchiveType.TAR else 'r:xz'
+      self._archive = tarfile.open(archive_path, mode=mode)
+    self._archive_path = archive_path
+
+  def list_files(self) -> typing.List[ArchiveMemberInfo]:
```
letitz, line 200:
```suggestion
  def list_files(self) -> List[ArchiveMemberInfo]:
```

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -34,111 +36,316 @@
...
+    # If the output directory is a symlink, get its actual path since we will be
+    # doing directory traversal checks later when unpacking the archive.
+    output_directory = os.path.realpath(path)
+
+    if not trusted:
+      if (_is_attempting_path_traversal(self._archive_path, output_directory,
+                                        member)):
+        return None
+
+    self._archive.extract(member=member, path=output_directory)
```
letitz, line 225: Consider extracting (no pun intended) the path traversal check to `ArchiveReader`:

```py
def extract(...):
  if attempting_traversal(...):
    return None

  return self._extract(...)

@abc.abstractmethod
def _extract(...):
  raise NotImplementedError
```

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -34,111 +36,316 @@
...
+    return os.path.realpath(os.path.join(output_directory, member))
+
+  def extractall(self, path, members=None, trusted=False) -> None:
+    to_extract = members if members is not None else self._archive.namelist()
+    #FIXME(paulsemel): we could try extracting that all.
```
letitz, line 230: Not sure what you mean here.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -34,111 +36,316 @@
...
+
+
+class ZipArchiveReader(ArchiveReader):
+  """A zip archive reader.
+  """
+
+  def __init__(self, file) -> None:
+    self._zip_archive = zipfile.ZipFile(file, mode='r')
+
+  def list_files(self) -> typing.List[ArchiveMemberInfo]:
```
letitz, line 242:
```suggestion
  def list_files(self) -> List[ArchiveMemberInfo]:
```

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -34,111 +36,316 @@
...
+
+  def extract(self, member, path, trusted=False):
+    # If the output directory is a symlink, get its actual path since we will be
+    # doing directory traversal checks later when unpacking the archive.
+    output_directory = os.path.realpath(path)
+
+    if not trusted:
+      if (_is_attempting_path_traversal(self._archive_path, output_directory,
+                                        member)):
+        return None
```
letitz, line 223: Can we log a warning or an error in this case?

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -34,111 +36,316 @@
...
+      # Create symlink if needed (only on unix platforms).
+      if (trusted and hasattr(os, 'symlink') and
+          oct(external_attr).startswith(SYMLINK_ATTRIBUTE)):
+        symlink_source = self._zip_archive.read(member)
+        if os.path.exists(extracted_path):
+          os.remove(extracted_path)
+        os.symlink(symlink_source, extracted_path)
+
+      return extracted_path
+    except:
```
letitz, line 302: Let's catch the exception and log something about it, instead of swallowing it silently.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -34,111 +36,316 @@
...
 
-  else:
-    logs.log_error('Unsupported compression type for file %s.' % archive_path)
+  Args:
+      archive_path (str): the path to the archive.
+      file_obj (obj, optional): a object-like containing the archive. Defaults
+      to None.
 
+  Raises:
+      this function raises if the file could not be open or if the archive type
```
letitz, line 328:
```suggestion
      If the file could not be opened or if the archive type
```

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -34,111 +36,316 @@
...
+      cannot be handled. See `get_archive_type` to check whether the archive
```
letitz, line 329:
```suggestion
      cannot be handled. See `get_archive_type()` and `is_archive()` to check whether the archive
```

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -163,171 +370,46 @@ def has_extension(extensions):
...
-  output_directory = os.path.realpath(output_directory)
-
-  archive_filename = os.path.basename(archive_path)
-  error_occurred = False
-
-  # Choose to unpack all files or ones matching a particular regex.
-  file_list = get_file_list(
-      archive_path, file_match_callback=file_match_callback)
+def unpack(reader, output_dir, trusted=False, file_match_callback=None):
+  """Unpacks the current archives opened with `reader`. If file_match_callback
```
letitz, line 379:
```suggestion
  """Unpacks the current archive opened with `reader`. If `file_match_callback`
```

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -163,171 +370,46 @@ def has_extension(extensions):
...
+  is None, unpacks all the archive. Otherwise, this only unpacks the files
+  matched by the callback.
+
+  Args:
+      reader (ArchiveReader): the archive reader
+      output_dir (_type_): the output directory to unpack the archive to.
```
letitz, line 385: This is a string, no?

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -163,171 +370,46 @@ def has_extension(extensions):
...
-  # doing directory traversal checks later when unpacking the archive.
-  output_directory = os.path.realpath(output_directory)
-
-  archive_filename = os.path.basename(archive_path)
-  error_occurred = False
-
-  # Choose to unpack all files or ones matching a particular regex.
-  file_list = get_file_list(
-      archive_path, file_match_callback=file_match_callback)
+def unpack(reader, output_dir, trusted=False, file_match_callback=None):
```
letitz, line 378: This seems like it could be a non-virtual method on `ArchiveReader`?

If you're strongly against adding concrete behavior to the base class, then consider the following design:

```py
class ArchiveReader:
  def __init__(impl: _ArchiveReaderImpl):
    self._impl = impl

  def extract(...):
    if traversal:
      return None
    return self._impl.extract(...)

  def unpack(...):
    ... do stuff ...
    for file in list:
      self._impl.extract(...)
    ...

class _ArchiveReaderImpl(abc.ABC):
  @abc.abstractmethod
  def extract(...):
    raise NotImplementedError

class _TarArchiveReaderImpl(ArchiveReaderImpl):
  def extract(...):
    ...

def open(file):
  if is_tar:
    return ArchiveReader(_TarArchiveReaderImpl(file))
  ...
```

Note that this lends itself to testing the shared code in `ArchiveReader` by injecting a mock impl.

letitz (changes requested):
Looking good! A bunch of small comments, one optional, design-level comment around inheritance vs composition.

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -151,13 +151,14 @@ def _make_space(requested_size, current_build_dir=None):
...
-def _make_space_for_build(build_local_archive,
+def _make_space_for_build(archive_reader,
                           current_build_dir,
                           file_match_callback=None):
   """Make space for extracting the build archive by deleting the least recently
   used builds."""
-  extracted_size = archive.extracted_size(
-      build_local_archive, file_match_callback=file_match_callback)
+  extracted_size = 0
+  if archive_reader:
```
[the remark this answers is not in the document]
paulsemel: So, this only existed because of `build_manager_test.py`, but I actually modified the latter to make this line not needed anymore, and forgot to remove it.

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -601,15 +607,15 @@ def _unpack_build(self,
...
     # Import here as this path is not available in App Engine context.
     from clusterfuzz._internal.bot.fuzzers import utils as fuzzer_utils
 
-    for archive_file in archive.iterator(archive_path):
-      if fuzzer_utils.is_fuzz_target_local(archive_file.name,
-                                           archive_file.handle):
-        fuzz_target = _normalize_target_name(archive_file.name)
+    for archive_file in reader.list_files():
+      if fuzzer_utils.is_fuzz_target_local(
+          archive_file.filename, reader.try_open(archive_file.filename)):
```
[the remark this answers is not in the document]
paulsemel: Right, I thought about that as well initially, but then figured it'd make the API weird for `ArchiveMemberInfo`.

src/clusterfuzz/_internal/system/archive.py
paulsemel: No, we only extract the directory, which means the directory will be created. I'll add a comment for this.

src/clusterfuzz/_internal/system/archive.py
paulsemel: We already do in the traversal check function, and I believe it is already pretty well descriptive.

src/clusterfuzz/_internal/system/archive.py
paulsemel: That's from previous stuff, good catch!

src/clusterfuzz/_internal/system/archive.py
paulsemel: I agree with you, but if you're ok with it, I'd like to do this in a second time, mostly because I'll have to design things in a way that it can be extended to take `.runtime_deps` into account.

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -157,8 +157,7 @@ def _make_space_for_build(archive_reader,
   """Make space for extracting the build archive by deleting the least recently
   used builds."""
   extracted_size = 0
```
letitz, line 159:
```suggestion
```

src/clusterfuzz/_internal/build_management/build_manager.py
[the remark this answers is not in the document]
letitz: Makes sense.

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -612,7 +611,7 @@ def _get_fuzz_targets_from_archive(self, reader):
     # Import here as this path is not available in App Engine context.
     from clusterfuzz._internal.bot.fuzzers import utils as fuzzer_utils
 
-    for archive_file in reader.list_files():
+    for archive_file in reader.list_members():
       if fuzzer_utils.is_fuzz_target_local(
           archive_file.filename, reader.try_open(archive_file.filename)):
```
letitz, line 616: 3 references to `archive_file.filename` should be changed to `archive_file.name`. The fact that this works means we must be missing a test?

src/clusterfuzz/_internal/system/archive.py
letitz: Fair enough!

src/clusterfuzz/_internal/system/archive.py
letitz: Got it.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -33,116 +39,357 @@
...
+    """
+    raise NotImplementedError
+
+  @abc.abstractmethod
+  def extract(self,
+              member: str,
+              path: Union[str, os.PathLike],
+              trusted: bool = False) -> str:
+    """Extracts `member` out of the archive to the provided path.
+    If `members` is a directory in the archive, only the directory itself will
```
letitz, line 110:
```suggestion
    If `member` is a directory in the archive, only the directory itself will
```

src/clusterfuzz/_internal/system/archive.py
letitz: Thanks!

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -66,53 +75,61 @@ def _is_attempting_path_traversal(archive_name, output_dir, filename) -> bool:
 
 @dataclasses.dataclass
 class ArchiveMemberInfo:
-  """Represents an archive member.
+  """Represents an archive member. A member can either be a file or a directory.
+  Members:
+    name: the name of the archive member.
```
letitz, line 80: Can this be a relative path for directory entries? e.g. `foo/bar`?

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -123,52 +140,54 @@ def close(self) -> None:
     raise NotImplementedError
 
   @abc.abstractmethod
-  def extractall(self, path, members=None, trusted=False) -> None:
+  def extract_all(self,
+                  path: Union[str, os.PathLike],
+                  members: List[str] = None,
```
letitz, line 145:
```suggestion
                  members: Optional[List[str]] = None,
```

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -123,52 +140,54 @@ def close(self) -> None:
...
+                  trusted: bool = False) -> None:
     """Extract the whole archive content or the members listed in `members`.
 
     Args:
-        path (str): the path where the members should be extracted.
-        Defaults to None.
-        members ([str], optional): the member names. Defaults to None.
-        trusted (bool, optional): whether the archive is trusted or not.
+        path: the path where the members should be extracted.
+        members (optional): the member names. Defaults to None.
```
letitz, line 151: The fact that these are optional, and their default value, is already documented in the function signature since they are kwargs.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -299,11 +328,16 @@ def extract(self, member, path, trusted=False):
         os.symlink(symlink_source, extracted_path)
 
       return extracted_path
-    except:
+    except Exception as e:
       # In case of errors, we try to extract whatever we can without errors.
+      logs.log_warn(
+          'An error occured while extracting the archive: %s.' % repr(e))
```
letitz, line 334:
```suggestion
          'An error occured while extracting %s from the archive: %s.' % (member, repr(e)))
```

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -315,22 +349,21 @@ class ArchiveError(Exception):
 
 
 # pylint: disable=redefined-builtin
-def open(archive_path, file_obj=None):
+def open(archive_path: str, file_obj: BinaryIO = None) -> ArchiveReader:
```
letitz, line 352:
```suggestion
def open(archive_path: str, file_obj: Optional[BinaryIO] = None) -> ArchiveReader:
```

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -315,22 +349,21 @@ class ArchiveError(Exception):
...
   """Opens the archive and gets the appropriate archive reader based on the
   `archive_path`.
```
letitz, line 355: We should document what happens when `file_obj` is not `None`.

letitz (commented):
A few more small comments.

src/clusterfuzz/_internal/system/archive.py
paulsemel: Yes!

src/clusterfuzz/_internal/build_management/build_manager.py
paulsemel: Well... those are hard to test, because we must mock what `list_members()` returns and ensure the members are "called". I think it is not necessary, mainly because this is a detail of the implementation of `build_manager.py`.
I think this is rather something to be caught with python type-checking rather than an actual test.

However, we do test what exactly `list_members()` does in `archive_test.py`, and we do ensure that it returns what's expected.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -330,13 +327,13 @@ def extract(self,
       return extracted_path
     except Exception as e:
       # In case of errors, we try to extract whatever we can without errors.
-      logs.log_warn(
-          'An error occured while extracting the archive: %s.' % repr(e))
+      logs.log_warn('An error occured while extracting %s the archive: %s.' %
```
letitz, line 330:
```suggestion
      logs.log_warn('An error occured while extracting %s from the archive: %s.' %
```

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -349,14 +346,15 @@ class ArchiveError(Exception):
 
 
 # pylint: disable=redefined-builtin
-def open(archive_path: str, file_obj: BinaryIO = None) -> ArchiveReader:
+def open(archive_path: str,
+         file_obj: Optional[BinaryIO] = None) -> ArchiveReader:
   """Opens the archive and gets the appropriate archive reader based on the
-  `archive_path`.
+  `archive_path`. If file_obj is not none, the binary file-like object will be
```
letitz, line 352:
```suggestion
  `archive_path`. If `file_obj` is not none, the binary file-like object will be
```

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -349,14 +346,15 @@ class ArchiveError(Exception):
...
+  used to read the archive instead of opening the file.
```
letitz, line 353:
```suggestion
  `archive_path`. If file_obj is not none, the binary file-like object will be
  used to read the archive instead of opening `archive_path`.
```

For some reason, I'm not able to comment on two lines at once.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -349,14 +346,15 @@ class ArchiveError(Exception):
...
 
   Args:
       archive_path: the path to the archive.
-      file_obj (optional): a object-like containing the archive. Defaults
-      to None.
+      file_obj: a object-like containing the archive.
```
letitz, line 357:
```suggestion
      file_obj: a file-like object containing the archive.
```

src/clusterfuzz/_internal/system/archive.py
letitz: Thanks! Let's document that then.

src/clusterfuzz/_internal/build_management/build_manager.py
letitz: I don't mean to suggest we should have an additional test for `ArchiveMemberInfo`, rather that this `build_manager` code seems unsufficiently tested. I can't figure out exactly what is supposed to be testing it, but in general we mock out `archive.open()` in `build_manager_test.py`, so I don't think the `list_members()` call will ever return a non-empty list, and we'll never hit this code. We might also be mocking out some function higher in the stack, in which case we'll never even execute this function.

Anyway, if you can't easily figure it out, let's leave it at that.

letitz (approved):
A couple last comments, LGTM otherwise!

paulsemel:
ping @oliverchang @jonathanmetzman, can you PTAL? This CL will unblock us on FuzzTest.

jonathanmetzman:
On some level I don't like this because it's a chrome specific solution when a more general one exists. In google3, fuzzers are seperated into seperate archives. I think this works as well no?

src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py
```diff
@@ -1343,7 +1343,8 @@ def use_peach_mutator(extra_env, grammar):
   unzipped = os.path.join(peach_dir, 'mutator')
   source = os.path.join(peach_dir, 'peach_mutator.zip')
 
-  archive.unpack(source, unzipped, trusted=True)
+  reader = archive.open(source)
+  archive.unpack(reader, unzipped, trusted=True)
```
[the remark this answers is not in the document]
oliverchang: +1. this feels cleaner.

oliverchang (approved):
Thank you again for this refactor, and very sorry for the extended delays here. 

This generally looks good, with just a +1 to one of @letitz's comments

oliverchang:
> On some level I don't like this because it's a chrome specific solution when a more general one exists. In google3, fuzzers are seperated into seperate archives. I think this works as well no?

I think this PR just does some refactoring to make the archive code cleaner and easier to extend. 

I'm OK with chrome-specific .runtime_deps file, but you do raise a good point: there's a larger change (to the chrome build process) that can make this all cleaner if we just split out build archives into individual targets (and don't selectively unpack at all).

paulsemel:
> > On some level I don't like this because it's a chrome specific solution when a more general one exists. In google3, fuzzers are seperated into seperate archives. I think this works as well no?
> 
> I think this PR just does some refactoring to make the archive code cleaner and easier to extend.

Agreed. I very honestly started working on this PR only to get the code cleaner. Then at some point when we started talking about Fuzztest support, we agreed that this idea would help ease things on our side, which is why it was mentioned in this PR. 

> I'm OK with chrome-specific .runtime_deps file, but you do raise a good point: there's a larger change (to the chrome build process) that can make this all cleaner if we just split out build archives into individual targets (and don't selectively unpack at all). Given the amount of work that may require, I'm happy to go with the .runtime_deps file approach for now.

I agree that the best solution would be to have separate archives. However, this could require a lot of work compared to `.runtime_deps` support, and thanks to this clean-up, the support for this will require a very low amount of code. wdyt @jonathanmetzman ?

jonathanmetzman:
> > On some level I don't like this because it's a chrome specific solution when a more general one exists. In google3, fuzzers are seperated into seperate archives. I think this works as well no?
> 
> I think this PR just does some refactoring to make the archive code cleaner and easier to extend.
> 
> I'm OK with chrome-specific .runtime_deps file, but you do raise a good point: there's a larger change (to the chrome build process) that can make this all cleaner if we just split out build archives into individual targets (and don't selectively unpack at all). Given the amount of work that may require, I'm happy to go with the .runtime_deps file approach for now.

It will be a long long project to switch I'm sure, though switching will solve a lot of problems (e.g. Windows fuzzers).
But it's probably unrealistic to expect this.
I suppose this is OK as a temporary measure, but I'm not so happy about complicatingthe existing build_manager code (which deals with Fuchsia builds, split builds, ordinary builds (UNPACK_ALL), builds that can be partially unpacked, cuttlefish kernel builds (can probably removed)) has too many 1 off solutions. It's fortunate that we don't need builds to be secret for the untrusted rearch, because modifying this mess would probably be impossible.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -33,116 +39,356 @@
...
+    for member in to_extract:
+      self.extract(member=member, path=path, trusted=trusted)
+
+
+class ArchiveError(Exception):
+  """ArchiveError"""
+
+
+# pylint: disable=redefined-builtin
+def open(archive_path: str,
```
jonathanmetzman, line 350: It's weird that we are returning an archive-file but calling it reader everywhere. I suppose that is because therew will be a conflict because this module is named "archive" but you can do `import ... as ...` or we can rename this to archivelib/archive_utils (but not now)

jonathanmetzman (commented):
First pass. Will look at archive.py more by monday.

jonathanmetzman (approved):
I did a relatively quick pass because Im a bit busy. It LGTM so I will merge to unblock @adetaylor

jonathanmetzman:
Ah sorry, I won't merge, I'll wait for Paul to fix my previous comments.

paulsemel:
Thanks @jonathanmetzman for spotting that we were missing the opportunity to use `with` here, that just makes things so much better!

jonathanmetzman:
LGTM

PR #3516, opened 2023-11-16
Add helper tip string for archive uploads

pgrace-google:
This message is shown in the quick uploads option, but some shepherds use the regular upload, where one has to know to hover to see this information.

crbug/1489702


jonathanmetzman (approved):
LGTM

src/appengine/private/components/upload-testcase/upload-form.html
```diff
@@ -276,6 +279,13 @@
             window.query.get(), this.paramKeys);
       }
 
+      shouldShowFiletypePicker(filePath) {
+        if (!filePath) {
+          return false;
+        }
+        return !filePath.includes('.html') && !filePath.includes('.js') && !filePath.includes('.pdf');
```
jonathanmetzman, line 286: Sorry why is this part needed?

pgrace-google: to show the message only when the file type is an archive! felt like it didnt need to be shown otherwise

jonathanmetzman (approved):
LGTM

PR #3518, opened 2023-11-16
[rearchitecture][analyze] Run uworker_main on cloud batch.

jonathanmetzman:
Make uworker_main of analyze_task execute on cloud batch. The next step is to drop the privileges of the service account we use.


oliverchang (commented):
LGTM

PR #3520, opened 2023-11-17
Improve batch execution.

jonathanmetzman:
1. Only do it on Linux.
2. return 0 on success.
3. Properly set env vars so clusterfuzz knows to execute utask main.
4. Dont do redundant updates before uworker_main.
5. log more


oliverchang (commented):
LGTM

PR #3528, opened 2023-11-21
Make analyze task run on tworkers

jonathanmetzman:
Avoid congestion issues caused by variant task - the most commonly run task :-(


oliverchang:
is this still required?

Issue #3531, opened 2023-11-21
Handle upgrades in new execution model


jonathanmetzman:
We need to specify the version of ClusterFuzz to use when launching uworker_main or postprocess. Otherwise changes can really break things.

PR #3533, opened 2023-11-21
Temporarily revert changes to output

jonathanmetzman:
They are breaking postprocess for older tasks.
https://pantheon.corp.google.com/errors/detail/CO6x7d-czdTsWA;filter=%5B%5D;time=PT6H?project=google.com:clusterfuzz


jonathanmetzman:
See https://github.com/google/clusterfuzz/issues/3531

PR #3536, opened 2023-11-22
Use protos directly in utasks

alhijazi:
This change removes the uworker_io wrappers and uses protos directly within the utasks. This change also removes the Json user defined proto.


src/clusterfuzz/_internal/bot/tasks/setup.py
```diff
@@ -19,13 +19,14 @@
 import time
 import zipfile
 
+from google.cloud.ndb import model
```
letitz, line 22: Using `model._entity_{to,from}_protobuf()` everywhere requires a bunch of lint-ignore comments, and does not ensure the right entity type is read from a protobuf. How about introducing some common helpers?

```
def model_to_protobuf(model: ndb.Model) -> entity_pb2.Entity:
  return ndb._model_to_protobuf(model)  # pylint: disable=protected-access

# Can be typed like this, but it's overkill without type checking.
from typing import Type
from typing import TypeVar

T = TypeVar('T', bound=ndb.Model)

def model_from_protobuf(entity: entity_pb2.Entity, model_type: Type[T]) -> T:
  model = ndb._model_from_protobuf(testcase)   # pylint: disable=protected-access
  assert isinstance(model, model_type)
  return model
```

We could even introduce specific helpers for each model type, e.g. `testcase_to_protobuf()`, but this is likely enough already.

src/clusterfuzz/_internal/bot/tasks/setup.py
```diff
@@ -227,12 +228,15 @@ def preprocess_setup_testcase(testcase, fuzzer_override=None):
...
   else:
-    setup_input = uworker_io.SetupInput()
+    setup_input = uworker_msg_pb2.SetupInput()
   setup_input.testcase_download_url = get_signed_testcase_download_url(testcase)
   return setup_input
 
 
-def setup_testcase(testcase, job_type, setup_input, metadata=None):
+def setup_testcase(testcase,
+                   job_type,
```
letitz, line 237: Optional: we can add type annotations while we're here:

```suggestion
def setup_testcase(testcase: data_types.Testcase,
                   job_type: str,
```

src/clusterfuzz/_internal/bot/tasks/setup.py
```diff
@@ -257,7 +261,7 @@ def setup_testcase(testcase, job_type, setup_input, metadata=None):
     _set_timeout_value_from_user_upload(testcase_id, metadata)
 
   # Update the fuzzer if necessary in order to get the updated data bundle.
-  if setup_input.fuzzer_name:
+  if setup_input.HasField("fuzzer_name"):
```
letitz, line 264: An empty fuzzer name is probably not valid, right? Can we just do:

```suggestion
  if setup_input.fuzzer_name:
```

src/clusterfuzz/_internal/bot/tasks/setup.py
```diff
@@ -524,34 +530,38 @@ def _set_fuzzer_env_vars(fuzzer):
...
+  data_bundles = list(
       ndb_utils.get_all_from_query(
-          data_types.DataBundle.query(data_types.DataBundle.name ==
-                                      update_input.fuzzer.data_bundle_name)))
-  logs.log('Data bundles: %s' % update_input.data_bundles)
+          data_types.DataBundle.query(
+              data_types.DataBundle.name == fuzzer.data_bundle_name)))
+  logs.log('Data bundles: %s' % data_bundles)
+
+  update_input.data_bundles.extend(  # pylint: disable=no-member
```
letitz, line 545: How come the linter complains about a missing member here? This PR seems to add a lot of such statements in places where I would not expect them.

src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
```diff
@@ -44,26 +44,22 @@ def uworker_main_no_io(utask_module, serialized_uworker_input):
 
   # Deal with the environment.
   set_uworker_env(uworker_input.uworker_env)  # pylint: disable=no-member
-  delattr(uworker_input, 'uworker_env')
+  uworker_input.ClearField("uworker_env")  # pylint: disable=no-member
```
letitz, line 47: Same question re: no-member.

Edit: there are a lot of these below. Same question applies, but I won't comment on all of them.

src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
```diff
@@ -437,6 +441,9 @@ def handle_build_setup_error(output):
         output.uworker_input.job_type,
         wait_time=testcase_fail_wait)
     return
+  testcase_upload_metadata = data_types.TestcaseUploadMetadata.query(
+      data_types.TestcaseUploadMetadata.testcase_id == int(
+          output.uworker_input.testcase_id)).get()
```
letitz, line 446: This could be extracted to a helper, there are 3 occurrences of this code in the file.

src/clusterfuzz/_internal/protos/uworker_msg.proto
```diff
@@ -192,21 +189,19 @@ enum ErrorType {
...
+  optional float crash_time = 4;
+  optional string crash_stacktrace_output = 5;
+  optional AnalyzeTaskOutput analyze_task_output = 6;
+  optional FuzzTaskOutput fuzz_task_output = 7;
+  optional MinimizeTaskOutput minimize_task_output = 8;
+  optional RegressionTaskOutput regression_task_output = 9;
+  optional ProgressionTaskOutput progression_task_output = 10;
+  optional SymbolizeTaskOutput symbolize_task_output = 11;
+  optional VariantTaskOutput variant_task_output = 12;
+  optional string error_message = 13;
```
letitz, line 206: Renumbering fields is dangerous if protos are being sent over the wire or stored on disk. I think this may already be the case, and that bit us during a previous deploy.

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -1863,32 +1862,43 @@ def run(self):
...
         known_crash_count=known_crash_count,
         testcases_executed=testcases_executed,
-        job_run_crashes=convert_groups_to_crashes(processed_groups),
     )
+    fuzz_task_output.job_run_crashes.extend(  # pylint: disable=no-member
+        convert_groups_to_crashes(processed_groups))
     if new_targets_count is not None:
-      fuzz_task_output.new_targets_count = new_targets_count
-    return uworker_io.UworkerOutput(fuzz_task_output=fuzz_task_output)
+      fuzz_task_output.new_targets_count = int(new_targets_count)
```
letitz, line 1876: It's surprising to me that a variable ending in `_count` is not an int already.

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -1863,32 +1862,43 @@ def run(self):
...
+    return uworker_msg_pb2.Output(fuzz_task_output=fuzz_task_output)
 
   def postprocess(self, uworker_output):
     """Handles postprocessing."""
     # TODO(metzman): Finish this.
     fuzz_task_output = uworker_output.fuzz_task_output
+    job_run_crashes = []
+    for crash in fuzz_task_output.job_run_crashes:
+      job_run_crashes.append({
```
letitz, line 1885: Optional: consider a list comprehension here:

```py
job_run_crashes = [
  {
    'is_new': crash.is_new,
    ...
  }
  for crash in fuzz_task_output.job_run_crashes
]
```

src/clusterfuzz/_internal/protos/uworker_msg.proto
```diff
@@ -42,7 +33,6 @@ message SymbolizeTaskInput {
 }
 
 message FuzzTaskInput {
```
letitz, line 35: Should this be removed entirely?

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -37,13 +38,14 @@
 from clusterfuzz._internal.system import environment
 
 
-def _maybe_clear_progression_last_min_max_metadata(testcase, uworker_output):
+def _maybe_clear_progression_last_min_max_metadata(
+    testcase, uworker_output: uworker_msg_pb2.Output):
```
letitz, line 42:
```suggestion
    testcase: data_types.Testcase, uworker_output: uworker_msg_pb2.Output):
```

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -315,8 +319,9 @@ def _testcase_reproduces_in_revision(testcase,
   _log_output(revision, result)
 
   if update_metadata:
-    progression_task_output.issue_metadata = _get_and_update_issue_metadata(
-        testcase)
+    progression_task_output.ClearField("issue_metadata")
```
letitz, line 322: Can you call `clear()` on the map field instead of clearing the field from the proto entirely?

```suggestion
    progression_task_output.issue_metadata.clear()
```

It avoids one string reference to a static field and expresses intent more precisely anyway.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -498,8 +502,8 @@ def find_fixed_range(uworker_input):
   if result and not result.is_crash():  # pylint: disable=no-member
     error_message = (
         f'Known crash revision {known_crash_revision} did not crash')
-    progression_task_output.crash_revision = int(max_revision)
-    return uworker_io.UworkerOutput(
+    progression_task_output.crash_revision = int(min_revision)
```
letitz, line 505: Should be `max_revision`?

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -592,20 +596,20 @@ def utask_main(uworker_input):
...
 
 
-def utask_postprocess(output):
+def utask_postprocess(output: uworker_msg_pb2.Output):
   """Trusted: Cleans up after a uworker execute_task, writing anything needed to
   the db."""
   testcase = data_handler.get_testcase_by_id(output.uworker_input.testcase_id)
   _maybe_clear_progression_last_min_max_metadata(testcase, output)
-  if output.progression_task_output:
+  if output.HasField("progression_task_output"):
```
letitz, line 604: This is checked twice, consider extracting it:

```py
task_output = None
if output.HasField("progression_task_output"):
  task_output = output.progression_task_output

if task_output:
  _update_issue_metadata(...)

...

if task_output and task_output.crash_on_latest:
  crash_on_latest(output)
  return
```

src/clusterfuzz/_internal/bot/testcase_manager.py
```diff
@@ -1196,13 +1196,14 @@ def check_for_bad_build(job_type, crash_revision):
...
         snapshot=process_handler.get_runtime_snapshot())
-  return uworker_io.BuildData(
+  return uworker_msg_pb2.BuildData(
       is_bad_build=is_bad_build,
       should_ignore_crash_result=crash_result.should_ignore(),
       build_run_console_output=build_run_console_output)
 
 
-def update_build_metadata(job_type, crash_revision, build_data):
+def update_build_metadata(job_type, crash_revision,
```
letitz, line 1205:
```suggestion
def update_build_metadata(job_type: str , crash_revision: int,
```

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
```diff
@@ -74,85 +66,16 @@ def upload_uworker_input(uworker_input, gcs_path):
...
-        deserialize_proto_field(element, None, is_input)
-        for element in initial_field_value
-    ]
-  elif isinstance(field_value, message.Message):
-    # This must come last! Otherwise it subsumes more specific types.
-    field_value = proto_to_deserialized_msg_object(field_value, is_input)
-  return field_value
-
-
 def deserialize_uworker_input(serialized_uworker_input):
```
letitz, line 69: We can add type annotations now. This goes for other functions below too.

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
```diff
@@ -219,329 +139,5 @@ def download_and_deserialize_uworker_output(output_url: str):
...
-  """Deserializes uworker's execute output for postprocessing. Returns a dict
-  that can be passed as kwargs to postprocess. Changes made db entities that
-  were modified during the untrusted portion of the task will be done to those
-  entities here."""
-  # Deserialize the proto.
-  uworker_output_proto = uworker_msg_pb2.Output()
-  uworker_output_proto.ParseFromString(uworker_output_str)
-  uworker_output = proto_to_deserialized_msg_object(
-      uworker_output_proto, is_input=False)
+  uworker_output.uworker_input.CopyFrom(uworker_input)  # pylint: disable=no-member
```
letitz, line 142: For example, why does the linter complain here?

src/clusterfuzz/_internal/bot/tasks/utasks/variant_task.py
```diff
@@ -111,34 +114,31 @@ def utask_main(uworker_input):
...
-    return uworker_io.UworkerOutput(
+    return uworker_msg_pb2.Output(
         error_type=uworker_msg_pb2.ErrorType.VARIANT_BUILD_SETUP,  # pylint: disable=no-member
     )
 
   # Disable gestures if we're running on a different platform from that of
   # the original test case.
-  use_gestures = (
-      uworker_input.testcase.platform == environment.platform().lower())
+  use_gestures = (testcase.platform == environment.platform().lower())
```
letitz, line 129:
```suggestion
  use_gestures = testcase.platform == environment.platform().lower()
```

src/clusterfuzz/_internal/tests/core/metrics/fuzzer_stats_test.py
```diff
@@ -252,10 +252,18 @@ def test_fuzz_task_upload_job_run_stats(self):
...
+    crashes = fuzz_task.convert_groups_to_crashes(groups)
+    job_run_crashes = []
+    for crash in crashes:
+      job_run_crashes.append({
+          'is_new': crash.is_new,
+          'count': crash.count,
+          'crash_type': crash.crash_type,
+          'crash_state': crash.crash_state,
+          'security_flag': crash.security_flag,
+      })
```
letitz, line 264: This is the third time this code is repeated, maybe extract it to a helper somewhere, maybe in `uworker_io`?

letitz (changes requested):
Looking good overall! Code still looks quite readable. Biggest comments I have are around linter disable statements, both `no-member` and `protected-access`. The latter  is easy to fix with a couple helper functions in `uworker_io`. The former is more concerning to me.

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
alhijazi: since we're reading it from an environment variable, it's a string, so I need to make sure it's properly converted to int.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -281,19 +284,20 @@ def _get_and_update_issue_metadata(testcase):
...
-                                     revision,
-                                     progression_task_output,
-                                     update_metadata=False):
+def _testcase_reproduces_in_revision(
+    testcase: data_types.Testcase,
+    testcase_file_path: str,
+    job_type: str,
+    revision: int,
+    progression_task_output: uworker_msg_pb2.ProgressionTaskOutput,
+    update_metadata=False):
```
letitz, line 293:
```suggestion
    update_metadata: bool = False):
```

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
letitz: I'd need @jonathanmetzman or @oliverchang to weigh in here. Maybe it's intentional, because in this branch we found that the target did *not* crash at `min_revision`?

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
letitz: Even better!

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
```diff
@@ -74,88 +70,21 @@ def upload_uworker_input(uworker_input, gcs_path):
...
-def serialize_uworker_input(uworker_input):
+def serialize_uworker_input(uworker_input: uworker_msg_pb2.Input) -> bytes:
   """Serializes and returns |uworker_input| as JSON. Can handle ndb entities."""
-  return uworker_input.serialize()
+  return uworker_input.SerializeToString()
 
 
-def serialize_and_upload_uworker_input(uworker_input) -> str:
+def serialize_and_upload_uworker_input(
+    uworker_input: uworker_msg_pb2.Input) -> str:
```
letitz, line 87:
```suggestion
    uworker_input: uworker_msg_pb2.Input) -> Tuple[str, str]:
```

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
```diff
@@ -219,334 +150,23 @@ def download_and_deserialize_uworker_output(output_url: str):
...
-    field.CopyFrom(entity_proto)
 
+from typing import Type
+from typing import TypeVar
 
-class SetupInput(UworkerInput):
-  """Input for setup.update_fuzzer_and_data_bundle and setup.setup_testcase in
-  uworker_main."""
-  PROTO_CLS = uworker_msg_pb2.SetupInput
+T = TypeVar('T', bound=ndb.Model)
```
letitz, line 165: I think this is overkill for now, because we do not have type-checking, so we can remove it. If you want to keep it, you'll want to move the imports to the top of the file.

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
```diff
@@ -219,334 +150,23 @@ def download_and_deserialize_uworker_output(output_url: str):
...
-  def __init__(self, testcase=None, error_type=None, **kwargs):
-    self.testcase = testcase
-    self.error_type = error_type
-    for key, value in kwargs.items():
-      setattr(self, key, value)
+def model_from_protobuf(pb2_model: entity_pb2.Entity, model_type: Type[T]) -> T:
+  """Helper function to get an entity from a proto."""
+  entity = model._entity_from_protobuf(pb2_model)  # pylint: disable=protected-access
+  assert isinstance(entity, model_type)
+  return entity
```
letitz, line 172:
```suggestion
def model_from_protobuf(entity_proto: entity_pb2.Entity, model_type: Type[T]) -> T:
  """Converts `entity_proto` to the `ndb.Model` of type `model_type` it encodes.
  
  Raises:
    AssertionError: if `entity_proto` does not encode a model of type `model_type`
  """
  entity = model._entity_from_protobuf(entity_proto)  # pylint: disable=protected-access
  assert isinstance(entity, model_type)
  return entity
```

src/clusterfuzz/_internal/bot/tasks/setup.py
```diff
@@ -227,12 +227,15 @@ def preprocess_setup_testcase(testcase, fuzzer_override=None):
...
+    setup_input = uworker_msg_pb2.SetupInput()
   setup_input.testcase_download_url = get_signed_testcase_download_url(testcase)
   return setup_input
 
 
-def setup_testcase(testcase, job_type, setup_input, metadata=None):
+def setup_testcase(testcase: data_types.Testcase,
+                   job_type: str,
+                   setup_input: uworker_msg_pb2.SetupInput,
+                   metadata=None):
```
letitz, line 238:
```suggestion
                   metadata: Optional[data_types.TestcaseUploadMetadata] = None):
```

src/clusterfuzz/_internal/bot/tasks/utasks/variant_task.py
```diff
@@ -111,34 +114,30 @@ def utask_main(uworker_input):
...
+        error_type=uworker_msg_pb2.ErrorType.UNHANDLED)
 
   # Check if we have an application path. If not, our build failed to setup
   # correctly.
   if not build_manager.check_app_path():
-    return uworker_io.UworkerOutput(
-        error_type=uworker_msg_pb2.ErrorType.VARIANT_BUILD_SETUP,  # pylint: disable=no-member
-    )
+    return uworker_msg_pb2.Output(
+        error_type=uworker_msg_pb2.ErrorType.VARIANT_BUILD_SETUP,)
```
letitz, line 124:
```suggestion
        error_type=uworker_msg_pb2.ErrorType.VARIANT_BUILD_SETUP)
```

letitz (commented):
A few nits and one substantive comment around `pylint-protobuf`, see `.pylintrc`.

.pylintrc
```diff
@@ -5,6 +5,8 @@ enable=
   indexing-exception,
   old-raise-syntax,
 
+load-plugins=pylint_protobuf
```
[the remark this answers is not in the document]
alhijazi: Thanks, I'll try integrating with it then!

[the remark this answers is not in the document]
letitz: No strong feelings here, sounds fine to me.

src/clusterfuzz/_internal/protos/uworker_msg.proto
```diff
@@ -16,15 +16,6 @@ syntax = "proto3";
...
 
-message Json {
-  string serialized = 1;
-}
-
-message UworkerEntityWrapper {
-  google.datastore.v1.Entity entity = 1;
-  Json changed = 2;
-}
-
```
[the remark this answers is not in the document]
letitz: Ah, that probably requires disentangling some uworker_io code to surgically extract json and entity conversion code. Don't sweat it if it's at all difficult!

src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
```diff
@@ -450,26 +452,19 @@ def handle_build_setup_error(output):
     uworker_msg_pb2.ErrorType.UNHANDLED,
 ] + setup.HANDLED_ERRORS
 
-# pylint: enable=no-member
-
 
 def _update_testcase(output):
   """Updates the testcase using the info passed from utask_main."""
-  if output.analyze_task_output is None:
+  if not output.HasField("analyze_task_output"):
```
jonathanmetzman, line 458: single quotes

src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
```diff
@@ -450,26 +452,19 @@ def handle_build_setup_error(output):
...
 
-  if analyze_task_output.absolute_path:
-    testcase.absolute_path = analyze_task_output.absolute_path
-
-  if analyze_task_output.minimized_arguments:
-    testcase.minimized_arguments = analyze_task_output.minimized_arguments
-
-  if analyze_task_output.crash_stacktrace:
-    testcase.crash_stacktrace = analyze_task_output.crash_stacktrace
+  testcase.absolute_path = analyze_task_output.absolute_path
```
jonathanmetzman, line 465: eh what if they are unset.

src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
```diff
@@ -478,10 +473,14 @@ def _update_testcase(output):
     testcase.crash_state = analyze_task_output.crash_state
     testcase.security_flag = analyze_task_output.security_flag
     if testcase.security_flag:
-      testcase.security_severity = analyze_task_output.security_severity
+      if analyze_task_output.HasField("security_severity"):
```
jonathanmetzman, line 476: single quotes and ugh

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -37,13 +37,14 @@
...
 
 
-def _maybe_clear_progression_last_min_max_metadata(testcase, uworker_output):
+def _maybe_clear_progression_last_min_max_metadata(
+    testcase: data_types.Testcase, uworker_output: uworker_msg_pb2.Output):
   """Clears last_progression_min and last_progression_max when
   clear_min_max_metadata is set to True"""
-  task_output = uworker_output.progression_task_output
-  if task_output is None:
+  if not uworker_output.HasField("progression_task_output"):
```
jonathanmetzman, line 44: single quotes

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -53,17 +54,21 @@ def _maybe_clear_progression_last_min_max_metadata(testcase, uworker_output):
 def _save_current_fixed_range_indices(testcase, uworker_output):
   """Save current fixed range indices in case we die in middle of task."""
   task_output = uworker_output.progression_task_output
+  last_progression_min = None
+  last_progression_max = None
+
+  if task_output.HasField("last_progression_min"):
```
jonathanmetzman, line 60: single quotes

src/clusterfuzz/_internal/protos/uworker_msg.proto
```diff
@@ -170,7 +166,7 @@ message ProgressionTaskOutput{
   // When set to True, we delete the metadata related to: last_progression_min
   // and last_progression_max from the testcase on postprocess.
   optional bool clear_min_max_metadata = 9;
-  optional Json issue_metadata = 10;
+  map<string, string> issue_metadata = 10;
```
jonathanmetzman, line 169: Is the value always a string? And not great that we can have legal json ({1: 1}) that wont serialize here.

src/clusterfuzz/_internal/protos/uworker_msg.proto
```diff
@@ -118,22 +108,28 @@ message AnalyzeTaskOutput {
...
   optional int64 testcases_executed = 6;
-  optional Json job_run_crashes = 7;
-  optional string fully_qualified_fuzzer_name = 8;
+  repeated CrashInfo job_run_crashes = 7;
   optional int32 new_targets_count = 9;
 }
 
 message MinimizeTaskOutput {
-  optional Json last_crash_result_dict = 1;
+  map<string, string> last_crash_result_dict = 1;
```
jonathanmetzman, line 131: address is a field here that seems to be an int based on some of your other files, so not sure we can say this is always a string.
Maybe we should leave a json field and have helper methods in uworker_io (name them copy_dict_from_json_field and save_dict_to_json_field to make clear that updating the dict wont work without calling again).

jonathanmetzman:
@oliverchang I think we should probably adopt this, what do you think?

oliverchang:
> @oliverchang I think we should probably adopt this, what do you think?

+1 this should simplify things greatly overall!

src/clusterfuzz/_internal/protos/uworker_msg.proto
alhijazi: this field is only populated with the return of `_extract_crash_result`
crash_address is a string as defined in CrashInfo class here:
https://github.com/google/clusterfuzz/blob/6877dd76c50cc5b25807f91d8534a31f125197e1/src/clusterfuzz/stacktraces/__init__.py#L34

src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/regression_task_test.py
```diff
@@ -392,14 +381,14 @@ def test_error_testcase_setup(self):
     """
     testcase = test_utils.create_generic_testcase()
 
-    output = uworker_io.UworkerOutput(
-        uworker_input=uworker_io.UworkerInput(
+    output = uworker_msg_pb2.Output(
+        uworker_input=uworker_msg_pb2.Input(
```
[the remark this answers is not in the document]
alhijazi: I agree here, I'll make sure we assign on construction when possible. Also worth noting that the linter complains if you try to assign a non-assignable field.
Example linter output:
```
| src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py:556:6: E5905: Field "Output.progression_task_output" does not support assignment (protobuf-no-assignment)
```

src/clusterfuzz/_internal/tests/core/bot/tasks/commands_test.py
```diff
@@ -204,8 +205,9 @@ def test_run_command_already_running_expired(self):
         time=datetime.datetime(1970, 1, 1),
         status='started').put()
 
-    self.mock.progression_utask_preprocess.return_value = uworker_io.UworkerInput(
-        job_type='job', testcase_id='123', uworker_env={})
+    self.mock.progression_utask_preprocess.return_value = uworker_msg_pb2.Input(
+        job_type='job', testcase_id='123')
+    self.mock.progression_utask_main.return_value = uworker_msg_pb2.Output()
```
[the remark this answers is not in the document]
alhijazi: Turns out it's not needed.

src/clusterfuzz/_internal/bot/tasks/utasks/variant_task.py
```diff
@@ -177,18 +173,18 @@ def utask_main(uworker_input):
...
+  variant_task_output = uworker_msg_pb2.VariantTaskOutput()
   variant_task_output.status = status
   variant_task_output.revision = int(revision)
   if crash_type is not None:
     variant_task_output.crash_type = crash_type
   if crash_state is not None:
     variant_task_output.crash_state = crash_state
-  variant_task_output.security_flag = security_flag
-  variant_task_output.is_similar = is_similar
+  variant_task_output.security_flag = bool(security_flag)
```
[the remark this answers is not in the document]
alhijazi: The linter complained about this

src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
alhijazi: These values are set only after `setup_testcase_and_build` has not resulted in errors.

At that point:
1. `testcase.absolute_path` is set to the value of `testcase_file_path` which is not None.
2. `testcase.minimized_arguments` is not None
3. `testcase.crash_stacktrace` is not None.

src/clusterfuzz/_internal/crash_analysis/stack_parsing/stack_parser.py
```diff
@@ -147,18 +147,18 @@ def to_proto(self):
     """Convert StackFrame to process_state.proto format for upload to crash/."""
```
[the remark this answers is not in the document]
alhijazi: Yes, pylint-protobuf complained.

src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/corpus_pruning_task_test.py
```diff
@@ -202,8 +201,10 @@ def test_preprocess(self):
     self.assertEqual(uworker_input.job_type, job_type)
     self.assertEqual(uworker_input.fuzzer_name, fuzzer_name)
     fuzz_target = data_handler.get_fuzz_target(fuzzer_name)
-    self.assertEqual(uworker_input.corpus_pruning_task_input.fuzz_target,
-                     fuzz_target)
+    self.assertEqual(
+        uworker_io.model_from_protobuf(
+            uworker_input.corpus_pruning_task_input.fuzz_target,
+            data_types.FuzzTarget), fuzz_target)
```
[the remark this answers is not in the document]
alhijazi: I think it's not that annoying and rather informative on what we intend to get.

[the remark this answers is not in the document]
alhijazi: also helpful to quickly find the class definition when reading the code.

alhijazi:
> First pass. Looks good except:
> 
> 1. JSON fields might not be handled correctly as maps.
     From what I encountered, I didn't see the need for anything more than a map<string,string>. For more complex types, we can create messages. I don't see the added value of JSON fields, maybe you can give me an example here.
> 2. CopyFrom/extend/update and HasField are all pretty bad IMO Maybe we can use a helper method get_field that returns None if unset. Also protoplus seems to offer assignment similar to uworker_io https://proto-plus-python.readthedocs.io/en/latest/messages.html#assigning-to-fields but I think it has some of the same issues my code had with references vs copies :-)
I eliminated all usages of `extend()`, most of `CopyFrom()` and most of `update()` by setting the fields directly at construction.
`HasField` is not used a lot. I think it's still a useful concept provided by protobuf and I don't see much value in hiding it. I would like the reader of the code to know that this is a special case that needs special handling.
That being said, we could probably remove the need to use `HasField` for scalar fields if (by design) we ensured `None`, `0`, and `''` had the same significance. It's a bit too late for that in some cases as this will require updating some database.
Regarding proto-plus offering assignment to fields, I think it's misleading to the reader of the code who would be expecting a reference while it's actually a copy.

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -20,6 +20,8 @@
 import random
 import re
 import time
+from typing import Any
```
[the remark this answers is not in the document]
alhijazi: it turns out the formatted puts them back on a line each.

jonathanmetzman (approved):
LGTM

PR #3538, opened 2023-11-24
[symbolize] Split symbolize task to trusted/untrusted parts

alhijazi:


src/clusterfuzz/_internal/bot/tasks/utasks/symbolize_task.py
```diff
@@ -59,18 +62,30 @@ def utask_main(uworker_input):
...
-    all_errors = uworker_handle_errors.get_all_handled_errors()
-    error.uworker_input.module_name = __name__
-    uworker_handle_errors.handle(error, all_errors)
-    return None
+    return error
 
   # Initialize variables.
   build_fail_wait = environment.get_value('FAIL_WAIT')
 
+  #TODO(alhijazi): check with @metzman if the blobstore is read-only and
```
oliverchang, line 87: It's not. We'll need to provide this ahead of time (unless we're handling this already somehow). 

Perhaps we can just inline in the actual stacktrace into `Testcase` in the proto message (if it's actually on the blob store) in some common setup routine for all tasks to make this easier.

WDYT @jonathanmetzman  ?

src/clusterfuzz/_internal/protos/uworker_msg.proto
```diff
@@ -89,7 +89,14 @@ message Input {
 }
 
 message SymbolizeTaskOutput {
- // TODO(metzman): Fill this out.
+  optional int64 build_fail_wait = 1;
```
oliverchang, line 92: We don't need to pass this around. Just read this from the environment variable on the tworker.

src/clusterfuzz/_internal/bot/tasks/utasks/symbolize_task.py
```diff
@@ -298,5 +295,50 @@ def get_symbolized_stacktraces(testcase_file_path, testcase,
...
+  testcase.crash_address = symbolize_task_output.crash_address
+  testcase.crash_state = symbolize_task_output.crash_state
+  testcase.crash_stacktrace = symbolize_task_output.crash_stacktrace
+
+  if not symbolize_task_output.symbolized:
+    data_handler.update_testcase_comment(
+        testcase, data_types.TaskState.ERROR,
+        'Unable to reproduce crash, skipping '
+        'stacktrace update')
+    # TODO(alhijazi): I am conserving previous behaviour, but shouldn't we
```
oliverchang, line 324: I believe this is to prevent retries of symbolization in these cases. This flag is used to control whether or not we want to attempt creating a symbolize task at all (e.g. https://github.com/google/clusterfuzz/blob/41a011f9cba88bdbefcae7bf258aea55cf05fc6f/src/clusterfuzz/_internal/bot/tasks/task_creation.py#L105). This is also the only place this flag is used.

oliverchang (commented):
Thanks!!

src/clusterfuzz/_internal/bot/tasks/utasks/symbolize_task.py
alhijazi: Noted, thanks! 
Will remove the TODO.

src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/symbolize_task_test.py
```diff
@@ -0,0 +1,209 @@
...
+    helpers.patch_environ(self)
+    helpers.patch(self, [
+        'clusterfuzz._internal.bot.tasks.setup.preprocess_setup_testcase',
+        'clusterfuzz._internal.build_management.build_manager.has_symbolized_builds'
+    ])
+    self.mock.preprocess_setup_testcase.return_value = uworker_io.SetupInput()
+    self.testcase = test_utils.create_generic_testcase()
+    self.testcase.fixed = 'Yes'
+    self.testcase.put()
+    # os.environ['JOB_NAME'] = 'symbolize'
```
jonathanmetzman, line 40: Can you take this out?

jonathanmetzman (approved):
LGTM

PR #3542, opened 2023-11-28
Fix setting OS custom field and add test

rmistry:


rmistry:
PTAL @oliverchang @jonathanmetzman 
CC @micahbales

src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
```diff
@@ -394,7 +394,13 @@ def save(self, new_comment=None, notify=True):
       if priority:
         self._data['issueState']['priority'] = priority
       if os:
-        self._data['issueState'][_CHROMIUM_OS_CUSTOM_FIELD_ID] = os
+        # TODO(rmistry): _update_issue will need to handle this as well.
```
rmistry, line 397: @oliverchang is this correct? Could the platform change via _update_issue ?

oliverchang: I think it can change in some circumstances, so yes.

oliverchang (approved):

PR #3545, opened 2023-11-29
Update llvm-symbolizer and libstdc++.so library

svasudevprasad:
The llvm-symbolizer was last updated in 2020. It seems to be causing issues with binaries built using later versions of clang. The symbolizer gives [??:?] result for filenames. Hence updated the llvm-symbolizer from llvm package 17.0.5 and libstdc++ to 6.0.32


jonatgoogle:
@oliverchang Could you please review this at your convenience? Thank you

oliverchang:
Thanks for contributing this! There's some subtle things here that are required to make this work properly though (e.g. needing to overwrite rpaths). 

I've instead opened https://github.com/google/clusterfuzz/pull/3552, which pulls our latest binaries from OSS-Fuzz (clang 15). 

Is clang 15 recent enough?

svasudevprasad:
Even with llvm-symbolizer v15.0.4, the issue was observed

oliverchang:
> Thank you for your inputs !
> 
> Even with llvm-symbolizer v15.0.4, the issue was observed during local testing.

OK, I've updated my PR to v18. I'll hold off on deploying this though until next week.

svasudevprasad:
Thank you for the update !

svasudevprasad:
#3552 - Can you please check if the library libstdc++.so also needs to be updated for LLVM 18 ?

oliverchang:
> #3552 - Can you please check if the library libstdc++.so also needs to be updated for LLVM 18 ?

It's no longer a dependency for that particular binary.

svasudevprasad:
Superseded by #3552.

PR #3546, opened 2023-11-29
Include 'chromium-testing' in testcase chromium checks.

oliverchang:
This should make e.g. blame task (which calls Predator) work.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/task_creation.py
+++ b/src/clusterfuzz/_internal/bot/tasks/task_creation.py
@@ -76,7 +76,7 @@ def create_blame_task_if_needed(testcase):
     return
 
   # Blame is only applicable to chromium project, otherwise bail out.
-  if testcase.project_name != 'chromium':
+  if testcase.is_chromium():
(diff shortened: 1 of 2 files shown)


jonathanmetzman (approved):
lgtm

src/clusterfuzz/_internal/bot/tasks/task_creation.py
```diff
@@ -76,7 +76,7 @@ def create_blame_task_if_needed(testcase):
     return
 
   # Blame is only applicable to chromium project, otherwise bail out.
-  if testcase.project_name != 'chromium':
+  if testcase.is_chromium():
```
rmistry, line 79: I was trying to understand more how predator integration works and I noticed this.

I think this needs to be `if not testcase.is_chromium()`

This may have turned off predator for chromium completely?

rmistry: https://github.com/google/clusterfuzz/pull/3569

PR #3547, opened 2023-11-29
Set OS through the update issue flow + more fixes

rmistry:
Fixes include using camelCase in API calls for consistency and using a list for customFields.


rmistry:
PTAL @oliverchang 
CC @micahbales

jonathanmetzman:
Deploying now. SHould be done in ~30 minutes

PR #3550, opened 2023-11-30
Experiment running multiple tasks in the same batch job as seperate tasks

jonathanmetzman:
This adds behavior to execute "combined tasks" combined tasks will be pub/sub messages for utasks that are executed under the final model (preprocess, uworker_main, postprocess on different bots). When dealing with a combined tasks a bot will execute preprocess of all the bots, take the input_urls returned by preprocess, and start a batch job where uworker_main will be run on different VMs as different tasks.
Also support handling the case here where tasks will need different VMs (not something a single job can support) by creating multiple jobs.
To support this, use a queue for utasks.
Also remove Pub/Sub filtering as this cannot be done dynamically.


src/clusterfuzz/_internal/google_cloud_utils/batch.py
```diff
@@ -125,18 +171,21 @@ def create_uworker_main_batch_job(module_name, cf_job, input_download_url):
...
-def _create_job(create_request):
+@retry.wrap(
+    retries=3,
+    delay=2,
+    function='google_cloud_utils.batch._send_create_job_request')
+def _send_create_job_request(create_request):
   return _batch_client().create_job(create_request)
 
 
 def get_spec(full_module_name, job_name):
```
oliverchang, line 187: nit: this wasn't introduced in this PR, but this function is a bit confusing. It seems to imply that `full_module_name`, `job_name` form a unique key for a spec, when it's not. I would suggest instead to have this function take in the exact parameters it needs (i.e. platform, and preemptible/non-preemptible). This is also a lot more modular.

oliverchang (commented):
Thanks! 

My one gripe with this overall is that it's tying our architecture very strongly to the specific limitations and details of Cloud Batch, which makes it more annoying if we want to swap out the underlying platform in the future, and adds a fair bit of complexity to our code. 

Does this end up being much different to us just provisioning time-limited GCE VMs on demand when we need to create tasks?  Or are there other considerations here?

jonathanmetzman:
> > > Thanks!
> > > My one gripe with this overall is that it's tying our architecture very strongly to the specific limitations and details of Cloud Batch, which makes it more annoying if we want to swap out the underlying platform in the future, and adds a fair bit of complexity to our code.
> > 
> > 
> > I mostly agree with this. But I guess it could make sense to (pre)process tasks a few at a time once all the tasks are migrated, we probably will get some slight improvements in performance.
> > > Does this end up being much different to us just provisioning time-limited GCE VMs on demand when we need to create tasks? Or are there other considerations here?
> > 
> > 
> > This shouldn't offer any improvement over that strategy in terms of throughput.
> 
> Right, but this (creating raw instances) could probably be a lot simpler in terms of implementation?

Should I explore creating raw instances then? Happy to throw this in the trash if raw instances (which will support Windows now) works.
I can write a design of this or even an implementation.

jonathanmetzman:
Also rmeove

> > > > > Thanks!
> > > > > My one gripe with this overall is that it's tying our architecture very strongly to the specific limitations and details of Cloud Batch, which makes it more annoying if we want to swap out the underlying platform in the future, and adds a fair bit of complexity to our code.
> > > > 
> > > > 
> > > > I mostly agree with this. But I guess it could make sense to (pre)process tasks a few at a time once all the tasks are migrated, we probably will get some slight improvements in performance.
> > > > > Does this end up being much different to us just provisioning time-limited GCE VMs on demand when we need to create tasks? Or are there other considerations here?
> > > > 
> > > > 
> > > > This shouldn't offer any improvement over that strategy in terms of throughput.
> > > 
> > > 
> > > Right, but this (creating raw instances) could probably be a lot simpler in terms of implementation?
> > 
> > 
> > Should I explore creating raw instances then? Happy to throw this in the trash if raw instances (which will support Windows now) works. I can write a design of this or even an implementation.
> 
> I certainly think this is worth a shot.

Ah actually I now remember. Batch gives us a queue. We don't have a queue with raw instances. I have to deal with this in FuzzBench by retrying which is bad IMO

jonathanmetzman:
> Also rmeove
> 
> > > > > > Thanks!
> > > > > > My one gripe with this overall is that it's tying our architecture very strongly to the specific limitations and details of Cloud Batch, which makes it more annoying if we want to swap out the underlying platform in the future, and adds a fair bit of complexity to our code.
> > > > > 
> > > > > 
> > > > > I mostly agree with this. But I guess it could make sense to (pre)process tasks a few at a time once all the tasks are migrated, we probably will get some slight improvements in performance.
> > > > > > Does this end up being much different to us just provisioning time-limited GCE VMs on demand when we need to create tasks? Or are there other considerations here?
> > > > > 
> > > > > 
> > > > > This shouldn't offer any improvement over that strategy in terms of throughput.
> > > > 
> > > > 
> > > > Right, but this (creating raw instances) could probably be a lot simpler in terms of implementation?
> > > 
> > > 
> > > Should I explore creating raw instances then? Happy to throw this in the trash if raw instances (which will support Windows now) works. I can write a design of this or even an implementation.
> > 
> > 
> > I certainly think this is worth a shot.
> 
> Ah actually I now remember. Batch gives us a queue. We don't have a queue with raw instances. I have to deal with this in FuzzBench by retrying which is bad IMO

The batch queue has priorities which is nice. But I suppose a queue (to deal with rate limiting in API calls to GCE) will be provided by existing queue system. Basically instance creation/preprocess will fail and then preprocess will be done again.

jonathanmetzman:
Using raw instances has a problem I did not realize.
There's an option to delete the instance on termination. Unfortunately, this only applies if the VM has reached its timelimit or is preempted. If it shuts itself down it is not deleted. So postprocess would have to delete the instance.

PR #3552, opened 2023-11-30
Upgrade llvm-symbolizer to llvm 18.

oliverchang:
This is taken from a chromium clang build.


jonathanmetzman (approved):
lgtm

jonathanmetzman:
Maybe deploy to chrome first so we can see if this affects things badly

oliverchang:
> Maybe deploy to chrome first so we can see if this affects things badly

Most things should be just using the llvm-symbolizer provided in the build zip, so hopefully this shouldn't be too bad.

I also upgraded this to 18 per https://github.com/google/clusterfuzz/pull/3545#issuecomment-1834790760. I can try this next week.

jonathanmetzman:
Do you think this will be fine given that oss-fuzz is still on clang-15?

davfu1:
> Maybe deploy to chrome first so we can see if this affects things badly

Android might be a good first deployment! This change is expected to fix a bug where our testcases have been missing line numbers for the last few months, due to the older symbolizer returning [??:?]:

https://clusterfuzz.corp.google.com/testcase-detail/5353756926476288
https://clusterfuzz.corp.google.com/testcase-detail/4840571869921280

PR #3554, opened 2023-12-01
Always do the substitution mapping of a label if it exists

rmistry:
[Config change](https://team-review.git.corp.google.com/c/gosst/clusterfuzz-config/+/2097721) to add hotlist mappings was submitted and deployed. But the `Security_Impact-*` labels are still not getting mapped to the specified hotlists.

I traced through the code and found that the substitution mapping only happens if it contains one of the known hardcoded wildcards- https://github.com/google/clusterfuzz/blob/6877dd76c50cc5b25807f91d8534a31f125197e1/src/clusterfuzz/_internal/issue_management/issue_filer.py#L244

This was unexpected because we have specified a couple more labels in the substitutions part of the config assuming they would be replaced by their hotlist IDs which will not work. Eg: `External-Fuzzer-Contribution` and `reward-topanel`.

This PR will always return the substitution mapping of a label if it exists. If no substitution mapping exists then the original label will be returned.

### DIFF
--- a/src/clusterfuzz/_internal/issue_management/issue_filer.py
+++ b/src/clusterfuzz/_internal/issue_management/issue_filer.py
@@ -246,7 +246,7 @@ def apply_substitutions(policy, label, testcase, security_severity=None):
       ]
 
   # No match found. Return mapped value if it exists else the original label
-  # will be returned..
+  # will be returned.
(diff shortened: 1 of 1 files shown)


rmistry:
PTAL @oliverchang , @jonathanmetzman 
CC @micahbales

PR #3555, opened 2023-12-04
Check substitution mapping before adding Chromium-related labels directly to issue.labels

rmistry:
Check substitution mapping before adding Chromium-related labels directly to issue.labels + add chromium-testing to two more places


rmistry:
PTAL @jonathanmetzman , @oliverchang 
CC @micahbales

PR #3558, opened 2023-12-10
Batch creation of tasks into a single batch job

jonathanmetzman:
Previously cloud batch had trouble keeping up with our variant tasks because may variant tasks are created at once and batch could only process 1 job every 2 seconds.
This change alleviates this problem by combining tasks that are created at the same time into a single batch job (each CF task will be a batch task that is part of the same batch job, instead of each CF task getting its own batch job)


jonathanmetzman:
@oliverchang What do you think of this approach?

src/clusterfuzz/_internal/base/tasks.py
```diff
@@ -346,6 +346,10 @@ def get_task():
   return task
 
 
+def get_payload(command, argument, job):
```
oliverchang, line 349: nit: s/get/construct/ is more descriptive. Also add a docstring.

src/clusterfuzz/_internal/bot/tasks/task_creation.py
```diff
@@ -13,17 +13,40 @@
...
 from clusterfuzz._internal.metrics import logs
 from clusterfuzz._internal.system import environment
 
 
-def mark_unreproducible_if_flaky(testcase, task_name, potentially_flaky):
+@dataclasses.dataclass
+class Task:
+  """Representing a Task."""
+
+  def __init__(self,
```
oliverchang, line 36: Per https://docs.python.org/3/library/dataclasses.html,

We don't need to write our own `__init__`.

src/clusterfuzz/_internal/google_cloud_utils/batch.py
```diff
@@ -71,48 +77,87 @@ def get_job_name():
   return 'j-' + str(uuid.uuid4()).lower()
 
 
-def create_uworker_main_batch_job(module_name, cf_job, input_download_url):
+class BatchTask:
```
oliverchang, line 80: docstring

src/clusterfuzz/_internal/google_cloud_utils/batch.py
```diff
@@ -71,48 +77,87 @@ def get_job_name():
...
+def _create_job(spec, input_urls):
+  """Creates and starts a batch job from |spec| that executes all tasks."""
+  task_group = batch.TaskGroup()
+  task_group.task_count = len(input_urls)
+  assert task_group.task_count < MAX_CONCURRENT_VMS_PER_JOB
+  task_environments = [
+      batch.Environment(variables={'UWORKER_INPUT_DOWNLOAD_URL': input_url})
+      for input_url in input_urls
+  ]
+  task_group.task_environments = task_environments
```
oliverchang, line 153: This may not end up mattering much, but I assume the individual tasks can't steal input URLs from other tasks?

oliverchang (approved):

src/clusterfuzz/_internal/google_cloud_utils/batch.py
jonathanmetzman: Yes they can't steal them.

jonathanmetzman: Although, the nice thing here is that the tasks are all based on the same testcase. So there's nothing for any of them to steal. Defense in depth!

PR #3559, opened 2023-12-11
Fix broken serialization

jonathanmetzman:


src/clusterfuzz/_internal/bot/tasks/utasks/variant_task.py
```diff
@@ -74,16 +74,18 @@ def utask_preprocess(testcase_id, job_type, uworker_env):
...
       testcase=uworker_io.entity_to_protobuf(testcase),
-      testcase_upload_metadata=uworker_io.entity_to_protobuf(
-          testcase_upload_metadata),
+      testcase_upload_metadata=),
       uworker_env=uworker_env,
       testcase_id=testcase_id,
       setup_input=setup_input,
   )
+  if testcase_upload_metadata:
+    uworker_input.testcase_upload_metadata = uworker_io.entity_to_protobuf(
```
letitz, line 87: I would have expected a `CopyFrom()` to be necessary here.

jonathanmetzman: You are correct: https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/tasks/utasks/variant_task.py#L86

PR #3563, opened 2023-12-11
Fix appending of newlines

backes:
If the crash_state holds multiple lines, we already append a '\n' also after the last line. If we use 'NULL' for the state, we do not append a '\n'. Hence also do not append the newline if the fuzz target is being used as the crash state.

### DIFF
--- a/src/clusterfuzz/stacktraces/__init__.py
+++ b/src/clusterfuzz/stacktraces/__init__.py
@@ -343,7 +343,8 @@ class StackParser:
       if state.crash_type not in ['V8 correctness failure']:
         state.crash_state = filter_addresses_and_numbers(state.crash_state)
 
-      # Truncate each line in the crash state to avoid excessive length.
+      # Truncate each line in the crash state to avoid excessive length. This
+      # also ensures a trailing '\n' in the crash state.
       original_crash_state = state.crash_state
       state.crash_state = ''
       for line in original_crash_state.splitlines():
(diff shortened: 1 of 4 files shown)


backes:
@oliverchang @jonathanmetzman PTAL
This is just a little refactoring, the actual change I wanted to make is #3564.

See https://groups.google.com/a/google.com/g/clusterfuzz-dev/c/Nq6xWPpGl-c/m/ruP1IVp5AAAJ for context.

src/clusterfuzz/stacktraces/__init__.py
```diff
@@ -366,11 +367,6 @@ def filter_crash_parameters(self, state):
     ]:
       state.crash_state = self.fuzz_target
 
-    # Add a trailing \n if it does not exist in crash state.
-    if (state.crash_state and state.crash_state != 'NULL' and
```
jonathanmetzman, line 370: Hmmm...but I wonder what cases were previously covered that no longer are.

backes: This code only had an effect if we replace the state either by the process name or the fuzz target. In both cases I don't see why we should append a newline. Note that for "NULL" we also (explicitly) did not do it.

backes: (note how ~20 lines above we already make sure that the state ends in a newline for all other cases)

backes:
Is your question resolved? If so, can you please merge this, so we can then rebase and merge the follow-up (#3564)?

oliverchang:
Thanks for this change! Note that changing this can have some repercussions though -- we may end up filing a bunch of duplicate bugs because the unique key for them (i.e. crash state) has changed.

Are there good reasons for this change, beyond consistency and aesthetics?

backes:
This was only for consistency. I don't see a reason to append a newline to the process name or fuzz target.

As all other states already are terminated by a newline, I don't think that this will affect too many reports, but I am also fine with living with this inconsistency and closing this PR.

PR #3564, opened 2023-12-11
Use fuzz target name instead of "NULL"

backes:
If no process name is available but we have a fuzz target name, use that
instead of "NULL".
This hopefully reduces the amount of "Abrt · NULL" reports we are
seeing.

### DIFF
--- a/src/clusterfuzz/_internal/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
+++ b/src/clusterfuzz/_internal/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py
@@ -1764,6 +1764,21 @@ class StackAnalyzerTestcase(unittest.TestCase):
                                   expected_state, expected_stacktrace,
                                   expected_security_flag)
 
+  def test_sanitizer_signal_abrt_fuzz_target(self):
+    """Same as above, but this time we specify a fuzz target which should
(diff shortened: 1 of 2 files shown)


backes:
@oliverchang @jonathanmetzman PTAL
Note that this is based on #3563; the difference is in this commit: https://github.com/google/clusterfuzz/commit/88fd71b15b04d5b753a826f5b9f8ad024903fbc8

See https://groups.google.com/a/google.com/g/clusterfuzz-dev/c/Nq6xWPpGl-c/m/ruP1IVp5AAAJ for context.

jonathanmetzman (approved):
LGTM

oliverchang:
This LGTM, other than my comment here in the other PR: https://github.com/google/clusterfuzz/pull/3563#issuecomment-1861821694

backes:
I rebased this now to not depend on #3563. The only difference is that the expected state now has a trailing newline.

Feel free to merge if this looks good.

oliverchang (approved):

backes:
Are there failing tests? Seems like I don't have access to the results.

oliverchang:
> Are there failing tests? Seems like I don't have access to the results.

Looks like flake. Trying again.

backes:
Thanks a lot! Looking forward to more helpful CF reports, and the respective fixes :D

PR #3566, opened 2023-12-11
Lock pipfile

jonathanmetzman:
I think a pretty big Windows breakage happened because the Pipfile wasn't actually locked, causing the init script to fail.


alhijazi (commented):
LGTM

PR #3567, opened 2023-12-11
[utask] Handle non-utask aware callers 

jonathanmetzman:
We must be able to handle cases where the caller doesn't preprocess first.


jonathanmetzman:
TBR to fix issue.

PR #3571, opened 2023-12-11
Undo combined batch execution

jonathanmetzman:


jonathanmetzman:
Just coding mistakes on my end. I reverted out of caution, but I think I can get it working.

PR #3583, opened 2023-12-12
Add support to run deprecated testcases

svasudevprasad:
If the testcase fails to match the current job, a check is done to see if the testcase has been deprecated. If it is deprecated check if it can run on the current device


src/clusterfuzz/_internal/bot/tasks/commands.py
```diff
@@ -323,10 +323,15 @@ def process_command_impl(task_name,
         # more details: https://github.com/google/clusterfuzz/issues/3347
         if (task_name != 'variant' and testcase_platform_id and
             not utils.fields_match(testcase_platform_id, current_platform_id)):
-          logs.log(
+
+          # Check if the devices or version is deprecated.
```
marktefftech, line 327: nit:

```
# Check if the device or branch is deprecated.
# If it is deprecated, try to execute on an updated platform.
```

src/clusterfuzz/_internal/platforms/android/constants.py
```diff
@@ -50,3 +50,20 @@
 }
 
 RELEASE_CONFIGURATION = 'next'
+
+DEPRECATED_DEVICE_LIST = [
```
marktefftech, line 54: Unfortunately these codenames on the left are internal, so they cannot be committed into open-source ClusterFuzz. We may have to find another way to determine if a testcase is considered deprecated due to the device.

harishdm: Mark, all these codenames are public at https://developers.google.com/android/drivers.

src/clusterfuzz/_internal/system/environment.py
```diff
@@ -1134,3 +1134,30 @@ def wrapper(*args, **kwargs):
     return None
 
   return wrapper
+
+
+def is_tescase_deprecated(platform_id=None):
+  """Whether or not the device is deprecated."""
```
marktefftech, line 1140:
```
"""Whether or not the device or branch is deprecated."""
```

src/clusterfuzz/_internal/bot/tasks/commands.py
```diff
@@ -323,10 +323,15 @@ def process_command_impl(task_name,
...
+          # If it is deprecated, try to find an alternate/replacement job
+          if not (environment.is_testcase_deprecated(testcase_platform_id)
+            and environment.can_testcase_run_on_platform(testcase_platform_id, current_platform_id)):
+            logs.log(
               'Testcase %d platform (%s) does not match with ours (%s), exiting'
               % (testcase.key.id(), testcase_platform_id, current_platform_id))
-          tasks.add_task(
+            tasks.add_task(
```
marktefftech, line 334: Before re-adding the task via `task.add_task`, don't we need to update the testcase information? If not, the testcase will look exactly the same the next time it comes through the queue, and it will never update.

So this:

`android:sunfish_hwasan:master` 

Needs to become something like this:

`android:<current_bot_codename>:<current_bot_branch>`

And then we should try to execute the updated testcase.

src/clusterfuzz/_internal/platforms/android/util.py
```diff
@@ -37,3 +37,44 @@ def get_local_path(device_path):
...
+    return False
+
+  # Check if devices is deprecated
+  if codename_fields[0] in android.constants.DEPRECATED_DEVICE_LIST:
+    return True
+
+  # Check if version is deprecated
+  # Currently only "main" or "m" is active.
+  # All other branches including "master" have been deprecated
+  if (platform_id_fields[2] <= 'v' or platform_id_fields[2] == 'master') and platform_id_fields[2] != 'm'
```
marktefftech, line 61: I don't think this code will run, can you compare strings using the `<=` operator?

src/clusterfuzz/_internal/system/environment.py
```diff
@@ -1134,3 +1134,30 @@ def wrapper(*args, **kwargs):
...
+    return False
+  if is_android(platform_id.upper()):
+    # FIXME: Handle this import in a cleaner way.
+    from clusterfuzz._internal.platforms import android
+
+    return android.util.is_tescase_deprecated(platform_id)
+
+  return False
+
+def can_testcase_run_on_platform(testcase_platform_id, current_platform_id):
```
marktefftech, line 1151: both of these functions `is_tescase_deprecated` and `can_testcase_run_on_platform` should first check if the environment is not Android, and immediately return `False` since this only applies to Android.

marktefftech (changes requested):
Overall I think you are on the right track. 

The only big piece missing is at some point in the code, we need to actually try to modify  / run the testcase itself on the new device or branch.

Currently I think this code just checks whether we *could* run this testcase on the current device, but it only re-adds the testcase back into the queue, so it will continue in a forever loop still.

src/clusterfuzz/_internal/platforms/android/constants.py
marktefftech: Thanks Harish you are correct. This comment can be ignored.

src/clusterfuzz/_internal/bot/tasks/commands.py
svasudevprasad: Done

src/clusterfuzz/_internal/platforms/android/constants.py
svasudevprasad: Closing this comment as per above

src/clusterfuzz/_internal/system/environment.py
svasudevprasad: Done

src/clusterfuzz/_internal/system/environment.py
svasudevprasad: Done. Checks added accordingly.

src/clusterfuzz/_internal/bot/tasks/commands.py
svasudevprasad: This is not required. Once the check passes, clusterfuzz downloads and runs the test case on the latest device. Screenshot attached for reference. 
image

src/clusterfuzz/_internal/platforms/android/util.py
svasudevprasad: This is working as intended.

src/clusterfuzz/_internal/bot/tasks/commands.py
svasudevprasad: Additionally, clusterfuzz is able to recognize this and add a log. 
image

However, currently there is no method to update the values or metadata in the original testcase and hence it will contain the value of the platform where it was observed initially. Hope it is fine.

src/clusterfuzz/_internal/platforms/android/util.py
```diff
@@ -37,3 +37,42 @@ def get_local_path(device_path):
...
+
+def is_testcase_deprecated(platform_id=None):
+  """Whether or not the Android device is deprecated."""
+
+  # Platform ID for Android is of the form - |android:{codename}_{sanitizer}:{build_version}|
+  platform_id_fields = platform_id.split(':')
+  if len(platform_id_fields) != 3:
+    return False
+
+  codename_fields = platform_id_fields[1].split('_')
```
marktefftech, line 50: Just for your consideration, I believe there are cases when no sanitizer is used. 

So the `.split('_')` method will be a no-operation. I do not think this is a problem.

src/clusterfuzz/_internal/platforms/android/util.py
```diff
@@ -37,3 +37,42 @@ def get_local_path(device_path):
...
+
+  # Check if devices is deprecated
```
marktefftech, line 52: nit: `device` (not plural)

src/clusterfuzz/_internal/platforms/android/util.py
```diff
@@ -37,3 +37,42 @@ def get_local_path(device_path):
...
+  if codename_fields[0] in android.constants.DEPRECATED_DEVICE_LIST:
+    return True
+
+  # Check if version is deprecated
```
marktefftech, line 56: ` Check if *branch* is deprecated`

src/clusterfuzz/_internal/platforms/android/util.py
```diff
@@ -37,3 +37,42 @@ def get_local_path(device_path):
...
+  # Currently only "main" or "m" is active.
```
marktefftech, line 57: nit: remove period from comments

src/clusterfuzz/_internal/platforms/android/util.py
```diff
@@ -37,3 +37,42 @@ def get_local_path(device_path):
...
+  # All other branches including "master" have been deprecated
+  if (platform_id_fields[2] <= 'v' or platform_id_fields[2] == 'master') and platform_id_fields[2] != 'm':
+    return True
```
marktefftech, line 60: This is a minor change, but I would prefer the following. Your decision though:

```
branch = platform_id_fields[2]
if (branch <= 'v' or branch == 'master') and branch != 'm':
    return True
```

src/clusterfuzz/_internal/platforms/android/util.py
```diff
@@ -37,3 +37,42 @@ def get_local_path(device_path):
...
+
+  # Platform ID for Android is of the form - |android:{codename}_{sanitizer}:{build_version}|
+  current_platform_id_fields = current_platform_id.split(':')
+  if len(current_platform_id_fields) != 3:
+    return False
+
+  # Deprecated testcase should run on any latest device and on main
+  # So ignore device information and check for current version
+  # If the current version is 'm' or main, run the test case
+  if current_platform_id_fields[2] == 'm':
```
marktefftech, line 75: The comment indicates you may need to add a check for main?

```
if current_platform_id_fields[2] in ['m', 'main]:
  ...
```

src/clusterfuzz/_internal/system/environment.py
```diff
@@ -1141,3 +1141,36 @@ def wrapper(*args, **kwargs):
...
+  """Whether or not the testcase can run on the current platform."""
+  if not is_android(testcase_platform_id.upper()) or not is_android(current_platform_id.upper()):
+    return False
+
+  if is_android_cuttlefish(testcase_platform_id.upper()) and is_android_cuttlefish(current_platform_id.upper()):
+    return True
+
+  if is_android(testcase_platform_id.upper()):
+    # FIXME: Handle this import in a cleaner way.
```
marktefftech, line 1171: I assume you noticed this import done similarly elsewhere in the file, but is it possible to just import this at the top of file in the standard way?

src/clusterfuzz/_internal/platforms/android/util.py
svasudevprasad: Marking this as resolved

src/clusterfuzz/_internal/platforms/android/util.py
svasudevprasad: Yes correct. This is done to remove the sanitizer part (if present) so that the device part is obtained for comparison.

src/clusterfuzz/_internal/platforms/android/util.py
svasudevprasad: Done

src/clusterfuzz/_internal/platforms/android/util.py
svasudevprasad: Done

src/clusterfuzz/_internal/platforms/android/util.py
svasudevprasad: Done

src/clusterfuzz/_internal/system/environment.py
svasudevprasad: Yes.  I took reference from line 215 in the same file - https://github.com/google/clusterfuzz/pull/3583/files#diff-ec10fe9c1dd83b9edcd526a0d57d8580bb8d764d4eab9c3c3e528d041e7a6487L215

I was not sure if it is possible to import this the standard way and did not want to break anything. Hence followed the existing method.

src/clusterfuzz/_internal/platforms/android/util.py
svasudevprasad: As per https://github.com/google/clusterfuzz/blob/7b4e4827866ae59b79e7481ed9ea0491efa2a63f/src/clusterfuzz/_internal/platforms/android/settings.py#L57, 'm' means main. Hence a check for 'm' seems sufficient as "main" will never be present.

marktefftech (approved):

marktefftech:
LGTM

@oliverchang or @jonathanmetzman , can you please take a look at your earliest convenience.

src/clusterfuzz/_internal/system/environment.py
```diff
@@ -1141,3 +1141,36 @@ def wrapper(*args, **kwargs):
...
 
   return wrapper
+
+
+def is_testcase_deprecated(platform_id=None):
+  """Whether or not the device or branch is deprecated."""
+  if not is_android(platform_id.upper()):
+    return False
+
+  if is_android_cuttlefish(platform_id.upper()):
```
jonathanmetzman, line 1151: This would be much cleaner as
```
if is_android and not is_android_cuttlefish(platform_id.upper()):
    from clusterfuzz._internal.platforms import android

    return android.util.is_testcase_deprecated(platform_id)
return False
```

src/clusterfuzz/_internal/system/environment.py
jonathanmetzman: Sorry I'm not sure this link shows this. Can you just import it at the top of the file instead.

src/clusterfuzz/_internal/platforms/android/util.py
```diff
@@ -37,3 +37,43 @@ def get_local_path(device_path):
   return os.path.join(
       root_directory,
       os.path.relpath(device_path, android.constants.DEVICE_FUZZING_DIR))
+
+
+def is_testcase_deprecated(platform_id=None):
+  """Whether or not the Android device is deprecated."""
+
+  # Platform ID for Android is of the form - |android:{codename}_{sanitizer}:{build_version}|
+  platform_id_fields = platform_id.split(':')
```
jonathanmetzman, line 46: hmmm...I'm not so happy with this string manipulation. seems fragile. WDYT @oliverchang

src/clusterfuzz/_internal/bot/tasks/commands.py
```diff
@@ -323,15 +323,28 @@ def process_command_impl(task_name,
...
               task_name,
               task_argument,
               job_name,
               wait_time=utils.random_number(1, TASK_RETRY_WAIT_LIMIT))
-          return None
+            return None
+
+          logs.log(
+              'Testcase %d platform (%s) can run on current platform (%s)'
```
jonathanmetzman, line 346: Use an f-string please.

src/clusterfuzz/_internal/bot/tasks/commands.py
```diff
@@ -323,15 +323,28 @@ def process_command_impl(task_name,
         # more details: https://github.com/google/clusterfuzz/issues/3347
         if (task_name != 'variant' and testcase_platform_id and
             not utils.fields_match(testcase_platform_id, current_platform_id)):
+
           logs.log(
+              'Testcase %d platform (%s) does not match with ours (%s), checking ...'
```
jonathanmetzman, line 328: f-string please.

src/clusterfuzz/_internal/system/environment.py
svasudevprasad: Done. Moved the import to the top and removed other duplicate imports

src/clusterfuzz/_internal/system/environment.py
svasudevprasad: Done

src/clusterfuzz/_internal/bot/tasks/commands.py
svasudevprasad: Done

src/clusterfuzz/_internal/bot/tasks/commands.py
svasudevprasad: Done

src/clusterfuzz/_internal/system/environment.py
```diff
@@ -24,6 +24,7 @@
 import yaml
 
 from clusterfuzz._internal import fuzzing
+from clusterfuzz._internal.platforms import android
```
jonathanmetzman, line 27: Hmm... so I meant to import at the top `if is_android`.
@marktefftech Do you know why we gated this import in the first place?

marktefftech: Unfortunately no, the imports have been this way since Oliver committed the original Android implementation in 2019

src/clusterfuzz/_internal/platforms/android/util.py
marktefftech: Friendly ping on this @oliverchang

src/clusterfuzz/_internal/bot/tasks/commands.py
```diff
@@ -317,21 +317,34 @@ def process_command_impl(task_name,
...
-              'Testcase %d platform (%s) does not match with ours (%s), exiting'
-              % (testcase.key.id(), testcase_platform_id, current_platform_id))
-          tasks.add_task(
-              task_name,
-              task_argument,
-              job_name,
-              wait_time=utils.random_number(1, TASK_RETRY_WAIT_LIMIT))
-          return None
+
+          logs.log(f'Testcase {testcase_id} platform {testcase_platform_id}\
```
jonathanmetzman, line 328: nit: generally you can use implicit string joining instead of backslashes

src/clusterfuzz/_internal/bot/tasks/commands.py
```diff
@@ -317,21 +317,34 @@ def process_command_impl(task_name,
...
+               does not match with ours {current_platform_id}, checking ...')
+
+          # Check if the device or branch is deprecated.
+          # If it is deprecated, try to execute on an updated platform.
+          if not (environment.is_testcase_deprecated(testcase_platform_id) and
+                  environment.can_testcase_run_on_platform(
+                      testcase_platform_id, current_platform_id)):
+            logs.log('Testcase %d platform (%s) does not match with ours\
```
jonathanmetzman, line 336: nit: Just use f-strings.

jonathanmetzman (approved):
LGTM

PR #3584, opened 2023-12-12
Use the "Component Tags" custom field to store additional components in google_issue_tracker

rmistry:
Google Issue Tracker has support for only one component per issue. The Chrome-on-Buganizer team has added a "Component Tags" custom field with which each issue can have multiple components (similar to behavior in Monorail).

This required refactoring of how components are handled in google_issue_tracker.

More context is in b/315853316


rmistry:
CC @micahbales

jonathanmetzman:
Something is wrong with our testing.

PR #3586, opened 2023-12-13
Add support for the ReleaseBlock custom field in google_issue_tracker and check substitution mappings

rmistry:
Add support for the ReleaseBlock custom field in google_issue_tracker. The "ReleaseBlock-Beta" label is added [here](https://github.com/google/clusterfuzz/blob/d439a7a9f8ce0ed8f79b33506a0833f1868e8f90/src/clusterfuzz/_internal/cron/cleanup.py#L1009) it will be now converted into a custom field for use in google_issue_tracker.

Also, check substitution mappings before adding Predator labels.


rmistry:
CC @micahbales

rmistry:
Note that the All-Tests failure looks unrelated to this PR

PR #3594, opened 2023-12-20
Fix is_production call

jonathanmetzman:


marktefftech:
A little late but LGTM... Thanks for addressing this.

PR #3598, opened 2024-01-02
[batch] Don't use external ips and try variant again.

jonathanmetzman:


jonathanmetzman:
@oliverchang PTAL
TBR

jonathanmetzman:
This should help fix the most recent crop of batch failures (along with the more than 4X disk), though the fact that batch falls over when given too many jobs for current quotas isn't good.

PR #3600, opened 2024-01-02
Migrate Minimize task

alhijazi:


src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
```diff
@@ -404,13 +406,11 @@ def utask_main(uworker_input):
...
   if environment.is_libfuzzer_job():
-    do_libfuzzer_minimization(testcase, testcase_file_path)
-    return None
+    return do_libfuzzer_minimization(testcase, testcase_file_path)
 
   if environment.is_engine_fuzzer_job():
     # TODO(ochang): More robust check for engine minimization support.
-    _skip_minimization(testcase, 'Engine does not support minimization.')
-    return None
+    return uworker_msg_pb2.Output()
```
jonathanmetzman, line 413: Why not just return None?

src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
```diff
@@ -503,34 +497,44 @@ def utask_main(uworker_input):
...
       testcase.security_severity = severity_analyzer.get_security_severity(
           testcase.crash_type, data_handler.get_stacktrace(testcase),
           uworker_input.job_type, bool(gestures))
+      minimize_task_output.security_severity_updated = True
+      if testcase.security_severity is not None:
+        minimize_task_output.security_severity = testcase.security_severity
 
     testcase.gestures = gestures
-    testcase.set_metadata('minimization_phase', MinimizationPhase.MAIN_FILE)
+    minimize_task_output.gestures.clear()
```
jonathanmetzman, line 511: Why do you clear it first?

src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
```diff
@@ -555,52 +563,115 @@ def utask_main(uworker_input):
...
+      testcase.security_severity = None
+
+  if minimize_task_output.HasField('minimization_phase'):
+    testcase.set_metadata('minimization_phase',
+                          minimize_task_output.minimization_phase)
+
+  if minimize_task_output.flaky_stack:
+    testcase.flaky_stack = minimize_task_output.flaky_stack
+
+  if minimize_task_output.HasField('minimized_arguments'):
```
jonathanmetzman, line 644: ooo don't like HasField.

src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
```diff
@@ -555,52 +563,115 @@ def utask_main(uworker_input):
...
+
+
 def utask_postprocess(output):
   """Postprocess in a trusted bot."""
+  update_testcase(output)
   if output.error_type != uworker_msg_pb2.ErrorType.NO_ERROR:
     uworker_handle_errors.handle(output, HANDLED_ERRORS)
     return
 
+  if environment.is_engine_fuzzer_job():
```
oliverchang, line 669: Unsure if this condition is correct here. 

The previous implementation only reaches this if it's an engine **AND** not libfuzzer (because there is a previous check for libFuzzer). 

We could also just move this to pre-process, and skip this step completely in such cases (i.e. a non libfuzzer engine case).

src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
oliverchang: are there any alternatives here? if not, seems OK to rely on these.

oliverchang (commented):
thank you!!! sorry for the very delayed review here.

src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
alhijazi: an abundance of caution. Not really necessary as this is the only place we set the field.

src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
alhijazi: I want to only update a testcase field if it was explicitly updated during `utask_main`. I don't see any other way than this for the moment, but I'm open to suggestions.

src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
alhijazi: Indeed thanks for catching that! I moved the check and the handling to preprocess.

src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
```diff
@@ -555,48 +570,105 @@ def utask_main(uworker_input):
...
+    uworker_msg_pb2.ErrorType.MINIMIZE_DEADLINE_EXCEEDED,
+    uworker_msg_pb2.ErrorType.MINIMIZE_DEADLINE_EXCEEDED_IN_MAIN_FILE_PHASE,
+    uworker_msg_pb2.ErrorType.LIBFUZZER_MINIMIZATION_UNREPRODUCIBLE,
+    uworker_msg_pb2.ErrorType.LIBFUZZER_MINIMIZATION_FAILED,
     uworker_msg_pb2.ErrorType.UNHANDLED
 ]
 
 
+def update_testcase(output: uworker_msg_pb2.Output):
+  """Updates the tescase using the values passed from utask_main. This done at
```
jonathanmetzman, line 627: "This done" Please fix this comment.

jonathanmetzman (commented):
LGTM

jonathanmetzman:
Going to merge this tomorrow.

PR #3601, opened 2024-01-02
Run variant on batch

jonathanmetzman:
Run variant on batch again.
Fix issue with subnetworks.

### DIFF
--- a/src/clusterfuzz/_internal/google_cloud_utils/batch.py
+++ b/src/clusterfuzz/_internal/google_cloud_utils/batch.py
@@ -131,17 +131,27 @@ def _get_task_spec(batch_workload_spec):
 
 def _get_allocation_policy(spec):
   """Returns the allocation policy for a BatchWorkloadSpec."""
-  instance_policy = batch.AllocationPolicy.InstancePolicy()
   disk = batch.AllocationPolicy.Disk()
   disk.image = 'batch-cos'
   disk.size_gb = spec.disk_size_gb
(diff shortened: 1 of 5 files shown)


src/clusterfuzz/_internal/google_cloud_utils/batch.py
```diff
@@ -131,17 +131,33 @@ def _get_task_spec(batch_workload_spec):
...
   disk.type = spec.disk_type
+  instance_policy = batch.AllocationPolicy.InstancePolicy()
   instance_policy.boot_disk = disk
   instance_policy.machine_type = spec.machine_type
   instances = batch.AllocationPolicy.InstancePolicyOrTemplate()
   instances.policy = instance_policy
+
+  # Don't use external ip addresses which use quota, cost money, and are
+  # unnecessary.
+  network_interface = batch.AllocationPolicy.NetworkInterface()
```
oliverchang, line 146: I think for Chromium we kept them because some testcases may need to access the external internet?

oliverchang (commented):
LGTM

src/clusterfuzz/_internal/google_cloud_utils/batch.py
jonathanmetzman: Hmmm...well I'll have to reopen my quota request :-(

I thought not having external ip means no one can access your bot from external networks, but you can still can reach the internet through a NAT (like a home router).

PR #3615, opened 2024-01-04
Improve batch

jonathanmetzman:
1. Don't retry tasks.
2. Set longer timeout so variant can complete.

### DIFF
--- a/src/clusterfuzz/_internal/google_cloud_utils/batch.py
+++ b/src/clusterfuzz/_internal/google_cloud_utils/batch.py
@@ -31,9 +31,8 @@ from . import credentials
 
 _local = threading.local()
 
-MAX_DURATION = '3600s'
-RETRY_COUNT = 1
-TASK_COUNT = 1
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
Also log when we are saving time on preprocess.

src/clusterfuzz/_internal/google_cloud_utils/batch.py
```diff
@@ -31,9 +31,8 @@
 
 _local = threading.local()
 
-MAX_DURATION = '3600s'
-RETRY_COUNT = 1
-TASK_COUNT = 1
+MAX_DURATION = f'{int(60 * 60 * 2.5)}s'
+RETRY_COUNT = 0
```
oliverchang, line 35: Any reason for removing the retry?

jonathanmetzman: It doesn't work at all. It tries to restart a docker container (ClusterFuzz's) that cannot actually be restarted.

Issue #3618, opened 2024-01-08
Mitigate too many jobs in Fuzzer


oliverchang:
For some of our deployments, we now have too many jobs and we're running into index entry limits on the `Fuzzer` entity.

PR #3619, opened 2024-01-08
Don't index jobs for Fuzzer entities.

oliverchang:
Some of our deployments are hitting index limits, so we can't do this anymore.

Generally this shouldn't impact performance very much, because there are not too many Fuzzer entities in all of our deployments. 

For the existing usages of this index, just loop through and filter out the Fuzzer entities that match.

Three usages are impacted:

- Jobs page: Tested and verified that this doesn't impact performance much (due to pagination).
- update_mappings_for_job: Whenever a Fuzzer/Job association is modified. Will monitor the performance of our project setup cron jobs.
- external_users: This only impacts fuzzer stats, testcase upload, which are both one-off loads that shouldn't impact performance much. 

For #3618.

### DIFF
--- a/src/appengine/handlers/jobs.py
+++ b/src/appengine/handlers/jobs.py
@@ -56,8 +56,8 @@ def _job_to_dict(job):
   result = job.to_dict()
   result['id'] = job.key.id()
   # Adding all associated fuzzers with each job.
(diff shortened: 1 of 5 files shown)


Alan32Liu (approved):

jonathanmetzman:
LGTM

jonathanmetzman:
Thanks!

PR #3633, opened 2024-01-11
Fix appengine python3.7 deprecation

jonathanmetzman:


oliverchang (commented):
thanks!

jonathanmetzman:
I guess this isn't really a fix as it's only for test cdoe.

PR #3635, opened 2024-01-11
[DO NOT MERGE] Experiment with receiving tasks and then merging

jonathanmetzman:


src/clusterfuzz/_internal/base/tasks.py
```diff
@@ -304,15 +344,42 @@ def get_task():
     logs.log_error('Failed to get any fuzzing tasks. This should not happen.')
     time.sleep(TASK_EXCEPTION_WAIT_INTERVAL)
 
+    return None
   return task
 
 
+def get_utasks_queue():
+  return 'utasks'
```
oliverchang, line 352: Is there a way to avoid this additional queue/indirection? 

Is the main reason why we have a dedicated one just because not all tasks work with batch yet? 

If we need this control on a task by task basis, perhaps we should start annotating tasks with attributes encoding the task name (which can be filtered on in Pub/Sub) ,

oliverchang: Alternatively, we can just use this queue as an intermediate step, and remove it later once everything is migrated.

oliverchang: Thinking more, I suppose the other benefit of having this queue is we can batch fuzz utasks this way as well with the same mechanism. I'm fine with this approach in that case, as I can't think of any other way to reduce the number of requests/Batch Jobs we make.

oliverchang, line 352: add a TODO to create a terraform configuration for this.

src/clusterfuzz/_internal/base/tasks.py
```diff
@@ -188,9 +228,13 @@ def get_regular_task(queue=None):
   application_id = utils.get_application_id()
   while True:
```
oliverchang, line 229: nit: not really relevant for this PR, but can we add max_messages=1 to line 230 here to be more explicit about the expected behaviour?

src/clusterfuzz/_internal/base/tasks.py
```diff
@@ -188,9 +228,13 @@ def get_regular_task(queue=None):
...
-
     if not messages:
       return None
+    task, retry = handle_single_message(messages)
+    if task:
+      return task
+    if not retry:
+      return None
 
     try:
```
oliverchang, line 239: can we delete all the lines that follow here? These seem all handled in `handle_single_message`.

PR #3637, opened 2024-01-11
Allow deployment to subset of bots

jonathanmetzman:


src/local/butler/deploy.py
```diff
@@ -93,6 +93,7 @@ def _deploy_app_prod(project,
                      package_zip_paths,
                      deploy_appengine=True):
   """Deploy app in production."""
+  test_deployment = bool(environment.get_value('USE_TEST_DEPLOYMENT'))
   if deploy_appengine:
```
oliverchang, line 97: should we force this to be false if we have test deployment set? 

It seems pretty dangerous to deploy to app engine otherwise. also similarly, with our other parts of deployment (k8s, bigquery, queues etc).

src/local/butler/deploy.py
```diff
@@ -93,6 +93,7 @@ def _deploy_app_prod(project,
                      package_zip_paths,
                      deploy_appengine=True):
   """Deploy app in production."""
+  test_deployment = bool(environment.get_value('USE_TEST_DEPLOYMENT'))
```
oliverchang, line 96: would prefer an explicit CLI argument here for this.

oliverchang (commented):
nice idea! just some comments on the approach

PR #3638, opened 2024-01-12
Fix incorrect consumption of .options files

aditya-wazir:
Ensures that changes made by set_sanitizer_options() are reflected in the configuration during the initial fuzzer run.

### DIFF
--- a/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py
@@ -269,8 +269,8 @@ class Engine(engine.Engine):
     Returns:
       A FuzzResult object.
     """
-    runner = libfuzzer.get_runner(target_path)
     libfuzzer.set_sanitizer_options(target_path, fuzz_options=options)
+    runner = libfuzzer.get_runner(target_path)
 
     # Directory to place new units.
     if options.merge_back_new_testcases:


marktefftech:
LGTM

jonathanmetzman:
1. Can you explain why this didn't work before and why it does work now? FYI get_runner does not create any processes, so it shouldn't matter when you set the environment.
2. Why do we not care about this happening in cleanse/minimize/prune etc? I think we should add this there.
3. Please add a test for this functionality.

aditya-wazir:
1.) Even with the .options file containing detect_leaks=0 flag, When executing the fuzzer using ClusterFuzz, a peculiar behavior is observed: during the initial run (fuzzing round 0) only, detect_leaks is set to 1. However, starting from the second iteration onward, the value of detect_leaks is consistently set to 0 . The root cause of this behavior is explained below:
   - In the engine.py file, within the fuzz(), the first runner is instantiated using get_runner(). This, in turn, returns an AndroidLibFuzzerRunner object. When the init() method of AndroidLibFuzzerRunner is invoked, it calls ProcessRunner.init() with default_args obtained from [_get_default_args](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py#L803)().
   - Within [_get_default_args](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py#L803)(), [get_sanitizer_options_for_display](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/system/environment.py#L329)() is utilized to retrieve the default ASAN_OPTIONS. At this point, the options file has not been read, so the default ASAN_OPTIONS contains detect_leaks=1. This default configuration is then passed to init(), where _default_args is assigned the default ASAN_OPTIONS.
   - Subsequently, after the get_runner() call, [set_sanitizer_options](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py#L1271)() is executed. This method internally reads the .options file, which contains flags such as { 'detect_leaks': '0'}. These flags override the default ASAN_OPTIONS, changing detect_leaks from 1 to 0. The updated ASAN_OPTIONS are then set internally using environment.py and [set_memory_tool_options](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/system/environment.py#L818)().
   - When runner.fuzz() is called, it eventually invokes the run() method. In run(), the command is assigned a value using [get_command](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/system/new_process.py#L267)(). This method appends _default_args of ProcessRunner to the command and returns it. Since, during the first iteration, ProcessRunner._default_args still contains the default ASAN_OPTIONS with detect_leaks=1, this value is present in the command. However, from the second run onward, the changes applied by [set_sanitizer_options](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py#L1271)() result in detect_leaks=0, and this updated configuration is reflected in the command. As clusterfuzz runs for default argument during initial run, it can report leaks bugs even if the options file has detect_leaks=0.
   - In order to address the problem, we have modified the calling order of [set_sanitizer_option](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py#L1271)() before initializing the running using the get_running() method in this CL. This ensures that changes made by [set_sanitizer_options](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py#L1271)() are reflected in the configuration during the initial fuzzer run.
   
2.) Following methods in engine.py are initializing runner and then subsequently calling [set_sanitizer_options](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py#L1271)()):
  - [fuzz](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py#L259)()
  -  [reproduce()](https://github.com/google/clusterfuzz/blob/8d52cf79e880aac660e91b5bc06943713e0acc10/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py#L375C7-L375C16)
  - [minimize_corpus()](https://github.com/google/clusterfuzz/blob/8d52cf79e880aac660e91b5bc06943713e0acc10/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py#L490C7-L490C22)
  - [minimize_testcase()](https://github.com/google/clusterfuzz/blob/8d52cf79e880aac660e91b5bc06943713e0acc10/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py#L538C7-L538C24)
  - [cleanse()](https://github.com/google/clusterfuzz/blob/8d52cf79e880aac660e91b5bc06943713e0acc10/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py#L573C7-L573C14)
Regarding concerns about other methods like  [reproduce()](https://github.com/google/clusterfuzz/blob/8d52cf79e880aac660e91b5bc06943713e0acc10/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py#L375C7-L375C16), [minimize_corpus()](https://github.com/google/clusterfuzz/blob/8d52cf79e880aac660e91b5bc06943713e0acc10/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py#L490C7-L490C22), [minimize_testcase()](https://github.com/google/clusterfuzz/blob/8d52cf79e880aac660e91b5bc06943713e0acc10/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py#L538C7-L538C24), [cleanse()](https://github.com/google/clusterfuzz/blob/8d52cf79e880aac660e91b5bc06943713e0acc10/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py#L573C7-L573C14), these methods are invoked after [fuzz](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py#L259)() in the execution sequence. With the bug fix applied in [fuzz](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py#L259)(), [set_sanitizer_options](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py#L1271)() is called before initializing the runner in [fuzz](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py#L259)(). This ensures that environment.py already contains the correct flags as per the .options file.
In  [reproduce()](https://github.com/google/clusterfuzz/blob/8d52cf79e880aac660e91b5bc06943713e0acc10/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py#L375C7-L375C16), [minimize_corpus()](https://github.com/google/clusterfuzz/blob/8d52cf79e880aac660e91b5bc06943713e0acc10/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py#L490C7-L490C22), [minimize_testcase()](https://github.com/google/clusterfuzz/blob/8d52cf79e880aac660e91b5bc06943713e0acc10/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py#L538C7-L538C24), [cleanse()](https://github.com/google/clusterfuzz/blob/8d52cf79e880aac660e91b5bc06943713e0acc10/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py#L573C7-L573C14), no default flags are being passed, as the default values have already been overridden and set in environment.py by the time these methods are executed. Therefore, the bug fix applied in [fuzz](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py#L259)() covers the subsequent methods in the execution flow, maintaining consistency in the configuration.

3.) We are currently working on writing the test for this fix.

marktefftech:
@aditya-wazir any update on this PR and the new test?

aditya-wazir:
Hi Mark ,
We are working on the unit test. We have designed and implemented the test. While testing, we are facing some issues that we are currently resolving. We will upload the unit test code in a couple of days

aditya-wazir:
Hi @marktefftech @jonathanmetzman ,
We analyzed the unit-test for our bug fix, and the following are our observations.
The crash is specific to the Android and hence it needs a device to run. Since tests run on the Linux cloud, when attempting to execute the fuzz() method results in an error since no device is connected. Hence we feel the unit-test for this is infeasible.

jonathanmetzman:
>The crash is specific to the Android and hence it needs a device to run. Since tests run on the Linux cloud, when attempting to execute the fuzz() method results in an error since no device is connected. Hence we feel the unit-test for this is infeasible.

Sorry, I don't think this is right. Almost all functionality in ClusterFuzz is for use with something that doesn't exist during testing, and they can be tested anyway. You can abstract away the functionality you need and write a test that ensures this (e.g. if you need two functions called in a certain order, you can write a test for this.

jonathanmetzman (approved):
LGTM
I'll merge this even though it lacks a test because it's not really adding code to maintain. If it breaks it breaks.

svasudevprasad:
> Can you resolve the merge conflicts please?

Done. Can you please check ?

PR #3639, opened 2024-01-12
Fix bug in bots with subqueue due to platform checking

marktefftech:


src/clusterfuzz/_internal/bot/tasks/commands.py
```diff
@@ -276,10 +276,12 @@ def process_command_impl(task_name,
       logs.log_error(error_string)
       raise errors.BadStateError(error_string)
 
-    # A misconfiguration led to this point. Clean up the job if necessary.
     job_queue_suffix = tasks.queue_suffix_for_platform(job.platform)
-    bot_queue_suffix = tasks.default_queue_suffix()
```
oliverchang, line 281: Would it make more sense to modify `default_queue_suffix` to handle this instead? I.e. extract the base platform there?

marktefftech: While this is a good suggestion, the current code may be best.

The `*_queue_suffix` comparison done here is checking the bot's "Base" queue, since it is intentionally more generic so this bot can run both generic and device-specific tasks.

For example,

```
job_queue_suffix = '-android'
bot_queue_suffix = '-android-pixel6'
```

The above code throws and error, even though we want the bot to be able to run both `android` and `pixel6` tasks. If we instead put this logic in `default_queue_suffix` there will be no error here, but this will cause problems elsewhere (`regular_queue` for example)

marktefftech: Changed the variable to `bot_base_queue_suffix` for clarity.

jonathanmetzman (approved):
LGTM

jonathanmetzman:
Will deploy this tomorrow

PR #3643, opened 2024-01-17
Make progression a UTask

jonathanmetzman:
Progression will be hard to combine tasks into jobs because there's so many of them that are started by a single instance.
Instead, just create each batch job. This should work according to my math because progression runs 3-6X less often than variant.


oliverchang (commented):
LGTM

PR #3645, opened 2024-01-18
Fix handling of flags by launch_cvd, add new flags

marktefftech:


src/clusterfuzz/_internal/platforms/android/adb.py
```diff
@@ -420,13 +420,13 @@ def start_cuttlefish_device(use_kernel=False):
   device_memory_mb = environment.get_value('DEVICE_MEMORY_MB',
                                            DEFAULT_DEVICE_MEMORY_MB)
   launch_cvd_command_line = (
-      f'{launch_cvd_path} -daemon -memory_mb {device_memory_mb} '
-      '-report_anonymous_usage_stats Y')
+      f'sudo {launch_cvd_path} --daemon --memory_mb={device_memory_mb} '
```
oliverchang, line 423: Is backwards compatibiltiy a concern here? i.e. would  we be running this on devices/configurations  where we  need the old cmdline style?

oliverchang (approved):

jonathanmetzman (approved):
LGTM.
@marktefftech Can you confirm this is OK, or that any gains are outweighed by the losses? I will deploy if you confirm.

src/clusterfuzz/_internal/platforms/android/adb.py
marktefftech: I can confirm that all of our cuttlefish instances (150 pre / 50 normal) launch via this command line, neither physical device nor hosts will reach this section of code. There should be no concerns of retro-compat.

After pair programming with Hamzeh@, we discovered that these flags were originally implemented incorrectly and have never been consumed since launch. The command in '{cmd}' succeeds, but the options are not consumed.

Changes in upstream cuttlefish code cause this implementation to fail, leading out current virtual device fleet in a complete stalemate.

marktefftech:
@jonathanmetzman thank you for the diligence. Please see my response to Oliver for greater detail, but the summary is that this has always been a faulty command.

Moreover, our Cuttlefish fleet has been out of commission due to upstream changes that pointed to this LOC. This change only affects the Android fleet and atm we have can only gain from fixing it.

PR #3647, opened 2024-01-18
Ensure that uworker_env only has string keys.

jonathanmetzman:
Otherwise we will get errors when adding it to to a proto. This bug pattern is very likely in ClusterFuzz because ClusterFuzz's env vars don't need to be strings (this seems like the wrong decision IMO), but their representation in proto must be strings since they must be a single type.

This fixes an error introduced here:
https://github.com/google/clusterfuzz/pull/3536
This fixes this bug: https://pantheon.corp.google.com/errors/detail/CPCzn83Hxtr1ag;service=;version=;filter=%5B%22%5C%22_collections_abc.py%5C%22%22%5D?e=-13802955&mods=logs_tg_prod&project=google.com:cluster-fuzz

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
@@ -21,6 +21,14 @@ from clusterfuzz._internal.system import environment
 
 
 def ensure_uworker_env_type_safety(uworker_env):
+  """Converts all values in |uworker_env| to str types.
+  ClusterFuzz parses env var values so that the type implied by the value
+  (which in every OS I've seen is a string), is the Python type of the value.
+  E.g. if "DO_BLAH=1" in the environment, environment.get_value('DO_BLAH') is 1,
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
CC @letitz @alhijazi This is relevant to our discussion about using protos directly or abstracting them.

jonathanmetzman:
I guess as long as ClusterFuzz allows any type for env var values, we probably cannot represent a uworker_env using a string->string dictionary?

src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
```diff
@@ -20,11 +20,17 @@
 from clusterfuzz._internal.system import environment
 
 
+def ensure_uworker_env_type_safety(uworker_env):
```
oliverchang, line 23: Can you add some docstring comments explaining why this is important?

oliverchang (approved):

marktefftech:
LGTM

PR #3649, opened 2024-01-22
Do not overwrite OOM with generic "Abrt"

backes:
In V8 fuzzers with ASan enabled we see crashes where after a detected out-of-memory we see an ASan "Abrt" report which overwrites the out-of-memory state. This is not helpful, we should classify such crashes as out-of-memory.

This also adds some more tests using existing test data but setting the `REPORT_OOMS_AND_HANGS` environment variable.


backes:
@oliverchang This is a follow-up, now that we are able to identify where our "Abrt - NULL" crashes were coming from (see https://github.com/google/clusterfuzz/pull/3564 and https://crbug.com/1519393).

backes:
@jonathanmetzman Are you available for a review?

jonathanmetzman:
LGTM
@oliverchang WDYT?

backes:
ping :)

oliverchang (approved):

PR #3651, opened 2024-01-23
Use signed URLs for corpus

jonathanmetzman:


jonathanmetzman:
This is somewhat early but I want feedback since it is such an invasive change.

src/clusterfuzz/_internal/google_cloud_utils/storage.py
```diff
@@ -1170,3 +1200,89 @@ def get_signed_download_url(remote_path, minutes=SIGNED_URL_EXPIRATION_MINUTES):
...
+
+
+def _error_tolerant_download_signed_url_to_file(url, path):
+  try:
+    return download_signed_url_to_file(url, path), url
+  except Exception:
+    return None
+
+
+def _error_tolerant_upload_signed_url(url, path):
```
oliverchang, line 1212: Maybe I'm missing this in this large PR, but who calls this?

src/clusterfuzz/_internal/fuzzing/corpus_manager.py
```diff
@@ -603,25 +360,228 @@ def get_regressions_corpus_gcs_url(self):
...
+  num_files = _count_corpus_files(directory)
+  logs.log(f'{num_files} corpus files downloaded.')
+  return True
+
+
+def fuzz_target_corpus_upload_files(corpus, filepaths):
+  results = storage.upload_signed_urls(corpus.corpus.upload_urls, filepaths)
+  # Make sure we don't reuse upload_urls.
+  urls_remaining = corpus.corpus.upload_urls[:len(results)]
+  del corpus.corpus.upload_urls[:]
```
oliverchang, line 566: Is there a `.clear()` method we can use here?

src/clusterfuzz/_internal/fuzzing/corpus_manager.py
```diff
@@ -603,25 +360,228 @@ def get_regressions_corpus_gcs_url(self):
...
+  if fuzz_target_corpus.HasField('regressions_corpus'):
+    regressions_dir = os.path.join(directory, 'regressions')
+    _sync_corpus_to_disk(fuzz_target_corpus.regressions_corpus, regressions_dir)
+
+  num_files = _count_corpus_files(directory)
+  logs.log(f'{num_files} corpus files downloaded.')
+  return True
+
+
+def fuzz_target_corpus_upload_files(corpus, filepaths):
```
oliverchang, line 562: docstrings?

src/clusterfuzz/_internal/fuzzing/corpus_manager.py
```diff
@@ -603,25 +360,228 @@ def get_regressions_corpus_gcs_url(self):
...
+  results = storage.download_signed_urls(corpus.corpus_urls, directory)
+  for filepath, upload_url in results:
+    corpus.filenames_to_delete_urls_mapping[filepath] = (
+        corpus.corpus_urls[upload_url])
+
+  # TODO(metzman): Add timeout and tolerance for missing URLs.
+  return results.count(None) < MAX_SYNC_ERRORS
+
+
+def fuzz_target_corpus_sync_to_disk(fuzz_target_corpus, directory) -> bool:
```
oliverchang, line 548: Rather than all these `fuzz_target_corpus_` prefixed functions, could we perhaps just create a lightweight Python class that wraps the proto and adds these as methods? I think that would greatly aid readability.

src/clusterfuzz/_internal/fuzzing/corpus_manager.py
```diff
@@ -603,25 +360,228 @@ def get_regressions_corpus_gcs_url(self):
...
+
+  num_files = _count_corpus_files(directory)
+  logs.log(f'{num_files} corpus files downloaded.')
+  return True
+
+
+def fuzz_target_corpus_upload_files(corpus, filepaths):
+  results = storage.upload_signed_urls(corpus.corpus.upload_urls, filepaths)
+  # Make sure we don't reuse upload_urls.
+  urls_remaining = corpus.corpus.upload_urls[:len(results)]
```
oliverchang, line 565: Can you add some comments to explain the overall logic here?

src/clusterfuzz/_internal/fuzzing/corpus_manager.py
```diff
@@ -603,25 +360,228 @@ def get_regressions_corpus_gcs_url(self):
...
+    regressions_dir = os.path.join(directory, 'regressions')
+    _sync_corpus_to_disk(fuzz_target_corpus.regressions_corpus, regressions_dir)
+
+  num_files = _count_corpus_files(directory)
+  logs.log(f'{num_files} corpus files downloaded.')
+  return True
+
+
+def fuzz_target_corpus_upload_files(corpus, filepaths):
+  results = storage.upload_signed_urls(corpus.corpus.upload_urls, filepaths)
```
oliverchang, line 563: What happens if len(filepaths) > len(upload_urls) ?

src/clusterfuzz/_internal/fuzzing/corpus_manager.py
oliverchang: It may also be better to wrap this consuming logic in another helper. 

e.g.

```python
def upload_and_consume_upload_urls(corpus.corpus.upload_urls, filepaths):
 ...
```

oliverchang (commented):
nice! do you have any performance numbers here? e.g. compared to gsutil?

src/clusterfuzz/_internal/fuzzing/corpus_manager.py
jonathanmetzman: Nope. this is the only method we can use sadly.

src/clusterfuzz/_internal/fuzzing/corpus_manager.py
```diff
@@ -603,25 +360,228 @@ def get_regressions_corpus_gcs_url(self):
...
+
+  return url
+
+
+def get_proto_corpus(bucket_name, bucket_path):
+  """Returns a proto representation of a corpus."""
+  gcs_url = _get_gcs_url(bucket_name, bucket_path)
+  corpus_urls = {}
+  # TODO(metzman): Allow this step to be skipped by trusted fuzzers.
+  for corpus_element_url in storage.list_blobs(gcs_url):
```
alhijazi, line 472: Not sure this would work. It seems to me that `list_blobs` only return names, not full path.

src/clusterfuzz/_internal/fuzzing/corpus_manager.py
```diff
@@ -603,25 +360,228 @@ def get_regressions_corpus_gcs_url(self):
...
+  Returns:
+    A string giving the GCS url.
+  """
+  backup_dir = gcs_url_for_backup_directory(backup_bucket_name, fuzzer_name,
+                                            project_qualified_target_name)
+  backup_file = str(date) + os.extsep + BACKUP_ARCHIVE_FORMAT
+  return f'{backup_dir.rstrip("/")}/{backup_file}'
+
+
+def backup_corpus(backup_bucket_name, corpus, directory):
```
alhijazi, line 376: `backup_corpus` is called within `utask_main`, it should use signed urls

src/clusterfuzz/_internal/google_cloud_utils/storage.py
```diff
@@ -152,6 +154,14 @@ def upload_signed_url(self, data, signed_url):
     """Uploads |data| to |signed_url|."""
     raise NotImplementedError
 
+  def sign_delete_url(self, remote_path, minutes=SIGNED_URL_EXPIRATION_MINUTES):
+    """Signs a DELETE URL for a remote file."""
+    raise NotImplementedError
+
+  def delete_signed_url(self, signed_url):
```
alhijazi, line 161: rename to `delete_resource` ?

src/clusterfuzz/_internal/google_cloud_utils/storage.py
jonathanmetzman: upload_signed_urls. We don't want to kill the uploading process if one thread exceptions.

jonathanmetzman:
> nice! do you have any performance numbers here? e.g. compared to gsutil?

For large corpora, it's about 25% slower. For smallish ones, it can go from, ~10 seconds to 22 seconds (I think a lot of the time is spent signing upload URLs that won't be used).

I've gotten it even faster than gsutil in the few tests I did when I use aiohttp and async, at the cost of making the code harder to understand.
Since the profiler works on 3.11 (https://github.com/google/clusterfuzz/issues/3738) I'll rerintroduce it and measure how long corpus operations take. If it's significant, I'll optimize it.

src/clusterfuzz/_internal/fuzzing/corpus_manager.py
jonathanmetzman: You are right.

src/clusterfuzz/_internal/google_cloud_utils/storage.py
jonathanmetzman: Ehh I think current message makes sense.

src/clusterfuzz/_internal/fuzzing/corpus_manager.py
jonathanmetzman: I'll leave this as a TODO.

PR #3656, opened 2024-01-23
Fix dependency issue

jonathanmetzman:
Don't depend on bot code in App Engine.

### DIFF
--- a/src/clusterfuzz/_internal/base/task_utils.py
+++ b/src/clusterfuzz/_internal/base/task_utils.py
@@ -1,4 +1,4 @@
-# Copyright 2023 Google LLC
+# Copyright 2024 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


src/clusterfuzz/_internal/base/task_utils.py
```diff
@@ -12,7 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Module providing utilities for utask users. This module should not depend on
-any other module in tasks to prevent circular imports."""
+any other module in tasks to prevent circular imports and issues with
```
letitz, line 15: Now that it's moved to `base/`, do we still need this warning? It seems to me that the directory structure implies this already.

letitz (approved):
LGTM with a single optional comment.

src/clusterfuzz/_internal/base/task_utils.py
jonathanmetzman: I don't think it's implied.

jonathanmetzman:
This faces the appengine 500 error we are seeing.

PR #3658, opened 2024-01-23
Handle postprocess tasks which don't have an eta.

jonathanmetzman:
Fixes https://pantheon.corp.google.com/errors/detail/COXs1_PEy9an-wE;time=PT6H?project=google.com:clusterfuzz


letitz (approved):
LGTM

PR #3229, opened 2023-07-16
Upgrade to Python3.11

jonathanmetzman:
1. Change Docker images to use Python3
2. Change local scripts to support Python3
3. Fix installation of Chromium dependencies
4. Remove profiling code, the profiling package only supports up to Python3.9
5. Recompile all protos and change some uworker serialization code that has to work differently due to some changes in dependencies
6. Add many no-member pylint exemptions since pylint is now misunderstanding proto members.
7. Change platforms for binary downloads during deploy
8. Upgrade packages that no longer work or have to be upgraded by because of depdency/dependent upgrades
[Used everywhere]
a. grpcio
b. grpcio-tools
c. protobuf
[development]
d. parameterized
e. WebTest
[bots]
f. google-cloud-batch
g. google-cloud-core
h. google-cloud-datastore
i. google-cloud-logging
j. google-cloud-monitoring
k. google-cloud-ndb
l. google-cloud-storage
m. google-api-python-client
n. google-auth
o. google-api-core
[appengine]
p. cryptography
q. firebase-admin
r. PyJWT

9. Remove no longer used pip packages:
a. twisted
b. lxml

PR #3663, opened 2024-01-23
Update to python3.11

jonathanmetzman:


jonathanmetzman:
I probably didn't run this one at all. https://github.com/google/clusterfuzz/pull/3229 with later patches was functional except on windows.

PR #3667, opened 2024-01-23
Save money on variant tasks

jonathanmetzman:
Run variant tasks on preemptibles on batch.


src/clusterfuzz/_internal/google_cloud_utils/batch.py
```diff
@@ -33,7 +33,9 @@
 _local = threading.local()
 
 MAX_DURATION = f'{int(60 * 60 * 2.5)}s'
-RETRY_COUNT = 0
+
+PREEMPTIBLE_RETRY_COUNT = 1
```
oliverchang, line 37: Is one time enough? What if the task gets preempted again?

jonathanmetzman: I bumped it up to 4.
I changed things so that all bots return 0 when the python code completes, regardless of exceptions, so that we never retry those tasks using this mechanism. I will use a different mechanism for retrying there.

jonathanmetzman: I changed it to 4. Also, now we will only retry on preemption. I'm going to use a different mechanism for application failures I think.

oliverchang (approved):

jonathanmetzman:
I'll close this for now as I don't think it will save that much since we've gotten the variant craziness under control.

PR #3671, opened 2024-01-24
Ensure handled errors in utasks are mapped.

alhijazi:


jonathanmetzman:
I think this is no longer needed because I landed Titouan's refactor of error handling.

PR #3674, opened 2024-01-25
Migration script

micahbales:
This update adds roll_back and batch_size arguments to the monorail/issuetracker migration script.

We need feedback on how best to run this script. It seems that we are intended to run it with butler.py, but it's not clear whether butler supports additional arguments, and it seems to require arguments (like script_name) that we are unsure where we get values for. Thank you for your review, and please advise!

### DIFF
--- a/src/local/butler/scripts/migration/monorail_to_issuetracker_migrator.py
+++ b/src/local/butler/scripts/migration/monorail_to_issuetracker_migrator.py
@@ -1,4 +1,4 @@
-# Copyright 2023 Google LLC
+# Copyright 2024 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/src/local/butler/scripts/migration/monorail_to_issuetracker_migrator.py
+++ b/src/local/butler/scripts/migration/monorail_to_issuetracker_migrator.py
@@ -14,10 +14,13 @@
 """Migrate Issue Ids from Monorail to Issue Tracker."""
 
 import csv
+import os
 
 from google.cloud import ndb
 
 from clusterfuzz._internal.datastore import data_types
+
+DEFAULT_BATCH_SIZE = 500
 
 
 def execute(args):
@@ -25,14 +28,24 @@ def execute(args):
   and/or group_bug_information fields to reflect the Issue Tracker issue
   id rather than the Monorail issue id."""
 
+  # Read the required enviroment variables.
+  file_loc = os.environ.get('FILE_LOC')
+  if not file_loc:
+    raise ValueError('Must specify FILE_LOC env variable')
+  project_name = os.environ.get('PROJECT_NAME')
+  if not project_name:
+    raise ValueError('Must specify PROJECT_NAME env variable')
+  batch_size = os.environ.get('BATCH_SIZE', DEFAULT_BATCH_SIZE)
+  roll_back = os.environ.get('ROLL_BACK', False)
+
   issue_id_dict = get_monorail_issuetracker_issue_id_dictionary(
-      args.file_loc, args.roll_back)
+      file_loc, roll_back)
 
   testcases = []
 
   for testcase in data_types.Testcase.query(
       # only target testcases in single project
-      data_types.Testcase.project_name == args.project_name,):
+      data_types.Testcase.project_name == project_name,):
     testcase_updated = False
     if testcase.bug_information and issue_id_dict.get(testcase.bug_information):
       testcase.bug_information = issue_id_dict[testcase.bug_information]
@@ -48,7 +61,7 @@ def execute(args):
       print(f'We will update testcase id: {testcase.key.id()}')
       testcases.append(testcase)
 
-    if args.non_dry_run and len(testcases) > args.batch_size:
+    if args.non_dry_run and len(testcases) >= batch_size:
       ndb.put_multi(testcases)
       print(f'Updated {len(testcases)}')
       testcases = []
--- a/src/local/butler/scripts/migration/monorail_to_issuetracker_migrator.py
+++ b/src/local/butler/scripts/migration/monorail_to_issuetracker_migrator.py
@@ -61,7 +61,7 @@ def get_monorail_issuetracker_issue_id_dictionary(file_loc, roll_back):
   # (ex. row: "600469, 40003765")
   with open(file_loc, 'r') as csvfile:
     reader = csv.reader(csvfile)
-    fieldnames = ['key', 'value']
+    fieldnames = ['monorail_id', 'issuetracker_id']
     reader = csv.DictReader(csvfile, fieldnames=fieldnames)
     for row in reader:
       key_id = row[fieldnames[1]] if roll_back else row[fieldnames[0]]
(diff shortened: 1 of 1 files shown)


micahbales:
@rmistry @oliverchang PTAL

rmistry:
@jonathanmetzman PTAL

src/local/butler/scripts/migration/monorail_to_issuetracker_migrator.py
```diff
@@ -43,15 +44,15 @@ def execute(args):
 
     testcases.append(testcase)
```
letitz, line 45: `testcase` should only be appended if modified above - no need to `put` it otherwise.

src/local/butler/scripts/migration/monorail_to_issuetracker_migrator.py
```diff
@@ -25,7 +25,8 @@ def execute(args):
   and/or group_bug_information fields to reflect the Issue Tracker issue
   id rather than the Monorail issue id."""
 
-  issue_id_dict = get_monorail_issuetracker_issue_id_dictionary(args.file_loc)
+  issue_id_dict = get_monorail_issuetracker_issue_id_dictionary(
+      args.file_loc, args.roll_back)
 
   testcases = []
```
letitz, line 32: As @oliverchang noted yesterday in #3648, you don't need to wrap the values assigned to fields in `ndb.{String,Integer}Property()`. In fact, it might break things - those functions are used to *define* fields in `ndb.Model` classes. You can just do:

```
testcase.bug_information = issue_id_dict[testcase.bug_information]
```
And similarly for `group_bug_information`.

src/local/butler/scripts/migration/monorail_to_issuetracker_migrator.py
```diff
@@ -63,8 +64,8 @@ def get_monorail_issuetracker_issue_id_dictionary(file_loc):
...
     for row in reader:
-      monorail_issue_id = row[fieldnames[0]]
-      issuetracker_issue_id = row[fieldnames[1]]
-      issue_id_dictionary[monorail_issue_id] = issuetracker_issue_id
+      key_id = row[fieldnames[1]] if roll_back else row[fieldnames[0]]
+      value_id = row[fieldnames[0]] if roll_back else row[fieldnames[1]]
+      issue_id_dictionary[key_id] = value_id
 
-  return issue_id_dictionary  # { monorail_issue_id: issuetracker_issue_id }
+  return issue_id_dictionary  # { key_id: value_id }
```
letitz, line 71: This comment is now superfluous: keys map to values :)

src/local/butler/scripts/migration/monorail_to_issuetracker_migrator.py
```diff
@@ -63,8 +64,8 @@ def get_monorail_issuetracker_issue_id_dictionary(file_loc):
     fieldnames = ['monorail_id', 'issuetracker_id']
     reader = csv.DictReader(csvfile, fieldnames=fieldnames)
     for row in reader:
-      monorail_issue_id = row[fieldnames[0]]
-      issuetracker_issue_id = row[fieldnames[1]]
-      issue_id_dictionary[monorail_issue_id] = issuetracker_issue_id
+      key_id = row[fieldnames[1]] if roll_back else row[fieldnames[0]]
+      value_id = row[fieldnames[0]] if roll_back else row[fieldnames[1]]
+      issue_id_dictionary[key_id] = value_id
```
letitz, line 69: This works, but I personally find it needlessly obscure. Also, you're defining `reader` twice, shadowing the first `csv.reader` with a `csv.DictReader` who confuses things more than it helps, IMO. How about:

```suggestion
for monorail_id, issuetracker_id in reader:
  if roll_back:
    issue_id_dictionary[issuetracker_id] = monorail_id
  else:
    issue_id_dictionary[monorail_id] = issuetracker_id
```

Or:

```suggestion
issue_id_dictionary = {
  monorail_id: issuetracker_id
  for monorail_id, issuetracker_id in reader
}
  
if roll_back:
  # Reverse ID mapping.
  issue_id_dictionary = {v: k for k,v in issue_id_dictionary}
```

src/local/butler/scripts/migration/monorail_to_issuetracker_migrator.py
```diff
@@ -43,15 +44,15 @@ def execute(args):
 
     testcases.append(testcase)
 
-    if args.non_dry_run and len(testcases) > 499:  # maximum batch size is 500
+    if args.non_dry_run and len(testcases) > args.batch_size:
       ndb.put_multi(testcases)
       testcases = []
 
   if args.non_dry_run and len(testcases) > 0:
     ndb.put_multi(testcases)
```
letitz, line 53: Can we add logging?

letitz (changes requested):

src/local/butler/scripts/migration/monorail_to_issuetracker_migrator.py
rmistry: Done.

src/local/butler/scripts/migration/monorail_to_issuetracker_migrator.py
rmistry: Done. I suspect once we can run it locally we will have to tweak this logging to see what is useful to see.

src/local/butler/scripts/migration/monorail_to_issuetracker_migrator.py
rmistry: Done. Thanks for the catch!

src/local/butler/scripts/migration/monorail_to_issuetracker_migrator.py
rmistry: Done

src/local/butler/scripts/migration/monorail_to_issuetracker_migrator.py
rmistry: Used the 1st suggestion. Thanks!

rmistry:
@letitz PTAL.
Note that I still cannot run it locally, so more fixes may be required.

letitz (approved):
LGTM

rmistry:
Failure seems unrelated-
```
  File "/workspace/src/clusterfuzz/_internal/tests/core/bot/tasks/task_types_test.py", line 47, in test_linux
    self.assertTrue(task_types.is_remote_utask('variant', job_name))
AssertionError: False is not true
```

src/local/butler/scripts/migration/monorail_to_issuetracker_migrator.py
```diff
@@ -25,33 +25,40 @@ def execute(args):
   and/or group_bug_information fields to reflect the Issue Tracker issue
```
jonathanmetzman, line 25: I don't think we can accept arbitrary command line args. Let's use env vars instead.

jonathanmetzman (approved):
lgtm

src/local/butler/scripts/migration/monorail_to_issuetracker_migrator.py
rmistry: Done. PTAL.

rmistry:
Rebased. Should be ready to be merged now.

oliverchang (approved):

PR #3679, opened 2024-01-26
[rearchitecture] Record how long utasks take

jonathanmetzman:
Record how long utasks take in two ways:
1. From start to finish
2. How long each part takes.

So we can figure out why batch seems to fall behind the queue-based system.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
@@ -50,7 +50,7 @@ def tworker_preprocess_no_io(utask_module, task_argument, job_type,
     logs.log_error('No uworker_input returned from preprocess')
     return None
 
-  uworker_input.preprocess_start_time = start
+  uworker_input.preprocess_start_time.CopyFrom(start)
 
   assert not uworker_input.module_name
   uworker_input.module_name = utask_module.__name__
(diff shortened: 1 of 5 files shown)


src/clusterfuzz/_internal/metrics/monitoring_metrics.py
```diff
@@ -156,6 +156,29 @@
     ],
 )
 
+UTASK_FINE_RUN_TIME = monitor.CounterMetric(
```
letitz, line 159: I think you want a CumulativeDistribution for this instead of a counter. A counter is good for tracking e.g. the number of items processed over time. Here, we don't want to know "how much time has this task spent processing utasks", but rather "what's the distribution of task durations (which will allow us to compute median, 90%, 99% durations)".

src/clusterfuzz/_internal/metrics/monitoring_metrics.py
```diff
@@ -156,6 +156,29 @@
...
+    field_spec=[
+        monitor.StringField('task'),
+        monitor.StringField('job'),
+        monitor.StringField('subtask'),
+        monitor.StringField('mode'),
+        monitor.StringField('platform'),
+    ],
+)
+
+UTASK_TOTAL_RUN_TIME = monitor.CounterMetric(
```
letitz, line 171: Same here.

src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
```diff
@@ -144,10 +182,51 @@ def uworker_bot_main():
   return 0
 
 
+def batch_record_finished_metrics(start, utask_module, job_type, subtask):
```
letitz, line 185: Nit: This sounds like a batch operation where you're recording finished metrics multiple times in a single call. The one below sounds like we're queueing a task to record metrics later. I suggest defining mode constants (or an enum) for `"batch"` and `"queue"` and passing those as arguments to `record_finished_metrics` directly.

PR #3697, opened 2024-01-29
Dont postprocess on non-linux bots

jonathanmetzman:
We don't want to spend their constrained time doing something a Linux bot can do.

### DIFF
--- a/src/clusterfuzz/_internal/base/tasks.py
+++ b/src/clusterfuzz/_internal/base/tasks.py
@@ -274,6 +274,9 @@ def get_postprocess_task():
   # This should only be run on non-preemptible bots.
   if not task_utils.is_remotely_executing_utasks():
     return None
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/base/tasks.py
```diff
@@ -274,6 +274,8 @@ def get_postprocess_task():
   # This should only be run on non-preemptible bots.
   if not task_utils.is_remotely_executing_utasks():
     return None
+  if not environment.platform().lower() == 'linux':
```
letitz, line 277: A comment here would be helpful to explain the intent behind the exclusion:

```suggestion
  # Postprocess is platform-agnostic, so we run all such tasks on our
  # most generic and plentiful bots.
  if not environment.platform().lower() == 'linux':
```

letitz (approved):
LGTM with a single comment.

PR #3698, opened 2024-01-29
Fix e2e subtask duration metric collection

letitz:
* Adds unit tests, which reproduced the bugs
* Fixes an inversion in the computation of the task duration, which was always negative
* Records the e2e duration from preprocess start time to subtask end time instead of only the current subtask's duration, as was the intent


jonathanmetzman (approved):
LGTM

PR #3699, opened 2024-01-30
Reduce wait times on user-requested tasks. 

jonathanmetzman:
Delays on user-requested tasks are harmful as they make users wait
and make ClusterFuzz less usable.
They probably don't provide much value on user-requested tasks since
these are unlikely to be requesting so many per second that a delay
prevents DoS.
Stop deferring tasks to prevent delays.
Potential TODO: Create mechanism for delay-less utasks, since the utask puller can be pulling other utasks for up to 60 seconds, and then might spend time preprocessing

### DIFF
--- a/src/appengine/handlers/upload_testcase.py
+++ b/src/appengine/handlers/upload_testcase.py
@@ -523,11 +523,14 @@ class UploadHandlerCommon:
           metadata.fuzzer_binary_name = target_name
           metadata.put()
 
+          # Use wait_time=None to execute the task ASAP, since it is
+          # user-facing.
           tasks.add_task(
               'unpack',
               str(metadata.key.id()),
               job_type,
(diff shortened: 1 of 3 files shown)


letitz (approved):
LGTM with a couple questions:

- can this be tested without undue effort?
- this does not progagate to later tasks, right? As in, once analyze is done it will kick off lots of other tasks, and those will have normal delays?

jonathanmetzman:
> LGTM with a couple questions:
> 
> * can this be tested without undue effort?

We can probably test some trivial parts, like the eta of the tasks put on the Q being time.time(). Want me to do this?

> * this does not progagate to later tasks, right? As in, once analyze is done it will kick off lots of other tasks, and those will have normal delays?

It does not propagate.

PR #3700, opened 2024-01-30
[analyze][rearchitecture] Reduce delays on untrusted parts of analyze

jonathanmetzman:
Instead of preprocessing and then putting on the queue for scheduling,
schedule right away. This eliminates up to 1 minute of time spent
pulling tasks from the queue and can improve execution speed on batch
by:
1. Allowing us to set a higher priority for the job.
2. Causing batch to schedule the job earlier because the job
requires less CPUs than jobs with many tasks. These CPUs may not
be available in times of high-usage.


letitz (approved):
Classic question: how hard is it to test this?

LGTM otherwise, with some nits below.

PR #3701, opened 2024-01-30
[Minimize] upload testcase using signed url

alhijazi:


alhijazi:
@jonathanmetzman 
Now we generate a `blob_name` in preprocess and create a signed_url using it. Then from `utask_main` upload the minimized_testcase where necessary.
One difference in behaviour is that while before, we were storing multiple blobs in storage at each call to `blobs.write_blob` and then storing the key in the testcase, now we only upload to one blob and overwrite it where necessary.

src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
```diff
@@ -355,6 +356,18 @@ def execute_parallel_runs(self, runs, instances=None):
     return results
 
 
+def _get_minimize_task_input():
+  blobs_bucket = storage.blobs_bucket()
+  blob_name = blobs.generate_new_blob_name()
+
+  gcs_path = storage.get_cloud_storage_file_path(blobs_bucket, blob_name)
+  if storage.get(gcs_path):
+    raise RuntimeError('UUID collision found: %s' % blob_name)
```
jonathanmetzman, line 365: f-string

jonathanmetzman (approved):
LGTM

PR #3703, opened 2024-01-30
Do not log error when app name is empty.

### LINKED ISSUE #3702: Please revert/fix/or finish https://github.com/google/clusterfuzz/commit/b38691b82ea080c2309588a72f797609779c3de5
It's the most common error on ClusterFuzz. As far as I can tell it happens all the time for libFuzzer jobs.
@letitz I think you authored it, could you please look into this.

letitz:
If `APP_NAME` is empty, then we should not expect to find a file with that name in search directories.

Fixes #3702.


marktefftech (approved):

jonathanmetzman (approved):
LGTM

PR #3711, opened 2024-02-01
Avoid crashing handler when testcase.platform is null.

letitz:
Some testcases have a null platform field. Account for that and avoid crashing when attempting to serve the testcase details page.

Fixes crbug.com/1504812.


letitz:
No! There seem to be a bunch of such testcases. It might be a symptom of another bug! Do you know if the platform is always expected to be there?

That said, if the DB does not enforce that the field must be set then we should allow it here.

jonathanmetzman:
> No! There seem to be a bunch of such testcases. It might be a symptom of another bug! Do you know if the platform is always expected to be there?
> 
> That said, if the DB does not enforce that the field must be set then we should allow it here.

I think it's expected. The db is nosql. It will accept anything.

jonathanmetzman:
I'll merge this but if you could look for what's causing thsi if you have free time, that would be great.

PR #3716, opened 2024-02-04
Check substitution mapping before adding any label directly to issue.labels + Update issue tracker URL

rmistry:
It is safe to do substitution mapping everytime because if a mapping is not found then it just returns the original label (see https://github.com/google/clusterfuzz/pull/3554/files)


rmistry:
PTAL @jonathanmetzman 
CC @micahbales

jonathanmetzman (approved):
lgtm

rmistry:
I updated this to do more substitution mapping.

PTAL @jonathanmetzman 
CC @micahbales

rmistry:
Tests passed.

@jonathanmetzman please merge and deploy when you can. TIA!

PR #3718, opened 2024-02-04
Force all chromium labels to go through the substitution mapping and add more logging

rmistry:


rmistry:
PTAL @jonathanmetzman 
CC @micahbales

rmistry:
Oh wait, there is a test failure

rmistry:
I don't think this will work. It would also lose the labels.added and labels.removed part.

Maybe I should only look at labels.added..

rmistry:
OK PTAL

jonathanmetzman (commented):
LGTM

jonathanmetzman (approved):
LGTM

PR #3721, opened 2024-02-04
Sanitize OS custom field. Chrome->ChromeOS

rmistry:


rmistry:
PTAL @jonathanmetzman 
CC @micahbales

PR #3722, opened 2024-02-04
Do get issue call right after modify issue call 

rmistry:
This is a workaround for the obfuscated emails issue

### DIFF
--- a/src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
+++ b/src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
@@ -73,8 +73,8 @@ def _sanitize_oses(oses):
   The OS custom field no longer has the 'Chrome' value.
   It was replaced by 'ChromeOS'.
   """
-  for i in range(len(oses)):
-    if oses[i] == 'Chrome':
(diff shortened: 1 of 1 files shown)


rmistry:
PTAL @jonathanmetzman 
CC @micahbales

src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
```diff
@@ -73,8 +73,8 @@ def _sanitize_oses(oses):
   The OS custom field no longer has the 'Chrome' value.
   It was replaced by 'ChromeOS'.
   """
-  for i in range(len(oses)):
-    if oses[i] == 'Chrome':
+  for i, os in enumerate(oses):
```
jonathanmetzman, line 76: nit: call this os_field so it's not conlficting with Python's os module.

jonathanmetzman (approved):

src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
rmistry: Done.

PR #3723, opened 2024-02-05
Do not update component tags if they are the same as on the issue

rmistry:
Context in b/323757748


rmistry:
CC @micahbales

jonathanmetzman:
LGTM

PR #3725, opened 2024-02-05
Use substition mappings before checking for predator labels in issues 

rmistry:
This is to hopefully stop the spam we are seeing in b/323757748


rmistry:
PTAL @jonathanmetzman 
CC @micahbales

jonathanmetzman (approved):
lgtm

PR #3731, opened 2024-02-06
Allow handling of blobs uploaded without filename metadata

alhijazi:


src/clusterfuzz/_internal/google_cloud_utils/storage.py
```diff
@@ -651,8 +651,11 @@ def __init__(self,
       self.size = size
     else:
       gcs_object = get(get_cloud_storage_file_path(bucket, object_path))
-
-      self.filename = gcs_object['metadata'].get(BLOB_FILENAME_METADATA_KEY)
+      if 'metadata' in gcs_object.keys():
+        self.filename = gcs_object['metadata'].get(BLOB_FILENAME_METADATA_KEY,
+                                                   None)
+      if not self.filename:
```
jonathanmetzman, line 657: I don't think we should log if it's totally normal behavior.

alhijazi: Could you confirm if filename is ever necessary? I didn't find a reason to believe so. 
I can remove it now, or we can leave it and then remove it later, as you wish.

jonathanmetzman: Let's remove later and land now.

jonathanmetzman:
Btw I suspect this is actually just fixing a symptom not the cause since this is fixing how we are handling minimized_keys.

PR #3733, opened 2024-02-06
Handle files without names

jonathanmetzman:


jonathanmetzman:
@oliverchang We have some blobs now without names because it's not really possible with signed URLs. Should we just use the name of the file in GCS as the name (since it's guaranteed to be unique way to identify the file (on disk for example))? Or are we worried about leaking the blob names (I don't think it actually matters in this case, but maybe in some other ones).

jonathanmetzman:
Fixes https://pantheon.corp.google.com/errors/detail/CNO7jNykldzt9gE;time=PT1H?project=cluster-fuzz&e=-13802955&mods=logs_tg_prod

jonathanmetzman:
Fixes https://g-issues.chromium.org/issues/324033361?pli=1&authuser=0

alhijazi (approved):

PR #3735, opened 2024-02-08
[fuzz task] Fix issue breaking blackbox fuzzer stats

jonathanmetzman:
Indentation was messed up causing the fuzzer name to only be set for engine fuzzers. Thanks to protobuf's brilliant defaulting to zero-ish values instead of None or exceptioning for unset fields, this was not discovered earlier.


jonathanmetzman:
CC @mi-ac

oliverchang (approved):

PR #3737, opened 2024-02-09
Fix progression task no-crash error message.

letitz:
`known_crash_revision` is not guaranteed to be equal to `min_revision`, which is what is being tested. Include the right revision in the error message if the crash does not reproduce at `min_revision`.

Bug: crbug.com/40229463

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
@@ -536,8 +536,7 @@ def find_fixed_range(uworker_input):
     return error
 
   if result and not result.is_crash():
-    error_message = (
-        f'Minimum revision r{min_revision} did not crash.')
+    error_message = (f'Minimum revision r{min_revision} did not crash.')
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -536,8 +536,7 @@ def find_fixed_range(uworker_input):
     return error
 
   if result and not result.is_crash():
-    error_message = (
-        f'Known crash revision {known_crash_revision} did not crash')
+    error_message = (f'Minimum revision r{min_revision} did not crash.')
```
alhijazi, line 539: No need for parentheses anymore.

alhijazi (approved):
LGTM % 1 small change.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
letitz: Ah, of course.

PR #3745, opened 2024-02-13
Try static_discovery in discovery call 

rmistry:
Trying solution from googleapis/google-api-python-client#1263 for b/324565392


rmistry:
PTAL @jonathanmetzman

jonathanmetzman (approved):
LGTM

PR #3751, opened 2024-02-13
Execute corpus_pruning on batch

jonathanmetzman:


jonathanmetzman:
It's not ready to fully execute on batch (meaning it uses trusted code), but this will test batch's scalability.

alhijazi (commented):
LGTM

PR #3754, opened 2024-02-13
Get signing credentials for non-gce hosts.

jonathanmetzman:
This will reduce the number of calls we make to the credential API.


oliverchang (approved):

jonathanmetzman:
I merged this manually because github is broken now.

PR #3755, opened 2024-02-13
Sign URLs locally

jonathanmetzman:
This is faster and less likely to run into rate-limiting issues. We can probably also get rid of hard to debug multirprocessing code.


jonathanmetzman:
I'm going to do some profiling before I merge this.
The code is copied from here: https://cloud.google.com/storage/docs/access-control/signing-urls-manually

oliverchang:
This is no longer relevant right?

PR #3756, opened 2024-02-13
Remove existing FoundIn labels before updating.

alhijazi:


alhijazi:
> Why do we need to do this?
It was requested in:
https://issues.chromium.org/issues/324491446

PR #3760, opened 2024-02-14
Do fuzzer selection in preprocess

jonathanmetzman:
Handle fuzzer selection in a way that is compatible with the fact that corpus access doesn't happen during uworker_main.
This means we will not be able to record fuzz targets during main or access the corpus then, so fuzzer selection must happen first.


src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -1593,45 +1593,82 @@ def get_testcase_variant(testcase_id, job_type):
...
+  # reasonably and won't try to DoS us by putting bogus fuzzers in the db.
+  # This should be changed by limiting the number of fuzz targets saved and
+  # putting an expiration on them.
+  if not binaries:
+    logs.log_error('Expected binaries.')
+    return None
+
+  project = get_project_name(job_type)
+
+  Target = collections.namedtuple('Target', ['engine', 'project', 'binary'])
```
oliverchang, line 1645: Why do we need these? Can we just construct the real FuzzTarget entities here instead?

src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -1593,45 +1593,82 @@ def get_testcase_variant(testcase_id, job_type):
...
+          job=job_type,
+          engine=engine_name,
+          last_run=time_now) for key_name in mapping
+  }
+
+  jobs = get_or_create_multi(mapping, data_types.FuzzTargetJob)
+  for job in jobs:
+    if not job.last_run:
+      job.last_run = utils.utcnow()
+      ndb.put_multi(jobs)
```
oliverchang, line 1669: is this meant to be indented at its current level?

src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -1593,45 +1593,82 @@ def get_testcase_variant(testcase_id, job_type):
...
+
+  return fuzz_targets
```
oliverchang, line 1671: Is this return value used anywhere?

src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -1593,45 +1593,82 @@ def get_testcase_variant(testcase_id, job_type):
...
+
+  project = get_project_name(job_type)
+
+  Target = collections.namedtuple('Target', ['engine', 'project', 'binary'])
+
+  mapping = {
+      data_types.fuzz_target_fully_qualified_name(engine_name, project, binary):
+      Target(engine_name, project, binary) for binary in binaries
+  }
+  fuzz_targets = get_or_create_multi(mapping, data_types.FuzzTarget)
```
oliverchang, line 1651: We probably don't need to save the result into a variable do we?

src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -1593,45 +1593,82 @@ def get_testcase_variant(testcase_id, job_type):
...
       binary_name=binary_name,
       job_type=job_type)
-  return fuzz_target
+  return result
+
+
+def get_or_create_multi(mapping, data_type):
+  """Gets or creates multiple db entities."""
+  keys = list(mapping.keys())
+  entities = ndb.get_multi([ndb.Key(data_type, key) for key in mapping])
```
oliverchang, line 1616: use the ndb_utils variants for these: https://github.com/google/clusterfuzz/blob/7bff389c58f58bab0c58682697671660ebfc94da/src/clusterfuzz/_internal/datastore/ndb_utils.py#L53

These handle higher batch counts than what ndb allows by default.

src/clusterfuzz/_internal/protos/uworker_msg.proto
```diff
@@ -168,6 +168,8 @@ message FuzzTaskOutput {
   optional StoreFuzzerRunResultsOutput fuzzer_run_results = 8;
   optional int32 new_targets_count = 9;
   optional int32 fuzzer_revision = 10;
+  repeated string fuzz_targets = 11;
+  optional string used_fuzzer = 12;
```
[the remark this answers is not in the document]
jonathanmetzman: These are all fuzz targets in the build, not the used fuzz targets.
That's under used_fuzzer.

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -1779,13 +1781,17 @@ def run(self):
...
+    if fuzz_target:
+      environment.set_value('FUZZ_TARGET', fuzz_target)
     build_setup_result = build_manager.setup_build(
-        environment.get_value('APP_REVISION'), target_weights=target_weights)
+        environment.get_value('APP_REVISION'))
+
+    output_to_report_targets = self.save_fuzz_targets(fuzz_target,
+                                                      build_setup_result)
+    if output_to_report_targets:
+      return output_to_report_targets
```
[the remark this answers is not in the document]
jonathanmetzman: Yeah. I think it makes more sense to skip fuzzing.

src/clusterfuzz/_internal/datastore/data_handler.py
jonathanmetzman: Good catch.

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -2000,10 +2048,20 @@ def utask_preprocess(fuzzer_name, job_type, uworker_env):
   )
 
 
+def save_fuzz_targets(output):
```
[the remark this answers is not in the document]
jonathanmetzman: Done.
A quesiton I had when making this (which I've now documented, is how much do we trust this? I think an untrusted engine could potentially misuse this for DoS purposes, but I don't think that's really worth defending against. WDYT?

src/clusterfuzz/_internal/datastore/data_handler.py
jonathanmetzman: Done.

src/clusterfuzz/_internal/datastore/data_handler.py
jonathanmetzman: Yes. In preprocess.

src/clusterfuzz/_internal/datastore/data_handler.py
jonathanmetzman: I think we do because we need to put the fuzzers and return them.

src/clusterfuzz/_internal/protos/uworker_msg.proto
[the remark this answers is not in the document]
oliverchang: I'm confused :) My comment was to replace "fuzzer" with "fuzz_target" to be consistent with terminology. Is that correct?

src/clusterfuzz/_internal/datastore/data_handler.py
oliverchang: Might be missing something but is this return value actually used? Are you talking about:

```python
def handle_fuzz_no_fuzz_target_selected(output):
  save_fuzz_targets(output)
  # Try again now that there are some fuzz targets.
  utask_preprocess(output.uworker_input.fuzzer_name,
                   output.uworker_input.job_type,
                   output.uworker_input.uworker_env)
```

src/clusterfuzz/_internal/protos/uworker_msg.proto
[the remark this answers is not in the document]
jonathanmetzman: You are right. I was confused. Let me change that term.
I think there was difficulty in me understanding what happened here becuase we were incorrectly still doing this in the blackbox fuzzer case as an artifact of when engine fuzzers ran there.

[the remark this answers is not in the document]
jonathanmetzman: Actually I don't think we need to record which fuzzer we used at all.

src/clusterfuzz/_internal/datastore/data_handler.py
jonathanmetzman: In preprocess, we call record_fuzz_target, this calls record_fuzz_targets under the hood. So we do need to return the value.

oliverchang: Ah, missed that!

PR #3764, opened 2024-02-16
Fix more problems backing up corpus

jonathanmetzman:


src/clusterfuzz/_internal/fuzzing/corpus_manager.py
```diff
@@ -361,7 +361,13 @@ class ProtoFuzzTargetCorpus(FuzzTargetCorpus):
   """Implementation of GCS corpus that uses protos (uworker-compatible) for fuzz
   targets."""
 
-  def __init__(self, proto_corpus):  # pylint: disable=super-init-not-called
+  def __init__(self, engine, project_qualified_target_name, proto_corpus):  # pylint: disable=super-init-not-called
+    # TODO(metzman): Do we need project_qualified_target_name?
+
+    # This is used to let AFL share corpora with libFuzzer.
+    self._engine = os.getenv('CORPUS_FUZZER_NAME_OVERRIDE', engine)
```
oliverchang, line 368: Can we push this logic to a bit earlier in the call stack?

i.e. Is `engine` in the context of this file always the `CORPUS_FUZZER_NAME_OVERRIDE` (with `engine` as a default)? I see this logic in various different places in this file. 

I wonder if we can do this logic earlier on in a more generalized location, so that classes like these can just take in the correct `engine` to begin with.

jonathanmetzman: You want callers to make this call when creating this class?
I think that's a bit annoying to repeat, but I'll change it if you want.

oliverchang: More that if there is a single place where we can set the engine and do this logic, and have this engine propagate everywhere. e.g. in corpus_pruning_task. 

This way we reduce the dependency on global state.

PR #3767, opened 2024-02-20
Consider job.platform's base platform for task compatibility

marktefftech:
Since introducing the ability to [create jobs with new queues](https://github.com/google/clusterfuzz/blob/f44d93d51bfdec7767679a38787a818b8949c67e/src/clusterfuzz/_internal/cron/project_setup.py#L881) this means that a `job.platform` can equal something like `ANDROID:<model>`.

Therefore we consider the "base" platform of both the job and the bot.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/commands.py
+++ b/src/clusterfuzz/_internal/bot/tasks/commands.py
@@ -276,23 +276,23 @@ def process_command_impl(task_name,
       logs.log_error(error_string)
       raise errors.BadStateError(error_string)
 
-    job_queue_suffix = tasks.queue_suffix_for_platform(
+    job_base_queue_suffix = tasks.queue_suffix_for_platform(
         environment.base_platform(job.platform))
     bot_platform = environment.platform().lower()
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/bot/tasks/commands.py
```diff
@@ -276,7 +276,8 @@ def process_command_impl(task_name,
       logs.log_error(error_string)
       raise errors.BadStateError(error_string)
 
-    job_queue_suffix = tasks.queue_suffix_for_platform(job.platform)
+    job_queue_suffix = tasks.queue_suffix_for_platform(
```
oliverchang, line 279: nit: can we rename this to `job_base_queue_suffix` for consistency?

oliverchang (approved):
LGTM with a nit

src/clusterfuzz/_internal/bot/tasks/commands.py
marktefftech: Done.

oliverchang (approved):

PR #3768, opened 2024-02-21
Let V8 Sandbox Violations overwrite other crash types

saelo:
Previously, we would only check for the V8 Sandbox Violation crash type if no other crash type was detected earlier on. As such, if a sandbox violation (i.e. memory corruption outside of the V8 sandbox) was first detected by ASan, the resulting report would be treated as an ASan error rather than a Sandbox Violation. However, we prefer to always label these crashes as V8 Sandbox Violations, regardless of the underlying type of memory corruption. This CL modifies the logic accordingly.


jonathanmetzman (approved):
LGTM

PR #3772, opened 2024-02-22
Fix use of repeated field

jonathanmetzman:


oliverchang (commented):
LGTM

jonathanmetzman:
@alhijazi Watch out for this bug btw. Repeated fields cannot use HasField.

PR #3774, opened 2024-02-23
Improve logging so we can see tasks postprocess and execute

jonathanmetzman:


src/clusterfuzz/_internal/protos/uworker_msg.proto
```diff
@@ -99,19 +99,22 @@ message Input {
...
-  optional RegressionTaskInput regression_task_input = 17;
-  optional SymbolizeTaskInput symbolize_task_input = 18;
-  optional string module_name = 19;
-  optional google.protobuf.Timestamp preprocess_start_time = 20;
+  optional VariantTaskInput variant_task_input = 8;
+  optional string fuzzer_name = 9;
+  optional SetupInput setup_input = 10;
+  optional AnalyzeTaskInput analyze_task_input = 11;
+  optional CorpusPruningTaskInput corpus_pruning_task_input = 12;
+  optional FuzzTaskInput fuzz_task_input = 13;
```
oliverchang, line 107: nit: in general it's bad practice to  change these numbers, as it breaks backwards compatibility.

src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
```diff
@@ -79,25 +79,41 @@ def ensure_uworker_env_type_safety(uworker_env):
...
                                                 uworker_env)
   if not uworker_input:
     logs.log_error('No uworker_input returned from preprocess')
     return None
 
+  logs.log_error('Preprocess finished.')
+
+  task_payload = environment.get_value('TASK_PAYLOAD')
+  if task_payload:
+    uworker_input.uworker_env['INITIAL_TASK_PAYLOAD'] = task_payload
```
oliverchang, line 99: Can we just use TASK_PAYLOAD here instead? Why the distinction with the "INITIAL_" prefix? 

If this prefix is required for a specific reason, please leave a comment in the code here.

PR #3779, opened 2024-02-25
Make task timeouts consistent with ttask lengths.

jonathanmetzman:


jonathanmetzman:
@letitz I think some of the progressions are failing due to the inconsistency, which was fine when only variant ran on batch.

PR #3780, opened 2024-02-26
Stack analyzer: Ignore centipede-specific stack frames

mi-ac:
This will remove common centipede internals from stack traces to prevent too eager grouping. Most fuzztests only differ in top-stack. The second and third stack frame are identical centipede frames, which results in Clusterfuzz grouping otherwise unique cases together.

The associated Chromium bug is: https://crbug.com/326493640


oliverchang (approved):

jonathanmetzman (approved):
LGTM

PR #3785, opened 2024-02-27
Use dedicated account for signing

jonathanmetzman:


src/Pipfile
```diff
@@ -33,6 +33,7 @@ requests = "==2.21.0"
 sendgrid = "==6.0.4"
 # Hack: We are using this to specify App Engine packages,
 # since there is no support for custom package sets.
+google-cloud-secret-manager = "==2.18.0"
```
oliverchang, line 36: nit: this should come before the comment. The comment is referring to the dev-packcages bit.

src/clusterfuzz/_internal/google_cloud_utils/credentials.py
```diff
@@ -58,22 +62,45 @@ def get_default(scopes=None):
...
+  """Returns signing credentials for signing URLs."""
+  if _use_anonymous_credentials():
+    return None
+
+  if service_account_path is not None:
     # Handle cases like android and Mac where bots are run outside of Google
     # Cloud Platform and don't have access to metadata server.
     signing_creds = service_account.Credentials.from_service_account_file(
-        google_application_credentials, scopes=_SCOPES)
+        service_account, scopes=_SCOPES)
```
oliverchang, line 103: Does it work if we pass the JSON decoded service account dict to https://google-auth.readthedocs.io/en/master/reference/google.oauth2.service_account.html#google.oauth2.service_account.Credentials.from_service_account_info ? 

This way we avoid the complexity of writing out to a file and keep everything in memory.

jonathanmetzman: I'll check tomorrow (also when i'll write tests).

oliverchang: I think it should work given the example here: https://google-auth.readthedocs.io/en/master/reference/google.oauth2.service_account.html#domain-wide-delegation

```
service_account_info = json.load(open('service_account.json'))
credentials = service_account.Credentials.from_service_account_info(
    service_account_info)
```

jonathanmetzman: It will add some complexity because we are already using the file based version on mac/linux/android. Maybe I'll just stop supporting GOOGLE_APPLICATION_CREDENTIALS for signing entirely?

oliverchang: Yes let's keep things homogeneous and use this account everywhere. The existing service accounts should have access to these secrets.

oliverchang (commented):
LGTM

PR #3787, opened 2024-02-27
Minimizer: Skip minimization early if not supported

mi-ac:
This only adds a minimization task if it's supported (black-box fuzzers and libfuzzer). E.g. centipede doesn't support minimization yet and currently minimization errors out, spamming the logs and the test cases.

This gracefully starts all post-minimization tasks from the beginning if minimization is not supported. Existing code for skipping minimization is untouched, since it's still executed when users select minimization explicitly.

This PR doesn't add a test since there are no tests of task_creation yet (and the hurdle for adding the first test seems rather high).


jonathanmetzman (approved):
LGTM

PR #3790, opened 2024-02-27
Dont save fuzz targets with incorrect names

jonathanmetzman:
The "name" at this point is a qualified named. Don't save this so we don't end up with fuzz targets like libFuzzer_libFuzzer...
We save the actual name of the binary later on. Just use that.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
@@ -44,7 +44,6 @@ from clusterfuzz._internal.crash_analysis.crash_result import CrashResult
 from clusterfuzz._internal.crash_analysis.stack_parsing import stack_analyzer
 from clusterfuzz._internal.datastore import data_handler
 from clusterfuzz._internal.datastore import data_types
-from clusterfuzz._internal.datastore import fuzz_target_utils
 from clusterfuzz._internal.datastore import ndb_utils
 from clusterfuzz._internal.fuzzing import corpus_manager
 from clusterfuzz._internal.fuzzing import fuzzer_selection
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
Sorry this isn't ready yet.

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -1786,10 +1785,14 @@ def run(self):
...
-      output_to_report_targets = self.record_fuzz_targets_to_output(
-          build_setup_result)
+      self.fuzz_task_output.fuzz_targets.extend(build_setup_result.fuzz_targets)
+      if not self.fuzz_task_output.fuzz_targets:
+        logs.log_error('No fuzz targets.')
+
       if not fuzz_target:
-        return output_to_report_targets
+        return uworker_msg_pb2.Output(
+            fuzz_task_output=self.fuzz_task_output,
```
alhijazi, line 1794: You set the fuzz_task_output here but not in the consequent Output's returned in case of error. 
Is that intentional?

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -2042,6 +2044,8 @@ def save_fuzz_targets(output):
   """Saves fuzz targets that were seen in the build to the database."""
   if not output.fuzz_task_output.fuzz_targets:
     return
+
+  logs.log(f'Saving fuzz targets: {output.fuzz_task_output.fuzz_targets}.')
```
alhijazi, line 2048: I think `save_fuzz_targets` should be called before calling error_handling in `utask_postprocess`.

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
jonathanmetzman: Yep.

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
jonathanmetzman: I'm not sure it will have any practical imapct and I'm generally opposed to running any code before error handling.

alhijazi (approved):

PR #3793, opened 2024-02-27
Make logs functions consistent with stdlib's logging module.

jonathanmetzman:


jonathanmetzman:
This is something that I always get tripped up on when programming in ClusterFuzz. Especially `logs.log(` (which really should be log.info.

PR #3794, opened 2024-02-28
Disable blob deletion in minimize_task

jonathanmetzman:


oliverchang (approved):
What are the exceptions?

PR #3796, opened 2024-02-28
build_manager: add fuzzer binary to custom build back

paulsemel:


paulsemel:
@alhijazi can you PTAL?

PR #3797, opened 2024-02-28
[Minimize] Fix cleaning up blobs in postprocess

alhijazi:


src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
```diff
@@ -633,8 +633,8 @@ def _cleanup_unused_blobs_from_storage(output: uworker_msg_pb2.Output):
   delete_stacktrace_blob = True
 
   if output.HasField('minimize_task_output'):
-    testcase_blob_key = output.minimize_task_output.minimized_keys
-    if testcase_blob_key.startswith(data_types.BLOBSTORE_STACK_PREFIX):
+    # If minimized_keys was set, we should not cleanup the corresponding blob.
```
jonathanmetzman, line 636: Do we need to make this change on 642?

jonathanmetzman:
Could you make the fix for the existing cases please?

src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
alhijazi: No, since the difference with stacktraces is that they "sometimes" contain a blob key, in which case the key itself is prefixed by `BLOBSTORE_STACK_PREFIX`. Whereas minimized_keys always contains a blob_name when set.

jonathanmetzman:
I'm going to have to wait to merge this until next week.

src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
jonathanmetzman: Why are we even cleaning up these in the first place? It's not clear to me.

alhijazi: `blobs.get_blob_signed_upload_url` creates the a blob and returns its name so that subsequent calls don't end up using the same blob for a different task.
If we don't end up using this object during main, we don't need to keep holding it, thus the cleanup in postprocess.

jonathanmetzman (approved):
LGTM

PR #3798, opened 2024-02-28
Workaround to use fuzzed_keys instead of minimized_keys for missing m…

alhijazi:
…inimzed testcases

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/setup.py
+++ b/src/clusterfuzz/_internal/bot/tasks/setup.py
@@ -345,8 +345,14 @@ def _get_testcase_key_and_archive_status(testcase):
   """Returns the testcase's key and whether or not it is archived."""
   if _is_testcase_minimized(testcase):
     key = testcase.minimized_keys
-    archived = bool(testcase.archive_state & data_types.ArchiveStatus.MINIMIZED)
-    return key, archived
+    # TODO(alhijazi): Remove this check once the issue with blob cleanup
+    # in minimize postprocess is resolved.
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/bot/tasks/setup.py
```diff
@@ -345,8 +345,16 @@ def _get_testcase_key_and_archive_status(testcase):
...
-    archived = bool(testcase.archive_state & data_types.ArchiveStatus.MINIMIZED)
-    return key, archived
+    # TODO(alhijazi): Remove this check once the issue with blob cleanup
+    # in minimize postprocess is resolved.
+    if storage.get(blobs.get_gcs_path(key)):
+      archived = bool(
+          testcase.archive_state & data_types.ArchiveStatus.MINIMIZED)
+      return key, archived
+    else:
+      logs.log(
```
alhijazi, line 355: @jonathanmetzman, I think we could even schedule a minimize_task here. WDYT?

jonathanmetzman (approved):
LGTM

src/clusterfuzz/_internal/bot/tasks/setup.py
jonathanmetzman: WHy?

alhijazi: Minimize task won't be rescheduled for this issue. It had terminated properly and we had mistakenly deleted the minimized testcase. We could reschedule minimize here to regenerate the minimized testcase and that would be used later by progression task.

PR #3800, opened 2024-02-28
Fix FUZZ_TARGET setting

jonathanmetzman:


jonathanmetzman:
This error is partly the fault of ClusterFuzz's error prone use of environment variables.
1. There's no reason env vars should have been used here and it's essentially a global variable (and not even a constant one).
2. environment.set_value converts any type to string before saving it to the environment while get_value parses env vars as python code. This does not work for nonbuiltin types, and this error was hard to spot because we put a proto in the environment. I think I will improve set_value by checking that get_value returns something equal.

PR #3802, opened 2024-02-29
Set initial task name as task name when updating comments. 

jonathanmetzman:
So that users see e.g. when minimized task ends, not
that postprocess ends.


src/clusterfuzz/_internal/system/environment.py
```diff
@@ -1039,6 +1039,13 @@ def set_value(environment_variable, value, env=None):
                                                value_str)
 
 
+def get_initial_task_name():
```
oliverchang, line 1042: Can you add some docstrings and comments explaining what this is?

oliverchang (approved):

PR #3805, opened 2024-03-05
Set consistent minimization keys also when minimization is skipped

mi-ac:
The _skip_minimization routine in minimize_task sets the minimized_keys before bailing out. This was missing in PR https://github.com/google/clusterfuzz/pull/3787.  Various other things in Clusterfuzz check these keys, e.g. the grouping task waits for minimization to finish by checking for content in minimized_keys. Without this fix, grouping never runs on respective test cases and non-grouped tests don't show up in the testcases list.

This doesn't retain the previous behavior, which set minimized_keys to fuzzed_keys. Setting an 'NA' string is more intuitive and is also used in a few other places already that mark minimization as not available.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/task_creation.py
+++ b/src/clusterfuzz/_internal/bot/tasks/task_creation.py
@@ -247,6 +247,9 @@ def create_tasks(testcase):
   else:
     # Environments that don't support minimization skip directly to other
     # tasks.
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
Do you want to add some tests here?

jonathanmetzman (approved):
LGTM

jonathanmetzman:
Minimizer then grouper is a bit beyond the tests we run in CF, but could be interesting to try.

PR #3806, opened 2024-03-05
Dont make proto backwards incompatible.

jonathanmetzman:


jonathanmetzman:
I may have reqested the reordering but it was a mistake.

PR #3808, opened 2024-03-06
Ensure preemptive hosts only run fuzz tasks for android

marktefftech:


src/clusterfuzz/_internal/platforms/android/settings.py
```diff
@@ -159,6 +159,11 @@ def is_mte_build():
   return 'mte' in build_flavor
 
 
+def using_preemptible_host():
+  """Return True if using a peemptible host."""
+  return '-pre-' in environment.get_value('DOCKER_HOST', '')
```
oliverchang, line 164: See my question on the other internal CL :) 

I would've expected the hostname to propagate to inside the container, as that's what we expect in our other environments. 

DOCKER_HOST also seems like it's leaking abstraction a bit (the fact that this is running in a docker container). Perhaps we could just call this HOSTNAME if we do end up needing an env var?

PR #3817, opened 2024-03-06
Set every attribute in __init__

jonathanmetzman:


jonathanmetzman:
This isn't ready yet.

oliverchang (commented):
we should consider https://docs.python.org/3/library/dataclasses.html for some of these, which makes defining attributes (with types and default values) a lot easier.

jonathanmetzman:
> we should consider https://docs.python.org/3/library/dataclasses.html for some of these, which makes defining attributes (with types and default values) a lot easier.

I'm going to convert this to a proto anyway, but was doing some intermediate refactoring.

PR #3820, opened 2024-03-06
Dont use multiprocessing for signing URLs.

jonathanmetzman:
It might be hanging and is uneeded now that signing is CPU-bound and done locally.


oliverchang (commented):
lgtm

PR #3824, opened 2024-03-07
Fix handling of Fuchsia builds

jonathanmetzman:


src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -125,6 +125,12 @@ def do_multiarmed_bandit_strategy_selection(uworker_env):
                         uworker_env)
 
 
+def has_standard_build():
```
oliverchang, line 128: add a docstring.

PR #3825, opened 2024-03-07
Optimize corpus operations

jonathanmetzman:
Dont create upload or delete URLs unnecessarily.


jonathanmetzman:
This shuld save at least 15 seconds on each corpus pruning run in Chrome ClusterFuzz. It saves ~6 minutes in Cuttlefish because cuttlefish is so much slower signing URLs. This should also be investigated.

PR #3826, opened 2024-03-07
Downgrade pyOpenSSL+cryptography to support Xenial.

jonathanmetzman:
We want to use cryptography to sign URLs quickly instead of in pure python.
Android runs on xenial and can't support this version of cryptography (which depends on a pyOpenSSL downgrade).
This fix may also help our Windows bots.


jonathanmetzman:
I think it would be good for Android to be on the same distro (and even same image) as ClusterFuzz.

oliverchang (commented):
LGTM

PR #3810, opened 2024-03-06
Dont save uploaded testcases to regression corpus

jonathanmetzman:

PR #3830, opened 2024-03-08
Let utasks perform untrusted behavior in main

jonathanmetzman:


jonathanmetzman:
This should be the last or second CL to land in the rearchitecture. It will allow force trusted workers to abort when they are handling an untrusted worker, when combined with https://github.com/google/clusterfuzz/pull/3810

PR #3832, opened 2024-03-08
Do not call stop_cvd when no device is running

aditya-wazir:
Validate device state prior to 'stop_cvd' execution; proceed only if the current state is 'device'.


marktefftech (approved):

marktefftech:
LGTM

PR #3836, opened 2024-03-11
[zip] Favor fileobj over filepath

jonathanmetzman:


jonathanmetzman:
I think this needs tests and to provide the same behavior for .tar.gz. Paul could you add those once this lands?

jonathanmetzman:
We also can't accept zip files that aren't meant to be tested independently. Unsure if this fixes that.

jonathanmetzman:
On second thought I don't think we can improve the velocity at which we merge commits from Chrome fuzzing if I spend so much time fixing issues with them. @paulsemel Can you please take over this CL from me?

oliverchang (approved):

jonathanmetzman:
I merged this since Paul is OOO and someone is blocked now.

PR #3840, opened 2024-03-11
Handle exceptions in our task scheduler.

oliverchang:
This can happen due to e.g. job deletions, and block testcases from getting progression scheduled.


jonathanmetzman (approved):
LGTM

PR #3848, opened 2024-03-13
Reduce the severity of HWASAN not being supported by sanitizer.py

ParisMeuleman:
@jonathanmetzman noticed a spike in these errors, we are tracking the action of fixing the option handling in https://issues.chromium.org/u/2/issues/329346758, meanwhile, we can reduce the severity of the log.

I am not sure of the actual fix, the HWAsan documentation is a bit lacking. It seems we might be able to set those as an environment variable in a wrap.sh. Can/should this be done by clusterfuzz is to be determined. 

Engine fuzzers are not uzing this (most of android fuzzing IIUC?)

### DIFF
--- a/src/clusterfuzz/_internal/platforms/android/sanitizer.py
+++ b/src/clusterfuzz/_internal/platforms/android/sanitizer.py
@@ -57,13 +57,13 @@ def get_options_file_path(sanitizer_tool_name):
   # Reduce the log severity for HWASAN
   # TODO(crbug.com/329346758) properly handle HWASAN options.
   if sanitizer_tool_name.lower() == "hwasan":
-    logs.log('Unsupported sanitizer: ' + sanitizer_tool_name)
+    logs.log(f'Unable to set options for sanitizer: {sanitizer_tool_name}')
     return None
 
   sanitizer_filename = SANITIZER_TOOL_TO_FILE_MAPPINGS.get(
       sanitizer_tool_name.lower())
   if sanitizer_filename is None:
-    logs.log_error('Unsupported sanitizer: ' + sanitizer_tool_name)
+    logs.log_error(f'Unable to set options for sanitizer: {sanitizer_tool_name}')
     return None
 
   return os.path.join(sanitizer_directory, sanitizer_filename)
--- a/src/clusterfuzz/_internal/platforms/android/sanitizer.py
+++ b/src/clusterfuzz/_internal/platforms/android/sanitizer.py
@@ -63,7 +63,8 @@ def get_options_file_path(sanitizer_tool_name):
   sanitizer_filename = SANITIZER_TOOL_TO_FILE_MAPPINGS.get(
       sanitizer_tool_name.lower())
   if sanitizer_filename is None:
-    logs.log_error(f'Unable to set options for sanitizer: {sanitizer_tool_name}')
+    logs.log_error(
+        f'Unable to set options for sanitizer: {sanitizer_tool_name}')
     return None
 
   return os.path.join(sanitizer_directory, sanitizer_filename)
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/platforms/android/sanitizer.py
```diff
@@ -54,6 +54,12 @@ def get_options_file_path(sanitizer_tool_name):
   sanitizer_directory = ('/system' if settings.get_sanitizer_tool_name() else
                          constants.DEVICE_TMP_DIR)
 
+  # Reduce the log severity for HWASAN
+  # TODO(crbug.com/329346758) properly handle HWASAN options.
+  if sanitizer_tool_name.lower() == "hwasan":
+    logs.log('Unsupported sanitizer: ' + sanitizer_tool_name)
```
jonathanmetzman, line 60: Can you use an f-string?

jonathanmetzman:
I don't have a great understanding of this some I'm a little worried about turning this off for the Android team's ClusterFuzz instance.

ParisMeuleman:
> I don't have a great understanding of this some I'm a little worried about turning this off for the Android team's ClusterFuzz instance.

makes sense, I guess involving @marktefftech is our best course of action? 

FYI regarding the follow up, I'm testing adding the HWASAN_OPTIONS through a [wrap.sh in the apk](https://developer.android.com/ndk/guides/wrap-script#debugging_when_using_wrapsh). I had some success but still need to test some more.

src/clusterfuzz/_internal/platforms/android/sanitizer.py
marktefftech: Can you change the log to be more indicative of the issue here?

We _do_ support hwasan as a sanitizer, but from what I can see, we _do not_ support setting additional `HWASAN_OPTIONS` via a metadata file.

So maybe:

```
    logs.log('Unable to set additional sanitizer options for HWASAN: ' + sanitizer_tool_name)
```

marktefftech:
This logging change LGTM

As far as a long-term solution, it would be nice to follow the [existing convention](https://github.com/google/clusterfuzz/blob/e2c5925a92fb3bd0c63523a33bdca3cc89f0f651/src/clusterfuzz/_internal/system/environment.py#L889-L904) of setting the `ADDITIONAL_<sanitizer>_OPTIONS`.

This will keep things the same for hwasan as they are for other sanitizers. We can set the options via configuration, for example, using the ClusterFuzz job definitions.

Take the `engine_asan` template as an example, which sets `ADDITIONAL_ASAN_OPTIONS = quarantine_size_mb=64.......`

src/clusterfuzz/_internal/platforms/android/sanitizer.py
ParisMeuleman: +1. I'll modify the existing log and this diminished severity one.

ParisMeuleman:
> This logging change LGTM
> 
> As far as a long-term solution, it would be nice to follow the [existing convention](https://github.com/google/clusterfuzz/blob/e2c5925a92fb3bd0c63523a33bdca3cc89f0f651/src/clusterfuzz/_internal/system/environment.py#L889-L904) of setting the `ADDITIONAL_<sanitizer>_OPTIONS`.
> 
> This will keep things the same for hwasan as they are for other sanitizers. We can set the options via configuration, for example, using the ClusterFuzz job definitions.
> 
> Take the `engine_asan` template as an example, which sets `ADDITIONAL_ASAN_OPTIONS = quarantine_size_mb=64.......`

So using the additional_hwasan_options sounds good. Our hwasan job actually had some (but mislabelled).
The issue here is that we are not using engine fuzzers, which as I understand have their own way of setting options. Instead we have a blackbox fuzzer where we:
  - generate files (fuzzer's job)
  - serve files over HTTP[s], with port forwarding to the host
  - open chrome (a hwasan build) on one of those file
  - observe the result
  - repeat

Opening Chrome with those hwasan options set is the challenge:
  - We have the option to add a wrap.sh script in the APK, but this means the options need to be decided at build time. 
  - Have a wrap.sh set using ```adb shell setprop wrap.$PACKAGE_NAME```. This seems more flexible and seems to work with my local tests.

PR #3864, opened 2024-03-14
Make test_for_reproducibility uworker safe.

jonathanmetzman:
All tasks that take a testcase as an argument use this function.

### DIFF
--- a/src/clusterfuzz/_internal/protos/uworker_msg.proto
+++ b/src/clusterfuzz/_internal/protos/uworker_msg.proto
@@ -114,7 +114,11 @@ message Input {
   optional string module_name = 18;
   optional google.protobuf.Timestamp preprocess_start_time = 19;
   repeated string global_blacklisted_functions = 20;
-  optional google.datastore.v1.Entity fuzz_target = 21;
+  optional TestCaseManagerInput testcase_manager_input = 21;
+}
+
+message TestCaseManagerInput {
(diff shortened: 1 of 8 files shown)


jonathanmetzman:
Follow up to finish testcase_manager.py https://github.com/google/clusterfuzz/pull/3865

src/clusterfuzz/_internal/protos/uworker_msg.proto
```diff
@@ -114,6 +114,11 @@ message Input {
   optional string module_name = 18;
   optional google.protobuf.Timestamp preprocess_start_time = 19;
   repeated string global_blacklisted_functions = 20;
+  optional TestCaseManagerInput testcase_manager_input = 21;
+}
+
+message TestCaseManagerInput {
+  optional google.datastore.v1.Entity fuzz_target = 1;
```
oliverchang, line 121: Would it make sense to consolidate the `fuzz_target` here with what's in FuzzTaskInput? 

e.g. either `FuzzTaskInput` would use `TestCaseManagerInput`, or we move `fuzz_target` up into `Input` ?

jonathanmetzman: I think hoisting it into Input makes sense.

oliverchang (approved):
LGTM assuming we hoist `fuzz_target` up into `Input`.

PR #3865, opened 2024-03-14
Make test_for_crash_with_retries uworker_safe

jonathanmetzman:


jonathanmetzman:
This should land after https://github.com/google/clusterfuzz/pull/3864 I broke this up because both PRs are risky in that they touch all of ClusterFuzz's important functionality.

PR #3870, opened 2024-03-15
Reland 00c0b532ef5a2ccd4b390304db11cb62414b378d

jonathanmetzman:
Most of the failures were either unrelated or temporary due to change in proto and updated clients not using the new proto.


jonathanmetzman:
One bug was fixed in this reland: https://github.com/google/clusterfuzz/pull/3870/commits/b29d1bf589a64a2e713d4b997d2b2e5d403231c9#diff-4e357c1cb796c84a085445e5c0182c2a0bc5d46c0271c1fe594a7d41947ed9d8R586

PR #3871, opened 2024-03-15
Dont update layouttests on batch

jonathanmetzman:
It's probably only needed for fuzzing, and right now it breaks untrusted service account.


oliverchang (commented):
LGTM

PR #3877, opened 2024-03-17
Move a privileged operation from main into preprocess

jonathanmetzman:


jonathanmetzman:
Failing test is spurious.

PR #3886, opened 2024-03-18
Make data bundle setup work on untrusted.

jonathanmetzman:
Reenable untrusted service account for minimize.


oliverchang (commented):
lgtm

Issue #3893, opened 2024-03-19
Handle insufficient privileges in progression task


jonathanmetzman:
```
during handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/mnt/scratch0/clusterfuzz/src/python/bot/startup/run_bot.py", line 128, in task_loop
    sys.exit(utasks.uworker_bot_main())
  File "/mnt/scratch0/clusterfuzz/src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py", line 220, in uworker_bot_main
    uworker_main(input_download_url)
  File "/mnt/scratch0/clusterfuzz/src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py", line 202, in uworker_main
    uworker_output = utask_module.utask_main(uworker_input)
  File "/mnt/scratch0/clusterfuzz/src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py", line 676, in utask_main
    return find_fixed_range(uworker_input)
  File "/mnt/scratch0/clusterfuzz/src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py", line 551, in find_fixed_range
    update_metadata=True)
  File "/mnt/scratch0/clusterfuzz/src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py", line 361, in _testcase_reproduces_in_revision
    issue_metadata = _get_and_update_issue_metadata(testcase)
  File "/mnt/scratch0/clusterfuzz/src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py", line 313, in _get_and_update_issue_metadata
    metadata = engine_common.get_all_issue_metadata_for_testcase(testcase)
  File "/mnt/scratch0/clusterfuzz/src/clusterfuzz/_internal/bot/fuzzers/engine_common.py", line 460, in get_all_issue_metadata_for_testcase
    fuzz_target = testcase.get_fuzz_target()
  File "/mnt/scratch0/clusterfuzz/src/clusterfuzz/_internal/datastore/data_types.py", line 727, in get_fuzz_target
    target = ndb.Key(FuzzTarget, name).get()
  File "/mnt/scratch0/clusterfuzz/src/third_party/google/cloud/ndb/_options.py", line 89, in wrapper
    return wrapped(*pass_args, **kwargs)
  File "/mnt/scratch0/clusterfuzz/src/third_party/google/cloud/ndb/utils.py", line 110, in positional_wrapper
[...]

PR #3894, opened 2024-03-19
Handle fuzz target metadata correctly and in an untrusted manner.

jonathanmetzman:
Previously analyze task was not retrieving the fuzz target metadata properly (this needs the build, analyze was doing it in postprocess) and progression was doing it in an unprivileged way (using the db in utask_main). Change this so we can retrieve the metadata in utask_main and save it in postprocess.
Leave a TODO in corpus_pruning which isn't migrated yet.


jonathanmetzman:
Fixes: https://github.com/google/clusterfuzz/issues/3893

src/clusterfuzz/_internal/bot/fuzzers/engine_common.py
```diff
@@ -451,16 +451,12 @@ def get_all_issue_metadata(fuzz_target_path):
   return metadata
 
 
-def get_all_issue_metadata_for_testcase(testcase):
+def get_fuzz_target_issue_metadata(fuzz_target):
```
alhijazi, line 454: We used to check whether the fuzz_target is None, how come we don't anymore?

jonathanmetzman: We do now, this was clearly a mistake. I thought it would be handled before calling this.
https://github.com/google/clusterfuzz/pull/3898

alhijazi: Great, thanks!

alhijazi (commented):
LGTM

PR #3904, opened 2024-03-21
Dont omit stats

jonathanmetzman:
This should fix a recent breakage where stats are being recorded, but are essentially empty. When I made this mistake, I assumed all of the stats fields in the class definitions were all that mattered, in fact they are a minimum, and many more stats matter.


jonathanmetzman:
CC @mi-ac

oliverchang (approved):

PR #3907, opened 2024-03-26
Update pipenv to latest version 2023.12.1

letitz:
This would allow use of package categories, which might help simplify dependency management.

Curious to see whether this breaks things or not, most notably CI.


letitz:
Did not realize there was an existing `pipenv` branch, closing this PR in favor of https://github.com/google/clusterfuzz/pull/3908.

PR #3908, opened 2024-03-26
Update pipenv to latest version 2023.12.1

letitz:
This would allow use of package categories, which might help simplify dependency management.

Curious to see whether this breaks things or not, most notably CI.

### DIFF
--- a/local/install_python_deps_linux.bash
+++ b/local/install_python_deps_linux.bash
@@ -16,7 +16,7 @@
 
 # Setup pipenv and install python dependencies.
 echo If this fails, you may need to build older Python from source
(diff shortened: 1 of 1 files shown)


letitz:
I want to test it out locally to see if it works on my machine (tm), then I guess we can try merging!

letitz:
I can confirm that `local/install_deps.bash` works with the updated pipenv. I'll see about updating the version used on Mac too.

letitz:
On Mac `local/Brewfile` is used to install pipenv using homebrew, with no version specification.

I think this is good to merge!

letitz:
Friendly ping!

jonathanmetzman (approved):

PR #3911, opened 2024-03-26
Fix serializing additional issue metadata

alhijazi:


src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -599,6 +600,10 @@ def get_issue_description(testcase,
 
   # Add additional body text from metadata.
   issue_metadata = testcase.get_metadata('issue_metadata', {})
+  try:
+    issue_metadata = json.loads(issue_metadata)
+  except:
+    pass
```
jonathanmetzman, line 606: This seems pretty hacky. I think we need to store the metadata field as something other than a string:string map.

alhijazi: I will think about a better way to do this, I don't have many ideas as the issue_metadata could be a dictionary of dictionaries

jonathanmetzman: I think we need to make the field a json-serialized string instead.

alhijazi: https://github.com/google/clusterfuzz/pull/3915

jonathanmetzman:
Superseded by https://github.com/google/clusterfuzz/pull/4174

PR #3914, opened 2024-03-27
Fix query when throttling issue filing per project.

jonathanmetzman:
"and" is not the right way to combine query conditions. "X and Y" is Y and will only filter on Y.
Add a test to verify query correctness.
This fixes a bug preventing issue filing for some of our users.

Also log throttling as an error. It's a pretty serious step that deserves attention if we have to throttle issue filing.

### DIFF
--- a/src/clusterfuzz/_internal/cron/triage.py
+++ b/src/clusterfuzz/_internal/cron/triage.py
@@ -400,14 +400,14 @@ class Throttler:
     """Gets the number of bugs that have been filed for a given job
     within a given time period."""
     return data_types.FiledBug.query(
-        data_types.FiledBug.job_type == job_type and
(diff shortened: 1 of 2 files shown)


oliverchang (commented):
awesome  find.  lgtm

src/clusterfuzz/_internal/cron/triage.py
```diff
@@ -461,7 +461,7 @@ def should_throttle(self, testcase):
       if count_per_job < job_bugs_filing_max:
         self._bug_filed_per_job_per_24hrs[testcase.job_type] = count_per_job + 1
         return False
-      logs.log(
+      logs.log_error(
```
oliverchang, line 464: hmm actually, not convinced  these  should be errors though. should  we  make them WARN instead?

PR #3916, opened 2024-03-27
Make storage module more flexible by accpeting more file-like objects.

jonathanmetzman:
Do this as part of landing: https://github.com/google/clusterfuzz/pull/3834/files


oliverchang (approved):
thanks!

PR #3919, opened 2024-04-02
Update severity field on existing Google IssueTracker issues

letitz:
This seems to have never worked.

It's unclear to me whether we should support the label being removed as well or not. I can't find out whether the severity field is translated to a severity label when reading back from the tracker, in which case edits to remove the label should probably be reflected in the severity field. Otherwise, I think we can only add the label.

Also adds an error log in the case where there are too many severity labels, so we can notice and diagnose such issues.

Doing this revealed a subtle bug in the existing custom handling code for other "fake" labels, like OS. Label values are found with a case-sensitive match in `_extract_label()` or `_get_labels()`, but removed with a case-insensitive match in `LabelStore.remove_by_prefix()`. This should probably be fixed, but I did not want to change existing behavior beyond the scope of this fix.

Fixes https://crbug.com/332396351.

### DIFF
--- a/src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
+++ b/src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
@@ -17,6 +17,9 @@
 
 import datetime
 import enum
+from typing import List
+from typing import Optional
+from typing import Sequence
 import urllib.parse
 
 from google.auth import exceptions
@@ -53,7 +56,8 @@ class IssueTrackerPermissionError(IssueTrackerError):
   """Permission error."""
 
 
-def _extract_all_labels(labels, prefix):
+def _extract_all_labels(labels: issue_tracker.LabelStore,
+                        prefix: str) -> List[str]:
   """Extract all label values."""
   results = []
   labels_to_remove = []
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
Could someone from Chrome help review this before me?

letitz:
Sured, adding @alhijazi for a first pass.

src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
```diff
@@ -625,10 +679,10 @@ def save(self, new_comment=None, notify=True):
       if foundin_values:
         self._data['issueState']['foundInVersions'] = foundin_values
 
-      severity_text = _extract_label(self.labels, 'Security_Severity-')
-      logs.log('google_issue_tracker: severity_text: %s' % severity_text)
-      severity = _get_severity_from_crash_text(severity_text)
-      self._data['issueState']['severity'] = severity
+      severity = _get_severity_from_labels(self.labels) or _DEFAULT_SEVERITY
+      if severity is not None:
```
alhijazi, line 683: severity will always be not None here.

alhijazi (approved):

src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
letitz: Oh right, missed this in the latest commit. Fixed.

PR #3922, opened 2024-04-05
[fuzztest] Fix fuzztest replay for centipede engine

mi-ac:
This fixes reproduction when fuzztests are executed via the centipede engine (currently the default in Chromium's centipede jobs). Reproduction requires setting an environment variable FUZZTEST_REPLAY with the input, similar to the native googlefuzztest engine.

The tests simulate a fuzz-test target using a bash script and validate a typical check failure similar to:
https://clusterfuzz.com/testcase-detail/5164253664706560

Google bug: https://crbug.com/332679477


mi-ac:
I wonder if integration tests are currently executed at all? I ran them locally with:

TEST_COVERAGE_BUCKET=/home/machenbach/v8/github/clusterfuzz/bla TEST_BACKUP_BUCKET=/home/machenbach/v8/github/clusterfuzz/bla  TEST_QUARANTINE_BUCKET=/home/machenbach/v8/github/clusterfuzz/bla  TEST_CORPUS_BUCKET=/home/machenbach/v8/github/clusterfuzz/bla  INTEGRATION=1 python butler.py py_unittest -t core -m -p centipede_engine_*

My new test passes like this, but an existing test failed. But I have no idea what to properly set in all these test buckets... so maybe it's just that.

mi-ac:
Note, detecting a fuzztest target by string.endswith is maybe not very future proof, but should cover the current fuzztests that exist. It's also used in the code here to determine fuzztest targets:
https://github.com/google/clusterfuzz/blob/bb28a45a205493e5bc6dfcfeb63102d4679f32cd/src/clusterfuzz/_internal/bot/fuzzers/utils.py#L61

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
```diff
@@ -267,8 +267,13 @@ def reproduce(self, target_path, input_path, arguments, max_time):  # pylint: di
           f':{constants.RSS_LIMIT_MB_FLAGNAME}={rss_limit}'
           f':{constants.TIMEOUT_PER_INPUT_FLAGNAME}={timeout}:')
 
-    runner = new_process.UnicodeProcessRunner(sanitized_target, [input_path])
-    result = runner.run_and_wait(timeout=max_time)
+    if sanitized_target.endswith(('_fuzztest', '_fuzz_test')):
```
jonathanmetzman, line 270: Can we do something else less hacky here? Can you check for FUZZTEST_MODE env var? I can set the job to use it.

mi-ac: I can do that and it would be fine for V8. For Chromium it would require some more follow up work, since there's currently one shared job for fuzztest and legacy-libfuzzer centepede. Setting a flag on the job level would require this to have two jobs and probably some more logic to not pick fuzztests in one job and only fuzztests in the other.

mi-ac: But maybe Chromium is fine to switch to the googlefuzztest engine anyways? Then no further fixes are required.

mi-ac: Done. PTAL

jonathanmetzman (approved):
lgtm

PR #3923, opened 2024-04-05
Indicate the latest revision the crash reproduces on

alhijazi:
It is misleading to say that the job reproduces on tip of tree when in reality it could be running on some stale builds.


jonathanmetzman (approved):
LGTM

src/clusterfuzz/_internal/cron/cleanup.py
```diff
@@ -761,7 +761,7 @@ def notify_closed_issue_if_testcase_is_open(policy, testcase, issue):
   if issue.status in [policy.status('fixed'), policy.status('verified')]:
     issue_comment = (
         f'ClusterFuzz testcase {testcase.key.id()} is still reproducing '
-        f'on the latest available build{crash_revision}.'
+        f'on the latest available build {crash_revision}.'
```
alhijazi, line 764: Btw this was intentional, not to leave a space between `build` and `.` when crash revision is empty.

alhijazi: space added on line 760

PR #3925, opened 2024-04-08
Re-use existing symbolized release build if possible

letitz:
Symbolize task first downloads a regular build to run under ASAN a few times with larger redzones in an attempt to refine the crash type, before downloading symbolized release and debug builds to gather symbolized stack traces. Sometimes, as seen in a Chrome fuzzing job, the regular build is the same as the symbolized release build: downloading and extracting it twice wastes valuable disk space.

This changes makes `SymbolizedBuild` re-use the existing build if it determines it to be compatible: that is, it is extracted to the same base build directory, shares the same revision and came from the same URL. I believe (though I am not certain) that both `RegularBuild` and `SymbolizedBuild` extract their zips in similar-enough fashions such that the build can really be re-used.

See also https://crbug.com/333014970#comment2.

### DIFF
--- a/src/clusterfuzz/_internal/build_management/build_manager.py
+++ b/src/clusterfuzz/_internal/build_management/build_manager.py
@@ -19,6 +19,7 @@ import re
 import shutil
 import subprocess
 import time
+from typing import Optional
 
 from clusterfuzz._internal.base import errors
 from clusterfuzz._internal.base import utils
@@ -328,9 +329,9 @@ class Build(BaseBuild):
   """Represent a build type at a particular revision."""
 
   def __init__(self,
-               base_build_dir,
-               revision,
-               build_prefix='',
+               base_build_dir: str,
+               revision: int,
+               build_prefix: str = '',
                fuzz_targets=None):
     super().__init__(base_build_dir)
     self.revision = revision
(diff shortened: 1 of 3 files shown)


letitz:
Alternative approaches to consider:

- the ASAN redzone logic could move to analyze task instead, since it already has the right build and is conceptually in charge of determining the crash type
- builds could be extracted to content-addressed storage directories:
  - in other words, make the build dir a hash of the zip and extract arguments
  - this is perhaps difficult due to the unpack code's reliance on environment variables - identifying the set of arguments that affects the unpacking process is perhaps hard

letitz:
@alhijazi PTAL for a first pass. Does this strike you as too hacky? I could be convinced that moving the redzone stuff to analyze task is the right thing to do instead.

jonathanmetzman:
> Symbolize task first downloads a regular build to run under ASAN a few times with larger redzones in an attempt to refine the crash type, before downloading symbolized release and debug builds to gather symbolized stack traces. Sometimes, as seen in a Chrome fuzzing job, the regular build is the same as the symbolized release build: downloading and extracting it twice wastes valuable disk space.
> 
> This changes makes `SymbolizedBuild` re-use the existing build if it determines it to be compatible: that is, it is extracted to the same base build directory, shares the same revision and came from the same URL. I believe (though I am not certain) that both `RegularBuild` and `SymbolizedBuild` extract their zips in similar-enough fashions such that the build can really be re-used.
> 
> See also https://crbug.com/333014970#comment2.

I'm not sure this will ever be useful. We launch a new instance each time we symbolize, so there will never be an earlier build on disk.

letitz:
There is, because symbolize task first downloads a regular (possible non-symbolized build) first.

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -764,22 +765,71 @@ def setup(self):
...
   def build_dir(self):
     return self._build_dir
 
+  @property
+  def reuse_release_build(self):
+    return self._reuse_release_build
+
   def _unpack_builds(self):
     """Download and unpack builds."""
     if not shell.remove_directory(self.build_dir, recreate=True):
```
alhijazi, line 835: Wouldn't this delete the existing symbolized build (if any)?

alhijazi: My bad, self.build_dir is a function call that returns `_build_dir`. The existing build is not present under this directory.

alhijazi:
> Alternative approaches to consider:
> 
> * the ASAN redzone logic could move to analyze task instead, since it already has the right build and is conceptually in charge of determining the crash type
> * builds could be extracted to content-addressed storage directories:
>   
>   * in other words, make the build dir a hash of the zip and extract arguments
>   * this is perhaps difficult due to the unpack code's reliance on environment variables - identifying the set of arguments that affects the unpacking process is perhaps hard

I would propose a 3rd approach:
Given that we always launch a new instance for symbolize (as Jonathan mentioned), we could just remove the regular build after we're done using it (this would save the necessary disk space).

I took a look at the solution, I think it could work but I find it risky. Just deleting the downloaded regular build should solve the disk space issue with minimal code changes.

letitz:
I like your approach, it's much simpler. It's still a bit sad that we would download and extract the same zip all over again - that can easily waste an hour. Still, that's no worse than today and would solve the issue at hand! I'll try that in a new PR.

letitz:
Abandoning this since #3932 landed.

PR #3926, opened 2024-04-08
More logging in `Build._unpack_build()`

letitz:
- Adds a log at the start stating input URL and output directory
- Changes log calls to use f-strings
- Catches exceptions with a name in order to print details in error logs

WARNING: There is a behavior change here, since we no longer use a bare `except` we will no longer catch exceptions like `KeyboardInterrupt`. I think this is better, as we probably should not be catching such exceptions without re-raising them immediately.


letitz:
@paulsemel PTAL, this is ready for a first pass.

paulsemel (approved):
Thanks, I love this. I think `build_manager.py` was lacking some logs.

jonathanmetzman (approved):
LGTM

PR #3927, opened 2024-04-08
Refactor `regression_task.found_regression_near_extreme_revisions`.

letitz:
Prework for changing the behavior of `check_earliest_revisions()` to perform a binary search for the first good build instead of just checking the first few and erroring out if unsuccessful.

See also https://crbug.com/333014970#comment2

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/regression_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/regression_task.py
@@ -147,34 +147,30 @@ def _testcase_reproduces_in_revision(
   return result.is_crash(), None
 
 
-def found_regression_near_extreme_revisions(
+def check_latest_revisions(
     testcase: data_types.Testcase,
     testcase_file_path: str,
     job_type: str,
-    revision_list: List[int],
-    min_index: int,
-    max_index: int,
+    revision_range: List[int],
     fuzz_target: Optional[data_types.FuzzTarget],
-    regression_task_output: uworker_msg_pb2.RegressionTaskOutput,  # pylint: disable=no-member
+    output: uworker_msg_pb2.RegressionTaskOutput,  # pylint: disable=no-member
 ) -> Optional[uworker_msg_pb2.Output]:  # pylint: disable=no-member
-  """Test to see if we regressed near either the min or max revision.
-  Returns a uworker_msg_pb2.Output or None.
-   The uworker_msg_pb2.Output contains either:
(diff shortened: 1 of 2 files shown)


letitz:
@alhijazi PTAL

src/clusterfuzz/_internal/bot/tasks/utasks/regression_task.py
```diff
@@ -182,52 +181,76 @@ def found_regression_near_extreme_revisions(
...
+        revision,
         regression_task_output,
         fuzz_target,
         should_log=False)
+
     if error:
+      # Skip bad build errors only.
       if error.error_type == uworker_msg_pb2.REGRESSION_BAD_BUILD_ERROR:
-        # If we find a bad build, potentially try another.
-        if min_index + 1 >= max_index:
```
alhijazi, line 211: I think a subtle issue we're no longer handling is when the revision_range is small enough to have an overlap between `revision_range[:EXTREME_REVISIONS_TO_TEST]` and `revision_range[-EXTREME_REVISIONS_TO_TEST - 1:-1]`
See also line 169 in the initial file

letitz: Thanks for pointing it out! I think the handling has not changed, actually.

On line 171 in the new file:

```py
  for revision in reversed(revision_range[-EXTREME_REVISIONS_TO_TEST - 1:-1])
```

This will return a slice with less than `EXTREME_REVISIONS_TO_TEST` elements if `revision_range` had less than `EXTREME_REVISIONS_TO_TEST + 1` elements in the first place. I think that's the same handling as line 169 in the original file.

Similarly for `checker_earliest_revisions()`, we do:

```py
  revision_range = revision_range[:EXTREME_REVISIONS_TO_TEST]
  for revision in revision_range:
```

Which will iterate over less than `EXTREME_REVISIONS_TO_TEST` revisions if `revision_range` is too small, just like it was handled on line 211 originally.

letitz:
```sh
$ python3            
[...]                                                                                                                     
>>> [1, 2, 3][-4:-1]                                                                                                            
[1, 2]
```

alhijazi (approved):

letitz:
@jonathanmetzman PTAL, this is a prereq for https://github.com/google/clusterfuzz/pull/3934 which aims to fix an issue where regression task can get stuck.

letitz:
Very late ping, this is ready to land again.

PR #3932, opened 2024-04-09
Delete regular build before symbolization.

letitz:
Alternative attempt to fix issues identified in https://crbug.com/333014970, as suggested by @alhijazi on #3925. 

After the bot downloads a regular build and uses it to try reproducing the crash with larger ASAN redzone sizes, this change makes the bot delete said regular build. It will then download a symbolized release build and (optionally) a symbolized debug build.

This is still a bit silly in the case where the regular build and symbolized release build are one and the same: we download and unpack the same (potentially-large) zip twice. However, it is much less complex and risky than #3925, which is a good tradeoff.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/symbolize_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/symbolize_task.py
@@ -107,7 +107,7 @@ def utask_main(uworker_input):
     build_revision = testcase.crash_revision
 
   # Set up a custom or regular build based on revision.
-  build_manager.setup_build(build_revision)
+  build = build_manager.setup_build(build_revision)
 
   # Get crash revision used in setting up build.
   crash_revision = environment.get_value('APP_REVISION')
@@ -157,6 +157,10 @@ def utask_main(uworker_input):
           break
 
       redzone /= 2
+
+  # We no longer need this build, delete it to save some disk space. We will
+  # download a symbolized release build to perform the symbolization.
+  build.delete()
 
   # We should have atleast a symbolized debug or a release build.
   symbolized_builds = build_manager.setup_symbolized_builds(crash_revision)


letitz:
Thanks for the review! Waiting for all tests to finish running before sending over to Jonathan.

letitz:
Cross-posting from https://333014970:

I started wondering if this would help - after all, the task *did* try to make space for the new download before giving up and failing. TL;DR: it would help, the current code does not delete the regular build.

The logic starts in `build_manager._make_space()` [1], which calls `build_manager._evict_build()` [2] in a loop until enough space has been freed. This iterates over all directories in the `BUILD_DIR` environment variable, choosing the least-recently-used build directory apart from any ancestor to the current build. This will skip the symbolized build's base build directory [3], which is the same as the regular build's base build directory in our case (which is also one the reasons why we could reuse it). In our case, `_make_space()` fails to find any directory to delete - as evidenced by the absence of logs stating that a directory was evicted [4].

[1] https://github.com/google/clusterfuzz/blob/e5e8cebaa27e859eba1143bbfd1378401f106ddb/src/clusterfuzz/_internal/build_management/build_manager.py#L438-L443  
[2] https://github.com/google/clusterfuzz/blob/e5e8cebaa27e859eba1143bbfd1378401f106ddb/src/clusterfuzz/_internal/build_management/build_manager.py#L107  
[3] https://github.com/google/clusterfuzz/blob/e5e8cebaa27e859eba1143bbfd1378401f106ddb/src/clusterfuzz/_internal/build_management/build_manager.py#L132  
[4] https://github.com/google/clusterfuzz/blob/e5e8cebaa27e859eba1143bbfd1378401f106ddb/src/clusterfuzz/_internal/build_management/build_manager.py#L149

letitz:
@jonathanmetzman PTAL, I think this will enable us to revert the temporary disk size bump we landed to get symbolize task to succeed.

jonathanmetzman (approved):
LGTM

PR #3933, opened 2024-04-09
Improve logging when reproducing crashes with larger ASAN redzones.

letitz:
- Adds a bunch of logs to clarify what is happening at each step
- Removes `sym_crash_type`, which was always equal to `testcase.crash_type`:
  - `testcase.crash_type` seems to never be mutated after it is assigned to `sym_crash_type`
  - `sym_crash_type` is never set to a different value:
     - `sym_crash_type` is only assigned `state.crash_type` if `state.crash_type == testcase.crash_type`
     - in other words, `sym_crash_type` is only ever assigned the same value it initially was set to

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/symbolize_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/symbolize_task.py
@@ -140,9 +140,10 @@ def utask_main(uworker_input):
         security_flag = crash_result.is_security_issue()
 
         if crash_analyzer.ignore_stacktrace(state.crash_stacktrace):
-          logs.log(f'Skipping crash with ASAN redzone size {redzone}: ' +
-                   'stack trace should be ignored.',
-                   stacktrace=state.crash_stacktrace)
+          logs.log(
+              f'Skipping crash with ASAN redzone size {redzone}: ' +
+              'stack trace should be ignored.',
+              stacktrace=state.crash_stacktrace)
         elif security_flag != testcase.security_flag:
           logs.log(f'Skipping crash with ASAN redzone size {redzone}: ' +
                    f'mismatched security flag: old = {testcase.security_flag}, '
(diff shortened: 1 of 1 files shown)


paulsemel (approved):
SGTM!

src/clusterfuzz/_internal/bot/tasks/utasks/symbolize_task.py
```diff
@@ -184,7 +196,7 @@ def utask_main(uworker_input):
                                  old_crash_stacktrace, sym_crash_state))
 
   symbolize_task_output = uworker_msg_pb2.SymbolizeTaskOutput(
-      crash_type=sym_crash_type,
+      crash_type=testcase.crash_type,
```
jonathanmetzman, line 199: Hmm..shouldn't this be state.crash_type in some cases?

letitz: `state.crash_type` is always equal to `testcase.crash_type`, see line 151.

jonathanmetzman: I'm confused. Doesn't 151 imply they are not always the same?

letitz: It's confusing!

On line 151 what happens is that we *skip* the new big-redzone crash if its type does not match the original crash type from `testcase`. In other words, we only use an ASAN big-redzone crash if its type is the same as `testcase.crash_type`.

Overall, the only parameters that are allowed to change are the address and crash state. I don't really understand what the crash state entails, so I can't tell if the crash type being fixed means the comment above on line 121 is a lie or not:

```
# A UAF might actually turn out to be OOB read/write with a bigger redzone.
```

letitz: I removed it because I think it was useless. I tried to lay out my reasoning in the PR description, but I had a hard time making it really clear. Basically, I think we were assigning the same value twice to `sym_crash_type`.

jonathanmetzman: Oh sorry for not reading the description well enough then.

letitz: No worries! Thanks for checking!

jonathanmetzman (approved):
LGTM

PR #3934, opened 2024-04-09
Search for min revision exponentially backwards

letitz:
This PR radically reworks the way regression task starts off, while leaving bisection untouched.

Before this change, on first run, regression task would:

1. iterate linearly through a handful of the latest revisions to see if the regression was very recent
   1. if so, return that regression range
2. iterate linearly through a handful of the earliest revisions to see if the testcase reproduced there
   1. if so, then return regression range `0:earliest_revision`
   2. if none of these earliest builds were good, fail with `REGRESSION_BAD_BUILD_ERROR` _and get stuck_
3. start bisecting

Step 2.ii. was responsible for lots of issues for Chrome's clusterfuzz instance as the earliest builds would sometimes stop working due to docker image changes (usually related to system libs).

This change reworks this to broadly behave as follows:

1. Search exponentially backwards from the max revision for a suitable min revision
   1. If no such revision can be found, find the first good revision and return regression range `0:earliest good revision`
   2. If we time out, set up state on the testcase to remember where we left off and reschedule regression task
2. Start bisecting 

Note that this means that there are now 3 possible states when we start regression task:

1. Starting from scratch: `last_regression_max` and `last_regression_min` are `None`
2. (new) Resuming the search for a min revision: `last_regression_max` is an integer, `last_regression_min` is None
3. Resuming bisection: `last_regression_max` and `last_regression_min` are integers

Bug: https://crbug.com/396344382


letitz:
@alhijazi PTAL while I work out how to write some kind of integration test.

alhijazi (approved):

letitz:
@jonathanmetzman this is ready for review, finally :)

oliverchang:
(ignore the auto-close messages -- they were caused by a misconfiguration that should now be fixed).

letitz:
Ping! Would love to get this in sooner rather than later.

jonathanmetzman:
I'll look today

src/clusterfuzz/_internal/bot/tasks/utasks/regression_task.py
```diff
@@ -149,114 +153,197 @@ def _testcase_reproduces_in_revision(
...
-    a. The earliest good build crashes, in which case the regression range is
-       set to [0, min_good_revision).
-    b. An error occurred.
   """
-  # Test to see if we crash in the oldest revision we can run. This is a pre-
-  # condition for our binary search. If we do crash in that revision, it
-  # implies that we regressed between the first commit and our first revision,
-  # which we represent as 0:|min_revision|.
+  assert max_index >= 0, max_index
+  assert max_index < len(revision_list), max_index
```
jonathanmetzman, line 232: I think this could be a problem in google3/android where builds are deleted. I suppose it would be a somewhat rare race condition (if GCS deleted it as we were doing regession). WDYT?

src/clusterfuzz/_internal/bot/tasks/utasks/regression_task.py
```diff
@@ -527,6 +642,8 @@ def handle_regression_build_setup_error(output: uworker_msg_pb2.Output):  # pyli
       wait_time=build_fail_wait)
 
 
+# TODO: Delete this once all uworkers are past
```
jonathanmetzman, line 645: I'm not sure what this means, and can you make an issue for this?

jonathanmetzman (commented):
Mostly looks awesome, just one question.

src/clusterfuzz/_internal/bot/tasks/utasks/regression_task.py
```diff
@@ -149,114 +153,196 @@ def _testcase_reproduces_in_revision(
...
+
+    d. If another error occurred:
+
+        None, None, output
+
+      where:
+
+        output.error_type indicates the error that occurred
+        output.regression_task_output = regression_task_output
+        output.regression_task_output.last_regression_max is set
```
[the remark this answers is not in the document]
letitz: Thanks :) I just hope it accurate!

jonathanmetzman:
You want this merged now?

src/clusterfuzz/_internal/bot/tasks/utasks/regression_task.py
letitz: Right! This case is handled by the code on line 442. If the max revision we had left off at was deleted, we will choose a max revision higher than the last one we used, and in all cases confirm that the testcase reproduces at `max_index` before attempting to search for anything - see line 453.

letitz: See also `TestFindMinRevision.test_resume_missing_crash_revision` that exercises how this function is called in that case. Do you want a corresponding test in `UTaskMainTest`?

src/clusterfuzz/_internal/bot/tasks/utasks/regression_task.py
letitz: Ah yes, forgot about cleaning this up. Will do.

letitz: Done.

letitz:
I think this can indeed be merged now. Best to do it while I'm still around to debug issues :) I'd be happy to deploy tomorrow.

jonathanmetzman:
Deploying now!

PR #3935, opened 2024-04-09
Fix fuzz task run method log message

gustavogaldinoo:
Fixing log message to include actual fuzz target in log message instead of raw stirng


jonathanmetzman (approved):
LGTM

PR #3947, opened 2024-04-16
Avoid interactive prompts in docker image builds.

### LINKED ISSUE #3609: ./build.sh for docker images fails locally because of keyboard-configuration
It has an interactive prompt that never resumes even when interacted with.

letitz:
Turns out that setting `DEBIAN_FRONTEND=noninteractive` does not always work. It specifically fails when installing `keyboard-configuration`, which asks for a keyboard layout interactively.

Using debconf in this way seems to work, OTOH. I built the chromium images successfully this way.

Fixes #3609 and https://crbug.com/41493477.


jonathanmetzman (approved):
LGTM

Thanks a bunch!

PR #3952, opened 2024-04-18
Add clear error when projects.json cannot load

marktefftech:


oliverchang (approved):
thanks!

jonathanmetzman (commented):
To be clear, the purpose of this isn't to make the code more robust right?
I'm not clear on why ProjectSetupError is clearer. using the original error will be more specific. E.g. whether it's because the http request failed or the json decoding failed.

marktefftech:
> To be clear, the purpose of this isn't to make the code more robust right? I'm not clear on why ProjectSetupError is clearer. using the original error will be more specific. E.g. whether it's because the http request failed or the json decoding failed.

The goal is to make the code more robust by catching the previously uncaught exception, which looks like this:

```
File "/usr/local/lib/python3.7/json/decoder.py", line 353, in raw_decode
    obj, end = self.scan_once(s, idx)
json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 11 column 5 (char 310)
```

Since the exception is uncaught it is not clear while file (`gcs_url`) is causing the issue.

I've updated the logic to catch specifically for `json.decoder.JSONDecodeError` and include this in the `ProjectSetupError`, so we don't lose the original exception.

Does this look better?

PR #3706, opened 2024-01-30
Record task duration metrics even in case of exceptions

letitz:
Using a context manager class to wrap task execution. Improves on #3698.

PR #3955, opened 2024-04-19
Record subtask duration metric.

letitz:
Builds on #3706 to record the subtask duration on top of the end-to-end duration.


src/clusterfuzz/_internal/metrics/monitoring_metrics.py
```diff
@@ -175,6 +175,23 @@
     ],
 )
 
+UTASK_DURATION_SECS = monitor.CumulativeDistributionMetric(
```
jonathanmetzman, line 178: Shouldn't this have subtask in the name?

letitz: Fair point! The existing e2e metric does not have it, so I went for consistency. I can switch both over if you'd rather.

letitz: Done.

jonathanmetzman (approved):
lgtm

PR #3956, opened 2024-04-22
Remove migrated cron jobs from AppEngine, k8s jobs fix and tf examples

andrei-near:
- Removing cron jobs migrated to k8s workloads.
- Adding missing REDIS_HOST in k8s workflows yaml definitions. Without this, [envsubst](https://github.com/google/clusterfuzz/blob/master/src/local/butler/deploy.py#L468) will not work and workload jobs will fail.
- Adding Terraform main.tf example along with resource import command

### DIFF
--- a/configs/test/gae/prod3/cron.yaml
+++ b/configs/test/gae/prod3/cron.yaml
@@ -14,21 +14,6 @@
 
 cron:
   # Python crons start here.
-  - description: Backup
-    url: /backup
-    schedule: every day 01:47
-    target: cron-service
-
-  - description: Batch FuzzerJobs.
-    url: /batch-fuzzer-jobs
-    schedule: every 30 minutes
-    target: cron-service
-
-  - description: Build crash stats.
-    url: /build-crash-stats
-    schedule: every 5 minutes
-    target: cron-service
-
   - description: Cleanup.
     url: /cleanup
     schedule: every 1 minutes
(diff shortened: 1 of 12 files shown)


jonathanmetzman:
> * Removing cron jobs migrated to k8s workloads.
> * Adding missing REDIS_HOST in k8s workflows yaml definitions. Without this, [envsubst](https://github.com/google/clusterfuzz/blob/master/src/local/butler/deploy.py#L468) will not work .
> * Adding Terraform main.tf example along with resource import command

Is there a way we can pass REDIS_HOST to every job without explicitly doing so in each config file? I don't love the repetition for a feature we don't ned.

andrei-near:
> > * Removing cron jobs migrated to k8s workloads.
> > * Adding missing REDIS_HOST in k8s workflows yaml definitions. Without this, [envsubst](https://github.com/google/clusterfuzz/blob/master/src/local/butler/deploy.py#L468) will not work .
> > * Adding Terraform main.tf example along with resource import command
> 
> Is there a way we can pass REDIS_HOST to every job without explicitly doing so in each config file? I don't love the repetition for a feature we don't need.

An option would be changes in this [code](https://github.com/google/clusterfuzz/blob/master/src/local/butler/deploy.py#L468) but this however will not make it very obvious the k8s workloads are using Redis. I'd prefer to have it explicit in yaml that having to dig through code to find out how it is provisioned.
This PR is rather a fix for current situation when workloads will fail initialisation if REDIS_HOST env is missing but also some examples about new terraform configs required by default

Issue #3957, opened 2024-04-22
Allow custom builds in nonprivileged execution


jonathanmetzman:
We'll do this by uploading *new* custom builds to a bucket that the untrusted worker can read from.
We'll only do this in projects that will use untrusted workers.
This will make it impossible to use older custom builds. The custom builds that are still in use, I will re-upload.

I'm doing this because it's too confusing to have some linux jobs that are privileged and others that aren't especially not for a reason like this.

jonathanmetzman:
CC @pgrace-google

PR #3958, opened 2024-04-22
Upload new custom builds to a specified bucket when available.

jonathanmetzman:
This will allow the untrusted worker to access the builds without special handling.
Fixes: https://github.com/google/clusterfuzz/issues/3957

### DIFF
--- a/src/clusterfuzz/_internal/build_management/build_manager.py
+++ b/src/clusterfuzz/_internal/build_management/build_manager.py
@@ -862,7 +862,15 @@ class CustomBuild(Build):
 
     build_local_archive = os.path.join(self.build_dir,
                                        self.custom_binary_filename)
-    if not blobs.read_blob_to_disk(self.custom_binary_key, build_local_archive):
+    custom_builds_bucket = environment.get_value('CUSTOM_BUILDS_BUCKET')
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
I suppose we could force this for all users of ClusterFuzz.
Note this approach will break old builds of past custom builds. But we can reupload them to get them working (regression task might not work though, I probably need to copy them to the new bucket)

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -862,7 +862,15 @@ def _unpack_custom_build(self):
 
     build_local_archive = os.path.join(self.build_dir,
                                        self.custom_binary_filename)
-    if not blobs.read_blob_to_disk(self.custom_binary_key, build_local_archive):
+    custom_builds_bucket = environment.get_value('CUSTOM_BUILDS_BUCKET')
```
oliverchang, line 865: per the other (internal) CL: would it make more sense to make this a project.yaml config instead of an env var?

jonathanmetzman: Yup

oliverchang (approved):

PR #3961, opened 2024-04-26
Download stable build info from gcs

aditya-wazir:
Retrieves stable build info for the Cuttlefish
branch and target by downloading build info from
stable_build_info.json from gcs.

### DIFF
--- a/src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
+++ b/src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
@@ -43,8 +43,7 @@ STABLE_CUTTLEFISH_BUILD = {
 }
 
 DEFAULT_STABLE_CUTTLEFISH_BUILD_INFO = (
-    "gs://test-blobs-bucket/fuzzers/target-cuttlefish/stable_build_info.json"
-)
+    "gs://test-blobs-bucket/fuzzers/target-cuttlefish/stable_build_info.json")
 
 def execute_request_with_retries(request):
   """Executes request and retries on failure."""
@@ -170,19 +169,20 @@ def get_client():
 
 def get_stable_build_info():
   """Return stable artifact for cuttlefish branch and target."""
+  logs.log('Reached get_stable_build_info')
   stable_build_info = STABLE_CUTTLEFISH_BUILD
-  gcs_build_info_url = DEFAULT_STABLE_CUTTLEFISH_BUILD_INFO
 
-  gcs_url_override = (
-    environment.get_value('OVERRIDE_STABLE_CUTTLEFISH_BUILD_INFO')
-    )
-  if gcs_url_override:
-    gcs_build_info_url = gcs_url_override
+  try:
+    build_info_data = storage.read_data(DEFAULT_STABLE_CUTTLEFISH_BUILD_INFO)
+    if build_info_data:
+      logs.log('Loading stable cuttlefish image from %s' %
+               DEFAULT_STABLE_CUTTLEFISH_BUILD_INFO)
+      stable_build_info = json.loads(build_info_data)
+  except Exception as e:
+    logs.log_error(
+        'Error loading remote data: %s!\nUsing default build info!' % e)
 
-  build_info_data = storage.read_data(gcs_build_info_url)
-  if build_info_data:
-    stable_build_info = json.loads(build_info_data)
-
+  logs.log('Using stable cuttlefish image - %s' % stable_build_info)
   return stable_build_info
 
 def get_latest_artifact_info(branch, target, signed=False):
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
```diff
@@ -163,13 +168,32 @@ def get_client():
...
+def get_stable_build_info():
+  """Return stable artifact for cuttlefish branch and target."""
+  stable_build_info = STABLE_CUTTLEFISH_BUILD
+  gcs_build_info_url = DEFAULT_STABLE_CUTTLEFISH_BUILD_INFO
+
+  gcs_url_override = environment.get_value('OVERRIDE_STABLE_CUTTLEFISH_BUILD_INFO')
+  if gcs_url_override:
+    gcs_build_info_url = gcs_url_override
+
+  build_info_data = storage.read_data(gcs_build_info_url)
```
svasudevprasad, line 182: Can we add a log for the URL and Build info for debug purposes ?

src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
```diff
@@ -163,13 +168,34 @@ def get_client():
 
   return client
 
+def get_stable_build_info():
+  """Return stable artifact for cuttlefish branch and target."""
+  stable_build_info = STABLE_CUTTLEFISH_BUILD
+  gcs_build_info_url = DEFAULT_STABLE_CUTTLEFISH_BUILD_INFO
+
+  gcs_url_override = (
+    environment.get_value('OVERRIDE_STABLE_CUTTLEFISH_BUILD_INFO')
```
marktefftech, line 177: Where is the `OVERRIDE_STABLE_CUTTLEFISH_BUILD_INFO` environment variable being defined?

I think overall we want to avoid adding any new environment variables.

It seems like this override would be a new, optional option that can be set, but I don't think we need to provide this option.

src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
marktefftech: This also seems prone to failure (for example, if the file gets corrupted or deleted).

Can we wrap this in a try / catch block in case it fails?

src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
```diff
@@ -163,13 +168,34 @@ def get_client():
...
+    )
+  if gcs_url_override:
+    gcs_build_info_url = gcs_url_override
+
+  build_info_data = storage.read_data(gcs_build_info_url)
+  if build_info_data:
+    stable_build_info = json.loads(build_info_data)
```
marktefftech, line 184: Again, there is a chance that the data is not formatted correctly in json. Should we consider wrapping this in a try / catch block as well?

src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
```diff
@@ -163,13 +168,34 @@ def get_client():
...
     return None
 
+  if environment.is_android_cuttlefish():
+    stable_build = get_stable_build_info()
+    if stable_build:
+      return stable_build
+
   # TODO(https://github.com/google/clusterfuzz/issues/3950)
   # After stabilizing the Cuttlefish image, revert this
   stable_build = STABLE_CUTTLEFISH_BUILD
```
marktefftech, line 201: Should we go ahead and delete this, since this logic is captured above in line#173 

`stable_build_info = STABLE_CUTTLEFISH_BUILD`

svasudevprasad (approved):

marktefftech:
Looks like there is a centipede-related unit test failing, but it's not clear whether or not this breakage is associated with this change:

```
  File "/workspace/src/clusterfuzz/_internal/tests/core/bot/fuzzers/centipede/centipede_engine_test.py", line 388, in test_minimize_corpus
    self.assertLess(len(os.listdir(minimized_corpus)), 6)
AssertionError: 6 not less than 6
```

src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
```diff
@@ -164,6 +169,25 @@ def get_client():
...
 
+def get_stable_build_info():
+  """Return stable artifact for cuttlefish branch and target."""
+  logs.log('Reached get_stable_build_info')
+  stable_build_info = STABLE_CUTTLEFISH_BUILD
+
+  try:
+    build_info_data = storage.read_data(DEFAULT_STABLE_CUTTLEFISH_BUILD_INFO)
+    if build_info_data:
+      logs.log('Loading stable cuttlefish image from %s' %
```
jonathanmetzman, line 180: try to use f-strings in the future please.

jonathanmetzman (approved):
LGTM

Issue #3964, opened 2024-04-26
Ensure that impact tasks aren't run on non-chrome projects


jonathanmetzman:
They are currently being run on the proprietary google code clusterfuzz. I think schedule_impact_tasks is responsible.

PR #3966, opened 2024-04-26
Dont schedule impact task outside of chrome

jonathanmetzman:
Fixes: #3964


oliverchang (approved):

oliverchang (commented):
actually, we can probably fix this a bit better by adding this logic inside tasks_scheduler instead? that would catch more cases if there were other places scheduling this. 

This cron also should just not be running on non-chromium projects also.

Issue #3965, opened 2024-04-26
Fix issues with zero size thread pools.


jonathanmetzman:

jonathanmetzman:
@vitorguidi When you get a chance post the stack here please.

PR #3967, opened 2024-04-26
Dont use map on a zero length list

jonathanmetzman:
Fixes: https://github.com/google/clusterfuzz/issues/3965

Doing so caused this error: `ChildProcessError: [Errno 10] No child processes`

Vitor discovered this issue.


vitorguidi (approved):
Tested this out in a local clusterfuzz environment, submitting a libfuzzer job without any corpus files. Verified the crash happens w/o the fix, and stops with this new code.

oliverchang (approved):

PR #3968, opened 2024-04-29
Move bisection request to earlier in progression_task.

oliverchang:
This needs to run even in the case that the crash still reproduces on the latest revision.


jonathanmetzman (approved):
LGTM

jonathanmetzman:
Do we want to land this soon?

PR #3971, opened 2024-05-02
Add logs to redo task

svasudevprasad:
As discussed as logs to debug check if the redo tasks are getting pushed to queue

### DIFF
--- a/src/clusterfuzz/_internal/base/tasks.py
+++ b/src/clusterfuzz/_internal/base/tasks.py
@@ -720,7 +720,7 @@ def redo_testcase(testcase, tasks, user_email):
   # Log the task for debug purposes
   logs.log_info(
       f'{utils.current_date_time()} : Adding testcase id {testcase_id} \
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/base/tasks.py
```diff
@@ -747,6 +747,12 @@ def redo_testcase(testcase, tasks, user_email):
         wait_time=wait_time)
 
   if progression:
+    # Log the task for debug purposes
+    testcase.comments += '[%s] : Redo task: Adding testcase id %s task \
```
jonathanmetzman, line 751: Can you please just call logs.log_info instead? I think adding it to the comment will annoy users and will only be useful in the very short run.
Also can you hoist this up so it's done on every redo task, not just progression.

svasudevprasad: Done

jonathanmetzman:
I merged this without formatting. Try to run it in the future please :-)

PR #3973, opened 2024-05-02
[rearchitecture] Handle cases when target is not in zip.

jonathanmetzman:
We should be able to handle cases when all fuzzers in a build are renamed.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
@@ -1549,7 +1549,7 @@ class FuzzingSession:
     """Run fuzzing engine."""
     fuzz_target_name = environment.get_value('FUZZ_TARGET')
     if not fuzz_target_name:
-      raise FuzzTaskError('No fuzz targets found.')
+      raise FuzzTaskError('No fuzz targets set.')
     environment.set_value('FUZZER_NAME',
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -1564,9 +1570,13 @@ def do_engine_fuzzing(self, engine_impl):
...
     for fuzzing_round in range(environment.get_value('MAX_TESTCASES', 1)):
       logs.log(f'Fuzzing round {fuzzing_round}.')
-      result, current_fuzzer_metadata, fuzzing_strategies = run_engine_fuzzer(
-          engine_impl, self.fuzz_target.binary, sync_corpus_directory,
-          self.testcase_directory)
+      try:
+        result, current_fuzzer_metadata, fuzzing_strategies = run_engine_fuzzer(
+            engine_impl, self.fuzz_target.binary, sync_corpus_directory,
+            self.testcase_directory)
+      except FuzzTargetNotFoundError:
```
oliverchang, line 1577: can you add a comment here explaining how we handle these exceptions? I.e. we treat missing targets as not reproducing (e.g. if they were deleted). 

 Also should we be logging this?

oliverchang (approved):

Simonnnn98 (commented):
Images

PR #3988, opened 2024-05-12
Use environment.get_value instead of os.environ

jonathanmetzman:
It's more consistent and it can be important when things like ints are put into the environment.


oliverchang (commented):
lgtm

PR #3992, opened 2024-05-13
Add bot logs when a regular task is received

svasudevprasad:


jonathanmetzman (commented):
If this is about those unresponsive testcases, I think we think that the bots aren't logging to stackdriver.

PR #3996, opened 2024-05-15
Add the possibility to push tasks to swarming

alhijazi:


configs/test/swarming/swarming.yaml
```diff
@@ -0,0 +1,32 @@
+# Copyright 2023 Google LLC
```
alhijazi, line 1: Todo: Fix copyright year.

src/clusterfuzz/_internal/protos/swarming.proto
```diff
@@ -0,0 +1,1177 @@
+// Copyright 2022 The LUCI Authors. All rights reserved.
```
alhijazi, line 1: copied over from luci code, don't know what the procedure is here wrt copyright and license?
@jonathanmetzman

letitz (commented):
Looking pretty good! Mostly code-level comments, with one bigger question around how to handle the third-party dependency on luci protos.

jonathanmetzman:
Are we going to be used the docker image in swarming?

alhijazi:
> Are we going to be used the docker image in swarming?

Yes, I probably should allow specifying the image from the config similar to batch. I will do the necessary change

src/clusterfuzz/_internal/protos/swarming.proto
jonathanmetzman: Probably needs to go in protos/third_party/.
Also, how will this be kept up to date?

src/clusterfuzz/_internal/swarming_utils/swarming.py
```diff
@@ -0,0 +1,123 @@
...
+                      swarming_pb2.StringPair(key='pool', value=swarming_pool)
+                  ],
+                  cas_input_root=swarming_pb2.CASReference(
+                      cas_instance=cas_instance,
+                      digest=swarming_pb2.Digest(
+                          hash=digest_hash, size_bytes=digest_size_bytes)),
+                  execution_timeout_secs=86400,
+                  env=[
+                      swarming_pb2.StringPair(
+                          key='UWORKER_INPUT_DOWNLOAD_URL', value=download_url),
```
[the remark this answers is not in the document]
alhijazi: Indeed

src/clusterfuzz/_internal/swarming_utils/swarming.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def _get_new_task_spec(command: str, job_name: str, download_url: str):
+  """Gets the configured specifications for a swarming task."""
+  job = _get_job(job_name)
+  config_name = job.platform
+  if command == 'fuzz':
+    config_name += '-PREEMPTIBLE'
+  else:
+    config_name += '-NONPREEMPTIBLE'
```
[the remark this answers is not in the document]
alhijazi: You are right. This was copied from batch handling.
I think we would still want to specify fuzzing session duration in the config, I'll modify the config to allow that.

src/clusterfuzz/_internal/swarming_utils/swarming.py
```diff
@@ -0,0 +1,123 @@
...
+  except ValueError:
+    return False
+
+
+def get_task_name():
+  return 't-' + str(uuid.uuid4()).lower()
+
+
+def _get_job(job_name: str):
+  """Returns the Job entity named by |job_name|. This function was made to make
```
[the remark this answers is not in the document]
alhijazi: Yes I was planning to add tests already, but wanted to share the first draft asap. I will add ones with the next update.

src/clusterfuzz/_internal/protos/swarming.proto
alhijazi: I talked to swarming folks, they don't expect this will change, and even if it changes, changes will be backward compatible, meaning we wouldn't need to update this file (at least not often).
I trimmed down the file to only the parts necessary for a NewTaskRequest message.

src/clusterfuzz/_internal/base/utils.py
```diff
@@ -124,6 +124,20 @@ def fetch_url(url):
   return response.text
 
 
+@retry.wrap(
+    retries=URL_REQUEST_RETRIES,
+    delay=URL_REQUEST_FAIL_WAIT,
+    function='base.utils.post_url')
+def post_url(url, data, headers):
```
letitz, line 131: Let's document this method better:

- argument meanings, and ideally their types
- response meaning and type
- any errors it can raise
- the fact that it retries
- what timeout is applied

src/clusterfuzz/_internal/swarming/__init__.py
```diff
@@ -74,15 +67,24 @@ def _get_new_task_spec(command: str, job_name: str, download_url: str):
...
-  service_account = instance_spec['service_account']
+  service_account = instance_spec['service_account_email']
+  # If this task request slice is not scheduled after waiting this long,
+  # the task state will be set to EXPIRED.
+  expiration_secs = instance_spec['expiration_secs']
+  # Maximum number of seconds the task can run before its process is
+  # forcibly terminated and the task results in TIMED_OUT.
+  execution_timeout_secs = instance_spec['execution_timeout_secs']
+  if command == 'fuzz':
+    execution_timeout_secs = swarming_config.get('fuzzing_session_duration')
```
letitz, line 78: If this is set to the exact duration of the fuzzing session used by CF, won't this mean the task will always time out? I assume CF will run for just a little longer than this duration.

src/clusterfuzz/_internal/swarming/__init__.py
```diff
@@ -93,13 +95,14 @@ def _get_new_task_spec(command: str, job_name: str, download_url: str):
...
+                  execution_timeout_secs=execution_timeout_secs,
                   env=[
+                      swarming_pb2.StringPair(key='UWORKER', value='True'),
+                      swarming_pb2.StringPair(key='SWARMING_BOT', value='True'),
                       swarming_pb2.StringPair(
-                          key='UWORKER_INPUT_DOWNLOAD_URL', value=download_url),
-                      swarming_pb2.StringPair(key='UWORKER', value=True),
-                      swarming_pb2.StringPair(key='SWARMING_BOT', value=True)
-                  ]))
+                          key='DOCKER_IMAGE', value=docker_image)
```
letitz, line 103: Is this environment variable required? If not, maybe we should skip specifying it when `instance_spec['docker_image']` is None.

src/clusterfuzz/_internal/swarming/__init__.py
```diff
@@ -118,6 +121,4 @@ def push_swarming_task(command, download_url, job_type):
   }
   swarming_server = _get_swarming_config().get('swarming_server')
   url = f'https://{swarming_server}/prpc/swarming.v2.Tasks/NewTask'
-  r = requests.post(
-      url=url, data=task_spec.SerializeToString(), headers=headers)
-  logs.log(r.status_code)
+  utils.post_url(url=url, data=task_spec.SerializeToString(), headers=headers)
```
letitz, line 124: I'm surprised this works. I would have expected the HTTP endpoint to ingest JSON, but we're passing a binary-serialized protobuf?

src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/utasks_test.py
```diff
@@ -40,9 +40,11 @@ class TworkerPreprocessTest(unittest.TestCase):
   def setUp(self):
     monitor.metrics_store().reset_for_testing()
     helpers.patch(self, [
+        'clusterfuzz._internal.swarming.is_swarming_task',
         'clusterfuzz._internal.bot.tasks.utasks.uworker_io.get_uworker_output_urls',
         'clusterfuzz._internal.bot.tasks.utasks.uworker_io.serialize_and_upload_uworker_input',
     ])
+    self.mock.is_swarming_task.return_value = False
```
letitz, line 47: We should have tests for the swarming case as well.

src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
```diff
@@ -0,0 +1,75 @@
...
+from clusterfuzz._internal.tests.test_libs import test_utils
+
+
+@test_utils.with_cloud_emulators('datastore')
+class GetSpecFromConfigTest(unittest.TestCase):
+  """Tests for get_spec_from_config."""
+
+  def setUp(self):
+    self.maxDiff = None
+    helpers.patch(self, ['clusterfuzz._internal.swarming._get_task_name'])
```
letitz, line 31: Instead of mocking out a private helper, can we instead do:

```python
spec = _get_new_task_spec(...)
self.assertRegex(spec.name, 't-[a-z0-9\-]*')  # Or even just "starts with `t-`"

spec.name = ''
self.assertEquals(spec, expected_spec)  # Where `expected_spec` has an empty name
```

src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
```diff
@@ -0,0 +1,75 @@
...
+
+from clusterfuzz._internal import swarming
+from clusterfuzz._internal.datastore import data_types
+from clusterfuzz._internal.protos import swarming_pb2
+from clusterfuzz._internal.tests.test_libs import helpers
+from clusterfuzz._internal.tests.test_libs import test_utils
+
+
+@test_utils.with_cloud_emulators('datastore')
+class GetSpecFromConfigTest(unittest.TestCase):
```
letitz, line 26: There are a couple of untested branches in `_get_new_task_spec()`:

- when there is no matching instance spec, we raise a `ValueError`
- when the command is `"fuzz"`, we use the special fuzzing session duration
- when the docker image is unspecified on the instance spec, we use an empty string instead

src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
```diff
@@ -0,0 +1,75 @@
...
+"""Swarming tests."""
+import base64
+import unittest
+
+from clusterfuzz._internal import swarming
+from clusterfuzz._internal.datastore import data_types
+from clusterfuzz._internal.protos import swarming_pb2
+from clusterfuzz._internal.tests.test_libs import helpers
+from clusterfuzz._internal.tests.test_libs import test_utils
+
```
letitz, line 23: We should also test `push_swarming_task()`.

src/clusterfuzz/_internal/protos/swarming.proto
letitz: I see. Can we add a monitoring metric (maybe in a followup PR) to record the rate of successful and unsuccessful task pushes? It would be good to know in real time if we start to fail to schedule tasks due to a change in the swarming API, for example.

src/clusterfuzz/_internal/cron/project_setup.py
```diff
@@ -915,6 +915,7 @@ def _sync_job(self, project, info, corpus_bucket_name, quarantine_bucket_name,
               f'{key} = {str(value).encode("unicode-escape").decode("utf-8")}\n'
           )
 
+      job.requires_gpu = bool(job.get_environment().get('REQUIRES_GPU'))
```
letitz, line 918: As I just discovered earlier, this file is used to sync config files to the database. Was the intent to set the `requires_gpu` field on jobs when UI changes are made to set the `REQUIRES_GPU`?

src/clusterfuzz/_internal/swarming/__init__.py
alhijazi: Maybe I should rename this to `fuzzing_task_duration` instead because that's what I meant by this config entry

src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
alhijazi: it ended up being convenient for the `push_swarming_task` test so I had to keep it

src/clusterfuzz/_internal/swarming/__init__.py
alhijazi: Thanks for catching that, I had made a mistake here.

src/clusterfuzz/_internal/swarming/__init__.py
alhijazi: Renamed to `fuzz_task_duration`. It is not exactly clear to me how long fuzz tasks run on batch currently. 
@jonathanmetzman  could you please advise on this?

src/clusterfuzz/_internal/cron/project_setup.py
alhijazi: I removed this part. Indeed, I was thinking about a way to update a field in the job. Not needed for the moment, would be useful if we end up adding a UI option

src/clusterfuzz/_internal/protos/swarming.proto
alhijazi: Yeah I can add that in a follow up pr!

src/clusterfuzz/_internal/base/utils.py
```diff
@@ -124,6 +124,31 @@ def fetch_url(url):
...
+  environment variable `URL_BLOCKING_OPERATIONS_TIMEOUT`.
+  Args:
+    url: the url to post to.
+    data: request data.
+    headers: request headers.
+  Returns:
+    the contents of the response, in unicode.
+  Raises:
+    raises an `HTTPError`, if one occurred."""
+  operations_timeout = environment.get_value('URL_BLOCKING_OPERATIONS_TIMEOUT')
```
jonathanmetzman, line 144: maybe let's use a sane value here like 60. 10 minutes seems too long to wait.

src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
```diff
@@ -40,6 +41,9 @@ class _Mode(enum.Enum):
   # `uworker_main` tasks are executed on bots via a Pub/Sub queue.
   QUEUE = "queue"
 
+  # `uworker_main` tasks are executed on swarming.
+  SWARMING = "swarming"
```
jonathanmetzman, line 45: nit: single quotes.

src/clusterfuzz/_internal/bot/tasks/task_types.py
```diff
@@ -133,7 +134,11 @@ def execute(self, task_argument, job_type, uworker_env):
       return
 
     logs.log('Queueing utask for remote execution.', download_url=download_url)
-    tasks.add_utask_main(command, download_url, job_type)
+    if batch.is_remote_task(command, job_type):
+      tasks.add_utask_main(command, download_url, job_type)
+    else:
+      assert swarming.is_swarming_task(command, job_type)
+      swarming.push_swarming_task(command, download_url, job_type)
```
jonathanmetzman, line 141: Wont this break fuzz_task, which hashn't been migrated yet?

src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
```diff
@@ -55,6 +59,14 @@ def _timestamp_now() -> Timestamp:
   return ts
 
 
+def _get_execution_mode(utask_module, job_type):
```
jonathanmetzman, line 62: Does anything use QUEUE? Why does it exist then?

src/clusterfuzz/_internal/swarming/__init__.py
```diff
@@ -0,0 +1,135 @@
...
+                      swarming_pb2.StringPair(key='pool', value=swarming_pool)
+                  ],
+                  cas_input_root=swarming_pb2.CASReference(
+                      cas_instance=cas_instance,
+                      digest=swarming_pb2.Digest(
+                          hash=digest_hash, size_bytes=digest_size_bytes)),
+                  execution_timeout_secs=execution_timeout_secs,
+                  env=[
+                      swarming_pb2.StringPair(key='UWORKER', value='True'),
+                      swarming_pb2.StringPair(key='SWARMING_BOT', value='True'),
```
jonathanmetzman, line 107: What code will be starting the task on the batch side?

src/clusterfuzz/_internal/protos/swarming_pb2_grpc.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2023 Google LLC
```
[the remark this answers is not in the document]
alhijazi: It was generated with the script so I kept it, but I don't think it's needed.

src/clusterfuzz/_internal/base/utils.py
alhijazi: Fixed

src/clusterfuzz/_internal/bot/tasks/task_types.py
alhijazi: I thought fuzz_task was already migrated. we can disable scheduling fuzz_task on gpu machines until it's migrated, right?

src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
alhijazi: Tasks that run locally use QUEUE. Local tasks call `tworker_preprocess_no_io`,`uworker_main_no_io`, and `tworker_postprocess_no_io`.
For batch/swarming, they use the same path, so we need a way to determine which is used.

src/clusterfuzz/_internal/swarming/__init__.py
alhijazi: You mean on the swarming side right?
This is to be figured out once we receive the machines and start setting things up. It will be similar to the existing startup scripts for linux/mac

src/clusterfuzz/_internal/bot/tasks/task_types.py
alhijazi: Until fuzz_task in migrated, we won't be able to run fuzzing sessions on the gpu machines, however we can still run the other utasks.

This won't affect the existing jobs that are not supposed to run swarming.
For now, the criterion to determine whether a job needs to run on swarming is whether the job has an env variable `REQUIRES_GPU`.

alhijazi:
Hello @jonathanmetzman,
I think we are ready to land this cl.

I was able to launch a utask_main on swarming running as `untrusted-worker@clusterfuzz.google.com.iam.gserviceaccount.com`

https://chrome-swarming.appspot.com/task?d=true&id=6b85123ff11fa010
This fails because I provide a dummy signed download url.

The run is logged on gcp here:
https://pantheon.corp.google.com/logs/query;query=labels.bot_name%3D%22mac-391-h516.golo.chromium.org%22;cursorTimestamp=2024-08-19T20:12:37.888873Z;startTime=2024-08-19T10:20:26.461Z?e=-13802955&mods=logs_tg_prod&project=google.com:clusterfuzz

configs/test/swarming/swarming.yaml
```diff
@@ -0,0 +1,41 @@
+# Copyright 2024 Google LLC
```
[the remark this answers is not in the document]
alhijazi: Yes indeed, I will do that in a follow-up change

src/clusterfuzz/_internal/swarming/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+from clusterfuzz._internal.base import utils
+from clusterfuzz._internal.config import local_config
+from clusterfuzz._internal.datastore import data_types
+from clusterfuzz._internal.google_cloud_utils import credentials
+from clusterfuzz._internal.protos import swarming_pb2
+
+
+def _job_requires_gpu(job: data_types.Job) -> bool:
+  """Checks whether the job should run on devices with gpu."""
+  requires_gpu = job.get_environment().get('REQUIRES_GPU')
```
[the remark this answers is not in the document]
alhijazi: `test_job_requires_gpu` in `swarming_tests` relies on this impl. I can remove the tests if you think we should instead go with environment.get_value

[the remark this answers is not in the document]
jonathanmetzman: I think you should replace it sorry.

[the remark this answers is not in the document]
alhijazi: Done.

jonathanmetzman (approved):
lgtm

PR #3999, opened 2024-05-17
Fix http server not being started on uworker bots

paulsemel:
Since the startup paths are different, uworker bots are not starting a HTTP server. This might be needed in some cases.


jonathanmetzman (approved):
lgtm

PR #4000, opened 2024-05-20
Make fuzz crashes uworker safe

jonathanmetzman:
This is probably the most important part of migrating fuzz_task.
This superseeds: #3834 and #3816


src/clusterfuzz/_internal/fuzzing/corpus_manager.py
```diff
@@ -609,7 +609,7 @@ def get_proto_corpus(bucket_name,
       gcs_url, num_uploads=max_upload_urls)
   corpus = uworker_msg_pb2.Corpus(
       corpus_urls=corpus_urls,
-      upload_urls=upload_urls,
+    upload_urls=upload_urls,
```
oliverchang, line 612: erroneous change?

oliverchang (approved):

PR #4002, opened 2024-05-21
Ensure the fuzzer job entity is updated.

jonathanmetzman:
Ensure that the fuzzer job entity is updated when the job definition is updated. Otherwise the fuzzerjob will never be updated.
e.g. I have a job with platform linux, ClusterFuzz creates a fuzzerjob i change the platform windows because I made a mistake. The FuzzerJob should be updated so that my fuzzer has a chance to run on windows.
This PR tries to keep the FuzzerJob and Job consistent. But I think a better fix would be to not have two sources of truth about a job.

### DIFF
--- a/src/clusterfuzz/_internal/fuzzing/fuzzer_selection.py
+++ b/src/clusterfuzz/_internal/fuzzing/fuzzer_selection.py
@@ -50,7 +50,7 @@ def update_mappings_for_fuzzer(fuzzer, mappings=None):
       mapping = data_types.FuzzerJob()
     mapping.fuzzer = fuzzer.name
     mapping.job = job_name
-    mapping.platform = job.platform
+    mapping.platform = jobs[job_name].platform
     new_mappings.append(mapping)
 
   ndb_utils.put_multi(new_mappings)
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/fuzzing/fuzzer_selection.py
```diff
@@ -35,24 +35,19 @@ def update_mappings_for_fuzzer(fuzzer, mappings=None):
...
   old_mappings = {}
-  for entity in entities:
-    old_mappings[entity.job] = entity
+  for fuzzer_job in fuzzer_job_entities:
+    old_mappings[fuzzer_job.job] = fuzzer_job
 
   new_mappings = []
+  jobs = ndb_utils.get_all_from_query(
+      data_types.Job.query(data_types.Job.name.IN(mappings)))
+  jobs = {job.name: job for job in jobs}
```
oliverchang, line 46: where is this used?

oliverchang (commented):
hmm, can you give a bit more context about the issue and what this is fixing? it's not very clear to me just looking at the PR.

oliverchang (approved):

PR #4003, opened 2024-05-21
Exclude Linux jobs from having to sign trusted agreement

pgrace-google:
We should not be making people sign the agreement when they are running a Linux job

### DIFF
--- a/src/appengine/handlers/upload_testcase.py
+++ b/src/appengine/handlers/upload_testcase.py
@@ -392,7 +392,7 @@ class UploadHandlerCommon:
         'trustedAgreement') == TRUSTED_AGREEMENT_TEXT.strip()
 
     if (not trusted_agreement_signed and
-        task_utils.is_remotely_executing_utasks()):
+        task_utils.is_remotely_executing_utasks() and platform_id != "Linux"):
       # Trusted agreement was not signed even though the job has privileges and
       # there are other jobs that don't have privileges.
       raise helpers.EarlyExitError(


src/appengine/handlers/upload_testcase.py
```diff
@@ -392,7 +392,7 @@ def do_post(self):
         'trustedAgreement') == TRUSTED_AGREEMENT_TEXT.strip()
 
     if (not trusted_agreement_signed and
```
jonathanmetzman, line 394: I'm not sure I like this.
1. Won't it be an issue if we have linux jobs that are trusted?
2. I think is_remotely_executing_utasks is returning the wrong result and should be fixed instead.

pgrace-google: 1. We don't have linux jobs that are trusted at the moment, right? It would need updating later on for sure - esp once we get hardware
2. hmm do you think linux platform check should live within is_remotely_executing_utasks?

src/appengine/handlers/upload_testcase.py
```diff
@@ -392,7 +392,7 @@ def do_post(self):
...
-        task_utils.is_remotely_executing_utasks()):
+        task_utils.is_remotely_executing_utasks() and platform_id != "Linux"):
```
jonathanmetzman, line 395: nit: single quotes.

src/appengine/handlers/upload_testcase.py
pgrace-google: chatted offline - to make this clearly temporary and submit for now until we have more robust configs set up

src/appengine/handlers/upload_testcase.py
pgrace-google: done!

jonathanmetzman (approved):
lgtm

PR #4005, opened 2024-05-22
[WIP] Bumping deps to an intermediary step compatible with Python3.11

vitorguidi:
### Motivation
Some of our current dependencies are incompatible with Python 3.11. This PR attempts to bump the least possible amount of dependencies (more specifically, grpcio, grpcio-tools, psutil and markupsafe), so we can isolate deployment issues between dependency bumps and the python upgrade itself.


src/Pipfile
```diff
@@ -35,13 +31,27 @@ sendgrid = "==6.0.4"
 # Hack: We are using this to specify App Engine packages,
 # since there is no support for custom package sets.
 future = "==0.17.1"
+grpcio = "==1.62.2"
+psutil = "5.9.4"
+protobuf = "3.20.3"
+cryptography = "==37.0.4"
```
jonathanmetzman, line 37: This was previously in alphabetical order can you preserve this?

src/Pipfile
```diff
@@ -35,13 +31,27 @@ sendgrid = "==6.0.4"
 # Hack: We are using this to specify App Engine packages,
 # since there is no support for custom package sets.
 future = "==0.17.1"
+grpcio = "==1.62.2"
```
jonathanmetzman, line 34: The "Hack" comment is supposed to be above dev-packages. Can you move it there please?

src/platform_requirements.txt
```diff
@@ -1,2 +1,16 @@
-grpcio==1.38.0
+grpcio==1.62.2
```
jonathanmetzman, line 1: This doesn't seem right that we need to add so many platform_requirements.

src/platform_requirements.txt
```diff
@@ -1,2 +1,16 @@
...
 protobuf==3.20.3
+grpcio = "==1.62.2"
+psutil = "5.9.4"
+markupsafe = "2.1.5"
+protobuf = "3.20.3"
+cryptography = "==37.0.4"
+pyyaml = "==6.0.0"
+bcrypt = "4.1.2"
+cffi = "==1.15.1"
+lazy-object-proxy= "==1.9.0"
```
jonathanmetzman, line 11: What is this new package?

src/Pipfile
```diff
@@ -35,13 +31,27 @@ sendgrid = "==6.0.4"
...
+cryptography = "==37.0.4"
+pyyaml = "==6.0.0"
+bcrypt = "4.1.2"
+cffi = "==1.15.1"
+lazy-object-proxy= "==1.9.0"
+pynacl = "==1.5.0"
+wrapt = "==1.16.0"
+google-cloud-profiler = "==4.1.0"
+google-crc32c = "==1.5.0"
+typed-ast = "==1.5.5"
```
jonathanmetzman, line 46: Why do we need ast code in prod?

src/platform_requirements.txt
vitorguidi: There are transitive dependencies which are C extensions, and were previously not declared.

src/platform_requirements.txt
vitorguidi: Ditto

jonathanmetzman: :facepalm: since youre adding it, you can have the honor of figuring out what it is.

src/platform_requirements.txt
jonathanmetzman: Ah.
So I don't always love the idea of pinning dependencies because it can confuse what we are actually interested in and makes it harder to upgrade the main package. Can we just leave these out?

src/platform_requirements.txt
jonathanmetzman: Can you find out?

src/appengine/handlers/testcase_detail/show.py
```diff
@@ -19,6 +19,7 @@
 
 from flask import request
 import jinja2
+from markupsafe import escape
```
jonathanmetzman, line 22: We don't import functions, only packages. See python styleguide for more details on this.

jonathanmetzman, line 22: We don't import functions, only packages/modules. See python styleguide for more details on this.

src/clusterfuzz/_internal/tests/appengine/handlers/base_handler_test.py
```diff
@@ -146,7 +146,7 @@ def test_forbidden_not_logged_in(self):
     app = webtest.TestApp(flaskapp)
     response = app.get('/', expect_errors=True)
     self.assertEqual(response.status_int, 302)
-    self.assertEqual('http://localhost/login?dest=http%3A%2F%2Flocalhost%2F',
+    self.assertEqual('/login?dest=http%3A%2F%2Flocalhost%2F',
```
jonathanmetzman, line 149: Do we know why this has changed?

src/platform_requirements.txt
vitorguidi: I would rather declare every single C dependency. Otherwise pipenv lock will choose those for us, and I have no assurance that whatever it chooses will be compatible with python 3.11.

src/platform_requirements.txt
vitorguidi: If you mean lazy-object-proxy, this is a transitive dependency for pylint-protobuf
cffi is a transitive dependency for, among others, pyopenssl
bcrypt is a transitive dependency for fabric

src/clusterfuzz/_internal/tests/appengine/handlers/base_handler_test.py
vitorguidi: Default behavior on flask, that changed among versions. Both are valid, see https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.2

src/platform_requirements.txt
jonathanmetzman: most of these are defined in the pipfile and you can just define them there.

src/Pipfile
vitorguidi: This could probably be in the dev dependencies. It is a transitive dependency for pylint-protobuf

src/platform_requirements.txt
jonathanmetzman: Hmmm...I'm not actually sure we need it.
@oliverchang WDYT?

oliverchang: It's also surprising to me that we need to add this many. 

We package all our zips for all platforms (including Mac, windows) on Linux. this means that by default, we would be pulling all of the Linux versions of those C extensions and distributing them to Mac and windows. 

Can we validate that we actually need all of these C extensions? 

Re Pipfile vs platform_requirements: 

the platform_requirements.txt file was meant as a hacky override to "re-install" some of the results of the Pipfile resolution / install with platform-specific ones instead when are we are packaging up our  zips here: https://github.com/google/clusterfuzz/blob/7f39a19e5b87295ca9c3f14557eb3e82d7c0dc88/src/local/butler/package.py#L85

vitorguidi: We do need all these extensions, [proof](https://gist.github.com/vitorguidi/b2a6599e7f3c8bbfdf1741eb6525d571) (run pip install pipdeptree && pipdeptree so you can see dependency trees). 

The only margin we have to simplify this, is if the transitive C dependencies will be obtained from pip install on Mac and Windows. If that is actually the case, the only packages that should belong in platform_dependencies are the **C dependencies which are top level packages**.

If that is the case however, why do we not place psutil and pyyaml in platform_dependencies? How are these two specific packages able to propagate correctly if they are top leve, and not transitive, dependencies?

src/appengine/handlers/testcase_detail/show.py
```diff
@@ -17,8 +17,9 @@
 import html
 import re
 
-from flask import request
```
jonathanmetzman, line 20: Thanks for changing this!

src/Pipfile
```diff
@@ -5,7 +5,12 @@ name = "pypi"
 
 [packages]
 antlr4-python3-runtime = "==4.7.2"
+bcrypt = "==4.1.2"
```
jonathanmetzman, line 8: So my understanding is that only direct dependencies are supposed to go in this file. But I think youre adding indirect ones right? Could you find out what is the correct practice?

vitorguidi: I am adding the indirect ones which are C extensions, and that is intentional. I want to be absolutely sure the binary wheels will be available once we do the version bump, and if I do not pin these, when we lock the dependencies it is possible that pip will resolve some version that makes it harder to upgrade (or introduces regressions).

jonathanmetzman: OK

src/platform_requirements.txt
vitorguidi: Follow up on this: investigate what transitive C dependnecies are required on windows and Mac, and how they end up getting installed if we only pull grpcio/proto for multiplatform, and the rest is linux only

oliverchang: +1.  Everything in the root Pipfile (https://github.com/google/clusterfuzz/blob/master/Pipfile) is expected to be installed on the system. This will certainly require C-extensions (and possibly transitively). 

src/Pipfile (which is separate from the root Pipfile) contains the packages that we download and re-package into zips for deployment. In theory, only grpcio and protobuf should require C extensions there, because we've never had any issue on Mac/Windows with just patching in the platform-specific versions of these two (which is what platform_requirements.txt does). 

Eventually, we do want to move everything from the root Pipfile into platform_requirements.txt, but perhaps we can do this more iteratively. e.g. just move a minimal seet that we have to absolutely upgrade for Python 3.11 support for now (e.g. psutil?)

oliverchang: and perhaps with PYTHONPATH/sys.path ordering, we don't even have to uninstall the existing version of psutil on the system if the newer one we provide in the source zip takes precedence over the system one.

see https://github.com/google/clusterfuzz/blob/5d4bd65fa370a01b7d0006d404ca6c27fafa5e21/src/clusterfuzz/_internal/base/modules.py#L47 for how we mess with python module search paths on startup.

src/platform_requirements.txt
vitorguidi: Now that I think of it, we can just get rid of lazy-object-proxy and typed-ast from dev dependencies, these do not run any appengine or clusterfuzz code, only the lint/format in butler.py, no reason to pin them

vitorguidi: Hmm there is a >0 chance that @jonathanmetzman  PR was actually correct, and Windows only broke because it was not running psutil >= 5.9.4. The largest version I saw in our windows init scripts was 5.7.0, which is not compatible with py 3.11.

I might be wrong, but this seems very plausible

src/platform_requirements.txt
vitorguidi: Hmm there is a >0 chance that @jonathanmetzman https://github.com/google/clusterfuzz/pull/3663 was actually correct, and Windows only broke because it was not running psutil >= 5.9.4. The largest version I saw in our windows init scripts was 5.7.0, which is not compatible with py 3.11.

I might be wrong, but this seems very plausible

jonathanmetzman: I'm confused, i thought this was your theory. I thought i succeeded in upgrading the version on Windows.

vitorguidi: I was working under hte assumption that the update succeeded on linux but not on windows . Is that the case?

PR #4008, opened 2024-05-27
Set component id if a numeric component is available in the component…

alhijazi:
… list

### DIFF
--- a/src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
+++ b/src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
@@ -686,8 +686,14 @@ class Issue(issue_tracker.Issue):
       # Make sure self.labels contains only hotlist IDs.
       self._filter_labels()
 
-      if self.component_id:
-        self._data['issueState']['componentId'] = int(self.component_id)
+      # Best effort setting on componentId from list of components.
+      # Set the componentId to the first encountered numeric component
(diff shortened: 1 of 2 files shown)


alhijazi:
@rmistry is there a way to get a componentid from component tag?
Does predator return component tags or ids?

src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
```diff
@@ -686,8 +686,14 @@ def save(self, new_comment=None, notify=True):
       # Make sure self.labels contains only hotlist IDs.
       self._filter_labels()
 
-      if self.component_id:
-        self._data['issueState']['componentId'] = int(self.component_id)
+      # Best effort setting on componentId from list of components.
+      # Set the componentId to the first encountered numeric component
+      # in the component list.
+      for component in list(self.components):
+        if component.isnumeric():
```
jonathanmetzman, line 693: hmmm...will component ever be an int? this will fail in that case.

alhijazi: It will not be an int, we are doing the same thing a few lines above:
https://github.com/google/clusterfuzz/blob/22e11083b540518248d512141c45c25c7f560f2e/src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py#L665

PR #4013, opened 2024-06-03
tasks: add support for both 'jobs-android' and device specific subqueues

paulsemel:
In chrome, we need to have generic jobs that can be scheduled on any devices. This PR will make sure generic tasks keep getting pulled even when subqueues are set-up. This PR is preliminary work for us to add subqueues on our android devices.

### DIFF
--- a/src/clusterfuzz/_internal/base/tasks.py
+++ b/src/clusterfuzz/_internal/base/tasks.py
@@ -122,6 +122,11 @@ def default_queue_suffix():
   return queue_suffix_for_platform(environment.platform())
 
 
+def generic_android_queue(prefix=JOBS_PREFIX):
+  """Get the generic 'android' queue that is not tied to a specific device."""
+  return prefix + queue_suffix_for_platform(environment.platform())
+
+
 def regular_queue(prefix=JOBS_PREFIX):
   """Get the regular jobs queue."""
   return prefix + default_queue_suffix()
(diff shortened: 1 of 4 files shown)


src/clusterfuzz/_internal/bot/tasks/commands.py
```diff
@@ -339,10 +339,16 @@ def process_command_impl(task_name,
             logs.log('Testcase %d platform (%s) does not match with ours\
                  (%s), exiting' % (testcase.key.id(), testcase_platform_id,
                                    current_platform_id))
+            # If this happens on android, we need to make sure we're not
+            # scheduling this task on a subqueue, since we know the current
+            # platform doesn't match.
+            queue = tasks.generic_android_queue() if environment.is_android(
```
paulsemel, line 345: This might not work actually... if a task runs on the correct device but not the correct branch of android, we might end up in an infinite loop there. We might need to check whether platform actually matches to correctly select the appropriate queue.

jonathanmetzman:
Mark do you understand how this is different from Haiku's queing system for Android?

src/clusterfuzz/_internal/bot/tasks/commands.py
paulsemel: Fixed this. The idea is that we want to make sure we're trying to schedule on a different device to schedule on the generic android queue. In theory, we should almost never schedule on the generic queue, except while migrating to using subqueues. Even for generic jobs, once the testcase is pinned to a device, we should start scheduling the next tasks on the subqueue so that we keep scheduling on the correct set of devices.

src/clusterfuzz/_internal/base/tasks.py
```diff
@@ -190,6 +195,21 @@ def get_regular_task(queue=None):
       return task
 
 
+def get_android_regular_task():
+  """Get an android regular task."""
+  if not environment.get_value('CF_ALLOWS_LISTENING_GENERIC_QUEUE') or \
```
marktefftech, line 200: As discussed in meet, let's make this default as opposed to introducing new env var

Issue #4016, opened 2024-06-05
Fix issue where global variables are causing tests to pass when they should fail


jonathanmetzman:
```
======================================================================
ERROR: test_setup_fuzz_0_True (tests.core.build_management.build_manager_test.RegularLibFuzzerBuildTest)
Tests setting up a build during fuzzing [with unpack_all='True'].
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/user/home.local/share/virtualenvs/clusterfuzz-H0A0yd4x/lib/python3.7/site-packages/parameterized/parameterized.py", line 620, in standalone_func
    return func(*(a + p.args), **p.kwargs, **kw)
  File "/user/homeprojects/clusterfuzz/src/clusterfuzz/_internal/tests/core/build_management/build_manager_test.py", line 483, in test_setup_fuzz
    self.assertEqual('target2', os.environ['FUZZ_TARGET'])
  File "/usr/local/lib/python3.7/os.py", line 681, in __getitem__
    raise KeyError(key) from None
KeyError: 'FUZZ_TARGET'

======================================================================
ERROR: test_setup_fuzz_with_extra_0_True (tests.core.build_management.build_manager_test.RegularLibFuzzerBuildTest)
Tests setting up a build during fuzzing with an extra build [with unpack_all='True'].
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/user/home.local/share/virtualenvs/clusterfuzz-H0A0yd4x/lib/python3.7/site-packages/parameterized/parameterized.py", line 620, in standalone_func
    return func(*(a + p.args), **p.kwargs, **kw)
  File "/user/homeprojects/clusterfuzz/src/clusterfuzz/_internal/tests/core/build_management/build_manager_test.py", line 527, in test_setup_fuzz_with_extra
    self.assertEqual('target2', os.environ['FUZZ_TARGET'])
  File "/usr/local/lib/python3.7/os.py", line 681, in __getitem__
    raise KeyError(key) from None
KeyError: 'FUZZ_TARGET'
```

jonathanmetzman:
This failure was caused by removing the line: `from clusterfuzz._internal.bot.tasks.utasks import fuzz_task` from build_manager_test

PR #4018, opened 2024-06-05
Fix failing unittests.

jonathanmetzman:
Including one caused by a nonhermetic test relying on environment variables set outside of the test.

### DIFF
--- a/src/clusterfuzz/_internal/tests/core/build_management/build_manager_test.py
+++ b/src/clusterfuzz/_internal/tests/core/build_management/build_manager_test.py
@@ -26,16 +26,15 @@ from pyfakefs import fake_filesystem_unittest
 
 from clusterfuzz._internal.base import errors
 from clusterfuzz._internal.base import utils
+# Add this so some badly written code doesn't cause a test failure.
+# TODO(https://github.com/google/clusterfuzz/issues/4016): Fix this.
+from clusterfuzz._internal.bot.tasks.utasks import fuzz_task  # pylint: disable=unused-import
 from clusterfuzz._internal.build_management import build_manager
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/tests/core/build_management/build_manager_test.py
```diff
@@ -26,6 +26,10 @@
 
 from clusterfuzz._internal.base import errors
 from clusterfuzz._internal.base import utils
+# Add this so some badly written code doesn't cause a test failure.
+# TODO(https://github.com/google/clusterfuzz/issues/4016): Fix this.
+from clusterfuzz._internal.bot.tasks.utasks import \
+    fuzz_task  # pylint: disable=unused-import
```
oliverchang, line 32: Is this import alone adding some kind of env var somewhere? where exactly is it?

oliverchang (approved):
thanks for the detailed context in this PR!

src/clusterfuzz/_internal/tests/core/build_management/build_manager_test.py
jonathanmetzman: I have no idea to be honest. Given that it's importing many other modules, it could be anywhere :facepalm:

jonathanmetzman:
Related: https://github.com/google/clusterfuzz/issues/4016

src/clusterfuzz/_internal/tests/core/build_management/build_manager_test.py
jonathanmetzman: Yes I think it's adding some kind of env var.

PR #4021, opened 2024-06-07
Remove ADB logging to avoid exceptions in logging large messages

jonathanmetzman:
One of the most common exceptions in ClusterFuzz is logging messages that are too long. This is bad and leads to warning fatigue. If we need this debugging capability it can be added back. But otherwise, we shouldn't have spurious exceptions firing so often.

### DIFF
--- a/src/clusterfuzz/_internal/platforms/android/adb.py
+++ b/src/clusterfuzz/_internal/platforms/android/adb.py
@@ -111,8 +111,7 @@ def create_directory_if_needed(device_directory):
 def directory_exists(directory_path):
   """Return whether a directory exists or not."""
   expected = '0'
-  result = run_shell_command(
-      '\'test -d "%s"; echo $?\'' % directory_path)
+  result = run_shell_command('\'test -d "%s"; echo $?\'' % directory_path)
   return result == expected
 
 
@@ -203,8 +202,7 @@ def factory_reset():
 def file_exists(file_path):
   """Return whether a file exists or not."""
   expected = '0'
-  result = run_shell_command(
-      '\'test -f "%s"; echo $?\'' % file_path)
+  result = run_shell_command('\'test -f "%s"; echo $?\'' % file_path)
   return result == expected
 
 
--- a/src/clusterfuzz/_internal/platforms/android/adb.py
+++ b/src/clusterfuzz/_internal/platforms/android/adb.py
@@ -82,14 +82,13 @@ def copy_local_directory_to_remote(local_directory, remote_directory):
   """Copies local directory contents to a device directory."""
   create_directory_if_needed(remote_directory)
   if os.listdir(local_directory):
-    run_command(['push', '%s/.' % local_directory, remote_directory], True,
-                True)
+    run_command(['push', '%s/.' % local_directory, remote_directory], True)
 
 
 def copy_local_file_to_remote(local_file_path, remote_file_path):
   """Copies local file to a device file."""
   create_directory_if_needed(os.path.dirname(remote_file_path))
-  run_command(['push', local_file_path, remote_file_path], True, True)
+  run_command(['push', local_file_path, remote_file_path], True)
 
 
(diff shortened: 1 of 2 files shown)


letitz:
Removing visibility into production failures goes counter to where I'd like ClusterFuzz to end up, as a user/operator. Can we instead truncate log messages?

marktefftech:
I'm wondering if there is another way to truncate logs, such as suggested by letitz@.

Removing the error logging from adb pretty much removes any helpful logs for debugging an issue with an android device, since this is their exclusive means of communication.

Can you provide an example of where you're seeing these messages flood the logs? Is it a single error being thrown many times, or many errors being thrown a single time?

jonathanmetzman:
> I'm wondering if there is another way to truncate logs, such as suggested by letitz@.
> 
> Removing the error logging from adb pretty much removes any helpful logs for debugging an issue with an android device, since this is their exclusive means of communication.
> 
> Can you provide an example of where you're seeing these messages flood the logs? Is it a single error being thrown many times, or many errors being thrown a single time?

Could you truncate the logs? I don't really know which ones are spitting out too much.
Here is the error: https://pantheon.corp.google.com/errors/detail/CPWHwu6d_4SnLg?project=google.com:clusterfuzz
It is very frequent

I think dumping the output of arbitrary fuzzing processes into logs isn't a good idea. The outputs can be anything, and not everything can go in the logs :-)

letitz:
> I think dumping the output of arbitrary fuzzing processes into logs isn't a good idea. The outputs can be anything, and not everything can go in the logs :-)

We include stderr from subprocesses in other places, e.g. https://github.com/google/clusterfuzz/blob/0d0ac1871b0d8c7f463257d2023233268997fd05/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py#L522-L532. 

I think it's oftentimes the only way we have to diagnose issues in production, so I would be really sad to see this PR land as is. It seems reasonable to me to instead add some common logic in the logging code to truncate excessively-long messages.

jonathanmetzman:
I'm going through with this. The error alerts are not usable anymore due to false positives.

ParisMeuleman:
Hey there, I was looking into reenabling some logs on android as we lack visibility. 
AFAIK we already truncate: https://github.com/google/clusterfuzz/blob/f918c367f25e7e6a24b837ab7be94e84af872c5a/src/clusterfuzz/_internal/metrics/logs.py#L553
though we have a specific limit for stackdriver that we don't use when we are not using fluentd.
we could probably truncate more aggresively always, or switch on 
‎_cloud_logging_enabled https://github.com/google/clusterfuzz/blob/f918c367f25e7e6a24b837ab7be94e84af872c5a/src/clusterfuzz/_internal/metrics/logs.py#L97

Do you remember the specific logs @jonathanmetzman regarding the exceeding size?

jonathanmetzman:
> Hey there, I was looking into reenabling some logs on android as we lack visibility. AFAIK we already truncate:
> 
> https://github.com/google/clusterfuzz/blob/f918c367f25e7e6a24b837ab7be94e84af872c5a/src/clusterfuzz/_internal/metrics/logs.py#L553
> 
> 
> though we have a specific limit for stackdriver that we don't use when we are not using fluentd.
> we could probably truncate more aggresively always, or switch on
> ‎_cloud_logging_enabled
> https://github.com/google/clusterfuzz/blob/f918c367f25e7e6a24b837ab7be94e84af872c5a/src/clusterfuzz/_internal/metrics/logs.py#L97
> 
> Do you remember the specific logs @jonathanmetzman regarding the exceeding size?
> 
> github.com/[/pull/4411](https://github.com/google/clusterfuzz/pull/4411) is reenabling logs with proposed fix above

I don't unfortunately.

ParisMeuleman:
Thanks, well then I think we should go through with #4411, and if the issues reappear we will tune it/revert it.
I searched but I believe those logs were not retained? or I have the wrong keywords.

PR #4022, opened 2024-06-07
Dont log failures to log metrics in Android as errors.

jonathanmetzman:
This is one of the top errors and contributes to error fatigue.

### DIFF
--- a/src/clusterfuzz/_internal/metrics/monitor.py
+++ b/src/clusterfuzz/_internal/metrics/monitor.py
@@ -108,7 +108,12 @@ class _FlusherThread(threading.Thread):
         if time_series:
           create_time_series(project_path, time_series)
       except Exception:
-        logs.log_error('Failed to flush metrics.')
+        if environment.is_android():
+          # FIXME: This exception is extremely common on Android. We are already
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/metrics/monitor.py
```diff
@@ -108,7 +108,12 @@ def run(self):
         if time_series:
           create_time_series(project_path, time_series)
       except Exception:
-        logs.log_error('Failed to flush metrics.')
+        if environment.is_android():
+          # FIXME: This exception is extremely common on Android. We are already
+          # aware of the problem, don't make more noise about it.
+          logs.log_warn('Failed to flush metrics.')
```
letitz, line 114: Just making sure: will this log the exception and its stack trace too?

letitz (approved):

src/clusterfuzz/_internal/metrics/monitor.py
jonathanmetzman: It won't.

letitz: That's too bad.

PR #4025, opened 2024-06-10
Bumping to python 3.11

vitorguidi:


src/clusterfuzz/_internal/bot/tasks/update_task.py
```diff
@@ -217,14 +224,31 @@ def update_source_code():
...
+    logs.info(f'absolute_filepath = {absolute_filepath}')
+
     if os.path.altsep:
+      logs.info(
+          f'altsep is not None. Old absolute_filepath = {absolute_filepath}'
+      )
       absolute_filepath = absolute_filepath.replace(os.path.altsep, os.path.sep)
+      logs.info(f'New absolute filepath = {absolute_filepath}')
 
-    if os.path.realpath(absolute_filepath) != absolute_filepath:
```
oliverchang, line 224: Let's try removing it. I can't think of why this would be needed for our source update use case, since this seems like something that's more aimed at prevent directory traversal. 

The other potential case is symlinks -- we could potentially just guard against that better by just checking if the entry in the zip is a symlink. I believe this is represented as a bit here: https://docs.python.org/3/library/zipfile.html#zipfile.ZipInfo.external_attr. IMO we can just skip this too unless we run into issues here.

vitorguidi: I deployed #4127 and did not observe any file being skipped, on any windows or linux production machine in internal. I will delete this line of code since there is no evidence it is relevant, and revisit it if anything breaks during rollout phase.

image

Disclaimer: the update task pulled the same source code it was running, so this is not representative of all possible update task scenarios.

vitorguidi: It remained functional after getting rid of this code, looks good to proceed without this check for now 
image

oliverchang: SGTM!

infra/k8s/schedule-upload-reports.yaml
```diff
@@ -0,0 +1,41 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: schedule-upload-reports-tasks
```
vitorguidi, line 18: This is probably a ghost cronjob. We can delete it manually and not commit this to version control, wdyt?

vitorguidi:
One issue happened but is probably not related to the update itself.

Manage VMs is breaking because it is trying to update one of our clusters beyond the nodes limit in GCP:

image

Two other errors are probably related to the above

image
image

All leads to believe this is an implementation issue in manage_vms.py and not related to the 311 update.

configs/test/gce/android-init.bash
```diff
@@ -97,7 +97,7 @@ sed -i 's/flush_interval 5s/flush_interval 60s/' \
 sudo service google-fluentd restart
 
 echo "Installing ClusterFuzz package dependencies."
-pip install crcmod==1.7 psutil==5.6.6 cryptography==3.3.2 pyOpenSSL==19.0.0
+pip install crcmod==1.7 psutil==5.9.4 cryptography==37.0.4 pyOpenSSL==22.0.0
```
jonathanmetzman, line 100: Did you make this change in our internal repo?

configs/test/gce/windows-init.ps1
```diff
@@ -63,7 +63,7 @@ setx /M PYTHONIOENCODING "UTF-8"
...
+$s = "if not exist $registrySetupFilePath ( EXIT )`nw32tm /resync`nnetsh winhttp import proxy source=ie`nnfsadmin client config protocol=tcp+udp UseReservedPorts=yes`nnfsadmin client stop`nnfsadmin client start`nset NFS_HOST=$nfsHost`nset NFS_VOLUME=$nfsVolume`nset NFS_ROOT=$nfsRoot`nmount -o anon -o nolock -o retry=10 $nfsHost`:/$nfsVolume $nfsRoot`nnet start w32time`nw32tm /resync`nset PREEMPTIBLE=$preemptible`nset QUEUE_OVERRIDE=$queueOverride`nset USER=bot`nset BOT_TMPDIR=c:\tmp`nset PYTHONPATH=c:\clusterfuzz\src`nset ROOT_DIR=c:\clusterfuzz`nset PATH=c:\java\bin;c:\python311;c:\python27;c:\Windows\System32;c:\nodejs;c:\Program Files (x86)\Windows Kits\10\Debuggers\x64;c:\Program Files (x86)\Google\Cloud SDK\google-cloud-sdk\bin;%PATH%`nc: `ncd \ `ncd clusterfuzz\src\python\bot\startup `npython -W ignore run.py"
```
jonathanmetzman, line 66: Did you make this change in our itnernal repo?

infra/k8s/backup.yaml
```diff
@@ -26,11 +26,13 @@ spec:
         spec:
           containers:
           - name: backup
-            image: gcr.io/clusterfuzz-images/base
+            image: gcr.io/clusterfuzz-images/base:vguidi_py311_testing_v1
```
jonathanmetzman, line 29: this is intentional right?

local/tests/ci_tests.bash
```diff
@@ -14,13 +14,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-IMAGE=gcr.io/clusterfuzz-images/ci
+IMAGE=gcr.io/clusterfuzz-images/ci:bc696af-202407021946
```
jonathanmetzman, line 17: should this be vguidi?

jonathanmetzman (approved):
lgtm

local/tests/ci_tests.bash
vitorguidi: I built this like, 1.5mo ago, but this is indeed python311.
Can change it for consistency and readability to point to the vguidi tag

infra/k8s/backup.yaml
vitorguidi: Yes. If I changed latest, everybody else that is not on 311 right now would break

configs/test/gce/windows-init.ps1
vitorguidi: Yes, it was done in the progressive rollout script for windows.

configs/test/gce/android-init.bash
vitorguidi: No. What is this script responsible for?

local/install_deps_linux.bash
```diff
@@ -25,7 +25,9 @@ while [ "$1" != "" ]; do
 done
 
 if [ -z "$PYTHON" ]; then
-  if which python3.10 > /dev/null; then
+  if which python3.11 > /dev/null; then
+    export PYTHON='python3.11'
+  elif which python3.10 > /dev/null; then
     export PYTHON='python3.10'
   elif which python3.9 > /dev/null; then
```
vitorguidi, line 32: Need to install pipenv through apt

configs/test/gce/android-init.bash
oliverchang: (Closing thread from chat). I believe this is an outdated / stale config file that's no longer required.

local/install_deps_linux.bash
oliverchang: Any reason in particular for this? We've had issues with different users having different `pipenv` versions in the past, since there are incompatibilities between versions.

vitorguidi: The local deps script breaks for PYTHON=python3.11 ./local/instal_deps.bash. Installing it through apt was the only solution I found (Jonathan's approach on last year's attempt)

oliverchang: What's the specific breakage? CC @jonathanmetzman as well. 

I'm worried this will introduce more inconsistencies down the line.

vitorguidi: This is the badness that happens attempting to install pipenv through pip

image

oliverchang (approved):

PR #4034, opened 2024-06-12
Remove fuzz target weight cap.

### LINKED ISSUE #4019: Remove cap on weight based on fuzz targets

letitz:
Allow engine fuzzer/job multipliers to keep up with the number of fuzz targets behind each engine/job combination.

This puts fuzz targets on approximately the same footing as blackbox fuzzers. They are still not strictly equivalent, since the logic here sets the engine's weight to the count of fuzz targets behind it, not the sum of those targets' weights. Consider the following scenario:

- fuzzer A: weight = 1.0, multiplier = 5.0
- fuzzer B: weight = 1.0, multiplier = 1.0 
- libfuzzer: weight  = 1.0, multiplier = 2.0
   - target C: weight 5.0
   - target D: weight 1.0

Fuzzer A has a 5/8 chance of being picked for a fuzzing session, whereas even though target C has a weight of 5.0, it still has only a 2/8 * 5/6 chance of being picked for a fuzzing session.

That said, if most multipliers are ~1.0 and most target weights are ~1.0, then a target weight of `x` results in a probability roughly equal to `x / sum(all fuzzer and target weights)`.

Fixes #4019.


jonathanmetzman (approved):
LGTM

Thanks for doing this!

PR #4039, opened 2024-06-19
Primitive progressive rollout implementation

vitorguidi:
### Motivation

There is no available mechanism by which two clusterfuzz versions can coexist in production. This PR assumes that, during the GCE VM creation, an user data script will:

- Startup scripts will be able to query the GCP api and figure out candidate/prod weights in the fleet composition, and choose docker image (linux) or python interpreter (windows) during initialization.
- The clusterfuzz docker container will be launched, during VM startup, with the CLUSTERFUZZ_STAGE environment variable, which can assume the values "prod" and "candidate".

If these assumptions hold, we can assume that clusterfuzz will start with the correct package ({platform}-3.zip or {platfor}-3-candidate.zip), and pull the corresponding file in all future update task runs.

This also requires change to butler deploy/package/remote, in order to take into account the two possible releases.
This PR belongs to [this effort](https://github.com/google/clusterfuzz/issues/4059).

### DIFF
--- a/butler.py
+++ b/butler.py
@@ -276,6 +276,8 @@ def main():
       '--prod', action='store_true', help='Deploy to production.')
   parser_deploy.add_argument(
       '--targets', nargs='*', default=['appengine', 'k8s', 'zips'])
+  parser_deploy.add_argument(
+      '--stage', '-s', choices=['prod', 'candidate'])
 
   parser_run_server = subparsers.add_parser(
       'run_server', help='Run the local Clusterfuzz server.')
--- a/butler.py
+++ b/butler.py
@@ -259,7 +259,7 @@ def main():
       'package', help='Package clusterfuzz with a staging revision')
   parser_package.add_argument(
       '-p', '--platform', choices=['linux', 'macos', 'windows', 'all'])
-  parser_package.add_argument('-r', '--release', choices=['prod', 'candidate'])
+  parser_package.add_argument('-r', '--release', choices=['prod', 'candidate'], default='prod')
 
   parser_deploy = subparsers.add_parser('deploy', help='Deploy to Appengine')
   parser_deploy.add_argument(
@@ -275,7 +275,7 @@ def main():
       '--prod', action='store_true', help='Deploy to production.')
   parser_deploy.add_argument(
       '--targets', nargs='*', default=['appengine', 'k8s', 'zips'])
-  parser_deploy.add_argument('--release', '-r', choices=['prod', 'candidate'])
+  parser_deploy.add_argument('--release', '-r', choices=['prod', 'candidate'], default='prod')
 
   parser_run_server = subparsers.add_parser(
       'run_server', help='Run the local Clusterfuzz server.')
--- a/butler.py
+++ b/butler.py
@@ -259,8 +259,7 @@ def main():
       'package', help='Package clusterfuzz with a staging revision')
   parser_package.add_argument(
       '-p', '--platform', choices=['linux', 'macos', 'windows', 'all'])
-  parser_package.add_argument(
-      '-r', '--release', choices=['prod', 'candidate'])
+  parser_package.add_argument('-r', '--release', choices=['prod', 'candidate'])
 
   parser_deploy = subparsers.add_parser('deploy', help='Deploy to Appengine')
   parser_deploy.add_argument(
@@ -276,8 +275,7 @@ def main():
       '--prod', action='store_true', help='Deploy to production.')
   parser_deploy.add_argument(
       '--targets', nargs='*', default=['appengine', 'k8s', 'zips'])
-  parser_deploy.add_argument(
-      '--release', '-r', choices=['prod', 'candidate'])
+  parser_deploy.add_argument('--release', '-r', choices=['prod', 'candidate'])
 
   parser_run_server = subparsers.add_parser(
       'run_server', help='Run the local Clusterfuzz server.')
--- a/butler.py
+++ b/butler.py
@@ -260,7 +260,7 @@ def main():
   parser_package.add_argument(
       '-p', '--platform', choices=['linux', 'macos', 'windows', 'all'])
   parser_package.add_argument(
-      '-s', '--stage', choices=['prod', 'candidate'])
+      '-r', '--release', choices=['prod', 'candidate'])
 
   parser_deploy = subparsers.add_parser('deploy', help='Deploy to Appengine')
   parser_deploy.add_argument(
@@ -277,7 +277,7 @@ def main():
   parser_deploy.add_argument(
       '--targets', nargs='*', default=['appengine', 'k8s', 'zips'])
   parser_deploy.add_argument(
-      '--stage', '-s', choices=['prod', 'candidate'])
+      '--release', '-r', choices=['prod', 'candidate'])
 
   parser_run_server = subparsers.add_parser(
       'run_server', help='Run the local Clusterfuzz server.')
--- a/butler.py
+++ b/butler.py
@@ -259,6 +259,8 @@ def main():
       'package', help='Package clusterfuzz with a staging revision')
   parser_package.add_argument(
       '-p', '--platform', choices=['linux', 'macos', 'windows', 'all'])
+  parser_package.add_argument(
+      '-s', '--stage', choices=['prod', 'candidate'])
 
   parser_deploy = subparsers.add_parser('deploy', help='Deploy to Appengine')
   parser_deploy.add_argument(
--- a/docker/base/Dockerfile
+++ b/docker/base/Dockerfile
@@ -97,12 +97,19 @@ RUN wget https://download.java.net/java/GA/jdk15.0.2/0d1cfde4252546c6931946de8db
     tar -xzv --strip-components=1 -f openjdk-15.0.2_linux-x64_bin.tar.gz --directory $JAVA_HOME && \
     rm -rf openjdk*.tar.gz $JAVA_HOME/jmods $JAVA_HOME/lib/src.zip
 
-# Install Python 3.
+# Install Python 3.7
 RUN curl -sS https://www.python.org/ftp/python/3.7.7/Python-3.7.7.tgz | tar -C /tmp -xzv && \
     cd /tmp/Python-3.7.7 && \
     ./configure --enable-optimizations --enable-loadable-sqlite-extensions && make altinstall && \
     rm -rf /tmp/Python-3.7.7
 RUN pip3.7 --no-cache-dir install pipenv==2022.8.5
+
+# Install Python3.11
+RUN curl -sS https://www.python.org/ftp/python/3.11.4/Python-3.11.4.tgz | tar -C /tmp -xzv && \
+    cd /tmp/Python-3.11.4 && \
+    ./configure --enable-optimizations --enable-loadable-sqlite-extensions && make altinstall && \
+    rm -rf /tmp/Python-3.11.4
+RUN pip3.11 --no-cache-dir install pipenv==2022.8.5
 
 # Install Node.js
 COPY setup_19.x /data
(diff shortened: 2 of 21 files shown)


docker/base/Dockerfile
```diff
@@ -131,7 +138,8 @@ ENV ROOT_DIR $INSTALL_DIRECTORY/clusterfuzz
 ENV UPDATE_WEB_TESTS True
 ENV PYTHONPATH $INSTALL_DIRECTORY/clusterfuzz/src
 ENV RUN_CMD "python3.7 $ROOT_DIR/src/python/bot/startup/run.py"
-ENV DEPLOYMENT_ZIP "linux-3.zip"
+ENV PYTHON_RUN_CMD = "$ROOT_DIR/src/python/bot/startup/run.py"
```
jonathanmetzman, line 141: nit: Don't have an equal for consistency purposes.

docker/base/Dockerfile
```diff
@@ -148,9 +156,11 @@ ENV LANG en_US.UTF-8
 ENV PYTHONIOENCODING UTF-8
 
 COPY setup_common.sh setup_clusterfuzz.sh start_clusterfuzz.sh setup_mock_metadata.sh Pipfile Pipfile.lock start.sh /data/
-RUN cd /data && \
-    # Make pip3.7 the default so that pipenv install --system works.
-    mv /usr/local/bin/pip3.7 /usr/local/bin/pip && \
-    pipenv install --deploy --system
+#Moving this to start.sh so we can choose interpreter version at runtime
+RUN cd /data
+#RUN cd /data && \
```
jonathanmetzman, line 161: Do we not need this anymore?

docker/base/start.sh
```diff
@@ -13,6 +13,17 @@
...
 # limitations under the License.
 
+DEPLOYMENT_ZIP="linux3.zip"
+if [[$CLUSTERFUZZ_STAGE == "candidate" ]]; then
+    DEPLOYMENT_ZIP="linux3-candidate.zip"
+fi
+export DEPLOYMENT_ZIP
+export RUN_CMD="python${PYTHON_INTERPRETER_VERSION} $PYTHON_RUN_CMD"
+
+#Assert pipenv install --system works
```
jonathanmetzman, line 23: put a space after hash in comments.

src/local/butler/deploy.py
```diff
@@ -92,7 +92,8 @@ def _deploy_app_prod(project,
                      yaml_paths,
                      package_zip_paths,
                      deploy_appengine=True,
-                     test_deployment=False):
+                     test_deployment=False,
```
jonathanmetzman, line 95: Should we delete the test_deployment feature?

src/local/butler/package.py
```diff
@@ -97,7 +98,10 @@ def package(revision,
   target_zip_name = constants.LEGACY_ZIP_NAME
   if platform_name:
     if python3:
-      target_zip_name = platform_name + '-3.zip'
+      if stage=="prod":
```
jonathanmetzman, line 101: Spaces in between equal and other other toakens and single quotes.

butler.py
```diff
@@ -259,6 +259,8 @@ def main():
       'package', help='Package clusterfuzz with a staging revision')
   parser_package.add_argument(
       '-p', '--platform', choices=['linux', 'macos', 'windows', 'all'])
+  parser_package.add_argument(
```
jonathanmetzman, line 262: Can we use a different name than "stage" which is used to mean the appengine stuff.
Why do you think it's worth using this?

docker/base/Dockerfile
```diff
@@ -131,7 +138,8 @@ ENV ROOT_DIR $INSTALL_DIRECTORY/clusterfuzz
 ENV UPDATE_WEB_TESTS True
 ENV PYTHONPATH $INSTALL_DIRECTORY/clusterfuzz/src
 ENV RUN_CMD "python3.7 $ROOT_DIR/src/python/bot/startup/run.py"
-ENV DEPLOYMENT_ZIP "linux-3.zip"
+ENV PYTHON_RUN_CMD = "$ROOT_DIR/src/python/bot/startup/run.py"
+#ENV DEPLOYMENT_ZIP "linux-3.zip"
```
jonathanmetzman, line 142: Why do we have commented out code.

butler.py
vitorguidi: No reason in particular, it is the first word that came to my mind to represent this concept.
To have a more clear meaning, this could be a `release`, for which we would have prod and candidate. Wdyt?

docker/base/Dockerfile
vitorguidi: This should have been erased

src/local/butler/deploy.py
vitorguidi: What is this feature? I have never used it

vitorguidi: Oh it is the test-targets thing. We could delete it if this proves successful, we leave the test cluster alive and perform our releases on it with this approach (might be a good idea to create a windows one for that purpose too)

vitorguidi:
> Do we also need to change the windows startup script? And Android and Mac?

I would rather not touch Android and Mac. It is too much friction to implement prog rollout in an environment we do not control.

As far as windows goes, since we do have our own fleet, a change to the windows startup is on the way. Will be in the PR soon

src/local/butler/deploy.py
vitorguidi: Now that I think of it, I would rather keep this feature. At some point in the future I would like to pin releases (git commit) in our deployments, so we can have granular control over what code is executed in each cluster.

For now, this is still useful.

docker/build.sh
```diff
@@ -15,28 +15,29 @@
 
 IMAGES=(
   gcr.io/clusterfuzz-images/base
-  gcr.io/clusterfuzz-images/high-end
+  #gcr.io/clusterfuzz-images/high-end
```
oliverchang, line 18: Are all these changes temporary for testing purposes?

docker/base/setup_clusterfuzz.sh
```diff
@@ -18,15 +18,19 @@ if [ -z "$DEPLOYMENT_BUCKET" ]; then
...
 if [[ -z "$LOCAL_SRC" ]]; then
   # Set up ClusterFuzz
   if [[ -d clusterfuzz ]]; then
     rm -rf clusterfuzz
   fi
 
-  gsutil cp gs://$DEPLOYMENT_BUCKET/$DEPLOYMENT_ZIP .
-  unzip -q -o $DEPLOYMENT_ZIP
+  # DEPLOYMENT_ZIP might be test-deployment/linux-3.zip, so we do not extract DEPLOYMENT_ZIP directly
+
```
oliverchang, line 31: nit: remove newline.

src/clusterfuzz/_internal/bot/tasks/update_task.py
```diff
@@ -39,10 +39,18 @@
 from clusterfuzz._internal.system import process_handler
 from clusterfuzz._internal.system import shell
 
+CLUSTERFUZZ_RELEASE = os.getenv('CLUSTERFUZZ_RELEASE')
```
oliverchang, line 42: nit: just `os.getenv('CLUSTERFUZZ_RELEASE', 'prod')`

src/clusterfuzz/_internal/bot/tasks/update_task.py
```diff
@@ -39,10 +39,18 @@
...
+if not CLUSTERFUZZ_RELEASE:
+  CLUSTERFUZZ_RELEASE = 'prod'
+
 TESTS_LAST_UPDATE_KEY = 'tests_last_update'
 TESTS_UPDATE_INTERVAL_DAYS = 1
 
-MANIFEST_FILENAME = 'clusterfuzz-source.manifest.3'
+REMOTE_MANIFEST_FILENAME = 'clusterfuzz-source.manifest'
```
oliverchang, line 49: It feels like there's a lot of repetition everywhere with this logic to compute the suffixes. Can we instead just create helpers for this that can be used everywhere?

e.g. something along the lines of 

```python
def get_source_zip_suffix():
  ...

def get_source_zip_name():
  ...
```

We can remove some of these constants if necessary and just make them dynamically computed based on reading from env.

src/local/butler/deploy.py
```diff
@@ -525,11 +542,18 @@ def execute(args):
     deploy_k8s = False
     deploy_zips = True
 
+  release = args.release
+  if args.release == 'candidate':
```
oliverchang, line 546: can we just use `args.release` directly? `argparse` validates the choices for us right?

src/local/butler/package.py
```diff
@@ -136,11 +140,18 @@ def package(revision,
 
 
 def execute(args):
+  """Execute the butler package command."""
+  release = 'prod'
+  if args.release == 'candidate':
```
oliverchang, line 145: same comment here as above re just using `args.release` directly.

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -553,7 +553,9 @@ def emit(level, message, exc_info=None, **extras):
               'path': path_name,
               'line': line_number,
               'method': method_name
-          }
+          },
+          'release': os.environ.get('CLUSTERFUZZ_RELEASE', 'prod'),
```
oliverchang, line 557: nit: let's create a helper for this too. e.g. `get_clusterfuzz_release()` in `utils`.

docker/build.sh
vitorguidi: Yes, the images that do not belong in the tester workflow have been commented. Will uncomment these before merging

src/clusterfuzz/_internal/metrics/logs.py
vitorguidi: I noticed the release/docker_image fields are not being logged over to GCP. This requires some debugging

src/local/butler/deploy.py
vitorguidi: Good catch. It does
```
clusterfuzzvitorguidi@vitorguidi:~/projects/clusterfuzz$ python3.11 butler.py package --platform=linux --release=loremipsum
error: argument -r/--release: invalid choice: 'loremipsum' (choose from 'prod', 'candidate')
usage: butler.py package [-h] [-p {linux,macos,windows,all}] [-r {prod,candidate}]

options:
  -h, --help            show this help message and exit
  -p {linux,macos,windows,all}, --platform {linux,macos,windows,all}
  -r {prod,candidate}, --release {prod,candidate}
  ```

File: src/clusterfuzz/_internal/bot/tasks/update_task.py (hunk 11 above)
Comment (vitorguidi) [reply src/clusterfuzz/_internal/bot/tasks/update_task.py]: Moved this to clusterfuzz._internal.base.utils so butler package/deploy and update_task can share this code

File: docker/build.sh [hunk 15]
```diff
@@ -31,7 +31,7 @@ IMAGES=(
 )
 
 function docker_push {
```
jonathanmetzman, line 33: Remove this commented out line please.

docker/fuchsia/start.sh
```diff
@@ -13,6 +13,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+DEPLOYMENT_ZIP="linux-3.zip"
```
jonathanmetzman, line 16: I think we might need to rebuild the google images for this to be used.

docker/oss-fuzz/host/Dockerfile
```diff
@@ -20,4 +20,5 @@ ENV NUM_WORKERS_PER_HOST 16
 ENV DISABLE_MOUNTS True
 ENV TRUSTED_HOST True
 
+
```
jonathanmetzman, line 23: remove this line

src/clusterfuzz/_internal/base/utils.py
```diff
@@ -227,6 +227,38 @@ def get_application_id():
...
+  if sys.version_info.major == 3:
+    suffix += '-3'
+  if release == 'candidate':
+    suffix += '-candidate'
+  return suffix
+
+
+def get_platform_deployment_filename(platform, release):
+  """Return the platform deployment filename."""
+  # Expects linux, macos or windows
```
jonathanmetzman, line 253: End this with a period.

jonathanmetzman (commented):
lgtm with nits

docker/fuchsia/start.sh
vitorguidi: The plan is to use progressive rollout to validate on internal, we should not need to perform this operation for now.

docker/build.sh
vitorguidi: This is the line that pushes latest, forgot to uncomment this guy to restore the file to its original state.

oliverchang (approved):

PR #4043, opened 2024-06-20
Delete unused spiel about sandboxed jobs.

jonathanmetzman:
It was deceptive and isn't used anymore.


pgrace-google (approved):
ooh i should have deleted this earlier - thank you!

PR #4048, opened 2024-06-20
Make logs functions consistent with stdlib's logging module

jonathanmetzman:
Superseeds: https://github.com/google/clusterfuzz/pull/3793.
Examples:
1. Most logging libraries don't make you write "log_" before each function, but in CF we do. (e.g. "logs.log_error" instead of "logs.error".
2. Most logging libraries call it "warning" not "warn". CF does "warn".
3. Most logging libraries have an "info" function. In CF, this is called "log".


src/appengine/handlers/external_update.py
```diff
@@ -41,7 +41,7 @@ def _mark_as_fixed(testcase, revision):
 def _mark_errored(testcase, revision, error):
   """Mark testcase as errored out."""
   message = 'Received error from external infra, marking testcase as NA.'
-  logs.log_warn(message, error=error, testcase_id=testcase.key.id())
+  logs.warning(message, error=error, testcase_id=testcase.key.id())
```
vitorguidi, line 44: Can you elaborate a bit on why the current state is inconsistent with stdlib? It is not entirely clear from just reading this code

jonathanmetzman: I updated the description to better explain this. But in short, the CF logging library is hard to use for anyone that programs in other codebases.

vitorguidi: A couple tests broke because mocks failed due to a [change in the function signatures](https://github.com/google/clusterfuzz/actions/runs/9604133225/job/26488916685?pr=4048). LGTM besides that

PR #4049, opened 2024-06-20
Remove syzkaller

jonathanmetzman:
It's not used anymore, keeping it around complicates things, if we need it again, we can revert.


src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -1243,9 +1198,6 @@ def setup_regular_build(revision,
     build_class = build_setup_host.RemoteRegularBuild
   elif environment.platform() == 'FUCHSIA':
     build_class = FuchsiaBuild
-  elif (environment.is_android_cuttlefish() and
```
jonathanmetzman, line 1246: Is this an OK change?

jonathanmetzman:
TODO: THere are some changes to adb that probably should not be made.

src/clusterfuzz/_internal/build_management/build_manager.py
jonathanmetzman: @marktefftech

src/clusterfuzz/_internal/crash_analysis/stack_parsing/stack_analyzer.py
```diff
@@ -27,8 +27,7 @@ def linkify_kernel_or_lkl_stacktrace_if_needed(crash_info):
   """Linkify Android Kernel or lkl stacktrace."""
   kernel_prefix = ''
   kernel_hash = ''
-  if (environment.is_android_kernel() and
```
jonathanmetzman, line 30: I think this change is from a revert but @marktefftech do you know what it should be now?

src/clusterfuzz/_internal/build_management/build_manager.py
marktefftech: Yes this should be fine.

marktefftech:
Acknowledging that you've tagged me for review at individual LOC, it's my understanding that our kernel fuzzing is now completely decoupled from ClusterFuzz (replaced by Syzkaller).

So, although I don't own the kernel code, I think it's safe to go ahead and eliminate anything to do with kernel fuzzing assuming it's only being used by Android.

marktefftech (approved):

jonathanmetzman:
> (Let me know if you want me to loop in a member of the kernel team, but like I said, this should be fine).

Yeah if you can loop in someone who worked on this and can confirm we don't need it anymore, that would be great.
Maybe @hzawawy ?

eprucka3:
> Yeah if you can loop in someone who worked on this and can confirm we don't need it anymore, that would be great.

I can confirm; we no longer use Clusterfuzz with our Syzkaller infra.

hzawawy:
> > Yeah if you can loop in someone who worked on this and can confirm we don't need it anymore, that would be great.
> 
> I can confirm; we no longer use Clusterfuzz with our Syzkaller infra.

+1

Issue #4047, opened 2024-06-20
Investigate if chromium-tests-syncer works


jonathanmetzman:
I saw a linting error recently that implies it's been nonfunctional for a while.

PR #4054, opened 2024-06-20
Fix linting error in chromium-tests-syncer

jonathanmetzman:
Related: #4047

### DIFF
--- a/src/python/other-bots/chromium-tests-syncer/run.py
+++ b/src/python/other-bots/chromium-tests-syncer/run.py
@@ -84,7 +84,8 @@ def unpack_crash_testcases(crash_testcases_directory):
 
     # Un-pack testcase.
     try:
-      setup.unpack_testcase(testcase)
+      testcase_download_url = setup.get_signed_testcase_download_url(testcase)
+      setup.unpack_testcase(testcase, testcase_download_url)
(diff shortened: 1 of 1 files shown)


src/python/other-bots/chromium-tests-syncer/run.py
```diff
@@ -84,7 +84,8 @@ def unpack_crash_testcases(crash_testcases_directory):
 
     # Un-pack testcase.
     try:
-      setup.unpack_testcase(testcase)
```
vitorguidi, line 87: The PR is about fixing a linting error, but before this change it seems like this would always [return None](https://github.com/google/clusterfuzz/blob/6d5b96fbfc2ab9bb02e08a34a447f3c9302c81c0/src/clusterfuzz/_internal/bot/tasks/setup.py#L378). Did it work before? Will it work now?

jonathanmetzman: I'm confused what the question is.
To be honest, I'm not sure if this code is even running anymore.

jonathanmetzman: I suspect not because I haven't seen this error anywhere.

Issue #4058, opened 2024-06-21
Deployment broken by python package upgrade


jonathanmetzman:
```
python butler.py run setup --config-dir ../clusterfuzz-config/configs/internal --non-dry-run
None
afl fuzzer already exists
Updating stats metrics.
libFuzzer fuzzer already exists
Updating stats metrics.
honggfuzz fuzzer already exists
Updating stats metrics.
googlefuzztest fuzzer already exists
Updating stats metrics.
syzkaller fuzzer already exists
Updating stats metrics.
centipede fuzzer already exists
Updating stats metrics.
Template with name afl already exists.
Template with name centipede already exists.
Template with name engine_asan already exists.
Template with name engine_msan already exists.
Template with name engine_ubsan already exists.
Template with name honggfuzz already exists.
Template with name googlefuzztest already exists.
Template with name libfuzzer already exists.
Template with name syzkaller already exists.
Template with name prune already exists.
Traceback (most recent call last):
  File "butler.py", line 391, in <module>
    sys.exit(main())
  File "butler.py", line 377, in main
    return command.execute(args)
  File "src/local/butler/run.py", line 48, in execute
    script.execute(args)
  File "src/local/butler/scripts/setup.py", line 401, in execute
    setup_metrics(args.non_dry_run)
  File "src/local/butler/scripts/setup.py", line 372, in setup_metrics
    project_path = client.project_path(project_name)
AttributeError: 'MetricServiceClient' object has no attribute 'project_path'
```

Issue #4059, opened 2024-06-21
Upgrade to python3.11


jonathanmetzman:
All PRs related to this effort should link here.

eduarddfinity:
Hello clusterfuzz team

do you have an ETA for migrating to python3.11? 
Unfortunately we've been locked out of our clusterfuzz instance since last month, because cross origin redirect login is not supported in firebase. I know this has been fixed in clusterfuzz but we cannot update our instance as app engine no longer allows python3.7 apps to be deployed.

Thanks in advance!

oliverchang:
I'll let @vitorguidi comment in more detail on this, but we are actively working on this upgrade and should hopefully be completing this in the coming weeks / month.

vitorguidi:
> Hello clusterfuzz team
> 
> do you have an ETA for migrating to python3.11? Unfortunately we've been locked out of our clusterfuzz instance since last month, because cross origin redirect login is no longer supported in firebase. I know this has been fixed in clusterfuzz but we cannot update our instance as app engine no longer allows python3.7 apps to be deployed.
> 
> Thanks in advance!

We already have all dependencies bumped, the next step is switching the interpreter and performing a canary with production instances. This should be done next week, and it will be possible to more accurately gauge how complex the process will be, but Oliver's estimate is what I am expecting.

vitorguidi:
#4025 has already been merged and we have been running clusterfuzz in production for >1 week at Python3.11, so it should be good now @eduarddfinity

nguyendon:
@vitorguidi are you all running from a stable release branch or are you just running off of `master`?

nguyendon:
@jonathanmetzman is running off of master just for an internal instance or is that what OSS-Fuzz runs on as well?

PR #4060, opened 2024-06-21
Fix deploy.

jonathanmetzman:
Metric setup is broken because of Python package upgrade.

Fixes: https://github.com/google/clusterfuzz/issues/4058
Related: https://github.com/google/clusterfuzz/issues/4059


jonathanmetzman:
We can avoid more issues like that by using the code from my downgrade.

PR #4063, opened 2024-06-25
Ensure correct ordering of 'time_series' list.

aditya-wazir:
Maintain separate 'gauge_series' for gauge metrics. Combined gauge series with the main 'time_series' list only before the flush operation ensuring that all metric are in order based on corresponding 'start_time' in 'time_series' list

### DIFF
--- a/src/clusterfuzz/_internal/metrics/monitor.py
+++ b/src/clusterfuzz/_internal/metrics/monitor.py
@@ -74,6 +74,10 @@ class _MockMetric:
     return self._mock_method
 
 
+def _time_series_sort_key(ts):
+  return ts.points[-1].interval.start_time
+
+
 class _FlusherThread(threading.Thread):
   """Flusher thread."""
 
@@ -81,9 +85,6 @@ class _FlusherThread(threading.Thread):
     super().__init__()
     self.daemon = True
     self.stop_event = threading.Event()
-
-  def time_series_sort_key(self, ts):
-    return ts.points[-1].interval.start_time
 
   def run(self):
     """Run the flusher thread."""
@@ -120,12 +121,12 @@ class _FlusherThread(threading.Thread):
           time_series.append(series)
 
           if len(time_series) == MAX_TIME_SERIES_PER_CALL:
-            time_series.sort(key=self.time_series_sort_key)
+            time_series.sort(key=_time_series_sort_key)
             create_time_series(name=project_path, time_series=time_series)
             time_series = []
 
         if time_series:
-          time_series.sort(key=self.time_series_sort_key)
+          time_series.sort(key=_time_series_sort_key)
           create_time_series(name=project_path, time_series=time_series)
       except Exception:
         if environment.is_android():
--- a/src/clusterfuzz/_internal/metrics/monitor.py
+++ b/src/clusterfuzz/_internal/metrics/monitor.py
@@ -82,6 +82,9 @@ class _FlusherThread(threading.Thread):
     self.daemon = True
     self.stop_event = threading.Event()
 
+  def time_series_sort_key(self, ts):
+    return ts.points[-1].interval.start_time
+
   def run(self):
     """Run the flusher thread."""
     create_time_series = _retry_wrap(_monitoring_v3_client.create_time_series)
@@ -117,18 +120,12 @@ class _FlusherThread(threading.Thread):
           time_series.append(series)
 
           if len(time_series) == MAX_TIME_SERIES_PER_CALL:
-            time_series.sort(
-                key=
-                lambda ts: (ts.points[-1].interval.start_time.second, ts.points[-1].interval.start_time.nanosecond)  # pylint: disable=line-too-long
-            )
+            time_series.sort(key=self.time_series_sort_key)
             create_time_series(name=project_path, time_series=time_series)
             time_series = []
 
         if time_series:
-          time_series.sort(
-              key=
-              lambda ts: (ts.points[-1].interval.start_time.second, ts.points[-1].interval.start_time.nanosecond)  # pylint: disable=line-too-long
-          )
+          time_series.sort(key=self.time_series_sort_key)
           create_time_series(name=project_path, time_series=time_series)
       except Exception:
         if environment.is_android():
--- a/src/clusterfuzz/_internal/metrics/monitor.py
+++ b/src/clusterfuzz/_internal/metrics/monitor.py
@@ -116,8 +116,8 @@ class _FlusherThread(threading.Thread):
                    f'start_time: {metric_st_sec}.{metric_st_ns}, '
                    f'end_time: {metric_et_sec}.{metric_et_ns}')
 
-          if (series.metric_kind ==
-              metric_pb2.MetricDescriptor.MetricKind.GAUGE):
+          if (series.metric_kind == metric_pb2.MetricDescriptor.MetricKind.GAUGE  # pylint: disable=no-member
+             ):
             gauge_series.append(series)
           else:
             time_series.append(series)
--- a/src/clusterfuzz/_internal/metrics/monitor.py
+++ b/src/clusterfuzz/_internal/metrics/monitor.py
@@ -94,7 +94,6 @@ class _FlusherThread(threading.Thread):
           return
 
         time_series = []
-        gauge_series = []
         end_time = time.time()
         for metric, labels, start_time, value in _metrics_store.iter_values():
           if (metric.metric_kind == metric_pb2.MetricDescriptor.MetricKind.GAUGE  # pylint: disable=no-member
@@ -115,21 +114,25 @@ class _FlusherThread(threading.Thread):
                    f'metric_kind: {series.metric_kind}, '
                    f'start_time: {metric_st_sec}.{metric_st_ns}, '
                    f'end_time: {metric_et_sec}.{metric_et_ns}')
-
-          if (series.metric_kind == metric_pb2.MetricDescriptor.MetricKind.GAUGE  # pylint: disable=no-member
-             ):
-            gauge_series.append(series)
-          else:
-            time_series.append(series)
+          time_series.append(series)
 
           if len(time_series) == MAX_TIME_SERIES_PER_CALL:
-            time_series.extend(gauge_series)
+            time_series.sort(
+                key=lambda ts: (
+                    ts.points[-1].interval.start_time.second,
+                    ts.points[-1].interval.start_time.nanosecond
+                )
+            )
             create_time_series(name=project_path, time_series=time_series)
-            gauge_series = []
             time_series = []
 
         if time_series:
-          time_series.extend(gauge_series)
+          time_series.sort(
+              key=lambda ts: (
+                  ts.points[-1].interval.start_time.second,
+                  ts.points[-1].interval.start_time.nanosecond
+              )
+          )
           create_time_series(name=project_path, time_series=time_series)
       except Exception:
         if environment.is_android():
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/metrics/monitor.py
```diff
@@ -108,13 +109,21 @@ def run(self):
...
 
           if len(time_series) == MAX_TIME_SERIES_PER_CALL:
+            time_series.extend(gauge_series)
             create_time_series(project_path, time_series)
+            gauge_series = []
             time_series = []
 
         if time_series:
+          time_series.extend(gauge_series)
           create_time_series(project_path, time_series)
```
marktefftech, line 127: Did you have the chance to run this locally and verify that the data structure `monitoring_v3.enums.MetricDescriptor.MetricKind.GAUGE` is compatible downstream in `create_time_series`?

I just want to make sure this won't cause any unexpected issues when creating the time series.

It may make sense to add or modify the tests in `src/clusterfuzz/_internal/tests/core/metrics/monitor_test.py` to verify this

aditya-wazir: Hi Mark,
Replaced monitoring_v3.enums.MetricDescriptor.MetricKind.GAUGE with metric_pb2.MetricDescriptor.MetricKind.GAUGE, and it's now working fine. However, when verifying whether our changes are compatible downstream in create_time_series, I'm encountering a "Permission Denied" issue when trying to call create_time_series.
 
This change sorts the existing series which includes the GAUGE metric in an ascending order of start_time. No new type of metrics has been added. Hence the existing MetricKind.GAUGE should be compatible with downstream in create_time_series.
 
Additionally, added a test for GaugeMetrics to monitor_test.py. Our changes include logical modifications to the FlusherThread.run() API. However, the corresponding test is marked with @unittest.skip, so it does not actually test our changes.

src/clusterfuzz/_internal/metrics/monitor.py
```diff
@@ -114,13 +115,21 @@ def run(self):
...
-          time_series.append(series)
+
+          if (series.metric_kind == metric_pb2.MetricDescriptor.MetricKind.GAUGE  # pylint: disable=no-member
+             ):
+            gauge_series.append(series)
+          else:
+            time_series.append(series)
 
           if len(time_series) == MAX_TIME_SERIES_PER_CALL:
+            time_series.extend(gauge_series)
```
letitz, line 126: It seems to me we should check the combined length of `time_series` and `gauge_series` against `MAX_TIME_SERIES_PER_CALL` if we're to concatenate them. Otherwise we get this weird behavior where the max only applies to a subset of the entries in the list we send to the monitoring service, whose length is effectively unbounded.

src/clusterfuzz/_internal/tests/core/metrics/monitor_test.py
```diff
@@ -157,6 +157,21 @@ def test_cumulative_distribution_metric_fixed(self):
...
+  def test_gauge_metric_failure(self):
+    """Test gauge metric failure."""
+    self.mock.check_module_loaded.return_value = False
+    gauge = monitor.GaugeMetric(
+        'g', 'desc', field_spec=[
+            monitor.StringField('name'),
+        ])
+    gauge.set(5)
+    self.assertIsInstance(gauge, monitor._MockMetric)
+
```
letitz, line 174: Can we have a test for the thing that is actually changing in this PR? i.e. the fact that the time series sent to the monitoring service is actually ordered?

src/clusterfuzz/_internal/metrics/monitor.py
letitz, line 126: IIUC this relies on the fact that gauge metric points are tagged with `start_time = end_time`, which makes them out of order with other metrics. However it seems possible to have time series points that have an even later start time, e.g.:

- time series X: start = 1, end = 2
- gauge series Y: start = 3, end = 3
- time series Z: start = 4, end = 5

Right?

It seems safer to me to sort the time series list in the way that the monitoring service expects it to be before sending it off.

letitz (commented):
Drive-by review. I am very glad someone is finally fixing this - we have lots of spammy logs about out-of-order time series points. Thanks for sending this in!

src/clusterfuzz/_internal/metrics/monitor.py
marktefftech: Ack, thanks for adding the test as well.

src/clusterfuzz/_internal/metrics/monitor.py
aditya-wazir: Done. Added sort based solution instead.

src/clusterfuzz/_internal/tests/core/metrics/monitor_test.py
aditya-wazir: Added TestFlusherThread test. In the following test we are using 2 metrics , i.e,  BOT_COUNT(Gauge) and HOST_INCONSISTENT_COUNT(Cumulative) in this order and mocking create_time_series method to retrieve 'time_series' list passed to create_time_series in run() api.
Once time_series list is retrieved we are checking whether the start_time of corresponding metrics is in ascending order or not. Based on that test is passing with fix applied and failing if no fix is applied

src/clusterfuzz/_internal/metrics/monitor.py
aditya-wazir: Done. Added Sorting based solution.

src/clusterfuzz/_internal/tests/core/metrics/monitor_test.py
```diff
@@ -157,6 +158,47 @@ def test_cumulative_distribution_metric_fixed(self):
...
+    monitor._flusher_thread = monitor._FlusherThread()
+    monitor.FLUSH_INTERVAL_SECONDS = 1
+    monitoring_metrics.BOT_COUNT.set(1, {'revision': '1'})
+    monitoring_metrics.HOST_INCONSISTENT_COUNT.increment()
+    os.environ['BOT_NAME'] = 'bot-1'
+
+    monitor._flusher_thread.start()
+    mock_create_time_series = mock_client.return_value.create_time_series
+    while not mock_create_time_series.called:
+      pass
```
letitz, line 195: You can synchronize this with a `Queue` instead:

```py
call_queue = queue.Queue()
monitor._monitoring_v3_client.create_time_series.side_effect = (
    lambda *args : call_queue.put(args))
[...]

args = call_queue.get(timeout=10)
time_series = args[2]['time_series']

File: src/clusterfuzz/_internal/metrics/monitor.py [hunk 5]
```diff
@@ -117,10 +117,18 @@ def run(self):
           time_series.append(series)
 
           if len(time_series) == MAX_TIME_SERIES_PER_CALL:
+            time_series.sort(
+                key=
+                lambda ts: (ts.points[-1].interval.start_time.second, ts.points[-1].interval.start_time.nanosecond)  # pylint: disable=line-too-long
```
letitz, line 122: Let's extract this to a helper function:

```py
def time_series_sort_key(ts):
  return ts.points[-1].interval.start_time
```

src/clusterfuzz/_internal/tests/core/metrics/monitor_test.py
letitz: Thanks very much!

src/clusterfuzz/_internal/tests/core/metrics/monitor_test.py
```diff
@@ -157,6 +158,47 @@ def test_cumulative_distribution_metric_fixed(self):
...
+    monitor._flusher_thread.stop()
+    time_series = mock_create_time_series.mock_calls[0][2]['time_series']
+    for i in range(1, len(time_series)):
+      self.assertLessEqual(time_series[i - 1].points[0].interval.start_time,
+                           time_series[i].points[0].interval.start_time)
```
letitz, line 200: This comparison working suggests we don't need to manually construct a tuple of seconds and nanoseconds as the sort key - these timestamp objects are already comparable?

letitz:
I went looking for the docs for creating time series. Here's what they say about the array of time series objects [1]:

> Adds at most one data point to each of several time series. The new data point must be more recent than any other point in its time series. Each TimeSeries value must fully specify a unique time series by supplying all label values for the metric and the monitored resource.

As for time series objects [2]:

> points[]
> [...]
> When creating a time series, this field must contain exactly one point

The first one suggests to me that we only need to ensure that points are sorted within each time series. Sorting the list globally should also work, but this still begs the question: why are gauge series the problem we're trying to solve?

[1] https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.timeSeries/create  
[2] https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TimeSeries

src/clusterfuzz/_internal/tests/core/metrics/monitor_test.py
aditya-wazir: Done.

src/clusterfuzz/_internal/metrics/monitor.py
aditya-wazir: Done.

src/clusterfuzz/_internal/tests/core/metrics/monitor_test.py
aditya-wazir: Yes, timestamp objects are comparable. changed sorting condition accordingly.

aditya-wazir:
> I went looking for the docs for creating time series. Here's what they say about the array of time series objects [1]:
> 
> > Adds at most one data point to each of several time series. The new data point must be more recent than any other point in its time series. Each TimeSeries value must fully specify a unique time series by supplying all label values for the metric and the monitored resource.
> 
> As for time series objects [2]:
> 
> > points[]
> > [...]
> > When creating a time series, this field must contain exactly one point
> 
> The first one suggests to me that we only need to ensure that points are sorted within each time series. Sorting the list globally should also work, but this still begs the question: why are gauge series the problem we're trying to solve?
> 
> [1] https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.timeSeries/create [2] https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TimeSeries

While it's true that points within a time series must be sorted, the requirement for creating time series objects is that the timestamps across all series must be in chronological order.
The challenge with gauge metrics is that as gauge metric's start_time is assigned the current timestamp('start_time = end_time', where end_time = time.time()) within the _FlusherThread.run() function, they can introduce out-of-order timestamps when combined with other metric types, which conflicts with the chronological order requirement for creating time series objects.
To ensure correct time series creation, we're implementing a sorting based solution. This approach ensures that the timestamps are in ascending order, meeting the API's requirements.

src/clusterfuzz/_internal/metrics/monitor.py
```diff
@@ -82,6 +82,9 @@ def __init__(self):
     self.daemon = True
     self.stop_event = threading.Event()
 
+  def time_series_sort_key(self, ts):
```
letitz, line 85: Nit: This does not need to be a public method on this class. We can make it a private free function.

letitz (approved):
LGTM with a single nit. Thanks for working on this!

src/clusterfuzz/_internal/metrics/monitor.py
aditya-wazir: Done.

marktefftech (approved):

PR #4069, opened 2024-06-29
use request host value instead of auth domain when proxying

quanyangsc:
This modifies how the target URL for proxying auth requests is instantiated, instead of using auth.auth_domain, it uses request.host which achieves the same behavior, but is more robust in the sense that it'd allow setups that are behind some kind of beyond corp/zero trust proxy more easily.

### DIFF
--- a/src/appengine/handlers/auth.py
+++ b/src/appengine/handlers/auth.py
@@ -33,7 +33,7 @@ class Handler(base_handler.Handler):
     """Handle a get request."""
     # We use `request.url` which is already the full URL.
     del extra
-    target_url = request.url.replace(auth.auth_domain(),
+    target_url = request.url.replace(request.host,
                                      auth.real_auth_domain(), 1)
     logging.info('Forwarding auth request to: %s', target_url)
     response = requests.get(target_url, timeout=60)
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
Can you be more clear about what problem this is solving?

quanyangsc:
@jonathanmetzman : sure thing.

Imagine a set up where clusterfuzz is behind a virtualhost proxy, the front domain is `front-server.com`, and the back domain is `back-server.com`.
The main blocker here is that for the oauth proxy to work, `auth.auth_domain()` needs to return `front-server.com` so that the oauth redirect URI works correctly. But, in this setup,  the value of `request.url` is `https://back-server.com/__/auth/path`, and so `request.url.replace("front-server.com", auth.real_auth_domain(), 1)` does nothing.

All the code is trying to do here is replace the host value in the URL `https://back-server.com/__/auth/path` with the value returned from `auth.real_auth_domain()`, which we can achieve with the change proposed in this PR.

quanyangsc:
@jonathanmetzman I can't view the reason why the clusterfuzz-testing-public job is failing, could you take a look?

PR #4088, opened 2024-07-03
Chore/revert uworker safe

vitorguidi:
We apparently had a drop in succesful batch runs from #4000 and subsequent PRs, so reverting for good measure


jonathanmetzman:
I don't think we should revert this yet.
1. The brekage is not user-visible yet as it only affects fuzz_task.
2. I am addressing the errors now and should have fixes for all of them in the next 30 minutes.
3. The errors don't appear easy to reproduce locally so having them occur in prod is valuable teaching moments.

jonathanmetzman:
I also somewhat doubt the batch metric is correctly reading this. Most batch jobs are probably not the result of fuzzing from that day.

vitorguidi:
> I also somewhat doubt the batch metric is correctly reading this. Most batch jobs are probably not the result of fuzzing from that day.

What would be desirable for us to debug this kind of issue in the future? Success rate, given a fuzzing task start time?

PR #3946, opened 2024-04-15
Store the backup corpus using a signed url

alhijazi:

PR #4110, opened 2024-07-09
Take CORPUS_FUZZER_NAME_OVERRIDE into account when generating backup …

alhijazi:
…signed urls

This fixes the issue introduced in
https://github.com/google/clusterfuzz/pull/3946/files#diff-4e357c1cb796c84a085445e5c0182c2a0bc5d46c0271c1fe594a7d41947ed9d8L511

Before the change above, we used the corpus.engine for the backup url which took into account the CORPUS_FUZZER_NAME_OVERRIDE.


oliverchang (approved):

jonathanmetzman (approved):
LGTM

Can you monitor this to make sure it's working as intended? Maybe add some temporary logging?

PR #4119, opened 2024-07-16
Added Debugging Logs

aditya-wazir:
Logs have been added to debug and resolve the following issues:
1. OSError: [Error 28] No space left on device
2. AttributeError: 'NoneType' object has no attribute 'stdin'
3. Line is too long for a stacktrace

### DIFF
--- a/src/clusterfuzz/_internal/system/shell.py
+++ b/src/clusterfuzz/_internal/system/shell.py
@@ -261,6 +261,13 @@ def create_directory(directory, create_intermediates=False, recreate=False):
     else:
       os.mkdir(directory)
   except:
+    dfh = subprocess.run(
+        "df -h", shell=True, capture_output=True, check=False, text=True)
+    dfhi = subprocess.run(
+        "df -hi", shell=True, capture_output=True, check=False, text=True)
+    logs.info(
+        f"DEBUG:\nDisk Free(h) :\n{dfh.stdout}\nDisk Free(hi):\n{dfhi.stdout}")
+
     logs.error('Unable to create directory %s.' % directory)
     return False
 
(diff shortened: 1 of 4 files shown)


src/clusterfuzz/_internal/system/shell.py
```diff
@@ -261,6 +261,13 @@ def create_directory(directory, create_intermediates=False, recreate=False):
     else:
       os.mkdir(directory)
   except:
+    dfh = subprocess.run(
```
oliverchang, line 264: This will always break on Windows right? Can we guard this with a check for linux?

jonathanmetzman:
nit: single quotes please.

src/clusterfuzz/_internal/system/shell.py
aditya-wazir: Yes the `df` command breaks on Windows. I have added an appropriate check to avoid this issue.

vitorguidi: Is there any abstraction we can use , that would work on windows?
Apparently, psutil is able to[ fetch disk information ](https://psutil.readthedocs.io/en/latest/index.html)on a platform agnostic way.

It would be nicer to not use any platform specific logic here, if possible

oliverchang: +1. That reminded me we actually do something similiar using psutil here already: https://github.com/google/clusterfuzz/blob/578b401b655e95211b6ba9b805986832622b213f/src/clusterfuzz/_internal/system/shell.py#L346 

I'm not sure if psutil provides any information on inodes though (`df -i`).

aditya-wazir: Thank you for the suggestion. We use these commands to investigate the `OSError: no space left on device`. While `psutil` covers disk usage, it doesn't provide inode information. Inode count is essential because running out of inodes can also cause this error. Therefore, it is preferable to use the `df` command.

oliverchang (approved):

PR #4124, opened 2024-07-26
Improve monitoring logs

letitz:
* Removes spammy debug logs from `FlusherThread.run()`
* Adds warning logs when time series uploads fail
* Adds exception details when the flusher thread runs into an exception

Also touches a couple more things that the linter was complaining about.


letitz:
This builds on #4125 and since github does not do diffbases well, it includes both PRs' changes. Still, I think it can be reviewed (esp. if you filter only for the right commits).

src/clusterfuzz/_internal/metrics/monitor.py
```diff
@@ -300,8 +309,8 @@ def monitoring_v3_time_series(self, time_series, labels, start_time, end_time,
     time_series.metric_kind = self.metric_kind
     time_series.value_type = self.value_type
 
-    interval = monitoring_v3.types.TimeInterval()
-    point = monitoring_v3.types.Point(interval=interval)
+    interval = _TimeInterval()
```
vitorguidi, line 312: Can you share some evidence that this change is able to send metrics?
Metrics are not fully covered in automated testing, so running this [unit test locally](https://github.com/google/clusterfuzz/blob/412b2e16153e29a0f2ce0ddf4cb4333af280e399/src/clusterfuzz/_internal/tests/core/metrics/monitor_test.py#L161) will ensure metrics are still making it to GCP

letitz: Ran the `monitor_test.py` file, which showed me that the path to `TimeSeries` was not correct (it's in `monitoring_v3.types`, not `monitoring_v3.types.metric`. The docs [1] agree. Not sure how this worked before?

As for running `JonathanDebugTest`, I just tried and it exploded with:

```
ERROR: test_flush (tests.core.metrics.monitor_test.JonathanDebugTest)
Sets up the flusher thread so we can debug it.
  File "/usr/local/google/home/titouan/code/clusterfuzz/src/clusterfuzz/_internal/tests/core/metrics/monitor_test.py", line 177, in test_flush
    monitor._flusher_thread.run()
  File "/usr/local/google/home/titouan/code/clusterfuzz/src/clusterfuzz/_internal/metrics/monitor.py", line 110, in run
    project_path = _monitoring_v3_client.common_project_path(  # pylint: disable=no-member
AttributeError: 'MetricServiceClient' object has no attribute 'common_project_path'

```

Which is weird, since `MetricServiceClient` definitely has a `common_project_path` attribute according to the docs [2].

[1] https://cloud.google.com/python/docs/reference/monitoring/2.15.1/google.cloud.monitoring_v3.types#timeseries
[2] https://cloud.google.com/python/docs/reference/monitoring/2.15.1/google.cloud.monitoring_v3.services.metric_service.MetricServiceClient#google_cloud_monitoring_v3_services_metric_service_MetricServiceClient_common_project_path

letitz: Ah, I think my `src/third_party` folder is wacky, trying to fix it.

letitz: Ok, the test ran locally with no issues for a minute.

jonathanmetzman (approved):
lgtm

PR #4125, opened 2024-07-26
Fix lint issues.

letitz:
Encountered when drafting PR #4124. The tree was in a bad state, this fixes some of it.

### DIFF
--- a/src/clusterfuzz/_internal/metrics/monitor.py
+++ b/src/clusterfuzz/_internal/metrics/monitor.py
@@ -51,6 +51,12 @@ INITIAL_DELAY_SECONDS = 16
 MAXIMUM_DELAY_SECONDS = 2 * 60  # 2 minutes.
 MAX_TIME_SERIES_PER_CALL = 200
 
+# Since `monitoring_v3` is only conditionally imported, pylint complains about
+# any accesses to its members. Define type aliases here once and for all, for
+# use below.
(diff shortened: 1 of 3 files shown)


src/clusterfuzz/_internal/metrics/monitor.py
```diff
@@ -51,6 +51,12 @@
 MAXIMUM_DELAY_SECONDS = 2 * 60  # 2 minutes.
 MAX_TIME_SERIES_PER_CALL = 200
 
+# Since `monitoring_v3` is only conditionally imported, pylint complains about
+# any accesses to its members. Define type aliases here once and for all, for
+# use below.
+_TimeInterval = monitoring_v3.types.TimeInterval  # pylint: disable=no-member
```
oliverchang, line 57: This will cause an exception in the case that `monitoring_v3 = None` above right?

oliverchang (commented):
thanks!

src/clusterfuzz/_internal/metrics/monitor.py
oliverchang: I reverted this specific change in https://github.com/google/clusterfuzz/pull/4125/commits/ade6f127bce74b9b932c0ec3c119bba140b14107 for now.

oliverchang (approved):

letitz:
Thanks for fixing this! Sorry, I was extremely busy yesterday and did not find the time to respond to your comments.

PR #4126, opened 2024-07-29
progression: Fix handling of errors.BuildNotFound in build_setup.

oliverchang:
This was necessary to properly close testcases where the build is no longer valid (e.g. a deleted fuzz target).

`errors.BuildNotFound` can be raised in `build_manager.build_setup()`, but it isn't handled properly. This was previously handled here prior to the utask refactor: https://github.com/google/clusterfuzz/blob/0149b701bf4143451ba3fc5251d76ceb121bbe93/src/clusterfuzz/_internal/bot/tasks/progression_task.py#L458

Context: https://github.com/google/clusterfuzz/pull/2490


alhijazi (approved):
LGTM, thanks!
You probably need to rebase since test failures seem unrelated to your change.

PR #4129, opened 2024-07-30
Added 'None' check for testcase_id in issue_redirector.py

Onkar1300:
Redirect the page to 'Testcase not found' instead of reporting error log when accessing issue with 'None' testcase_id

### DIFF
--- a/src/appengine/handlers/issue_redirector.py
+++ b/src/appengine/handlers/issue_redirector.py
@@ -24,7 +24,9 @@ class Handler(base_handler.Handler):
 
   def get(self, testcase_id=None):
     """Redirect user to the correct URL."""
-    testcase = helpers.get_testcase(testcase_id)
+    testcase = None
+    if testcase_id:
+      testcase = helpers.get_testcase(testcase_id)
     issue_url = helpers.get_or_exit(
         lambda: issue_tracker_utils.get_issue_url(testcase),
         'Issue tracker for testcase (id=%s) is not found.' % testcase_id,


jonathanmetzman:
Not sure this is a worthwhile change. Why do you want it?

Onkar1300:
> Not sure this is a worthwhile change. Why do you want it?

Hi Jonathan,
If https://clusterfuzz.corp.google.com/issue is accesed without any testcase_id, the argument to '[get_testcase()](https://github.com/google/clusterfuzz/blob/8c2f3bcebd061dfdd40470ee2e2ddab6bf6637e1/src/appengine/handlers/issue_redirector.py#L27)' is None which eventually leads to the following error: 
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
To resolve the issue, a None check is added to redirect to None 'testcase_id' doesn't exist. instead of reporting TypeError.

jonathanmetzman:
> > Not sure this is a worthwhile change. Why do you want it?
> 
> Hi Jonathan, If https://clusterfuzz.corp.google.com/issue is accesed without any testcase_id, the argument to '[get_testcase()](https://github.com/google/clusterfuzz/blob/8c2f3bcebd061dfdd40470ee2e2ddab6bf6637e1/src/appengine/handlers/issue_redirector.py#L27)' is None which eventually leads to the following error: TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType' To resolve the issue, a None check is added to redirect to None 'testcase_id' doesn't exist. instead of reporting TypeError.

When are you directed to this URL though?

Onkar1300:
> > > Not sure this is a worthwhile change. Why do you want it?
> > 
> > 
> > Hi Jonathan, If https://clusterfuzz.corp.google.com/issue is accesed without any testcase_id, the argument to '[get_testcase()](https://github.com/google/clusterfuzz/blob/8c2f3bcebd061dfdd40470ee2e2ddab6bf6637e1/src/appengine/handlers/issue_redirector.py#L27)' is None which eventually leads to the following error: TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType' To resolve the issue, a None check is added to redirect to None 'testcase_id' doesn't exist. instead of reporting TypeError.
> 
> When are you directed to this URL though?

Whenever web-security-scanner@google.com tries to access https://clusterfuzz.corp.google.com/issue, it triggers this error log in ClusterFuzz logs

jonathanmetzman:
Why is this bot trying to access that URL? Can the bot be fixed?

Onkar1300:
> Why is this bot trying to access that URL? Can the bot be fixed?

it seems like for security checks, [web-security-scanner@google.com](mailto:web-security-scanner@google.com) tries to access https://clusterfuzz.corp.google.com/issue which raises the above error. We are not aware how to fix the bot such that it does not access https://clusterfuzz.corp.google.com/issue.

jonathanmetzman:
OK, then I don't think we should make this change. We shouldn't special case things so a security scanner doesn't get a 500.

PR #4130, opened 2024-07-31
Added Debugging Logs:

aditya-wazir:
Logs have been added to debug and resolve the following issues:
1. AttributeError: 'NoneType' object has no attribute 'stdin'
2. Line is too long for a stacktrace

### DIFF
--- a/src/clusterfuzz/_internal/system/environment.py
+++ b/src/clusterfuzz/_internal/system/environment.py
@@ -356,7 +356,7 @@ def get_llvm_symbolizer_path():
         stderr=subprocess.DEVNULL)
     if return_code == 0:
       # llvm-symbolize works, return it.
-      logs.info(f'Debug: Environment Symbolizer Path :- {llvm_symbolizer_path}')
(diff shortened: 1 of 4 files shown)


src/clusterfuzz/_internal/system/environment.py
```diff
@@ -365,10 +367,12 @@ def get_llvm_symbolizer_path():
 
   # Make sure that we have a default llvm-symbolizer for this platform.
   if not os.path.exists(llvm_symbolizer_path):
+    logs.info(f'Debug: None Symbolizer Path :- {llvm_symbolizer_path}')
```
jonathanmetzman, line 370: Maybe we should make a debug log function. I don't like writing "debug" in an info log function.

aditya-wazir: Hi Jonathan, 
These logs are for investigating the root cause of ```AttributeError: 'NoneType' object has no attribute 'stdin'``` issue. Once we debug this issue, we will remove the logs.

PR #4132, opened 2024-07-31
Rename `environment.get_value_string()` to `get_value_raw()`.

letitz:
This is a prelude to fixing type errors in `environment.py`.

One of the big sources of type errors so far has been the fact that `environment.get_value()` is ~untyped. Many places in the codebase assume that it returns e.g. a string, or an int. We can introduce `get_value_int()` for example for ints that checks at runtime that the environment variable evaluated to an int, and use that instead where applicable to appease the
type checker (and surface errors earlier at runtime).

There is a need for a `get_value_string()` that tries to evaluate environment variable contents, i.e. supports quoted values. Many variables are expected to be strings, yet the pervasive use of `environment.get_value()` and `environment.set_value()` means it is hard to tell whether they are sometimes quoted or not.

This PR renames the current `get_value_string()` function to make space for a future `get_value_string()` that evaluates variables.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
@@ -1172,11 +1172,10 @@ def process_crashes(crashes: List[Crash], context: Context,
 
 def _get_issue_metadata_from_environment(variable_name):
   """Get issue metadata from environment."""
-  values = str(environment.get_value_string(variable_name, '')).split(',')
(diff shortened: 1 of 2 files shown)


jonathanmetzman (approved):
lgtm
Happy to land this when you are.

letitz:
Thanks for the review! I've rebased and made the type annotation more precise, to avoid errors like:

```py
foo = environment.get_value_raw('foo', '')
bar = foo.split(',')   # type checker complains `foo` might be None
```

PR #4154, opened 2024-08-06
Add Tunk Stable versions to deprecated testcase check

svasudevprasad:
Updated can_testcase_run_on_platform to check for
Android Trunk Stable versions - 'Z', 'A' and 'B'

### DIFF
--- a/src/clusterfuzz/_internal/platforms/android/util.py
+++ b/src/clusterfuzz/_internal/platforms/android/util.py
@@ -81,4 +81,11 @@ def can_testcase_run_on_platform(testcase_platform_id, current_platform_id):
   if current_platform_id_fields[2] == 'm':
     return True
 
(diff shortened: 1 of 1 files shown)


jonathanmetzman (approved):
lgtm

jonathanmetzman:
Can you explain why this is needed?
And please do this in the future.

oliverchang:
> Can you explain why this is needed? And please do this in the future.

+1 ! Please see https://google.github.io/eng-practices/review/developer/cl-descriptions.html

svasudevprasad:
This change is required so that deprecated test cases run on devices with the latest Android version. Currently only "m" or "main" was considered as the latest version. 

Trunk stable based releases should also be considered as the latest version. Hence this change adds the trunk stable based versions - Z, A, B to the list of versions considered as "latest"

PR #4155, opened 2024-08-06
Allow restricting issue access using job's AUTOMATIC_LABELS

alhijazi:


alhijazi:
This is related to: https://issues.chromium.org/issues/357539001

src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
```diff
@@ -697,6 +707,15 @@ def save(self, new_comment=None, notify=True):
       access_limit = self._issue_access_limit
       if access_limit:
         self._data['issueState']['accessLimit'] = {'accessLevel': access_limit}
+      print(list(self.labels))
```
oliverchang, line 710: remove print?

src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
```diff
@@ -683,6 +691,8 @@ def save(self, new_comment=None, notify=True):
       self._set_severity(severity)
       self.labels.remove_by_prefix(_SEVERITY_LABEL_PREFIX)
 
+      # Check for an access_limit label before filtering
+      limit_view_label = _extract_label(self.labels, 'LIMIT_')
```
oliverchang, line 695: Should we combine `_extract_label` and `_get_access_limit_for_label` into a single function? i.e. `_extract_access_limit_from_labels(self.labels)` ?

This consolidates the logic a bit.

src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
```diff
@@ -48,6 +48,14 @@ class IssueAccessLevel(str, enum.Enum):
   LIMIT_VIEW_TRUSTED = 'LIMIT_VIEW_TRUSTED'
 
 
+def _get_access_limit_for_label(label: str):
+  """Returns IssueAccessLevel from string lable or None."""
+  try:
+    return IssueAccessLevel('LIMIT_' + label)
+  except:
+    return None
```
oliverchang, line 56: should we log a warning/error in these cases?

src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
```diff
@@ -697,6 +707,15 @@ def save(self, new_comment=None, notify=True):
       access_limit = self._issue_access_limit
       if access_limit:
         self._data['issueState']['accessLimit'] = {'accessLevel': access_limit}
+      print(list(self.labels))
+      if access_limit == IssueAccessLevel.LIMIT_NONE:
+        # Try to set accessLimit from labels if any.
+        access_limit = _get_access_limit_for_label(limit_view_label)
```
oliverchang, line 713: hmm, should we always be overriding the access limit when it's specified in a job? it seems a bit confusing to me that this only applies when the default access level is limit_none.

src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
alhijazi: Done.

src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
alhijazi: Done.

src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
alhijazi: I am not sure what the previous behaviour was. I was worried this would allow to open up some restricted issues by mistake. This is why I decided only to set the access level from labels if it's set to None.

alhijazi: Alternatively, we could set the custom access level from labels only if it is greater or equal to the current access level limit. WDYT?
Otherwise, we could just override the value always, but we would risk opening up some restricted issues.

oliverchang: Thanks for the explanation! What you have seems good to me. We can adjust this later if this ever becomes a problem.

PR #4168, opened 2024-08-15
Fix AttributeError: 'NoneType' object has no attribute 'corpus_pruning_task_input'

jonathanmetzman:
This was caused by the roll forward of oss-fuzz.


oliverchang (commented):
LGTM, thanks!

PR #4169, opened 2024-08-15
Fix deadline in manage_vms

jonathanmetzman:
Fix problem in oss-fuzz where manage_vms doesn't complete, causing old VMs/containers to hang around.
Do this by increasing the deadline to 24 hours for manage_vms to compelte. Also, temporarily change the order of manage_vms so that the previously unresized groups will be resized first. This should hopefully get rid of weird errors we are seeing.

### DIFF
--- a/infra/k8s/manage-vms.yaml
+++ b/infra/k8s/manage-vms.yaml
@@ -21,7 +21,7 @@ spec:
   concurrencyPolicy: Forbid
   jobTemplate:
     spec:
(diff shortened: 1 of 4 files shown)


vitorguidi (approved):
LGTM

However, gotta deploy this from the [3.7 branch](https://github.com/google/clusterfuzz/pull/4163), right?

oliverchang (commented):
lgtm

PR #4186, opened 2024-08-19
Update install_deps_linux.bash to install google-cloud-cli instead.

oliverchang:
This was renamed to google-cloud-cli. Referencing google-cloud-sdk may even cause old/stale versions to be installed in some setups.


jonathanmetzman (approved):
lgtm

PR #4192, opened 2024-08-19
[fuzz_task] Get fuzz target weights in preprocess

jonathanmetzman:
Move fuzzer selection to preprocess because we need to access the db to do it now.
https://github.com/google/clusterfuzz/issues/3008


jonathanmetzman:
This is not ready.

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -99,12 +99,6 @@ class FuzzErrorCode:
     ['success', 'testcase_file_paths', 'fuzzer_metadata'])
 
 
-def has_standard_build():
-  if environment.platform() == 'FUCHSIA':
```
vitorguidi, line 103: Why is this being removed and not showing up somewhere else? Is fuchsia deprecated?

PR #4194, opened 2024-08-20
LogError: Testcase comments truncated

Onkar1300:
Reduce error logging noise for the above issue by using 'logs.info' instead of 'logs.error'


oliverchang (approved):

jonathanmetzman:
Hmmm...on the one hand this is an error that fires a lot, but it usually highlights a real issue.

PR #4199, opened 2024-08-21
Remove unneeded use of FUZZ_TARGET

jonathanmetzman:


jonathanmetzman:
I'm going to try to get rid of this env var.

PR #4211, opened 2024-08-27
Support OSS-Fuzz custom fields for google issue tracker.

oliverchang:


oliverchang:
@rmistry can you please check if we're using the issue tracker correctly here with the new custom fields we're supporting here?

src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
```diff
@@ -31,11 +32,15 @@
 _NUM_RETRIES = 3
 _ISSUE_TRACKER_URL = 'https://issues.chromium.org/issues'
 
+# TODO: Make these configuration settings instead of hardcoded values in code.
 # These custom fields use repeated enums.
 _CHROMIUM_OS_CUSTOM_FIELD_ID = '1223084'
```
rmistry, line 37: As you are likely aware but making it explicit here- If there are any assumptions about these custom fields existing in the code then they will fail in the non-Chromium tracker. Doing a quick check it should be OK, but mentioning incase you see errors about that later.

src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
```diff
@@ -559,6 +570,32 @@ def _update_issue(self, new_comment=None, notify=True):
...
 
+    # Special case: OSS-Fuzz "Reported" custom field.
+    added_reported = _get_labels(self.labels.added, 'Reported-')
+    if added_reported:
+      try:
+        # Assume there is only one entry.
+        year, month, day = _parse_date_label(added_reported[0])
+        custom_field_entries.append({
+            'customFieldId': _OSS_FUZZ_REPORTED_CUSTOM_FIELD_ID,
+            'dateValue': {
```
rmistry, line 581: Just curious- Where did you find that this was the way to update a "Date" customfield?

rmistry: Looks correct to me according to this example- https://yaqs.corp.google.com/eng/q/4249097869912440832

rmistry:
> @rmistry can you please check if we're using the issue tracker correctly here with the new custom fields we're supporting here?

LGTM (I did not notice any obvious issues)

src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
oliverchang: Ack yep -- there aren't overlaps currently between Chromium and OSS-Fuzz.

src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
oliverchang: Thanks! I was basing this off the service protobuf definitions, but I wasn't sure how the JSON transcoding worked.

PR #4221, opened 2024-09-10
Fix tests-syncer

jonathanmetzman:
Fixes: https://g-issues.chromium.org/issues/365801496

### DIFF
--- a/docker/chromium/tests-syncer/Dockerfile
+++ b/docker/chromium/tests-syncer/Dockerfile
@@ -30,8 +30,9 @@ RUN apt-get update && \
 RUN apt-get update && \
     apt-get install -y \
         git \
+	python-is-python3 \
         subversion \
-        zip
+        zip	
 
 RUN git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git \
     /data/depot_tools
--- a/docker/chromium/tests-syncer/Dockerfile
+++ b/docker/chromium/tests-syncer/Dockerfile
@@ -32,7 +32,7 @@ RUN apt-get update && \
         git \
 	python-is-python3 \
         subversion \
-        zip	
+        zip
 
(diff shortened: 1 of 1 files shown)


docker/chromium/tests-syncer/Dockerfile
```diff
@@ -30,8 +30,9 @@ RUN apt-get update && \
 RUN apt-get update && \
```
jonathanmetzman, line 30: I think it's super confusing this is FROM gcr.io/clusterfuzz-images/base and not FROM gcr.io/clusterfuzz-images/chromium/base

docker/chromium/tests-syncer/Dockerfile
```diff
@@ -30,8 +30,9 @@ RUN apt-get update && \
...
     apt-get install -y \
         git \
+	python-is-python3 \
         subversion \
-        zip
+        zip
```
vitorguidi, line 35: Nit: spaces

jonathanmetzman:
cc @mi-ac

docker/chromium/tests-syncer/Dockerfile
```diff
@@ -30,6 +30,7 @@ RUN apt-get update && \
 RUN apt-get update && \
     apt-get install -y \
         git \
+	python-is-python3 \
```
vitorguidi, line 33: What magic does this do, in order to fix the problem?

vitorguidi (approved):
lgtm

docker/chromium/tests-syncer/Dockerfile
jonathanmetzman: there's a chrome script that runs "python". Since python2 was replaced, this doesn't exist on most systems anymore.
Fixing this actually allows the initialization to complete and for syncing to happen instead of being in an init loop.

PR #4225, opened 2024-09-12
Try to fix tests syncer

jonathanmetzman:
The existing setup doesn't terminate. Probably because there's over 157k testcases.

### DIFF
--- a/src/python/other-bots/chromium-tests-syncer/run.py
+++ b/src/python/other-bots/chromium-tests-syncer/run.py
@@ -37,7 +37,7 @@ from clusterfuzz._internal.system import environment
 from clusterfuzz._internal.system import shell
 
 MAX_TESTCASE_DIRECTORY_SIZE = 10 * 1024 * 1024  # in bytes.
-MAX_TESTCASES = 10000
+MAX_TESTCASES = 500
 STORED_TESTCASES_LIST = []
 
 # pylint: disable=broad-exception-raised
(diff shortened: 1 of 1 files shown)


src/python/other-bots/chromium-tests-syncer/run.py
```diff
@@ -37,14 +37,20 @@
...
 
 
 def unpack_crash_testcases(crash_testcases_directory):
   """Unpacks the old crash testcases in the provided directory."""
+  count = 0
   for testcase in ndb_utils.get_all_from_model(data_types.Testcase):
+    count += 1
+    if count >= MAX_TESTCASES:
+      logs.info(f'{MAX_TESTCASES} testcases reached.')
+      break
```
vitorguidi, line 53: What is the idea here?  This script will run several times and consume the old test cases slowly, or is this a hard cap untill some sort of corpus pruning task runs?

vitorguidi (approved):
lgtm

src/python/other-bots/chromium-tests-syncer/run.py
jonathanmetzman: Hard cap.

jonathanmetzman: There's never any pruning for here. This source has basically been polluted over the years and is unusable now.

jonathanmetzman: Maybe I will manually prune it, though I don't know how much it will move the needle.

mi-ac: It'd be great if we could get the latest 500 though...

PR #4231, opened 2024-09-13
Use fast clone for Chromium sources

mi-ac:
The old code used unnecessary gclient routines that clone the repository with dependencies. For web tests the dependencies are unnecessary. The new approach takes under 3 minutes compared to over 2 hours.

### DIFF
--- a/src/python/other-bots/chromium-tests-syncer/run.py
+++ b/src/python/other-bots/chromium-tests-syncer/run.py
@@ -235,20 +235,8 @@ def main():
   shell.create_directory(crash_testcases_directory)
   unpack_crash_testcases(crash_testcases_directory)
 
-  # Sync web tests.
-  logs.info('Syncing web tests.')
-  src_directory = os.path.join(tests_directory, 'src')
-  gclient_file_path = os.path.join(tests_directory, '.gclient')
-  if not os.path.exists(gclient_file_path):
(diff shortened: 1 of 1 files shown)


src/python/other-bots/chromium-tests-syncer/run.py
```diff
@@ -235,20 +235,8 @@ def main():
   shell.create_directory(crash_testcases_directory)
   unpack_crash_testcases(crash_testcases_directory)
 
-  # Sync web tests.
-  logs.info('Syncing web tests.')
```
letitz, line 239: It was nice to have a log statement for debugging. Does `clone_git_repository()` do its own logging?

letitz (commented):
I don't suppose there are any tests for this code?

src/python/other-bots/chromium-tests-syncer/run.py
mi-ac: Yes it does.

mi-ac:
This never had any tests.

src/python/other-bots/chromium-tests-syncer/run.py
letitz: Thanks for confirming.

letitz (approved):
LGTM

PR #4232, opened 2024-09-13
Switch old svn repository entries to git

mi-ac:
The svn repositories don't exist anymore. The tree entries can all be found in the WebKit git repository.


letitz (approved):
LGTM

Issue #4238, opened 2024-09-15
Investigate database contention possibly breaking much of OSS-Fuzz


jonathanmetzman:
Logs show a lot of database contention. And oss-fuzz is in an unhealthy state right now.
My theory is that contention caused by widespread failures (caused by ) is breaking things in the rest of oss-fuzz.
https://github.com/google/oss-fuzz/issues/12417 is for sure caused by contention.

jonathanmetzman:
Some other issues, though possibly caused by the big upgrade.
https://github.com/google/oss-fuzz/issues/12450#issuecomment-2338286927
https://github.com/google/oss-fuzz/issues/12446

jonathanmetzman:
I've also removed builds/projects that are exceptioning and looping: https://github.com/google/oss-fuzz/pull/12501
https://github.com/google/oss-fuzz/pull/12500
https://github.com/google/oss-fuzz/pull/12499
https://github.com/google/oss-fuzz/pull/12498

jonathanmetzman:
Fixed

PR #4239, opened 2024-09-15
Skip the fuzzer output junk for engine fuzzers.

jonathanmetzman:
It's intended for blackbox fuzzers, it's not needed at all (should probably be deprecated though some chrome folks use it), and seems to be having a hand in the contention breaking parts of oss-fuzz
Mitigates https://github.com/google/clusterfuzz/issues/4238


src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -708,6 +708,8 @@ def store_fuzzer_run_results(testcase_file_paths, fuzzer, fuzzer_command,
                              generated_testcase_string, fuzz_task_input):
   """Store fuzzer run results in database."""
   # Upload fuzzer script output to bucket.
+  if not environment.is_engine_fuzzer_job():
```
oliverchang, line 711: Can you add comments to all of these for context?

oliverchang (commented):
lgtm

PR #4244, opened 2024-09-15
Install Java 17 on images

jonathanmetzman:


jonathanmetzman:
THis doesn't work. We need java_15

vitorguidi:
> THis doesn't work. We need java_15

Why? What broke?

jonathanmetzman:
> > THis doesn't work. We need java_15
> 
> Why? What broke?

JNI

jonathanmetzman:
Confirmed this works with both versions of Java.

oliverchang (approved):

PR #4245, opened 2024-09-16
Docker base: Don't build Python with -j.

oliverchang:
This seems to lead to linker issues with gcov.

Likely upstream bug: https://bugs.python.org/issue29365


jonathanmetzman:
lgtm. Sorry about that.

PR #4246, opened 2024-09-16
Make docker build.sh accept an argument again.

oliverchang:
Otherwise building this on Cloud Build no longer works because there is no git metadata there:

https://github.com/google/clusterfuzz/blob/master/docker/cloudbuild.yaml


vitorguidi (approved):
lgtm

PR #4247, opened 2024-09-16
Fix OS custom field updates in Google Issue Tracker

letitz:
Skip empty OS custom fields in API calls to Google Issue Tracker, since those are rejected.
Also deduplicate values while we're at it.

The continuing duplication of code between new issues and updated issues makes me sad,
but this PR does not improve anything on that front.

Fixes: https://crbug.com/366955327
Bug: https://crbug.com/367205692


jonathanmetzman (approved):
lgtm

PR #4248, opened 2024-09-17
Populate platform and bot_name after all utask_main runs.

letitz:
ClusterFuzz has been writing crashes to BigQuery with empty `platform` columns,
because it takes that value from `uworker_msg.Output.platform_id`, which AFAICT
is only populated when `fuzz_task` is run on a remote instance.

Bug: https://crbug.com/367205692


jonathanmetzman (approved):
lgtm

PR #4254, opened 2024-09-18
remove the unnecessary free disk space limit

paulsemel:
With Chromium archive sizes and extracted size being bigger, we cannot afford the 10GB surplus for the build directory.


jonathanmetzman:
@oliverchang Do you know why we have these limits at all?

paulsemel:
> @oliverchang Do you know why we have these limits at all?

I decided to completely remove the limit. Let's wait for @oliverchang  on why this was used, and I'll set this back if needed.

PR #4260, opened 2024-09-19
google_issue_tracker: support configurable URLs.

oliverchang:
OSS-Fuzz needs something different to avoid an unnecessary redirect.


jonathanmetzman (approved):
lgtm

PR #4262, opened 2024-09-20
Update tests-syncer caps and logging

mi-ac:
- This reduces the crashtest cap from 250000 to 25000.
- Add more logging to better see how far we get.
- Exclude also centipede and googlefuzztest cases similar to libfuzzer.

Context: http://b/365801496

### DIFF
--- a/src/python/other-bots/chromium-tests-syncer/run.py
+++ b/src/python/other-bots/chromium-tests-syncer/run.py
@@ -36,8 +36,10 @@ from clusterfuzz._internal.system import archive
 from clusterfuzz._internal.system import environment
 from clusterfuzz._internal.system import shell
 
+ENGINE_FUZZER_NAMES = ['afl', 'centipede', 'googlefuzztest', 'libFuzzer']
 MAX_TESTCASE_DIRECTORY_SIZE = 10 * 1024 * 1024  # in bytes.
-MAX_TESTCASES = 250000
+MAX_TESTCASES = 25000
+TESTCASES_REPORT_INTERVAL = 2500
 STORED_TESTCASES_LIST = []
 
 # pylint: disable=broad-exception-raised
@@ -51,6 +53,9 @@ def unpack_crash_testcases(crash_testcases_directory):
     if count >= MAX_TESTCASES:
       logs.info(f'{MAX_TESTCASES} testcases reached.')
       break
+    if count % TESTCASES_REPORT_INTERVAL == 0:
+      logs.info(f'Processed {count} testcases.')
+
     testcase_id = testcase.key.id()
 
     # 1. If we have already stored the testcase, then just skip.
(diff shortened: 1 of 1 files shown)


src/python/other-bots/chromium-tests-syncer/run.py
```diff
@@ -85,7 +89,9 @@ def unpack_crash_testcases(crash_testcases_directory):
 
     # 8. Skip in-process fuzzer testcases, since these are only applicable to
     # fuzz targets and don't run with blackbox binaries.
-    if testcase.fuzzer_name and testcase.fuzzer_name in ['afl', 'libFuzzer']:
+    if testcase.fuzzer_name and testcase.fuzzer_name in [
```
letitz, line 92: We can definitely push the filtering down to datastore and avoid iterating through obviously-useless entries.

You'll want to use `Testcase.query`, like: https://github.com/google/clusterfuzz/blob/ad6131e9d61cdbe7838d26b55d82200faeddf84f/src/local/butler/scripts/backfiler.py#L31-L35

src/python/other-bots/chromium-tests-syncer/run.py
```diff
@@ -51,6 +52,9 @@ def unpack_crash_testcases(crash_testcases_directory):
     if count >= MAX_TESTCASES:
       logs.info(f'{MAX_TESTCASES} testcases reached.')
       break
+    if count % TESTCASES_REPORT_INTERVAL == 0:
```
letitz, line 55: Can also do it based on time: every minute, output progress.

src/python/other-bots/chromium-tests-syncer/run.py
```diff
@@ -85,7 +89,9 @@ def unpack_crash_testcases(crash_testcases_directory):
 
     # 8. Skip in-process fuzzer testcases, since these are only applicable to
     # fuzz targets and don't run with blackbox binaries.
-    if testcase.fuzzer_name and testcase.fuzzer_name in ['afl', 'libFuzzer']:
+    if testcase.fuzzer_name and testcase.fuzzer_name in [
+        'afl', 'centipede', 'googlefuzztest', 'libFuzzer'
```
letitz, line 93: Maybe define a constant for this at the top, e.g. `ENGINE_FUZZER_NAMES`.

letitz (approved):
LGTM with a couple suggestions.

src/python/other-bots/chromium-tests-syncer/run.py
mi-ac: Done.

src/python/other-bots/chromium-tests-syncer/run.py
mi-ac: I'd leave that as a follow-up, in case we need it at all. Maybe a better query will reduce this iteration enough so that we don't need this.

src/python/other-bots/chromium-tests-syncer/run.py
mi-ac: That'd be great. Will try in a follow up.

src/python/other-bots/chromium-tests-syncer/run.py
```diff
@@ -36,8 +36,10 @@
 from clusterfuzz._internal.system import environment
 from clusterfuzz._internal.system import shell
 
+ENGINE_FUZZER_NAMES = ['afl', 'centipede', 'googlefuzztest', 'libFuzzer']
```
jonathanmetzman, line 39: Can you get the list from here: https://github.com/google/clusterfuzz/blob/ad6131e9d61cdbe7838d26b55d82200faeddf84f/src/clusterfuzz/_internal/cron/fuzzer_and_job_weights.py#L54

jonathanmetzman (approved):

PR #4263, opened 2024-09-20
[build_manager] add support for remote zip

paulsemel:
This adds support for remote ZIP.

As of now, performances are quite good locally, and the read ahead mechanism should keep reasonable performance. Also, given that the ClusterFuzz bots are having HDD, numbers might even be better there, as we're only storing on disk when unpacking the build.

The memory consumption of this new feature is contant: it uses at most (and most of the time) 50 MB of RAM.

### DIFF
--- a/src/clusterfuzz/_internal/bot/fuzzers/utils.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/utils.py
@@ -13,10 +13,13 @@
 # limitations under the License.
 """Fuzzer utils."""
 
+import functools
 import os
 import re
 import stat
 import tempfile
+from typing import Callable
+from typing import Optional
 
 from clusterfuzz._internal.base import utils
 from clusterfuzz._internal.metrics import logs
@@ -30,7 +33,7 @@ BLOCKLISTED_TARGET_NAME_REGEX = re.compile(r'^(jazzer_driver.*)$')
 EXTRA_BUILD_DIR = '__extra_build'
 
 
-def is_fuzz_target_local(file_path, file_handle=None):
+def is_fuzz_target(file_path, file_opener: Optional[Callable] = None):
   """Returns whether |file_path| is a fuzz target binary (local path)."""
   if '@' in file_path:
     # GFT targets often have periods in the name that get misinterpreted as an
@@ -53,7 +56,7 @@ def is_fuzz_target_local(file_path, file_handle=None):
     # Ignore files with disallowed extensions (to prevent opening e.g. .zips).
     return False
 
-  if not file_handle and not os.path.exists(file_path):
+  if not file_opener and not os.path.exists(file_path):
     # Ignore non-existent files for cases when we don't have a file handle.
     return False
 
@@ -72,24 +75,27 @@ def is_fuzz_target_local(file_path, file_handle=None):
     logs.warning('Tried to read from non-regular file: %s.' % file_path)
     return False
 
-  # Use already provided file handle or open the file.
-  local_file_handle = file_handle or open(file_path, 'rb')
+  # Either use the file opener or open the file ourselves.
+  if not file_opener:
+    file_opener = functools.partial(open, mode='rb')
+  try:
+    with file_opener(file_path) as file_handle:
+      result = False
+      for pattern in FUZZ_TARGET_SEARCH_BYTES:
+        # TODO(metzman): Bound this call so we don't read forever if something
+        # went wrong.
+        file_handle.seek(0)
+        result = utils.search_bytes_in_file(pattern, file_handle)
+        if result:
+          break
 
-  result = False
-  for pattern in FUZZ_TARGET_SEARCH_BYTES:
-    # TODO(metzman): Bound this call so we don't read forever if something went
-    # wrong.
-    local_file_handle.seek(0)
-    result = utils.search_bytes_in_file(pattern, local_file_handle)
-    if result:
-      break
+      file_handle.close()
 
-  if not file_handle:
-    # If this local file handle is owned by our function, close it now.
-    # Otherwise, it is caller's responsibility.
-    local_file_handle.close()
-
-  return result
+      return result
+  except Exception as e:
+    # In case we could not open the file, we consider it's not a fuzzer.
+    logs.warning(f'Could not open {file_path}: {e}')
+    return False
 
 
 def get_fuzz_targets_local(path):
@@ -103,7 +109,7 @@ def get_fuzz_targets_local(path):
         continue
 
       file_path = os.path.join(root, filename)
-      if is_fuzz_target_local(file_path):
+      if is_fuzz_target(file_path):
         fuzz_target_paths.append(file_path)
 
   return fuzz_target_paths
--- a/src/clusterfuzz/_internal/build_management/build_archive.py
+++ b/src/clusterfuzz/_internal/build_management/build_archive.py
@@ -190,13 +190,9 @@ class DefaultBuildArchive(BuildArchive):
     from clusterfuzz._internal.bot.fuzzers import utils as fuzzer_utils
 
     for archive_file in self.list_members():
-      file_content = self.try_open(archive_file.name)
-      if fuzzer_utils.is_fuzz_target_local(archive_file.name, file_content):
+      if fuzzer_utils.is_fuzz_target(archive_file.name, self.open):
         fuzz_target = fuzzer_utils.normalize_target_name(archive_file.name)
         self._fuzz_targets[fuzz_target] = archive_file.name
-
-      if file_content:
-        file_content.close()
 
     return list(self._fuzz_targets.keys())
 
@@ -299,23 +295,19 @@ class ChromeBuildArchive(DefaultBuildArchive):
     return res
 
 
-# pylint: disable=redefined-builtin
-def open(archive_path: str) -> BuildArchive:
-  """Opens the archive and gets the appropriate build archive based on the
-  `archive_path`. The resulting object is usable as a normal archive reader,
-  but provides additional feature related to build handling.
+def open_with_reader(reader: archive.ArchiveReader) -> BuildArchive:
+  """Open the archive and gets the appropriate build archive based on the
+  provided archive information.
 
   Args:
-      archive_path: the path to the archive.
+      reader: the archive reader.
 
   Raises:
-      If the file could not be opened or if the archive type cannot be handled.
+    If the archive reader cannot be handled.
 
   Returns:
-      the build archive.
+      The build archive.
   """
-  reader = archive.open(archive_path)
-
   # Unfortunately, there is no good heuristic for determining which build
   # archive implementation to use.
   # Hopefully, we can search in the archive whether some files are present and
@@ -328,3 +320,41 @@ def open(archive_path: str) -> BuildArchive:
   if reader.file_exists(args_gn_path):
     return ChromeBuildArchive(reader)
   return DefaultBuildArchive(reader)
+
+
+def open(archive_path: str) -> BuildArchive:  # pylint: disable=redefined-builtin
+  """Opens the archive and gets the appropriate build archive based on the
+  `archive_path`. The resulting object is usable as a normal archive reader,
+  but provides additional feature related to build handling.
+
+  Args:
+      archive_path: the path to the archive.
+
+  Raises:
+      If the file could not be opened or if the archive type cannot be handled.
+
+  Returns:
+      The build archive.
+  """
+  reader = archive.open(archive_path)
+  return open_with_reader(reader)
+
+
+def open_uri(uri: str) -> BuildArchive:
+  """Opens a build archive over HTTP. This is only compatible with chromium as
+  of now.
+
+  Args:
+      uri: the URI pointing to the zip file.
+
+  Returns:
+      The build archive.
+  """
+  reader = archive.ZipArchiveReader(archive.HttpZipFile(uri))
+  return open_with_reader(reader)
+
+
+def unzip_over_http_compatible(build_url: str) -> bool:
+  """Whether the build URL is compatible with unzipping over HTTP.
+  """
+  return archive.HttpZipFile.is_uri_compatible(build_url)
--- a/src/clusterfuzz/_internal/build_management/build_manager.py
+++ b/src/clusterfuzz/_internal/build_management/build_manager.py
@@ -14,11 +14,13 @@
 """Build manager."""
 
 from collections import namedtuple
+import contextlib
 import os
 import re
 import shutil
 import subprocess
 import time
+from typing import Optional
 
 from clusterfuzz._internal.base import errors
 from clusterfuzz._internal.base import utils
@@ -402,10 +404,92 @@ class Build(BaseBuild):
     if instrumented_library_paths:
       self._patch_rpaths(instrumented_library_paths)
 
-  def _unpack_build(self, base_build_dir, build_dir, build_url):
+  @contextlib.contextmanager
+  def _download_and_open_build_archive(self, base_build_dir: str,
+                                       build_dir: str, build_url: str):
+    """Downloads the build archive at `build_url` and opens it.
+
+    Args:
+        base_build_dir: the base build directory
+        build_dir: the current build directory
+        build_url: the build URL
+
+    Yields:
+        the build archive
+    """
+    # Download build archive locally.
+    build_local_archive = os.path.join(build_dir, os.path.basename(build_url))
+
+    # Make the disk space necessary for the archive available.
+    archive_size = storage.get_object_size(build_url)
+    if archive_size is not None and not _make_space(archive_size,
+                                                    base_build_dir):
+      shell.clear_data_directories()
+      logs.log_fatal_and_exit(
+          'Failed to make space for download. '
+          'Cleared all data directories to free up space, exiting.')
+
+    logs.info(f'Downloading build from {build_url} to {build_local_archive}.')
+    try:
+      storage.copy_file_from(build_url, build_local_archive)
+    except Exception as e:
+      logs.error(f'Unable to download build from {build_url}: {e}')
+      raise
+
+    try:
+      with build_archive.open(build_local_archive) as build:
+        yield build
+    finally:
+      shell.remove_file(build_local_archive)
+
+  def _open_build_archive(self, base_build_dir: str, build_dir: str,
+                          build_url: str, http_build_url: Optional[str],
+                          unpack_everything: Optional[bool]):
+    """Gets a handle on a build archive for the current build. Depending on the
+    provided parameters, this function might download the build archive into
+    the build directory or directly use remote HTTP archive.
+
+    Args:
+        unpack_everything: wether we should unpack the whole archive or try
+        selective unpacking.
+        base_build_dir: the base build directory.
+        build_dir: the current build directory.
+        build_url: the build URL.
+        http_build_url: the HTTP build URL.
+
+    Raises:
+        if an error occurred while accessing the file over HTTP or while
+        downloading the file on disk.
+
+    Returns:
+        the build archive.
+    """
+    # We only want to use remote unzipping if we're not unpacking everything and
+    # if the HTTP URL is compatible with remote unzipping.
+    allow_unpack_over_http = environment.get_value(
+        'ALLOW_UNPACK_OVER_HTTP', default_value=False)
+    can_unzip_over_http = (
+        allow_unpack_over_http and not unpack_everything and http_build_url and
+        build_archive.unzip_over_http_compatible(http_build_url))
+
+    if not can_unzip_over_http:
+      return self._download_and_open_build_archive(base_build_dir, build_dir,
+                                                   build_url)
+    logs.info("Opening an archive over HTTP, skipping archive download.")
+    assert http_build_url
+    return build_archive.open_uri(http_build_url)
+
+  def _unpack_build(self,
+                    base_build_dir,
+                    build_dir,
+                    build_url,
+                    http_build_url=None):
     """Unpacks a build from a build url into the build directory."""
     # Track time taken to unpack builds so that it doesn't silently regress.
     start_time = time.time()
+
+    unpack_everything = environment.get_value(
+        'UNPACK_ALL_FUZZ_TARGETS_AND_FILES')
 
     logs.info(f'Unpacking build from {build_url} into {build_dir}.')
 
@@ -419,27 +503,9 @@ class Build(BaseBuild):
       _handle_unrecoverable_error_on_windows()
       return False
 
-    # Download build archive locally.
-    build_local_archive = os.path.join(build_dir, os.path.basename(build_url))
-
-    # Make the disk space necessary for the archive available.
-    archive_size = storage.get_object_size(build_url)
-    if archive_size is not None and not _make_space(archive_size,
-                                                    base_build_dir):
-      shell.clear_data_directories()
-      logs.log_fatal_and_exit(
-          'Failed to make space for download. '
-          'Cleared all data directories to free up space, exiting.')
-
-    logs.info(f'Downloading build from {build_url}.')
     try:
-      storage.copy_file_from(build_url, build_local_archive)
-    except Exception as e:
-      logs.error(f'Unable to download build from {build_url}: {e}')
-      return False
-
-    try:
-      with build_archive.open(build_local_archive) as build:
+      with self._open_build_archive(base_build_dir, build_dir, build_url,
+                                    http_build_url, unpack_everything) as build:
         unpack_everything = environment.get_value(
             'UNPACK_ALL_FUZZ_TARGETS_AND_FILES')
 
@@ -463,8 +529,7 @@ class Build(BaseBuild):
               'Cleared all data directories to free up space, exiting.')
 
         # Unpack the local build archive.
-        logs.info(
-            f'Unpacking build archive {build_local_archive} to {build_dir}.')
+        logs.info(f'Unpacking build archive {build_url} to {build_dir}.')
         trusted = not utils.is_oss_fuzz()
 
         build.unpack(
@@ -473,7 +538,7 @@ class Build(BaseBuild):
             trusted=trusted)
 
     except Exception as e:
-      logs.error(f'Unable to unpack build archive {build_local_archive}: {e}')
+      logs.error(f'Unable to unpack build archive {build_url}: {e}')
       return False
 
     if unpack_everything:
@@ -483,9 +548,6 @@ class Build(BaseBuild):
       # such so that it is not re-used.
       partial_build_file_path = os.path.join(build_dir, PARTIAL_BUILD_FILE)
       utils.write_data_to_file('', partial_build_file_path)
-
-    # No point in keeping the archive around.
-    shell.remove_file(build_local_archive)
 
     elapsed_time = time.time() - start_time
     elapsed_mins = elapsed_time / 60.
@@ -605,10 +667,20 @@ class RegularBuild(Build):
                revision,
                build_url,
                build_prefix='',
-               fuzz_target=None):
+               fuzz_target=None,
+               http_build_url=None):
+    """RegularBuild constructor. See Build constructor for other parameters.
+
+    Args:
+        http_build_url: the http build URL. E.g.
+        http://storage.com/foo/bar.zip. Defaults to None.
+        build_url: the GCS bucket URL where the build is stored. E.g.
+        gs://foo/bar.zip.
+    """
     super().__init__(
         base_build_dir, revision, build_prefix, fuzz_target=fuzz_target)
     self.build_url = build_url
+    self.http_build_url = http_build_url
 
     if build_prefix:
       self.build_dir_name = build_prefix.lower()
@@ -630,7 +702,7 @@ class RegularBuild(Build):
     build_update = not self.exists()
     if build_update:
       if not self._unpack_build(self.base_build_dir, self.build_dir,
-                                self.build_url):
+                                self.build_url, self.http_build_url):
         return False
 
       logs.info('Retrieved build r%d.' % self.revision)
@@ -1116,6 +1188,9 @@ def setup_regular_build(revision,
 
     return None
 
+  # build_url points to a GCP bucket, and we're only converting it to its HTTP
+  # endpoint so that we can use remote unzipping.
+  http_build_url = build_url.replace('gs://', 'https://storage.googleapis.com/')
   base_build_dir = _base_build_dir(bucket_path)
 
   build_class = RegularBuild
@@ -1133,7 +1208,8 @@ def setup_regular_build(revision,
       revision,
       build_url,
       build_prefix=build_prefix,
-      fuzz_target=fuzz_target)
+      fuzz_target=fuzz_target,
+      http_build_url=http_build_url)
   if build.setup():
     result = build
   else:
--- a/src/clusterfuzz/_internal/system/archive.py
+++ b/src/clusterfuzz/_internal/system/archive.py
@@ -15,6 +15,7 @@
 
 import abc
 import dataclasses
+import io
 import os
 import tarfile
 from typing import BinaryIO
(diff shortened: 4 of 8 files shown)


src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -328,3 +331,34 @@ def open(archive_path: str) -> BuildArchive:
...
+
+  Args:
+      uri: the gs:// storage URI.
+
+  Returns:
+      the build archive.
+  """
+  assert unzip_over_http_compatible(uri)
+  match = CHROMIUM_GS_MATCHER.match(uri)
+  uri = f'https://commondatastorage.googleapis.com/{match.group(1)}'
```
letitz, line 348: I think there might be a helper for this `gs://` to `https://` URL conversion somewhere in CF code already

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -328,3 +331,34 @@ def open(archive_path: str) -> BuildArchive:
...
+  reader = archive.ZipArchiveReader(archive.HttpZipFile(uri))
+  args_gn_path = os.path.join(reader.root_dir(), 'args.gn')
+  if reader.file_exists(args_gn_path):
+    return ChromeBuildArchive(reader)
+  return DefaultBuildArchive(reader)
+
+
+def unzip_over_http_compatible(build_url: str) -> bool:
+  """Whether the build URL is compatible with unzipping over HTTP.
+  As for now, we're only checking for chromium compatible URLs.
```
letitz, line 358: Why?

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -328,3 +331,34 @@ def open(archive_path: str) -> BuildArchive:
...
+  """
+  match = CHROMIUM_GS_MATCHER.match(build_url)
+  if not match:
+    return False
+  uri = f'https://commondatastorage.googleapis.com/{match.group(1)}'
```
letitz, line 363: We're matching against chromium buckets and extracting the URI twice: once in the assertion that calls this function and again immediately after. If we drop the chromium requirement, then we can simply move the `is_uri_compatible()` assertion into `HttpZipFile`'s constructor.

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -190,13 +193,13 @@ def list_fuzz_targets(self) -> List[str]:
...
-      if fuzzer_utils.is_fuzz_target_local(archive_file.name, file_content):
+      # file_content = self.try_open(archive_file.name)
+      if fuzzer_utils.is_fuzz_target_local(archive_file.name, self):
         fuzz_target = fuzzer_utils.normalize_target_name(archive_file.name)
         self._fuzz_targets[fuzz_target] = archive_file.name
 
-      if file_content:
-        file_content.close()
+      # if file_content:
+      #   file_content.close()
```
letitz, line 202: Leftover comments.

src/clusterfuzz/_internal/bot/fuzzers/utils.py
```diff
@@ -30,7 +30,7 @@
 EXTRA_BUILD_DIR = '__extra_build'
 
 
-def is_fuzz_target_local(file_path, file_handle=None):
+def is_fuzz_target_local(file_path, file_opener=None):
```
letitz, line 33: The fact that this could return true for remote zip archive is confusing me. Can you explain a bit more about what this code is doing and how it's used?

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -425,27 +425,38 @@ def _unpack_build(self,
...
-          'Cleared all data directories to free up space, exiting.')
-
-    logs.info(f'Downloading build from {build_url}.')
-    try:
-      storage.copy_file_from(build_url, build_local_archive)
-    except Exception as e:
-      logs.error(f'Unable to download build from {build_url}: {e}')
-      return False
+    should_download = not utils.is_chromium(
+    ) or not build_archive.unzip_over_http_compatible(build_url)
```
letitz, line 429: Suggest inverting this for easier reading:

```py
can_unzip_over_http = utils.is_chromium() and build_archive.compatible(url)
```

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -425,27 +425,38 @@ def _unpack_build(self,
...
+      except Exception as e:
+        logs.error(f'Unable to download build from {build_url}: {e}')
+        return False
+
+      opener = build_archive.open
+      file_arg = build_local_archive
+    else:
+      logs.info("Using an archive over HTTP, skipping archive download.")
+      opener = build_archive.open_uri
+      file_arg = build_url
```
letitz, line 456: Consider extracting this to a helper method with early returns, or two helper methods - one for each branch.

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -425,27 +425,38 @@ def _unpack_build(self,
...
+
+      logs.info(f'Downloading build from {build_url}.')
+      try:
+        storage.copy_file_from(build_url, build_local_archive)
+      except Exception as e:
+        logs.error(f'Unable to download build from {build_url}: {e}')
+        return False
+
+      opener = build_archive.open
+      file_arg = build_local_archive
```
letitz, line 452: You can bind `opener` to `file_arg` and avoid having to thread both variables below, by passing a callback with no arguments instead.

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -328,3 +331,34 @@ def open(archive_path: str) -> BuildArchive:
...
+      the build archive.
+  """
+  assert unzip_over_http_compatible(uri)
+  match = CHROMIUM_GS_MATCHER.match(uri)
+  uri = f'https://commondatastorage.googleapis.com/{match.group(1)}'
+  reader = archive.ZipArchiveReader(archive.HttpZipFile(uri))
+  args_gn_path = os.path.join(reader.root_dir(), 'args.gn')
+  if reader.file_exists(args_gn_path):
+    return ChromeBuildArchive(reader)
+  return DefaultBuildArchive(reader)
```
letitz, line 353: This is copied from `open()` and should be shared instead

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -425,27 +425,38 @@ def _unpack_build(self,
...
+    else:
+      logs.info("Using an archive over HTTP, skipping archive download.")
+      opener = build_archive.open_uri
+      file_arg = build_url
 
     try:
-      with build_archive.open(build_local_archive) as build:
+      with opener(file_arg) as build:
         unpack_everything = environment.get_value(
```
letitz, line 460: It seems to me that if we want to unpack everything, unzipping over http may be slower than downloading then unzipping all at once. Is it still interesting in order to conserve disk space?

letitz (commented):
Only had time to review part of this  - did not get down to `archive.py`, but hopefully this helps already.

src/clusterfuzz/_internal/build_management/build_manager.py
paulsemel: You're right. I meant to condition that also with the `unpack_everything` variable but forgot to do it. That's only interesting for partial unzipping.

src/clusterfuzz/_internal/bot/fuzzers/utils.py
paulsemel: Yeah, I think the function name is not really correct, since it can operate on a `file_handle`, which might not necessarily be local.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -407,6 +412,103 @@ class ArchiveType:
...
+@dataclasses.dataclass
+class CacheBlock:
+  """Represents a cache entry for the HttpZipFile."""
+  start: int
+  end: int
+  content: Sequence[bytes]
+
+
+class HttpZipFile:
+  """This class is a very simple file-object representation of a file over HTTP.
```
letitz, line 424: Let's link to the python docs for file objects.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -407,6 +412,103 @@ class ArchiveType:
...
+
+  def tell(self) -> int:
+    """Provides a tell implementation. Returns the current curso position.
+
+    Returns:
+        the current cursor position.
+    """
+    return self._pos
+
+  def _fetch_from_http(self, start, end) -> Sequence[bytes]:
```
letitz, line 478:
```suggestion
  def _fetch_from_http(self, start: int, end: int) -> Sequence[bytes]:
```

Also let's document whether the range is inclusive of start and end, etc. IIRC the `Range` header uses inclusive ranges.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -407,6 +412,103 @@ class ArchiveType:
...
+    resp = self.session.get(self.uri, headers={'Range': f'bytes={start}-{end}'})
+    return resp.content
+
+  def _fetch_from_cache(self, start, end) -> Sequence[bytes]:
```
letitz, line 482: Same comment.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -407,6 +412,103 @@ class ArchiveType:
...
+    if self._current_block.start > start or self._current_block.end < end:
+      read_ahead_end = min(self.file_size - 1, end + REMOTE_HTTP_MIN_READ_SIZE)
```
letitz, line 484: `REMOTE_HTTP_MIN_READ_SIZE` does not behave as I expected: I thought this would be the minimum size we would fetch in one request, but this is actually a fixed readahead value.

I thought:

* read [0, 1M) -> fetch [0, 50M)
* read [0, 100M) -> fetch [0, 100M)

But this does:

* read [0, 1M) -> fetch [0, 51M)
* read [0, 100M) -> fetch [0, 150M)

Which is a totally fair policy, let's just make it clearer to the reader. How about:

* rename `REMOTE_HTTP_MIN_READ_SIZE` -> `HTTP_READAHEAD_BYTES`
* document it with a comment explaining its behavior

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -407,6 +412,103 @@ class ArchiveType:
...
+
+    Returns:
+        a sequence of bytes.
+    """
+    if size == -1:
+      size = self.file_size - self._pos
+    read_size = min(self.file_size - self._pos, size)
+    end_range = self._pos + read_size - 1
+    if read_size > REMOTE_HTTP_MIN_READ_SIZE:
+      content = self._fetch_from_http(self._pos, end_range)
```
letitz, line 505: This means we won't cache these bytes - is that intentional? Why not always delegate to `_fetch_from_cache()` and let it determine whether or not to fetch more bytes from the wire?

src/clusterfuzz/_internal/tests/core/build_management/build_manager_test.py
```diff
@@ -323,6 +324,21 @@ def test_setup(self):
...
+    os.environ['RELEASE_BUILD_URL_PATTERN'] = (
+        'https://example.com/path/file-release-([0-9]+).zip')
+    self.mock.get_build_urls_list.return_value = [
+        'gs://path/file-release-10.zip',
+        'gs://path/file-release-2.zip',
+        'gs://path/file-release-1.zip',
+    ]
+    self.mock.time.return_value = 1000.0
+    build = build_manager.setup_regular_build(2)
+    self.assertIsNotNone(build.http_build_url)
```
letitz, line 340: Let's assert the exact value of the URL, since it should be deterministic.

src/clusterfuzz/_internal/tests/core/build_management/build_manager_test.py
```diff
@@ -671,6 +694,53 @@ def __eq__(_, fuzz_target):  # pylint: disable=no-self-argument
               trusted=True)
       ])
 
+  @parameterized.parameterized.expand(['True', 'False'])
```
letitz, line 697: Why not parameterize by actual boolean values instead of strings?

letitz (commented):
Looking good overall! A bunch of small comments.

src/clusterfuzz/_internal/system/archive.py
paulsemel: Yep, that's because the way those archive library works is that they only read what's needed. If the read is greater than 50 MB, that means we're reading a huge file, and we won't need it in following reads. This is only useful for small reads tbh.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -407,6 +412,103 @@ class ArchiveType:
...
+    try:
+      res = requests.head(uri, timeout=5)
+      return res.headers.get('Accept-Ranges') is not None
+    except:
+      return False
+
+  def __init__(self, uri):
+    self.uri = uri
+    self.session = requests.Session()
+    # This slows down the overall performances.
```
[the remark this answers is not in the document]
paulsemel: Yep!

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -402,7 +404,70 @@ def _post_setup_success(self, update_revision=True):
...
+      try:
+        with build_archive.open(build_local_archive) as build:
+          yield build
+      finally:
+        shell.remove_file(build_local_archive)
+    else:
+      logs.info("Opening an archive over HTTP, skipping archive download.")
+      assert http_build_url
+      with build_archive.open_uri(http_build_url) as build:
+        yield build
```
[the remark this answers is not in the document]
paulsemel: Ok, I scratched my head around to make this cleaner while still keeping the correct constraints (read the other comment). WDYT?

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -402,7 +404,70 @@ def _post_setup_success(self, update_revision=True):
...
+      logs.info(f'Downloading build from {build_url}.')
+      try:
+        storage.copy_file_from(build_url, build_local_archive)
+      except Exception as e:
+        logs.error(f'Unable to download build from {build_url}: {e}')
+        raise
+
+      try:
+        with build_archive.open(build_local_archive) as build:
+          yield build
```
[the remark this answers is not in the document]
paulsemel: No, I kind of need the context manager here because I want to delete the file after we used it. I changed the code a bit, and I think that's cleaner, WDYT?

src/clusterfuzz/_internal/bot/fuzzers/utils.py
```diff
@@ -72,22 +75,23 @@ def is_fuzz_target_local(file_path, file_handle=None):
...
+    return False
 
   result = False
   for pattern in FUZZ_TARGET_SEARCH_BYTES:
     # TODO(metzman): Bound this call so we don't read forever if something went
     # wrong.
-    local_file_handle.seek(0)
-    result = utils.search_bytes_in_file(pattern, local_file_handle)
+    file_handle.seek(0)
+    result = utils.search_bytes_in_file(pattern, file_handle)
```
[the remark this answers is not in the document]
paulsemel: Yes, but:
1. We ultimately only download a fuzz target twice, because the other ones won't be selected to be unpacked, thus those will only be downloaded once.
2. This is already true anyways. And yes, I agree this is bad, especially the code you're pointing to.

src/clusterfuzz/_internal/bot/fuzzers/utils.py
```diff
@@ -103,7 +107,7 @@ def get_fuzz_targets_local(path):
         continue
 
       file_path = os.path.join(root, filename)
-      if is_fuzz_target_local(file_path):
+      if is_fuzz_target(file_path):
```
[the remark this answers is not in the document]
paulsemel: Agreed!

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -407,6 +412,103 @@ class ArchiveType:
...
+    self.session.headers.pop('Accept-Encoding')
+    resp = self.session.head(self.uri)
+    self.file_size = int(resp.headers.get('Content-Length', default=0))
+    self._current_block = CacheBlock(0, 0, [])
+    self._pos = 0
+    assert resp.headers.get('Accept-Ranges') is not None
```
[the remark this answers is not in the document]
paulsemel: I honestly do not think this should be a concern. Tbh, nothing prevents that from happening after we created the object, and we cannot really prevent that as well (and we'll end up in a very uncomfortable situation as well). Should we really try to prevent from that? (that's an honest question tbf)

src/clusterfuzz/_internal/bot/fuzzers/utils.py
```diff
@@ -72,24 +75,26 @@ def is_fuzz_target_local(file_path, file_handle=None):
...
-    local_file_handle.close()
-
-  return result
+  # Either use the file opener or open the file ourselves.
+  if not file_opener:
+    file_opener = functools.partial(open, mode='rb')
+  try:
+    with file_opener(file_path) as file_handle:
+      result = False
+      for pattern in FUZZ_TARGET_SEARCH_BYTES:
```
letitz, line 84: Note for future work: If searching these bytes take a long time, we might want to search for them "in parallel" by pushing the pattern list into `search_bytes_in_file`, which can scan each chunk of the file for all patterns (avoiding the need to re-read the file X times) or even reach for something like Aho-Corasick: https://pyahocorasick.readthedocs.io/en/latest/

src/clusterfuzz/_internal/bot/fuzzers/utils.py
```diff
@@ -72,24 +75,26 @@ def is_fuzz_target_local(file_path, file_handle=None):
...
+        result = utils.search_bytes_in_file(pattern, file_handle)
+        if result:
+          break
+
+      file_handle.close()
+
+      return result
+  except:
+    # In case we could not open the file, we consider it's not a fuzzer.
+    return False
```
letitz, line 97: Let's log the exception we're swallowing here as a warning - it's always risky to silently swallow exceptions.

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -328,3 +320,43 @@ def open(archive_path: str) -> BuildArchive:
...
+      archive_path: the path to the archive.
+
+  Raises:
+      If the file could not be opened or if the archive type cannot be handled.
+
+  Returns:
+      the build archive.
+  """
+  reader = archive.open(archive_path)
+  return open_with_reader(reader=reader)
```
letitz, line 341:
```suggestion
  return open_with_reader(reader)
```

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -328,3 +320,43 @@ def open(archive_path: str) -> BuildArchive:
...
+  of now.
+
+  Args:
+      uri: the URI pointing to the zip file.
+
+  Returns:
+      the build archive.
+  """
+  reader = archive.ZipArchiveReader(archive.HttpZipFile(uri))
+  return open_with_reader(reader=reader)
```
letitz, line 355:
```suggestion
  return open_with_reader(reader)
```

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -463,8 +525,7 @@ def _unpack_build(self, base_build_dir, build_dir, build_url):
               'Cleared all data directories to free up space, exiting.')
 
         # Unpack the local build archive.
-        logs.info(
-            f'Unpacking build archive {build_local_archive} to {build_dir}.')
+        logs.info(f'Unpacking build archive {build_url} to {build_dir}.')
```
letitz, line 528: It seems we're no longer logging the path to the zip file we downloaded (in the event we did download the full zip) anywhere. Can you add that to the log on line 432?

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -436,10 +434,74 @@ def _unpack_build(self, base_build_dir, build_dir, build_url):
...
-      return False
+      raise
 
     try:
       with build_archive.open(build_local_archive) as build:
+        yield build
+    finally:
+      shell.remove_file(build_local_archive)
+
+  def _open_build_archive(self, unpack_everything: Optional[bool],
```
letitz, line 445: Let's document `unpack_everything` along with other args.

It also feels a bit weird to pass `unpack_everything` as the first argument - I would have put it last but that's just taste.

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -605,10 +663,18 @@ def __init__(self,
                revision,
                build_url,
                build_prefix='',
-               fuzz_target=None):
+               fuzz_target=None,
+               http_build_url=None):
+    """RegularBuild constructor. See Build constructor for other parameters.
+
+    Args:
+        http_build_url: the http build URL. It differs from build_url as
```
letitz, line 671: Let's also document `build_url` while we're here.

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -605,10 +663,18 @@ def __init__(self,
...
+        build_url points to the GCP bucket URL. Defaults to None.
```
letitz, line 672: I would give an explicit example: `gs://foo/bar.zip` - otherwise sure, they both are GCP bucket URLs, in a sense?

src/clusterfuzz/_internal/system/archive.py
letitz: Ah, I see. Can you explain that in a comment for the future reader? 

Optional comment / musing follows below:

It results in a behavior I find counterintuitive, though that does not terribly matter:

* read(49M) -> fetches 99MB and caches it
* read(50M) -> fetches 50MB, does not cache it

I think this could be addressed by doing something like:

```py
HTTP_MAX_BYTES = 100 * 1024 * 1024

def fetch_size(size):
  """How many bytes to fetch over HTTP to serve `read(size)`."""
  cache_size = min(size + HTTP_READAHEAD_BYTES, HTTP_MAX_BYTES)
  return max(size, cache_size)
```

Then at least you don't get a cliff effect for network fetches. But is it useful? Probably not?

letitz (commented):
LGTM with a few last comments.

The only major thing remaining is `_maybe_get_http_build_url()`. I would love to switch that to a straightforward string manipulation instead of relying on environment variables.

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -328,3 +320,43 @@ def open(archive_path: str) -> BuildArchive:
   if reader.file_exists(args_gn_path):
     return ChromeBuildArchive(reader)
   return DefaultBuildArchive(reader)
+
+
+# pylint: disable=redefined-builtin
+def open(archive_path: str) -> BuildArchive:
```
jonathanmetzman, line 326: I think it's generally better to put this pylint directive on the same line were disabling the check for.
The way it is currently done, it is disabling this check for the whole file.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -407,6 +415,147 @@ class ArchiveType:
...
+    end: the end of the byte range in the file (inclusive).
+    content: the sequence of bytes for this range.
+  """
+  start: int
+  end: int
+  content: bytes
+
+
+class HttpZipFile(io.IOBase):
+  """This class is a very simple file-like object representation of a file over
```
jonathanmetzman, line 432: Should probably read "of a zip file"

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -407,6 +415,147 @@ class ArchiveType:
...
+      res = requests.head(uri, timeout=HTTP_REQUEST_TIMEOUT)
+      return res.headers.get('Accept-Ranges') is not None
+    except Exception as e:
+      logs.warning(f'Request to {uri} failed: {e}')
+      return False
+
+  def __init__(self, uri: str):
+    self.uri = uri
+    self.session = requests.Session()
+    # This slows down the overall performances, because encodding something
```
jonathanmetzman, line 452: should probably be compressing instead of encoding.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -407,6 +415,147 @@ class ArchiveType:
...
+    start: the start of the byte range in the file.
+    end: the end of the byte range in the file (inclusive).
+    content: the sequence of bytes for this range.
+  """
+  start: int
+  end: int
+  content: bytes
+
+
+class HttpZipFile(io.IOBase):
```
jonathanmetzman, line 431: Is all this code original? if it's inspired from some other implementation, maybe it should be linked to?

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -1076,6 +1121,27 @@ def _get_latest_revision(bucket_paths):
...
+def _maybe_get_http_build_url(revision) -> Optional[str]:
+  """Tries to get the HTTP build URL if it exists.
+
+  Args:
+      revision: the revision we're currently handling.
+
+  Returns:
+      the build URL.
+  """
+  http_build_url_pattern = environment.get_value('RELEASE_BUILD_URL_PATTERN')
```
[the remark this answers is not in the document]
jonathanmetzman: This is a good point. I don't think we have any function for this but I think replacing gs:// with https://storage.googleapis.com/ should work.

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -1076,6 +1142,35 @@ def _get_latest_revision(bucket_paths):
...
+  if not http_build_url_pattern:
+    return None
+  dir_name = os.path.dirname(http_build_url_pattern)
+  basename = os.path.basename(http_build_url_pattern)
```
jonathanmetzman, line 1158: If you want to use the path functions, I think you really should use posixpath, because on windows, I think this breaks.

src/clusterfuzz/_internal/build_management/build_manager.py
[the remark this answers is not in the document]
jonathanmetzman: I think debug and symbolized builds are a full chrome thing, and aren't used for engine fuzzer builds.

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -1076,6 +1142,35 @@ def _get_latest_revision(bucket_paths):
   return None
 
 
+def _maybe_get_http_build_url(revision) -> Optional[str]:
```
jonathanmetzman, line 1145: One thing to maybe note is that by converting the build URL to HTTP, I think we lose the ability to authenticate without any extra work. This may not be a problem in current chrome builds but may be a problem in the future.

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -436,10 +434,74 @@ def _unpack_build(self, base_build_dir, build_dir, build_url):
...
+    Raises:
+        if an error occurred while accessing the file over HTTP or while
+        downloading the file on disk.
+
+    Returns:
+        the build archive.
+    """
+    # We only want to use remote unzipping if we're not unpacking everything and
+    # if the HTTP URL is compatible with remote unzipping.
+    can_unzip_over_http = not unpack_everything and http_build_url\
```
jonathanmetzman, line 467: Can you use parenthesis here instead of backslashes.

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -1076,6 +1142,35 @@ def _get_latest_revision(bucket_paths):
...
+def _maybe_get_http_build_url(revision) -> Optional[str]:
+  """Tries to get the HTTP build URL if it exists.
+
+  Args:
+      revision: the revision we're currently handling.
+
+  Returns:
+      the build URL.
+  """
+  http_build_url_pattern = environment.get_value('RELEASE_BUILD_URL_PATTERN')
```
jonathanmetzman, line 1154: RELEASE_BUILD_URL_PATTERN isn't actually used anywhere. I don't think we should use it and we should probably remove it from the job definitions. Instead we should use RELEASE_BUILD_BUCKET_PATH

src/clusterfuzz/_internal/build_management/build_manager.py
jonathanmetzman, line 467: I think we should probably gate this feature on an env var. We can add this env var to windows_libfuzzer_chrome_asan at first (which is already broken) and then expand it to more jobs or make it opt-out.

src/clusterfuzz/_internal/system/archive.py
paulsemel: Nope, that's original code I wrote.

oliverchang (commented):
Happy to get this merged if this resolves Chrome's immediate issues. 

That said, would it also be worthwhile also scoping out what it would take for Chrome to migrate to split builds? i.e. https://github.com/google/clusterfuzz/blob/b529f775fcf2bb7efc95991701a67ffa05701b8d/src/clusterfuzz/_internal/build_management/build_manager.py#L653

There's a fair bit of complexity already with the different build types we have, and long term we should aim to simplify this.

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -1116,6 +1188,9 @@ def setup_regular_build(revision,
 
     return None
 
+  # build_url points to a GCP bucket, and we're only converting it to its HTTP
+  # endpoint so that we can use remote unzipping.
+  http_build_url = build_url.replace('gs://', 'https://storage.googleapis.com/')
```
letitz, line 1193: Optional: now that the transformation is so simple, you could push it into `HttpZipFile` if you wanted.

src/clusterfuzz/_internal/system/archive.py
```diff
@@ -39,6 +42,11 @@
 ARCHIVE_FILE_EXTENSIONS = (
     ZIP_FILE_EXTENSIONS + TAR_FILE_EXTENSIONS + LZMA_FILE_EXTENSIONS)
 
+# This is the size of the internal buffer that we're using to cache HTTP
+# range bytes.
+HTTP_BUFFER_SIZE = 70 * 1024 * 1024  # 50 MB
```
letitz, line 47:
```suggestion
HTTP_BUFFER_SIZE = 50 * 1024 * 1024  # 50 MB
```

src/clusterfuzz/_internal/system/archive.py
letitz: Thanks for adding a comment. It only explains what is happening, but not *why*, which is really what's interesting here. In other words, I can see for myself from reading the code that we don't cache large requests, but it does not tell me why that is. You explained it well in your PR comment, can you do that in the code?

letitz (approved):
LGTM with a couple last comments.

letitz:
@oliverchang right, I had filed crbug.com/333965940 to look into that. I think it's the right direction longer term, but I think it significantly more work than this change.

src/clusterfuzz/_internal/build_management/build_manager.py
paulsemel: That's true, although I think we should rather push it in the `_unpack` function maybe? Let's keep it here for now, there will be subsequent changes once we validate that this works as expected.

paulsemel:
Note: whenever this lands, don't forget that you need to add the `ALLOW_UNPACK_OVER_HTTP = True` environment variable on the job so that this kicks in.

src/clusterfuzz/_internal/build_management/build_manager.py
jonathanmetzman: E.g. we fuzzed widevine-based builds in the past.

PR #4264, opened 2024-09-20
Fix autohealing policy update logic.

letitz:
There are two logic fixes:

1. API field names are in `camelCase`, not `snake_case`.
2. Previously, we would treat trying to update a single field as an error and delete the policy.
   It is only an error to *supply* a single field in the config, but we may legitimately find a single field needs updating.

Also moved config validation to loading time. It would likely be best to explode if the auto-healing policy in the config has a single field set, instead of ignoring it with a warning, but this preserves behavior for now.

### DIFF
--- a/src/clusterfuzz/_internal/cron/manage_vms.py
+++ b/src/clusterfuzz/_internal/cron/manage_vms.py
@@ -19,9 +19,9 @@ import copy
 import itertools
 import json
 import logging
+from typing import Optional
 
 from google.cloud import ndb
-from tuping import Optional
 
 from clusterfuzz._internal.base import utils
 from clusterfuzz._internal.config import local_config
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
Awesome thanks! Do you think some tests can be added, or would you like me to just merge as-is.

jonathanmetzman:
OK I suppose this is fine to merge now.
@vitorguidi Can you merge and deploy this change since I am waiting for my workstation to work again.

vitorguidi:
> OK I suppose this is fine to merge now. @vitorguidi Can you merge and deploy this change since I am waiting for my workstation to work again.

Will do as soon as I stabilize the cronjob metrics

letitz:
Please don't merge this yet!

> Looking into this made me realize that I need to update the instance group creation path as well as the update path, now that Cluster.auto_healing_policy has a different type

letitz:
I've updated the code path for instance group creations. I don't have time to write any more tests here unfortunately, so if tests are green I think we should land this and watch `manage_vms` in production for a bit to make sure it runs OK.

jonathanmetzman (approved):
lgtm

PR #4266, opened 2024-09-20
Add asynchronous downloading to speed up signed URL downloading.

jonathanmetzman:
The speed ups are up to 20x on my own machine.
It's even faster with one core than 16 parallel processes.


src/Pipfile
```diff
@@ -51,4 +52,5 @@ itsdangerous = "==2.0.1"
 Jinja2 = "==3.1.4"
 PyJWT = "==2.7.0"
 requests-toolbelt = "==0.9.1"
-werkzeug = "==2.2.2"
\ No newline at end of file
+werkzeug = "==2.2.2"
+aiohttp = "==3.10.5"
```
jonathanmetzman, line 56: Is this the right way to add a package to both appengine and gce?
The package will be used by a depedency of storage.py which is used in appengine.
Maybe I should exclude the package and do `try: import ... except:...`

jonathanmetzman:
TODO: Do this for other http methods.

vitorguidi:
> The speed ups are up to 20x on my own machine. It's even faster with one core than 16 parallel processes.

Do we have any idea of how long the downloading step takes from past profiling data, on fuzz task? (back when those profiling images were around)

src/Pipfile.lock
```diff
@@ -14,13 +14,135 @@
...
+                "sha256:f29930bc2921cef955ba39a3ff87d2c4398a0394ae217f41cb02d5c26c8b1b77",
+                "sha256:f489a2c9e6455d87eabf907ac0b7d230a9786be43fbe884ad184ddf9e9c1e385",
+                "sha256:f5bf3ead3cb66ab990ee2561373b009db5bc0e857549b6c9ba84b20bc462e172",
+                "sha256:f6f18898ace4bcd2d41a122916475344a87f1dfdec626ecde9ee802a711bc569",
+                "sha256:f8112fb501b1e0567a1251a2fd0747baae60a4ab325a871e975b7bb67e59221f",
+                "sha256:fd31f176429cecbc1ba499d4aba31aaccfea488f418d60376b911269d3b883c5"
+            ],
+            "index": "pypi",
+            "markers": "python_version >= '3.8'",
+            "version": "==3.10.5"
```
vitorguidi, line 121: Are these dep bumps supported in our platforms? 

There should be complaints from https://github.com/google/clusterfuzz/blob/044ec0848c18544561b11a79033498d02b697309/src/local/butler/common.py#L295 during butler deploy, if not.

vitorguidi: Might also need to rebuild docker, since httpaio is a new dep with C extensions to access async syscalls

jonathanmetzman: I'm pretty sure we're not using the C extensions.

jonathanmetzman: But I'll know what to look out for if I see a problem.

jonathanmetzman:
> > The speed ups are up to 20x on my own machine. It's even faster with one core than 16 parallel processes.
> 
> Do we have any idea of how long the downloading step takes from past profiling data, on fuzz task? (back when those profiling images were around)

We got rid of profiling before we switched to this new method of downloading, but seeing as the new method can take up to 15 minutes, I think the savings will be substantial.

src/Pipfile
oliverchang: non-dev dependencies here will go into `src/third_party`, which is shared by appengine and gce. 

if we want appengine-**only** dependencies, we can just specify it in `dev-packages`. So, I think we can just remove line 56 here to achieve what you want.

src/clusterfuzz/_internal/system/fast_http.py
```diff
@@ -0,0 +1,86 @@
...
+    ]
+    return await asyncio.gather(*tasks)
+
+
+def _error_tolerant_download_file(session: aiohttp.ClientSession, url: str,
+                                  path: str) -> bool:
+  try:
+    return _async_download_file(session, url, path)
+  except:
+    return False
```
oliverchang, line 75: Should we log an error here?

src/clusterfuzz/_internal/system/fast_http.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+async def _async_download_file(session: aiohttp.ClientSession, url: str,
+                               path: str) -> bool:
+  async with session.get(url) as response:
+    with open(path, 'wb') as fp:
+      while True:
+        chunk = await response.content.read(1024)
+        if not chunk:
+          return True
```
oliverchang, line 85: we never return `False` ?

src/clusterfuzz/_internal/system/fast_http.py
```diff
@@ -0,0 +1,86 @@
...
+@contextlib.contextmanager
+def _pool(pool_size=_POOL_SIZE):
+  if (environment.get_value('PY_UNITTESTS') or
+      environment.platform() == 'WINDOWS'):
+    yield futures.ThreadPoolExecutor(pool_size)
+  else:
+    yield futures.ProcessPoolExecutor(pool_size)
+
+
+def download_urls(urls: List[str], filepaths: List[str]) -> str:
```
oliverchang, line 39: the return type looks wrong.

src/clusterfuzz/_internal/system/fast_http.py
```diff
@@ -0,0 +1,86 @@
...
+  """Downloads multiple |urls| to |filepaths| in parallel and
+  asynchronously. Tolerates errors. Returns a list of whether each
+  download was successful."""
+  # TOOD(metzman): Move the multiprocessing code over from storage.
+  assert len(urls) == len(filepaths)
+  chunks = []
```
oliverchang, line 45: nit: is `batches` a better term here? 

`chunk` is overloaded in this file (in `_async_download_file` where it has a different meaning).

src/clusterfuzz/_internal/system/fast_http.py
```diff
@@ -0,0 +1,86 @@
...
+
+from clusterfuzz._internal.system import environment
+
+_POOL_SIZE = multiprocessing.cpu_count()
+
+
+@contextlib.contextmanager
+def _pool(pool_size=_POOL_SIZE):
+  if (environment.get_value('PY_UNITTESTS') or
+      environment.platform() == 'WINDOWS'):
```
oliverchang, line 33: Can you add a comment explaining why such a special case exists?

src/clusterfuzz/_internal/system/fast_http.py
```diff
@@ -0,0 +1,86 @@
...
+    yield futures.ThreadPoolExecutor(pool_size)
+  else:
+    yield futures.ProcessPoolExecutor(pool_size)
+
+
+def download_urls(urls: List[str], filepaths: List[str]) -> str:
+  """Downloads multiple |urls| to |filepaths| in parallel and
+  asynchronously. Tolerates errors. Returns a list of whether each
+  download was successful."""
+  # TOOD(metzman): Move the multiprocessing code over from storage.
```
oliverchang, line 43: Can you point to the exact code in this comment?

src/clusterfuzz/_internal/google_cloud_utils/storage.py
```diff
@@ -22,6 +22,7 @@
 import shutil
 import threading
 import time
+from typing import List
```
oliverchang, line 25: I think we can get rid of this and just use `list` directly now that we dropped Python 3.7 support.

oliverchang (approved):

PR #4267, opened 2024-09-20
Optimize tests-syncer crashtest query

mi-ac:
This limits the crashtest list already when executing the query instead of fetching all and pruning later. This also sorts the tests newest first, so that we rather archive newer tests if the cap is reached.

Context: http://b/365801496

### DIFF
--- a/src/python/other-bots/chromium-tests-syncer/run.py
+++ b/src/python/other-bots/chromium-tests-syncer/run.py
@@ -48,7 +48,16 @@ STORED_TESTCASES_LIST = []
 def unpack_crash_testcases(crash_testcases_directory):
   """Unpacks the old crash testcases in the provided directory."""
   count = 0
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
I'm not 100% sure this will work. We use a nosql database that I've noticed can't do basic queries SQL can.
I'm happpy to merge and see what happens.

jonathanmetzman:
You know, if you rename main to `def execute(_):` and move it to `src/local/butler/scripts/` you can execute this locally which will help you test this. You can even just test portions (I would probably do that).

command to run:
```
python butler.py run --config-dir ../clusterfuzz-config/configs/internal --non-dry-run $YOUR_SCRIPT
```

mi-ac:
Thanks, will try that in case we need to revert. I copy/pasted this from other queries for test cases in the repo, which had similar constructs.

PR #4268, opened 2024-09-22
Fix tests-syncer crashtest query

mi-ac:
Currently the bug-information inequality isn't compatible with the other parts of the query. This fixes the query using an existing index.

Context: http://b/365801496


jonathanmetzman (approved):
lgtm

PR #4270, opened 2024-09-24
Instrumenting cronjob exit codes

vitorguidi:
### Motivation

Kubernetes signals that cronjobs fail, after retries, through [events](https://kubernetes.io/docs/reference/kubernetes-api/cluster-resources/event-v1/).

GKE does not make it easy to alert on this failure mode. For cronjobs without retries, the failure will be evident in the GCP panel. For cronjobs with retries, there will be a success indicator for the last successful cronjob, and the failure will be registered under the events panel.

### Alternatives considered

Options available to monitor failing cronjobs are:

- the container/restart_count metric, from [GKE](https://cloud.google.com/monitoring/api/metrics_kubernetes). This will be flaky, since a job might succeed on the third attempt. Also, it is not easy to pinpoint the cronjob, since we get the container name as a label
- alert on a log based metric from the cluster events. The output of kubectl get events gets dumped to cloud logging, so we can create a metric on events of the form "Saw completed job: oss-fuzz-apply-ccs-28786460, status: Failed". However this requires regex manipulation, has to be manually applied across all projects, and also makes it hard to derive the failing cronjob from the container name. It also adds a hard dependency on kubernetes.

### Solution

The proposed solution is to reuse the builtin clusterfuzz metrics implementation, and add a gauge metric CLUSTERFUZZ_CRON_EXIT_CODE, with the cron name as a label.

 If the metric is at 1, then the cronjob is unhealthy, otherwise it is healthy. An alert must be set for when it reaches the 1 state, for every label.
 
 Since cronjobs are ephemeral, there is no need for a thread to continuously flush metrics. The option to use monitoring without a flushing thread was added. The same approach can be used to fix metrics for swarming/batch.
 
 Also, the flusher thread was changed to make sure that leftover metrics are flushed before it stops.
 
 Note: this PR is part of this [initiative](https://github.com/google/clusterfuzz/issues/4271)

### DIFF
--- a/src/clusterfuzz/_internal/metrics/monitor.py
+++ b/src/clusterfuzz/_internal/metrics/monitor.py
@@ -114,11 +114,12 @@ class _FlusherThread(threading.Thread):
     project_path = _monitoring_v3_client.common_project_path(  # pylint: disable=no-member
         utils.get_application_id())
 
+    should_stop = False
     while True:
+      # Make sure there are no metrics left to flush after monitor.stop() is called
+      if self.stop_event.wait(FLUSH_INTERVAL_SECONDS):
+          should_stop = True
       try:
-        if self.stop_event.wait(FLUSH_INTERVAL_SECONDS):
-          return
-
         time_series = []
         end_time = time.time()
         for metric, labels, start_time, value in _metrics_store.iter_values():
@@ -146,6 +147,8 @@ class _FlusherThread(threading.Thread):
           logs.warning(f'Failed to flush metrics: {e}')
         else:
           logs.error(f'Failed to flush metrics: {e}')
+      if should_stop:
+        return
 
   def stop(self):
     self.stop_event.set()
--- a/src/clusterfuzz/_internal/metrics/monitor.py
+++ b/src/clusterfuzz/_internal/metrics/monitor.py
@@ -116,9 +116,9 @@ class _FlusherThread(threading.Thread):
 
     should_stop = False
     while True:
-      # Make sure there are no metrics left to flush after monitor.stop() is called
+      # Make sure there are no metrics left to flush after monitor.stop()
       if self.stop_event.wait(FLUSH_INTERVAL_SECONDS):
-          should_stop = True
+        should_stop = True
       try:
         time_series = []
         end_time = time.time()
--- a/src/clusterfuzz/_internal/metrics/monitor.py
+++ b/src/clusterfuzz/_internal/metrics/monitor.py
@@ -100,6 +100,37 @@ class _MockMetric:
 def _time_series_sort_key(ts):
   return ts.points[-1].interval.start_time
 
+def flush_metrics():
+  project_path = _monitoring_v3_client.common_project_path(  # pylint: disable=no-member
+      utils.get_application_id())
+  try:
+    time_series = []
+    end_time = time.time()
+    for metric, labels, start_time, value in _metrics_store.iter_values():
+      if (metric.metric_kind == metric_pb2.MetricDescriptor.MetricKind.GAUGE  # pylint: disable=no-member
+          ):
+        start_time = end_time
+
+      series = _TimeSeries()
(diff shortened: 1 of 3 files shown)


src/python/bot/startup/run_cron.py
```diff
@@ -55,12 +57,35 @@ def main():
   except ImportError:
     pass
 
+  # Few metrics get collected per job run, so
+  # no need for a thread to continuously push those
+  monitor.initialize(use_flusher_thread=False)
```
oliverchang, line 62: Is there enough of a benefit to do this special casing ? 

If there's not much performance benefit, it's probably better to keep things more simple and uniform (i.e. always have a thread).

src/clusterfuzz/_internal/metrics/monitor.py
```diff
@@ -111,41 +145,14 @@ def __init__(self):
...
-        if environment.is_android():
-          # FIXME: This exception is extremely common on Android. We are already
-          # aware of the problem, don't make more noise about it.
-          logs.warning(f'Failed to flush metrics: {e}')
-        else:
-          logs.error(f'Failed to flush metrics: {e}')
+      # Make sure there are no metrics left to flush after monitor.stop()
+      if self.stop_event.wait(FLUSH_INTERVAL_SECONDS):
+        should_stop = True
+      flush_metrics()
```
oliverchang, line 153: Can you add details of this change to PR description also? Is this addressing a problem with the final set of metrics not being flushed?

oliverchang (approved):

src/python/bot/startup/run_cron.py
vitorguidi: It is mostly because of batch and swarming. Suppose a task takes 1m to run. We would have to wait for the entirety of FLUSH_INTERVAL_SECONDS (10m hardcoded) for them to dump, which would cause burn compute for no reason.

This thread is a weird concept for cron/batch/swarming, where only one task gets executed and the bot immediately quits. Hence this special casing, which IMO fits the ephemeral use cases better. The thread still makes sense for the use case of long lived bots, though.

jonathanmetzman: This is an interesting tradeoff, but 10 minutes seems tolerable for most of our cron jobs and 10 minutes is probably overly conservative anyway and can be reduced.

vitorguidi: For crons maybe, but for batch/swarming where there is stuff that runs under a minute, it is really hard for me to see value in using this thread.

letitz: I tend to agree that as long as we make sure the thread flushes before stopping, it seems like a micro-optimization to avoid having a flusher thread.

vitorguidi: What value of FLUSH_INTERVAL_SECONDS should be chosen for this thread to be an universal solution?
If there is such a value that will prevent getting rate limited by the GCP metrics endpoint, and also not decrease the throughput of batch, I am fine with it.

However, this would take experimentation, which is not a really wise time investment. I would rather not have the trouble of choosing and stabilizing this value for short lived use cases.

oliverchang: I think the interval isn't really critical to this decision. For shorter batch / cron jobs, could we instrument the interpreter exit instead to flush metrics? i.e. have `monitor.initialize()` register an atexit handler. 

That way, we keep things uniform in that we initialize this the same way, while not worrying about flush intervals.

vitorguidi: I can refactor this in the upcoming batch PR.
There is one last complexity: kubernetes can kill cronjobs due to timeout with a SIGTERM, followed by a SIGKILL. This SIGTERM has to be treated somehow

https://stackoverflow.com/questions/59973600/what-does-shutdown-look-like-for-a-kubernetes-cron-job-pod-when-its-being-termi

PR #4272, opened 2024-09-25
Pointing k8s cronjobs to the prod code release and a non hacky docker image

vitorguidi:
Docker images were temporarily set to vguidi_py311_testing_v1 to execute the migration process for 311.

This PR sets it to a proper tag, and makes the cronjobs pull from the production release instead of k8s.

This operation should not touch the clusterfuzz source code repository, we need to move these resources to an internal config repo (and kill butler deploy for kubernetes resources, swap for argo)

### DIFF
--- a/infra/k8s/backup.yaml
+++ b/infra/k8s/backup.yaml
@@ -26,11 +26,11 @@ spec:
         spec:
           containers:
           - name: backup
-            image: gcr.io/clusterfuzz-images/base:vguidi_py311_testing_v1
+            image: gcr.io/clusterfuzz-images/base:091c6c2-202409251610
             imagePullPolicy: Always
             env:
             - name: CLUSTERFUZZ_RELEASE
-              value: "candidate"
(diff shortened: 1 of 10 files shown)


infra/k8s/backup.yaml
```diff
@@ -26,11 +26,11 @@ spec:
         spec:
           containers:
           - name: backup
-            image: gcr.io/clusterfuzz-images/base:vguidi_py311_testing_v1
+            image: gcr.io/clusterfuzz-images/base:091c6c2-202409251610
```
jonathanmetzman, line 29: Are we pinning these images but running the bots on latest?

vitorguidi: Bots will be pinned as well, a CL will show up internally soon

jonathanmetzman: OK!

oliverchang (commented):
LGTM!!!

PR #4275, opened 2024-09-26
Add index for an improved test-syncer query

mi-ac:
This prepares relanding a more efficient query to the tests-syncer that also pre-filters the bug information. When testing the new query locally, gae suggests this index to serve the query.

Context: http://b/365801496


letitz (approved):
LGTM, though I've never touched these files so take it with a grain of salt :)

PR #4276, opened 2024-09-26
add support for predator component id

paulsemel:
This is only when dealing with google_issue_tracker.

### DIFF
--- a/src/clusterfuzz/_internal/cron/cleanup.py
+++ b/src/clusterfuzz/_internal/cron/cleanup.py
@@ -136,8 +136,8 @@ def cleanup_testcases_and_issues():
       update_fuzz_blocker_label(policy, testcase, issue,
                                 top_crashes_by_project_and_platform_map)
       logs.info('maybe updated fuzz blocker')
-      update_component_labels(policy, testcase, issue)
-      logs.info('maybe updated component labels')
+      update_component_labels_and_id(policy, testcase, issue)
+      logs.info('maybe updated component labels and component id')
       update_issue_ccs_from_owners_file(policy, testcase, issue)
       logs.info('maybe updated issueccs')
       update_issue_owner_and_ccs_from_predator_results(policy, testcase, issue)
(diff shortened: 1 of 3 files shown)


src/clusterfuzz/_internal/cron/cleanup.py
```diff
@@ -1070,6 +1072,11 @@ def update_component_labels(policy, testcase, issue):
   for filtered_component in filtered_components:
     issue.components.add(filtered_component)
 
+  # This is very specific to google_issue_tracker, so this attribute might not
+  # be available for other issue trackers.
+  if component_id and hasattr(issue, 'component_id'):
+    issue.component_id = component_id
```
letitz, line 1078: Would this attribute assignment explode for other issue trackers, or would the assigned value simply be ignored?  If the latter, we can set it regardless of whether issue has a `component_id` attribute yet.

src/clusterfuzz/_internal/tests/appengine/handlers/cron/cleanup_test.py
```diff
@@ -1363,7 +1363,7 @@ def test_top_crashes_match_and_label_removed(self):
 
 @test_utils.with_cloud_emulators('datastore')
 class UpdateComponentsTest(unittest.TestCase):
```
letitz, line 1365: Let's add a test that checks `component_id` is assigned.

src/clusterfuzz/_internal/cron/cleanup.py
letitz, line 1078: Have you checked that if this attribute is set, the we will make the right API call to buganizer to update it on the issue? Ideally there is a test for this already :)

letitz (commented):
Thanks for sending this in!

src/clusterfuzz/_internal/cron/cleanup.py
paulsemel: No, that would explode, which is why it's conditioned.

paulsemel: So that part is already tested in `google_issue_tracker_test.py`

src/clusterfuzz/_internal/tests/appengine/handlers/cron/cleanup_test.py
paulsemel: I think that's the most I can do about it, because otherwise it would require tons of work to rework how the cleanup.py and cleanup_test.py work. Those are really tied to monorail initially, and aren't really made to have per-tracker cases. I think that would be too much work just for that feature.

src/clusterfuzz/_internal/cron/cleanup.py
letitz: Awesome, thanks for checking.

src/clusterfuzz/_internal/tests/appengine/handlers/cron/cleanup_test.py
letitz: Good enough, thanks!

letitz (approved):
LGTM

letitz:
@jonathanmetzman I'll merge this and let you deploy.

PR #4278, opened 2024-09-26
Fix AFL minimization

jonathanmetzman:
Minimization tasks need to be created for AFL even though AFL itself doesn't support minimization.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/task_creation.py
+++ b/src/clusterfuzz/_internal/bot/tasks/task_creation.py
@@ -248,7 +248,9 @@ def create_tasks(testcase):
     testcase.put()
     return
 
-  if environment.is_minimization_supported():
+  if environment.is_minimization_supported() or environment.get_value(
+    'MINMIZE_FUZZER_OVERRIDE'
+  ):
     # For supported environments, just create the minimize task for now.
     # Once minimization is complete, it automatically creates the rest of the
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/bot/tasks/task_creation.py
```diff
@@ -248,7 +248,9 @@ def create_tasks(testcase):
     testcase.put()
     return
 
-  if environment.is_minimization_supported():
+  if environment.is_minimization_supported() or environment.get_value(
+    'MINMIZE_FUZZER_OVERRIDE'
```
oliverchang, line 252: misspelling? 

Also, would it make sense to just move this env var check into `is_minimization_supported` ?

oliverchang (changes requested):

jonathanmetzman:
This was broken by https://github.com/google/clusterfuzz/pull/3787

src/clusterfuzz/_internal/bot/tasks/task_creation.py
jonathanmetzman: I think it makes sense not to treat minimization as supported here: https://github.com/google/clusterfuzz/pull/3787/files#diff-e8255271eeeadc2bda46b215fbc7b0bb160ef1116f6e27ff895990d88caafa5fR382

PR #4279, opened 2024-09-27
utils: accelerate byte search in big file handles

paulsemel:
The current implementation was unperformant for multiple reasons:
- it searched for lines, which means it forced the internal implementation to search for '\n' in the bytes.
- as mentioned in the old comment, it was actually not memory stable, because the length of the line could be huge for big files, and thus didn't really prevent OOM issues.
- was not exactly correct, because prevented from finding patterns containing '\n'.

On huge file (such as chrome, which is why I started investigating this), this new implementation improves by a factor of 5 the performances of this function, and thus tremendously reducing the amount of time spent here.

### DIFF
--- a/src/clusterfuzz/_internal/base/utils.py
+++ b/src/clusterfuzz/_internal/base/utils.py
@@ -685,7 +685,7 @@ def search_bytes_in_file(search_bytes, file_handle) -> bool:
       file_handle: the file-object handle.
 
   Returns:
-      whether the pattern was found in the file.
+      Whether the pattern was found in the file.
   """
   # Invariant: `buf[:offset]` contains valid bytes.
   buf = bytearray(DEFAULT_SEARCH_BUFFER_LENGTH)
--- a/src/clusterfuzz/_internal/base/utils.py
+++ b/src/clusterfuzz/_internal/base/utils.py
@@ -70,6 +70,9 @@ THREAD_POOL = None
 LOCAL_SOURCE_MANIFEST = os.path.join('src', 'appengine', 'resources',
                                      'clusterfuzz-source.manifest')
 
+# The length we're reading from the file-object in search_bytes_in_file.
+DEFAULT_SEARCH_BUFFER_LENGTH = 10 * 1024 * 1024  # 10MB
+
 
 def utcnow():
   """Return datetime.datetime.utcnow(). We need this method because we can't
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/base/utils.py
```diff
@@ -70,6 +70,8 @@
 LOCAL_SOURCE_MANIFEST = os.path.join('src', 'appengine', 'resources',
                                      'clusterfuzz-source.manifest')
 
+DEFAULT_SEARCH_BUFFER_LENGTH = 10 * 1024 * 1024  # 10MB
```
letitz, line 73: Let's document this constant.

src/clusterfuzz/_internal/base/utils.py
```diff
@@ -673,16 +675,34 @@ def restart_machine():
     os.system('sudo shutdown -r now')
 
 
-def search_bytes_in_file(search_bytes, file_handle):
+def search_bytes_in_file(search_bytes,
+                         file_handle,
+                         buffer_length=DEFAULT_SEARCH_BUFFER_LENGTH):
   """Helper to search for bytes in a large binary file without memory
   issues.
   """
```
letitz, line 683: Let's document at least `buffer_length`, though we might as well document the whole function while we're at it.

src/clusterfuzz/_internal/base/utils.py
```diff
@@ -673,16 +675,34 @@ def restart_machine():
     os.system('sudo shutdown -r now')
 
 
-def search_bytes_in_file(search_bytes, file_handle):
+def search_bytes_in_file(search_bytes,
+                         file_handle,
+                         buffer_length=DEFAULT_SEARCH_BUFFER_LENGTH):
```
letitz, line 680: It does not seem that anyone except tests passes a non-default buffer length. Let's remove the parameter.

src/clusterfuzz/_internal/base/utils.py
```diff
@@ -673,16 +675,34 @@ def restart_machine():
...
-  # line.
-  for line in file_handle:
-    if search_bytes in line:
+  assert buffer_length >= len(search_bytes)
+  buf = bytearray(buffer_length)
+  offset = 0
+  length = len(search_bytes) - 1
+  # We read right after the previously copied offset. This ensures we're
+  # perfectly matching the byte sequence above.
+  while read_len := file_handle.readinto(memoryview(buf)[offset:]):
```
letitz, line 690: Let's immediately wrap `buf` in a `memoryview`. I think things will still work, and it will ensure we never copy `buf` accidentally?

src/clusterfuzz/_internal/base/utils.py
```diff
@@ -673,16 +675,34 @@ def restart_machine():
...
 
+    # We copy exactly one byte less than the actual length of the sequence
+    # we're trying to match. This is to prevent missing stuff like:
+    # sequence: 'abcd'
+    # read #1: 'abc' -> no match
+    # read #2: 'd' -> no match
+    # With the copy, we get:
+    # read #1: 'abc': -> no match
+    # copy: 'abc'
+    # read #2: 'abcd': -> match!
```
letitz, line 702: This comment might be better placed on `length`, to explain the -1 calculation.

src/clusterfuzz/_internal/base/utils.py
```diff
@@ -673,16 +675,34 @@ def restart_machine():
...
   """Helper to search for bytes in a large binary file without memory
   issues.
   """
-  # TODO(aarya): This is too brittle and will fail if we have a very large
-  # line.
-  for line in file_handle:
-    if search_bytes in line:
+  assert buffer_length >= len(search_bytes)
+  buf = bytearray(buffer_length)
+  offset = 0
```
letitz, line 686:
```suggestion
  # Invariant: `buf[:offset]` contains valid bytes.
  buf = bytearray(buffer_length)
  offset = 0
```

letitz (approved):
Very neat! A few nits but LGTM overall.

One comment about naming: "to fasten" means to attach, as in "fasten your seatbelts" - you mean "accelerate".

jonathanmetzman (approved):
lgtm

PR #4280, opened 2024-09-30
Update get_battery_level_and_temperature()

Onkar1300:
Fetch a valid battery 'level' instead of 'Capacity level' to prevent timeout in wait_until_good_state() of battery.py


jonathanmetzman (approved):
lgtm

PR #4289, opened 2024-10-02
Marking closed build issues as Fixed to avoid having a verifier/assignee

vitorguidi:
While trying to close bugs through the oss-fuzz-build-stats cron, we get a 400 reply from buganizer complaining about invalid input. This boils down to:

```
[redacted] Issues with status VERIFIED must have an assignee, Issues with status VERIFIED must have a verifier
```

To avoid this, we change these to be fixed instead.


jonathanmetzman (approved):
LGTM

jonathanmetzman:
Sorry. didn't mean to merge. Can you revert this if tests fail or if you need to make more changes?

Issue #4271, opened 2024-09-25
ClusterFuzz monitoring improvements


vitorguidi:
This issue serves as an umbrella for the monitoring initiative

PR #4291, opened 2024-10-03
Flushing metrics on ephemeral executions

vitorguidi:
As things currently stand, metrics are flushed by a background thread every 10m. However, for the batch/swarming use case, we will lose metrics for jobs that finish before this interval.  To handle that, monitor.stop will be called before the bot exits.

Finally, sigterm is handled in run_bot to avoid losing metrics when instances get preempted

This PR is part of #4271.

### DIFF
--- a/src/clusterfuzz/_internal/tests/core/metrics/monitor_test.py
+++ b/src/clusterfuzz/_internal/tests/core/metrics/monitor_test.py
@@ -15,10 +15,8 @@
 # pylint: disable=protected-access
 
 import os
-import queue
 import time
 import unittest
-from unittest.mock import patch, MagicMock
 
 from clusterfuzz._internal.metrics import monitor
 from clusterfuzz._internal.metrics import monitoring_metrics
@@ -62,7 +60,7 @@ class InitializeTest(unittest.TestCase):
   def test_initialize(self):
     """Tests initialization."""
     monitor.initialize()
-    self.assertEqual(2, self.mock.start.call_count)
+    self.assertEqual(1, self.mock.start.call_count)
 
 
 class MonitorTest(unittest.TestCase):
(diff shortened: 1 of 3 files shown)


src/clusterfuzz/_internal/tests/core/metrics/monitor_test.py
```diff
@@ -61,7 +62,7 @@ def test_error(self):
   def test_initialize(self):
     """Tests initialization."""
     monitor.initialize()
-    self.assertEqual(1, self.mock.start.call_count)
+    self.assertEqual(2, self.mock.start.call_count)
```
vitorguidi, line 65: This would check if flusher thread started, so asserted for one. Now that there is a ticker and a flusher thread, it must be 2

src/python/bot/startup/run_bot.py
```diff
@@ -125,7 +126,17 @@ def task_loop():
...
         # Batch/Swarming tasks only run one at a time.
-        sys.exit(utasks.uworker_bot_main())
+        uworker_return_code = None
+        exception = None
+        try:
+          uworker_return_code = utasks.uworker_bot_main()
+        except Exception as e:
+          exception = e
+        finally:
+          monitor.stop()
```
letitz, line 136: You might be able to do this more simply with a context manager:

```py
with monitor.Daemon() as monitoring_daemon:
  sys.exit(utasks.uworker_bot_main())
```

vitorguidi: I never questioned myself as to why this with thing would show up in Python code. Thanks for calling that out!

src/python/bot/startup/run_bot.py
```diff
@@ -84,6 +85,23 @@ def lease_all_tasks(task_list):
...
+  logs.info('Handling sigterm, stopping monitoring daemon.')
+  monitor.stop()
+  logs.info('Sigterm handled, metrics flushed.')
+
+
+@contextlib.contextmanager
+def wrap_with_monitoring():
+  """Wraps execution so we flush metrics on exit"""
+  try:
+    monitor.initialize()
```
letitz, line 98: Note for further work: Now that the `monitor` object is only used in this function, I think you can get rid of the global objects in `monitor.py` and wrap the whole thing in e.g. a `Monitor` class that lives on this function's stack.

jonathanmetzman (approved):

PR #4294, opened 2024-10-04
Fixing oss-fuzz-apply-ccs

vitorguidi:
The cron oss-fuzz-apply-ccs has been broken for a bit. It boils down to this error:

```
'dict_values' object has no attribute 'remove'
  File "/usr/local/google/home/vitorguidi/projects/clusterfuzz/src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py", line 120, in _extract_label
    labels.remove(label)
    ^^^^^^^^^^^^^
  File "/usr/local/google/home/vitorguidi/projects/clusterfuzz/src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py", line 137, in _get_oss_fuzz_reported_value
    added_reported = _extract_label(labels, 'Reported-')
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/google/home/vitorguidi/projects/clusterfuzz/src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py", line 604, in _update_issue
    added_reported = _get_oss_fuzz_reported_value(self.labels.added)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/google/home/vitorguidi/projects/clusterfuzz/src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py", line 824, in save
    result = self._update_issue(new_comment=new_comment, notify=notify)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/google/home/vitorguidi/projects/clusterfuzz/src/clusterfuzz/_internal/cron/oss_fuzz_apply_ccs.py", line 95, in main
    issue.save(new_comment=comment, notify=True)
  File "/usr/local/google/home/vitorguidi/projects/clusterfuzz/src/local/butler/scripts/run_cron.py", line 24, in execute
    oss_fuzz_apply_ccs.main()
  File "/usr/local/google/home/vitorguidi/projects/clusterfuzz/src/local/butler/run.py", line 48, in execute
    script.execute(args)
  File "/usr/local/google/home/vitorguidi/projects/clusterfuzz/butler.py", line 388, in main
    return command.execute(args)
           ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #4295, opened 2024-10-07
Fix instance creation logic.

letitz:
New instance groups may not have auto-healing policies. This broke `manage_vms`.

I broke it in https://github.com/google/clusterfuzz/pull/4264 - who knew integration tests would have been a good idea?

Fixes: https://issues.chromium.org/371829507

### DIFF
--- a/src/clusterfuzz/_internal/cron/manage_vms.py
+++ b/src/clusterfuzz/_internal/cron/manage_vms.py
@@ -160,8 +160,15 @@ def _template_needs_update(current_template, new_template, resource_name):
 
 
 def _auto_healing_policy_to_dict(
-    policy: compute_engine_projects.AutoHealingPolicy) -> Dict[str, Any]:
-  """Converts `policy` into its JSON API representation."""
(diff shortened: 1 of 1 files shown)


vitorguidi:
> New instance groups may not have auto-healing policies. This broke `manage_vms`.
> 
> I broke it in #4264 - who knew integration tests would have been a good idea?
> 
> Fixes: https://issues.chromium.org/371829507

lgtm, but can you run this locally to make sure it succeeds and post the evidence? There are no integration tests, but this can be tested. Instructions here https://github.com/google/clusterfuzz/pull/4294

Change the config dir env vars to your local copy of clusterfuzz-config/configs/internal.

letitz:
Just tried running `manage_vms.main()` locally by hacking `src/local/butler/scripts/run_locally.py` as you did in #4294. It just hangs there waiting for all cluster updates to be applied, and logs nothing to the console. I got a bit worried I would be modifying the state of production in a bad way, so I bailed out. Can we land and observe how `manage_vms` does?

vitorguidi (approved):
lgtm

PR #4296, opened 2024-10-07
Remove job mappings info logs

svasudevprasad:
HOST_JOB_SELECTION is working as intended. The mappings are chosen based on the environment variable. Hence remove the info logs to reduce clutter


jonathanmetzman (approved):
lgtm

PR #4297, opened 2024-10-08
Optimize get_artifacts_for_build() with regex filtering

aditya-wazir:
Introduces regex-based filtering directly in the API to improve performance and reduce number of calls to ab api server.

### DIFF
--- a/src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
+++ b/src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
@@ -119,10 +119,19 @@ def download_artifact(client, bid, target, attempt_id, name, output_directory,
   return output_path
 
 
-def get_artifacts_for_build(client, bid, target, attempt_id='latest'):
+def get_artifacts_for_build(client, bid, target, attempt_id='latest',
+                            regexp=''):
   """Return list of artifacts for a given build."""
-  request = client.buildartifact().list(
-      buildId=bid, target=target, attemptId=attempt_id)
+  if regexp == '':
+    request = client.buildartifact().list(
+        buildId=bid, target=target, attemptId=attempt_id)
+  else:
+    request = client.buildartifact().list(
+        buildId=bid,
+        target=target,
+        attemptId=attempt_id,
+        nameRegexp=regexp,
+        maxResults=100)
 
   request_str = (f'{request.uri}, {request.method}, '
                  f'{request.body}, {request.methodId}')
@@ -239,7 +248,7 @@ def get(bid, target, regex, output_directory, output_filename=None):
 def run_script(client, bid, target, regex, output_directory, output_filename):
   """Download artifacts as specified."""
   artifacts = get_artifacts_for_build(
-      client=client, bid=bid, target=target, attempt_id='latest')
+      client=client, bid=bid, target=target, attempt_id='latest', regexp=regex)
   if not artifacts:
     logs.error(f'Artifact could not be fetched for target {target}, '
                f'build id {bid}.')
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
```diff
@@ -118,10 +118,19 @@ def download_artifact(client, bid, target, attempt_id, name, output_directory,
...
+  if regexp == '':
+    request = client.buildartifact().list(
+        buildId=bid, target=target, attemptId=attempt_id)
+  else:
+    request = client.buildartifact().list(
+        buildId=bid,
+        target=target,
+        attemptId=attempt_id,
+        nameRegexp=regexp,
+        maxResults=100)
```
jonathanmetzman, line 133: Why is maxResults passed here but not above?

src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
```diff
@@ -118,10 +118,19 @@ def download_artifact(client, bid, target, attempt_id, name, output_directory,
   return output_path
 
 
-def get_artifacts_for_build(client, bid, target, attempt_id='latest'):
+def get_artifacts_for_build(client, bid, target, attempt_id='latest',
+                            regexp=''):
   """Return list of artifacts for a given build."""
-  request = client.buildartifact().list(
-      buildId=bid, target=target, attemptId=attempt_id)
+  if regexp == '':
```
marktefftech, line 124: nit: `if not regexp:` reads better

src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
```diff
@@ -118,10 +118,19 @@ def download_artifact(client, bid, target, attempt_id, name, output_directory,
   return output_path
 
 
-def get_artifacts_for_build(client, bid, target, attempt_id='latest'):
+def get_artifacts_for_build(client, bid, target, attempt_id='latest',
+                            regexp=''):
```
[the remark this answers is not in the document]
aditya-wazir: Done.

src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
aditya-wazir: The maxResults=100 parameter is passed in the else block to ensure that all artifacts matching the provided regular expression are retrieved in a single request. This is because the default maximum result limit of 9 artifacts might be insufficient when using a regex that matches a large number of artifacts.
In earlier approach(if-case) , there were no regex based filtering and we were fetching artifacts name one by one using buildartifact.list() api resulting in more number of api calls. and no need of passing 'maxResults' argument.

src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
aditya-wazir: Done.

aditya-wazir:
> This change looks good.
> 
> @aditya-wazir it would be really interesting to see about adding a test for files located in the `src/clusterfuzz/_internal/platforms/android/` directory. As far as I can tell, there are not any (at least, not in `src/clusterfuzz/_internal/tests/core`).
> 
> Could you please investigate the level of effort to add a new test file for android here? Maybe we could just start with a single `fetch_artifact_test.py` file, and build on the structure thereafter.

Hi Mark,
We will start looking into feasibility of adding tests for src/clusterfuzz/_internal/platforms/android/ directory. We will start with fetch_artifact_test.py and then upload tests for other python components as part of different PR

jonathanmetzman (approved):
lgtm

PR #4298, opened 2024-10-08
Fully qualify autohealing policy URLs.

letitz:
URLs returned by the GCE API are fully-qualified, so we should fully qualify URLs from our configs before comparing them for equality in `manage_vms`.

Bug: https://crbug.com/371829507


vitorguidi (approved):
lgtm. The motivation behind this change is a bit unclear though, is it because instances are always being recreated, since it will always think the auto healing policy has changed?

Edit: hmm I think you meant [b/372055499](b/372055499), that is the problem indeed. LGTM

PR #4300, opened 2024-10-08
Update get_battery_level_and_temperature()

Onkar1300:
Update regex to fetch a valid battery 'level' instead of 'Capacity level' to prevent timeout in wait_until_good_state() of battery.py

### DIFF
--- a/src/clusterfuzz/_internal/platforms/android/battery.py
+++ b/src/clusterfuzz/_internal/platforms/android/battery.py
@@ -40,7 +40,7 @@ def get_battery_level_and_temperature():
   output = adb.run_shell_command(['dumpsys', 'battery'])
 
   # Get battery level.
-  m_battery_level = re.match(r'.*level: (\d+).*', output, re.DOTALL)
+  m_battery_level = re.match(r'.*\n[\t ]*level: (\d+).*', output, re.DOTALL)
   if not m_battery_level:
     logs.error('Error occurred while getting battery status.')
     return None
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/platforms/android/battery.py
```diff
@@ -40,7 +40,7 @@ def get_battery_level_and_temperature():
   output = adb.run_shell_command(['dumpsys', 'battery'])
 
   # Get battery level.
-  m_battery_level = re.match(r'.*level: (\d+).*', output, re.DOTALL)
+  m_battery_level = re.match(r'.*\n[\t ]*level: (\d+).*', output, re.DOTALL)
```
vitorguidi, line 43: Can you write a simple unit test for this, so it becomes more explicit what kind of input we are expecting here? As someone who is not dealing with android on an everyday basis, it is hard to understand:

* Why the previous code broke
* Why this change solves the above

Onkar1300: Sure. I have updated battery_test.py

> * Why the previous code broke
- Due to the recent [change](https://android-review.git.corp.google.com/c/platform/frameworks/base/+/3164137), 'Capacity level', was added in dumpsys battery output. Hence, with the current regex, 'get_battery_level_and_temperature()' fetches 'Capacity level' instead of 'level'.
- This led to indefinite recursion in the [while loop](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/platforms/android/battery.py#L84) since 'Capacity level' has range (-1,5) and it would never pass the [check](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/platforms/android/battery.py#L94).
> * Why this change solves the above
- With the update regex, we are matching for 'level' that is preceded by only spaces/tabs and not characters. This ensures that the correct battery level is fetched.

vitorguidi (approved):
lgtm

PR #4310, opened 2024-10-09
Upgrade pyright

jonathanmetzman:


Pipfile
```diff
@@ -9,6 +9,7 @@ crcmod = "==1.7"
 future = "==0.17.1"
 protobuf = "==4.23.4"
 psutil = "==5.9.4"
+pyright = "==1.1.384"
```
oliverchang, line 12: we should update this on line 27 instead right?

PR #4313, opened 2024-10-10
Improve corpus management

jonathanmetzman:
1. Don't add an extra slash after the directory we are uploading to (e.g. gs://bucket/target-dir//blob). This is a benign but confusing bug.
2. Raise errors when we fail to delete or upload corpus.
3. Name variables better so it's clear what determines whether a corpus element is deleted.
4. Fix b/372698387 "Failed to delete" error by including deletion URLs for quarantine corpus. This is speculative but is definitely an issue. It's unclear why this is only happening on Windows.
5. Log how many files we delete and upload.


vitorguidi (commented):
lgtm

PR #4314, opened 2024-10-11
Allowing id token on oauth for appengine

vitorguidi:
### Motivation

In order to get healthchecks on appengine handlers, we need to use GCP Uptimes. They authenticate through an [oauth id token](https://cloud.google.com/monitoring/uptime-checks#create). As things currently stand, oauth on appengine only supportes access tokens, so this PR solves that.

### Alternatives considered

It would be ideal if we only did one api call, either to validade an id or access token. However, the [specification](https://auth0.com/docs/secure/tokens/id-tokens/id-token-structure) for oauth2 does not present a standard way of, given a token, differentiating between the two. For that reason, two api calls to GCP are made.

Part of #4271

### DIFF
--- a/src/appengine/libs/handler.py
+++ b/src/appengine/libs/handler.py
@@ -161,9 +161,21 @@ def unsupported_on_local_server(func):
   return wrapper
 
 
-def _validate_access_token(authorization):
-  """Validates a JWT as an access or id token, or raises."""
-  access_token = authorization.split(' ')[1]
+def validate_id_token(access_token):
+  """Validates a JWT as an id token."""
+  response_id_token = requests.get(
+      'https://www.googleapis.com/oauth2/v3/tokeninfo',
+      params={'id_token': access_token},
+      timeout=HTTP_GET_TIMEOUT_SECS)
+
+  if response_id_token.status_code == 200:
+    return response_id_token
+
+  return None
+
+
+def validate_access_token(access_token):
+  """Validates a JWT as an access token."""
   response_access_token = requests.get(
       'https://www.googleapis.com/oauth2/v3/tokeninfo',
       params={'access_token': access_token},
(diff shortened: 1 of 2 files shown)


src/appengine/libs/handler.py
```diff
@@ -161,6 +161,27 @@ def wrapper(self, *args, **kwargs):
   return wrapper
 
 
+def _validade_access_token(authorization):
```
oliverchang, line 164: nit: fix spelling of validate

src/appengine/libs/handler.py
```diff
@@ -161,6 +161,27 @@ def wrapper(self, *args, **kwargs):
...
+  access_token = authorization.split(' ')[1]
+  response_access_token = requests.get(
+      'https://www.googleapis.com/oauth2/v3/tokeninfo',
+      params={'access_token': access_token},
+      timeout=HTTP_GET_TIMEOUT_SECS)
+  response_id_token = requests.get(
```
oliverchang, line 170: can we short circuit this if the access token response succeeds? i.e. we don't make the id token request.

src/appengine/libs/handler.py
```diff
@@ -161,6 +161,27 @@ def wrapper(self, *args, **kwargs):
...
+      'https://www.googleapis.com/oauth2/v3/tokeninfo',
+      params={'id_token': access_token},
+      timeout=HTTP_GET_TIMEOUT_SECS)
+  if response_access_token.status_code != 200 and response_id_token:
```
oliverchang, line 174: should this second part be `response_id_token.status_code != 200` ?

oliverchang (changes requested):

oliverchang (approved):

oliverchang (approved):

PR #4315, opened 2024-10-11
Remove support for privileged jobs in batch.

jonathanmetzman:
We don't need to test this anymore.

### DIFF
--- a/src/clusterfuzz/_internal/google_cloud_utils/batch.py
+++ b/src/clusterfuzz/_internal/google_cloud_utils/batch.py
@@ -58,11 +58,6 @@ BatchWorkloadSpec = collections.namedtuple('BatchWorkloadSpec', [
     'machine_type',
 ])
 
-_UNPRIVILEGED_TASKS = {
-    'analyze', 'corpus_pruning', 'symbolize', 'regression', 'variant',
-    'minimize', 'progression'
-}
-
 
 def _create_batch_client_new():
   """Creates a batch client."""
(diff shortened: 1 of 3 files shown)


src/clusterfuzz/_internal/google_cloud_utils/batch.py
```diff
@@ -281,10 +276,6 @@ def _get_spec_from_config(command, job_name):
     config_name += '-PREEMPTIBLE'
   else:
     config_name += '-NONPREEMPTIBLE'
-  # TODO(metzman): Get rid of this when we stop doing privileged operations in
```
jonathanmetzman, line 284: This cannot land before https://team-review.git.corp.google.com/c/gosst/clusterfuzz-config/+/2350609 which removes the UNPRIVILEGED suffix from the unprivileged config, and gets rid of the privileged configs.

src/clusterfuzz/_internal/google_cloud_utils/batch.py
```diff
@@ -281,10 +276,8 @@ def _get_spec_from_config(command, job_name):
...
-  # utasks.
-  if command in _UNPRIVILEGED_TASKS:
-    config_name += '-UNPRIVILEGED'
+  # TODO(metzman): Remove this since all jobs will be unprivileged.
```
jonathanmetzman, line 279: I'm not removing UNPRIVILEGED right away because it could be unsafe if I were to do this without landing the rename in https://team-review.git.corp.google.com/c/gosst/clusterfuzz-config/+/2350609

oliverchang (approved):

letitz:
How does this relate to the recent discussion thread on trusted vs untrusted workers? I think we might want to preserve the ability to schedule tasks on batch with a privileged service account.

jonathanmetzman:
> How does this relate to the recent discussion thread on trusted vs untrusted workers? I think we might want to preserve the ability to schedule tasks on batch with a privileged service account.

It was actually seperate from that discussion (which I didn't totally see when I made this because I was semi-OOO). I'll probably drop this PR.

PR #4316, opened 2024-10-11
[rearchitecture][fuzz_task] Move corpus signing to preprocess.

jonathanmetzman:


jonathanmetzman:
I'm going to fix the test tomorrow. I don't think it's particularly a helpful test for this change. It's only failing because I changed the function signature, and it hasn't uncovered a real bug.

oliverchang (approved):

PR #4321, opened 2024-10-11
Fixing oauth annotation order

vitorguidi:
### Motivation

Uptime healthchecks are breaking for fuzzer/jobs/corpora. This happens because the check_user_access annotation is placed BEFORE the oauth one, which leads to the verification being [asserted](https://github.com/google/clusterfuzz/blob/master/src/appengine/libs/access.py#L89) before credentials are fetched.

This PR fixes the annotation order, making authentication happen before authorization.

Part of #4271


oliverchang (commented):
lgtm

PR #4322, opened 2024-10-11
Adding instrumentation for bugs opened/closed

vitorguidi:
### Motivation

Opening and closing bugs is the last step in the clusterfuzz user journey, and we currently do not have that instrumentation. This PR:

* Enables monitoring in the kubernetes environment
* Enables monitoring on cron jobs
* Moves the wrap_with_monitoring context manager from run_bot.py to monitoring.py, so it can get reused in run_cron.py
* Collects bugs opened metrics from the triage cronjob
* Collects bugs closed metrics from the cleanup cronjob

The only relevant label for these metrics is the fuzzer name.

For bug filing, we measure how many attempts:
* Succeeded
* Failed
* Got throttled

For bug closing, we measure how many attempts:
* Succeeded
* Failed

Part of #4271

### DIFF
--- a/src/clusterfuzz/_internal/cron/triage.py
+++ b/src/clusterfuzz/_internal/cron/triage.py
@@ -30,7 +30,8 @@ from clusterfuzz._internal.issue_management import issue_filer
 from clusterfuzz._internal.issue_management import issue_tracker_policy
 from clusterfuzz._internal.issue_management import issue_tracker_utils
 from clusterfuzz._internal.metrics import crash_stats
-from clusterfuzz._internal.metrics import logs, monitoring_metrics
+from clusterfuzz._internal.metrics import logs
+from clusterfuzz._internal.metrics import monitoring_metrics
 
 from . import grouper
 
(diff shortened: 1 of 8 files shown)


src/clusterfuzz/_internal/cron/triage.py
```diff
@@ -261,6 +261,7 @@ def _file_issue(testcase, issue_tracker, throttler):
 
   if throttler.should_throttle(testcase):
     _add_triage_message(testcase, 'Skipping filing as it is throttled.')
+    monitoring_metrics.ISSUE_FILING_THROTTLED.increment({'fuzzer_name': testcase.fuzzer_name})
```
vitorguidi, line 264: One alternative would be to have a single ISSUE_FILLING metric, and a label to indicate throttled/success/failure, to define less metric types.

src/clusterfuzz/_internal/metrics/monitoring_metrics.py
```diff
@@ -200,3 +200,41 @@
...
+
+ISSUE_FILING_THROTTLED = monitor.CounterMetric(
+    'issues/filing/throttled',
+    description='Bugs opened through triage task.',
+    field_spec=[
+        monitor.StringField('fuzzer_name'),
+        monitor.StringField(''),
+    ])
+
+ISSUE_FILINGS_FAILED = monitor.CounterMetric(
```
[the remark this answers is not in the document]
vitorguidi: No, will fix.

src/clusterfuzz/_internal/cron/triage.py
alhijazi: I think it might be more convenient to have a single issue_filing metric with 3 possible values (success, failure, throttling). Same for ISSUE_CLOSING which can have two values(success, failure)

vitorguidi: Will send a followup one to simplify this.

PR #4329, opened 2024-10-15
Get rsync working correctly

jonathanmetzman:
1. Correctly track which files are preserved, new and deleted by using hashes instead of filenames.
2. Use filenames (with hashes) instead of paths which change when pruning (there is a merge directory).
3. Ensure we aren't using default values from proto's maps.
4. Add tests.

### DIFF
--- a/src/clusterfuzz/_internal/fuzzing/corpus_manager.py
+++ b/src/clusterfuzz/_internal/fuzzing/corpus_manager.py
@@ -93,6 +93,15 @@ def _count_corpus_files(directory):
   return shell.get_directory_file_count(directory)
 
 
+def rename_file_to_sha(filepath, directory=None):
+  if directory is None:
+    directory = os.path.dirname(filepath)
+  sha1sum = utils.file_hash(filepath)
+  new_filepath = os.path.join(directory, sha1sum)
+  shutil.move(filepath, new_filepath)
+  return new_filepath
+
+
 def legalize_filenames(file_paths):
   """Convert the name of every file in |file_paths| a name that is legal on
   Windows. Returns list of legally named files."""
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/fuzzing/corpus_manager.py
```diff
@@ -93,34 +93,43 @@ def _count_corpus_files(directory):
...
-      legally_named.append(new_file_path)
+      # Hash file to get new name since it also lets us get rid of duplicates,
+      # will not cause collisions for different files and makes things more
+      # consistent (since libFuzzer uses hashes).
+      new_filepath = rename_file_to_sha(filepath)
+      legally_named.append(new_filepath)
     except OSError:
-      failed_to_move_files.append((file_path, new_file_path))
-  if failed_to_move_files:
+      failed_to_move_filepaths.append(filepath)
```
vitorguidi, line 127: Can we log this event?

src/clusterfuzz/_internal/fuzzing/corpus_manager.py
```diff
@@ -409,26 +417,32 @@ def rsync_from_disk(self,
...
+        del filenames_to_delete_dict[filename]
+      else:
+        # We only need to upload if it wasn't uploaded already.
+        filepaths_to_upload.append(filepath)
+
+    logs.info('Uploading corpus.')
+    results = self.upload_files(filepaths_to_upload)
+    logs.info('Done uploading corpus.')
+    filenames_to_delete = list(filenames_to_delete_dict.values())
+    assert ((len(filenames_to_delete) != len(
```
vitorguidi, line 438: I do not understand this assertion, what is going on here?

vitorguidi (approved):
lgtm

PR #4330, opened 2024-10-15
Enriching BOT_COUNT metric

vitorguidi:
### Motivation

As a request from the Chrome team, it would be nice to know what:
* OS type
* OS version
* Release (prod/candidate)

a bot corresponds to. This PR implements that.

Part of #4271

### DIFF
--- a/src/clusterfuzz/_internal/system/environment.py
+++ b/src/clusterfuzz/_internal/system/environment.py
@@ -787,7 +787,7 @@ def platform():
   raise ValueError('Unsupported platform "%s".' % sys.platform)
 
 
-def platform_vesion():
+def platform_version():
   """Return the operating system type, unless an override is provided."""
   environment_override = get_value('OS_OVERRIDE')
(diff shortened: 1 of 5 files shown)


src/clusterfuzz/_internal/system/environment.py
```diff
@@ -786,6 +787,14 @@ def platform():
   raise ValueError('Unsupported platform "%s".' % sys.platform)
 
 
+def platform_vesion():
```
jonathanmetzman, line 790: Is this needed? Also it's spelled wrong.

vitorguidi: Chrome folks asked for it, as per design doc. I think it is a good idea to have it, so we can, for instance, figure out what glibc to expect in every environment without having to ask around or read internal config/code

vitorguidi:
> Does this mean that the change wasn't merged?

Yes, it is now being done in #4333

PR #4336, opened 2024-10-17
Adding fuzzing session duration metric

vitorguidi:
### Motivation

We currently lack metrics for fuzzing session duration. This PR adds that as a histogram metric, with granularity by fuzzer, job and platform.

Part of #4271

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
@@ -1852,13 +1852,12 @@ class FuzzingSession:
     self.fuzz_task_output.crash_groups.extend(crash_groups)
 
     fuzzing_session_duration = time.time() - start_time
-    labels = {
-      'fuzzer': self.fuzzer_name,
-      'job': self.job_type,
-      'platform': environment.platform()
-    }
     monitoring_metrics.FUZZING_SESSION_DURATION.add(
-      fuzzing_session_duration, labels)
+        fuzzing_session_duration, {
+            'fuzzer': self.fuzzer_name,
+            'job': self.job_type,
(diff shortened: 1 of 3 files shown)


oliverchang (approved):

jonathanmetzman (approved):
lgtm

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -1698,6 +1698,7 @@ def do_blackbox_fuzzing(self, fuzzer, fuzzer_directory, job_type):
 
   def run(self):
     """Run the fuzzing session."""
+    start_time = time.time()
```
alhijazi, line 1701: LGTM, with one note on this now including fuzzer setup time, I wonder how much time this takes and whether it's more accurate to move the start time to line 1775 just before `do_engine_fuzzing` and `do_blackbox_fuzzing` are called.

alhijazi: You could also add a comment before the `FUZZING_SESSION_DURATION`definition to state that this includes fuzzer setup and data bundles update times.

alhijazi (approved):

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
vitorguidi: I think it is worth measuring the setup time as well, otherwise we would be measuring fuzzing time (which is being tracked in #4356 ) + the time to upload metadata to datastore, which would be dominated by the fuzzing time, and probably make the metric redundant.

Adding the comment as requested

PR #4339, opened 2024-10-18
Instrumenting build retrieval and unpacking times

vitorguidi:
### Motivation

We currently lack metrics for build retrieval and unpacking times. This PR adds that, with granularity by fuzz target and job type.

There are two different implementations for build downloading/unpacking:

- In the Build class, from which RegularBuild, SplitTargetBuild, FuchsiaBuild and SymbolizedBuild inherit the downloading/unpacking behavior
- In the CustomBuild class, which implements its own logic

There are two possible cases for downloading/unpacking: clusterfuzz either downloads the whole build and unpacks it locally, or unpacks it remotely. This is the case for all build types except CustomBuild, which does not perform remote unpacking.

For build retrieval over http, we do not track download time. For all the other cases, it suffices to keep track of start/finish time for download and unpacking.

Finally, a _build_type is added to the constructor of the Build class, from which all other inherit. This is used to track the build type (debug or release), and is only mutated by SymbolizedBuild when attempting to fetch a debug build.

Part of #4271

### DIFF
--- a/src/clusterfuzz/_internal/build_management/build_manager.py
+++ b/src/clusterfuzz/_internal/build_management/build_manager.py
@@ -562,8 +562,7 @@ class Build(BaseBuild):
 
     elapsed_time = time.time() - start_time
     monitoring_metrics.JOB_BUILD_RETRIEVAL_TIME.add(
-        elapsed_time,
-        {
+        elapsed_time, {
             'fuzz_target': self.fuzz_target,
             'job': os.getenv('JOB_TYPE'),
             'platform': environment.platform(),
--- a/src/clusterfuzz/_internal/build_management/build_manager.py
+++ b/src/clusterfuzz/_internal/build_management/build_manager.py
@@ -322,8 +322,7 @@ class Build(BaseBuild):
                base_build_dir,
                revision,
                build_prefix='',
-               fuzz_target=None,
-               job_type=None):
+               fuzz_target=None):
     super().__init__(base_build_dir)
     self.revision = revision
     self.build_prefix = build_prefix
@@ -333,7 +332,6 @@ class Build(BaseBuild):
     # This is used by users of the class to instruct the class which fuzz
     # target to unpack.
     self.fuzz_target = fuzz_target
-    self.job_type = job_type
 
   def _reset_cwd(self):
     """Reset current working directory. Needed to clean up build
@@ -670,8 +668,7 @@ class RegularBuild(Build):
                build_url,
                build_prefix='',
                fuzz_target=None,
-               http_build_url=None,
-               job_type=None):
+               http_build_url=None):
     """RegularBuild constructor. See Build constructor for other parameters.
 
     Args:
@@ -681,7 +678,7 @@ class RegularBuild(Build):
         gs://foo/bar.zip.
     """
     super().__init__(
-        base_build_dir, revision, build_prefix, fuzz_target=fuzz_target, job_type=job_type)
+        base_build_dir, revision, build_prefix, fuzz_target=fuzz_target)
     self.build_url = build_url
     self.http_build_url = http_build_url
 
@@ -1102,7 +1099,7 @@ def _full_fuzz_target_path(bucket_path, fuzz_target):
   return bucket_path.replace('%TARGET%', _base_fuzz_target_name(fuzz_target))
 
 
-def _setup_split_targets_build(bucket_path, fuzz_target, revision=None, job_type=None):
+def _setup_split_targets_build(bucket_path, fuzz_target, revision=None):
   """Set up targets build."""
   bucket_path = environment.get_value('FUZZ_TARGET_BUILD_BUCKET_PATH')
   if not fuzz_target:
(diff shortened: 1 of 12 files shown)


oliverchang (changes requested):
Rather than propagating this variable everywhere, can we just use the `JOB_NAME` env var? i.e. https://github.com/google/clusterfuzz/blob/ff71dd24500c49807840d41d3649cbf9eb53e707/src/clusterfuzz/_internal/build_management/build_manager.py#L83

This should be set in the uworkers at this point right @jonathanmetzman  ?

jonathanmetzman:
> Rather than propagating this variable everywhere, can we just use the `JOB_NAME` env var? i.e.
> 
> https://github.com/google/clusterfuzz/blob/ff71dd24500c49807840d41d3649cbf9eb53e707/src/clusterfuzz/_internal/build_management/build_manager.py#L83
> 
> This should be set in the uworkers at this point right @jonathanmetzman ?

Yes!

jonathanmetzman:
> > 'JOB_NAME'
> 
> > Rather than propagating this variable everywhere, can we just use the `JOB_NAME` env var? i.e.
> > https://github.com/google/clusterfuzz/blob/ff71dd24500c49807840d41d3649cbf9eb53e707/src/clusterfuzz/_internal/build_management/build_manager.py#L83
> > 
> > This should be set in the uworkers at this point right @jonathanmetzman ?
> 
> That simplifies things a lot, thanks for the suggestion!
> 
> It seems like it does propagate to uworkers
> 
> https://github.com/google/clusterfuzz/blob/a1904926747580bd6388b025720f25506e35a491/src/clusterfuzz/_internal/bot/tasks/commands.py#L141

yep env vars definitely propagate to workers.

jonathanmetzman (approved):
lgtm

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -550,6 +561,17 @@ def _unpack_build(self,
       utils.write_data_to_file('', partial_build_file_path)
 
     elapsed_time = time.time() - start_time
+    monitoring_metrics.JOB_BUILD_RETRIEVAL_TIME.add(
+        elapsed_time,
+        {
+            # The concept of a fuzz target does not apply
```
oliverchang, line 567: is this comment correct here? we are setting  `fuzz_target` here.

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -872,6 +897,18 @@ def _unpack_custom_build(self):
                                      build_local_archive):
       return False
 
+    build_download_time = time.time() - download_start_time
+    monitoring_metrics.JOB_BUILD_RETRIEVAL_TIME.add(
+        build_download_time,
+        {
+            # The concept of a fuzz target does not apply
+            # to blackbox fuzzers
```
oliverchang, line 905: nit: it's more that custom builds don't support fuzz targets today, so a better comment would be:

"Custom builds don't support fuzz targets."

(We could probably fix this if needed -- I think we did support this in the past right @jonathanmetzman ?)

oliverchang (approved):
Looks good with some nits.

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -550,6 +561,14 @@ def _unpack_build(self,
       utils.write_data_to_file('', partial_build_file_path)
 
     elapsed_time = time.time() - start_time
+    monitoring_metrics.JOB_BUILD_RETRIEVAL_TIME.add(
```
alhijazi, line 564: It seems that the unpack time here will sometimes also include the download time since we call `_open_build_archive` on line 518, which can lead to a download.

 This is not the same behaviour as for the Custom Builds

src/clusterfuzz/_internal/metrics/monitoring_metrics.py
```diff
@@ -32,6 +32,19 @@
         monitor.BooleanField('bad_build'),
     ])
 
+JOB_BUILD_RETRIEVAL_TIME = monitor.CumulativeDistributionMetric(
+    'task/fuzz/job/build_retrieval_time',
+    bucketer=monitor.FixedWidthBucketer(width=0.05, num_finite_buckets=20),
+    description=('Distribution of fuzz task\'s build retrieval times. '
```
vitorguidi, line 38: This metric name is misleading, this instrumentation applies for all tasks that invoke build_manager.setup_build, fix it.

src/clusterfuzz/_internal/metrics/monitoring_metrics.py
```diff
@@ -32,6 +32,19 @@
...
+                 '(grouped by fuzzer/job)'),
+    field_spec=[
+        monitor.StringField('fuzz_target'),
```
vitorguidi: removing this to avoid cardinality explosion

alhijazi (approved):

PR #4341, opened 2024-10-21
[Monitoring] Adding a build age metric

vitorguidi:
### Motivation

We currently lack awareness on how old builds are during fuzz task. This PR implements that, by making the assumption that the Last Update Time metadata field in GCS is a good proxy for build age. [Documentation reference](https://cloud.google.com/storage/docs/json_api/v1/objects#resource)

### Approach

Symbolized and custom builds do not matter, thus all builds of interest will be fetched from ```build_manager.setup_regular_build```. Logic for collecting all bucket paths and the latest revision was refactored, so that ```setup_regular_build``` can also figure out the latest revision for a given build and conditionally emit the proposed metric.

### Testing strategy

!Todo: test this for fuzz, analyze, progression

Locally ran tasks, with instructions from #4343 and #4345 , and verified the _emmit_build_age_metric function gets invoked and produces sane output. 

Commands used:
```
fuzz libFuzzer libfuzzer_asan_log4j2
```
image

```
progression 4992158360403968 libfuzzer_asan_qt
```
image

```
analyze 4992158360403968 libfuzzer_asan_qt (disclaimer: build revision was overriden mid flight to force a trunk build, since this testcase was already tied to a crash revision)
```
image

Part of #4271


src/clusterfuzz/_internal/metrics/monitoring_metrics.py
```diff
@@ -32,6 +32,18 @@
         monitor.BooleanField('bad_build'),
     ])
 
+JOB_BUILD_AGE = monitor.CumulativeDistributionMetric(
+    'task/fuzz/job/build_age',
+    bucketer=monitor.FixedWidthBucketer(width=0.05, num_finite_buckets=20),
+    description=('Distribution of fuzz task\'s build age. '
+                 '(grouped by fuzzer/job)'),
+    field_spec=[
+        monitor.StringField('fuzz_target'),
```
vitorguidi, line 41: Figure out how to collect this only for fuzz task

jonathanmetzman:
Does this need a review?

vitorguidi:
> Does this need a review?

I am triggering the reviews once the code gets to a good enough state, I will add a WIP tag to the next ones to signal that.

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -404,6 +406,32 @@ def _post_setup_success(self, update_revision=True):
     if instrumented_library_paths:
       self._patch_rpaths(instrumented_library_paths)
 
+  def _emmit_build_age_metric(self, gcs_path):
+    """Emmits a metric to track the age of a build."""
+    try:
+      last_update_time = storage.get(gcs_path).get('updated')
+      if isinstance(last_update_time, str):
+        # storage.get returns two different types for the updated field:
```
oliverchang, line 414: we should fix this instead. Can you file a bug and add a TODO?

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -486,6 +514,7 @@ def _unpack_build(self,
                     http_build_url=None):
     """Unpacks a build from a build url into the build directory."""
     # Track time taken to unpack builds so that it doesn't silently regress.
+    self._emmit_build_age_metric(build_url)
```
oliverchang, line 517: is this correct when we already have line 906?

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -868,9 +897,13 @@ def _unpack_custom_build(self):
...
       storage.copy_file_from(gcs_path, build_local_archive)
-    elif not blobs.read_blob_to_disk(self.custom_binary_key,
+    else:
+      gcs_path = blobs.get_gcs_path(self.custom_binary_key)
+      if not blobs.read_blob_to_disk(self.custom_binary_key,
                                      build_local_archive):
-      return False
+        return False
+
+    self._emmit_build_age_metric(gcs_path)
```
oliverchang, line 906: I don't think this function (`_unpack_build`) is the correct place to track this metric. 

This gets called to legitimately unpack older builds in e.g. regression_task where we are performing bisection to find the first build that introduces a crash. 

We need to do this only in the `setup_trunk_build` path, which is the path for setting up the **latest** build.

oliverchang (changes requested):

src/clusterfuzz/_internal/build_management/build_manager.py
vitorguidi: https://github.com/google/clusterfuzz/issues/4357
Opened and assigned to myself

src/clusterfuzz/_internal/build_management/build_manager.py
vitorguidi: Line 906 only applies for _CustomBuild_, that overrides **setup()** and does not call **_unpack_build()** from the _Build_ class.

All other classes end up calling **_unpack_build()** from the _Build_ class, more specifically through the **setup()** method inherited from _RegularBuild_ (with the exception of _SymbolizedBuild_, that overrides **setup()** but ends up calling the **_unpack_build()** method in the _Build_ class anyway). 

So this does look solid to me.

src/clusterfuzz/_internal/build_management/build_manager.py
vitorguidi: If we do it on setup_trunk_build though, the code path for setup_custom_binary will never get hit, and we will not emmit metrics for chrome fuzzers, right?

setup_trunk_build calls setup_regular_build under the hood:

https://github.com/google/clusterfuzz/blob/24f976051fbe6b02660cc02f9d5c87986b01c981/src/clusterfuzz/_internal/build_management/build_manager.py#L1158

And if we only allow for classes that inherit from RegularBuild, we will not instrument SymbolizedBuild nor CustomBuild, which are legitimate use cases for us, right?

Having said that, if the concern is instrumenting other tasks that legitimately will download old builds, we can use the "task_name" label to filter out for only fuzz task. Wdyt?

src/clusterfuzz/_internal/build_management/build_manager.py
oliverchang: Ack, I misread this. My bad!

src/clusterfuzz/_internal/build_management/build_manager.py
oliverchang: We don't need to track this at all for custom builds. Custom builds are manually uploaded builds with no expectation they'll be updated regularly. They're typically used for one-off experiments. 

For `SymbolizedBuild` etc -- that is not used for fuzzing.

I don't think we should tie the task name to whether or not we want to alert on it or not. That's tying an assumption that's a bit far removed from the actual thing we're checking -- "when we pull what we expect to be the latest builds, are they actually kept up to date?".

re `setup_trunk_build` -- I don't think we should (or can) call emit there, because the full zip url is not constructed there -- it's `setup_regular_build` that has this. We just need to limit the cases when we emit to when we know `setup_regular_build` is setting up the latest. i.e. we might need to just add a boolean there that is set when called from `setup_trunk_build`. Perhaps there's something more elegant here that achieves the same.

vitorguidi: Thanks for the clarification! Will implement that.

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -868,9 +898,13 @@ def _unpack_custom_build(self):
...
       storage.copy_file_from(gcs_path, build_local_archive)
-    elif not blobs.read_blob_to_disk(self.custom_binary_key,
+    else:
+      gcs_path = blobs.get_gcs_path(self.custom_binary_key)
+      if not blobs.read_blob_to_disk(self.custom_binary_key,
                                      build_local_archive):
-      return False
+        return False
+
+    self._emit_build_age_metric(gcs_path)
```
alhijazi, line 907: If I understand correctly, this will be emitted for all tasks.
Tasks like progression and regression that try many versions, starting with the latest to check if the crash occurs or not. I don't think this approach would work there. We rather need to just verify that what we consider the latest version on such job/task is not too old, not every version tried.

As for analyze_task, it was pointed out that the uploader can specify the version they want to test on, in that case, we should not emit a metric.

src/clusterfuzz/_internal/metrics/monitoring_metrics.py
```diff
@@ -32,6 +32,19 @@
         monitor.BooleanField('bad_build'),
     ])
 
+JOB_BUILD_AGE = monitor.CumulativeDistributionMetric(
+    'task/fuzz/job/build_age',
```
alhijazi, line 36: `task/fuzz/job/build_age` makes it sound like this metric is only collected for `fuzz_task`

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -404,6 +406,33 @@ def _post_setup_success(self, update_revision=True):
...
 
+  def _emit_build_age_metric(self, gcs_path):
+    """Emits a metric to track the age of a build."""
+    try:
+      last_update_time = storage.get(gcs_path).get('updated')
+      # TODO(vitorguidi): standardize return type between fs and gcs.
+      if isinstance(last_update_time, str):
+        # storage.get returns two different types for the updated field:
+        # the gcs api returns string, and the local filesystem implementation
+        # returns a datetime.datetime object normalized for UTC
```
jonathanmetzman, line 417: nit: end with period.

src/clusterfuzz/_internal/build_management/build_manager.py
vitorguidi: Another not on this, it is possible that setup_regular_build gets invoked with the latest revision, coming from some code path other than setup_trunk_build.

setup_regular_build is indeed the correct place to emit this metric, IFF the revision in question is the latest.

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -1199,8 +1202,64 @@ def _get_latest_revision(bucket_paths):
...
+    return None
+  build_url = revisions.find_build_url(bucket_path, build_urls, revision)
+  if not build_url:
+    logs.error(
+        'Error getting build url for job %s (r%d).' % (job_type, revision))
+    return None
+  return build_url
+
+
+def _get_build_bucket_paths():
```
oliverchang, line 1246: nit: add docstring

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -1199,8 +1202,64 @@ def _get_latest_revision(bucket_paths):
...
+      bucket_paths.append(bucket_path)
+    else:
+      logs.info('Bucket path not found for %s' % env_var)
+  return bucket_paths
+
+
+def setup_trunk_build(fuzz_target, build_prefix=None):
   """Sets up latest trunk build."""
+  bucket_paths = _get_build_bucket_paths()
+  if len(bucket_paths) == 0:
```
oliverchang, line 1260: nit: just `if not bucket_paths` per https://google.github.io/styleguide/pyguide.html#214-truefalse-evaluations

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -1221,24 +1280,24 @@ def setup_trunk_build(bucket_paths, fuzz_target, build_prefix=None):
...
-  build_url = revisions.find_build_url(bucket_path, build_urls, revision)
-  if not build_url:
-    logs.error(
-        'Error getting build url for job %s (r%d).' % (job_type, revision))
+  build_url = _get_build_url(bucket_path, revision, job_type)
 
+  if not build_url:
     return None
 
+  all_bucket_paths = _get_build_bucket_paths()
```
oliverchang, line 1295: nit: rather than recompute this stuff here, can we just pass a `is_latest` argument to `setup_regular_build` (which defaults to false)?

oliverchang (approved):

src/clusterfuzz/_internal/build_management/build_manager.py
vitorguidi: It is possible that a latest build gets fetched without going through the setup_trunk_build codepath, so I don't think that would work.

oliverchang: Yes, it is possible, but I don't think we need to care about them? 

We really only care about fuzz_task here (which calls setup_trunk_build). Yes, there are some cases where another task will happen to setup the latest build, but I don't think we need to track those?

vitorguidi: The design doc states that some non fuzz taks are relevant, so maybe these non-trunk cases are relevant. @alhijazi any thoughts on this?

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -900,9 +901,11 @@ def _unpack_custom_build(self):
         os.makedirs(directory)
       gcs_path = f'/{custom_builds_bucket}/{self.custom_binary_key}'
       storage.copy_file_from(gcs_path, build_local_archive)
-    elif not blobs.read_blob_to_disk(self.custom_binary_key,
+    else:
+      gcs_path = blobs.get_gcs_path(self.custom_binary_key)
```
alhijazi, line 905: This line doesn't seem to be used.

src/clusterfuzz/_internal/build_management/build_manager.py
alhijazi: For tasks like progression, we care about running the latest build as well to be able to tell if an issue was fixed or not. This impacts the time between a fix being introduced and clusterfuzz marking an issue as fixed.
Same for analyze, we would like to know if a given issue reproduces at head.

alhijazi (approved):
LGTM % comment

PR #4344, opened 2024-10-21
[batch] Support executing fuzz task

jonathanmetzman:
This is needed for local debugging.
It won't affect production yet.


vitorguidi (approved):
lgtm, since it does not touch prod

PR #4346, opened 2024-10-21
Add a task id for easy tracking of tasks. 

jonathanmetzman:
This should make it easy to get the logs for the
preprocess, postprocess, and main parts of a task.


vitorguidi:
We could expand on this idea in the future, to get some form of tracing over the lifecycle of entities within clusterfuzz, for better observability.

PR #4354, opened 2024-10-23
Remove incorrect assertion.

jonathanmetzman:
I now understand what .get() does when a query returns more than one result, it gets an arbitrary one, it doesn't error. So don't error here.

Fixes: https://pantheon.corp.google.com/errors/detail/CKjxheXi1cOF-AE;filter=%5B%5D;time=PT1H;locations=global?e=-13802955&invt=AbfeBw&mods=logs_tg_prod&project=google.com:clusterfuzz


oliverchang (commented):
lgtm

PR #4356, opened 2024-10-23
Tracking fuzzing time on engine fuzzers

vitorguidi:
### Motivation

Clusterfuzz only tracks fuzzing time for blackbox fuzzers at the moment, this PR extends the tracking to engine fuzzers as well by emmiting the JOB_TOTAL_FUZZ_TIME and FUZZER_TOTAL_FUZZ_TIME metrics.

Since all engine fuzzing is single process/single threaded, it suffices to track start and end time for each test case run. The only difference in behavior is that only libfuzzer indicates a timeout, and thus all other engines are expected to concentrate their metrics on timeout=False.

### Testing strategy

Ran a fuzz task locally and verified the code path for _TrackFuzzTime is reached and produces sane output.
Command used:
```
fuzz libFuzzer libfuzzer_asan_log4j2
```
image

Part of #4271

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
@@ -1496,12 +1496,12 @@ class FuzzingSession:
       logs.info(f'Fuzzing round {fuzzing_round}.')
       try:
         with _TrackFuzzTime(self.fully_qualified_fuzzer_name,
-                          self.job_type) as tracker:
-          result, current_fuzzer_metadata, fuzzing_strategies = run_engine_fuzzer(
+                            self.job_type) as tracker:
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -1507,7 +1511,7 @@ def do_engine_fuzzing(self, engine_impl):
         logs.error(f'{self.fuzz_target.binary} is not in the build.')
         return [], {}
 
-      fuzzer_metadata.update(current_fuzzer_metadata)
+      fuzzer_metadata.update(cur_fuzzer_metadata)
```
vitorguidi, line 1514: This is to avoid linting failure due to a line too long

oliverchang (approved):

PR #4365, opened 2024-10-31
Add rate limiting

jonathanmetzman:


src/clusterfuzz/_internal/base/task_rate_limiting.py
```diff
@@ -0,0 +1,95 @@
...
+    self.task_argument = task_argument
+    self.job_name = job_name
+
+  @property
+  def readable_name(self):
+    return ' '.join([self.task_name, self.task_argument, self.job_name])
+
+  def record_task(self, success: bool) -> None:
+    """Records a task and whether it completed succesfully."""
+    if self.task_name in {'uworker_main', 'postprocess', 'preprocess'}:
```
oliverchang, line 45: can you put these into a constant somewhere? 

they're also duplicated on line 61.

src/clusterfuzz/_internal/base/task_rate_limiting.py
```diff
@@ -0,0 +1,95 @@
...
+  TASK_RATE_LIMIT_MAX_ERRORS = 100
+  TASK_RATE_LIMIT_MAX_COMPLETIONS = 2000
+
+  def __init__(self, task_name, task_argument, job_name):
+    self.task_name = task_name
+    self.task_argument = task_argument
+    self.job_name = job_name
+
+  @property
+  def readable_name(self):
```
oliverchang, line 40: how about just `__str__` ?

src/clusterfuzz/_internal/datastore/data_types.py
```diff
@@ -1114,6 +1114,18 @@ class TaskStatus(Model):
   time = ndb.DateTimeProperty()
 
 
+class WindowRateLimitTask(Model):
```
oliverchang, line 1117: Did you consider using redis for this? What would be the considerations that led to choosing datastore here? maintenance cost?

oliverchang, line 1117: will these have asome kind of TTL set to auto cleanup?

src/clusterfuzz/_internal/base/task_rate_limiting.py
```diff
@@ -0,0 +1,95 @@
...
+from clusterfuzz._internal.datastore import ndb_utils
+from clusterfuzz._internal.metrics import logs
+from clusterfuzz._internal.system import environment
+
+
+def _get_datetime_now():
+  return datetime.datetime.now()
+
+
+class TaskRateLimiter:
```
oliverchang, line 28: Can you add some brief comments to explain this class and the general rate limiting logic?

src/clusterfuzz/_internal/base/task_rate_limiting.py
```diff
@@ -0,0 +1,95 @@
+# Copyright 2024 Google LLC
```
oliverchang, line 1: Would it make more sense to place this in `base/tasks/rate_limiter.py` ?

src/clusterfuzz/_internal/base/task_rate_limiting.py
```diff
@@ -0,0 +1,95 @@
...
+  def is_rate_limited(self) -> bool:
+    """Checks if the given task is rate limited."""
+    if self.task_name in {'uworker_main', 'postprocess', 'preprocess'}:
+      # Don't rate limit these fake tasks.
+      return False
+    if environment.get_value('COMMAND_OVERRIDE'):
+      # A user wants to run this task.
+      return False
+    window_start = _get_datetime_now() - self.TASK_RATE_LIMIT_WINDOW
+    query = data_types.WindowRateLimitTask.query(
```
oliverchang, line 68: should we set `keys_only` here to make this more efficient? we don't care about the contents right?

src/clusterfuzz/_internal/datastore/data_types.py
```diff
@@ -1114,6 +1114,18 @@ class TaskStatus(Model):
...
+  """Records the completion of a task. This cannot be part of TaskStatus
+  because it will have a different lifecycle (it's not needed after the window
+  completes)."""
+  # TODO(metzman): Consider using task_id.
+  timestamp = ndb.DateTimeProperty(auto_now_add=True)
```
oliverchang, line 1122: do we need to add a index for querying on these?

src/clusterfuzz/_internal/bot/tasks/commands.py
```diff
@@ -217,23 +214,32 @@ def run_command(task_name,
...
   except errors.InvalidTestcaseError:
     # It is difficult to try to handle the case where a test case is deleted
     # during processing. Rather than trying to catch by checking every point
     # where a test case is reloaded from the datastore, just abort the task.
     logs.warning('Test case %s no longer exists.' % task_argument)
+    rate_limiter.record_task(success=False)
   except BaseException:
     # On any other exceptions, update state to reflect error and re-raise.
     if should_update_task_status(task_name):
       data_handler.update_task_status(task_state_name,
```
oliverchang, line 237: should we put the `rate_limiter.record_task` call in `update_task_status` instead?

oliverchang (commented):
nice!

src/clusterfuzz/_internal/datastore/data_types.py
jonathanmetzman: I think there probably won't be much of a difference in practice between the two approaches. I think Redis might be a bit better for the short term (6 hour) rate limitng but probably worse for the 60-day rate limiting. WDYT?

jonathanmetzman: Yes. I've mentioned this in the docstring now.

src/clusterfuzz/_internal/base/task_rate_limiting.py
jonathanmetzman: Nope, we need the status.

src/clusterfuzz/_internal/bot/tasks/commands.py
```diff
@@ -211,23 +208,32 @@ def run_command(task_name,
       raise AlreadyRunningError
 
   result = None
+  rate_limiter = task_rate_limiting.TaskRateLimiter(task_name, task_argument,
+                                                    job_name)
+  if rate_limiter.is_rate_limited():
+    logs.error(f'Rate limited task: {task_name} {task_argument} {job_name}')
+    if task_name == 'fuzz':
+      # Wait 10 seconds. We don't want to try again immediately because if we
```
oliverchang, line 216: nit: replace 10 with FAIL_WAIT ?

oliverchang (approved):

PR #4367, opened 2024-10-31
Fix issue dealing with lack of fuzz target picked by preprocess

jonathanmetzman:
Handle this gracefully as it's needed for jobs.


oliverchang (commented):
lgtm

PR #4373, opened 2024-11-01
update google issue query to handle field searches

pgrace-google:


src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
```diff
@@ -1122,7 +1122,11 @@ def _parse_datetime(date_string):
 
 def _get_query(keywords, only_open):
   """Gets a search query."""
-  query = ' '.join('"{}"'.format(keyword) for keyword in keywords)
+  string_keywords = [word for word in keywords if ':' not in word]
```
jonathanmetzman, line 1125: Can we make this function accept two kinds of keywords so we don't need this logic in this function?

pgrace-google: you know what i love that

pgrace-google: nvm i take it back - that means we would have to adjust the interface function for find_issues and find_issues_url

but we also would only really have to update the google_internal_tracker function implementations so i guess its not that bad

pgrace-google: ok nvm lint is failing i gotta update monorail and jira too but now im feeling the sunk cost fallacy so im gonna update it anyway

jonathanmetzman: Feel free to disable (using pylint) or delete anything in monorail that's failing. It's totally dead code.

jonathanmetzman (approved):
lgtm

PR #4374, opened 2024-11-01
Refactor fuzzingsession

jonathanmetzman:
Start preparing to seperate code dealing with engine and non-engine fuzzing.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
@@ -1724,6 +1724,7 @@ class FuzzingSession:
             testcases_metadata, crashes)
 
   def run_common(self):
+    """Prepare to run either blackbox or engine fuzzing."""
     # Update LSAN local blacklist with global blacklist.
     global_blacklisted_functions = (
         self.uworker_input.fuzz_task_input.global_blacklisted_functions)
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -1737,11 +1743,21 @@ def run(self):
...
 
     self.testcase_directory = environment.get_value('FUZZ_INPUTS')
+    return None
 
+  def run(self):
+    """Run the fuzzing session."""
+    start_time = time.time()
+    result = self._run_common()
+    if result:
+      return result
```
vitorguidi, line 1757: Probably worth moving the metric from #4336 to a decorator as you pointed out, we are losing instrumentation in this early return.

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -1737,11 +1743,21 @@ def run(self):
...
+
+    # TODO(metzman): Move this section into seperate functions for engine and
+    # blackbox fuzzing instead of sandwhiching with ifs for engine fuzzing.
```
vitorguidi, line 1760: Is this being done at this PR, or in a follow up one?

jonathanmetzman: A following one. That's generally what TODOs mean.

PR #4377, opened 2024-11-05
Gate async HTTP on a config

jonathanmetzman:
The async HTTP usage has been shown to cause OOMs in OSS-Fuzz that cause significant problems,
in particular causing NTP to fail to sync.

### DIFF
--- a/src/clusterfuzz/_internal/google_cloud_utils/storage.py
+++ b/src/clusterfuzz/_internal/google_cloud_utils/storage.py
@@ -38,6 +38,7 @@ from clusterfuzz._internal.base import utils
 from clusterfuzz._internal.config import local_config
 from clusterfuzz._internal.metrics import logs
 from clusterfuzz._internal.system import environment
+from clusterfuzz._internal.system import fast_http
 from clusterfuzz._internal.system import shell
 
 from . import credentials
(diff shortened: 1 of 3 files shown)


src/clusterfuzz/_internal/google_cloud_utils/storage.py
```diff
@@ -1153,6 +1158,17 @@ def blobs_bucket():
   return local_config.ProjectConfig().get('blobs.bucket')
 
 
+def use_async_http():
+  global _USES_ASYNC_HTTP
```
oliverchang, line 1162: nit: just use `memoize` here. 

e.g. https://github.com/google/clusterfuzz/blob/2d57351ad8e22597a94849db6e013b12d9dcdc3c/src/clusterfuzz/_internal/base/utils.py#L904

oliverchang (approved):

src/clusterfuzz/_internal/google_cloud_utils/storage.py
jonathanmetzman: I couldn't figure out if it works with no arguments and it seems a little less readable to be honest. I'll switch it if you think it works.

oliverchang: it should work -- the example linked has no arguments.

vitorguidi (approved):
lgtm

PR #4378, opened 2024-11-06
Fix OSS-fuzz builds status reporting.

oliverchang:
There were some hardcoded constants that needed to be updated for the Google Issue Tracker migration.

We'll also stop adding the 'Project' custom field for now, because it's hard to propagate custom field values and it's not essential for build failure issues.

This should hopefully fix
https://github.com/google/oss-fuzz/issues/12536#issuecomment-2447332796.


oliverchang:
TBR

PR #4379, opened 2024-11-06
google_issue_tracker: Set default priority.

oliverchang:
Otherwise new issues may error out because it's a required field.


jonathanmetzman (commented):
lgtm

PR #4380, opened 2024-11-06
add logs at the end of unpacking

paulsemel:
We are currently logging progression when unpacking, but we're missing the last log. This can be useful to monitor what's going on in CF.


jonathanmetzman (approved):
lgtm

PR #4381, opened 2024-11-06
[Monitoring] Collecting a metric for the age of untriaged testcases

vitorguidi:
### Motivation

Once a testcase is generated (or manually uploaded), followup tasks (analyze/progression) are started. This happens by publishing to a pubsub queue, both for the manually uploaded case, and for the fuzzer generated case.

If for any reason the messages are not processed, the testcase gets stuck. To get better visibility into these stuck testcases, the UNTRIAGED_TESTCASE_AGE metric is introduced, to pinpoint how old these testcases that have not yet been triaged are(more precisely, gone through analyze/regression/impact/progression tasks).

### Attention points

Testcase.timestamp mutates in analyze task:

https://github.com/google/clusterfuzz/blob/6ed80851ad0f6f624c5b232b0460c405f0a018b5/src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py#L589

 This makes it unreliable as a source of truth for testcase creation time. To circumvent that, a new ```created``` field is added to the Testcase entity, from which we can derive the correct creation time.

Since this new field will only apply for testcases created after this PR is merged, Testcase.timestamp will be used instead to calculate the testcase age when the new field is missing.

### Testing strategy

Ran the triage cron locally, and verified the codepath for the metric is hit and produces sane output (reference testcase: 4505741036158976).
image

Part of #4271

### DIFF
--- a/src/clusterfuzz/_internal/cron/triage.py
+++ b/src/clusterfuzz/_internal/cron/triage.py
@@ -298,21 +298,15 @@ def _file_issue(testcase, issue_tracker, throttler):
   return filed
 
 
-def _emmit_untriaged_testcase_age_metric(critical_tasks_completed: bool,
-                                         testcase: data_types.Testcase):
+def _emit_untriaged_testcase_age_metric(critical_tasks_completed: bool,
+                                        testcase: data_types.Testcase):
   """Emmits a metric to track age of untriaged testcases."""
   if critical_tasks_completed:
     return
   if not testcase.timestamp:
     return
   current_time = datetime.datetime.utcnow()
-  # testcase.timestamp is mutated in analyze task.
-  # taking the minimum between these two timestamps
-  # avoids the need to backfill creation_timestamp
-  testcase_creation_time = testcase.timestamp
-  if testcase.creation_timestamp:
-    testcase_creation_time = testcase.creation_timestamp
-  testcase_age = current_time - testcase_creation_time
+  testcase_age = current_time - testcase.get_created_time()
   testcase_age = testcase_age.total_seconds()
   monitoring_metrics.UNTRIAGED_TESTCASE_AGE.add(
       testcase_age,
@@ -363,7 +357,7 @@ def main():
       continue
 
     # Emmit the metric for testcases that should be triaged.
-    _emmit_untriaged_testcase_age_metric(critical_tasks_completed, testcase)
+    _emit_untriaged_testcase_age_metric(critical_tasks_completed, testcase)
 
     # Skip if we are running progression task at this time.
     if testcase.get_metadata('progression_pending'):
--- a/src/clusterfuzz/_internal/cron/triage.py
+++ b/src/clusterfuzz/_internal/cron/triage.py
@@ -298,6 +298,26 @@ def _file_issue(testcase, issue_tracker, throttler):
   return filed
 
 
+def _emmit_untriaged_testcase_age_metric(
+    critical_tasks_completed: bool,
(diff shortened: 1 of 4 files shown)


src/clusterfuzz/_internal/cron/triage.py
```diff
@@ -298,6 +298,30 @@ def _file_issue(testcase, issue_tracker, throttler):
   return filed
 
 
+def _emmit_untriaged_testcase_age_metric(critical_tasks_completed: bool,
+                                         testcase: data_types.Testcase):
+  """Emmits a metric to track age of untriaged testcases."""
```
oliverchang, line 303: nit: emit is the correct spelling.

src/clusterfuzz/_internal/datastore/data_types.py
```diff
@@ -423,6 +423,9 @@ class Testcase(Model):
   # Timestamp.
   timestamp = ndb.DateTimeProperty()
 
+  # Original creation timestamp.
+  creation_timestamp = ndb.DateTimeProperty()
```
oliverchang, line 427: nit: add `indexed=False` here, unless we plan on querying on this property. 

Also, I'd rename this to `created`

src/clusterfuzz/_internal/datastore/data_types.py
```diff
@@ -423,6 +423,9 @@ class Testcase(Model):
   # Timestamp.
   timestamp = ndb.DateTimeProperty()
 
+  # Original creation timestamp.
```
oliverchang, line 426: Can we add some comments to the existing `timestamp` property above? To make it clear how this is different? 

I think something like

"# The time when this Testcase is considered valid. This is usually when the Testcase is created in Datastore when created in e.g. `fuzz_task`. However, there are also cases where the `Testcase` is a placeholder (e.g. awaiting `analyze`), which means this timestamp will reflect when `analyze` completes its analysis).

src/clusterfuzz/_internal/cron/triage.py
```diff
@@ -298,6 +298,30 @@ def _file_issue(testcase, issue_tracker, throttler):
...
+  if critical_tasks_completed:
+    return
+  if not testcase.timestamp:
+    return
+  current_time = datetime.datetime.utcnow()
+  # testcase.timestamp is mutated in analyze task.
+  # taking the minimum between these two timestamps
+  # avoids the need to backfill creation_timestamp
+  testcase_creation_time = testcase.timestamp
+  if testcase.creation_timestamp:
```
oliverchang, line 313: does this `if` exist purely for historical testcases?  

Instead of having this `if` here forever, I'd recommend just adding a `get_created_time()` (or similarly named) method to `Testcase` to perform this logic, with some comments explaining this.

oliverchang (commented):
nice!

jonathanmetzman (approved):
lgtm

src/clusterfuzz/_internal/metrics/monitoring_metrics.py
```diff
@@ -273,6 +273,17 @@
         monitor.StringField('status'),
     ])
 
+UNTRIAGED_TESTCASE_AGE = monitor.CumulativeDistributionMetric(
+    'issues/untriaged_testcase_age',
+    description='Age of testcases that were not yet triaged '
+    '(have not yet completed analyze, progression,'
```
alhijazi, line 279: Progression does not seem to be part of critical tasks, you probably meant regression here.

PR #4383, opened 2024-11-06
Log pid even better

jonathanmetzman:


jonathanmetzman:
So far it looks like the python process is not starting from run_bot. This might help figure what it's actually running before it causes the issue.

PR #4385, opened 2024-11-07
Add a field for TTL in WindowTaskRateLimit.

jonathanmetzman:


jonathanmetzman:
It's questionable whether we should have two fields when having an expiry field is sufficient. But I think it's less understandable for a human to understand what the entity means if it's only timestamp is 6 hours from when it completed instead of when it completed.
datastore TTL needs the expiry timestamp field. So going with both seems best.

jonathanmetzman:
Test failure is spurious.

PR #4386, opened 2024-11-07
[rate limitng] Record errors properly.

jonathanmetzman:
We can't record an error after an exception is raised. Record it first.


oliverchang (commented):
LGTM

PR #4393, opened 2024-11-08
build_manager: do not fetch fuzz_targets when not necessary

paulsemel:
For non engine fuzzers, we do not need to have the full list of fuzz targets, since those builds are only containing the APP_NAME. For that reason, and when the build is fully unpacked, lazily fetch the fuzzing targets only when requested by the user of the class.

This change will tremendously speed up the unpacking step for some of our fuzzers, see https://docs.google.com/document/d/1OepfVcuG2XNXLxgZIXVwgE-uOhX9RO5RfH0TSg_wmPU/.


jonathanmetzman:
> For non engine fuzzers, we do not need to have the full list of fuzz targets, since those builds are only containing the APP_NAME. For that reason, and when the build is fully unpacked, lazily fetch the fuzzing targets only when requested by the user of the class.

What do you mean by fuzz targets in this context?
Are chrome or content shell fuzz targets here? they usually aren't referred to like this.

> 
> This change will tremendously speed up the unpacking step for some of our fuzzers, see https://docs.google.com/document/d/1OepfVcuG2XNXLxgZIXVwgE-uOhX9RO5RfH0TSg_wmPU/.

Nice!

I'm pretty nervous about this change though. The code here is pretty fragile and breaking it could completely break fuzzing in a non-obvious way (by not recording which fuzzers are in a build, the next build can't pick them).

jonathanmetzman:
I think to deploy this we will need to do partial deployments across all our instances of CF (since they each have their own way of dealing with builds)

jonathanmetzman:
WIll deploy this during US working hours

PR #4396, opened 2024-11-11
Give corpus pruning enough time to complete

jonathanmetzman:


src/clusterfuzz/_internal/google_cloud_utils/batch.py
```diff
@@ -33,7 +33,8 @@
 
 _local = threading.local()
 
-MAX_DURATION = f'{60 * 60 * 6}s'
+DEFAULT_MAX_RUN_DURATION = f'{6 * 60 * 60}s'
+CORPUS_PRUNING_MAX_RUN_DURATION = f'{24 * 60 * 60}s'
```
oliverchang, line 37: Can you just use https://github.com/google/clusterfuzz/blob/597d1e15ad665af8fdc82c88a11ff97cfbbc59a0/src/clusterfuzz/_internal/base/tasks/__init__.py#L56 as the source of truth instead?

oliverchang (changes requested):

oliverchang (approved):

PR #4397, opened 2024-11-12
Start to schedule fuzz tasks on batch in OSS-Fuzz

jonathanmetzman:
Start to schedule fuzz tasks on batch in OSS-Fuzz

The scheduler will work differenly in OSS-Fuzz and Chrome. This only implements and OSS-Fuzz version.
This version will use job and project weights to decide which fuzzing jobs to schedule. It then adds these tasks
to the queue for other bots to preprocess and then for the utask_main_scheduler to actually schedule on batch.
For now, we will only do this for 100 CPUs.
1. Add a cron job to run the scheduler every 15 minutes.
2. Improve region handling in batch (still far from complete).
3. Add function for bulk adding of tasks to queue for use by scheduler.
4. Make fuzz tasks less of a priority on batch than others.

### DIFF
--- a/configs/test/batch/batch.yaml
+++ b/configs/test/batch/batch.yaml
@@ -70,4 +70,4 @@ project: 'test-clusterfuzz'
 # TODO(metzman): Come up with a better system than this. I think a system where
 # we have a list of zones and associated config (such as subnets) to pick where
 # to launch tasks is ideal.
-zone: 'us-central1-a'
+region: 'us-central1-a'
--- a/configs/test/batch/batch.yaml
+++ b/configs/test/batch/batch.yaml
@@ -67,3 +67,7 @@ mapping:
     preemptible: true
     machine_type: n1-standard-1
 project: 'test-clusterfuzz'
+# TODO(metzman): Come up with a better system than this. I think a system where
+# we have a list of zones and associated config (such as subnets) to pick where
+# to launch tasks is ideal.
+zone: 'us-central1-a'
--- a/configs/test/batch/batch.yaml
+++ b/configs/test/batch/batch.yaml
@@ -69,4 +69,4 @@ project: 'test-clusterfuzz'
 # TODO(metzman): Come up with a better system than this. I think a system where
 # we have a list of zones and associated config (such as subnets) to pick where
 # to launch tasks is ideal.
-region: 'us-central1-a'
+region: 'us-central1'
--- a/configs/test/batch/batch.yaml
+++ b/configs/test/batch/batch.yaml
@@ -20,7 +20,6 @@ mapping:
     disk_size_gb: 110
     disk_type: pd-standard
     service_account_email: test-clusterfuzz-service-account-email
-    subnetwork: null
     gce_region: 'gce-region'
     gce_zone: 'gce-zone'
     network: 'projects/google.com:clusterfuzz/global/networks/networkname'
--- a/infra/k8s/schedule-fuzz.yaml
+++ b/infra/k8s/schedule-fuzz.yaml
@@ -1,0 +1,41 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: backup
+spec:
+  schedule: "*/20 * * * *"
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    spec:
+      activeDeadlineSeconds: 1200  # 20 minutes.
+      template:
+        spec: 
+          containers:
+          - name: backup
+            image: gcr.io/clusterfuzz-images/base:091c6c2-202409251610
+            imagePullPolicy: Always
+            env:
+            - name: CLUSTERFUZZ_RELEASE
+              value: "prod"
+            - name: RUN_CMD
+              value: "python3.11 $ROOT_DIR/src/python/bot/startup/run_cron.py schedule_fuzz"
+            - name: IS_K8S_ENV
+              value: "true"
+            - name: DISABLE_MOUNTS
+              value: "true"
+          restartPolicy: OnFailure
+      backoffLimit: 3
--- a/src/clusterfuzz/_internal/base/tasks/__init__.py
+++ b/src/clusterfuzz/_internal/base/tasks/__init__.py
@@ -609,10 +609,8 @@ def add_utask_main(command, input_url, job_type, wait_time=None):
       extra_info={'initial_command': initial_command})
 
 
-def bulk_add_tasks(
-    tasks,
-    queue=None,
-    eta_now=False):
+def bulk_add_tasks(tasks, queue=None, eta_now=False):
+  """Adds |tasks| in bulk to |queue|."""
 
   # Old testcases may pass in queue=None explicitly, so we must check this here.
   if queue is None:
@@ -628,17 +626,17 @@ def bulk_add_tasks(
     for task in tasks:
       task.eta = now
 
-  pubsub.PubSubClient()
+  pubsub_client = pubsub.PubSubClient()
+  pubsub_messages = [task.to_pubsub_message() for task in tasks]
   pubsub_client.publish(
-      pubsub.topic_name(utils.get_application_id(), queue),
-      [task.to_pubsub_message() for task in tasks])
-
+      pubsub.topic_name(utils.get_application_id(), queue), pubsub_messages)
 
 
 def add_task(command,
              argument,
              job_type,
              queue=None,
+             wait_time=None,
              extra_info=None):
   """Add a new task to the job queue."""
   if wait_time is None:
--- a/src/clusterfuzz/_internal/base/tasks/__init__.py
+++ b/src/clusterfuzz/_internal/base/tasks/__init__.py
@@ -609,18 +609,38 @@ def add_utask_main(command, input_url, job_type, wait_time=None):
       extra_info={'initial_command': initial_command})
 
 
+def bulk_add_tasks(
+    tasks,
+    queue=None,
+    eta_now=False):
+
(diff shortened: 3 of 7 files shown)


jonathanmetzman:
This implementation is bad, I think the job weights are influenced by the project weights. We always run ffmpeg.

src/clusterfuzz/_internal/base/concurrency.py
```diff
@@ -0,0 +1,31 @@
...
+import multiprocessing
+
+from clusterfuzz._internal.system import environment
+
+POOL_SIZE = multiprocessing.cpu_count()
+
+
+@contextlib.contextmanager
+def make_pool(pool_size=POOL_SIZE):
+  # Don't use processes on Windows and unittests to avoid hangs.
```
oliverchang, line 26: Do you know why we hang? is this because of the start method for multiprocessing being "spawn" on windows? or something else?

infra/k8s/schedule-fuzz.yaml
```diff
@@ -0,0 +1,41 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: backup
```
oliverchang, line 18: can we update the names from "backup" everywhere?

src/clusterfuzz/_internal/base/tasks/__init__.py
```diff
@@ -368,14 +368,19 @@ def __init__(self,
                eta=None,
                is_command_override=False,
                high_end=False,
-               extra_info=None):
+               extra_info=None,
+               is_from_queue=False):
```
oliverchang, line 372: is this also for testing only? this seems like a rather confusing flag to add here here. 

Can you also explain a bit why you need this and how it's helpful for testing?

oliverchang (commented):
nice! some initial questions/comments

infra/k8s/schedule-fuzz.yaml
jonathanmetzman: Ohhh good find.

src/clusterfuzz/_internal/base/tasks/__init__.py
jonathanmetzman: It's for testing fuzzing on batch in prod.
Basically, this PR adds a fuzz task scheduler. But we need a mechanism to ensure these fuzz tasks get scheduled on batch instead of just run by regular bots.
This is only needed for that purpose and I will remove it when we have moved all of fuzzing to batch.

src/clusterfuzz/_internal/base/concurrency.py
jonathanmetzman: I'm not sure to be honest. The comment isn't really detailed enough for me to know exactly what happened. My mistake.

infra/k8s/schedule-fuzz.yaml
```diff
@@ -0,0 +1,41 @@
...
+spec:
+  schedule: "*/30 * * * *"
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    spec:
+      activeDeadlineSeconds: 1200  # 20 minutes.
+      template:
+        spec:
+          containers:
+          - name: backup
```
oliverchang, line 28: you need to change this also.

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -0,0 +1,227 @@
...
+      break
+    if quota['metric'] == 'CPUS':
+      cpu_quota = quota
+      continue
+    if quota['metric'] == 'PREEMPTIBLE_CPUS':
+      preemptible_quota = quota
+      continue
+  assert preemptible_quota or cpu_quota
+
+  if not preemptible_quota['limit']:
```
oliverchang, line 60: can you explain in a comment how this logic works? Why can we switch between preemptilbe and nonpreemptible quota?

configs/test/batch/batch.yaml
```diff
@@ -67,3 +66,7 @@ mapping:
     preemptible: true
     machine_type: n1-standard-1
 project: 'test-clusterfuzz'
+# TODO(metzman): Come up with a better system than this. I think a system where
+# we have a list of zones and associated config (such as subnets) to pick where
+# to launch tasks is ideal.
+region: 'us-central1'
```
oliverchang, line 72: what's using this exactly? 

can you also explain a bit why we can't just go with "a system where
# we have a list of zones and associated config (such as subnets) to pick where
# to launch tasks is ideal." in this PR? Are there anything blockign this?

src/clusterfuzz/_internal/base/tasks/__init__.py
oliverchang: OK! Please add a tODO here in that case.

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -0,0 +1,227 @@
...
+    for job in ndb_utils.get_all_from_query(data_types.Job.query()):
+      jobs[job.name] = FuzzerJob(
+          job=job.name,
+          project=job.project,
+          queue=tasks.queue_for_platform(job.platform))
+
+    fuzzer_job_weight_by_project = collections.defaultdict(int)
+    fuzzer_jobs = {}
+    fuzzer_job_query = ndb_utils.get_all_from_query(
+        data_types.FuzzerJob.query())
```
oliverchang, line 134: Is this performant enough? Do we need to use `FuzzerJobs` (plural) here? (Which IIRC we created to address performance issues).

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -0,0 +1,227 @@
...
+        continue
+
+      fuzzer_job.weight = fuzzer_job_db.actual_weight * project_weight
+      key = get_fuzzer_job_key(fuzzer_job_db.fuzzer, fuzzer_job_db.job)
+      fuzzer_jobs[key] = fuzzer_job
+
+      fuzzer_job_weight_by_project[fuzzer_job.project] += (
+          fuzzer_job_db.actual_weight)
+
+    for key, fuzzer_job in list(fuzzer_jobs.items()):
```
oliverchang, line 154: Do we actually need this key? Or even this `fuzzer_jobs` dict at all? Can it just be a list instead?

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+class OssfuzzFuzzTaskScheduler(BaseFuzzTaskScheduler):
+  """Fuzz task scheduler for OSS-Fuzz."""
+
+  def get_fuzz_tasks(self) -> Dict[str, tasks.Task]:
+    # TODO(metzman): Handle high end.
+    # A job's weight is determined by its own weight and the weight of the
+    # project is a part of. First get project weights.
+    logs.info('Getting projects.')
```
oliverchang, line 112: Is it more useful to instead log the total weight we computed?

i.e. "Computed total project CPU weight of X from Y projects"

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -0,0 +1,227 @@
...
+  def get_fuzz_tasks(self):
+    raise NotImplementedError('Child class must implement.')
+
+  def _get_cpus_per_fuzz_job(self, job_name):
+    del job_name
+    # TODO(metzman): Actually implement this.
+    return 2
+
+
+class FuzzerJob:
```
oliverchang, line 84: this is a bit confusing. can we rename this something to differentiate it more from data_types.FuzzerJob? Something like "FuzzTaskCandidate" ?

oliverchang (commented):
nice!!

src/clusterfuzz/_internal/cron/schedule_fuzz.py
jonathanmetzman: It is very performant. I'm not sure what the original perf issue was.
Locally (where I have a bunch of cores, thus improving the speed of the actual bottleneck, putting the tasks on the queue) we can schedule ~125 tasks in ~15 seconds.

src/clusterfuzz/_internal/cron/schedule_fuzz.py
jonathanmetzman: Youre right. I think there was some left over structure from my initial implementations which were more complex since they didn't have the FuzzerJob class.

src/clusterfuzz/_internal/cron/schedule_fuzz.py
jonathanmetzman: These logs are sort of only for tracking control flow, they probably can be removed now that I know this code works pretty well. Should I?

I don't think the weight thing is important to log because i don't think it will be a meaningful number. The total is just a denominator basically right?

configs/test/batch/batch.yaml
jonathanmetzman: It's old. I removed it.
There's nothing blocking it to be honest. But I think it's a bit complex to check multiple regions for CPU usage instead of just one.
I think we can simplify our implementation if we just force all preemptible (fuzz) tasks into us-central1 and all non-preemptible tasks (everything else) into us-east4. We won't need to do anything complicated to pick the region and we will never starve the non-fuzz tasks (and will have great latency for these). The problem with this approach is:
1. Cloud might not appreciate upping our CPU usage by 50k in one region and prefer we split the workload.
2. We may want to fuzz less if we are doing a lot of other tasks, the approach I give above means that our spending is uncapped, it's whatever fuzzing capacity is + how many non-fuzz tasks are requested).

src/clusterfuzz/_internal/cron/schedule_fuzz.py
jonathanmetzman: I'm going to explain, but basically in our chrome instance (https://pantheon.corp.google.com/iam-admin/quotas?e=-13802955&mods=logs_tg_prod&project=google.com:clusterfuzz&pageState=(%22allQuotasTable%22:(%22f%22:%22%255B%255D%22))) there is no preemptible CPU quota, just one CPU quota. That means we need to obey the CPU quota when the preemptible CPU doesn't exist (ie. is 0).

oliverchang: ack, thanks! Clarifying that in a comment would be very helpful for future contributors.

PR #4400, opened 2024-11-12
Upload crashes to logs bucket in permissionless way

jonathanmetzman:
This puts crashes in the output for uploading. That can probably be replaced with a signed_url.
It can probably be improved if the crash is included elsewhere.


src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -2067,13 +2073,22 @@ def _to_engine_output(output: str, return_code: int,
       output=bytes(truncated_output, 'utf-8'),
       return_code=return_code,
       timestamp=proto_timestamp)
+
+  if crash_path is None:
+    return engine_output
+  if os.path.getsize(crash_path) > 10 * 1024**2:
+    return engine_output
```
oliverchang, line 2080: should we log an error ? this seems pretty bad if we're not uploading a testcase right?

src/clusterfuzz/_internal/bot/testcase_manager.py
```diff
@@ -534,7 +535,7 @@ def _do_run_testcase_and_return_result_in_queue(crash_queue,
       # Don't upload uninteresting testcases (no crash) or if there is no log to
       # correlate it with (not upload_output).
       if upload_output:
-        upload_testcase(file_path, log_time)
+        upload_testcase(file_path, log_time, None)
```
oliverchang, line 538: wrong order of arguments?

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
jonathanmetzman: It's just a testcase in logs. Really not that important.

src/clusterfuzz/_internal/bot/testcase_manager.py
jonathanmetzman: Youre right.
I think because of how many edge cases our code has there's no actual misbehavior resulting from this bug

PR #4401, opened 2024-11-12
Log uworker input and output URLs.

jonathanmetzman:
This will help debug.
Because these are not signed, one needs permissions to use these.


oliverchang (commented):
LGTM

PR #4402, opened 2024-11-13
Dont do untrusted worker stuff on batch

jonathanmetzman:
This is breaking fuzzing on batch in oss-fuzz.
Looking forward to the day when we can get rid of untrusted worker. Living side by side with uworker is unpleasant.


src/clusterfuzz/_internal/bot/tasks/setup.py
```diff
@@ -467,6 +467,12 @@ def _prepare_update_data_bundle(fuzzer, data_bundle):
   return data_bundle_directory
 
 
+def take_trusted_host_path():
+  if environment.is_uworker():
```
oliverchang, line 471: Would it make sense to move this check into `is_trusted_host`? or is it specific to this case?

oliverchang (commented):
lgtm

PR #4403, opened 2024-11-13
Make find_main_crash uworker safe

jonathanmetzman:


src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -364,13 +365,8 @@ def __init__(self, crashes: List[Crash], context,
...
 
     self.crashes = crashes
-    if context.fuzz_target:
-      fully_qualified_fuzzer_name = context.fuzz_target.fully_qualified_name()
-    else:
-      fully_qualified_fuzzer_name = context.fuzzer_name
-
     self.main_crash, self.one_time_crasher_flag = find_main_crash(
-        crashes, fully_qualified_fuzzer_name, context.test_timeout, upload_urls)
+        crashes, context.fuzz_target, context.test_timeout, upload_urls)
```
oliverchang, line 369: is this correct? we need to fallback to the fuzzer name if it's not a fuzz target.

PR #4406, opened 2024-11-14
Skipping CCd users without GAIA accounts for buganizer issues

vitorguidi:
### Motivation

oss_fuzz_build_status and oss_fuzz_apply_ccs are broken due to buganizer rejecting emails without associated accounts. This serves as a workaround so at least some notifications can suceeed. (https://github.com/google/oss-fuzz/issues/12536)

### Why this works

Buganizer throws 400 errors listing out all rejected emails in the exception, so we can just parse the non gaia emails from that and complete the execution by ignoring those.

### Testing strategy

Ran both cronjobs locally with the proposed new code, and verified they ran to completion.


src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
```diff
@@ -68,6 +70,24 @@ def _get_access_limit_from_labels(labels: issue_tracker.LabelStore):
...
+def retry_on_invalid_gaia_accounts(func):
+  """Decorator to retry saving an issue, skipping CCs
+    without a GAIA account."""
+
+  @functools.wraps(func)
+  def wrapper(self, *args, **kwargs):
+    try:
+      return func(self, *args, **kwargs)
+    except Exception as e:
+      if "HttpError 400" in str(e):
```
oliverchang, line 82: nit: single quotes for all strings.

src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
```diff
@@ -719,7 +739,8 @@ def _update_issue(self, new_comment=None, notify=True):
                                       issueId=str(self.id)))
     return result
 
-  def save(self, new_comment=None, notify=True):
+  @retry_on_invalid_gaia_accounts
+  def save(self, skip_emails=[], new_comment=None, notify=True):
```
oliverchang, line 743: nit: add `skip_emails` to the very end, in case there are callers that are relying on argument order.

oliverchang, line 743: would it make sense to log `skip_emails` here as a single place that logs all the skipped emails for a given issue?

src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
```diff
@@ -68,6 +70,24 @@ def _get_access_limit_from_labels(labels: issue_tracker.LabelStore):
...
+  """Decorator to retry saving an issue, skipping CCs
+    without a GAIA account."""
+
+  @functools.wraps(func)
+  def wrapper(self, *args, **kwargs):
+    try:
+      return func(self, *args, **kwargs)
+    except Exception as e:
+      if "HttpError 400" in str(e):
+        emails_to_skip = re.findall(r'[\w\.-]+@[\w\.-]+', str(e))
```
oliverchang, line 83: Do multiple emails get listed in the exception? Or is it one at a time?

oliverchang, line 83: I believe the "+" character itself is also common in emails in the part before the "@", and isn't captured here. 

e.g. "foo+alerts@domain.com"

vitorguidi: I have no idea to be honest, I will do some code archeology on buganizer and get back to you.

oliverchang: ack. my main question is if we need to run this in a loop up to a certain number of retries.

vitorguidi: No, one retry is enough. Buganizer concatenates all failed emails, linked source reference in internal chat

src/clusterfuzz/_internal/cron/oss_fuzz_apply_ccs.py
```diff
@@ -67,7 +68,16 @@ def main():
...
 
-    issue.save(notify=True)
+    try:
+      issue.save(notify=True)
+    except Exception as e:
+      error_count += 1
+      logging.error('Failed to apply ccs for test case '
+                    '%s: %s.', testcase.key, e)
+
+  if error_count:
```
oliverchang, line 78: any reason this needs to be an int instead of just of a bool?

oliverchang (approved):

jonathanmetzman (approved):
lgtm

PR #4409, opened 2024-11-15
doc: Remove html b-tags from oss_fuzz_build_status.py

maflcko:
They are unsupported on the new tracker and aren't needed anyway. Example: https://issues.oss-fuzz.com/issues/379066972#comment1


jonathanmetzman:
Thanks!

oliverchang (approved):

PR #4410, opened 2024-11-15
Handle deleted fuzz targets in split builds

jonathanmetzman:
Fixes: https://buganizer.corp.google.com/issues/377701732


jonathanmetzman:
I'm concerned I'm not dealing with this properly for non-split builds.

oliverchang (approved):
thanks!

PR #4413, opened 2024-11-18
Increasing number of retries for buganizer

vitorguidi:
### Motivation

oss-fuzz-apply-ccs is failing. A possible cause is request rate limiting on the buganizer API. This PR increases the number of retries for buganizer requests, in order to fix this particular issue.

### DIFF
--- a/src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
+++ b/src/clusterfuzz/_internal/issue_management/google_issue_tracker/issue_tracker.py
@@ -32,7 +32,7 @@ from clusterfuzz._internal.issue_management import issue_tracker
 from clusterfuzz._internal.issue_management.google_issue_tracker import client
 from clusterfuzz._internal.metrics import logs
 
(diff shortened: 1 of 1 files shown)


jonathanmetzman (approved):
lgtm

jonathanmetzman:
I wonder if retries is a good way to deal with rate limiting.
Are these isues filed concurrently?

oliverchang (approved):
Do we still want this? Or have we determined a better root cause for the issues we're seeing?

vitorguidi:
> Do we still want this? Or have we determined a better root cause for the issues we're seeing?

This is actually the same thing as build-stats, 400 response (maybe from the gaia issue , I need to look further into it)

```
 <HttpError 400 when requesting https://issuetracker.googleapis.com/v1/issues/42478377:modify?alt=json returned "Request contains an invalid argument.". Details: "Request contains an invalid argument.">.
 ```
 
 Closing this, since the root cause is something else.

PR #4414, opened 2024-11-19
[Monitoring] Adding a blackbox fuzzer testcase generation time metric

vitorguidi:
### Motivation

Chrome folks need to know how long on average a fuzzer takes to generate a testcase. This PR implements that.

Part of #4271

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
@@ -1581,22 +1581,25 @@ class FuzzingSession:
 
     # Run the fuzzer to generate testcases. If error occurred while trying
     # to run the fuzzer, bail out.
-    testcase_generation_start_time = time.time()
+    testcase_generation_start = time.time()
     generate_result = self.generate_blackbox_testcases(
         fuzzer, job_type, fuzzer_directory, testcase_count)
     if not generate_result.success:
       return None, None, None, None
-    testcase_generation_finish_time = time.time()
-    elapsed_testcase_generation_time = testcase_generation_finish_time - testcase_generation_start_time
+    testcase_generation_finish = time.time()
+    elapsed_testcase_generation_time = testcase_generation_finish
+    elapsed_testcase_generation_time -= testcase_generation_start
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -1581,10 +1581,22 @@ def do_blackbox_fuzzing(self, fuzzer, fuzzer_directory, job_type):
 
     # Run the fuzzer to generate testcases. If error occurred while trying
     # to run the fuzzer, bail out.
+    testcase_generation_start_time = time.time()
```
jonathanmetzman: We don't want to include failures? What if there's exceptions? I think having a context manager can help here and avoid cluttering up the code with metrics code.

vitorguidi: I don't see how to include failures here. If an exception is thrown on line 1587, then we do not really care about partial successes (and we do not even have a way to get a partial result from the inner function, since it is throwing anyway).

Am I missing something obvious here?

vitorguidi: Re the context manager, an auxiliary function is simpler IMO, since there is really nothing special to do when exceptions show up

jonathanmetzman: Auxiliary sgtm!

alhijazi (approved):

src/clusterfuzz/_internal/metrics/monitoring_metrics.py
```diff
@@ -140,6 +140,19 @@
     ],
 )
 
+TESTCASE_GENERATION_AVERAGE_TIME = monitor.CumulativeDistributionMetric(
+    'task/fuzz/fuzzer/testcase_generation_duration',
+    bucketer=monitor.FixedWidthBucketer(width=0.05, num_finite_buckets=20),
```
vitorguidi, line 145: This would cap the metric at 1, just like the others before.

PR #4415, opened 2024-11-19
[Monitoring] Adding metric to track time from testcase creation to bug filing

vitorguidi:
### Motivation

As per Chrome request, it is desirable to know how long it takes for an issue to be opened, from the moment a testcase is created.

Part of #4271

### DIFF
--- a/src/clusterfuzz/_internal/cron/triage.py
+++ b/src/clusterfuzz/_internal/cron/triage.py
@@ -254,6 +254,16 @@ def _check_and_update_similar_bug(testcase, issue_tracker):
   return False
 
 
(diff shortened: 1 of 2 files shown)


oliverchang (approved):
LGTM, but note there are some legitimate cases where we might delay filing of a bug (e.g. https://github.com/google/clusterfuzz/blob/c46d537b208b1e21d13aeb0786e8b52c4a64aea6/src/clusterfuzz/_internal/cron/triage.py#L249).

Other cases include if a testcase is part of a deduplicated group. 

If we plan to build alerting policies on these, we may need to give these some special consideration.

oliverchang:
CC @alhijazi @letitz

alhijazi (approved):

PR #4416, opened 2024-11-20
Make task rate limiting opt in

jonathanmetzman:
Make the rate limiting added in https://github.com/google/clusterfuzz/commit/c26e353f230793f148860718fad866cbc04b82f3 opt-in temporarily so we can deploy to other instances without worrying if this damages their fuzzing.


oliverchang (approved):
Can you give some context for this change?

PR #4422, opened 2024-11-22
[tworker] Speed up fuzz_task preprocess

jonathanmetzman:
Attempt to make the tworkers faster so they can scale up to OSS-Fuzz.
Make the following changes based on profiling.

1. Sign URLs in parallel.
2. Don't do postprocess tasks. Pulling from a queue which we don't get anything is relatively expensive.
3. Don't clean up after tasks since preprocess doesn't change the state of the machine.
4. Don't sign delete URLs for corpus in fuzz task, we don't delete in fuzz task anyway.
5. Limit the number of upload URLs for corpus in fuzz task to the number we will actually use.
6. Memoize to avoid parsing YAML often.
7. (unrelated to preprocess) Remove overly-conservative, useless limitation on async downloads.
8. Remove unneeded call to last_updated.
9. Remove unnecessary sleep when task is rate limited. This sleep only makes sense for oss-fuzz hosts.

### DIFF
--- a/src/clusterfuzz/_internal/base/concurrency.py
+++ b/src/clusterfuzz/_internal/base/concurrency.py
@@ -21,8 +21,9 @@ from clusterfuzz._internal.system import environment
 POOL_SIZE = multiprocessing.cpu_count()
 
 
-class InProcessPool:
-
+class SingleThreadPool:
+  """Single thread pool for when it's not worth using Python's thread
+  implementation."""
   def __init__(self, size):
     del size
 
--- a/src/clusterfuzz/_internal/base/concurrency.py
+++ b/src/clusterfuzz/_internal/base/concurrency.py
@@ -24,6 +24,7 @@ POOL_SIZE = multiprocessing.cpu_count()
 class SingleThreadPool:
   """Single thread pool for when it's not worth using Python's thread
   implementation."""
+
   def __init__(self, size):
     del size
 
@@ -41,7 +42,7 @@ def make_pool(pool_size=POOL_SIZE, cpu_bound=False, max_pool_size=None):
   if (environment.get_value('PY_UNITTESTS') or
       environment.platform() == 'WINDOWS'):
     if cpu_bound:
-      yield InProcessPool(pool_size)
+      yield SingleThreadPool(pool_size)
     else:
       yield futures.ThreadPoolExecutor(pool_size)
   else:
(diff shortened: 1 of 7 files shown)


src/clusterfuzz/_internal/base/concurrency.py
```diff
@@ -21,11 +21,30 @@
 POOL_SIZE = multiprocessing.cpu_count()
 
 
+class InProcessPool:
```
oliverchang, line 24: can you add docstrings? 

Also -- this is more of a "SingleThreadedPool" right?

src/clusterfuzz/_internal/google_cloud_utils/storage.py
```diff
@@ -1351,12 +1352,17 @@ def _sign_urls_for_existing_file(
...
+  return _sign_urls_for_existing_file(url, include_delete_urls)
+
+
 def sign_urls_for_existing_files(urls,
                                  include_delete_urls) -> List[Tuple[str, str]]:
   logs.info('Signing URLs for existing files.')
-  result = [
-      _sign_urls_for_existing_file(url, include_delete_urls) for url in urls
-  ]
+  args = ((url, include_delete_urls) for url in urls)
```
oliverchang, line 1363: Is `include_delete_urls` the same value for the entire call? 

It's not clear to me why we need this `_mappable_sign_urls_for_existing_file` in that case, and all these list comprehensions. 

Can we avoid all of these and just do:

```python
pool.map(lambda url: _sign_urls_for_existing_file(url, include_delete_urls), urls)
```

?

jonathanmetzman: I can try, but the reason I'm programming like closures don't exist is because I think they've caused me issues with multiprocessing and we don't have test coverage of multiprocessing.

oliverchang: Can you expand on the kind of issues you encountered?

It's not clear to me how using a lambda here would introduce multiprocessing issues. Was it to do with variable capture? 

In any case, having lots of these functions like `_mappable_sign_urls_for_existing_file` that only serve the purpose of changing the structure of some internal representation to make our implementation details fit seems non-ideal. We should either avoid this (e.g. via a lambda), or change our underlying data structures in a way that avoids this need.

PR #4423, opened 2024-11-22
Make sure to schedule pruning for all jobs.

jonathanmetzman:
It seems like some projects are not getting pruning tasks. This might be because we were not querying jobs correctly until this patch.

### DIFF
--- a/src/clusterfuzz/_internal/cron/schedule_corpus_pruning.py
+++ b/src/clusterfuzz/_internal/cron/schedule_corpus_pruning.py
@@ -17,6 +17,7 @@ from clusterfuzz._internal.base import tasks
 from clusterfuzz._internal.base import utils
 from clusterfuzz._internal.datastore import data_types
 from clusterfuzz._internal.datastore import fuzz_target_utils
+from clusterfuzz._internal.datastore import ndb_utils
 from clusterfuzz._internal.metrics import logs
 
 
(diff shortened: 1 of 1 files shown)


oliverchang (commented):
hmm, this would be surprising if it made a difference, from what I remember about how `query()` worked. Did you have any specific reasons to believe this is the cause?

jonathanmetzman:
Hmmm...I think youre right. I think I've seen a bunch of targets where corpus_pruning isn't being scheduled (At all, so I don't think this is a batch issue).
https://pantheon.corp.google.com/logs/query;query=%22Adding%20corpus%20pruning%20task%22%0A%22sound-open-firmware%22;cursorTimestamp=2024-11-21T19:25:59.304007175Z;startTime=2024-11-01T20:31:02.259Z;endTime=2024-11-21T22:31:02.259Z?e=-13802955&mods=logs_tg_prod&project=clusterfuzz-external

jonathanmetzman:
> Hmmm...I think youre right. I think I've seen a bunch of targets where corpus_pruning isn't being scheduled (At all, so I don't think this is a batch issue). https://pantheon.corp.google.com/logs/query;query=%22Adding%20corpus%20pruning%20task%22%0A%22sound-open-firmware%22;cursorTimestamp=2024-11-21T19:25:59.304007175Z;startTime=2024-11-01T20:31:02.259Z;endTime=2024-11-21T22:31:02.259Z?e=-13802955&mods=logs_tg_prod&project=clusterfuzz-external

Eh turns out that project just has an i386 build. I'll look for more, becuase I've seen a lot of projects with many tens of thousands of corpus elements

PR #4425, opened 2024-11-22
Put corpus pruning tasks on preprocess queue

jonathanmetzman:


oliverchang (approved):

jonathanmetzman:
I think I'm going to abandon this approach and go with: https://github.com/google/clusterfuzz/pull/4435 instead.

PR #4428, opened 2024-11-22
Adding build revision metric

vitorguidi:
### Motivation

Chrome folks want to know what build revision is being used in fuzz task for every job in fuzz_task. This PR implements that.

### DIFF
--- a/src/clusterfuzz/_internal/build_management/build_manager.py
+++ b/src/clusterfuzz/_internal/build_management/build_manager.py
@@ -1226,6 +1226,16 @@ def _emit_build_age_metric(gcs_path):
   except Exception as e:
     logs.error(f'Failed to emit build age metric for {gcs_path}: {e}')
 
+def _emit_build_revision_metric(revision):
+  """Emits a gauge metric to track the build revision."""
+  monitoring_metrics.JOB_BUILD_REVISION.set(
+    revision,
+    labels = {
+        'job': os.getenv('JOB_NAME'),
+        'platform': environment.platform(),
+        'task': os.getenv('TASK_NAME'), 
+    }
+  )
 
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/metrics/monitoring_metrics.py
```diff
@@ -44,6 +44,17 @@
     ],
 )
 
+JOB_BUILD_REVISION = monitor.GaugeMetric(
+    'job/build_age',
```
letitz, line 48: Name should be different

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -1227,6 +1227,17 @@ def _emit_build_age_metric(gcs_path):
     logs.error(f'Failed to emit build age metric for {gcs_path}: {e}')
 
 
+def _emit_build_revision_metric(revision):
+  """Emits a gauge metric to track the build revision."""
+  monitoring_metrics.JOB_BUILD_REVISION.set(
+      revision,
+      labels={
+          'job': os.getenv('JOB_NAME'),
```
letitz, line 1235: I thought using `environment.get()` was preferred to raw `os.getenv()`.

vitorguidi: I do not know why to prefer one over the other, @jonathanmetzman wdyt?

letitz: Ah, I thought there was something to do with mocking, or a difference between `os.getenv()` and `os.environ.get()`, but it turns out I'm misremembering. There's probably no use then.

jonathanmetzman (approved):
lgtm

oliverchang (approved):

PR #4431, opened 2024-11-25
Speed up preprocessing significantly

jonathanmetzman:
1. Cache credentials. This should have been done long ago.
Not doing so causes a new process to be created each time credentials
are requested. This is probably the highest risk change.
2. Properly skip task cleanup. It's not needed in tworkers.
3. Cache leak sanitizer db queries in Redis.
4. Only request names when listing contents of GCS directory.
5. Don't check that there is no UUID collision when uploading corpus elements. This is basically impossible and wastes time. It is also benign and not a security risk if it happens within the same 24 hours (one fuzz task will overwrite the corpus for the same fuzzer).

### DIFF
--- a/src/clusterfuzz/_internal/fuzzing/leak_blacklist.py
+++ b/src/clusterfuzz/_internal/fuzzing/leak_blacklist.py
@@ -17,6 +17,7 @@ import os
 import re
 
 from clusterfuzz._internal.base import errors
+from clusterfuzz._internal.base import memoize
 from clusterfuzz._internal.datastore import data_handler
 from clusterfuzz._internal.datastore import data_types
 from clusterfuzz._internal.datastore import ndb_utils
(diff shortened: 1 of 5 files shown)


jonathanmetzman:
Some of the changes here are related to improving the speed of multiprocess URL signing. I will undo this, as I have a suspicion it's caused hosts to die.

src/clusterfuzz/_internal/fuzzing/leak_blacklist.py
```diff
@@ -65,8 +66,9 @@ def cleanup_global_blacklist():
   ndb_utils.delete_multi(blacklists_to_delete)
 
 
+@memoize.wrap(memoize.Memcache(60 * 10))
```
jonathanmetzman, line 69: I use memcach for this so it can expire.

PR #4435, opened 2024-11-25
[oss-fuzz] Run every preprocess in preprocess pool

jonathanmetzman:
Only do this in oss-fuzz for now.


jonathanmetzman:
Note that I am merging this into the oss-fuzz branch for now which I will only use for oss-fuzz until the changes on it are ready for chrome.

oliverchang (approved):

PR #4436, opened 2024-11-25
Handle new fuzz targets in split builds properly

jonathanmetzman:
Previously we tried to handle it by finding the fuzz targets in the list. But we still failed because it wasn't in the db. Save it to the db to solve this problem.


jonathanmetzman:
CC @marktefftech

marktefftech (approved):
Assuming `binary=fuzz_target_binary,` will pickup the binaries in the `FUZZ_TARGET_BUILD_BUCKET` location, this LGTM.

oliverchang (approved):

PR #4438, opened 2024-11-26
Disable untriaged testcase age metrics in triage.py

ParisMeuleman:
This seems to be causing crbug.com/380707237, disable it until we get a working solution.


ParisMeuleman:
closing in favor of the fix: https://github.com/google/clusterfuzz/pull/4439

PR #4441, opened 2024-11-26
Fix Analyze Task

alhijazi:
Fetching the testcase from the datastore has been wrongly moved before the call to `_update_testcase`. This leads to us holding an outdated version of the testcase.


letitz (approved):
I'm surprised no tests fail. It seems to me we could have a simple test that checks that after postprocess, at least one of the fields in the testcase in the DB was updated to reflect uworker output?

Issue #83, opened 2019-02-05
Document monitoring workflow


Dor1s:
e.g. how to use Error Reporting dashboard, check logs

I've also noticed a bunch of error during production testing, those might scare the people, might be good to mention some of them or maybe even downgrade to warnings?

PR #4444, opened 2024-11-26
Do safe preprocess optimizations.

jonathanmetzman:
1. Don't unnecessarily list features about blobs other than names
2. Memoize leak blacklist stuff.
3. Properly skip cleanup.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/commands.py
+++ b/src/clusterfuzz/_internal/bot/tasks/commands.py
@@ -80,6 +80,8 @@ class AlreadyRunningError(Error):
 def cleanup_task_state():
   """Cleans state before and after a task is executed."""
   # Cleanup stale processes.
+  if not environment.is_tworker():
+    return
   process_handler.cleanup_stale_processes()
 
   # Clear build urls, temp and testcase directories.
@@ -468,7 +470,6 @@ def process_command_impl(task_name, task_argument, job_name, high_end,
     return run_command(task_name, task_argument, job_name, uworker_env)
   finally:
     # Final clean up.
-    if not environment.is_tworker():
-      cleanup_task_state()
+    cleanup_task_state()
     if 'CF_TASK_ID' in os.environ:
       del os.environ['CF_TASK_ID']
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
This has some of the changes from https://github.com/google/clusterfuzz/pull/4431 while I wait for a review.

src/clusterfuzz/_internal/bot/tasks/commands.py
```diff
@@ -80,6 +80,8 @@ class AlreadyRunningError(Error):
 def cleanup_task_state():
   """Cleans state before and after a task is executed."""
   # Cleanup stale processes.
+  if not environment.is_tworker():
```
vitorguidi, line 83: How bad is it if this codepath gets hit by the uworker? Why not an assertion?

jonathanmetzman: I expect the codepath to be hit by both uworker and tworker (is that what you mean).
I just don't want to do the cleanup in a twoker. It's pretty slow and it's doing things a tworker doesn't need like killing processes and removing tmpdirs. tworker doesn't really need this because it's not actually running or downloading anything.

src/clusterfuzz/_internal/bot/tasks/commands.py
```diff
@@ -468,7 +470,6 @@ def process_command_impl(task_name, task_argument, job_name, high_end,
     return run_command(task_name, task_argument, job_name, uworker_env)
   finally:
     # Final clean up.
-    if not environment.is_tworker():
-      cleanup_task_state()
+    cleanup_task_state()
```
vitorguidi, line 473: I assume this got removed from #83, since you make sure that this code is not reached by the uworker?

jonathanmetzman: Sorry do you mean a different PR?

vitorguidi (approved):
lgtm, with some clarifications

src/clusterfuzz/_internal/bot/tasks/commands.py
vitorguidi: Line of code, sorry

jonathanmetzman: I find the question kind of confusing, but I'm no longer relying on that old if statement, the new one is better because the function is called from more than one place.
This doesn't really have anything to do with the untrusted arch.

PR #4446, opened 2024-11-26
Cache credentials

jonathanmetzman:
Getting credentials is pretty slow since it launches a gcloud process. We should just cache credentials. Supposedly the library itself handles refreshes. We are also caching in storage.py so there's proof it's safe.


jonathanmetzman:
This has some of the changes from https://github.com/google/clusterfuzz/pull/4431 while I wait for a review.

vitorguidi (approved):
lgtm

PR #4450, opened 2024-11-26
 Increase fuzz tasks scheduled from 1250->2000 every 5 minutes

jonathanmetzman:


oliverchang (approved):

jonathanmetzman:
3000 is I think close to what oss-fuzz does.

PR #4451, opened 2024-11-26
Fix issue scheduling tasks on preprocess queue

jonathanmetzman:
Also increase the number of utasks we can get from the queue


jonathanmetzman:
Test failure is spurious.

PR #4399, opened 2024-11-12
Support multiple regions in batch

jonathanmetzman:
This will be needed for oss-fuzz which has a workload that is way too big to run in one region.

PR #4453, opened 2024-11-26
Support multiple regions in batch

jonathanmetzman:
Use randomness to "load balance" for now.


jonathanmetzman:
Supersedes #4399

PR #4458, opened 2024-11-27
[Monitoring] Adding a metric for task outcome

vitorguidi:
### Motivation

We currently have no metric that tracks the error rate for each task. This PR implements that, and the error rate can be obtained by summing up the metric with outcome=failure, divided by the overall sum.

This is useful for SLI alerting.

Part of #4271

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/commands.py
+++ b/src/clusterfuzz/_internal/bot/tasks/commands.py
@@ -208,6 +208,12 @@ def get_command_object(task_name):
   # Force remote execution.
   return task_types.UTask(_COMMAND_MODULE_MAP[task_name])
 
+def _emit_task_outcome_metric(task, job, outcome):
+  monitoring_metrics.TASK_OUTCOME_COUNT.increment(labels={
+    'job': job,
+    'task': task,
+    'outcome': outcome,
+  })
 
 def run_command(task_name, task_argument, job_name, uworker_env):
   """Runs the command."""
@@ -249,14 +255,17 @@ def run_command(task_name, task_argument, job_name, uworker_env):
     # where a test case is reloaded from the datastore, just abort the task.
     logs.warning('Test case %s no longer exists.' % task_argument)
     rate_limiter.record_task(success=False)
+    _emit_task_outcome_metric(task_name, job_name, 'failure')
   except BaseException:
     # On any other exceptions, update state to reflect error and re-raise.
     rate_limiter.record_task(success=False)
+    _emit_task_outcome_metric(task_name, job_name, 'failure')
     if should_update_task_status(task_name):
       data_handler.update_task_status(task_state_name,
                                       data_types.TaskState.ERROR)
     raise
   else:
+    _emit_task_outcome_metric(task_name, job_name, 'success')
     rate_limiter.record_task(success=True)
 
   # Task completed successfully.
--- a/src/clusterfuzz/_internal/bot/tasks/commands.py
+++ b/src/clusterfuzz/_internal/bot/tasks/commands.py
@@ -209,14 +209,6 @@ def get_command_object(task_name):
   return task_types.UTask(_COMMAND_MODULE_MAP[task_name])
 
 
-def _emit_task_outcome_metric(task, job, outcome):
-  monitoring_metrics.TASK_OUTCOME_COUNT.increment(labels={
-      'job': job,
-      'task': task,
-      'outcome': outcome,
-  })
-
-
 def run_command(task_name, task_argument, job_name, uworker_env):
   """Runs the command."""
   task = get_command_object(task_name)
@@ -257,17 +249,14 @@ def run_command(task_name, task_argument, job_name, uworker_env):
     # where a test case is reloaded from the datastore, just abort the task.
     logs.warning('Test case %s no longer exists.' % task_argument)
     rate_limiter.record_task(success=False)
-    _emit_task_outcome_metric(task_name, job_name, 'failure')
   except BaseException:
     # On any other exceptions, update state to reflect error and re-raise.
     rate_limiter.record_task(success=False)
-    _emit_task_outcome_metric(task_name, job_name, 'failure')
     if should_update_task_status(task_name):
       data_handler.update_task_status(task_state_name,
                                       data_types.TaskState.ERROR)
     raise
   else:
-    _emit_task_outcome_metric(task_name, job_name, 'success')
     rate_limiter.record_task(success=True)
 
   # Task completed successfully.
--- a/src/clusterfuzz/_internal/bot/tasks/commands.py
+++ b/src/clusterfuzz/_internal/bot/tasks/commands.py
@@ -208,12 +208,14 @@ def get_command_object(task_name):
   # Force remote execution.
   return task_types.UTask(_COMMAND_MODULE_MAP[task_name])
 
+
 def _emit_task_outcome_metric(task, job, outcome):
   monitoring_metrics.TASK_OUTCOME_COUNT.increment(labels={
-    'job': job,
-    'task': task,
-    'outcome': outcome,
+      'job': job,
+      'task': task,
+      'outcome': outcome,
   })
+
 
 def run_command(task_name, task_argument, job_name, uworker_env):
   """Runs the command."""
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
@@ -75,6 +75,7 @@ class _MetricRecorder(contextlib.AbstractContextManager):
   Members:
     start_time_ns (int): The time at which this recorder was constructed, in
       nanoseconds since the Unix epoch.
+    saw_failure: indicates if utask_main returned an error code as a value.
   """
 
   def __init__(self, subtask: _Subtask):
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
@@ -75,14 +75,15 @@ class _MetricRecorder(contextlib.AbstractContextManager):
   Members:
     start_time_ns (int): The time at which this recorder was constructed, in
       nanoseconds since the Unix epoch.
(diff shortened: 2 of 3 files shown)


jonathanmetzman (commented):
utask_main is run with `utasks.uworker_bot_main` You might want to catch this too.

src/clusterfuzz/_internal/bot/tasks/commands.py
```diff
@@ -249,14 +257,17 @@ def run_command(task_name, task_argument, job_name, uworker_env):
     # where a test case is reloaded from the datastore, just abort the task.
     logs.warning('Test case %s no longer exists.' % task_argument)
     rate_limiter.record_task(success=False)
+    _emit_task_outcome_metric(task_name, job_name, 'failure')
   except BaseException:
     # On any other exceptions, update state to reflect error and re-raise.
     rate_limiter.record_task(success=False)
+    _emit_task_outcome_metric(task_name, job_name, 'failure')
```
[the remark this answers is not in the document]
vitorguidi: Actually this should not be here, would be better to collect success/failure per:

Task (fuzz/etc)
Stage (pre/main/post)
Outcome (ok/error)

Probably _internal/bot/tasks/utasks/__init__.py ?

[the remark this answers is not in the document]
vitorguidi: Yeah, nevermind this, reusing _MetricsRecorder is a better idea

jonathanmetzman (approved):
lgtm

src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
```diff
@@ -138,6 +138,11 @@ def __exit__(self, _exc_type, _exc_value, _traceback):
     monitoring_metrics.UTASK_SUBTASK_E2E_DURATION_SECS.add(
         e2e_duration_secs, self._labels)
 
+    outcome = 'error' if _exc_type else 'success'
```
vitorguidi, line 141: Is there any possible situation where we return an error code as a value, and not throw an exception, when a task fails in any of the pre/main/post steps?

vitorguidi: https://github.com/google/clusterfuzz/blob/089d9ef1a8f6bdb03a9046ad5c3286874384cf3a/src/clusterfuzz/_internal/protos/uworker_msg.proto#L355

Should probably be introspecting the output to pinpoint failure if the result is anything other than NO_ERROR?

vitorguidi: Can't really think of any other scenario where a task failure does not imply an exception.

src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
```diff
@@ -138,6 +140,14 @@ def __exit__(self, _exc_type, _exc_value, _traceback):
     monitoring_metrics.UTASK_SUBTASK_E2E_DURATION_SECS.add(
         e2e_duration_secs, self._labels)
 
+    # The only case where a task might fail without throwing, is in
+    # utask_main, by returning an ErrorType proto which indicates
+    # failure.
+    outcome = 'error' if _exc_type or self.saw_failure else 'success'
+    monitoring_metrics.TASK_OUTCOME_COUNT.increment({
+        **self._labels, 'outcome': outcome
```
letitz, line 148: Since we have a nice enum of possible errors, I'd much rather we be more specific here and record it as the outcome: e.g. `result: 'NO_ERROR'` or `result: 'BUILD_SETUP_FAILED'`. It would greatly help in debugging problems.

Exceptions can be bucketed as their own `result: 'UNHANDLED_EXCEPTION'`.

Do you think it will pose a problem with metric cardinality?

src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
```diff
@@ -80,6 +81,7 @@ def __init__(self, subtask: _Subtask):
     self.start_time_ns = time.time_ns()
     self._subtask = subtask
     self._labels = None
+    self.saw_failure = False
```
letitz, line 84: Public field should be documented in class docstring.

src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
vitorguidi: Hmm now that I think of it, we should get rid of job

For oss fuzz:

15 tasks
1300 projects ~ 1300 jobs
3 subtasks (pre main post)
2 modes (batch/queue)
3 platforms (linux mac win)
2 outcomes (fail/success)

This goes for around 700k, which will bite us. We could swap job out for all outcomes, and that will be around ~16k possible label combinations, which should be fine. Wdyt?

vitorguidi: @jonathanmetzman the current state of the subtask duration metric allows for around 350k different label combinations, [GCP's recommendation](https://cloud.google.com/logging/docs/logs-based-metrics/troubleshooting#too-many-time-series) is at most 30k. Should we get rid of job in this context manager as a whole?

jonathanmetzman: What's the limit? I think there's other things that can be removed
1. I think modes is definitely not needed, the two modes won't live side by side much longer.
2. I think platform may be unneeded, it can be obtained through the job name.
Would reducing cardinality by 6X help?

jonathanmetzman: Also, there's currently about 1000 platforms in oss-fuzz but that is going away. Will that help?

vitorguidi: It is actually 40 distinct values under that enum.
15 tasks * 40 outcomes * 1300 jobs * 3 subtasks > 1M, so yeah, no chance at all to get the discriminated outcomes, while keeping jobs.

Another option is to spin a separate metric, TASK_OUTCOME_BY_ERROR_TYPE, for which we track all the labels except job, and manage to stay under  30k distinct labels.

Wdyt @letitz

jonathanmetzman: Sorry why 40 outcomes?

vitorguidi: https://github.com/google/clusterfuzz/blob/06173d288ab32f2f1a61e77b083d92b483b00e77/src/clusterfuzz/_internal/protos/uworker_msg.proto#L355

38 ErrorTypes from the utask_main output, 'N/A' (to indicate no error), and 'UNHANDLED_EXCEPTION', as suggested by titouan

vitorguidi:
> ### Motivation
> We currently have no metric that tracks the error rate for each task. This PR implements that, and the error rate can be obtained by summing up the metric with outcome=failure, divided by the overall sum.
> 
> This is useful for SLI alerting.
> 
> Part of #4271

Merging this for the sake of quick iteration, we can revisit this if folks feel like it is necessary

jonathanmetzman:
I think this metric needs some temporary reworking.
image
As is, it's not very useful because it is too close to 100% error rate because ClusterFuzz is treating expected misbehavior (bad builds) as exceptional when they are not.
I think until ClusterFuzz is refactored, this metric would be more useful if it only measured exceptions and not handled errors.

src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
letitz: FIne with me to have a separate metric for error types that does not group by job.

> I think modes is definitely not needed, the two modes won't live side by side much longer.
> I think platform may be unneeded, it can be obtained through the job name.
Would reducing cardinality by 6X help?

Mode does not contribute much to cardinality since we have a mostly static mapping from (task, platform) to mode. Platform is actually free since as you point out, it is implied by job. So there are not 3 copies of each job, one for each platform. IIRC this is what actually matters, not the potential cardinality if we generated all possible labels.

letitz:
It may not be useful for alerting, but I for one find this data interesting. I would not have assumed that most tasks would end in "temporary failure".

Maybe that's the breakdown we want though here? "success", "temporary error"/"retry" and "failure"?

vitorguidi:
> > Breaking down by success/retry/failure is too janky, as we would have to map these 38 ErrorType enums to 3 sets, which would lead to polluting the codebase and make things hard to understand. Also, for every new ErrorType added, this metric would have to be updated, adding further friction to development.
> 
> I would have thought the opposite. Right now, if I want to know what it means for regression task to fail with `REGRESSION_NO_CRASH`, then I have to go read the code. If I could see easily that a certain error just means we'll retry the task, I would find the codebase easier to reason through. That said, this reminds me that errors don't map cleanly to retrying or not, such as when we notice a flake and might or might not retry depending on whether it's a first offense.
> 
> Ultimately I would like to know when there are jobs whose tasks are failing too often, so I can go investigate.

We can partition these errors on a best effort basis in three sets: things that unequivocally imply success (ie, NO_ERROR and ANALYZE_DOES_NOT_REPRODUCE), things for which there **MIGHT BE** a retry (ie, PROGRESSION_TIMEOUT, MINIMIZE_DEADLINE_EXCEEDED), and things that unequivocally imply failure (ie, FUZZ_DATA_BUNDLE_SETUP_FAILURE).

The above would belong in the TASK_OUTCOME_COUNT, under the 'outcome' label:
* success
* potential_retry
* failure

Wdyt? This solves the problem, it will at least be possible to filter for the unambiguous failures

letitz:
Fine with me! @alhijazi might still want to have a say

vitorguidi:
> Fine with me! @alhijazi might still want to have a say

Given the 30k different label limitation, we will not be able to drill down by ErrorType and job simultaneously:

* 38 error types + 'N/A' + 'UNHANDLED_EXCEPTION' as error_count
* 1300+ jobs in oss-fuzz

52k>30k, so this is as far as we can go. This requirement is impossible, unfortunately.

alhijazi:
> Fine with me! @alhijazi might still want to have a say

Fine with me also!

PR #4462, opened 2024-11-28
Optimize corpus signing by 6-8%

jonathanmetzman:
Use imap_unordered for these two improvements.
1. Async signing, we can start signing the arbitrary URLs before we are done listing and signing the existing URLs.
2. Ignoring the order means we can sign/list even faster.
3. Use of iterables means we don't need to use unnecessary memory.


oliverchang (commented):
nice!

PR #4472, opened 2024-12-03
Add logging of clusterfuzz version

ParisMeuleman:


src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -32,6 +33,29 @@
...
+  # (base.memoize > base.persistent_cache > base.utils > base.memoize)
+  from clusterfuzz._internal.base.utils import current_source_version
+  from clusterfuzz._internal.base.utils import utcnow
+  if os.getenv('PY_UNITTESTS'):
+    timestamp = utcnow().strftime('%Y%m%d%H%M%S-utc')
+    components = [timestamp, "PY_UNITTESTS"]
+    return '-'.join(components)
+  if _is_local():
+    timestamp = utcnow().strftime('%Y%m%d%H%M%S-utc')
+    components = [timestamp, os.environ['USER'], "LOCAL"]
```
jonathanmetzman, line 54: nit: single quotes.

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -32,6 +33,29 @@
...
+    return '-'.join(components)
```
jonathanmetzman, line 55: I think we already have the timestamp in the logs.

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -32,6 +33,29 @@
...
+def _current_source_version():
+  """Return the current source version:
+    - Arbitrary value local runs
+    - from base.utils.current_source_version():
+      - SOURCE_VERSION_OVERRIDE if set
+      - from the package's clusterfuzz-source.manifest if present
+      - None"""
+  # local imports to avoid circular imports
+  # (base.memoize > base.persistent_cache > base.utils > base.memoize)
+  from clusterfuzz._internal.base.utils import current_source_version
```
jonathanmetzman, line 46: Nit: we import modules not functions.

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -151,6 +175,8 @@ def get_logging_config_dict(name):
           get_handler_config('bot/logs/android_heartbeat.log', 1),
       'run_cron':
           get_handler_config('bot/logs/run_cron.log', 1),
+      'root':
```
jonathanmetzman, line 178: Why are we adding this?

src/clusterfuzz/_internal/metrics/logs.py
ParisMeuleman: yeah I was keeping the format of current_source_version but that's not really  useful.

src/clusterfuzz/_internal/metrics/logs.py
ParisMeuleman: Unfortunately I can't recall how I hit this up, but it was when running some scripts, anyhow, I removed this, I'll do a dedicated PR if I get this issue again.

jonathanmetzman:
I'm not entirely sure this is needed, given a log, you can always go back and find the version of CF that started the task.

ParisMeuleman:
> I'm not entirely sure this is needed, given a log, you can always go back and find the version of CF that started the task.

You mean if you go back enough in the logs of a given task you'll find the version? I drafted this as a result of the outages the week before, I did not have a clear indication that the failure logs where from a given version, but that might just be that I did not know where to look. If we had this information we would probably have been able to pinpoint the release and take a corrective action much faster.

jonathanmetzman:
> > I'm not entirely sure this is needed, given a log, you can always go back and find the version of CF that started the task.
> 
> You mean if you go back enough in the logs of a given task you'll find the version? I drafted this as a result of the outages the week before, I did not have a clear indication that the failure logs where from a given version, but that might just be that I did not know where to look. If we had this information we would probably have been able to pinpoint the release and take a corrective action much faster.

Exactly. Sorry this wasn't clear. I think I'm in favor of landing this, because it's better for something so important to be in an obvious place.

ParisMeuleman:
Hey @jonathanmetzman  are we going forward with this? Happy to rebase if needed.

PR #4473, opened 2024-12-04
build_manager: allow remote unpacking when unpacking everything

paulsemel:
Now that we are lazily checking for fuzzing targets, it makes sense to allow remote unpacking even when unpacking the full archive. Furthermore, it seems that remote unpacking performances are much higher than local unpacking on CF bots, so this might improve overall performances of the build_manager.


jonathanmetzman (approved):
lgtm

PR #4474, opened 2024-12-05
Reduce bug filing threshold for libfuzzer-based Android crashes

AashutoshMurthy:
Updated triage.py to file bugs for fuzzer crashes by lowering the threshold limit. These fuzzers run on Android platform and on libfuzzer engine.

### DIFF
--- a/src/clusterfuzz/_internal/cron/triage.py
+++ b/src/clusterfuzz/_internal/cron/triage.py
@@ -98,6 +98,60 @@ def _is_bug_filed(testcase):
   return False
 
 
+def _is_blocking_progress(testcase):
+  """Checks the crash frequency if it is reported on libfuzzer"""
+  if testcase.job_type.startswith('libfuzzer'):
+    # Get crash statistics data on this unreproducible crash for last X days.
+    last_hour = crash_stats.get_last_successful_hour()
+    if not last_hour:
+      # No crash stats available, skip.
+      return False
+
+    _, rows = crash_stats.get(
+        end=last_hour,
+        block='day',
+        days=data_types.FILE_CONSISTENT_UNREPRODUCIBLE_TESTCASE_DEADLINE,
+        group_by='reproducible_flag',
+        where_clause=(
+            'crash_type = %s AND crash_state = %s AND security_flag = %s' %
+            (json.dumps(testcase.crash_type), json.dumps(testcase.crash_state),
+             json.dumps(testcase.security_flag))),
+        group_having_clause='',
+        sort_by='total_count',
+        offset=0,
+        limit=1)
+
+    # Calculate total crash count and crash days count.
+    crash_days_indices = set()
+    total_crash_count = 0
+    for row in rows:
+      if 'groups' not in row:
+        continue
+
+      total_crash_count += row['totalCount']
+      for group in row['groups']:
+        for index in group['indices']:
+          crash_days_indices.add(index['hour'])
+
+    crash_days_count = len(crash_days_indices)
+    # Considers an unreproducible testcase as important if the crash
+    # occurred at least once everyday for the last 14 days and total
+    # crash count exceeded 14.
+    return (crash_days_count ==
+            data_types.FILE_CONSISTENT_UNREPRODUCIBLE_TESTCASE_DEADLINE and
+            total_crash_count >=
+            data_types.FILE_UNREPRODUCIBLE_TESTCASE_MIN_STARTUP_CRASH_THRESHOLD)
+
+  return False
+
+
+def is_crash_important_android(testcase):
+  """"Indicate if the android crash is important to file."""
+  if _is_blocking_progress(testcase):
+    return True
+  return False
+
+
 def _is_crash_important(testcase):
   """Indicate if the crash is important to file."""
   if not testcase.one_time_crasher_flag:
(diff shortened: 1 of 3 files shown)


src/clusterfuzz/_internal/cron/triage.py
```diff
@@ -98,6 +98,54 @@ def _is_bug_filed(testcase):
   return False
 
 
+def _is_startup_crash(testcase):
+  """Check if job contains both libfuzzer and android"""
```
jonathanmetzman, line 102: "Checks"
Also, it should be documented in this function name or at least docstring that this only applies to android/libFuzzer. I could only tell from the PR name which isn't going to be available when reading the code.

jonathanmetzman (commented):
Can you share the rationale for this please?

AashutoshMurthy:
> Can you share the rationale for this please?

This PR was created in response to the message chain on [b/372321538](https://b.corp.google.com/issues/372321538).
To be specific, [**Steven**'s comment](https://b.corp.google.com/issues/372321538#comment6) asked for the issue of bugs not being filed, to be fixed.
So, to not affect a large user base, we are reducing the threshold for fuzzers running on android platform and on libfuzzer engine.

src/clusterfuzz/_internal/cron/triage.py
AashutoshMurthy: Done.

src/clusterfuzz/_internal/tests/appengine/handlers/cron/triage_test.py
```diff
@@ -25,6 +25,65 @@
...
+        'clusterfuzz._internal.metrics.crash_stats.get_last_successful_hour',
+        'clusterfuzz._internal.metrics.crash_stats.get',
+        'clusterfuzz._internal.base.utils.utcnow',
+    ])
+    self.mock.utcnow.return_value = test_utils.CURRENT_TIME
+
+  def test_is_libfuzzer_android_startup_crash_1(self):
+    """If this unreproducible testcase (libfuzzer) is crashing frequently,
+    then it is an important crash."""
+    self.mock.get_last_successful_hour.return_value = 417325
```
marktefftech, line 43: What is the point of using this return value? I assume in production maybe it is stored as milliseconds and thus represented by an integer like this... if that is the case, please add a comment to clarify or modify the variable names to indicate such

src/clusterfuzz/_internal/cron/triage.py
```diff
@@ -98,6 +98,54 @@ def _is_bug_filed(testcase):
...
+        continue
+
+      total_crash_count += row['totalCount']
+      for group in row['groups']:
+        for index in group['indices']:
+          crash_days_indices.add(index['hour'])
+
+    crash_days_count = len(crash_days_indices)
+    # Considers an unreproducible testcase as important if the crash
+    # occurred atleast once everyday for the last 14 days and total
```
marktefftech, line 139: nit: at least*

src/clusterfuzz/_internal/tests/appengine/handlers/cron/triage_test.py
AashutoshMurthy: I used this value from an existing test as reference [here](https://github.com/google/clusterfuzz/blob/ca619ef34aeaf35abf9e7bbcfd48b8f497e07ba3/src/clusterfuzz/_internal/tests/appengine/handlers/cron/triage_test.py#L97).

src/clusterfuzz/_internal/cron/triage.py
```diff
@@ -379,7 +427,10 @@ def main():
     # Check if the crash is important, i.e. it is either a reproducible crash
     # or an unreproducible crash happening frequently.
     if not _is_crash_important(testcase):
-      continue
+      # Check if the crash is a startup crash, i.e. it is causing the fuzzer
+      # to crash on startup and not allowing the fuzzer to run longer
+      if not _is_libfuzzer_android_startup_crash(testcase):
```
marktefftech, line 432: Can we rearrange this logic to check that this is on the android platform. Something like:

```
def _is_libfuzzer_android_startup_crash(testcase): # <--- Rename to something more indicative? (ex. is_blocking_progress)
    """Checks if the given testcase is an android libfuzzer crash"""
    ...

def is_crash_important_android(testcase):
    if _is_libfuzzer_android_startup_crash:
        # important

if not _is_crash_important(testcase):
    if testcase.platform == android and not is_crash_important_android(testcase):
        continue
```

marktefftech (commented):
Thanks for proactively adding the tests! This is appreciated.

Left some minor comments

src/clusterfuzz/_internal/cron/triage.py
AashutoshMurthy: Done.

src/clusterfuzz/_internal/cron/triage.py
AashutoshMurthy: I have updated the logic as per your request. Please take a look.

src/clusterfuzz/_internal/cron/triage.py
```diff
@@ -98,6 +98,60 @@ def _is_bug_filed(testcase):
   return False
 
 
+def _is_blocking_progress(testcase):
```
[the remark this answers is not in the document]
svasudevprasad: Done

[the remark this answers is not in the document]
AashutoshMurthy: Done.

PR #4477, opened 2024-12-07
Fix putting tasks on queue

jonathanmetzman:
Allow utask_main scheduling


jonathanmetzman:
Spurious test failure.

PR #4479, opened 2024-12-09
Allow configuring default corpora bucket location.

letitz:
Allow instances to specify the GCS bucket location for data bundle buckets in `project.yaml` as a new key: `data_bundle_bucket_location`.

This will allow creating regional buckets instead of using the default `US` multi-region which results in high data transfer costs in Chrome's instance.


paulsemel (approved):
Thanks for working on this!

jonathanmetzman (approved):

PR #4483, opened 2024-12-10
Mitigate pagination issue listing blobs

jonathanmetzman:
Pagination doesn't happen properly with a filter.

### DIFF
--- a/src/clusterfuzz/_internal/google_cloud_utils/storage.py
+++ b/src/clusterfuzz/_internal/google_cloud_utils/storage.py
@@ -1077,8 +1077,7 @@ def get_blobs(cloud_storage_path, recursive=True):
     exception_types=_TRANSIENT_ERRORS)
 def list_blobs(cloud_storage_path, recursive=True):
   """Return blob names under the given cloud storage path."""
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/google_cloud_utils/storage.py
```diff
@@ -1077,8 +1077,7 @@ def get_blobs(cloud_storage_path, recursive=True):
     exception_types=_TRANSIENT_ERRORS)
 def list_blobs(cloud_storage_path, recursive=True):
   """Return blob names under the given cloud storage path."""
-  for blob in _provider().list_blobs(
-      cloud_storage_path, recursive=recursive, names_only=True):
+  for blob in _provider().list_blobs(cloud_storage_path, recursive=recursive):
```
vitorguidi, line 1080: Seems a bit cryptic why removing names_only implies pagination stops being an issue, how can I repro a request with pagination?

jonathanmetzman: It's the filter that causes this

PR #4499, opened 2024-12-13
[Monitoring] Partition uworker_output.ErrorType conditions into success, maybe_retry and failure outcomes

vitorguidi:
### Motivation

#4458 implemented a task outcome metric, so we can track error rates in utasks, by job/task/subtask.

As failures are expected for ClusterFuzz, initially only unhandled exceptions would be considered as actual errors. Chrome folks asked for a better partitioning of error codes, which is implemented here as the following outcomes:

* success: the task has unequivocally succeeded, producing a sane result
* maybe_retry: some transient error happened, and the task is potentially being retried. This might capture some unretriable failure condition, but it is a compromise we are willing to make in order to decrease false positives.
* failure: the task has unequivocally failed.

Part of #4271

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
@@ -84,6 +84,52 @@ class _MetricRecorder(contextlib.AbstractContextManager):
     self._subtask = subtask
     self._labels = None
     self.utask_main_failure = None
+    self._utask_success_conditions = [
+      uworker_msg_pb2.ErrorType.NO_ERROR,
+      uworker_msg_pb2.ErrorType.ANALYZE_NO_CRASH,
+      uworker_msg_pb2.ErrorType.PROGRESSION_BAD_STATE_MIN_MAX,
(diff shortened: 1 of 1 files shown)


jonathanmetzman (commented):
Please revert this PR.
I don't want to have categorize these conditions, especially not where they are defined.

vitorguidi:
There is a cleaner way to do this:

Add two boolean fields to uworker_output.ErrorType:

- is_success
- is_retryable_response

This way we can propagate the data to _MetricRecorder without listing out enums. Will revert this and implement the alternative approach

vitorguidi:
The maybe_retry outcome is kind of useless, it comprises a tiny tiny volume among all outcomes. It is better to split between success and error only

@letitz 
image

vitorguidi:
> The maybe_retry outcome is kind of useless, it comprises a tiny tiny volume among all outcomes. It is better to split between success and error only
> 
> @letitz @alhijazi image

Even discarding fuzz task (which does not retry and comprises most of the task volume), this is still a very small fraction
image

letitz:
Fair enough, thanks for trying!

PR #4504, opened 2024-12-16
Add Fuzzilli cases to the test-input archive

mi-ac:
This adds 9 archives from Fuzzilli to the general test-input archives used by fuzzers on Clusterfuzz. The Fuzzilli-side archives are refreshed every few days. We'll add a freshness metric in a follow up.

This was tested locally with butler.

See bug: https://crbug.com/362963886

### DIFF
--- a/src/python/other-bots/chromium-tests-syncer/run.py
+++ b/src/python/other-bots/chromium-tests-syncer/run.py
@@ -23,6 +23,7 @@ modules.fix_module_search_paths()
 import os
 import re
 import subprocess
+import tarfile
 import time
 
 from clusterfuzz._internal.base import utils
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
Note that we need to merge this into the google branch as well.

src/python/other-bots/chromium-tests-syncer/run.py
```diff
@@ -208,6 +209,39 @@ def create_gecko_tests_directory(tests_directory, gecko_checkout_subdirectory,
                            target_subdirectory)
 
 
+def create_fuzzilli_tests_directory(tests_directory):
+  """Create Fuzzilli tests directory from the autozilli GCS archives."""
+  logs.info('Syncing fuzzilli tests.')
+  fuzzilli_tests_directory = os.path.join(tests_directory, 'fuzzilli')
+  remote_archive_tmpl = 'gs://autozilli/autozilli-%d.tgz'
```
jonathanmetzman, line 216: Is this readable by anyone?

jonathanmetzman (approved):
lgtm

src/python/other-bots/chromium-tests-syncer/run.py
mi-ac: See: https://issues.chromium.org/issues/362963886#comment21

mi-ac:
> Note that we need to merge this into the google branch as well.

Who'd need to do that? Is that part of the deployment now?

jonathanmetzman:
> > Note that we need to merge this into the google branch as well.
> 
> Who'd need to do that? Is that part of the deployment now?

For now, you would, sorry. Or you could merge your changes into chrome to begin with and avoid having to do two PRs.
It's part of the deploy process but isn't automatic.

PR #4511, opened 2024-12-16
Fix fetching paginated results with a filter.

jonathanmetzman:
Filters mean that pagination isn't handled for you.

### DIFF
--- a/src/clusterfuzz/_internal/google_cloud_utils/storage.py
+++ b/src/clusterfuzz/_internal/google_cloud_utils/storage.py
@@ -237,7 +237,6 @@ class GcsProvider(StorageProvider):
 
     client = _storage_client()
     bucket = client.bucket(bucket_name)
-    properties = {}
 
     if recursive:
       delimiter = None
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/google_cloud_utils/storage.py
```diff
@@ -249,23 +248,37 @@ def list_blobs(self, remote_path, recursive=True, names_only=False):
...
+              'bucket': bucket_name,
+              'name': prefix,
+          }
+          yield properties
+
+      next_page_token = iterator.next_page_token
+      if next_page_token is None:
+        break
+      if iterations % 50 == 0:
+        logs.error('Might be infinite looping.')
```
vitorguidi, line 280: Why this magic number?

src/clusterfuzz/_internal/google_cloud_utils/storage.py
```diff
@@ -249,23 +248,37 @@ def list_blobs(self, remote_path, recursive=True, names_only=False):
...
-
-      yield properties
-
-    if not recursive:
-      # When doing delimiter listings, the "directories" will be in `prefixes`.
-      for prefix in iterator.prefixes:
-        properties['bucket'] = bucket_name
-        properties['name'] = prefix
+    iterations = 0
+    while True:
```
vitorguidi, line 252: Is there no metadata in the http/grpc call that hints how many elements will be returned?

src/clusterfuzz/_internal/google_cloud_utils/storage.py
jonathanmetzman: Is there a better heuristic for guessing we are in an infinite loop :-)
It's an arbitrary number that I don't think should be realistically reached.

src/clusterfuzz/_internal/google_cloud_utils/storage.py
jonathanmetzman: I don't think so.

src/clusterfuzz/_internal/google_cloud_utils/storage.py
vitorguidi: How many elements show up per page?
What 

The heuristic approach is reasonable, but it would be nice to lay out in the open how many files at most are expected to be in a bucket (build/corpus), and why 50 * items_per_page is a good upper bound.

vitorguidi (approved):
lgtm, modulo making the upper bound assumption explicit

src/clusterfuzz/_internal/google_cloud_utils/storage.py
vitorguidi: Having this upper threshold and counting during the iteration might also be more resilient, if GCP changes the pagination behavior under the hood and you get, let's say, 10x less items, this might break in a very subtle way.

PR #4518, opened 2024-12-17
Modify get_fastboot_path() to allow for custom binaries

marktefftech:
Custom binaries are often used emulators (i.e. the binary being used by the host).

This code follows the same strategy used for the ADB binary above in `get_adb_path()`.


svasudevprasad (approved):

jonathanmetzman (approved):
lgtm

PR #4520, opened 2024-12-18
Fix off by one error in assertion

jonathanmetzman:
Also delete some example code.


jonathanmetzman:
TBR to fix breakage

oliverchang (commented):
lgtm

PR #4523, opened 2024-12-18
Use python's implementations for random_weighted_choice 

jonathanmetzman:
The existing implementation looked buggy. It was fine but let's use
Python's stdlib implementation so we can be more sure and reduce complexity.


oliverchang (approved):
nice!

jonathanmetzman:
Do we think it makes sense to delete the tests.

PR #4526, opened 2024-12-19
Enable skipping minimization with an env var

mi-ac:
This enables a job or fuzzer config to disable minimization with an environment variable. We can specify this at the additional environment variables of particular fuzzers like Binaryen, that don't work well with Clusterfuzz' built-in minimization attempts.

Bug: https://crbug.com/380264190


jonathanmetzman (approved):
lgtm
I didn't realize this would be so easy to implement.
Thanks!

PR #4528, opened 2024-12-19
Try scheduling as much as available

jonathanmetzman:
Instead of trying to schedule a small amount of fuzz tasks every 2 minutes and hoping this leads
to fuzzing at full capacity, just schedule almost the full amount at once.

### DIFF
--- a/src/clusterfuzz/_internal/cron/schedule_fuzz.py
+++ b/src/clusterfuzz/_internal/cron/schedule_fuzz.py
@@ -59,13 +59,15 @@ def get_available_cpus(project: str, region: str) -> int:
   assert preemptible_quota or cpu_quota
 
   if not preemptible_quota['limit']:
-    # Preemptible quota is not set. Obey the CPU quota since that limitss us.
+    # Preemptible quota is not set. Obey the CPU quota since that limits us.
     quota = cpu_quota
   else:
     quota = preemptible_quota
   assert quota['limit'], quota
 
-  return quota['limit'] - quota['usage']
+  # TODO(metzman): Do this in a more configurable way.
+  limit = min(quota['limit'], 100_000)
+  return limit - quota['usage']
 
 
 class BaseFuzzTaskScheduler:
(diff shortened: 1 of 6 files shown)


vitorguidi (approved):
lgtm

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -186,15 +202,38 @@ def get_batch_regions(batch_config):
   return list(set(config['gce_region'] for config in mapping.values()))
 
 
+def get_available_cpus(project: str, regions: List[str]) -> int:
+  """Returns the available CPUs for fuzz tasks."""
+  # TODO(metzman): This doesn't distinguish between fuzz and non-fuzz
+  # tasks (nor preemptible and non-preemptible CPUs). Fix this.
+  waiting_tasks = sum(
+      batch.count_queued_or_scheduled_tasks(project, region)
```
vitorguidi, line 210: One option to simulate these behaviors is https://simpy.readthedocs.io/
https://brooker.co.za/blog/2022/04/11/simulation.html

It is hard to imagine what this behavios imply, by just the idea

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -186,15 +202,38 @@ def get_batch_regions(batch_config):
...
+      for region in regions)
+  waiting_tasks += count_unacked(project, 'preprocess')
+  waiting_tasks += count_unacked(project, 'utasks_main')
+  soon_commited_cpus = waiting_tasks * CPUS_PER_FUZZ_JOB
+  logs.info(f'Soon committed CPUs: {soon_commited_cpus}')
+  available_cpus = sum(
+      get_available_cpus_for_region(project, region) for region in regions)
+  available_cpus = max(available_cpus - soon_commited_cpus, 0)
```
vitorguidi, line 218: Now that we take the queue size into account, we can go back to running this very frequently, right?

vitorguidi (approved):
lgtm. offered a design time option to anticipate real world behavior of these batch scheduling policies

src/clusterfuzz/_internal/cron/schedule_fuzz.py
jonathanmetzman: Hmmm...I had the opposite thought, that because we can schedule so many more at once, there's no need to run it so often. I think there can be a slight delay between publishing and reaching the queue, so probably above 5 minutes makes most sense.

PR #4530, opened 2024-12-20
Close old non reproducible bugs

pgrace-google:
Clusterfuzz will either confirm bugs are reproducible, mark them as non reproducible and close, or say nothing. In the last case, we want to auto close any bugs that have been open for long enough for it to have tried reproducing and failed


pgrace-google:
done! maybe 1d is too aggressive though

jonathanmetzman (approved):
LGTM

Btw, new bureaucracy we added is you also need to cherry pick this onto the chrome branch. We're deploying from there for now while master isn't so stable.

PR #4481, opened 2024-12-10
Extend TESTCASE_UPLOAD_TRIAGE_DURATION to account for fuzzer generated test cases

vitorguidi:
### Motivation

[#4364](https://github.com/google/clusterfuzz/pull/4364) implemented the tracking for the time it takes clusterfuzz to complete several steps of the manually uploaded testcase lifecycle.

As per Chrome's request, the metric will now contain an 'origin' label, which indicates if the testcase was 'manually_uploaded'  or generated by a 'fuzzer'

PR #4547, opened 2024-12-23
[Monitoring] Enrich UNTRIAGED_TESTCASE_AGE metric to track testcases stuck in analyze

vitorguidi:
### Motivation

We currently have no way to tell if analyze task was successfully executed. The TESTCASE_UPLOAD_TRIAGE_DURATION metric from #4364 would only track duration for tasks that did finish.

An analyze_pending field is added to the Testcase entity in datastore, which is set to False by default, to True for manually uploaded testcases, and to False once analyze task postprocess runs.

It also increments the UNTRIAGED_TESTCASE_AGE metric from #4381 with a status label, so we can know at what step the testcase is stuck, thus allowing us to alert if analyze is taking longer to finish than expected.

The alert itself could be, for instance, P50 age of untriaged testcase (status=analyze_pending) > 3h.

Also, this retroactively addresses comments from #4481:

* Fixes docstring for emit_testcase_triage_duration_metric
* Removes assertions
* Renames TESTCASE_UPLOAD_TRIAGE_DURATION to TESTCASE_TRIAGE_DURATION, since it now accounts for fuzzer generated testcases
* Use a boolean "from_fuzzer" field, instead of "origin" string, in TESTCASE_TRIAGE_DURATION

### DIFF
--- a/src/clusterfuzz/_internal/cron/triage.py
+++ b/src/clusterfuzz/_internal/cron/triage.py
@@ -352,7 +352,7 @@ def _emit_untriaged_testcase_age_metric(testcase: data_types.Testcase,
     return
 
   logs.info(f'Emiting UNTRIAGED_TESTCASE_AGE for testcase {testcase.key.id()} '
-            f'(age = {testcase.get_age_in_seconds()})')
+            f'(age = {testcase.get_age_in_seconds()}), step = {step}')
   monitoring_metrics.UNTRIAGED_TESTCASE_AGE.add(
       testcase.get_age_in_seconds() / 3600,
       labels={
--- a/src/clusterfuzz/_internal/cron/triage.py
+++ b/src/clusterfuzz/_internal/cron/triage.py
@@ -343,7 +343,8 @@ PENDING_GROUPING = 'pending_grouping'
 PENDING_FILING = 'pending_filing'
 
 
-def _emit_untriaged_testcase_age_metric(testcase: data_types.Testcase, step: str):
+def _emit_untriaged_testcase_age_metric(testcase: data_types.Testcase,
+                                        step: str):
   """Emmits a metric to track age of untriaged testcases."""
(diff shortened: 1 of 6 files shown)


src/clusterfuzz/_internal/metrics/monitoring_metrics.py
```diff
@@ -231,17 +231,17 @@
     ],
 )
 
-TESTCASE_UPLOAD_TRIAGE_DURATION = monitor.CumulativeDistributionMetric(
-    'uploaded_testcase_analysis/triage_duration_secs',
+TESTCASE_TRIAGE_DURATION = monitor.CumulativeDistributionMetric(
+    'testcase_analysis/triage_duration_secs',
```
alhijazi, line 235: Let's rename this to triage_duration_hours (since the unit is hours)

alhijazi (approved):

src/clusterfuzz/_internal/cron/triage.py
```diff
@@ -490,16 +497,15 @@ def main():
...
 
-    # A testcase is untriaged, until immediately before a bug is opened
-    _emit_untriaged_testcase_age_metric(testcase)
-    _set_testcase_stuck_state(testcase, False)
-    _increment_untriaged_testcase_count(testcase.job_type, PENDING_FILING)
-
     # File the bug first and then create filed bug metadata.
     if not _file_issue(testcase, issue_tracker, throttler):
+      _emit_untriaged_testcase_age_metric(testcase, PENDING_FILING)
+      _increment_untriaged_testcase_count(testcase.job_type, PENDING_FILING)
```
jonathanmetzman, line 503: There's a lot of stuff we're doing here for metrics. Is any of it tracked in tests?

jonathanmetzman (approved):
lgtm

PR #4548, opened 2024-12-23
Delete OSSFuzzClustersManager.

jonathanmetzman:
It's not used anymore now that everything is run using batch.
Bug: b/381528482


jonathanmetzman:
Closing this to reopen against oss-fuzz. I may merge since manage_vms is already broken.

PR #4551, opened 2024-12-23
Remove untrusted_runner from misc files.

jonathanmetzman:
It's not used anymore.

### DIFF
--- a/src/clusterfuzz/_internal/system/environment.py
+++ b/src/clusterfuzz/_internal/system/environment.py
@@ -815,11 +815,6 @@ def reset_environment():
     # Add shared variables with values from _initial_environment.
     os.environ.update(_initial_environment)
 
-  if is_trusted_host():
-    from clusterfuzz._internal.bot.untrusted_runner import \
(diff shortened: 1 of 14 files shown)


src/clusterfuzz/_internal/system/environment.py
```diff
@@ -705,12 +705,6 @@ def is_posix():
   return os.name == 'posix'
 
 
-def is_trusted_host(ensure_connected=True):
```
vitorguidi, line 708: There is also a test file for this, environment_test.py. It should also be changed, right?

vitorguidi: The same goes for UntrustedRunnerIntegrationTest

jonathanmetzman: The problem with these PRs is that I think the untrusted runner tests provide coverage for non-untrusted code that is otherwise not covered. I should try to convert these to normal tests rather than get rid of them.

PR #4553, opened 2024-12-26
Delete OSSFuzzClustersManager

jonathanmetzman:
It's not used anymore now that everything is run using batch.
Also, bump fuzz tasks scheduled so we can postprocess on tworkers.

Bug: b/381528482


vitorguidi (commented):
lgtm

PR #4558, opened 2024-12-26
[Chrome deployment] Split utasks properly into success and error, add analyze step to testcase triage metric

vitorguidi:
This merges #4547 and #4516 to the chrome branch.

### DIFF
--- a/src/clusterfuzz/_internal/datastore/data_types.py
+++ b/src/clusterfuzz/_internal/datastore/data_types.py
@@ -583,6 +583,10 @@ class Testcase(Model):
   # Tracks if a testcase is stuck during triage.
   stuck_in_triage = ndb.BooleanProperty(default=False)
 
+  # Tracks if analyze task is pending.
+  # Defaults to false, since most testcases are fuzzer produced.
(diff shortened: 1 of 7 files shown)


src/clusterfuzz/_internal/datastore/data_types.py
```diff
@@ -583,6 +583,10 @@ class Testcase(Model):
   # Tracks if a testcase is stuck during triage.
   stuck_in_triage = ndb.BooleanProperty(default=False)
 
+  # Tracks if analyze task is pending.
```
jonathanmetzman, line 586: Do we want to track if just analyze is pending and not analyze/minimize/regressoin?

vitorguidi: Minimize/regression/impact are tracked together under critical tasks. The idea for this, is so we would be able to catch that pubsub incident.

Since we already have quite a few stuck testcases under critical tasks, we do not gain much by splitting until we clean up the old stuff.

jonathanmetzman:
lgtm

PR #4562, opened 2024-12-27
Check for upload permission based on google group

pgrace-google:
Although the IssueTracker component create controls are locked down to this group, its possible for folks to accidentally migrate bugs into this component. To prevent that from happening, check for the reporter's membership before proceeding with the upload. It is a bit hacky i admit.

### DIFF
--- a/src/clusterfuzz/_internal/cron/external_testcase_reader.py
+++ b/src/clusterfuzz/_internal/cron/external_testcase_reader.py
@@ -16,6 +16,7 @@
 import datetime
 import re
 
+from google.cloud import storage
 import requests
 
 from appengine.libs import form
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/cron/external_testcase_reader.py
```diff
@@ -31,6 +32,15 @@
 ISSUETRACKER_WONTFIX_STATE = 'NOT_REPRODUCIBLE'
 
 
+def has_vrp_upload_permission(reporter):
+  """Checks whether the given reporter has permission to upload."""
+  storage_client = storage.Client()
+  bucket = storage_client.bucket('clusterfuzz-vrp-uploaders')
+  blob = bucket.blob('vrp-uploaders')
+  members = blob.download_as_string().decode('utf-8').splitlines()[0].split(',')
```
jonathanmetzman, line 40: Might be a good idea to have a YAML file but I'm OK with this.

src/clusterfuzz/_internal/cron/external_testcase_reader.py
```diff
@@ -31,6 +32,15 @@
 ISSUETRACKER_WONTFIX_STATE = 'NOT_REPRODUCIBLE'
 
 
+def has_vrp_upload_permission(reporter):
```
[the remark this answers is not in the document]
pgrace-google: I was thinking of running the cron job every 5 min or so, but it should only need to be called when we have submissions, which would depend on the usage of it

but 1. good call - i was being stupid and calling it for each check, but i should update this to be at least just one call for each cron job run lol

and 2. redis sounds like also a good plan! ill do that in a separate CL -ive added a todo for me!!

src/clusterfuzz/_internal/cron/external_testcase_reader.py
pgrace-google: true!! ive added a comment for me to do!

jonathanmetzman (approved):
lgtm

PR #4564, opened 2024-12-27
Bound preprocess and parallel signing calls

jonathanmetzman:
Bound preprocess to 30 minutes (shouldn't be more than 5)
Bound parallel signing to 2 minutes, exception if we pass this.

### DIFF
--- a/src/clusterfuzz/_internal/base/concurrency.py
+++ b/src/clusterfuzz/_internal/base/concurrency.py
@@ -49,7 +49,7 @@ def make_pool(pool_size=POOL_SIZE, cpu_bound=False, max_pool_size=None):
     else:
       yield futures.ThreadPoolExecutor(pool_size)
   else:
-    yield multiprocessing.Pool(pool_size)
+    yield futures.ProcessPoolExecutor(pool_size)
 
 
 # TODO(metzman): Find out if batching makes things even faster.
--- a/src/clusterfuzz/_internal/base/concurrency.py
+++ b/src/clusterfuzz/_internal/base/concurrency.py
@@ -21,22 +21,8 @@ from clusterfuzz._internal.system import environment
 POOL_SIZE = multiprocessing.cpu_count()
 
 
-class SingleThreadPool:
-  """Single thread pool for when it's not worth using Python's thread
-  implementation."""
-
-  def __init__(self, size):
-    del size
-
-  def map(self, f, l):
-    return list(map(f, l))
-
-  def imap_unordered(self, f, l):
-    return list(map(f, l))
-
-
 @contextlib.contextmanager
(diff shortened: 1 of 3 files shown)


src/clusterfuzz/_internal/base/concurrency.py
```diff
@@ -21,35 +21,18 @@
...
   if (environment.get_value('PY_UNITTESTS') or
       environment.platform() == 'WINDOWS'):
-    if cpu_bound:
-      yield SingleThreadPool(pool_size)
-    else:
-      yield futures.ThreadPoolExecutor(pool_size)
+    yield futures.ThreadPoolExecutor(pool_size)
   else:
-    yield multiprocessing.Pool(pool_size)
+    yield futures.ProcessPoolExecutor(pool_size)
```
vitorguidi, line 35: Why processes and not threads?

src/clusterfuzz/_internal/base/tasks/__init__.py
```diff
@@ -616,6 +656,9 @@ def run(self):
         if time_left <= 0:
           logs.info('Lease reached maximum lease time of {} seconds, '
                     'stopping renewal.'.format(self._max_lease_seconds))
+          if self._ack_on_timeout:
+            logs.info('Acking on timeout')
+            self._message.ack()
```
vitorguidi, line 661: Why acking? Why is this preferred to exiting and retrying later?

src/clusterfuzz/_internal/base/concurrency.py
jonathanmetzman: This is mostly used for CPU bound tasks such as cryptographic signing. In addition, Python's threads suck and processes are usually faster for things like I/O bound stuff.

src/clusterfuzz/_internal/google_cloud_utils/storage.py
```diff
@@ -1377,17 +1378,24 @@ def get_arbitrary_signed_upload_url(remote_directory):
       get_arbitrary_signed_upload_urls(remote_directory, num_uploads=1))[0]
 
 
-def maybe_parallel_map(func, arguments):
+def parallel_map(func, argument_list):
```
vitorguidi, line 1381: The if statement is being removed because the rearch is complete, right?

src/clusterfuzz/_internal/base/tasks/__init__.py
jonathanmetzman: For fuzz tasks, I'd rather things starve for now than gum up the queue.

src/clusterfuzz/_internal/google_cloud_utils/storage.py
jonathanmetzman: Correct!

PR #4565, opened 2024-12-27
[Monitoring] Make monitoring reference the secondary project

vitorguidi:
Chrome is split in two projects internally, and the current way terraform is organized will create the dashboard in the wrong project.

This PR makes the dashboard reference the secondary project id, thus solving the problem.

Part of #4271

### DIFF
--- a/infra/terraform/monitoring.tf
+++ b/infra/terraform/monitoring.tf
@@ -12,7 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+provider "google" {
+  alias = "monitoring"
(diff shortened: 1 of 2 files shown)


jonathanmetzman (approved):
lgtm

jonathanmetzman:
I think the name secondary_project isn't so great, maybe "display_project" because it's the project where the dashboard is displayed.

vitorguidi:
> I think the name secondary_project isn't so great, maybe "display_project" because it's the project where the dashboard is displayed.

we might want to create more resources in this second project in the future, not only metrics. secondary catches this more general intention

PR #4572, opened 2025-01-02
[Monitoring] Make appengine crons emit metrics

vitorguidi:
Appengine crons are currently not instrumented, so we cannot get testcase metrics for the google 3 deployment. This PR fixes that.

Part of #4271


jonathanmetzman (approved):
lgtm

PR #4573, opened 2025-01-02
Ensure we don't rerun utaks on expired signed URLs for uworker_input.

jonathanmetzman:
If the URL is expired, we can't possibly complete the task. Doing this though requries a few steps:
1. Note that we fail to download a signed URL because of an expired token.
2. Note that the utask failed because of failing to download input due to signed URL expiration.
3. Ensure bots don't retry when encountering this error.


vitorguidi (approved):
The approach is solid, lgtm'ing on the basis of we cannot know beforehand if this is actually the issue

PR #4574, opened 2025-01-02
Fix catching of own exception

jonathanmetzman:


jonathanmetzman:
Errr wrong code pushed. One sec

vitorguidi (approved):

jonathanmetzman:
Here it is

PR #4575, opened 2025-01-02
Force unit tests to run before deployments

vitorguidi:
Followup for the analyze/triage chrome incident

### DIFF
--- a/src/local/butler/deploy.py
+++ b/src/local/butler/deploy.py
@@ -530,6 +530,8 @@ def execute(args):
   _enforce_tests_pass()
   _enforce_safe_day_to_deploy()
 
+  assert False
+
   # Build templates before deployment.
   appengine.build_templates()
 
--- a/src/local/butler/deploy.py
+++ b/src/local/butler/deploy.py
@@ -459,6 +459,9 @@ def _is_safe_deploy_day():
 
 
 def _enforce_tests_pass():
+  config = local_config.Config()
+  if not config.get('project.enforce_tests_before_deploy', False):
+    return
   py_unittest.run_tests(target='core', parallel=True)
   py_unittest.run_tests(target='appengine', parallel=True)
 
--- a/src/local/butler/deploy.py
+++ b/src/local/butler/deploy.py
@@ -530,8 +530,6 @@ def execute(args):
   _enforce_tests_pass()
   _enforce_safe_day_to_deploy()
 
-  assert False
-
   # Build templates before deployment.
   appengine.build_templates()
 
--- a/src/local/butler/deploy.py
+++ b/src/local/butler/deploy.py
@@ -29,6 +29,7 @@ from local.butler import appengine
 from local.butler import common
 from local.butler import constants
 from local.butler import package
+from local.butler import py_unittest
 from src.clusterfuzz._internal.base import utils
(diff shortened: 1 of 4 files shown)


src/local/butler/py_unittest.py
```diff
@@ -215,7 +217,12 @@ def run_tests_parallel(args, test_directory, top_level_dir):
     sys.exit(1)
 
 
-def execute(args):
+def run_tests(target=None,
```
vitorguidi, line 220: This was moved just so butler/deploy.py can access it as a public method, and to break the dependency on the arg object from the parser.

oliverchang (approved):

src/local/butler/deploy.py
```diff
@@ -457,6 +458,11 @@ def _is_safe_deploy_day():
   return day_now_in_ny not in {4, 5, 6}  # The days of the week are 0-indexed.
 
 
+def _enforce_tests_pass():
+  py_unittest.run_tests(target='core', parallel=True)
```
oliverchang, line 462: Actually, we need a flag to disable this, so that our cloudbuild deployments (where tests are already run prior to deploy). 

See cloudbuild.yaml in clusterfuzz-config.

oliverchang (changes requested):

src/local/butler/deploy.py
vitorguidi: It is unclear: what happens in cloudbuild deployments that depends on butler.py?

From reading the yaml, it seems like "pipenv deploy" is executed, this has nothing to do with butler right?

oliverchang: The deploy script ends up ruinning butler: https://github.com/google/clusterfuzz/blob/master/docker/ci/deploy

vitorguidi: I think there is a simpler approach. We can do the unit tests conditionally only for the chrome deployment, as suggested by @jonathanmetzman 

Once we complete the deployment work, butler deploy will be deprecated, so under no conditions an artifact will be produced without the tests succeeding.

I would rather avoid touching the CI CD stuff, since this is only a temporary incident followup requested by a subset of users.

jonathanmetzman: The idea sounds like something I would suggest because I'm trying to minimize the impact of this.
I don't think it matters very much either way, but I am leaning towards this change being unnecessary.
It was done to deal with the multibranch situation which encouraged merging code (that was approved to be merged into another branch) without PRs (since who wants to open a PR twice), which led to one breakage.
The multibranch situation should be gone in two weeks I'm guessing, so I think this PR is a slight overcorrection (I think it was originally my suggestion even) that won't really have much impact beyond making deploys take longer (which is bad!).

src/local/butler/deploy.py
```diff
@@ -457,6 +458,14 @@ def _is_safe_deploy_day():
   return day_now_in_ny not in {4, 5, 6}  # The days of the week are 0-indexed.
 
 
+def _enforce_tests_pass():
+  config = local_config.Config()
+  if not config.get('project.enforce_tests_before_deploy', False):
```
vitorguidi, line 463: there is an accompanying PR in the configs repo for this

oliverchang (approved):

PR #4577, opened 2025-01-02
Make schedule_fuzz conform to cronjob API.

jonathanmetzman:
Our cronjobs are expected to return. Don't infite loop in schedule_fuzz.


jonathanmetzman:
Merging to fix deployment.

PR #4580, opened 2025-01-04
Update issue type and component for VRP uploaded bugs

pgrace-google:
Question 0: dont review this yet im running tests just to make sure its right
Question 1: i imagine there is a better place to put the component ids, but i havent been able to find it. where is it??


src/clusterfuzz/_internal/cron/cleanup.py
```diff
@@ -1056,6 +1059,19 @@ def update_component_labels_and_id(policy, testcase, issue):
   component_id = _get_predator_result_item(testcase,
                                            'suspected_buganizer_component_id')
 
+  # These bugs were filed via the direct upload and hence specific to the
+  # google issue tracker.
+  if hasattr(issue,
```
jonathanmetzman, line 1064: when would it not have component_id?
You can replace this with getattr(issue, 'component_id', None) == VRP_UPLOAD_COMPONENT_ID.

pgrace-google: i believe an issue wouldnt have component_id for non google issue tracker (monorail, jira) 

but ooh thank you done!

pgrace-google:
(just rebasing)

PR #4583, opened 2025-01-07
[centipede] add support for gathering stats for centipede runs

paulsemel:
This PR adds supports so that CF supports Centipede stats. This will help understand better how centipede fuzzers are performing on ClusterFuzz.

### DIFF
--- a/src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
@@ -14,10 +14,15 @@
 """Centipede engine interface."""
 
 from collections import namedtuple
+import csv
 import os
 import pathlib
 import re
 import shutil
+from typing import Dict
+from typing import List
+from typing import Optional
+from typing import Union
 
 from clusterfuzz._internal.bot.fuzzers import dictionary_manager
 from clusterfuzz._internal.bot.fuzzers import engine_common
@@ -28,6 +33,7 @@ from clusterfuzz._internal.metrics import logs
 from clusterfuzz._internal.system import environment
 from clusterfuzz._internal.system import new_process
 from clusterfuzz.fuzz import engine
+from clusterfuzz.stacktraces import constants as stacktraces_constants
 
 _CLEAN_EXIT_SECS = 10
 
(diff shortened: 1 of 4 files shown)


paulsemel:
Note: two very interesting metrics won't be added yet: `new_features` and `new_tests_added`. I want to do that in a second time, because that's more work and that'll allow splitting the PR in two different PRs. It's ok since we allow adding nullable fields to the schema, so we just won't have those data when the first PR lands.

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
```diff
@@ -72,9 +78,68 @@ def _set_sanitizer_options(fuzzer_path):
...
+
+  Returns:
+      a dictionary containing the stats.
+  """
+  if not os.path.exists(stats_file):
+    return None
+  with open(stats_file, 'r') as statsfile:
+    csvreader = csv.reader(statsfile)
+    l = list(csvreader)
+    # If the binary could not run at all, the file will be empty or with only
```
oliverchang, line 97: should we also handle exceptions generally?

I'd expect there to be (hopefully rare) cases the file is malformed because centipede was killed in the middle of writing the file.

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
```diff
@@ -72,9 +78,68 @@ def _set_sanitizer_options(fuzzer_path):
...
+    # the column description line.
+    if len(l) <= 1:
+      return None
+    return {
+        l[0][i]: float(l[-1][i]) if '.' in l[-1][i] else int(l[-1][i])
```
oliverchang, line 102: Can you add a comment to explain what this is doing? Or to include an example line that we're parsing? 

Also -- is "." reliable enough of an indicator that something is a float? e.g. is "1e-1" a possible input here? 

It'll likely be a little more robust if we just use `isdigit` to check for int validity instead.

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
paulsemel: Yes, I think you're right, let's be careful here and gracefully fail.

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
paulsemel: Arg.. I meant to rework the function but actually forgot to do it. WDYT now?

oliverchang (approved):
thanks!

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
```diff
@@ -72,9 +78,79 @@ def _set_sanitizer_options(fuzzer_path):
...
+  Args:
+      stats_file: the path to Centipede stats file.
+
+  Returns:
+      a dictionary containing the stats.
+  """
+  try:
+    with open(stats_file, 'r') as statsfile:
+      csvreader = csv.reader(statsfile)
+      rows = list(csvreader)
```
jonathanmetzman, line 95: nit: can unindent after this point.

jonathanmetzman (approved):
LGTM

There's an additional step where I think we have to edit something in datastore (?) to get the stats to show up on fuzzer-stats.

PR #4584, opened 2025-01-07
Log time series that failed to upload

vitorguidi:
We are currently facing the following error:

```
Duplicate TimeSeries encountered. Only one point can be written per TimeSeries per request.
```

It probably happens because the custom metric store/upload stuff is missing this restriction in [create time series request](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.timeSeries/create).

In order to further troubleshoot the issue, logging is required, to figure out exactly what metric is being published with >1 point.


jonathanmetzman (approved):
lgtm

PR #4585, opened 2025-01-07
Allow scheduling more fuzz tasks at one time.

jonathanmetzman:


oliverchang (approved):
any rationale behind this number?  maybe a comment?

PR #4588, opened 2025-01-08
Fix trimmed labels being used in other metrics, due to shallow copy

vitorguidi:
This  addresses the errors on b/384820142

The main issue is that, in _MetricsRecorder, trimmed_labels was a shallow copy for self._labels, which had job removed, thus causing many utask duration with different jobs to reduce to the same labels (deduplication seems to only take into account the labels declared in monitoring_metrics.py)

The other error is related to saturation, where time series would be written too often. This probably happens due to the loss of the job label, causing many requests to a subset of the original labels. It can possibly go away once this lands, otherwise more investigation is needed.


jonathanmetzman (approved):
lgtm

PR #4591, opened 2025-01-08
Fix dashboard escape for $ character in  promql queries

vitorguidi:
The last terraform definition was wrongly escaping the curly braces with "\\". The actual intended way to do it is to double the dollar sign

Stack overflow reference: https://stackoverflow.com/questions/55796685/terraform-escaping-bash-command-substitution

The concrete problem that this causes, is that every promql query would fail in the dashboard

Testing strategy: terraform validate ran successfully

image


jonathanmetzman (approved):
lgtm

vitorguidi:
> Member
It worked
image

PR #4601, opened 2025-01-14
[batch] Support projects with different disk sizes

jonathanmetzman:
We may need to do this and have per-job templates because chrome needs it. Maybe I can get rid of this and replace it with per-job templates and make every oversized oss-fuzz project use the same config with the same disk size (right now it varies).

### DIFF
--- a/src/clusterfuzz/_internal/cron/project_setup.py
+++ b/src/clusterfuzz/_internal/cron/project_setup.py
@@ -752,8 +752,14 @@ class ProjectSetup:
         return base_project
     return project
 
-  def _sync_job(self, project, info, corpus_bucket_name, quarantine_bucket_name,
-                logs_bucket_name, backup_bucket_name, oss_fuzz_gb=None):
+  def _sync_job(self,
+                project,
+                info,
+                corpus_bucket_name,
+                quarantine_bucket_name,
+                logs_bucket_name,
+                backup_bucket_name,
+                oss_fuzz_gb=None):
     """Sync the config with ClusterFuzz."""
     # Create/update ClusterFuzz jobs.
     job_names = []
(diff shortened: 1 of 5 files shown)


src/clusterfuzz/_internal/cron/project_setup.py
```diff
@@ -808,11 +809,14 @@ def _sync_job(self, project, info, corpus_bucket_name, quarantine_bucket_name,
             project, info, template.engine, template.memory_tool,
             template.architecture)
       base_project_name = self._get_base_project_name(project)
+      oss_fuzz_project = ndb.Key(data_types.OssFuzzProject, project).get()
```
oliverchang, line 818: do we need to add a check for OSS-Fuzz here, and only do this for OSS-Fuzz? otherwise we'll get None errors?

oliverchang: i misread the code -- disregard.

src/clusterfuzz/_internal/cron/project_setup.py
```diff
@@ -751,8 +752,14 @@ def _get_base_project_name(self, project):
...
-  def _sync_job(self, project, info, corpus_bucket_name, quarantine_bucket_name,
-                logs_bucket_name, backup_bucket_name):
+  def _sync_job(self,
+                project,
+                info,
+                corpus_bucket_name,
+                quarantine_bucket_name,
+                logs_bucket_name,
+                backup_bucket_name,
+                oss_fuzz_gb=None):
```
oliverchang, line 762: what's passing this  in? do we need this as an argument?

oliverchang (approved):

src/clusterfuzz/_internal/cron/project_setup.py
jonathanmetzman: Ah I messed up a cherry pick. Let me fix this.

PR #4603, opened 2025-01-14
centipede: fix average_exec_per_sec stat

paulsemel:


jonathanmetzman (approved):
lgtm

jonathanmetzman:
Could you fix the failing intergration tests

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
```diff
@@ -313,6 +313,9 @@ def fuzz(self, target_path, options, reproducers_dir, max_time):  # pylint: disa
         'actual_duration': actual_duration,
         'fuzzing_time_percent': fuzzing_time_percent,
     })
+    fuzz_time_secs_avg = stats.get('FuzzTimeSec_Avg', 0.0)
+    num_execs_avg = stats.get('NumExecs_Avg', 0.0)
+    stats['average_exec_per_sec'] = num_execs_avg / fuzz_time_secs_avg
```
letitz, line 318: If `FuzzTimeSec_Avg` is absent, or zero, you will divide by zero here. Default it to 1.0 instead?

letitz: You can still divide by zero here if FuzzTimeSec_Avg is present and set to 0, but that might be impossible.

paulsemel: You're right, let's be cautious here.

paulsemel:
I think we're good here now. One test is failing, but that's not related to that change I believe.

PR #4604, opened 2025-01-14
Delete cronjob used for gRPC

jonathanmetzman:
We don't need these certs anymore now that we aren't using gRPC


vitorguidi (approved):
lgtm, but the cron needs manual deleting in k8s. Butler does not destroy them if absent

jonathanmetzman:
Done.

PR #4606, opened 2025-01-14
Check for more preconditions before deploying

letitz:
Specifically:

* set up application-default gcloud credentials
* if deploying to k8s, check that `kubectl` and `terraform` are in the `PATH`

### DIFF
--- a/src/local/butler/deploy.py
+++ b/src/local/butler/deploy.py
@@ -533,6 +533,9 @@ def execute(args):
   _enforce_tests_pass()
   _enforce_safe_day_to_deploy()
 
+  # Needed for some subsequent steps.
+  common.execute('gcloud auth application-default login')
+
   # Build templates before deployment.
   appengine.build_templates()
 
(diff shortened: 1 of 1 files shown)


vitorguidi (approved):
lgtm

src/local/butler/deploy.py
```diff
@@ -533,6 +533,9 @@ def execute(args):
   _enforce_tests_pass()
   _enforce_safe_day_to_deploy()
 
+  # Needed for some subsequent steps.
+  common.execute('gcloud auth application-default login')
```
jonathanmetzman, line 537: Hmmm...We don't use the gcb deploy so much but I'm not sure this works there. Also does it force me to login each deploy?

jonathanmetzman (changes requested):

src/local/butler/deploy.py
letitz: I think it would force you to log in on each deploy, yes. That was required to deploy to appengine when I tried - the culprit was `python butler.py run setup` on line 438.

jonathanmetzman: Let's leave this out please. It might save you some time but it will break out GCB deploy and make things more annoying for me to deploy (particularly because I SSH into the machine I deploy from).

PR #4610, opened 2025-01-15
Don't include chrome restrictions on uploads in oss-fuzz

jonathanmetzman:
Don't force users to sign trusted agreement.


src/appengine/handlers/upload_testcase.py
```diff
@@ -393,6 +393,7 @@ def do_post(self):
         'trustedAgreement') == TRUSTED_AGREEMENT_TEXT.strip()
 
     if (not trusted_agreement_signed and
+        utils.is_chromium() and
```
vitorguidi, line 396: This is not self evident:
* Why does this check exist?
* Why does a trust agreement need signing?
* Why is it ok to perform untrusted (?) uploads in oss-fuzz, but not in chrome? (ie, this change)

jonathanmetzman: Done

PR #4620, opened 2025-01-16
Revert #4575 and #4612 (#4616)

jonathanmetzman:
We've decided they are unnecessary.
More importantly, they are breaking deploys.


vitorguidi (approved):
lgtm

PR #4626, opened 2025-01-21
Avoid overwriting minimized keys if minimization round fails.

letitz:
Previous logic would overwrite `minimized_keys` if say round 5 of libfuzzer minimization failed, even if round 4 succeeded. This resulted in tasks completing minimization successfully, but not storing `minimized_keys` to reflect that in the database.

Instead, we should take the result of the last successful minimization round. This both aligns with the logic around crash results and testcase file names, and prevents another side effect of this bug: we were previously deleting blobs on GCS that had been uploaded during successful minimization rounds :/

Note: it seems a similar bug affects the cleanse step, but I did not change logic there as it is OSS-Fuzz specific and I was not sure whether or not it could be intentional. See [`minimize_task.py` line 1686](https://github.com/google/clusterfuzz/pull/4626/files#diff-e8255271eeeadc2bda46b215fbc7b0bb160ef1116f6e27ff895990d88caafa5fR1686).

There are exceedingly few tests for minimize task, so I did not write a test for this either - the effort required seemed too high.

Bug: https://crbug.com/389589679

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
@@ -1370,9 +1370,9 @@ def do_js_minimization(test_function, get_temp_file, data, deadline, threads,
 
 
 def _run_libfuzzer_testcase(fuzz_target,
-                            testcase,
-                            testcase_file_path,
-                            crash_retries=1):
+                            testcase: data_types.Testcase,
+                            testcase_file_path: str,
+                            crash_retries: int = 1) -> CrashResult:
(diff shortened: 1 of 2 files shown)


alhijazi (approved):

vitorguidi (approved):
The only business critical path that this touches seems to be grouper

https://github.com/google/clusterfuzz/blob/41e5571a97f270d420ab6e42585262a631b1fe62/src/clusterfuzz/_internal/cron/grouper.py#L406

lgtm, since the immediate consequence is that we will stop filing bugs that are actually duplicates, if anything goes bad

letitz:
Thanks! Merging.

PR #4628, opened 2025-01-21
Merge deploy reverts to chrome

vitorguidi:
This merges #4624 and #4616 into chrome, to unblock deployments

### DIFF
--- a/src/local/butler/deploy.py
+++ b/src/local/butler/deploy.py
@@ -27,7 +27,6 @@ from local.butler import appengine
 from local.butler import common
 from local.butler import constants
 from local.butler import package
-from local.butler import py_unittest
 from src.clusterfuzz._internal.base import utils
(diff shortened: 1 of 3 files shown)


src/local/butler/deploy.py
```diff
@@ -456,14 +455,6 @@ def _is_safe_deploy_day():
   return day_now_in_ny not in {4, 5, 6}  # The days of the week are 0-indexed.
```
letitz, line 455: We're still missing the fix: we need to import pytz and fix the signature of the `now()` function.

letitz (changes requested):

letitz (approved):
LGTM, thanks!

letitz:
Merged to unblock https://github.com/google/clusterfuzz/pull/4627.

PR #4634, opened 2025-01-23
Update min_revision so that progression task can progress without errors

svasudevprasad:
If the min_revision (based on last_tested_crash_revision or known_crash_revision) no longer exists, then use the first revision in the list so that progression does not error out with "Build not found"


jonathanmetzman (approved):
LGTM

PR #4638, opened 2025-01-24
Clean up external_testcase_reader for testing round 1

pgrace-google:
removing some of the safeguards to run a prod test

### DIFF
--- a/src/clusterfuzz/_internal/cron/external_testcase_reader.py
+++ b/src/clusterfuzz/_internal/cron/external_testcase_reader.py
@@ -174,7 +174,6 @@ def handle_testcases(tracker, config):
   if len(issues) == 0:
     return
 
-  # TODO(pgrace) Cache in redis.
   vrp_uploaders = get_vrp_uploaders(config)
 
   # Rudimentary rate limiting -
@@ -194,8 +193,7 @@ def handle_testcases(tracker, config):
       continue
 
     # Submit valid testcases.
-    # TODO(pgrace) replace with 0 once testing is complete
-    attachment_metadata = attachment_metadata[6]
+    attachment_metadata = attachment_metadata[0]
     attachment = tracker.get_attachment(
         attachment_metadata['attachmentDataRef']['resourceName'])
     submit_testcase(issue.id, attachment, attachment_metadata['filename'],


configs/test/gae/prod3/cron.yaml
```diff
@@ -34,6 +34,11 @@ cron:
     schedule: every 1 minutes
     target: cron-service
 
+  - description: Handle external testcases.
+    url: /external-testcase-reader
+    schedule: every 6 hours
+    target: cron-service
```
vitorguidi, line 40: Any particular reason to have this in GAE?
AFAIK our crons only run on kubernetes

src/appengine/handlers/cron/external_testcase_reader.py
```diff
@@ -0,0 +1,26 @@
...
+from clusterfuzz._internal.cron import external_testcase_reader
+from handlers import base_handler
+from libs import handler
+
+
+class Handler(base_handler.Handler):
+  """Handle externally submitted testcases."""
+
+  @handler.cron()
+  def get(self):
```
vitorguidi, line 25: ditto

configs/test/gae/prod3/cron.yaml
pgrace-google: no not at all - its because my abilities cap at finding the first patching pattern. let me fix that lol. I remember we discussed most (if not all) cron jobs are in k8s, but then I promptly ignored that apparently when i found this file.

i kept the server.py changes for appengine in case it becomes useful for testing, because I think we said appengine cron jobs are used for testing! lmk if that doesnt make any sense

src/appengine/handlers/cron/external_testcase_reader.py
pgrace-google: ill keep this file in case we need for testing, unless that doesnt make any sense!

src/appengine/server.py
```diff
@@ -112,6 +113,7 @@ def register_routes(flask_app, routes):
 cron_routes = [
     ('/cleanup', cleanup.Handler),
     ('/external-update', external_update.Handler),
+    ('/external-testcase-reader', external_testcase_reader.Handler),
```
vitorguidi, line 116: Not really sure what this is for, @jonathanmetzman  or @oliverchang know better.

src/appengine/handlers/cron/external_testcase_reader.py
vitorguidi: There are ways to test it locally, invoking it through the debugger
https://github.com/google/clusterfuzz/blob/8270d0b063aedb5b372f35f55e5f556f672487b6/src/local/butler/scripts/run_cron.py#L4

src/appengine/server.py
pgrace-google: i thin if the appengine cron job isnt needed for testing this seems irrelevant?? removing!

src/appengine/handlers/cron/external_testcase_reader.py
pgrace-google: that had the problem of ACLs 😿

jonathanmetzman (approved):
LGTM

jonathanmetzman:
Can you fix the merge conflicts?

pgrace-google:
done!

Issue #3347, opened 2023-09-20
Create Subqueues to Enable Device-Specific Fuzzing and Testcases


marktefftech:
<h2> Problem </h2>

There is a mismatch between the original device that found a testcase and the new device now trying to reproduce that testcase. Here is what it looks like in the bot logs:

```
2023-09-20 10:48:16,693 - run_bot - INFO - Executing command "progression 6251137650982912 libfuzzer_hwasan_android_device"
2023-09-20 10:48:17,102 - run_bot - INFO - Testcase 6251137650982912 platform (<platform>:<device>:<branch>) does not match with ours (<platform>:<device>:<branch>), exiting
2023-09-20 10:48:17,502 - run_bot - INFO - Task complete, stopping renewal.
```

<h2> Solution </h2>

[Code](https://github.com/google/clusterfuzz/pull/3306/)

This feature request is used to create device-specific queues in ClusterFuzz, so that:

1. A fuzzer can be designed to run only on specific devices
2. Bots will subscribe to queues that contain compatible testcases only

Steps:

* Add a `projects.json` file to your ClusterFuzz environment, which will be consumed by the hourly cron job located in `src/appengine/handlers/cron/project_setup.py`
* Configure the QUEUE_OVERRIDE variable in the format of `<PLATFORM>:<DEVICE>`
* * The semicolon (':') delimiter is known to ClusterFuzz as a SUBQUEUE_IDENTIFIER, indicating the need for device-specific task queues.

marktefftech:
@jonathanmetzman is it possible to get triage access (per [github](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels#applying-a-label)) so I can add labels?

jonathanmetzman:
Done!

marktefftech:
Thank you for the quick response. Maybe the permissions take time to propagate, but I still cannot add labels per the docs linked above.

jonathanmetzman:
Let me know what access you need then and I can give it

marktefftech:
Apologies I never saw this, but I had to accept access via an email invitation.

PR #4643, opened 2025-01-27
Support engineless fuzzers in project setup

ParisMeuleman:
We want to use subqueues #3347 for our android jobs/bots as we waste a lot of machine time picking up incompatible tasks.

Currently the subqueues can be only assigned through project setup.
Our Android fuzzing uses engineless fuzzers, which are not supported by project setup.
This adds support for the later.


jonathanmetzman (approved):
lgtm

jonathanmetzman:
Can you fix the failing test.

ParisMeuleman:
> Also, you should be committing directly to master, the plan is (hopefully) to drop the chrome branch since all of its additional commits are supposed to be cherry picks.

I wasn't sure, I went for Chrome first to avoid any impacts on internal/oss fuzz as they are heavily using project setup and not chrome.

jonathanmetzman:
That makes sense.
It looks like the chrome branch is going to stay gone, I got rid of it this morning, so my advice is even more necessary :-)

ParisMeuleman:
@jonathanmetzman shall we move forward with this ?

PR #4645, opened 2025-01-28
Try to improve corpus pruning preprocess time.

jonathanmetzman:
1. Limit corpus signing to 50,000 URLs.
2. Treat a lack of successful corpus pruning as a semi-failure and force removal of corpus files in this case.
3. Only fetch the names of files when only the names are used, to improve speed.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/corpus_pruning_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/corpus_pruning_task.py
@@ -1097,12 +1097,12 @@ def utask_preprocess(fuzzer_name, job_type, uworker_env):
       last_execution_metadata and
       last_execution_metadata.status == data_types.TaskState.FINISHED)
 
-  # # Make sure we're the only instance running for the given fuzzer and
-  # # job_type.
-  # if not data_handler.update_task_status(task_name,
-  #                                        data_types.TaskState.STARTED):
-  #   logs.info('A previous corpus pruning task is still running, exiting.')
-  #   return None
+  # Make sure we're the only instance running for the given fuzzer and
+  # job_type.
+  if not data_handler.update_task_status(task_name,
+                                         data_types.TaskState.STARTED):
+    logs.info('A previous corpus pruning task is still running, exiting.')
+    return None
 
   setup_input = (
       setup.preprocess_update_fuzzer_and_data_bundles(fuzz_target.engine))
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/corpus_pruning_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/corpus_pruning_task.py
@@ -160,7 +160,7 @@ async def _limit_corpus_size(corpus_url):
       assert deleting
       blobs_to_delete.append(blob)
       if len(blobs_to_delete) == GOOGLE_CLOUD_MAX_BATCH_SIZE:
-        delete_success = await fast_http.delete_batch(
+        delete_success = await fast_http.delete_gcs_blobs_batch(
             session, bucket, blobs_to_delete, creds.token)
         if delete_success:
           num_deleted += GOOGLE_CLOUD_MAX_BATCH_SIZE
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/corpus_pruning_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/corpus_pruning_task.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Corpus pruning task."""
+import asyncio
 import collections
 import datetime
 import json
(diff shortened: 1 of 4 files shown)


vitorguidi:
> 1. Limit corpus signing to 50,000 URLs.
> 2. Treat a lack of successful corpus pruning as a semi-failure and force removal of corpus files in this case.
> 3. Only fetch the names of files when only the names are used, to improve speed.

This ultimately intends to bump up our fuzzing hours for oss fuzz, right?
Would be nice to see the profiling results that motivated this change, if you still have that 👀

src/clusterfuzz/_internal/bot/tasks/utasks/corpus_pruning_task.py
```diff
@@ -1076,12 +1083,13 @@ def utask_preprocess(fuzzer_name, job_type, uworker_env):
   # If our last execution failed, shrink to a randomized corpus of usable size
   # to prevent corpus from growing unbounded and recurring failures when trying
   # to minimize it.
-  if last_execution_failed:
+  if last_execution_failed or last_execution_not_successful:
```
vitorguidi, line 1086: Why OR'ing here? Failing and not success are synonyms, right?

jonathanmetzman:
> > 1. Limit corpus signing to 50,000 URLs.
> > 2. Treat a lack of successful corpus pruning as a semi-failure and force removal of corpus files in this case.
> > 3. Only fetch the names of files when only the names are used, to improve speed.
> 
> This ultimately intends to bump up our fuzzing hours for oss fuzz, right? Would be nice to see the profiling results that motivated this change, if you still have that 👀

This is a good point, but I don't have profiling results because the scale here was so absurd I gave up waiting. cjson_read_fuzzer had 4.5 million testcases

src/clusterfuzz/_internal/bot/tasks/utasks/corpus_pruning_task.py
jonathanmetzman: Not quite. Not attempting is not succeeding, it is also not failing.
This distinction is proably stupid and I can get rid of it. But I'm scared of doing very destructive actions (deleting testcases if there are more than 10k) in more cases so I make the distinction.

src/clusterfuzz/_internal/bot/tasks/utasks/corpus_pruning_task.py
```diff
@@ -118,18 +118,22 @@ def _get_corpus_file_paths(corpus_path):
...
+def _limit_corpus_size(corpus_url, not_failed):
   """Limit number of files and size of a corpus."""
   corpus_count = 0
   corpus_size = 0
   deleted_corpus_count = 0
   bucket, _ = storage.get_bucket_name_and_path(corpus_url)
   logs.info('Limiting corpus size.')
+  size_limit = CORPUS_SIZE_LIMIT_FOR_FAILURES
+  num_files_limit = CORPUS_FILES_LIMIT_FOR_FAILURES
+  if not_failed:
```
oliverchang, line 130: can you add a comment to explain this logic a bit more?

src/clusterfuzz/_internal/bot/tasks/utasks/corpus_pruning_task.py
```diff
@@ -118,18 +118,22 @@ def _get_corpus_file_paths(corpus_path):
   ]
 
 
-def _limit_corpus_size(corpus_url):
+def _limit_corpus_size(corpus_url, not_failed):
```
oliverchang, line 121: nit: s/not_failed/last_execution_failed/ to be clearer.

src/clusterfuzz/_internal/bot/tasks/utasks/corpus_pruning_task.py
```diff
@@ -118,18 +118,22 @@ def _get_corpus_file_paths(corpus_path):
...
+  num_files_limit = CORPUS_FILES_LIMIT_FOR_FAILURES
+  if not_failed:
+    size_limit *= 5
+    num_files_limit *= 5
   for corpus_file in storage.get_blobs(corpus_url):
     corpus_count += 1
     corpus_size += corpus_file['size']
-    if (corpus_count > CORPUS_FILES_LIMIT_FOR_FAILURES or
-        corpus_size > CORPUS_SIZE_LIMIT_FOR_FAILURES):
+    if (corpus_count > num_files_limit or corpus_size > size_limit):
```
oliverchang, line 136: nit: unnecessary parentheses now.

src/clusterfuzz/_internal/bot/tasks/utasks/corpus_pruning_task.py
oliverchang: +1 this is a bit confusing. can we either add a comment explaining this, or just change `last_execution_failed` to cover both cases ?

oliverchang (approved):

Issue #3950, opened 2024-04-17
Fix Android ability to launch Cuttlefish Images


marktefftech:
During the Android bot start up script, we are constantly polling the Android Build API [here](https://github.com/google/clusterfuzz/blob/d3854a4d00417cdae053b0b36418c5754a67efce/src/clusterfuzz/_internal/platforms/android/fetch_artifact.py#L161)

Unfortunately this API usage is flaky and causes a lot of outages.

Temporarily we are going to hardcode the build to a stable one, and use this ticket to track the automation of checking build stability before pulling the image and running the launch command

PR #4652, opened 2025-01-30
Use stable cuttlefish build if available.

aditya-wazir:
Ensures that the stable Cuttlefish build id is used if a valid build ID (bid) is present in stable_build_info.json, else uses tip-of-tree build id for cuttlefish devices.


src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
```diff
@@ -210,7 +210,9 @@ def get_latest_artifact_info(branch, target, signed=False):
   # TODO(https://github.com/google/clusterfuzz/issues/3950)
   # After stabilizing the Cuttlefish image, revert this
   if environment.is_android_cuttlefish():
-    return get_stable_build_info()
+    build_info = get_stable_build_info()
+    if 'bid' in build_info and build_info['bid'] != '0':
```
marktefftech, line 214: What happens in the case that this condition fails?

It seems like we would continue on the make the request, which is not what we want.

I'm not sure why the stable build id would be `== 0`. Can you handle the failure condition more explicitly, please?

aditya-wazir: Hi Mark,
If the above if-condition fails, then control flow will go ahead and make request to ab API to boot tip-of-tree build id.
We added 'build_info['bid'] != '0'' check so that we can toggle between golden cuttlefish build and tip-of-tree build by setting bid as 0 in stable_build_info.json, this way way we can easily swap between tip of tree and golden image without deployment.

marktefftech: Thanks for clarifying. Can you please add a comment indicating that a `bid==0` is purposefully used for fetching a build from tip of tree?

marktefftech:
@aditya-wazir once this is closed, do you think we should go ahead and update this issue?

https://github.com/google/clusterfuzz/issues/3950

marktefftech (approved):

src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
aditya-wazir: Hi Mark,
I have added comment as requested.

jonathanmetzman (approved):
LGTM

PR #4656, opened 2025-01-31
Improve logs of the grouper cron job

letitz:
Always log testcase IDs when grouping two testcases. Also log the reason why testcases are grouped.


letitz:
Ping, this is a small change that would greatly help us debug grouping issues.

letitz:
@vitorguidi maybe you have a bit more bandwidth for a quick review?

PR #4657, opened 2025-01-31
Increase fuzz tasks scheduled to deal with reduced capacity

jonathanmetzman:
The execution of tasks is more jagged under batch.
1. By having a buffer of tasks, we might be able to smooth this out by having fewer temporary declines in tasks.
2. The fuzz task scheduling takes very long (>20minutes) because of how long batch task querying takes. This can leave gaps that we are slow to react to.


vitorguidi (commented):
lgtm

Issue #4672, opened 2025-02-06
Prevent corpus explosion in tiny projects


jonathanmetzman:
Tiny projects have been run too much since the batch switch. This is leading to corpus explosion which prevent pruning from running. A vicious cycle.

PR #4673, opened 2025-02-06
Try to fix corpus explosion issue through scheduling: #4672

jonathanmetzman:
A bug in the scheduling code made smaller projects run way too often. This has caused corpus explosions.

Fix this, don't accidentally ignore the project weight.


jonathanmetzman:
PTAL.
TBR to stop the bleeding.

PR #4677, opened 2025-02-07
Propagate fuzz target in fuzz_task

jonathanmetzman:


jonathanmetzman:
TODO: is it fuzz_target.name? or something else I need to propagate.

PR #4678, opened 2025-02-07
Allow Android queue to proceed with the command  process_command_impl

svasudevprasad:
If the job base quque is '-android', which is the default Android queue ignore the mismatch since with subqueues, it is expected. Let the command proceed


jonathanmetzman (approved):
This subqueue stuff is making the code pretty ugly.

PR #4674, opened 2025-02-07
centipede: create workdir in `prepare` instead of the constructor

paulsemel:
It turns out that the same class can be used for different fuzzing rounds, and the parent directory of the temp dir is being cleared in between each round. For that reason, we need to re-create a workdir in `prepare`.

PR #4679, opened 2025-02-10
Merge #4674 to 'chrome' branch

paulsemel:
centipede: create workdir in prepare instead of the constructor

It turns out that the same class can be used for different fuzzing rounds, and the parent directory of the temp dir is being cleared in between each round. For that reason, we need to re-create a workdir in prepare.


letitz:
Mentioning #4674 in a comment to help Github xrefs.

letitz (approved):
LGTM

PR #4681, opened 2025-02-10
Bump dependencies to fix twine urllib mismatch

vitorguidi:
We are getting this error during butler deploy:

```
| ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
| twine 5.1.1 requires urllib3>=1.26.0, but you have urllib3 1.24.3 which is incompatible.
```

There might have been a regression in the underlying python dependencies. We ran:

```
pipenv lock
```

And this will change the following deps:

[prod]
aiohappyballs: 2.4.2 > 2.4.6
aiosignal: 1.3.1 > 1.3.2
attrs: 24.2 > 25.1
cachetools: 5.5.0 > 5.5.1
deprecated: 1.2.14 > 1.2.18
frozenlist: 1.4.1 > 1.5.0
google-cloud-appengine-logging: 1.4.5 > 1.5.0
google-apis-commons-protos: 1.65  > 1.66
grpc-google-iam-v1: 0.13.1 > 0.14.0
pbr: 6.1.0 > 6.1.1
propcache: 0.2.1 (NEW)
proto-plus: 1.24.0 > 1.26.0
pyjwt: 2.9.0 > 2.10.1
setuptools: 75.1.0 > 75.8.0
six: 1.16.0 > 1.17.0
yarl: 1.13.1 > 1.18.3

[dev]
cachecontrol: 0.14.0 > 0.14.1
cachetools: 5.5.0 > 5.5.1
click: 8.1.7 > 8.1.8
google-cloud-firestore: 2.19.0 > 2.20.0
google-apis-common-protos: 1.65.0 > 1.66.0
markupsafe: 2.1.5 > 3.0.2
proto-plus: 1.24.0 > 1.26.0
pyjwt: 2.9.0 > 2.10.1

### Testing

We successfully deployed from butler with --release=chrome-tests-syncer and --target=zips, this unblocks deployments

PR #4685, opened 2025-02-10
Revert "Bump dependencies to fix twine urllib mismatch (#4681)"

vitorguidi:
This reverts commit ff6b717c98e055bf1f77994733982ec413f00a80.

The issue occurs only with --targets=zips,k8s, and we are moving the k8s part to argocd. Also, this does not really fix the problem, so reverting to prevent any surprises.


jonathanmetzman (approved):
lgtm

PR #4688, opened 2025-02-11
Fix detection of generic segfaults in V8

backes:
On non-sanitizer builds we just get the output `Received signal 11 <unknown> 000000000000` (see https://crbug.com/388616198).
This adjusts the regular expression for generic segfaults to detect this.


jonathanmetzman (approved):
lgtm

backes:
Thanks, Jonathan!

PR #4694, opened 2025-02-13
Ignore ABRT crashes due to Centipede timeouts.

letitz:
When Centipede detects a timeout, it sends `SIGABRT` to the runner process. If running an ASAN build with `handle_abort=2`, this causes the runner to crash with an ASAN report. ClusterFuzz should not try to parse that and treat it as an abort, it really is just a timeout.

Bug: https://crbug.com/396148611


src/clusterfuzz/stacktraces/constants.py
```diff
@@ -649,6 +649,7 @@
     'Security CHECK failure',
     'Security DCHECK failure',
     'Out-of-memory',
+    'Timeout',
```
jonathanmetzman, line 652: I'm a little nervous about this one, but I looked it up and it seems fine.

jonathanmetzman (approved):
lgtm

PR #4696, opened 2025-02-18
Resolve TypeError during FuzzContext creation

AashutoshMurthy:
This PR aims to resolve the following error:
`TypeError: bad argument type for built-in operation`

The value of `fuzzer_metadata` for the key `'issue_metadata'` is a `dict`, it should be a `str`.


jonathanmetzman:
I think this needs more thought. We are returning a dict elsewhere, do all users of this need a string.

AashutoshMurthy:
> I think this needs more thought. We are returning a dict elsewhere, do all users of this need a string? Why bother loading it.

I had a look at the usage of `issue_metadata`, and have summarised my findings in the chart below:

image

From this, I feel we can probably change the fix to convert `issue_metadata` from a `dict` to `str` in `fuzz_task.py`, as the same logic has been implemented elsewhere. Here's the [reference](https://github.com/google/clusterfuzz/commit/d6ce764a41d0d0208961fe3e88859637adc689f3) where `issue_metadata` is converted to a `str` using `json.dumps()` wherever needed.

jonathanmetzman:
> > I think this needs more thought. We are returning a dict elsewhere, do all users of this need a string? Why bother loading it.
> 
> I had a look at the usage of `issue_metadata`, and have summarised my findings in the chart below:
> 
> image
> 
> From this, I feel we can probably change the fix to convert `issue_metadata` from a `dict` to `str` in `fuzz_task.py`, as the same logic has been implemented elsewhere. Here's the [reference](https://github.com/google/clusterfuzz/commit/d6ce764a41d0d0208961fe3e88859637adc689f3) where `issue_metadata` is converted to a `str` using `json.dumps()` wherever needed.

Sure, let's do that.

jonathanmetzman (approved):
lgtm

PR #4698, opened 2025-02-19
Schedule fuzzing more aggressively

jonathanmetzman:
Try to avoid dips in fuzz tasks, especially when corpus pruning is scheduled.

### DIFF
--- a/src/clusterfuzz/_internal/cron/schedule_fuzz.py
+++ b/src/clusterfuzz/_internal/cron/schedule_fuzz.py
@@ -35,6 +35,10 @@ from clusterfuzz._internal.metrics import logs
 # TODO(metzman): Actually implement this.
 CPUS_PER_FUZZ_JOB = 2
 
+# Pretend like our CPU limit is 3% higher than it actually is so that we use the
+# full CPU capacity even when scheduling is slow.
+CPU_BUFFER_MULTIPLIER = 1.03
+
 
 def _get_quotas(creds, project, region):
   compute = discovery.build('compute', 'v1', credentials=creds)
@@ -72,7 +76,7 @@ def count_unacked(creds, project_id, subscription_id):
   return 0
 
 
-def get_available_cpus_for_region(creds, project: str, region: str) -> int:
+def get_cpu_limit_for_regions(creds, project: str, region: str) -> int:
   """Returns the number of available CPUs in the current GCE region."""
 
   quotas = _get_quotas(creds, project, region)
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -258,24 +262,26 @@ def get_available_cpus(project: str, regions: List[str]) -> int:
   # Add up all queued and scheduled.
   region_counts = [sum(tup) for tup in region_counts]
   logs.info(f'Region counts: {region_counts}')
-  if region_counts[0] > 10000:
+  if region_counts[0] > 50_000:
```
vitorguidi, line 265: Can you extract this to a constant so it is easier to understand this is a tunable, from reading the top of the file?

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -258,24 +262,26 @@ def get_available_cpus(project: str, regions: List[str]) -> int:
...
-            f'occupied): {available_cpus}')
-  available_cpus = max(available_cpus - soon_occupied_cpus, 0)
+            f'occupied): {cpu_limit}')
+  available_cpus = max(cpu_limit - soon_occupied_cpus, 0)
 
-  # Don't schedule more than 10K tasks at once. So we don't overload batch.
-  print('len_regions', len(regions))
+  # Don't schedule more than 50K tasks at once. So we don't overload batch.
   # This number is arbitrary, but we aren't at full capacity at lower numbers.
   available_cpus = min(available_cpus, 50_000 * len(regions))
```
vitorguidi, line 283: Same constant as previous comment

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -258,24 +262,26 @@ def get_available_cpus(project: str, regions: List[str]) -> int:
...
   logs.info('Actually free CPUs (before subtracting soon '
-            f'occupied): {available_cpus}')
-  available_cpus = max(available_cpus - soon_occupied_cpus, 0)
+            f'occupied): {cpu_limit}')
+  available_cpus = max(cpu_limit - soon_occupied_cpus, 0)
 
-  # Don't schedule more than 10K tasks at once. So we don't overload batch.
-  print('len_regions', len(regions))
+  # Don't schedule more than 50K tasks at once. So we don't overload batch.
   # This number is arbitrary, but we aren't at full capacity at lower numbers.
```
vitorguidi, line 282: Is there any way we can reach out to batch to tune this algorithm, or request them to come with some sort of API that does what you are intending?

Also, some metrics on their side that would allow us to introspect into the task queue state, so this is less magical?

jonathanmetzman:
Test failure is old. Will fix later.

src/clusterfuzz/_internal/cron/schedule_fuzz.py
jonathanmetzman: I'll ask them!

jonathanmetzman: I'm having trouble imagining what the API would look like though.

PR #4701, opened 2025-02-20
Add BUILD_ID environment variable support

aditya-wazir:
This commit adds support for the BUILD_ID environment variable. If set, it will be used to construct the build info.

### DIFF
--- a/src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
+++ b/src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
@@ -207,6 +207,12 @@ def get_latest_artifact_info(branch, target, signed=False):
   if not client:
     return None
 
+  build_id = environment.get_value('BUILD_ID')
+  if build_id is not None and build_id != '':
+    logs.info(f'BUILD_ID env var present with value {build_id}. '
+              f'Hence returning its value.')
+    return {"bid": build_id, "branch": branch, "target": target}
+
   # TODO(https://github.com/google/clusterfuzz/issues/3950)
   # After stabilizing the Cuttlefish image, revert this
   if environment.is_android_cuttlefish():


marktefftech:
@aditya-wazir is `BUILD_ID` used elsewhere, or is this a net-new environment variable being introduced?

Asking because I believe the ClusterFuzz team is wanting to reduce dependency on env variables over the longer-term, so if there's any other way to accomplish this without adding a new env var, we would want to consider that approach first.

Correct me if I understand incorrectly @jonathanmetzman

aditya-wazir:
> @aditya-wazir is `BUILD_ID` used elsewhere, or is this a net-new environment variable being introduced?
> 
> Asking because I believe the ClusterFuzz team is wanting to reduce dependency on env variables over the longer-term, so if there's any other way to accomplish this without adding a new env var, we would want to consider that approach first.
> 
> Correct me if I understand incorrectly @jonathanmetzman

Hi @marktefftech ,
BUILD_ID is a new environment variable introduced by us to send BUILD ID  directly through clusterfuzz run command. we are doing this so that the build can be configured via ATP tests. Without this, there is no mechanism to update the build id used for a test.

jonathanmetzman:
> > @aditya-wazir is `BUILD_ID` used elsewhere, or is this a net-new environment variable being introduced?
> > Asking because I believe the ClusterFuzz team is wanting to reduce dependency on env variables over the longer-term, so if there's any other way to accomplish this without adding a new env var, we would want to consider that approach first.
> > Correct me if I understand incorrectly @jonathanmetzman
> 
> Hi @marktefftech , BUILD_ID is a new environment variable introduced by us to send BUILD ID directly through clusterfuzz run command. we are doing this so that the build can be configured via ATP tests. Without this, there is no mechanism to update the build id used for a test.

Thanks mark for looking out for this.

Where will BUILD_ID be set, will it be in the job definition?

PR #4705, opened 2025-02-25
Ensure regression tasks don't start new tasks when not making progress

mi-ac:
This returns a bad-build error type after a regression task, when a known min/max input pair didn't shrink. Previously, a timeout error type was returned, after which post-process respawns the task with the same min/max pair, leading to a task loop.

This doesn't change behavior when either min or max are not known (e.g. when tasks are initially started) or when some progress is made (e.g. at least one improvement to min or max during several hours this task can run) or when other error conditions happen.

Chrome bug: https://crbug.com/396344382


jonathanmetzman (approved):
lgtm

PR #4706, opened 2025-02-26
[Centipede] Perform corpus minimization during a fuzzing session

alhijazi:
This change call's centipede's minimize_corpus method on each fuzzing round. 
This allows us to distill the corpus and only add useful units.
On a local run, this allowed to decrease the number of units added from 2000+ to 5 with the same coverage.
The cl also extracts some common functionality into engine_common.

### DIFF
--- a/src/clusterfuzz/_internal/bot/fuzzers/centipede/constants.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/centipede/constants.py
@@ -40,7 +40,7 @@ EXTRA_BINARIES_FLAGNAME = 'extra_binaries'
 EXIT_ON_CRASH_FLAGNAME = 'exit_on_crash'
 
 MAX_LEN_FLAGNAME = 'max_len'
-RUNS_FLAGNAME = 'runs'
+NUM_RUNS_FLAGNAME = 'num_runs'
 BATCH_SIZE_FLAGNAME = 'batch_size'
 
 NUM_RUNS_PER_MINIMIZATION = 100000
--- a/src/clusterfuzz/_internal/bot/fuzzers/centipede/constants.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/centipede/constants.py
@@ -39,6 +39,10 @@ BINARY_FLAGNAME = 'binary'
 EXTRA_BINARIES_FLAGNAME = 'extra_binaries'
 EXIT_ON_CRASH_FLAGNAME = 'exit_on_crash'
 
+MAX_LEN_FLAGNAME = 'max_len'
+RUNS_FLAGNAME = 'runs'
+BATCH_SIZE_FLAGNAME = 'batch_size'
+
 NUM_RUNS_PER_MINIMIZATION = 100000
 
 
--- a/src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
@@ -32,6 +32,7 @@ from clusterfuzz._internal.bot.fuzzers.centipede import constants
 from clusterfuzz._internal.metrics import logs
 from clusterfuzz._internal.system import environment
 from clusterfuzz._internal.system import new_process
+from clusterfuzz._internal.system import shell
 from clusterfuzz.fuzz import engine
 from clusterfuzz.stacktraces import constants as stacktraces_constants
 
@@ -44,6 +45,17 @@ TargetBinaries = namedtuple('TargetBinaries', ['unsanitized', 'sanitized'])
 
 class CentipedeError(Exception):
   """Base exception class."""
+
+
+class CentipedeOptions(engine.FuzzOptions):
+  """Centipede engine options."""
+
+  def __init__(self, corpus_dir, arguments, strategies, workdir,
+               new_corpus_dir):
+    super().__init__(corpus_dir, arguments, strategies)
+    # Directory to add new units
+    self.new_corpus_dir = new_corpus_dir
+    self.workdir = workdir
 
 
 def _get_runner(target_path):
@@ -198,11 +210,15 @@ class Engine(engine.Engine):
     # 1. Centipede-readable corpus file;
     # 2. Centipede-readable feature file;
     # 3. Crash reproducing inputs.
-    workdir = self._create_temp_dir('workdir')
+    workdir = engine_common.create_temp_fuzzing_dir('workdir')
     arguments[constants.WORKDIR_FLAGNAME] = str(workdir)
 
-    # Directory corpus_dir saves the corpus files required by ClusterFuzz.
-    arguments[constants.CORPUS_DIR_FLAGNAME] = corpus_dir
+    # Directory to place new units. While fuzzing, the new corpus
+    # elements are written to the first dir in the list of corpus directories.
+    new_corpus_dir = engine_common.create_temp_fuzzing_dir('new')
+    corpus_dirs = [new_corpus_dir, corpus_dir]
+    arguments[constants.CORPUS_DIR_FLAGNAME] = ','.join(
+        dir for dir in corpus_dirs)
 
     target_binaries = self._get_binary_paths(target_path)
     if target_binaries.unsanitized is None:
@@ -214,7 +230,8 @@ class Engine(engine.Engine):
       arguments[constants.EXTRA_BINARIES_FLAGNAME] = str(
           target_binaries.sanitized)
 
-    return engine.FuzzOptions(corpus_dir, arguments.list(), {})
+    return CentipedeOptions(corpus_dir, arguments.list(), {}, workdir,
+                            new_corpus_dir)
 
   def _get_binary_paths(self, target_path):
     """Gets the paths to the main and auxiliary binaries based on |target_path|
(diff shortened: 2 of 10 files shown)


alhijazi:
> So this is really intendended to be like libFuzzer's "corpus merging" and not replace corpus pruning task right?

Indeed, this is needed to make sure we only upload useful units from a single fuzzing session.
The corpus pruning task will deduplicate inputs uploaded by multiple fuzzing sessions.

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
```diff
@@ -461,11 +544,16 @@ def minimize_corpus(self, target_path, arguments, input_dirs, output_dir,
...
+      for file in os.listdir(crashes_dir):
+        crasher_path = os.path.join(crashes_dir, file)
+        shutil.copy(crasher_path, reproducers_dir)
+
     shutil.rmtree(minimized_corpus_workdir)
+    if not workdir:
+      # Only remove this directory if it was created in this method.
+      shutil.rmtree(full_corpus_workdir)
 
     return engine.ReproduceResult(result.command, result.return_code,
```
javanlacerda, line 558: Should it return a ReproduceResult? The libfuzzer engine returns a FuzzResult for minimize_corpus function. https://github.com/google/clusterfuzz/blob/0cd0f6e06e1606cc01e908f18078db957b3ba121/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py#L552. I think we should standarize this return to make it easier to be handled by the corpus pruning task.

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
```diff
@@ -198,11 +210,15 @@ def prepare(self, corpus_dir, target_path, build_dir):
...
     arguments[constants.WORKDIR_FLAGNAME] = str(workdir)
 
-    # Directory corpus_dir saves the corpus files required by ClusterFuzz.
-    arguments[constants.CORPUS_DIR_FLAGNAME] = corpus_dir
+    # Directory to place new units. While fuzzing, the new corpus
+    # elements are written to the first dir in the list of corpus directories.
+    new_corpus_dir = engine_common.create_temp_fuzzing_dir('new')
+    corpus_dirs = [str(new_corpus_dir), str(corpus_dir)]
+    arguments[constants.CORPUS_DIR_FLAGNAME] = ','.join(
+        dir for dir in corpus_dirs)
```
jonathanmetzman, line 221: I don't think the for part is necessary, just `corpus_dirs`

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
```diff
@@ -284,11 +301,44 @@ def fuzz(self, target_path, options, reproducers_dir, max_time):  # pylint: disa
...
+    except Exception as e:
+      corpus_minimization_failed = True
+      logs.error(f'corpus minimization failed: {e}')
+
+    if corpus_minimization_failed:
+      # If we fail to minimize, fall back to moving the new units
+      # from the new corpus_dir to the main corpus_dir.
+      engine_common.move_mergeable_units(options.new_corpus_dir,
+                                         options.corpus_dir)
+
```
jonathanmetzman, line 341: Can you leave a TODO (and maybe a self-reminder) to change this to a warning if this becomes a problem caused by user code rather than by ClusterFuzz or centipede?

jonathanmetzman: Also, I think there's no need for the extra variable, just do the move in the except block.

src/clusterfuzz/_internal/bot/fuzzers/engine_common.py
```diff
@@ -656,3 +657,34 @@ def get_log_header(command, time_executed):
...
   quoted_command = get_command_quoted(command)
   return f'Command: {quoted_command}\nTime ran: {time_executed}\n'
+
+
+def is_sha1_hash(possible_hash):
+  """Returns True if |possible_hash| looks like a valid sha1 hash."""
+  if len(possible_hash) != 40:
+    return False
+
+  hexdigits_set = set(string.hexdigits)
```
jonathanmetzman, line 667: Can you make this a global constant? I don't like recreating this for every testcase.
Even though it isn't your code.

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
```diff
@@ -284,11 +301,44 @@ def fuzz(self, target_path, options, reproducers_dir, max_time):  # pylint: disa
...
+          input_dirs=[options.new_corpus_dir],
+          output_dir=options.corpus_dir,
+          reproducers_dir=reproducers_dir,
+          max_time=time_for_minimize,
+          # Use the same workdir that was used for fuzzing.
+          # This allows us to skip rerunning the fuzzing inputs.
+          workdir=workdir)
+    except Exception as e:
+      corpus_minimization_failed = True
+      logs.error(f'corpus minimization failed: {e}')
```
jonathanmetzman, line 334: We don't need to log the exception type, error does magic to find it for us.

jonathanmetzman (approved):
LGTM with nits

src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py
```diff
@@ -1352,7 +1319,8 @@ def pick_strategies(strategy_pool, fuzzer_path, corpus_directory,
 
   # Generate new testcase mutations using radamsa, etc.
   if is_mutations_run:
-    new_testcase_mutations_directory = create_corpus_directory('mutations')
+    new_testcase_mutations_directory = engine_common.create_temp_fuzzing_dir(
```
vitorguidi, line 1322: What is the difference between a corpus directory and a temporary fuzzing dir? Why cant this function in commons be called `create_temp_dir`?

vitorguidi:
> This change call's centipede's minimize_corpus method on each fuzzing round. This allows us to distill the corpus and only add useful units. On a local run, this allowed to decrease the number of units added from 2000+ to 5 with the same coverage. The cl also extracts some common functionality into engine_common.

About deploying this change:

* It touches libfuzzer and hongfuzz. These are minor changes only moving helper functions to engine_commons, so it is unlikely to break. However, how can we be sure that libfuzzer sessions will not change behavior? What should we be watching once this gets deployed?

* Same question goes for centipede. How can we know that the fuzzing does not break, and that the uploaded corpus is indeed smaller?

LGTM, if there is a plan to answer those questions once this lands

vitorguidi (approved):
lgtm, conditioned on a strategy for observing the rollout being presented

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
alhijazi: Totally agree.
I didn't modify it in my PR since I don't use the function return value and wanted to leave the decision to the implementer of the corpus_pruning task to modify as needed.

src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py
alhijazi: The initial `create_corpus_directory` created a subdirectory in the fuzzing temp directory.
```
def create_corpus_directory(name):
  """Create a corpus directory with a give name in temp directory and return its
  full path."""
  new_corpus_directory = os.path.join(fuzzer_utils.get_temp_dir(), name)
  engine_common.recreate_directory(new_corpus_directory)
  return new_corpus_directory
```
For this reason, I extracted and renamed this method to `create_temp_fuzzing_dir` to indicate that we are creating a temporary directory for fuzzing

src/clusterfuzz/_internal/bot/fuzzers/centipede/constants.py
```diff
@@ -39,6 +39,10 @@
 EXTRA_BINARIES_FLAGNAME = 'extra_binaries'
 EXIT_ON_CRASH_FLAGNAME = 'exit_on_crash'
 
+MAX_LEN_FLAGNAME = 'max_len'
+RUNS_FLAGNAME = 'runs'
+BATCH_SIZE_FLAGNAME = 'batch_size'
```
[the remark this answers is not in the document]
alhijazi: We can probably plan something to change it in a follow up. Open to suggestions

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
alhijazi: I ended up modifying this to:
```
    arguments[constants.CORPUS_DIR_FLAGNAME] = f'{new_corpus_dir},{corpus_dir}'
```

Using corpus_dir directly leads to having the array brackets in the arguments
```
'[path/to/dir1, path/to/dir2]'
```

alhijazi:
> > This change call's centipede's minimize_corpus method on each fuzzing round. This allows us to distill the corpus and only add useful units. On a local run, this allowed to decrease the number of units added from 2000+ to 5 with the same coverage. The cl also extracts some common functionality into engine_common.
> 
> About deploying this change:
> 
> * It touches libfuzzer and hongfuzz. These are minor changes only moving helper functions to engine_commons, so it is unlikely to break. However, how can we be sure that libfuzzer sessions will not change behavior? What should we be watching once this gets deployed?
> * Same question goes for centipede. How can we know that the fuzzing does not break, and that the uploaded corpus is indeed smaller?
> 
> LGTM, if there is a plan to answer those questions once this lands

Af for breaking libfuzzer/hongfuzz, I took a close look at the changes and I am confident no behavioural change was done. As you can see, all I did was extract some methods to a common directory while ensuring we are doing exactly the same thing as before.

As for centipede, I plan to closely monitor this and make sure we are seeing the expected behaviour and that corpus_minimization is succeeding.
The desired effect for centipede will become clear once the corpus_pruning task starts running and we can start with a fresh corpus for our big fuzzers.

PR #4708, opened 2025-02-27
Update appengine imports for external_testcase_reader. 

pgrace-google:
(drive by add additional component filter in query for extra safety)

### DIFF
--- a/src/clusterfuzz/_internal/cron/external_testcase_reader.py
+++ b/src/clusterfuzz/_internal/cron/external_testcase_reader.py
@@ -19,18 +19,27 @@ import re
 from google.cloud import storage
 import requests
 
-from appengine.libs import form
-from appengine.libs import gcs
-from appengine.libs import helpers
 from clusterfuzz._internal.config import local_config
 from clusterfuzz._internal.issue_management.google_issue_tracker import \
     issue_tracker
+from clusterfuzz._internal.metrics import logs
 
 ACCEPTED_FILETYPES = [
     'text/javascript', 'application/pdf', 'text/html', 'application/zip'
 ]
 ISSUETRACKER_ACCEPTED_STATE = 'ACCEPTED'
 ISSUETRACKER_WONTFIX_STATE = 'NOT_REPRODUCIBLE'
+_CLUSTERFUZZ_GET_URL = (
+    'https://clusterfuzz.corp.google.com/upload-testcase/get-url-oauth')
+_UPLOAD_URL_PROPERTY = 'uploadUrl'
+_TESTCASE_ID_PROPERTY = 'id'
+
+
+class ExternalTestcaseReaderException(Exception):
+  """Error when uploading an externally submitted testcase.."""
+
+  def __init__(self, message):
+    super().__init__(message)
 
 
 def get_vrp_uploaders(config):
@@ -42,8 +51,7 @@ def get_vrp_uploaders(config):
   return members
 
 
-def close_issue_if_invalid(upload_request, attachment_info, description,
-                           vrp_uploaders):
+def close_issue_if_invalid(issue, attachment_info, description, vrp_uploaders):
   """Closes any invalid upload requests with a helpful message."""
   comment_message = (
       'Hello, this issue is automatically closed. Please file a new bug after'
@@ -51,10 +59,10 @@ def close_issue_if_invalid(upload_request, attachment_info, description,
   invalid = False
 
   # TODO(pgrace) Remove after testing.
-  if upload_request.id == 373893311:
+  if issue.id == 373893311:
     return False
 
-  if not upload_request.reporter in vrp_uploaders:
+  if not issue.reporter in vrp_uploaders:
     comment_message += (
         'You are not authorized to submit testcases to Clusterfuzz.'
         ' If you believe you should be, please reach out to'
@@ -90,8 +98,8 @@ def close_issue_if_invalid(upload_request, attachment_info, description,
     comment_message += (
         '\nPlease see the new bug template for more information on how to use'
         ' Clusterfuzz direct uploads.')
-    upload_request.status = ISSUETRACKER_WONTFIX_STATE
-    upload_request.save(new_comment=comment_message, notify=True)
+    issue.status = ISSUETRACKER_WONTFIX_STATE
+    issue.save(new_comment=comment_message, notify=True)
 
   return invalid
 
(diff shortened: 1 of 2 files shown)


pgrace-google:
(context failure logs: http://shortn/_BBefaIUwXU)

src/clusterfuzz/_internal/cron/external_testcase_reader.py
```diff
@@ -158,18 +158,21 @@ def handle_testcases(tracker, config):
...
+      query_filters=[
+          'componentid:1600865', 'componentid:1457062+', 'status:accepted'
+      ],
       only_open=True)
   for issue in older_issues:
     # Close out older bugs that may have failed to reproduce.
     if close_issue_if_not_reproducible(issue, config):
-      helpers.log('Closing issue {issue_id} as it failed to reproduce',
-                  issue.id)
+      logs.info('Closing issue %s as it failed to reproduce' % issue.id)
```
alhijazi, line 168: Prefer f-strings
```
logs.info(f'Closing issue {issue.id} as it failed to reproduce')
```

src/clusterfuzz/_internal/cron/external_testcase_reader.py
```diff
@@ -202,7 +205,7 @@ def handle_testcases(tracker, config):
     issue.status = ISSUETRACKER_ACCEPTED_STATE
     issue.assignee = 'clusterfuzz@chromium.org'
     issue.save(new_comment=comment_message, notify=True)
-    helpers.log('Submitted testcase file for issue {issue_id}', issue.id)
+    logs.info('Submitted testcase file for issue %s' % issue.id)
```
alhijazi, line 208: Same here:
```
logs.info(f'Submitted testcase file for issue {issue.id}')
```

src/clusterfuzz/_internal/cron/external_testcase_reader.py
```diff
@@ -189,7 +192,7 @@ def handle_testcases(tracker, config):
     reporters_map[issue.reporter] = reporters_map.get(issue.reporter, 1) + 1
     if close_issue_if_invalid(issue, attachment_metadata, commandline_flags,
                               vrp_uploaders):
-      helpers.log('Closing issue {issue_id} as it is invalid', issue.id)
+      logs.info('Closing issue %s as it is invalid' % issue.id)
```
alhijazi, line 195: Same comment for f-strings

src/clusterfuzz/_internal/cron/external_testcase_reader.py
```diff
@@ -158,18 +158,21 @@ def handle_testcases(tracker, config):
   # Handle bugs that were already submitted and still open.
   older_issues = tracker.find_issues_with_filters(
       keywords=[],
-      query_filters=['componentid:1600865', 'status:accepted'],
+      query_filters=[
+          'componentid:1600865', 'componentid:1457062+', 'status:accepted'
```
alhijazi, line 162: It might be worth extracting these to constants and adding a comment above each saying which component name they represent

src/clusterfuzz/_internal/cron/external_testcase_reader.py
```diff
@@ -202,7 +205,7 @@ def handle_testcases(tracker, config):
     issue.status = ISSUETRACKER_ACCEPTED_STATE
```
alhijazi, line 205: Unrelated to this change but worth looking into.
`submit_testcase` raises a TypeError  that is not caught anywhere. This will crash and block submitting other testcases.

I would suggest that instead of raising an exception, you log the error and return whether submit succeeded or not, then handle this here.

src/clusterfuzz/_internal/cron/external_testcase_reader.py
pgrace-google: love it!

src/clusterfuzz/_internal/cron/external_testcase_reader.py
pgrace-google: love the suggestion!! thank you!

pgrace-google:
updated!

i do not actually know if the problem of the imports is to be fixed by adding that to the path - but does that seem like a sensible solution to you??

alhijazi:
> updated!
> 
> i do not actually know if the problem of the imports is to be fixed by adding that to the path - but does that seem like a sensible solution to you??

So the issue is coming from these two imports:
```
from appengine.libs import form
from appengine.libs import gcs
```
Looking at the other cron jobs in the same directory, I noticed non of them use appengine imports.

Also, I noticed that for cron jobs like cleanup (https://github.com/google/clusterfuzz/blob/8e43b8b4bc7b984defe62654a2ec05365a7fdb7f/src/clusterfuzz/_internal/cron/cleanup.py#L4), there's an appengine cron handler called cleanup located here: https://github.com/google/clusterfuzz/blob/8e43b8b4bc7b984defe62654a2ec05365a7fdb7f/src/appengine/handlers/cron/cleanup.py#L4
Same pattern exists for  `predator_pull` and `triage` in https://github.com/google/clusterfuzz/tree/8e43b8b4bc7b984defe62654a2ec05365a7fdb7f/src/appengine/handlers/cron. 

The right approach might be similar to what was done for these three cron jobs.

Adding @jonathanmetzman as the code owner.

src/clusterfuzz/_internal/cron/external_testcase_reader.py
```diff
@@ -21,16 +21,18 @@
 
 from appengine.libs import form
 from appengine.libs import gcs
```
vitorguidi, line 23: It seems like you have to import this because the upload handler requires the CSRF token

https://github.com/google/clusterfuzz/blob/8e07445597abc24f3db26d622d8054adf9157873/src/appengine/handlers/upload_testcase.py#L636

Why not creating a different endpoint for external testcases uploads, and have it do this logic on your behalf? Arrange the code in such a way that the cron itself does not depend on anything from appengine. 


I am not that familiar with this codebase, so I do not know how viable this is. @oliverchang might be able to spot a more concrete way to pull this off

vitorguidi: The same applies for GCS. Is there a way to delegate this prepare upload call to a handler in the appengine side, and only have the cron request it, without having to depend on this import?

pgrace-google: I'm not sure how that might work - the prepare blob upload maybe I could set up as an appengine handler/endpoint, but the csrf token seems too sensitive to make an endpoint? and it appears that the other endpoints rely on this for authentication

Added Oliver as a review to get thoughts and ideas! 

Worst case idea: i can duplicate the two functions I need.

oliverchang: +1 to Vitor's comment. We can approve this PR to unblock the current breakage, but we should definitely look at splitting this out so that we don't need to depend on appengine here. this is very brittle and will likely break again in the future. 

it's definitely entirely possible to upload testcases to ClusterFuzz without depending on anything in ClusterFuzz (we have external clients that do this already).

We can go through the oauth version of the upload endpoint instead (https://github.com/google/clusterfuzz/blob/8e07445597abc24f3db26d622d8054adf9157873/src/appengine/handlers/upload_testcase.py#L649), which does not require a csrf token.

Similarly for the prepare_blob_upload part, there is already an existing endpoint: 
https://github.com/google/clusterfuzz/blob/e44a0a40e5e9945324a3d8e0c095fb1fa0ecc36c/src/appengine/handlers/upload_testcase.py#L272

pgrace-google: updated to rely on the oauth version!! i didnt know we had those :D thank you!!

PR #4707, opened 2025-02-27
[Centipede] Implement corpus prune for centipede on corpus prune task

javanlacerda:
It implements corpus pruning for centipede fuzzer.

PR #4710, opened 2025-03-01
bump yapf to 0.43.0

javanlacerda:
UNBLOCKS #4707 

It bumps the yapf library to 0.43.0
It's needed to support new python features, such as `match case`.

I've run the following command to validate it's working properly: 

`yapf -p -i src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py src/clusterfuzz/_internal/bot/tasks/utasks/corpus_pruning_task.py src/clusterfuzz/_internal/tests/core/bot/fuzzers/centipede/centipede_engine_test.py src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/corpus_pruning_task_test.py`

The given files were formatted correctly following the new style `google` instead of the deprecated one `chromium`


.style.yapf
```diff
@@ -1,5 +1,5 @@
 [style]
-based_on_style = chromium
+based_on_style = google
```
jonathanmetzman, line 2: Let's leave this as is for now. I don't want to pollute blame history with formatting changes.

javanlacerda: Hmm, but this is blocking this PR https://github.com/google/clusterfuzz/pull/4707. The current version doesn't support the `match case` usage. However I can modify the PR to use `if else` instead, and avoid this bump. WDYT?

jonathanmetzman: Chromium doesn't support `match case`? Bizzare.
I guess land this then.

.style.yapf
```diff
@@ -1,5 +1,5 @@
 [style]
-based_on_style = google
+based_on_style = pep8
```
jonathanmetzman, line 2: Hmm...Why switch to this? I think this is a much bigger change, can we just switch to google?

javanlacerda: But it seems that the behavior is different

javanlacerda: I reseted to the bump for 0.43.0 version.

.style.yapf
javanlacerda: It doesn't in the version 0.22 ;/. The version 0.29.1 supported but the lint behavior is different as well.

jonathanmetzman (changes requested):
Javan, on further thought I think it would be a bad idea to land this right now. Just get rid of the match case for now.
Landing this change will probably make it extremely difficult to merge the oss-fuzz branch back into the master branch.

PR #4711, opened 2025-03-03
Add the option to force the detection of OOMS and hangs, ignoring the redzone

ParisMeuleman:
the redzone prevents the detection of centipede timeouts, resulting in a lot of wrongly classified testcases https://crbug.com/396148611.

I wasn't able to find the reasons of the redzone limitation, so I went for a third value forcing.  Alternatively we might want to split ooms and hangs if only one is limited due to redzones?


ParisMeuleman:
Hey there, any updates ? :)

jonathanmetzman (approved):
lgtm

PR #4712, opened 2025-03-04
Modernize and add logs to corpus_pruning

alhijazi:


src/clusterfuzz/_internal/bot/tasks/setup.py
```diff
@@ -487,6 +487,7 @@ def update_data_bundle(
 
   time_before_sync_start = time.time()
 
+  result = None
```
jonathanmetzman, line 490: nit: I don't think this is helpful.

jonathanmetzman (approved):
LGTM with nits.

PR #4668, opened 2025-02-06
MNT Removing the Fluentd setup and updating to default to GCP logging

vitaliset:
Removes the configuration and setup of Fluentd across ClusterFuzz.

- Deleting Fluentd setup from multiple Docker and VM setup files.
- Modifying the logging configuration in `src/clusterfuzz/_internal/metrics/logs.py` to remove references to Fluentd and default the logging to google cloud logging.
- Updating related tests in `src/clusterfuzz/_internal/tests/core/metrics/logs_test.py` to reflect these changes.

b/389622962

PR #4715, opened 2025-03-06
FIX Revert "MNT Removing the Fluentd setup and updating to default to GCP logging (#4668)"

vitaliset:
This reverts commit 8e43b8b4bc7b984defe62654a2ec05365a7fdb7f.

### DIFF
--- a/configs/test/gce/android-init.bash
+++ b/configs/test/gce/android-init.bash
@@ -80,6 +80,22 @@ echo "Adding workaround to prevent /dev/random hangs."
 rm /dev/random
 ln -s /dev/urandom /dev/random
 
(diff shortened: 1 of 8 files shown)


jonathanmetzman:
> This reverts commit [8e43b8b](https://github.com/google/clusterfuzz/commit/8e43b8b4bc7b984defe62654a2ec05365a7fdb7f).

Can you give more context why this is needed?

vitaliset:
> Can you give more context why this is needed?

It's no longer necessary. During debugging with @vitorguidi, we considered reverting the change, but we later identified the root cause. Closing this PR.

PR #4716, opened 2025-03-06
Adding clusterfuzz-config commit sha to revision

javanlacerda:
Adding clusterfuzz-config commit hash to the revision. It will help during tracing sessions.

image


jonathanmetzman (approved):
lgtm

src/local/butler/common.py
```diff
@@ -93,9 +94,13 @@ def _compute_revision(timestamp, *extras):
     crbug.com/674173."""
   timestamp = timestamp.strftime('%Y%m%d%H%M%S-utc')
   _, git_sha = execute('git rev-parse --short HEAD')
-  git_sha = git_sha.strip().decode('utf-8')
+  clusterfuzz_git_sha = git_sha.strip().decode('utf-8')
 
-  components = [timestamp, git_sha, os.environ['USER']] + list(extras)
+  # Adding also the clusterfuzz-config version to the revision
+  _, git_sha = execute(f'git -C {environment.get_config_directory()} rev-parse --short HEAD')
+  clusterfuzz_config_git_sha = git_sha.strip().decode('utf-8')
```
vitorguidi, line 101: Please provide auxiliary functions that get only the clusterfuzz git sha, or the clusterfuzz config git sha. This will be hard to reuse for structured logging ,where both pieces of information should be different labels

src/local/butler/common.py
```diff
@@ -93,9 +94,13 @@ def _compute_revision(timestamp, *extras):
...
+
+  components = [timestamp, clusterfuzz_git_sha, clusterfuzz_config_git_sha, os.environ['USER']] + list(extras)
```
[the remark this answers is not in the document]
vitorguidi: [Here](https://github.com/google/clusterfuzz/blob/956647f683df5fc3cf1e6ac244b825b09874bc95/src/clusterfuzz/_internal/bot/tasks/update_task.py#L168) for instance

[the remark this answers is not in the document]
vitorguidi: This in particular will not break because timestamp comes first, worth double checking where else this might touch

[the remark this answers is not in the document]
javanlacerda: it makes sense. I'll move the config sha to the end, with this we will keep the order and for cases that splits this for getting any information, the indexes will keep the same. And I'll double check its usage anyways

src/local/butler/common.py
```diff
@@ -86,16 +86,32 @@ def compute_prod_revision():
   return _compute_revision(_utcnow())
 
 
+def _get_clusterfuzz_commit_sha():
+  _, git_sha = execute('git rev-parse --short HEAD')
+  return git_sha.strip().decode('utf-8')
```
vitorguidi, line 91: This will work only for the context of butler deploy/package, not for a running bot. You have to parse this information out of the manifest file

This is where the[ manifest file is produced](https://github.com/google/clusterfuzz/blob/master/src/local/butler/package.py#L129
)

This is where the [information is consumed](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/base/utils.py#L925)

The idea is that these functions are also provided in utils, so the bot can get them in runtime as well. Also, please use [memoize](https://github.com/google/clusterfuzz/blob/1d2c8d1cd84ef15e311790d099d6aa0f08a16ffa/src/appengine/libs/auth.py#L70) for these auxiliary functions, so that we do not need to read a file everytime we write a log

src/local/butler/common.py
```diff
@@ -86,16 +86,32 @@ def compute_prod_revision():
...
+
+
+def _get_clusterfuzz_config_commit_sha():
```
vitorguidi, line 94: ditto

src/local/butler/common.py
javanlacerda: Hmm, but looking into the code, the `compute_prod_revision` is used on butler deploy only.  the staging one is used on `windows.py` and `posix.py`. Is this staging version used in production?

src/local/butler/common.py
```diff
@@ -86,16 +86,32 @@ def compute_prod_revision():
...
   timestamp = timestamp.strftime('%Y%m%d%H%M%S-utc')
-  _, git_sha = execute('git rev-parse --short HEAD')
-  git_sha = git_sha.strip().decode('utf-8')
+  clusterfuzz_git_sha = _get_clusterfuzz_commit_sha()
+
+  # Adding also the clusterfuzz-config version to the revision
+  clusterfuzz_config_git_sha = _get_clusterfuzz_config_commit_sha()
 
-  components = [timestamp, git_sha, os.environ['USER']] + list(extras)
+  components = [
```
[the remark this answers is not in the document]
javanlacerda: Sure! 
image

Pipfile
```diff
@@ -24,6 +24,7 @@ pyfakefs = "==4.4.0"
 pylint = "~=2.4"
 pylint-protobuf = "==0.21.0"
 pyright = "==1.1.393"
+pyyaml = "==6.0.2"
```
vitorguidi, line 27: Is this needed or can it be rolled back?

javanlacerda: I'll make a commit removing, just to be sure that it's needed.

javanlacerda: Actually, I double checked it and it's needed. Pyyaml é installed through the setup script, but only in the "bots" context. https://github.com/google/clusterfuzz/blob/1d2c8d1cd84ef15e311790d099d6aa0f08a16ffa/src/setup.py#L53. 

This change will use pyyaml on the butler deploy context, then we need it on the root level dependencies.

vitorguidi (approved):
lgtm

PR #4719, opened 2025-03-07
Set a proper timeout for corpus minimization after a fuzzing session

alhijazi:
This change also adds the collected new_units_added stat to the centipede schema.

### DIFF
--- a/src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
@@ -189,6 +189,19 @@ class Engine(engine.Engine):
     return arguments
 
   # pylint: disable=unused-argument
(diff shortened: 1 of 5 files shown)


src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
```diff
@@ -188,6 +188,19 @@ def _get_arguments(self, fuzzer_path):
 
     return arguments
 
+  # pylint: disable=unused-argument
+  def fuzz_additional_processing_timeout(self, options):
```
jonathanmetzman, line 192: nit: This should be next to the function. Or better yet, just do `del options` and skip the pylint directive.

jonathanmetzman (approved):
lgtm

PR #4720, opened 2025-03-10
Skip 'rsync_from_disk()' call if the quarantine corpus is empty

dhavalts1989:
We noticed an increase in fuzzers failing in the Coverage Client Job. We observed that the major cause was a failure during corpus pruning. Calling 'rsync_from_disk()' with an [empty quarantine corpus](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/tasks/utasks/corpus_pruning_task.py#L702) leads to trigger [unintended assertion](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/fuzzing/corpus_manager.py#L442) in the corpus pruning job, which subsequently causes the Coverage Client Job to fail. To address this, we now skip the '[rsync_from_disk()](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/tasks/utasks/corpus_pruning_task.py#L702)' call if the quarantine corpus is empty, preventing unnecessary failures.


jonathanmetzman (approved):
lgtm

PR #4723, opened 2025-03-13
Dont treat jazzerjs as a target

jonathanmetzman:


jonathanmetzman:
Will fix: https://pantheon.corp.google.com/errors/detail/CO6GsvLh2Ney8gE;time=P30D;locations=global?e=-13802955&mods=logs_tg_prod&project=clusterfuzz-external in a few days when it stops being saved as a fuzz target.

vitorguidi:
> Will fix: https://pantheon.corp.google.com/errors/detail/CO6GsvLh2Ney8gE;time=P30D;locations=global?e=-13802955&mods=logs_tg_prod&project=clusterfuzz-external in a few days when it stops being saved as a fuzz target.

Why do you want to stop saving it as a fuzz target? The PR seems to be doing the opposite of it, allowlisting jazzer.js as a fuzz target. I do not understand the intention behind the fix

jonathanmetzman:
> > Will fix: https://pantheon.corp.google.com/errors/detail/CO6GsvLh2Ney8gE;time=P30D;locations=global?e=-13802955&mods=logs_tg_prod&project=clusterfuzz-external in a few days when it stops being saved as a fuzz target.
> 
> Why do you want to stop saving it as a fuzz target? The PR seems to be doing the opposite of it, allowlisting jazzer.js as a fuzz target. I do not understand the intention behind the fix

The PR is incomplete and I will fix it.
It is not a fuzz target, it is a binary that runs fuzz targets, if we treat it as a fuzz target, then we aren't actually fuzzing anything.

jonathanmetzman:
> > > Will fix: https://pantheon.corp.google.com/errors/detail/CO6GsvLh2Ney8gE;time=P30D;locations=global?e=-13802955&mods=logs_tg_prod&project=clusterfuzz-external in a few days when it stops being saved as a fuzz target.
> > 
> > 
> > Why do you want to stop saving it as a fuzz target? The PR seems to be doing the opposite of it, allowlisting jazzer.js as a fuzz target. I do not understand the intention behind the fix
> 
> The PR is incomplete and I will fix it. It is not a fuzz target, it is a binary that runs fuzz targets, if we treat it as a fuzz target, then we aren't actually fuzzing anything.

That's not what I've done here. Instead, here I made sure it always gets unpacked with every target.

PR #4727, opened 2025-03-18
Avoid categorizing dchecks as security issues

paulsemel:
We already have an environment variable to correctly handle that, and the behavior is pretty much the same between CHECKs and DCHECKs.


alhijazi:
Could you add a test in `src/clusterfuzz/_internal/tests/core/crash_analysis/stack_parsing/stack_analyzer_test.py`
similar to `test_check_failure_chrome` ?
You just need to make a file like
`src/clusterfuzz/_internal/tests/core/crash_analysis/stack_parsing/stack_analyzer_data/check_failure_chrome.txt`
, fill it with a chrome dcheck failure stack, and make sure it does what is intended.

alhijazi (approved):

alhijazi:
FYI @jonathanmetzman , @vitorguidi

PR #4729, opened 2025-03-19
FIX Prevent None values in GCP log labels

vitaliset:
While locally debugging on my machine for b/403539825, I noticed `BOT_NAME` was not set. This prevented me from sending logs to GCP, but the error message wasn't obvious.

To prevent this issue for other developers working with logs in the future, I suggest we add a default string value (e.g., "null") when `BOT_NAME` is not available. This should not interfere with production.


oliverchang (approved):
thanks!

vitorguidi (approved):

PR #4730, opened 2025-03-19
centipede: add default symbolizer path option to make the run faster

paulsemel:
The Sundew team confirmed that this flag was not necessary if we're not willing to log the features visually (seeing the functions being hit). This is not the case on ClusterFuzz, so we can safely ignore that flag. Note that adding this flag doesn't break the centipede statistics.


alhijazi (approved):

jonathanmetzman (approved):
lgtm

PR #4734, opened 2025-03-20
Add script to run grouper locally

ViniciustCosta:
Adding a script that modifies the grouper code enabling it to be executed locally.
This is intended to facilitate experimenting and analyzing the grouper outcome without touching production. There are mainly two features in it:
* Load and store a snapshot of the testcases attributes available for grouping, in order to be able to compare different versions of the grouper against the same state of the testcases.
* Enable executing the grouper with distinct variations, while storing the results in a graph-like data structure that makes it easier to inspect groups and undestand why testcases were grouped.

Note: since the code was copied and edited from the original grouper code, its is better to use the first commit to review changes.


src/local/butler/scripts/grouper_experiment.py
```diff
@@ -0,0 +1,728 @@
+# Copyright 2025 Google LLC
```
oliverchang, line 1: drive by comment: is there a way we could just call/re-use the actual grouper code here instead of duplicating this?

ViniciustCosta: I think so, but as far as I see, it would require adding a lot of conditions in the grouper code that could make it harder to understand and maintain the original code. IMO since this is just for experiments, it is ok to make it a copy of the grouper code for now.

oliverchang: thanks for clarifying! can you add this detail as a comment to this file to explain the changes you made for ease of local debugging?

oliverchang (approved):
thanks!

PR #4735, opened 2025-03-21
builds: consider missing APP_PATH blackbox builds as bad builds

paulsemel:
In cases where we have an APP_NAME and no APP_PATH set, it means we failed at finding the APP_NAME in the build. In such cases, it means we'll never be able to use them. For that reason, just mark those builds as bad.


jonathanmetzman (approved):
LGTM

PR #4736, opened 2025-03-21
[Logging] Adding progression structured log

javanlacerda:
It creates the structured log for progression tasks, defines de behavior to extract the information from the test case, and add this context behavior for progression preprocess, main and postprocess.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
@@ -731,10 +731,10 @@ def utask_postprocess(output: uworker_msg_pb2.Output):  # pylint: disable=no-mem
       _ERROR_HANDLER.handle(output)
       return
 
-    # If there is a fine grained bisection service available, request it. Both
-    # regression and fixed ranges are requested once. Regression is also requested
-    # here as the bisection service may require details that are not yet available
-    # (e.g. issue ID) at the time regress_task completes.
+    # If there is a fine grained bisection service available, request it.
+    # Both regression and fixed ranges are requested once. Regression is also
+    # requested here as the bisection service may require details that
+    # are not yet available (e.g. issue ID) at the time regress_task completes.
     bisection.request_bisection(testcase)
 
     if task_output and task_output.crash_on_latest:
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
@@ -436,38 +436,38 @@ def _set_regression_testcase_upload_url(
 def utask_preprocess(testcase_id, job_type, uworker_env):
   """Runs preprocessing for progression task."""
   testcase = data_handler.get_testcase_by_id(testcase_id)
-  if not testcase:
-    return None
+  with logs.progression_log_context(testcase):
+    if not testcase:
+      return None
+    if testcase.fixed:
+      logs.error(f'Fixed range is already set as {testcase.fixed}, skip.')
+      return None
 
-  if testcase.fixed:
-    logs.error(f'Fixed range is already set as {testcase.fixed}, skip.')
-    return None
+    # Set a flag to indicate we are running progression task. This shows pending
+    # status on testcase report page and avoid conflicting testcase updates by
+    # triage cron.
+    testcase.set_metadata('progression_pending', True)
+    data_handler.update_testcase_comment(testcase, data_types.TaskState.STARTED)
+    blob_name, blob_upload_url = blobs.get_blob_signed_upload_url()
+    progression_input = uworker_msg_pb2.ProgressionTaskInput(  # pylint: disable=no-member
+        custom_binary=build_manager.is_custom_binary(),
+        bad_revisions=build_manager.get_job_bad_revisions(),
+        blob_name=blob_name,
+        stacktrace_upload_url=blob_upload_url)
+    # Setup testcase and its dependencies.
+    setup_input = setup.preprocess_setup_testcase(testcase, uworker_env)
 
-  # Set a flag to indicate we are running progression task. This shows pending
-  # status on testcase report page and avoid conflicting testcase updates by
-  # triage cron.
-  testcase.set_metadata('progression_pending', True)
-  data_handler.update_testcase_comment(testcase, data_types.TaskState.STARTED)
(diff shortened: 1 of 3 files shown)


src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -684,8 +684,9 @@ def utask_main(uworker_input):
   """Executes the untrusted part of progression_task."""
   testcase = uworker_io.entity_from_protobuf(uworker_input.testcase,
                                              data_types.Testcase)
-  uworker_io.check_handling_testcase_safe(testcase)
-  return find_fixed_range(uworker_input)
+  with logs.progression_log_context(testcase):
+    uworker_io.check_handling_testcase_safe(testcase)
+    return find_fixed_range(uworker_input)
```
vitorguidi, line 689: Why here and not utask_main()?

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
```diff
@@ -713,53 +714,54 @@ def utask_postprocess(output: uworker_msg_pb2.Output):  # pylint: disable=no-mem
   """Trusted: Cleans up after a uworker execute_task, writing anything needed to
   the db."""
   testcase = data_handler.get_testcase_by_id(output.uworker_input.testcase_id)
```
vitorguidi, line 716: You can move this logic to the decorator itself, and wrap the function with

@ProgressionLogContext

This logic for fetching the testcase id and calling a with will repeat itself on 10 files like this, I think it will be cleaner with this suggestion

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -28,6 +28,11 @@
 import traceback
 from typing import Any
 from typing import NamedTuple
+from typing import TYPE_CHECKING
+
+# This is needed to avoid circular import
```
vitorguidi, line 33: It is unclear to me how this circular import will play out. Can you be more specific here?

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -631,9 +651,28 @@ def get_extras(self) -> NamedTuple:
...
+        return GenericLogStruct()
+
+    elif self == LogContextType.TESTCASE:
+      try:
+        testcase: "Testcase | None" = log_contexts.meta.get('testcase')
+        return TestcaseLogStruct(
+            testcase_id=testcase.key.id(),  # type: ignore
+            # The get_fuzz_target functions seems that always
+            # returns a FuzzTarget.
+            fuzz_target=testcase.get_fuzz_target().key.id(),  # type: ignore
```
vitorguidi, line 664: What if the testcase does not have a fuzz target? Blackbox fuzzers are like this

Null.key.id() => null ptr exception

Would be good to have an unit test for this scenario

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -706,3 +745,22 @@ def task_stage_context(stage: Stage):
       yield
     finally:
       log_contexts.delete_metadata('stage')
+
+
+@contextlib.contextmanager
+def testcase_log_context(testcase: "Testcase"):
```
vitorguidi, line 751: The testcase fetch from the database belongs here IMO

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
javanlacerda: Because it needs the testcase as argument, and it's recovered just in the line 685.

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
javanlacerda: It could be done, but notice that the way that testcase is recovered is different for some cases, we have the risk to not get it properly for the context. Also, we can create problem of circular import as almost everything imports logs and we would need to import he logic there.

src/clusterfuzz/_internal/metrics/logs.py
javanlacerda: The data_types loads logs, and logs imports the data_types. It creates an infinte loop of importing trial. This TYPE_CHECKING avoids this, because make it import the Testcase only during the Type checking execution and not during the runtime.

src/clusterfuzz/_internal/metrics/logs.py
javanlacerda: You're right. I took a look here and the `get_fuzz_target()` can return None for some cases. I'll handle this.

javanlacerda: fixed

src/clusterfuzz/_internal/metrics/logs.py
javanlacerda: I agree that it could be cleaner, but we cannot introduce this logic here avoid circular import. this information needs to be imputed on the log context

vitaliset (approved):

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -645,9 +665,33 @@ def get_extras(self) -> NamedTuple:
...
+    elif self == LogContextType.TESTCASE:
+      try:
+        testcase: "Testcase | None" = log_contexts.meta.get('testcase')
+        if not testcase:
+          error(
+              'Testcase not found in log context metadata', ignore_context=True)
+          return GenericLogStruct()
+
+        fuzz_target = testcase.get_fuzz_target()
+        fuzz_target = fuzz_target.key.id() if fuzz_target else 'unknown'
```
vitorguidi, line 680: :shark:

src/clusterfuzz/_internal/metrics/logs.py
vitorguidi: https://mypy.readthedocs.io/en/stable/runtime_troubles.html

src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py
javanlacerda: btw, it makes sense. I'll move the context to utask_main instead

vitorguidi: it is in uworker main, the git review line skip was deceiving here, lgtm

vitorguidi (approved):

PR #4738, opened 2025-03-24
Fix regression in regression task.

oliverchang:
After https://github.com/google/clusterfuzz/pull/3934, we stopped taking into account MIN_REVISION.

MIN_REVISION is usually used by jobs to indicate the minimum revision to start bisecting (because all builds prior are known to be "bad").

All jobs with MIN_REVISION set are currently failing regression because this is not respected.


oliverchang:
@vitorguidi @javanlacerda can you please help with reviewing/merging/deploying this tomorrow? This should hopefully resolve one of our current alerts on uworker error rates.

alhijazi (approved):

javanlacerda (approved):
LGTM!

PR #4722, opened 2025-03-13
[Logging] Task structured log

javanlacerda:
It implements structured logs for task context. It adds the following labels for logs on task context:
- Task Id (the UUID one)
- Task name
- Stage (preprocess, main, postprocess)

PR #4739, opened 2025-03-24
MNT Introduces a flush interval for GCP logs

vitaliset:
[AFAICT](https://cloud.google.com/logging/docs/agent/logging/configuration#cloud-fluentd-config) Google cloud Fluentd logs used to have a flush mechanism to log to GCP only when `buffer_chunk_limit` got over 512kb or every `flush_interval` (that [was set to 60 seconds](https://github.com/google/clusterfuzz/blob/af69605ef6c36af9b459d4ba6a5ebb055d363a17/configs/test/gce/android-init.bash#L95)). 

[Default GCP transport handler](https://cloud.google.com/python/docs/reference/logging/latest/google.cloud.logging_v2.handlers.transports.background_thread.BackgroundThreadTransport#google_cloud_logging_v2_handlers_transports_background_thread_BackgroundThreadTransport) has a `max_interval` (analogous to `flush_interval` IIUC) of 0 seconds. This means that we don't wait at all to send GCP logs. This is making our log writing rate much greater than usual, causing our logging quota to explode (b/405350230 and b/405972925).

This PR changes the `max_interval` and `grace_period` to a higher number.

Thanks @javanlacerda and @ViniciustCosta for taking a look into this with me. :)

### DIFF
--- a/src/clusterfuzz/_internal/metrics/logs.py
+++ b/src/clusterfuzz/_internal/metrics/logs.py
@@ -370,7 +370,7 @@ def configure_cloud_logging():
 
   # Rate for the BackgroundThreadTransport to flush the logs.
   handler.transport.worker._max_latency = int(
-      os.getenv('LOGGING_CLOUD_MAX_LATENCY', '60'))  # disable=protected-access
+      os.getenv('LOGGING_CLOUD_MAX_LATENCY', '60'))  # pylint: disable=protected-access
 
   def cloud_label_filter(record):
     # Update the labels with additional information.
--- a/src/clusterfuzz/_internal/metrics/logs.py
+++ b/src/clusterfuzz/_internal/metrics/logs.py
@@ -353,6 +353,7 @@ def configure_cloud_logging():
   """ Configure Google cloud logging, for bots not running on appengine nor k8s.
   """
   import google.cloud.logging
+  from google.cloud.logging.handlers.transports import BackgroundThreadTransport
 
   # project will default to the service account's project (likely from
   #   GOOGLE_APPLICATION_CREDENTIALS).
@@ -366,7 +367,8 @@ def configure_cloud_logging():
       'compute.googleapis.com/resource_name': socket.getfqdn().lower(),
       'bot_name': os.getenv('BOT_NAME', 'null'),
   }
-  handler = client.get_default_handler(labels=labels)
+  handler = client.get_default_handler(labels=labels)#, transport=BackgroundThreadTransport)
+  handler.transport.worker._max_latency = 60
 
   def cloud_label_filter(record):
     # Update the labels with additional information.
--- a/src/clusterfuzz/_internal/metrics/logs.py
+++ b/src/clusterfuzz/_internal/metrics/logs.py
@@ -353,7 +353,6 @@ def configure_cloud_logging():
   """ Configure Google cloud logging, for bots not running on appengine nor k8s.
   """
   import google.cloud.logging
-  from google.cloud.logging.handlers.transports import BackgroundThreadTransport
 
   # project will default to the service account's project (likely from
   #   GOOGLE_APPLICATION_CREDENTIALS).
--- a/src/clusterfuzz/_internal/metrics/logs.py
+++ b/src/clusterfuzz/_internal/metrics/logs.py
@@ -367,7 +367,9 @@ def configure_cloud_logging():
       'compute.googleapis.com/resource_name': socket.getfqdn().lower(),
       'bot_name': os.getenv('BOT_NAME', 'null'),
   }
-  handler = client.get_default_handler(labels=labels)#, transport=BackgroundThreadTransport)
+  handler = client.get_default_handler(labels=labels)
+
+  # Rate for the BackgroundThreadTransport to flush the logs
   handler.transport.worker._max_latency = 60
 
   def cloud_label_filter(record):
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -368,6 +368,9 @@ def configure_cloud_logging():
   }
   handler = client.get_default_handler(labels=labels)
 
+  # Rate for the BackgroundThreadTransport to flush the logs
+  handler.transport.worker._max_latency = 60
```
vitaliset, line 372: We can change it to 60 seconds (just like it was when fluentd was being used), but maybe setting it to 5 might be better because of the `grace_period` of [`BackgroundThreadTransport`](https://cloud.google.com/python/docs/reference/logging/latest/google.cloud.logging_v2.handlers.transports.background_thread.BackgroundThreadTransport#google_cloud_logging_v2_handlers_transports_background_thread_BackgroundThreadTransport). What do you think @oliverchang?

javanlacerda (approved):
LGTM! Thanks!

src/clusterfuzz/_internal/metrics/logs.py
ViniciustCosta: Wouldn't it be enough to set the `grace_period` to the same value or greater than the `max_latency`? Or is there an issue I'm not seeing?

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -368,6 +368,9 @@ def configure_cloud_logging():
   }
   handler = client.get_default_handler(labels=labels)
 
+  # Rate for the BackgroundThreadTransport to flush the logs
```
ViniciustCosta, line 371: Maybe try to check if the `batch_size` from BackgroundThreadTransport is enough compared to the `buffer_chunk_limit` from fluentD.

ViniciustCosta (approved):
LGTM!

src/clusterfuzz/_internal/metrics/logs.py
vitaliset: I think that the translation might be harder, but might be worth it if this doesn't fix the issue. Any thoughts @oliverchang?

oliverchang: +1 let's see how this goes first. we can tweak this in subsequent PRs and compare with the fluentd behaviour if this doesn't fix it sufficiently.

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -366,7 +368,18 @@ def configure_cloud_logging():
...
       'bot_name': os.getenv('BOT_NAME', 'null'),
   }
-  handler = client.get_default_handler(labels=labels)
+
+  class MaxLatencyTransport(BackgroundThreadTransport):
+
+    def __init__(self, client, name, **kwargs):
+      super().__init__(
+          client,
+          name,
```
oliverchang, line 377: do we need to set `grace_period` here? otherwise we might lose logs forever right?

oliverchang (approved):
LGTM with one question.

src/clusterfuzz/_internal/metrics/logs.py
oliverchang: the logs leading up to a process exit can be quite crucial (e.g. logs leading up to an exception that can be helpful for debugging), so IMO it's very important we keep them. 

How about something like `grace_period` 15 and `max_latency` 10? 15 seconds seems OK to me for teardown, and 10 seconds might offer big improvements already. I think we should make `grace_period` slightly larger per the recommendation in the docs: https://cloud.google.com/python/docs/reference/logging/latest/google.cloud.logging_v2.handlers.transports.background_thread.BackgroundThreadTransport

> The amount of time to wait for new logs before sending a new batch. It is strongly recommended to keep this smaller than the grace_period. This means this is effectively the longest amount of time the background thread will hold onto log entries before sending them to the server.

We can monitor the metrics after this and adjust more as required.

javanlacerda: Oliver's suggestion seems reasonable.

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -366,7 +368,19 @@ def configure_cloud_logging():
       'compute.googleapis.com/resource_name': socket.getfqdn().lower(),
       'bot_name': os.getenv('BOT_NAME', 'null'),
   }
-  handler = client.get_default_handler(labels=labels)
```
vitaliset, line 369: Depending on the machine, [`get_default_handler`](https://github.com/googleapis/python-logging/blob/42387bf63f8b362d24a094d83a0a8f66170758a4/google/cloud/logging_v2/client.py#L368-L390) gets `CloudLoggingHandler` or `StructuredLogHandler`. Despite the naming, `CloudLoggingHandler` also allows structured logging. I ran locally the structured introduced in #4722 and got the [correct logs](https://pantheon.corp.google.com/logs/query;cursorTimestamp=2025-03-25T20:56:33.252756Z;duration=PT5M;query=SEARCH%2528%22test%20carlolemos%20transport%22%2529%0Atimestamp%3D%222025-03-25T20:56:33.252756Z%22%0AinsertId%3D%221jsxldlfizp3xu%22%0Atimestamp%3D%222025-03-25T20:56:33.252756Z%22%0AinsertId%3D%221jsxldlfizp3xu%22?project=google.com:clusterfuzz), so I think using this handler instead of `get_default_handler` should be fine. 

Anyway, we have to pay attention when we deploy to make sure that we are receiving logs from all the different machine types.

PR #4740, opened 2025-03-24
[fix] move default labels to json formatter

javanlacerda:
It moves the `release` and `docker_image` default labels to into the json formatter. before it, this labels were omitted.

### DIFF
--- a/src/clusterfuzz/_internal/metrics/logs.py
+++ b/src/clusterfuzz/_internal/metrics/logs.py
@@ -204,6 +204,10 @@ class JsonFormatter(logging.Formatter):
             os.getpid(),
         'task_id':
             os.getenv('CF_TASK_ID', 'null'),
+        'release':
+            os.getenv('CLUSTERFUZZ_RELEASE', 'prod'),
+        'docker_image':
+            os.getenv('DOCKER_IMAGE', '')
     }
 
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -689,14 +691,18 @@ def clear(self):
...
+    log_contexts.delete(contexts)
 
 
 @contextlib.contextmanager
 def task_stage_context(stage: Stage):
   with wrap_log_context(contexts=[LogContextType.TASK]):
-    log_contexts.add_metadata('stage', stage)
-    yield
-    log_contexts.delete_metadata('stage')
+    try:
```
vitorguidi, line 704: Why the try catch? What is the present issue that this addresses, and how is it going to solve it?

javanlacerda: This is and ortogonal improvement. When the scope that is wrapped with the context has any error, the `tear down` for the context doesn't happen. I noticed that when updating the code for removing the labels from the logger call. It's also being addressed in the PR https://github.com/google/clusterfuzz/pull/4736, but I prefered to fix it asap on this pr

vitaliset (approved):

PR #4741, opened 2025-03-25
check clusterfuzz-config is up to date

javanlacerda:
It extends the logic to check that clusterfuzz git repo is up to date with origin/master to clusterfuzz-config

### DIFF
--- a/src/local/butler/deploy.py
+++ b/src/local/butler/deploy.py
@@ -370,19 +370,20 @@ def _update_redis(project):
                    '--project={project}'.format(project=project, region=region))
 
 
-def get_remote_sha():
+def get_remote_sha(git_dir: str = '.'):
   """Get remote sha of origin/master."""
-  _, remote_sha_line = common.execute('git ls-remote origin refs/heads/master')
+  _, remote_sha_line = common.execute(
+      f'git -C {git_dir} ls-remote origin refs/heads/master')
 
   return re.split(br'\s+', remote_sha_line)[0]
 
 
-def is_diff_origin_master():
+def is_diff_origin_master(git_dir: str = '.'):
   """Check if the current state is different from origin/master."""
-  common.execute('git fetch')
-  remote_sha = get_remote_sha()
-  _, local_sha = common.execute('git rev-parse HEAD')
-  _, diff_output = common.execute('git diff origin/master --stat')
+  common.execute(f'git -C {git_dir} fetch')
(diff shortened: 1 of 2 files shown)


src/local/butler/deploy.py
```diff
@@ -370,19 +370,20 @@ def _update_redis(project):
...
+def is_diff_origin_master(git_dir: str = '.'):
   """Check if the current state is different from origin/master."""
-  common.execute('git fetch')
-  remote_sha = get_remote_sha()
-  _, local_sha = common.execute('git rev-parse HEAD')
-  _, diff_output = common.execute('git diff origin/master --stat')
+  common.execute(f'git -C {git_dir} fetch')
+  remote_sha = get_remote_sha(git_dir)
+  _, local_sha = common.execute(f'git -C {git_dir} rev-parse HEAD')
+  _, diff_output = common.execute(f'git -C {git_dir} diff origin/master --stat')
```
vitorguidi, line 386: There is the edge case of candidate releases. We should enable a second branch to deplyo candidate, otherwise we lose our only prog rollout mechanism so far.

vitorguidi: butler deploy ... --release=candidate

vitorguidi (approved):

src/local/butler/deploy.py
javanlacerda: the `--force` will keep this ability, because we're just checking just as we do for clusterfuzz, but not blocking.

src/local/butler/deploy.py
```diff
@@ -554,15 +555,18 @@ def execute(args):
   appengine.build_templates()
 
   if not is_ci and not args.staging:
-    if is_diff_origin_master():
+    if is_diff_origin_master() or is_diff_origin_master(
+        environment.get_config_directory()):
```
ViniciustCosta, line 559: I think the `environment.get_config_directory` is not the right one to use here. Maybe you should retrieve the config directory from the `-c` or `--config-dir` argument from butler.

ViniciustCosta: Nevermind, I see that `CONFIG_DIR_OVERRIDE` is getting set with the `--config-dir` args when deploy is executed, so this should be fine.

ViniciustCosta (approved):
LGTM

oliverchang (commented):
LGTM, thanks!

PR #4745, opened 2025-03-28
[Logging] Add and propagate common logging metadata

ViniciustCosta:
### Description
This PR adds important common metadata that was missing from the current log entries:
- Clusterfuzz source version
- Clusterfuzz config version
- Instance ID
- OS type
- OS version

At first, to propagate the context containing these labels throughout the stack, we created global environment variables and initialized them in the entry points for the main clusterfuzz use cases. However, this changed during review in favor of propagating with a metadata in the logs context, check the updated comment: https://github.com/google/clusterfuzz/pull/4745#issuecomment-2802477155

In addition, this PR also enforces a more structured string to the manifest file in order to facilitate reading and parsing it later. Unit tests to validate this structure were added. The format follows: `<timestamp>-utc-<clusterfuzz commit sha>-<user name>-<clusterfuzz-config commit sha>-<prod|release>`. 

### Tests
* Onebox deployment to GCE VM (google3) ([Link to logs](https://cloudlogging.app.goo.gl/hmHgaSmb94HVuccp6)):
image

* Onebox deployment to GCE preemptible VMs (chrome) ([Link to logs](https://cloudlogging.app.goo.gl/VSV6PcnPy6AgWM4e7)):
image

### DIFF
--- a/src/clusterfuzz/_internal/base/utils.py
+++ b/src/clusterfuzz/_internal/base/utils.py
@@ -935,7 +935,8 @@ def get_version_commits() -> tuple[str, str] | None:
   """Return the commit hash for source and config revision."""
   source_version_data = (current_source_version() or '').split('-')
   if len(source_version_data) >= 5:
-    source_commit, config_commit = source_version_data[2], source_version_data[4]
+    source_commit, config_commit = source_version_data[2], source_version_data[
+        4]
     return (source_commit, config_commit)
   return None
 
--- a/src/clusterfuzz/_internal/base/utils.py
+++ b/src/clusterfuzz/_internal/base/utils.py
@@ -21,6 +21,7 @@ import hashlib
 import inspect
 import os
 import random
+import re
 import sys
 import time
 import urllib.parse
@@ -916,7 +917,7 @@ def current_project():
   return environment.get_value('PROJECT_NAME', default_project_name())
 
 
-def current_source_version():
+def current_source_version() -> str | None:
   """Return the current source revision."""
   source_version_override = environment.get_value('SOURCE_VERSION_OVERRIDE')
   if source_version_override:
@@ -930,18 +931,26 @@ def current_source_version():
       try:
         file_data = file_data.strip().decode('utf-8')
       except UnicodeDecodeError as e:
-        logs.error(f'Error decoding manifest file content - {str(e)}')
+        logs.error(f'Error decoding manifest file content - {str(e.reason)}')
         file_data = None
     return file_data
 
 
-def get_version_commits() -> tuple[str, str] | None:
-  """Return the commit hash for source and config revision."""
-  source_version_data = (current_source_version() or '').split('-')
-  if len(source_version_data) >= 5:
-    source_commit = source_version_data[2]
-    config_commit = source_version_data[4]
-    return (source_commit, config_commit)
+def parse_manifest_data(file_data: str | None) -> dict[str, str] | None:
+  """Parse the manifest revision data and return it as a dict."""
+  if not file_data:
+    return None
+
+  pattern = r'^([0-9]{14}-utc)-([0-9a-f]+)-(.*)-([0-9a-f]+)-(prod|staging)'
+  result = re.match(pattern, file_data)
+  if result:
+    return {
+        'timestamp': result.group(1),
+        'cf_commit_sha': result.group(2),
+        'user': result.group(3),
+        'cf_config_commit_sha': result.group(4),
+        'release': result.group(5)
+    }
   return None
 
 
@@ -957,10 +966,12 @@ def get_instance_name() -> str:
 
 def set_common_log_context() -> None:
   """Set env variables to propagate common context used by logs."""
-  source_version = get_version_commits()
-  if source_version:
-    environment.set_value('CF_VERSION', source_version[0])
-    environment.set_value('CF_CONFIG_VERSION', source_version[1])
+  parsed_source_version = parse_manifest_data(
+      file_data=current_source_version())
+  if parsed_source_version:
+    environment.set_value('CF_VERSION', parsed_source_version['cf_commit_sha'])
+    environment.set_value('CF_CONFIG_VERSION',
+                          parsed_source_version['cf_config_commit_sha'])
   environment.set_value('OS_TYPE', environment.platform())
   environment.set_value('INSTANCE_ID', get_instance_name())
 
--- a/src/clusterfuzz/_internal/base/utils.py
+++ b/src/clusterfuzz/_internal/base/utils.py
@@ -964,18 +964,6 @@ def get_instance_name() -> str:
   return environment.get_value('BOT_NAME', '')
 
 
-def set_common_log_context() -> None:
-  """Set env variables to propagate common context used by logs."""
-  parsed_source_version = parse_manifest_data(
-      file_data=current_source_version())
-  if parsed_source_version:
-    environment.set_value('CF_VERSION', parsed_source_version['cf_commit_sha'])
-    environment.set_value('CF_CONFIG_VERSION',
-                          parsed_source_version['cf_config_commit_sha'])
-  environment.set_value('OS_TYPE', environment.platform())
-  environment.set_value('INSTANCE_ID', get_instance_name())
-
-
 def read_from_handle_truncated(file_handle, max_len):
   """Read from file handle, limiting output to |max_len| by removing output in
   the middle."""
--- a/src/clusterfuzz/_internal/base/utils.py
+++ b/src/clusterfuzz/_internal/base/utils.py
@@ -927,7 +927,11 @@ def current_source_version():
   if os.path.exists(local_manifest_path):
     file_data = read_data_from_file(local_manifest_path, eval_data=False)
     if file_data is not None:
-      file_data = file_data.strip().decode('utf-8')
+      try:
+        file_data = file_data.strip().decode('utf-8')
+      except UnicodeDecodeError as e:
+        logs.error(f'Error decoding manifest file content - {str(e)}')
+        file_data = None
     return file_data
 
 
--- a/src/clusterfuzz/_internal/base/utils.py
+++ b/src/clusterfuzz/_internal/base/utils.py
@@ -922,13 +922,42 @@ def current_source_version():
   if source_version_override:
     return source_version_override
 
-  root_directory = environment.get_value('ROOT_DIR')
+  root_directory = environment.get_value('ROOT_DIR', '.')
   local_manifest_path = os.path.join(root_directory, LOCAL_SOURCE_MANIFEST)
   if os.path.exists(local_manifest_path):
-    return read_data_from_file(
-        local_manifest_path, eval_data=False).strip().decode('utf-8')
+    file_data = read_data_from_file(local_manifest_path, eval_data=False)
+    if file_data is not None:
+      file_data = file_data.strip().decode('utf-8')
+    return file_data
 
+
+def get_version_commits() -> tuple[str, str] | None:
+  """Return the commit hash for source and config revision."""
+  source_version_data = (current_source_version() or '').split('-')
+  if len(source_version_data) >= 5:
+    source_commit, config_commit = source_version_data[2], source_version_data[4]
+    return (source_commit, config_commit)
   return None
+
+
+def get_instance_name() -> str:
+  """Return the instance name based on the running environment."""
+  if environment.is_running_on_k8s():
+    return environment.get_value('HOSTNAME', '')
+  if environment.is_running_on_app_engine():
+    return environment.get_value('GAE_INSTANCE', '')
+  # Use bot name here as instance as that's more useful to us.
+  return environment.get_value('BOT_NAME', '')
+
+
+def set_common_log_context() -> None:
+  """Set env variables to propagate common context used by logs."""
+  source_version = get_version_commits()
+  if source_version:
+    environment.set_value('CF_VERSION', source_version[0])
+    environment.set_value('CF_CONFIG_VERSION', source_version[1])
+  environment.set_value('OS_TYPE', environment.platform())
+  environment.set_value('INSTANCE_ID', get_instance_name())
 
 
 def read_from_handle_truncated(file_handle, max_len):
--- a/src/clusterfuzz/_internal/base/utils.py
+++ b/src/clusterfuzz/_internal/base/utils.py
@@ -949,7 +949,7 @@ def parse_manifest_data(file_data: str | None) -> dict[str, str] | None:
         'cf_commit_sha': result.group(2),
         'user': result.group(3),
         'cf_config_commit_sha': result.group(4),
-        'release': result.group(5)
+        'appengine_release': result.group(5)
     }
   return None
 
--- a/src/clusterfuzz/_internal/base/utils.py
+++ b/src/clusterfuzz/_internal/base/utils.py
@@ -935,8 +935,8 @@ def get_version_commits() -> tuple[str, str] | None:
   """Return the commit hash for source and config revision."""
   source_version_data = (current_source_version() or '').split('-')
   if len(source_version_data) >= 5:
-    source_commit, config_commit = source_version_data[2], source_version_data[
-        4]
+    source_commit = source_version_data[2]
+    config_commit = source_version_data[4]
     return (source_commit, config_commit)
   return None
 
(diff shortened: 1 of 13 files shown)


ViniciustCosta:
@vitorguidi @javanlacerda I still have to add/fix the tests, but PTAL.

src/clusterfuzz/_internal/base/utils.py
```diff
@@ -922,15 +922,45 @@ def current_source_version():
...
-  root_directory = environment.get_value('ROOT_DIR')
+  root_directory = environment.get_value('ROOT_DIR', '.')
   local_manifest_path = os.path.join(root_directory, LOCAL_SOURCE_MANIFEST)
   if os.path.exists(local_manifest_path):
-    return read_data_from_file(
-        local_manifest_path, eval_data=False).strip().decode('utf-8')
-
+    file_data = read_data_from_file(local_manifest_path, eval_data=False)
+    if file_data is not None:
+      file_data = file_data.strip().decode('utf-8')
```
javanlacerda, line 930: Can this fail? Should we handle this for failing?

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -646,13 +655,26 @@ class LogContextType(enum.Enum):
...
   TASK = 'task'
   TESTCASE = 'testcase'
   PROGRESSION = 'progression'
 
   def get_extras(self) -> NamedTuple:
     """Get the structured log for a given context"""
-    if self == LogContextType.TASK:
+    if self == LogContextType.COMMON:
+      try:
+        return CommonLogStruct(
```
javanlacerda, line 667: You think we could implement this logic inside the class and then having this struct as default?

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -625,6 +626,14 @@ class GenericLogStruct(NamedTuple):
...
 
 
+class CommonLogStruct(NamedTuple):
+  clusterfuzz_version: str
+  clusterfuzz_config_version: str
+  instance_id: str
+  operating_system: str
+  os_version: str
+
+
```
javanlacerda, line 636:
```suggestion
class CommonLogStruct(NamedTuple):
  clusterfuzz_version: str = os.getenv('CF_VERSION', 'null')
  clusterfuzz_config_version: str = os.getenv('CF_CONFIG_VERSION', 'null')
  instance_id: str = os.getenv('INSTANCE_ID', 'null')
  operating_system: str = os.getenv('OS_TYPE', 'null')
  os_version: str = platform.release()


```

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -646,13 +655,26 @@ class LogContextType(enum.Enum):
...
+      try:
+        return CommonLogStruct(
+            clusterfuzz_version=os.getenv('CF_VERSION', 'null'),
+            clusterfuzz_config_version=os.getenv('CF_CONFIG_VERSION', 'null'),
+            instance_id=os.getenv('INSTANCE_ID', 'null'),
+            operating_system=os.getenv('OS_TYPE', 'null'),
+            os_version=platform.release())
+      except Exception as e:
+        error(str(e), ignore_context=True)
+        return GenericLogStruct()
```
javanlacerda, line 675:
```suggestion
      return CommonLogStruct()
```

javanlacerda: Then we could delete the generic struct and use the common instead

src/clusterfuzz/_internal/base/utils.py
ViniciustCosta: `read_data_from_file` deals with most exceptions while reading the file, but I suppose that the decoding may raise an exception. I will add a handler for this!

src/clusterfuzz/_internal/metrics/logs.py
ViniciustCosta: IMO it is better (for consistency) to have the Struct classes only define the structured fields and let all logic on how to retrieve their values in a single place, which is the `get_extras` method, in this case.

src/clusterfuzz/_internal/metrics/logs.py
ViniciustCosta: I think we would still need the generic struct in order to return it in case of some error/exception (including when trying to retrieve the common context).

src/clusterfuzz/_internal/metrics/logs.py
javanlacerda: It makes sense. it works fine as well

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -723,7 +745,7 @@ class LogContexts(metaclass=Singleton):
   """Class to keep the log contexts and metadata"""
 
   def __init__(self):
-    self.contexts: list[LogContextType] = []
+    self.contexts: list[LogContextType] = [LogContextType.COMMON]
```
javanlacerda, line 748: It need to be added to the `clear` function as well.

src/clusterfuzz/_internal/metrics/monitor.py
```diff
@@ -568,14 +568,7 @@ def _initialize_monitored_resource():
...
 
-  # Use bot name here instance as that's more useful to us.
-  if environment.is_running_on_k8s():
-    instance_name = environment.get_value('HOSTNAME')
-  elif environment.is_running_on_app_engine():
-    instance_name = environment.get_value('GAE_INSTANCE')
-  else:
-    instance_name = environment.get_value('BOT_NAME')
-  _monitored_resource.labels['instance_id'] = instance_name
+  _monitored_resource.labels['instance_id'] = utils.get_instance_name()
```
javanlacerda, line 571: 🆒

src/clusterfuzz/_internal/metrics/logs.py
ViniciustCosta: Good catch, fixed!

src/clusterfuzz/_internal/base/utils.py
```diff
@@ -922,15 +922,49 @@ def current_source_version():
...
+  """Return the instance name based on the running environment."""
+  if environment.is_running_on_k8s():
+    return environment.get_value('HOSTNAME', '')
+  if environment.is_running_on_app_engine():
+    return environment.get_value('GAE_INSTANCE', '')
+  # Use bot name here as instance as that's more useful to us.
+  return environment.get_value('BOT_NAME', '')
+
+
+def set_common_log_context() -> None:
```
jonathanmetzman, line 958: Why not put this in logs.py? Seems to make more sense there.

src/clusterfuzz/_internal/base/utils.py
```diff
@@ -922,15 +922,49 @@ def current_source_version():
...
+    return file_data
+
+
+def get_version_commits() -> tuple[str, str] | None:
+  """Return the commit hash for source and config revision."""
+  source_version_data = (current_source_version() or '').split('-')
+  if len(source_version_data) >= 5:
+    source_commit = source_version_data[2]
+    config_commit = source_version_data[4]
+    return (source_commit, config_commit)
```
jonathanmetzman, line 944: This is a bit dangerous because it's not kept in sync with the code creating this version string.
Can we do a test that ensures we parse the output correctly?

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -625,6 +626,14 @@ class GenericLogStruct(NamedTuple):
   pass
 
 
+class CommonLogStruct(NamedTuple):
```
jonathanmetzman, line 629: What do you think of doing set_common_log_context in the first call to a log function, so it's impossible to forget to do.

jonathanmetzman: I think this suggested approach will work much better with multiprocessing code since we typically use spawn.

src/clusterfuzz/_internal/base/utils.py
ViniciustCosta: I agree, and that was the initial intention, but problems with circular imports got in the way. Basically, the logs module is so ubiquitously imported that attempts to use methods from other modules will cause these issues with circular import.

src/clusterfuzz/_internal/base/utils.py
ViniciustCosta: Yeah, I agree. I will add some unit tests to the utils module to verify this as well. Thanks

src/clusterfuzz/_internal/tests/core/base/utils_test.py
```diff
@@ -557,3 +557,145 @@ def test_not_exists(self, buffer_length):
...
+
+  def test_missing_commits(self):
+    """Test parsing manifest data without commits versions."""
+    file_data = '20250402153042-utc--abcd312-cad6977-prod'
+    self.assertIsNone(utils.parse_manifest_data(file_data))
+
+    file_data = '20250402153042-utc-40773ac0-abcd312--prod'
+    self.assertIsNone(utils.parse_manifest_data(file_data))
+
+  def test_additional_prefix(self):
```
vitorguidi, line 698: What if an user hides a commit value? as in:

user = john-X
commit = -Y

Do we parse the commit as X or Y? You are using (.)* as the user representation, so this is a possible situation

vitorguidi:
> This PR adds important common metadata that was missing from the current log entries:
> 
> * Clusterfuzz source version
> * Clusterfuzz config version
> * Instance ID
> * OS type
> * OS version
> 
> In order to propagate the context containing these labels throughout the stack, we created global environment variables and initialized them in the entry points for the main clusterfuzz use cases.

It would be good to perform a candidate release/deployment to test this before we merge, since there might be edge cases or non obvious conditions where the logging might break.

However, since we are splitting the PR that sets up the logging context, from the one that applies it, we cannot do this on this PR. We should do this on the one that actually logs stuff.

@ViniciustCosta this is actually the case, right? We need to annotate the entrypoints with the log context before we are able to actually see the logs, which will be done in a follow up pr?

src/clusterfuzz/_internal/tests/core/base/utils_test.py
ViniciustCosta: I'm not sure if I got it, but I'm enforcing the string format in `common.py` (with unit tests) to be separated by '-'  containing all defined fields.
So, for instance, if the username is `john-X` and the CF/config commits are `Z ` and `Y`, the string would be `timestamp-utc-Z-john-X-Y-prod` , which would be correctly parsed (check `test_valid_manifest_staging` which tests a similar case).

ViniciustCosta:
> It would be good to perform a candidate release/deployment to test this before we merge, since there might be edge cases or non obvious conditions where the logging might break.

I agree with the candidate deploy since this will touch almost all logging, so if something breaks, we could potentially miss a lot of logs.

> However, since we are splitting the PR that sets up the logging context, from the one that applies it, we cannot do this on this PR. We should do this on the one that actually logs stuff.
> 
> @ViniciustCosta this is actually the case, right? We need to annotate the entrypoints with the log context before we are able to actually see the logs, which will be done in a follow up pr?

Not actually, the way this is being done is that the common context is added by default in the `LogsContexts` class, so we don't need to wrap the entrypoints with the log context. However, as it is now, we do need to set up the environment variable in those entrypoints, which it is being done (for most cases I found) in this PR, so it should log the common metadata.

ViniciustCosta:
Following our discussions, I've changed the approach on how to propagate the common context in order to avoid the need to set env vars in all clusterfuzz entry points.

Now, the common context is being propagated via metadata in the logs singleton. The information needed is retrieved the first time a logs emit is called and the remaining execution uses this same context.

One possible downside here is that, if it fails to retrieve the information the first time, the context will be empty throughout the execution. However, the trade-off is to avoid creating an overhead of having to get this metadata every time it logs and continuously failing to retrieve it.

@javanlacerda @jonathanmetzman @vitorguidi PTAL (sorry for the large PR, but a lot of it is unit tests and some code style).

src/clusterfuzz/_internal/base/utils.py
```diff
@@ -916,21 +917,53 @@ def current_project():
...
 
-def current_source_version():
+def current_source_version() -> str | None:
   """Return the current source revision."""
   source_version_override = environment.get_value('SOURCE_VERSION_OVERRIDE')
   if source_version_override:
     return source_version_override
 
-  root_directory = environment.get_value('ROOT_DIR')
+  root_directory = environment.get_value('ROOT_DIR', '.')
```
vitorguidi, line 926: Why this change? What behavior do you want to avoid with this default value?

src/clusterfuzz/_internal/base/utils.py
```diff
@@ -916,21 +917,53 @@ def current_project():
...
+        file_data = None
+    return file_data
 
+
+def parse_manifest_data(file_data: str | None) -> dict[str, str] | None:
+  """Parse the manifest revision data and return it as a dict."""
+  if not file_data:
+    return None
+
+  pattern = r'^([0-9]{14}-utc)-([0-9a-f]+)-(.*)-([0-9a-f]+)-(prod|staging)'
```
vitorguidi, line 944: I went out to read some more code on this, and this STAGING thing seems to be dead code.

It is related to deploying to a second APPENGINE environment. And I think this has not been used for a long time, and thus dead code.

It is not in the scope of this project to get rid of dead code though, but important to pay attention to.

src/clusterfuzz/_internal/base/utils.py
```diff
@@ -916,21 +917,53 @@ def current_project():
...
+  result = re.match(pattern, file_data)
+  if result:
+    return {
+        'timestamp': result.group(1),
+        'cf_commit_sha': result.group(2),
+        'user': result.group(3),
+        'cf_config_commit_sha': result.group(4),
+        'release': result.group(5)
```
vitorguidi, line 952: This log line is misleading. If anything, it should be the "candidate" or "prod" release from the python 3.7 update. If we are reading bot or cron logs, we dont really care about anything app engine related.

https://github.com/google/clusterfuzz/pull/4039

However, since we are killing it with staging environments from the deployment project, both "candidate" and appengine "staging" will end up being dead code which we should get rid of. 

The good news is, you can delete this log line, and the change that adds prod/staging to the manifest =)

src/clusterfuzz/_internal/base/utils.py
ViniciustCosta: Mainly because the next line tries to join this with the manifest file path, so if `ROOT_DIR` is not set, this could break (also the static type check keeps complaining about it).

src/clusterfuzz/_internal/base/utils.py
ViniciustCosta: I agree that we should get rid of the dead code (which seems a lot in our codebase), but IMO we should do this in another PR after making sure that this is not useful anymore and removing all other parts in the butler deploy that are responsible for deploying to this environment.

For instance, in the current documentation, there is still a reference for using this staging for testing UI changes [Stagin changes](https://google.github.io/clusterfuzz/contributing-code/staging-changes/), although I'm not sure if it still works, we should double check it.

src/clusterfuzz/_internal/base/utils.py
ViniciustCosta: This is not actually adding anything to the logs, this method just parses and returns the manifest file data in a structured way. For the common logs context, only the commit hashes are used in the structured fields.

Regarding the deletion of this "staging", as I commented, I think we should double check it and maybe delete all things related in another PR (I will create a bug to track this).

The change I made to the manifest only enforced a format to be easier to parse it later, the "staging" was already being added and I enforced that either "prod" or "staging" was the end of the string.

vitorguidi: Sgtm
Can we at least rename this dict field to appengine_release, to avoid confusion?

ViniciustCosta: Sure, I'll change it!

javanlacerda (approved):

vitorguidi:
> This PR adds important common metadata that was missing from the current log entries:
> 
> * Clusterfuzz source version
> * Clusterfuzz config version
> * Instance ID
> * OS type
> * OS version
> 
> In order to propagate the context containing these labels throughout the stack, we created global environment variables and initialized them in the entry points for the main clusterfuzz use cases.
> 
> In addition, this PR also enforces a more structured string to the manifest file in order to facilitate reading and parsing it later. Unit tests to validate this structure were added. The format follows: `<timestamp>-utc-<clusterfuzz commit sha>-<user name>-<clusterfuzz-config commit sha>-<prod|release>`.

LGTM, it is wise to do a candidate release on a subset of bots for us to have confidence this works

vitorguidi (approved):
lgtm. evidence was provided internally to a onebox deployment where the logs were showing, so I am comfortable merging the code

PR #4748, opened 2025-04-01
[Fix] Fix metadata to retrieve fuzz target in testcase-based logging

ViniciustCosta:
Latest changes to add the fuzz target to the testcase-based structured logs was failing when retrieving the fuzz target as a key ID: [error](https://pantheon.corp.google.com/errors/detail/CNuqgM-PtKvgaA;locations=global;time=P30D?project=google.com:clusterfuzz&utm_source=error-reporting-notification&utm_medium=email&utm_content=new-error&e=-13802955&mods=logs_tg_prod&inv=1&invt=Abtmmg). 
This PR attempts to fix it by using the "binary" field from the FuzzTarget entity, which is also used in the ClusterFuzz frontend to represent the fuzz target.

### DIFF
--- a/src/clusterfuzz/_internal/metrics/logs.py
+++ b/src/clusterfuzz/_internal/metrics/logs.py
@@ -683,10 +683,10 @@ class LogContextType(enum.Enum):
           return GenericLogStruct()
 
         fuzz_target = testcase.get_fuzz_target()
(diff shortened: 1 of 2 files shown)


javanlacerda (approved):
Thanks for fixing it!

PR #4749, opened 2025-04-01
bump ubuntu to 24

javanlacerda:
It bumps the ubuntu version for basic tests workflow as the 20.04 version will be deprecated. https://github.com/actions/runner-images/issues/11101


ViniciustCosta (approved):
Thanks!

PR #4752, opened 2025-04-03
MNT Truncate exception messages in logs to fit GCP limits

vitaliset:
Recently, @paulsemel [observed](https://chat.google.com/room/AAAALIZtBQE/PsfB7Jr1zFk/Vu0oVjjUFoo?cls=10) that some logs were not being received because they exceeded the size limit allowed by GCP (for [example](https://pantheon.corp.google.com/logs/query;cursorTimestamp=2025-04-01T08:19:42.923556Z;endTime=2025-04-01T08:30:00.000Z;query=%22regression%22%0Aseverity%20%3D%20%22ERROR%22%0A%22Failed%20to%20submit%22%0Atimestamp%3D%222025-04-01T08:19:42.923556Z%22%0AinsertId%3D%22nl5t4lfbvzgo8%22%0Atimestamp%3D%222025-04-01T08:19:42.923556Z%22%0AinsertId%3D%22nl5t4lfbvzgo8%22;startTime=2025-04-01T00:35:00.000Z?project=google.com:clusterfuzz)). This PR applies the same logic we use to truncate very large messages in the traceback of exceptions that are added to the log messages.

b/408173151

### DIFF
--- a/src/clusterfuzz/_internal/metrics/logs.py
+++ b/src/clusterfuzz/_internal/metrics/logs.py
@@ -35,10 +35,11 @@ if TYPE_CHECKING:
   from clusterfuzz._internal.datastore.data_types import Testcase
 
 # The maximum allowed log entry size is 256 KB for GCP. We set the
-# STACKDRIVER_LOG_MESSAGE_LIMIT to approximately 100 KB to accommodate both the
-# primary log message and potential traceback exceptions. This reserves roughly
-# 200 KB (100 KB each) for message content, leaving sufficient space for
-# structured logging metadata within the 256 KB total limit.
+# STACKDRIVER_LOG_MESSAGE_LIMIT to approximately 80 KB (under 100 KB) to
+# accommodate both the primary log message and potential traceback exceptions.
+# This reserves roughly up to 200 KB (up to 100 KB each) for message content,
+# leaving sufficient space for structured logging metadata within the 256 KB
+# total limit.
 STACKDRIVER_LOG_MESSAGE_LIMIT = 80000
 LOCAL_LOG_MESSAGE_LIMIT = 100000
 LOCAL_LOG_LIMIT = 500000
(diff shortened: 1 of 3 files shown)


src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -264,8 +264,13 @@ def update_entry_with_exc(entry, exc_info):
   if exc_info[0]:
     # we need to set the result of traceback.format_exception to the field
     # `message`. And we move our
-    entry['message'] += '\n' + ''.join(
+    formatted_exception = ''.join(
         traceback.format_exception(exc_info[0], exc_info[1], exc_info[2]))
+
+    # Truncate the formatted exception if it's too long.
+    truncated_exception = truncate(formatted_exception,
```
ViniciustCosta, line 271: Maybe you should truncate after appending it to the message?

ViniciustCosta (approved):
LGTM!

src/clusterfuzz/_internal/metrics/logs.py
vitaliset: I think the most important parts of a traceback are the beginning and the end. The current `truncate` [implementation](https://github.com/google/clusterfuzz/blob/ee841591eaf83591a8c680d22db29ce13b34dd58/src/clusterfuzz/_internal/metrics/logs.py#L177-L186) follows this logic by keeping equal portions from the beginning and the end of the provided string. If I truncate the appended message, I might miss the ending parts of the log or the beginning of the traceback.

IIUC, `clusterfuzz._internal.metrics.logs.STACKDRIVER_LOG_MESSAGE_LIMIT` aims to [keep the text size under 100KB](https://github.com/google/clusterfuzz/blob/ee841591eaf83591a8c680d22db29ce13b34dd58/src/clusterfuzz/_internal/metrics/logs.py#L37), while the log entry maximum is 256KB. So, applying it twice (once for the actual message and once for the traceback) seemed like an acceptable approach, but I don't mind revisiting it.

ViniciustCosta: Oh I see, that makes a lot of sense! Thanks!

jonathanmetzman: I agree. I think it would be good to put this explanation in a comment though!

jonathanmetzman (approved):
lgtm

PR #4754, opened 2025-04-03
[Logging] check and log exceptions in log context

javanlacerda:
It adds a catch for exceptions that are raised during the contexts, and log it to have this log with the structured logs.

### DIFF
--- a/src/clusterfuzz/_internal/metrics/logs.py
+++ b/src/clusterfuzz/_internal/metrics/logs.py
@@ -768,6 +768,11 @@ def task_stage_context(stage: Stage):
     try:
       log_contexts.add_metadata('stage', stage)
       yield
+    except Exception as e:
+      error(
+          message='Error during task.',
+          error_stack_trace=traceback.format_exc())
+      raise e
     finally:
       log_contexts.delete_metadata('stage')
 
@@ -778,6 +783,11 @@ def testcase_log_context(testcase: "Testcase"):
     try:
       log_contexts.add_metadata('testcase', testcase)
       yield
+    except Exception as e:
+      error(
+          message='Error during testcase context.',
+          error_stack_trace=traceback.format_exc())
+      raise e
     finally:
       log_contexts.delete_metadata('testcase')
 
--- a/src/clusterfuzz/_internal/metrics/logs.py
+++ b/src/clusterfuzz/_internal/metrics/logs.py
@@ -764,14 +764,13 @@ def wrap_log_context(contexts: list[LogContextType]):
 
 @contextlib.contextmanager
 def task_stage_context(stage: Stage):
+  """Creates a task context for a given stage"""
   with wrap_log_context(contexts=[LogContextType.TASK]):
     try:
       log_contexts.add_metadata('stage', stage)
       yield
     except Exception as e:
-      error(
-          message='Error during task.',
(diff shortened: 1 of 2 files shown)


ViniciustCosta:
Could you add some unit tests to check if this works correctly?

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -768,6 +768,11 @@ def task_stage_context(stage: Stage):
     try:
       log_contexts.add_metadata('stage', stage)
       yield
+    except Exception as e:
+      error(
+          message='Error during task.',
+          error_stack_trace=traceback.format_exc())
```
vitorguidi, line 774: Why do not just error(e), so we see the string representation of the string?

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -768,6 +768,11 @@ def task_stage_context(stage: Stage):
     try:
       log_contexts.add_metadata('stage', stage)
       yield
+    except Exception as e:
+      error(
+          message='Error during task.',
```
vitorguidi, line 773: IMO you can just omit the 'Error during task' message, since you most likely will read this when intentionally querying for the appropriate task context (task_name=X, job=Y, etc). It suffices to display the exception alone

src/clusterfuzz/_internal/metrics/logs.py
ViniciustCosta: The `error()` log already appends the exception string in the logs message

src/clusterfuzz/_internal/metrics/logs.py
javanlacerda: message is a requirement for the logger, you think we could send it as an empty string? I agree that it is redundant, but it can help as to be sure that the log was created excatly here, wdyt?

src/clusterfuzz/_internal/tests/core/metrics/logs_test.py
```diff
@@ -550,6 +550,39 @@ def test_progression_log_context(self):
...
+  def test_log_exception(self):
+    """It checks that task_stage_context catch and log the
+    error raised during the decorated scope
+    """
+    logger = mock.MagicMock()
+    self.mock.get_logger.return_value = logger
+
+    with logs.task_stage_context(logs.Stage.PREPROCESS):
+      try:
+        exception = Exception("msg")
```
jonathanmetzman, line 562: nit: single quote

jonathanmetzman (approved):
lgtm

src/clusterfuzz/_internal/tests/core/metrics/logs_test.py
```diff
@@ -550,6 +550,39 @@ def test_progression_log_context(self):
             },
         })
 
+  def test_log_exception(self):
```
jonathanmetzman, line 553: Change to: "Checks that the task_stage_context catches and logs the error raised in the decorated scope."
Can we change the function name to be more descriptive.

vitorguidi (approved):
Tyvm for fixing this, it is very valuable

src/clusterfuzz/_internal/metrics/logs.py
vitorguidi: sgtm

PR #4756, opened 2025-04-04
Add regression to structured logs

ViniciustCosta:


src/clusterfuzz/_internal/bot/tasks/utasks/regression_task.py
```diff
@@ -727,27 +728,33 @@ def utask_postprocess(output: uworker_msg_pb2.Output) -> None:  # pylint: disabl
...
-          output.uworker_input.testcase_id)
-      task_creation.mark_unreproducible_if_flaky(testcase, 'regression', False)
-
-  if output.error_type != uworker_msg_pb2.ErrorType.NO_ERROR:  # pylint: disable=no-member
-    _ERROR_HANDLER.handle(output)
-    return
-
-  save_regression_range(output)
+  # Try to retrieve the testcase early for logs context.
+  try:
```
javanlacerda, line 733: Is that try catch really necessary? I mean, we do have several other points on the code that assumes that it will exist correctly, in the preprocess, for instance. I'm just concerned if we are adding complexity to handle a `None` return for cases that will never happen.

ViniciustCosta: I was not sure about this, but my thoughts here were that the `get_testcase_by_id` raises an exception if it fails and without this try-catch, it could break the execution early on (although I agree that it is very unlikely). So to avoid failure solely due to adding this log context, I found it safer to handle it.

Also, there is logic to handle a `None` testcase in the log context ([here](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/metrics/logs.py#L681))

ViniciustCosta: But yeah, if this is adding unnecessary complexity, I am fine with removing it

vitorguidi: Remember that pre and post process run in different instances, and the code versions are not in sync atm. This is not a safe assumption to make.

javanlacerda: yep. but this is the as is, right? for progression, this validation doesn't exist as well. I think we don't need to solve a problem that doesn't exist

ViniciustCosta: What I meant was to add this handler only for cases like this, where I had to explicitly add the `get_testcase_by_id` to the beginning of the function for the logs (other cases we assume that it's working).
On the other hand, I agree that the `get_testcase_by_id` is used everywhere without this handling, so maybe it is not needed.

ViniciustCosta: I removed the try-catch in this case to avoid unnecessary complexity, since the next method called here tries to retrieve the testcase.

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -668,18 +666,19 @@ def get_extras(self) -> NamedTuple:
             task_argument=task_argument,
             stage=stage,
             task_job_name=task_job_name)
-      except Exception as e:
+      except:
         # This flag is necessary to avoid
-        # infinite loop in this context verification
-        error(str(e), ignore_context=True)
+        # infinite loop in this context verification.
+        error('Error retrieving context for task logs.', ignore_context=True)
```
javanlacerda, line 672: 👍

javanlacerda (approved):
LGTM! Thanks!

PR #4757, opened 2025-04-08
[Fix] Add fuzz target as argument for logs context to avoid access deny to DB

ViniciustCosta:
After adding structured logs to testcase-based tasks (https://github.com/google/clusterfuzz/pull/4736), a new error group started showing up due to insufficient permissions/no current context for accessing datastore. This is probably due to querying for the testcase's fuzz target during `utask_main` execution in logs.

This tries to solve it by adding the fuzz target as an argument for the logs context, since we can retrieve it from the datastore entity in trusted part and from the protobuf during the untrusted part of the task.

### DIFF
--- a/src/clusterfuzz/_internal/metrics/logs.py
+++ b/src/clusterfuzz/_internal/metrics/logs.py
@@ -32,6 +32,7 @@ from typing import TYPE_CHECKING
 
 # This is needed to avoid circular import
 if TYPE_CHECKING:
+  from clusterfuzz._internal.datastore.data_types import FuzzTarget
   from clusterfuzz._internal.datastore.data_types import Testcase
 
 # The maximum allowed log entry size is 256 KB for GCP. We set the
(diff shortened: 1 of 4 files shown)


src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -796,30 +802,35 @@ def task_stage_context(stage: Stage):
 
 
 @contextlib.contextmanager
```
javanlacerda, line 804: Could you add a comment describing why we need to receive the fuzz_target as well?

jonathanmetzman:
Can you fix the test failure

src/clusterfuzz/_internal/tests/core/metrics/logs_test.py
```diff
@@ -557,11 +557,14 @@ def test_progression_log_context(self):
         fuzzer_name="test_fuzzer", job_type='test_job')
     testcase.put()
 
-    with logs.progression_log_context(testcase):
+    with logs.progression_log_context(testcase, fuzz_target):
```
javanlacerda, line 560: Could you add a test case where the fuzz_target is not passed, but the testcase metadata has the binary?

ViniciustCosta: Done!

PR #4758, opened 2025-04-08
[Logging] Move error to warning for log context logs

javanlacerda:
We're moving the logging level from error to warning for exceptions that was caught during log context. We need to do this because this errors are already handled in the uper level, but we want to log it as warning to have visibility with the strucutred logs.

### DIFF
--- a/src/clusterfuzz/_internal/metrics/logs.py
+++ b/src/clusterfuzz/_internal/metrics/logs.py
@@ -795,7 +795,7 @@ def task_stage_context(stage: Stage):
       log_contexts.add_metadata('stage', stage)
       yield
     except Exception as e:
-      error(message='Error during task.')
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -795,7 +795,7 @@ def task_stage_context(stage: Stage):
       log_contexts.add_metadata('stage', stage)
       yield
     except Exception as e:
-      error(message='Error during task.')
+      warning(message='Error during task.')
```
ViniciustCosta, line 798: Maybe add a comment just making clear that this is will be handled and logged as an error in the top-level and we just want to log as warning for the context

ViniciustCosta (approved):
lgtm

PR #4761, opened 2025-04-09
[Logging] Add structured logs context to minimize task

ViniciustCosta:
Adding structured logs to the minimize task using the testcase-based context. For postprocess, it was necessary to add code to retrieve the testcase early on, in order to pass it to the context manager.


javanlacerda (approved):
LGTM, thanks!

PR #4762, opened 2025-04-14
[Logging] Extend testcase-based logs context to variant task

ViniciustCosta:
Adding structured logs to the variant task extending the testcase-based context.
For postprocess step, it was necessary to add code to retrieve the testcase early on, in order to pass it to the context manager.


javanlacerda (approved):
LGTM

PR #4764, opened 2025-04-15
Have a single index for last_run in FuzzTargetJob

vitorguidi:
#4763 wrongly added the last_run field to a preexisting composite index, it should have been a new index only for last_run


jonathanmetzman (approved):
lgtm

javanlacerda (approved):

PR #4774, opened 2025-04-24
[Logging] Add fuzz-related logs context to corpus pruning task

ViniciustCosta:


src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -750,6 +754,16 @@ def get_extras(self) -> NamedTuple:
...
         return GenericLogStruct()
 
+    if self == LogContextType.FUZZ:
+      try:
+        return FuzzLogStruct(
+            fuzzer=log_contexts.meta.get('fuzzer_name', 'null'),
+            job=log_contexts.meta.get('job_type', 'null'),
+            fuzz_target=log_contexts.meta.get('fuzz_target', 'null'))
+      except:
+        error('Error retrieving context for fuzzing-based logs.')
```
javanlacerda, line 764: You need to add the `ignore_context=True`

ViniciustCosta: Good catch, I'll fix that!

src/clusterfuzz/_internal/tests/core/metrics/logs_test.py
```diff
@@ -760,18 +762,67 @@ def test_missing_fuzz_target_in_log_context(self):
...
+        extra={
+            'extras': logs_extra,
+            'location': {
+                'path': os.path.abspath(__file__).rstrip('c'),
+                'line': statement_line,
+                'method': 'test_missing_fuzz_target_in_testcase_context'
+            },
+        })
+
+  def test_fuzz_logs_context(self):
```
vitorguidi, line 793: This tests the log being emited within the log context. We do not have integration tests that go through pre/main/postprocess though, so we are not testing the actual task execution flow here.

Can we get some evidence of this running on a candidate release of some sort, to make sure nothing weird happens and we actually get the logs?

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -695,21 +695,25 @@ class TaskLogStruct(NamedTuple):
...
-     and this context is used for define the adicional labels
-     to be added to the log.
+  """Log context types.
+  
+  This is the way to define the context for a given entrypoint and this
+  context is used for define the adicional labels to be added to the log.
   """
   COMMON = 'common'
   TASK = 'task'
+  FUZZ = 'fuzz'
```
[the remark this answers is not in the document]
ViniciustCosta: My first instinct was to call it "fuzzer" as well, but my idea here is that this is a fuzz-related context (not fuzz task), since only tasks that are related to fuzzing contain these fields (fuzzer, fuzz_target and job).

I don't mind changing it to "fuzzer" though, if it is clearer in your opinion :)

src/clusterfuzz/_internal/tests/core/metrics/logs_test.py
ViniciustCosta: Yeah, we have this gap of integration tests that hinders our testing capabilities for these kind of changes.
I will do a onebox deployment with candidate and paste the logs here.

src/clusterfuzz/_internal/metrics/logs.py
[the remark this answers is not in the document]
vitorguidi: Fuzzing is about findIng crashes, pruning the corpus has to do with checking coverage. This is why I am a bit reluctant about this word.

This is a nit though, so I am completely ok with going forward with either (or a third option, but I exhausted my creativity here to come up with yet another name)

src/clusterfuzz/_internal/tests/core/metrics/logs_test.py
ViniciustCosta: I executed the task preprocess stage locally and pasted the logs in the PR description (tyvm for the help on how to do this @vitorguidi 🤝)

src/clusterfuzz/_internal/metrics/logs.py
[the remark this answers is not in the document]
ViniciustCosta: My point here was more about the meaning of a log context that has the triad "fuzzer, fuzz_target and job" than the meaning of the task itself. But, if it was already misleading, I think "fuzzer" must be a better option (I also couldn't think of a third one haha). I'll update it, thanks.

vitorguidi (approved):
Thanks for the thorough testing! lgtm

javanlacerda (approved):
Receive it!

PR #4781, opened 2025-04-30
[Logging] Create and add structured logs for Cleanup task

ViniciustCosta:
### Description
Adding structured logging to the cleanup cronjob. In order to enable this, the following changes on logs module were done:
* Added a new log context for cronjobs.
* Added a filter to the logs handler for both GAE/GKE to propagate the `extras` argument from python logging to the `json_fields` metadata in google cloud logging ([GCP reference](https://cloud.google.com/python/docs/reference/logging/latest/std-lib-integration#automatic-metadata-detection)). (Note that this is a workaround to enable the observability project for now. As a long-term solution, we should try to centralize the logs handlers/filters for all environments). 

For the cleanup module:
* Set env vars for task name and ID in the cron entrypoints (for k8s and appengine).
* Instrumented the cleanup code with the cron-based log context and with the testcase-based context. This latest is done for each testcase within the clean up unneeded open testcases method. 

### Tests
Running the cronjob locally with the debugger and sending logs to GCP.
* Setting `IS_K8S_ENV` to true ([link to logs](https://cloudlogging.app.goo.gl/Us8aHBMUyNoaPFbL8)):
image

### DIFF
--- a/src/clusterfuzz/_internal/cron/cleanup.py
+++ b/src/clusterfuzz/_internal/cron/cleanup.py
@@ -100,6 +100,60 @@ def cleanup_reports_metadata():
   ndb_utils.delete_multi(uploaded_reports)
 
 
+def _cleanup_testcases_and_issues(testcase, jobs,
+                                  top_crashes_by_project_and_platform_map,
+                                  empty_issue_tracker_policy):
+  """Clean up unneeded open testcase and its associated issues."""
+  testcase_id = testcase.key.id()
+  logs.info(f'Processing testcase {testcase_id}.')
+
+  try:
+    issue = issue_tracker_utils.get_issue_for_testcase(testcase)
+    policy = issue_tracker_utils.get_issue_tracker_policy_for_testcase(testcase)
+    if not policy:
+      logs.info('No policy')
+      policy = empty_issue_tracker_policy
+
+    # Issue updates.
+    update_os_labels(policy, testcase, issue)
+    logs.info('maybe updated os')
+    update_fuzz_blocker_label(policy, testcase, issue,
+                              top_crashes_by_project_and_platform_map)
+    logs.info('maybe updated fuzz blocker')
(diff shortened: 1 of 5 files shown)


src/clusterfuzz/_internal/cron/cleanup.py
```diff
@@ -125,55 +179,10 @@ def cleanup_testcases_and_issues():
...
-      # Mark testcase as triage complete if both testcase and associated issue
-      # are closed. This also need to be done before the deletion rules.
-      mark_testcase_as_triaged_if_needed(testcase, issue)
-
-      # Testcase deletion rules.
-      delete_unreproducible_testcase_with_no_issue(testcase)
-    except Exception:
-      logs.error(f'Failed to process testcase {testcase_id}.')
+    with logs.testcase_log_context(testcase, testcase.get_fuzz_target()):
+      _cleanup_testcases_and_issues(testcase, jobs,
```
javanlacerda, line 183: Couldn't we add the `try` `catch` here, instead of inside the `_cleanup_testcases_and_issues` function?

javanlacerda: To avoid to indent the entire block. Just like the refactor we've done for setting the log context, but in this case, we already have the block we want to indent in a funcion

ViniciustCosta: As we discussed offline, it is better to keep the function as is so that it handles the exception.

src/python/bot/startup/run_cron.py
```diff
@@ -59,6 +60,11 @@ def main():
   task = sys.argv[1]
 
   task_module_name = f'clusterfuzz._internal.cron.{task}'
+
```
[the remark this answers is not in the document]
ViniciustCosta: Done, thanks

jonathanmetzman (approved):
LGTM

javanlacerda (approved):

PR #4784, opened 2025-05-06
Use backwards compatible helper function for io decompression. 

jonathanmetzman:
This helper saves code and catches a case we missed.


oliverchang (commented):
LGTM

PR #4785, opened 2025-05-07
[Logging] Add new log context for grouper-related structured logging

ViniciustCosta:
Changes to the logs module in order to enable structured logging for the grouper task:
* Adapting the testcase-based context to also accept the `TestcaseAttributes` class (used by the grouper).
* Adding a new log context for grouping, which is an edge case where pairs of testcases are processed each step.

In addition, this PR also adds unit tests and comments that were missing in the logs module.

### DIFF
--- a/src/clusterfuzz/_internal/metrics/logs.py
+++ b/src/clusterfuzz/_internal/metrics/logs.py
@@ -31,8 +31,11 @@ from typing import Any
 from typing import NamedTuple
 from typing import TYPE_CHECKING
 
+from google.cloud import ndb
+
 # This is needed to avoid circular import
 if TYPE_CHECKING:
+  from clusterfuzz._internal.cron.grouper import TestcaseAttributes
   from clusterfuzz._internal.datastore.data_types import FuzzTarget
   from clusterfuzz._internal.datastore.data_types import Testcase
 
@@ -694,6 +697,14 @@ def get_common_log_context() -> dict[str, str]:
     return {}
 
 
+def get_testcase_id(
+    testcase: 'Testcase | TestcaseAttributes') -> int | str | None:
+  """Return the ID for a testcase or testcase attribute object."""
+  if isinstance(testcase, ndb.Model):
+    return testcase.key.id()  # type: ignore
+  return getattr(testcase, 'id', None)
+
+
 class GenericLogStruct(NamedTuple):
   pass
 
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -694,6 +695,16 @@ def get_common_log_context() -> dict[str, str]:
     return {}
 
 
+def get_testcase_id(
+    testcase: 'Testcase | TestcaseAttributes') -> int | str | None:
+  """Return the ID for a testcase or testcase attribute object."""
+  from google.cloud import ndb
```
javanlacerda, line 701: I think it can be imported on the top of the file. As it is a third party lib, it should not cause any circular import problem.

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -925,7 +950,7 @@ def fuzzer_log_context(fuzzer_name: str, job_type: str,
 
 
 @contextlib.contextmanager
-def testcase_log_context(testcase: 'Testcase',
+def testcase_log_context(testcase: 'Testcase | TestcaseAttributes',
```
javanlacerda, line 953: I'm maybe missing something, but you added this new type we can have for testcase, but this context wasn't added anywhere. Should it be?

src/clusterfuzz/_internal/metrics/logs.py
ViniciustCosta: Not actually, I think it is only accessible after building the CF dependencies, so it only works during runtime.

It broke when I tried to import on the top level (`ModuleNotFoundError: No module named 'google.cloud'`) and you can see [here](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/metrics/logs.py#L361) that this is already done in this module.

javanlacerda: I see, thanks for the clarification.

src/clusterfuzz/_internal/metrics/logs.py
ViniciustCosta: Yes, the context was not added yet.
I tried to split this PR adding the new context from the one actually instrumenting the grouper code, to facilitate the review (I probably should have added this in the description, my bad).

javanlacerda (approved):
LGTM, since it will be used for instrumeting grouper task in a separate PR. Thanks!

src/clusterfuzz/_internal/metrics/logs.py
jonathanmetzman: Can you leave a comment explaining this?

jonathanmetzman (approved):
lgtm

PR #4787, opened 2025-05-07
[Logging] Add structured log context to Triage

ViniciustCosta:
### Description
This adds structured logging to the triage job.

To enable this, the triage code was instrumented with the cron log context and with the testcase-based context, which wraps each triage execution for a given testcase. To avoid indenting the whole function, a new private method was created for triaging a single testcase.

Note: grouper will be instrumented in a next PR after https://github.com/google/clusterfuzz/pull/4781 is merged.

### Test
* Running the cronjob locally with the debugger ([link to logs](https://cloudlogging.app.goo.gl/dh6JbXqhcSDRkWDK6)).
image

### DIFF
--- a/src/clusterfuzz/_internal/cron/triage.py
+++ b/src/clusterfuzz/_internal/cron/triage.py
@@ -416,6 +416,133 @@ def _emit_untriaged_testcase_age_metric(testcase: data_types.Testcase,
       })
 
 
+def _triage_testcase(testcase, excluded_jobs, all_jobs, throttler):
+  """Files a bug for a given testcase."""
+  testcase_id = testcase.key.id()
+  critical_tasks_completed = data_handler.critical_tasks_completed(testcase)
+
+  # Skip if testcase's job is removed.
+  if testcase.job_type not in all_jobs:
+    _set_testcase_stuck_state(testcase, False)
+    logs.info(f'Skipping testcase {testcase_id}, since its job was removed '
+              f' ({testcase.job_type})')
+    return
+
+  # Skip if testcase's job is in exclusions list.
+  if testcase.job_type in excluded_jobs:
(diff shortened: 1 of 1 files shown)


javanlacerda (approved):
LGTM. just a question, is triage being called through the common task flow? I'm wondering if we should add the task context as well.

ViniciustCosta:
> LGTM. just a question, is triage being called through the common task flow? I'm wondering if we should add the task context as well.

Triage is executed as a cronjob, so I added the cron-related context (which logs the task name/ID). The entry points for the cronjobs is already instrumented setting the env vars with this information (https://github.com/google/clusterfuzz/pull/4781).

src/clusterfuzz/_internal/cron/triage.py
```diff
@@ -448,131 +574,9 @@ def main():
...
-
-    _set_testcase_stuck_state(testcase, False)
-
-    _create_filed_bug_metadata(testcase)
-    issue_filer.notify_issue_update(testcase, 'new')
-
-    logs.info('Filed new issue %s for testcase %d.' % (testcase.bug_information,
-                                                       testcase_id))
+    with logs.testcase_log_context(testcase, testcase.get_fuzz_target()):
+      logs.info(f'Triaging {testcase_id}')
```
vitorguidi, line 578: I think a good test is to run this locally with the debugger, until you hit this line of code. Then you can exit early, and not go into any _triage_testcase code at all.

LGTM, will approve once evidence is granted

ViniciustCosta: Sure, I added this test to the PR description (note: I also had to comment out the grouper call). Thanks!

vitorguidi (approved):
lgtm

PR #4790, opened 2025-05-08
Check flag is_managed_by_argo_cd for k8s deploy

javanlacerda:
it adds a guard on deploy k8s function to check if it's configured to be managed by ArgoCD. We should deploy it only if it's not, to avoid overlap the deployments.

### DIFF
--- a/src/local/butler/deploy.py
+++ b/src/local/butler/deploy.py
@@ -502,6 +502,9 @@ def _enforce_safe_day_to_deploy():
 
 def _deploy_k8s(config_dir):
   """Deploys all k8s workloads."""
+  is_managed_by_argo_cd = environment.get_value('MANAGED_BY_ARGOCD', False)
+  if is_managed_by_argo_cd:
+    return
   k8s_dir = os.path.join('infra', 'k8s')
   k8s_instance_dir = os.path.join(config_dir, 'k8s')
   k8s_project = local_config.ProjectConfig().get('env.K8S_PROJECT')
--- a/src/local/butler/deploy.py
+++ b/src/local/butler/deploy.py
@@ -502,12 +502,13 @@ def _enforce_safe_day_to_deploy():
 
 def _deploy_k8s(config_dir):
   """Deploys all k8s workloads."""
-  is_managed_by_argo_cd = environment.get_value('MANAGED_BY_ARGOCD', False)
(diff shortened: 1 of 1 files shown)


src/local/butler/deploy.py
```diff
@@ -502,6 +502,9 @@ def _enforce_safe_day_to_deploy():
 
 def _deploy_k8s(config_dir):
   """Deploys all k8s workloads."""
+  is_managed_by_argo_cd = environment.get_value('MANAGED_BY_ARGOCD', False)
```
vitorguidi, line 505: This should be an application configuration in project.yaml, not an env var

vitorguidi: https://github.com/google/clusterfuzz/blob/16ef6544575f5fd64c45ecc0fb581fc1cbeb8167/src/local/butler/deploy.py#L489

> example of how this is done

javanlacerda: Definitely!

vitorguidi (approved):
lgtm

src/local/butler/deploy.py
```diff
@@ -502,9 +502,13 @@ def _enforce_safe_day_to_deploy():
 
 def _deploy_k8s(config_dir):
   """Deploys all k8s workloads."""
+  config = local_config.ProjectConfig()
+  is_argocd_enabled = config.get('argocd.enabled', False)
+  if is_argocd_enabled:
```
ViniciustCosta, line 507: What do you think about adding this check early-on, as it is done [here](https://github.com/google/clusterfuzz/blob/master/src/local/butler/deploy.py#L589), and setting the `deploy_k8s` to false? 

So that anything that touches this deployment is not executed, for instance, the terraform step [here](https://github.com/google/clusterfuzz/blob/master/src/local/butler/deploy.py#L462). Or maybe this is expected to be executed actually?

jonathanmetzman (commented):
LGTM

src/local/butler/deploy.py
javanlacerda: I apologize, I miss your comment before merging it. But we wouldn't like to avoid to deploy terraform when the k8s is managed by argocd, then it would misbehave.

vitaliset (commented):
LGTM

PR #4792, opened 2025-05-09
[Logging] Instrument Grouper code for structured logs

ViniciustCosta:


src/clusterfuzz/_internal/cron/grouper.py
```diff
@@ -215,108 +232,120 @@ def _group_testcases_based_on_variants(testcase_map):
...
   for testcase_1_id, testcase_1 in testcase_map.items():
     for testcase_2_id, testcase_2 in testcase_map.items():
-      # Rule: Don't group the same testcase and use different combinations for
-      # comparisons.
-      if testcase_1_id <= testcase_2_id:
-        continue
+      with logs.grouper_log_context(testcase_1, testcase_2):
+        # Rule: Don't group the same testcase and use different combinations for
+        # comparisons.
+        if testcase_1_id <= testcase_2_id:
```
vitorguidi, line 269: By using the grouper context, you are logging things with a different labels.

Suppose I want to see everything that happened the last 2 days for a given testcase id 123.

I would do a logs query for:

jsonpayload.extras.testcase_id = 123


However, I do miss this out on any grouper related logs, when this new context is being used, since I would have to do another filter:

jsonpayload.extras.testcase_id = 123 OR jsonpayload.extras.testcase_1_id = 123 or jsonpayload.extras.testcase_2_id = 123

It would be important to have this homogeneous access pattern, where we only reference testcase_id. Otherwise, whoever is debugging must be aware of a lower level implementation (grouper context) to do high level troubleshooting (testcase lifecycle). This might work for us, but not for customers who download logs through the UI, when we land tracing, for instance.


Two possible approaches to solve this issue:
* Make sure we log twice in the grouper Testcase, one in the Testcase context for each Testcase id
* Manually edit the log entries in the business logic, so we do two log.info (or w/e level), each referencing a different Testcase id

I think 2 is better, since the grouper is stable code and we should not be adding many more logs after this change. Wdyt?

ViniciustCosta: You're righ, that makes a lot of sense! I figured that option 1 (duplicating the log entry with each testcase context in the logs module) would be a better and feasible approach. I just updated, will try to add some tests with the local debugger, but PTAL.

Thanks for catching this!

src/clusterfuzz/_internal/cron/grouper.py
```diff
@@ -55,6 +55,13 @@ def __init__(self, testcase_id):
     self.is_leader = True
     self.issue_id = None
 
+  def get_metadata(self, key=None, default=None):
+    """Retrieve class attributes."""
+    # This is done to simulate the same method in the Testcase class.
```
jonathanmetzman, line 60: How does it simulate it? I'm not sure I understand.
https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/datastore/data_types.py#L704

jonathanmetzman (commented):
I'll try to do a full review later today.

src/clusterfuzz/_internal/cron/grouper.py
ViniciustCosta: Maybe my wording here was misleading, sorry about that. What I meant was that I added this method to have a similar behavior as the one you linked, so that doing `testcase.get_metadata()` works independently of the object being a `Testcase` or `TestcaseAttributes` (which is done in the logs).

src/clusterfuzz/_internal/cron/grouper.py
```diff
@@ -132,6 +139,56 @@ def matches_top_crash(testcase, top_crashes_by_project_and_platform):
   return False
 
 
+def _check_variant_grouping_candidate(testcase_1, testcase_2,
```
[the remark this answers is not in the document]
ViniciustCosta: Sure, I'll add this.

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -774,6 +770,23 @@ def get_extras(self) -> NamedTuple:
         common_ctx = get_common_log_context()
         log_contexts.add_metadata('common_ctx', common_ctx)
 
+    if self == LogContextType.GROUPER:
+      if log_contexts.meta.get('in_grouper_ctx'):
+        return
+      log_contexts.add_metadata('in_grouper_ctx', True)
+      # Duplicate the emit so that we have a log entry with the context of each
+      # testcase in the pair being grouped.
+      emit(*args, **kwargs)
```
vitorguidi, line 779: I am not really a fan of doing emit in a function that does a setup.

Wdyt about having some sort of simetry on these labels? The problem we are facing is to do a double emit, only using either testcase1 (and group1), or testcase2 (and group2)

We could have these labels with a prefix _some_$LABEL and _another_$LABEL, and we move this logic of double emission into emit itself.

For instance:
* one emission for _some_testcase and _some_group_id
* another emission for _another_testcase and _another_testcase_id

I think having some sort of convention for these symetric logs is a more generic and elegant solution. Wdyt?

ViniciustCosta: At first I wanted to avoid changing anything in the logs module aside from the structured logs we are working on, but I agree this approach is better!

As we discussed offline, I tried to implement a generic way of having symmetric log entries in the same emit call. For this case, this was done by passing a specific label for the logs containing a list with extras (`label : value`) for each entry. 
I found that simpler to use and understand instead of having to add prefixes to the labels themselves and then checking each extra field and figuring out which entry it belongs to. This also facilitates having an arbitrary number of symmetric emissions in the same call.

PTAL (lint is failing because of some problems with the dev env, I will try to fix later).

ViniciustCosta: Improved the PRs description and added the tests, PTAL.

PR #4793, opened 2025-05-13
[Deployment Revisited][Staging] Update create config with missing resources

vitorguidi:
### Motivation

The butler create config script is out of date. If we bootstrap clusterfuzz in the current state, it will not be fully functional, mostly due to new utask main executions model, powered by GCP batch, and default Redis versions automatically created by GCP being incompatible with the current Redis python library.

### Changes to butler deploy

* Removed redis management from butler deploy, since there was split brain where it was managed in both terraform and butler deploy
* deploy_k8s is now performed before deploy_appengine, since appengine depends on redis, and redis bootstrap is done on terraform

### Newly added resources

- Enabled batch API
- Enabled secrets manager API
- utask queues:
  - preprocess
  - postprocess
  - utask_main
- utask buckets:
- pubsub trigger, that publishes to the postprocess topic when the uworker-output bucket is written to
- Update permissions on default project/compute service account:
  - Storage object admin
  - Secret manager secret acessor
  - Metric publisher 
- gcs-signer service account
- gcs-signer secret, containing a JSON token from the gcs-signer service account
- untrusted-worker service account
  - Overall permissions:
    - batch agent reporter
    - error reporting writer
    -  logs writer
    - monitoring metric writer 
  - Deployment bucket permissions:
    - Storage viewer
    - GCS object lister
    -  Storage Legacy Bucket Reader 

### Newly added resources to terraform

- VPC Serverless connector for redis

### Attention points

The command to create a subscription from a GCS bucket will forcefully create the postprocess topic. Howevver, butler deploy will attempt to recreate and the operation should be idempotent, so it is actually fine to have it during butler create config for simplicity.


src/clusterfuzz/_internal/base/utils.py
```diff
@@ -1031,7 +1031,9 @@ def is_oss_fuzz():
 
 def is_chromium():
   """If this is an instance of chromium fuzzing."""
-  return default_project_name() in ('chromium', 'chromium-testing')
+  return default_project_name() in ('chromium', 'chromium-testing',
```
jonathanmetzman, line 1034: Maybe we can make an env var or config that allows a project to specify it is chromium instead of hardcoding who is?

src/local/butler/common.py
```diff
@@ -257,15 +257,15 @@ def _pipfile_to_requirements(pipfile_dir, requirements_path, dev=False):
     dev_arg = '--dev'
 
   return_code, output = execute(
-      f'python -m pipenv requirements {dev_arg}',
+      f'python3.11 -m pipenv requirements {dev_arg}',
```
jonathanmetzman, line 260: Would python3 work here since we are already in the venv?

src/local/butler/create_config.py
```diff
@@ -286,22 +344,73 @@ def execute(args):
...
+                         'Service account for GCS signer.')
+  add_service_account_role(gcloud, args.project_id, signer_service_account,
+                           'roles/storage.objectAdmin')
+  # Create gcs-signer-secret from the signer service account
+  json_key = get_json_key_from_service_account(gcloud, signer_service_account)
+  create_secret(gcloud, 'gcs-signer-secret', json_key)
   # Create buckets now that domain is verified.
   create_buckets(args.project_id, [bucket for _, bucket in bucket_replacements])
-
+  #Creates the required pubsub triggers from buckets
```
jonathanmetzman, line 397: nit: space

jonathanmetzman (approved):
LGTM 
with nits

src/local/butler/common.py
vitorguidi: Hmm some of these python things are actually incorrect, for instance:

```
 File "/usr/local/lib/python3.11/subprocess.py", line 413, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['python3', 'butler.py', 'deploy', '--force', '--targets', 'appengine', '--prod', '--config-dir', '../clusterfuzz-config/configs/chrome-development']' returned non-zero exit status 1.
```

I will run the actual bootstrap against a dev environment and see where it breaks, then come back and fix on the PR

src/local/butler/deploy.py
```diff
@@ -445,6 +422,11 @@ def _prod_deployment_helper(config_dir,
   }
 
   try:
+    # Appengine depends on Redis, which is managed by Terraform
+    # Therefore, we need to deploy Terraform first
+    if deploy_k8s:
+      _deploy_terraform(config_dir)
```
vitorguidi, line 428: We need to do deploy terraform on the first time deploy_appengine is invoked here. Otherwise, it will not find the redis cluster and bootstrapping will fail

javanlacerda (approved):
LGTM!

javanlacerda (approved):
A tip: You can run the yapf command with `-i` that autofix the lint (without needing to run the whole format script). You will need to remove the `-d` that means `diff`. it should look like `yapf -p -i <file1> <file2>`

PR #4795, opened 2025-05-19
[Deployments] Remove K8s resources

javanlacerda:
It remove the logic to deploy k8s and all k8s manifests from infra folder.

We're able to do that once we moved all resources to be managed by ArgoCD through Helm.

Also, we concentrated all Kubernetes manifests into the internal repository.


vitorguidi:
> It remove the logic to deploy k8s and all k8s manifests from infra folder.
> 
> We're able to do that once we moved all resources to be managed by ArgoCD through Helm.
> 
> Also, we concentrated all Kubernetes manifests into the internal repository.

I do not think we should delete the K8S files. You can move them to config/test/k8s instead, so external users can bootstrap clusterfuzz, and we also maintain these as documentation.

javanlacerda:
> > It remove the logic to deploy k8s and all k8s manifests from infra folder.
> > We're able to do that once we moved all resources to be managed by ArgoCD through Helm.
> > Also, we concentrated all Kubernetes manifests into the internal repository.
> 
> I do not think we should delete the K8S files. You can move them to config/test/k8s instead, so external users can bootstrap clusterfuzz, and we also maintain these as documentation.

It makes sense! I'll do that!

vitorguidi (approved):
lgtm

PR #4796, opened 2025-05-20
Pick region scheduling based on availability

jonathanmetzman:


jonathanmetzman:
This never landed because our k8s instances don't connect to redis

jonathanmetzman:
We should still try to do this.

PR #4797, opened 2025-05-20
Merge OSS-Fuzz branch into master

jonathanmetzman:
This merges the OSS-Fuzz branch into master.
The branches diverged when support for fuzzing on batch was added to OSS-Fuzz.
Among other things this:
Added scheduling of fuzz tasks and made a lot of optimizations to it, in order to deal with challenges filling up our capacity with fuzzing.
Optimized things like corpus signing, downloading and pruning as previously these had not been run at OSS-Fuzz scale and levels of diversity.


jonathanmetzman:
Most of the major commits here have been reviewed already. This code has already been deployed and at 2000+ lines is not really going to go through another real review.

PR #4804, opened 2025-05-27
Merg3

jonathanmetzman:


vitorguidi:
> In OSS-Fuzz, assume the queue for every task to be initially scheduled in is the preprocess task. Make a config for this so we won't do this in Chrome or google3.
> 
> Fixes the recent reproducible crashes breakage in oss-fuzz.

What queue were the tasks going to during the incident?
Where should they be going instead?
Why does this PR fix the above issue?

Not immediately obvious by just reading the code

jonathanmetzman:
Proof this is fixed: https://oss-fuzz.com/testcase-detail/4910577766170624
(this PR has been deployed to oss-fuzz)

jonathanmetzman:
> I'm getting a linter error that I think is a false positive, could someone sanity check my claim? I'm disabling the check for now.
> 
> ```
> src/clusterfuzz/_internal/base/tasks/__init__.py:719:53: W0613: Unused argument 'utask_main' (unused-argument)
> ```

I was looking at an old version of the code, the error is real.

PR #4807, opened 2025-05-28
[Tracing] Implement events module

ViniciustCosta:
Implementing the module responsible for handling clusterfuzz events (mainly testcase lifecycle events).
This module defines the event data classes and enables emitting these events throughout the codebase to be persisted in a generic repository. The default repository is the datastore and testcase lifecycle events are stored under the `TestcaseLifecycleEvent` entity.

In summary, this PR:
* Creates the generic data model to describe ClusterFuzz events.
* Extends this as base classes for testcase lifecycle events (implementing one specific event as an example).
* Creates an abstract repository interface for handling and persisting events.
* Implements the repository methods for Datastore.
* Creates the functions to emit the events, i.e., serialize and store them into the underlying repository.
* Adds unit tests for this module.

### DIFF
--- a/src/clusterfuzz/_internal/metrics/events.py
+++ b/src/clusterfuzz/_internal/metrics/events.py
@@ -187,7 +187,7 @@ class NDBEventRepository(IEventRepository, EventHandler):
       logs.error('Error deserializing Datastore entity to event.')
     return None
 
-  def store_event(self, event: Event) -> str | int | None:
+  def store_event(self, event: Event) -> int | None:
     """Stores a Datastore entity and returns its ID."""
     entity = self._serialize_event(event)
     if entity is None:
@@ -245,6 +245,8 @@ def config_handlers() -> None:
   ]
 
   for handler in event_handlers:
+    if handler is None:
+      continue
     if not isinstance(handler, EventHandler):
       logs.error(
           f'Event handler of type {type(handler)} should subclass EventHandler.'
--- a/src/clusterfuzz/_internal/metrics/events.py
+++ b/src/clusterfuzz/_internal/metrics/events.py
@@ -20,6 +20,7 @@ from dataclasses import dataclass
 from dataclasses import field
 from dataclasses import InitVar
 import datetime
+from enum import Enum
 from typing import Any
 
 from clusterfuzz._internal.config import local_config
@@ -27,6 +28,11 @@ from clusterfuzz._internal.datastore import data_handler
 from clusterfuzz._internal.datastore import data_types
 from clusterfuzz._internal.metrics import logs
 from clusterfuzz._internal.system import environment
+
+
+class EventTypes(Enum):
+  """Specific event types."""
+  TESTCASE_CREATION = 'testcase_creation'
 
 
 @dataclass(kw_only=True)
@@ -55,16 +61,16 @@ class Event:
 
 
 @dataclass(kw_only=True)
-class TestcaseEvent(Event):
+class BaseTestcaseEvent(Event):
   """Base class for testcase-related events."""
   # Testcase entity (only used in init to set the event data).
   testcase: InitVar[data_types.Testcase | None] = None
 
   # Testcase metadata (retrieved from the testcase entity, if available).
-  testcase_id: int | None = None
-  fuzzer: str | None = None
-  job: str | None = None
-  crash_revision: int | None = None
+  testcase_id: int | None = field(init=False, default=None)
+  fuzzer: str | None = field(init=False, default=None)
+  job: str | None = field(init=False, default=None)
+  crash_revision: int | None = field(init=False, default=None)
 
   def __post_init__(self, testcase=None, **kwargs):
     if testcase is not None:
@@ -76,7 +82,7 @@ class TestcaseEvent(Event):
 
 
 @dataclass(kw_only=True)
-class TaskEvent(Event):
+class BaseTaskEvent(Event):
   """Base class for task-related events."""
   # Task ID retrieved from environment var (if not directly set).
   task_id: str | None = None
@@ -88,18 +94,19 @@ class TaskEvent(Event):
 
 
 @dataclass(kw_only=True)
-class TestcaseCreationEvent(TestcaseEvent, TaskEvent):
+class TestcaseCreationEvent(BaseTestcaseEvent, BaseTaskEvent):
   """Testcase creation event."""
-  event_type: str = field(default='testcase_creation', init=False)
+  event_type: str = field(
+      default=EventTypes.TESTCASE_CREATION.value, init=False)
   # Either manual upload, fuzz task or corpus pruning.
   origin: str | None = None
   # User email, if testcase manually uploaded.
   uploader: str | None = None
 
 
-# Mapping of specific event types to their classes.
+# Mapping of specific event types to their data classes.
 _EVENT_TYPE_CLASSES = {
-    'testcase_creation': TestcaseCreationEvent,
+    EventTypes.TESTCASE_CREATION.value: TestcaseCreationEvent,
 }
 
 
--- a/src/clusterfuzz/_internal/metrics/events.py
+++ b/src/clusterfuzz/_internal/metrics/events.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Events definition and handling."""
+"""Events handling and emitting."""
 
 from abc import ABC
 from abc import abstractmethod
@@ -95,18 +95,9 @@ class TestcaseCreationEvent(TestcaseEvent, TaskEvent):
   uploader: str | None = None
 
 
-@dataclass(kw_only=True)
-class TestcaseRejectionEvent(TestcaseEvent, TaskEvent):
-  """Testcase rejection event."""
-  event_type: str = field(default='testcase_rejection', init=False)
-  # Reason for rejection (e.g., triage_duplicated)
-  reason: str | None = None
-
-
 # Mapping of specific event types to their classes.
 _EVENT_TYPE_CLASSES = {
     'testcase_creation': TestcaseCreationEvent,
-    'testcase_rejection': TestcaseRejectionEvent
 }
 
 
--- a/src/clusterfuzz/_internal/metrics/events.py
+++ b/src/clusterfuzz/_internal/metrics/events.py
@@ -101,3 +101,129 @@ class TestcaseRejectionEvent(TestcaseEvent, TaskEvent):
   event_type: str = field(default='testcase_rejection', init=False)
   # Reason for rejection (e.g., triage_duplicated)
   reason: str | None = None
+
+
+# Mapping of specific event types to their classes.
+_EVENT_TYPE_CLASSES = {
+    'testcase_creation': TestcaseCreationEvent,
+    'testcase_rejection': TestcaseRejectionEvent
+}
+
+
+class IEventRepository(ABC):
+  """Event repository abstract class (interface).
+
+  This class is responsable for defining the expected operations for event
+  storage and retrieval in a repository/database.
+  All concrete event repositories must implement these methods.
+  """
+
+  @abstractmethod
+  def serialize_event(self, event: Event) -> Any:
+    """Serialize an event into the underlying database entity."""
+
+  @abstractmethod
+  def deserialize_event(self, entity: Any) -> Event | None:
+    """Deserialize a database entity into an event."""
+
+  @abstractmethod
+  def store_event(self, entity: Any) -> str | int | None:
+    """Save an event entity into the underlying database and return its ID."""
+
+
+class NDBEventRepository(IEventRepository):
+  """Implements the event repository for Datastore.
+  
+  Handles conversion between Event objects and Datastore entities. If a new
+  Datastore entity model is needed, it relies on mapping the event types to
+  the correct entity.
+  """
+  # Maps `event_type` to a Datastore model.
+  # For now, only testcase lifecycle events are being traced.
+  _event_to_entity_map = {}
+  _default_entity = data_types.TestcaseLifecycleEvent
+
+  def _to_entity(self, event: Event) -> data_types.Model:
+    """Converts the event object into the Datastore entity."""
+    entity_model = self._event_to_entity_map.get(event.event_type,
+                                                 self._default_entity)
+
+    event_entity = entity_model(event_type=event.event_type)
+    for key, val in asdict(event).items():
+      setattr(event_entity, key, val)
+    return event_entity
+
+  def _from_entity(self, entity: data_types.Model) -> Event:
+    """Converts a Datastore entity into an event object."""
+    if not hasattr(entity, 'event_type'):
+      raise TypeError('Datastore model should contain an event_type.')
+
+    event_type = entity.event_type  # type: ignore
+    event_class = _EVENT_TYPE_CLASSES.get(event_type, None)
+    if event_class is None:
+      event = Event(event_type=event_type)
+    else:
+      event = event_class()
+    for key, val in entity.to_dict().items():
+      setattr(event, key, val)
+    return event
+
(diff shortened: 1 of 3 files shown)


ViniciustCosta:
I still need to add more unit tests and maybe fix minor stuff, but I think it's ready for a review. PTAL

src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -0,0 +1,221 @@
...
+class IEventRepository(ABC):
+  """Event repository abstract class (interface).
+
+  This class is responsable for defining the expected operations for event
+  storage and retrieval in a repository/database.
+  All concrete event repositories must implement these methods.
+  """
+
+  @abstractmethod
+  def serialize_event(self, event: Event) -> Any:
```
vitorguidi, line 113: Why does this belong in the interface? 

A repository allows me to read/create/update/delete Events (the Event classes you described). Why should I care about the internal representation of the datastore? 

Also, when we implement the frontend, I want to query events with a list of filters. For instance: give me all tasks events that belong to testcase_id = 123


IMO the serialization stuff should go away, and the read operations should be implemented.

src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+def get_repository() -> IEventRepository | None:
+  """Return the repository interface to handle and store events."""
+  if not _repository:
+    config_repository()
+  return _repository
+
+
+def emit(event: Event) -> str | int | None:
```
vitorguidi, line 210: We have two use cases which will arise from events so far:
1) Persisting them to the database
2) Sending buganizer notifications, if applicable

This function explicitly calls the persistence part. A better approach is to have some sort of dependency injection, so we can register **arbitrary handlers**.

One handler deals with persistence, the other deals with event emission.

Something along the lines of:

```
handlers = config.parseEventHandlers();
def emit(e: Event):
     for handler in handlers:
         handler.handle(e)
 
```
We can make the code more receptive to this future use cases, since we know ahead of time that we will need at least two different handlers.
Wdyt?

src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -0,0 +1,221 @@
...
+  @abstractmethod
+  def deserialize_event(self, entity: Any) -> Event | None:
+    """Deserialize a database entity into an event."""
+
+  @abstractmethod
+  def store_event(self, entity: Any) -> str | int | None:
+    """Save an event entity into the underlying database and return its ID."""
+
+
+class NDBEventRepository(IEventRepository):
```
vitorguidi, line 125: Where will the configuration for this live in?

I imagine in project.yaml. Something like:

```
events:
    storage_backend: ndb
    issue_tracker: buganizer
```

So an unit test for correct creation of this would be welcome

src/clusterfuzz/_internal/metrics/events.py
ViniciustCosta: IMO the SerDe may be useful for future functions, and it makes sense to have these since every repository should have a way of converting these events into the underlying representation (and vice-versa). Also, I don't think this interface is really supposed to be used outside of the events module. Other module should only really care for the top-level methods (in this case, the emit and further read/query methods).
 
Regarding update/delete operations, they may not be very applicable to events, which are supposed to be immutable (but it should be easy to add, if needed).

The read method I agree! I missed that and will add in a new commit, thanks!

ViniciustCosta: I will also update the "create" method (`store_events`) to receive the Event dataclass instead of the entity, this was a bad choice as well.

src/clusterfuzz/_internal/metrics/events.py
ViniciustCosta: Yes, exactly. But my initial idea was to use the datastore as the default and only add new config if we use a distinct storage. Do you think that it's better to add this config already setting the `ndb`?

Also, how would I add unit tests for this? Do you mean like mocking the `local_config.Project().get()` and checking if the correct repository is chosen (only one case since we only have datastore for now)?

The issue tracker I'm not sure why it is needed, I may be missing something, but it should be retrieved from the testcase itself, possibly using the [issue tracker utils]( https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/issue_management/issue_tracker_utils.py#L86). Maybe we need to only enable/disable it in the config?

src/clusterfuzz/_internal/metrics/events.py
vitorguidi: The reason why I digress on the serde methods belonging in the interface is that, as you said youself, it will not be used outside of the module.

An interface is about what users will see. Nobody using the events module, from the business logic, should be referencing low level database details. All we care about when manipulating this module is the 'class Event'. 

We NEVER should reference any lower level construct from the business logic, otherwise we are making the application functionality vulnerable to underlying library/infrastructure changes. 

The whole point of these abstractions is that, **if we ever submit a code change to these lower level constructs (NDBHandler, etc), we do not have to change any code in the application business logic**. Exposing the serde in the interface goes against this principle.

It would make sense to make it a private method in an abstract class, but since you are implementing an interface, only public methods should be exposed.

src/clusterfuzz/_internal/metrics/events.py
vitorguidi: I think it is a good call to add NDB in the config, so we make it very explicit that this module was design to have a flexible storage layer. And yes, the unit test is precisely what you described: mocking local_config.get.

Re the issue tracker, you are correct. Whenever we implement the issue tracking handle, we can derive that from the testcase at runtime. Nvmd that.

ViniciustCosta: Cool, I'll add the unit test and the config in clusterfuzz-config.

src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -0,0 +1,239 @@
...
+  @abstractmethod
+  def deserialize_event(self, entity: Any) -> Event | None:
+    """Deserialize a database entity into an event."""
+
+  @abstractmethod
+  def store_event(self, event: Event) -> str | int | None:
+    """Save an event into the underlying database and return its ID."""
+
+  @abstractmethod
+  def get_event(self, event_id: str | int,
```
vitorguidi, line 139: How do we get a set of events, given a constraint?

For instance:
* All events for testcase id X
* All events of type Y

We also need a method to retrieve several testcases. The set of filters should accomodate the future use case of the front end.

ViniciustCosta: I think these should be defined in future PRs. This one should focus on defining the baseline implementation for emitting these events and persisting them, for now  (also, as described by the definition of done of the associated task)

src/clusterfuzz/_internal/metrics/events.py
ViniciustCosta: I think you are overloading the concept of users in this case. As I said, this interface is not suppose to be used outside of the module itself, its responsibility is defining a generic way for any functions in this module to communicate with the storage backend (but python is not very friendly with oop concepts like public and private interfaces). Users from the business logic side, i.e., any other module that wants to emit or query for events are going to use the top-level functions (just like it's done for the logs module, where the logger itself is not exposed). Thus, nobody using the events module will reference any low level database detail independently of the serde available here (other modules will not do something like `events._repository.method()`).

I don't think defining these functions to be implemented by all repositories (which need to be able to convert between representations) would make it vulnerable to changes in the underlying infrastructure. Code in the application business logic would not have to be changed by any changes in these lower level handlers.

Anyhow, I'll remove them for the sake of this review, since they do not seem useful in the current use case.

vitorguidi: Hmm, now I get it.

Any business logic code (fuzz task, the front end) will do:

```
import src.clusterfuzz.internal.events (or w/e)
events.emit(x)
```

and the repository class is hidden inside the module, so no one will be able to access the serde stuff. LGTM, the actual problem I was worried about does not exist.

src/clusterfuzz/_internal/metrics/events.py
vitorguidi: Fine by me

src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -0,0 +1,276 @@
...
+        event = event_class()
+      for key, val in entity.to_dict().items():
+        if hasattr(event, key):
+          setattr(event, key, val)
+      return event
+    except:
+      logs.error('Error deserializing Datastore entity to event.')
+    return None
+
+  def store_event(self, event: Event) -> str | int | None:
```
vitorguidi: why str | int? If we have only one entity that represents the event in the database, in what situation would mismatching types be returned?

src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -0,0 +1,276 @@
...
+  """Return the repository used to handle and persist events."""
+  # Any other repository implementations should define here the project config
+  # event storage name.
+  storage_cfg = local_config.ProjectConfig().get('events.storage')
+  repository = None
+  if storage_cfg == 'datastore':
+    repository = NDBEventRepository()
+  else:
+    logs.warning('Events storage in project config not found/available.')
+  return repository
```
vitorguidi, line 233: If you return None here, and pass an event to this null object, you will get a runtime exception

src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -0,0 +1,276 @@
...
+  # Handlers config should be added here.
+  repository = get_repository()
+  event_handlers = [
+      repository,
+  ]
+
+  for handler in event_handlers:
+    if not isinstance(handler, EventHandler):
+      logs.error(
+          f'Event handler of type {type(handler)} should subclass EventHandler.'
```
vitorguidi: I think you should be more severe here and do an assertion, if we pass an object of some other type, something really weird is happening.

src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -0,0 +1,276 @@
...
+def emit(event: Event) -> None:
+  """Emit an event using the configured handlers.
+
+  Actions taken with emit depend on the project configuration. Mainly, it
+  should persist the event in a storage and send notifications in the issue
+  tracker, if available.
+  """
+  handlers = get_handlers()
+  if handlers is None:
+    logs.error('Failed setting events module handlers.')
```
vitorguidi: The message warns about a failure on building the handlers, you can be more precise and say that no handlers are registered and the event will not be processed.

src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -0,0 +1,276 @@
...
+    try:
+      entity_model = self._event_to_entity_map.get(event.event_type,
+                                                   self._default_entity)
+      event_entity = entity_model(event_type=event.event_type)
+      for key, val in asdict(event).items():
+        setattr(event_entity, key, val)
+      return event_entity
+    except:
+      logs.error(
+          f'Error serializing event of type {event.event_type} to Datastore.')
```
vitorguidi: Why not logging the whole event? It makes troubleshooting more actionable:

```
          f'Error serializing eventto datastore: {event}')
```

src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -0,0 +1,276 @@
...
+      if event_class is None:
+        event = Event(event_type=event_type)
+      else:
+        event = event_class()
+      for key, val in entity.to_dict().items():
+        if hasattr(event, key):
+          setattr(event, key, val)
+      return event
+    except:
+      logs.error('Error deserializing Datastore entity to event.')
```
vitorguidi: Same, consider logging the whole entity:
```
      logs.error('Error deserializing Datastore entity to event: {entity}')
```

src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -0,0 +1,276 @@
...
+  def store_event(self, event: Event) -> str | int | None:
+    """Stores a Datastore entity and returns its ID."""
+    entity = self._serialize_event(event)
+    if entity is None:
+      return None
+    try:
+      entity.put()
+      return entity.key.id()
+    except:
+      logs.error('Error storing Datastore event entity.')
```
vitorguidi: Same as previous comment

src/clusterfuzz/_internal/tests/core/metrics/events_test.py
```diff
@@ -0,0 +1,331 @@
...
+    event_tc_creation = events.TestcaseCreationEvent(
+        source=source, testcase=testcase, origin='fuzz_task')
+    event_type = event_tc_creation.event_type
+    eid = self.repository.store_event(event_tc_creation)
+    self.assertIsNotNone(eid)
+
+    event_entity = data_handler.get_entity_by_type_and_id(
+        data_types.TestcaseLifecycleEvent, eid)
+    self.assertIsNotNone(event_entity)
+    self.assertEqual(event_entity.event_type, event_type)
```
vitorguidi, line 284: This tests if an event got persisted, but there is a missing correctness check here: was it persisted, _with the correct contents?_

Might make sense to also assert on the data fields from the TestcaseCreationEvent class, and the parent classes too.

src/clusterfuzz/_internal/tests/core/metrics/events_test.py
```diff
@@ -0,0 +1,331 @@
...
+    event_entity.timestamp = date_now
+    event_entity.source = 'events_test'
+    self._set_common_event_fields(event_entity)
+    event_entity.put()
+    event_id = event_entity.key.id()
+
+    event = self.repository.get_event(event_id)
+    self.assertIsNotNone(event)
+    self.assertEqual(event.event_type, 'generic_event_test')
+    self.assertIsInstance(event, events.Event)
```
vitorguidi, line 308: This tests if _some event gets fetched, given an event id_, but does not assert correctness on the contents. Same suggestion as previous comment.

src/clusterfuzz/_internal/metrics/events.py
ViniciustCosta: Tbh I was not really sure if `key.id()` always returned an int. From reading the [docs](https://googleapis.dev/python/python-ndb/latest/key.html), ID could be an int or str, but since we are not setting it in the application for now, datastore uses an unique integer.  I'll change to only integer.

src/clusterfuzz/_internal/metrics/events.py
ViniciustCosta: This should be treated by the method calling this (which is done in `config_handlers`). Also, since it is a project config, it could be set to not persist events.

src/clusterfuzz/_internal/metrics/events.py
ViniciustCosta: Sure, I'll add that.

src/clusterfuzz/_internal/metrics/events.py
ViniciustCosta: I was reluctant on logging the whole event, since it may have a lot of fields and log messages could get large. But if that's not an issue, I'm ok with adding it.

src/clusterfuzz/_internal/tests/core/metrics/events_test.py
ViniciustCosta: The idea was that the serde functions were already being tested to ensure correctness of the content when converting, so just testing that the entity was uploaded to datastore was enough. But I can add the asserts for the data fields as well!

src/clusterfuzz/_internal/tests/core/metrics/events_test.py
ViniciustCosta: Ditto

src/clusterfuzz/_internal/metrics/events.py
ViniciustCosta: Probably someone adding a new handler without extending from the right class. I will raise an exception instead of continuing the loop.

vitorguidi (approved):

PR #4808, opened 2025-05-29
Adding the exporting features for the job-syncer cronjob

vitorguidi:
### Motivation

In order to have testing environments (dev/staging), we need some mechanism to automatically push workload definitions for the fuzzers/fuzz targets/etc.

This will be accomplished with the jobs_syncer cronjob, that allows importing from prod, and exporting to a test enviroment, the following entities:

* Fuzzer
* Job
* JobTemplate
* DataBundle

These are sufficient, along with their blobs, to kickstart fuzzing in a Clusterfuzz deployment

### Implementation

This PR will implement the Exporting functionality, in order to dump entity definitions and their blobs from production to a GCS bucket, from which these definitions can be exported to the testing environments.

The export will be done to a target bucket, which the cronjob will take through the EXPORT_BUCKET env var. Within this bucket, the following layout will be provided:

```
fuzzer/
    some_fuzzer/
        entity.proto
        blobstore_key
        sample_testcase
...
job/
    some_job/
        entity.proto
        custom_binary_key
...
databundle/
    entity.proto
    contents/
        blob1
        blob2
...
jobtemplate/
     entity.proto
```

### Testing strategy

Unit tests to verify serialization, deserialization, blob and entity copying.

### Refactors

The entity serialization/deserialization logic was moved from the uworker_io module to the data_types module, so it can be reused without imposing a dependency from the job exporter onto the uworker module (unrelated domains)

### Non goals

Importing entities will be adressed in a followup PR

### DIFF
--- a/src/clusterfuzz/_internal/cron/job_exporter.py
+++ b/src/clusterfuzz/_internal/cron/job_exporter.py
@@ -19,6 +19,7 @@ import os
 from google.cloud import ndb
 from google.protobuf import any_pb2
 
+from clusterfuzz._internal.bot.tasks.utasks import uworker_io
 from clusterfuzz._internal.datastore import data_types
 from clusterfuzz._internal.google_cloud_utils import blobs
 from clusterfuzz._internal.google_cloud_utils import gsutil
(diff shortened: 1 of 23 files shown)


src/clusterfuzz/_internal/cron/job_exporter.py
```diff
@@ -0,0 +1,160 @@
...
+    self._export_bucket = export_bucket
+
+  def _serialize(self, entity) -> bytes:
+    return uworker_io.entity_to_protobuf(entity).SerializeToString()
+
+  def _deserialize(self, proto_as_str: bytes) -> ndb.Model:
+    deserialized_any = any_pb2.Any()  # pylint: disable=no-member
+    # Parse the bytes into the Any message
+    deserialized_any.ParseFromString(proto_as_str)
+    return uworker_io.entity_from_protobuf(deserialized_any, self._target_cls)
```
jonathanmetzman, line 89: Sort of a weird dependency. I'm fine with it. But maybe it would be better to put this function in data_handler (also a bad module).

src/clusterfuzz/_internal/cron/job_exporter.py
```diff
@@ -0,0 +1,160 @@
...
+  def export_entities(self):
+    for entity in self._target_cls.query():
+      self._export_entity(entity)
+
+  def import_entities(self):
+    pass
+
+
+def main():
+  "Exports datastore entities and respective blobs"
```
jonathanmetzman: """

jonathanmetzman (approved):
lgtm

PR #4815, opened 2025-06-06
Update process_command_impl to add task to intented queue

svasudevprasad:
process_command_impl updated to accept the queue from which the task was obtained. This is then used to add the task back to  the queue incase it cannot be executed on the current hardware


jonathanmetzman:
Hi Vitor or Matheus, could you please take a look? I don't have the cycles to do this and a deploy this week.

PR #4818, opened 2025-06-10
[Tracing] Creating the `TestcaseRejectionEvent`

vitaliset:
This PR introduces the `TestcaseRejectionEvent` with associated tests and updates the `TestcaseLifecycleEvent` datastore data type.

I will later create a PR to introduce the emits in the desired steps of the life cycle.

b/394056013

### DIFF
--- a/src/clusterfuzz/_internal/tests/core/metrics/events_test.py
+++ b/src/clusterfuzz/_internal/tests/core/metrics/events_test.py
@@ -439,24 +439,3 @@ class EmitEventTest(unittest.TestCase):
     self.assertEqual(event_entity.fuzzer, 'fuzzer1')
     self.assertEqual(event_entity.job, 'test_content_shell_drt')
     self.assertEqual(event_entity.crash_revision, 1)
-
-  def test_emit_datastore_rejection_event(self):
-    """Test emit rejection event with datastore repository."""
-    self.project_config['events.storage'] = 'datastore'
-    os.environ['CF_TASK_ID'] = 'f61826c3-ca9a-4b97-9c1e-9e6f4e4f8868'
-
-    testcase = test_utils.create_generic_testcase()
-    events.emit(
-        events.TestcaseRejectionEvent(
-            source='events_test',
-            testcase=testcase,
-            rejection_reason='analyze_no_repro'))
-
-    # Assert that the event was stored in datastore.
-    all_events = data_types.TestcaseLifecycleEvent.query().fetch()
-    self.assertEqual(len(all_events), 1)
-    event_entity = all_events[0]
-
-    self.assertEqual(event_entity.event_type,
-                     events.EventTypes.TESTCASE_REJECTION.value)
-    self.assertEqual(event_entity.rejection_reason, 'analyze_no_repro')


src/clusterfuzz/_internal/datastore/data_types.py
```diff
@@ -1659,6 +1659,10 @@ class TestcaseLifecycleEvent(Model):
   # If testcase is manually uploaded, the user email.
   uploader = ndb.StringProperty()
 
+  ### Testcase Rejection
+  # Explanation for the testcase rejection.
+  reason = ndb.StringProperty()
```
ViniciustCosta: I would suggest using something like `rejection_reason` to be more explicit, since we have all fields in the same entity (remember to also update the field name in the dataclass).

src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -104,9 +105,18 @@ class TestcaseCreationEvent(BaseTestcaseEvent, BaseTaskEvent):
   uploader: str | None = None
 
 
+@dataclass(kw_only=True)
+class TestcaseRejectionEvent(BaseTestcaseEvent, BaseTaskEvent):
+  """Testcase rejection event."""
+  event_type: str = field(
+      default=EventTypes.TESTCASE_REJECTION.value, init=False)
+  reason: str | None = None
```
ViniciustCosta: Maybe also add a comment here describing the field.

src/clusterfuzz/_internal/tests/core/metrics/events_test.py
```diff
@@ -121,6 +121,21 @@ def test_testcase_creation_events(self):
...
+  def test_testcase_rejection_event(self):
+    """Test testcase rejection event class."""
+    event_type = events.EventTypes.TESTCASE_REJECTION.value
+    source = 'events_test'
+    testcase = test_utils.create_generic_testcase()
+
+    event_rejection = events.TestcaseRejectionEvent(
+        source=source,
+        testcase=testcase,
+        reason='triage spotted as a duplicate testcase')
```
ViniciustCosta: Nit: maybe use one of the reasons defined in the design doc (or use underscore instead of spaces).

src/clusterfuzz/_internal/tests/core/metrics/events_test.py
```diff
@@ -216,6 +231,18 @@ def test_serialize_testcase_specific_event(self):
...
+    """Test serializing a testcase rejection event."""
+    testcase = test_utils.create_generic_testcase()
+    event = events.TestcaseRejectionEvent(
+        source='events_test',
+        testcase=testcase,
+        reason='analyze not reproducing')
+
+    entity = self.repository._serialize_event(event)  # pylint: disable=protected-access
+    self.assertIsNotNone(entity)
+    self.assertEqual(entity.reason, 'analyze not reproducing')
```
ViniciustCosta: Try to add the assertions for the correct datastore model and for event content, similar to `test_serialize_testcase_specific_event`.

src/clusterfuzz/_internal/tests/core/metrics/events_test.py
```diff
@@ -266,6 +293,28 @@ def test_deserialize_testcase_event(self):
...
+    event_entity.fuzzer = 'fuzzer1'
+    event_entity.job = 'test_job'
+    event_entity.crash_revision = 2
+    event_entity.reason = 'analyze_flake_on_first_attempt'
+    event_entity.put()
+
+    event = self.repository._deserialize_event(event_entity)  # pylint: disable=protected-access
+    self.assertIsNotNone(event)
+    self.assertIsInstance(event, events.TestcaseRejectionEvent)
+    self.assertEqual(event.reason, 'analyze_flake_on_first_attempt')
```
ViniciustCosta: Also add the asserts for event entity content.

src/clusterfuzz/_internal/tests/core/metrics/events_test.py
```diff
@@ -362,3 +411,25 @@ def test_emit_datastore_event(self):
     self.assertEqual(event_entity.fuzzer, 'fuzzer1')
     self.assertEqual(event_entity.job, 'test_content_shell_drt')
     self.assertEqual(event_entity.crash_revision, 1)
+
+  def test_emit_datastore_rejection_event(self):
```
ViniciustCosta: I wonder if we need to test emit for every event type, since emit will simply call the same methods in the event handler that we already have tests for. IMO we don't need it, but I'm ok with having it.

vitaliset: I'm okay removing it. :)

PR #4819, opened 2025-06-10
[Tracing] Creating the `IssueFilingEvent`

vitaliset:
This PR introduces the `IssueFilingEvent` with associated tests and updates the `TestcaseLifecycleEvent` datastore data type.

I will later create other PRs to introduce the emits in the desired steps of the life cycle.

b/394059579


src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -53,6 +54,10 @@ class RejectionReason:
   ANALYZE_FLAKE_ON_FIRST_ATTEMPT = 'analyze_flake_on_first_attempt'
 
 
+class IssueTracker:
```
ViniciustCosta: I think is better to retrieve the issue tracker name from the testcase or project itself, no need to define constant values here (see https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/cron/triage.py#L509)

src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -130,10 +131,23 @@ class TestcaseRejectionEvent(BaseTestcaseEvent, BaseTaskEvent):
   rejection_reason: str | None = None
 
 
+@dataclass(kw_only=True)
+class IssueFilingEvent(BaseTestcaseEvent, BaseTaskEvent):
+  """Testcase creation event."""
```
ViniciustCosta: Please update the docstring for this event

src/clusterfuzz/_internal/tests/core/metrics/events_test.py
```diff
@@ -138,6 +138,24 @@ def test_testcase_rejection_event(self):
     self.assertEqual(event_rejection.rejection_reason,
                      events.RejectionReason.ANALYZE_NO_REPRO)
 
+  def test_issue_filing_event(self):
+    '''Test issue filing event class.'''
```
ViniciustCosta: Double quotes (Also true for the other methods added in this file)

PR #4821, opened 2025-06-11
[Tracing] Emiting `TestcaseRejectionEvent` during analyze task

vitaliset:
This PR introduces the emits of `TestcaseRejectionEvent` in the desired steps of the analyze task life cycle as stated in the "Testcase rejection" section of "Lightweight Clusterfuzz Tracing" doc.

We are also adding tests to the flow of the `handle_noncrash` function of analyze, including the verification of the emit.

The other cases where `TestcaseRejectionEvent` must be emitted (namely fuzz task, triage and corpus prunning) will be addressed in separate PRs.

b/394056013


src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
```diff
@@ -256,11 +257,22 @@ def handle_noncrash(output):
...
     return
   testcase_upload_metadata = testcase_utils.get_testcase_upload_metadata(
       output.uworker_input.testcase_id)
+  events.emit(
+      events.TestcaseRejectionEvent(
+          source='handle_noncrash',
+          testcase=testcase,
+          rejection_reason='analyze_no_repro'))
   data_handler.mark_invalid_uploaded_testcase(
       testcase, testcase_upload_metadata, 'Unreproducible')
```
vitaliset: Should we emit the event after the "logic execution" of the testcase (e.g., marking it as invalid)? @vitorguidi @ViniciustCosta @jonathanmetzman 

```suggestion
  rejection_event = (
      events.TestcaseRejectionEvent(
          source='handle_noncrash',
          testcase=testcase,
          rejection_reason='analyze_no_repro'))
  data_handler.mark_invalid_uploaded_testcase(
      testcase, testcase_upload_metadata, 'Unreproducible')
  events.emit(rejection_event)
```

ViniciustCosta: As we were discussing offline, I think this is a better way to do it, since we create the event before making changes to the testcase metadata, to assert that we have access to the data we need for the event, but only emit it after actually executing the method.

src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
```diff
@@ -256,11 +257,22 @@ def handle_noncrash(output):
     testcase.status = 'Unreproducible, retrying'
     testcase.put()
 
+    events.emit(
+        events.TestcaseRejectionEvent(
+            source='handle_noncrash',
+            testcase=testcase,
+            rejection_reason='analyze_flake_on_first_attempt'))
```
vitorguidi: This should be an enum within the events module, not a raw string. If for whatever reason we have to change it, and it is referenced in the code more than once (n times), we will have to search over the code base for this many occurences and replace it manually

ViniciustCosta: Hm I agree if all the reasons are mapped and well defined, since at first I thought there could be arbitrary reasons.
If this is true, we probably also need to update some other fields to be enums as well (e.g., `origin` from `TestcaseCreationEvent`) and not arbitrary values.

src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
```diff
@@ -256,11 +257,22 @@ def handle_noncrash(output):
     testcase.status = 'Unreproducible, retrying'
     testcase.put()
 
+    events.emit(
+        events.TestcaseRejectionEvent(
+            source='handle_noncrash',
```
vitorguidi: what happens if I refactor the function signature? I will have to change this manually. This will be very confusing to maintain. If you want to do it, consider using an auxiliary function that returns the function name from the current frame.

Another option is to get rid of this source field, and just log the event. We already get that for free in the logs module, and it would just be a matter of adding another handler that does:

```
logs.info(f'Event emitted: {event}')
```

The second option seems more atractive, since the high level event should contain higher level concepts, and not lower level implementation details (line of code)

It also lacks the task id from the design doc

ViniciustCosta: Regarding the `source`, my argument here is that there may be events with the same parameters emitted from distinct places in the code and it sounds useful to tell them apart.

I agree that passing the function name is not a good approach, but logging the event seems redundant and it makes it unnecessarily harder for users, since they'd need to query the logs to find it. A possible alternative to do it would be to add a default method in the events dataclass that retrieves the caller information, similar to how it is done in the logs. Wdyt?

Re lacks task id: task id is retrieved from the env variable by default in the events module.

vitaliset: I feel like we get the best of both words by having the logging handler and also having the events `emit` function retrieving the caller information. :)

src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
vitaliset: Introduced the `RejectionReason` enum.

vitaliset: By the way, I'm using the `.value` so the dataclass itself is of a simple type and that works well with datastore's input/retrieve logic.

I'm talking offline with @ViniciustCosta and maybe we can directly use the Enum here and handle the input/retrieve elsewhere (maybe using [annotations](https://docs.python.org/3/howto/annotations.html) to deal with type transformations - which might generate problems with modified enum reads...).

ViniciustCosta: Tbh this kind of a nit. My point was just that the useful part of Enum is actually not caring for the values, just using the constants. But in our cases (and this also true for other structs, like `EventTypes`) we are only referencing the value itself with `.value`.

However, having to deal with converting Enums back and forth from Datastore (or any other repo) may be adding unnecessary complexity (mainly when converting back to an event dataclass). An easy alternative, considering our needs, would be simply using the class attributes as constants and not inheriting from Enum (or something similar to this).
```
class RejectionReason:
  ANALYZE_FLAKE_ON_FIRST_ATTEMPT = 'analyze_flake_on_first_attempt'`
  ...
```

vitorguidi: Enum is just an implementation option, I am fine with the string constants. The crux of the issue is just having a single source of truth to define these values, and then we reference them everywhere else.

if the strings offer less friction, lgtm

src/clusterfuzz/_internal/bot/tasks/utasks/analyze_task.py
vitorguidi: Just do dump the thought process here, I am not a fan of this source field.

If we ever need to disambiguate two events trough a source location, then we are probably missing another information field.

For this event right now, we have the 'rejection_reason' field. It suffices.
Also, the code is a moving target: line X right now will move around as the code evolves, and will not be the same as line X in the future. This is very confusing, we will have to check files at specific revisions to make sense of old events.

I would rather drop this field. If we ever need to disambiguate, the event is anemic and needs more higher level information

ViniciustCosta: I see your point, I still think that this could be useful and we don't actually lose anything by having it. The issue with code changes is the same as using the logs. And I think having very specific values for the fields so that we can only call/create events from a specific location is not a very scalable thing to do.


Anyway, @vitaliset already removed the argument passing the function name and this shouldn't change anything in this PR. Maybe I will try to implement a default function to retrieve that info automatically from the events module. Thanks!

src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -41,6 +41,12 @@ class EventTypes(Enum):
   TESTCASE_REJECTION = 'testcase_rejection'
 
 
+class RejectionReason:
```
ViniciustCosta, line 44: Maybe update the events unit tests as well in order to use these values.

ViniciustCosta (approved):
lgtm, thanks!

PR #4823, opened 2025-06-12
[Deployment Revisited][Staging] Add Fuzzer import feature to the job-exporter cronjob

vitorguidi:
### Motivation

This PR adds the capability of importing Fuzzers from the exported data, implemented in #4808 .

### Implementation

Assuming that all data was exported to $SOME_BUCKET, the folder structure for fuzzers looks like this:

```
$SOME_BUCKET/
    fuzzer/
        entities
        some-fuzzer/
            entity.proto
            blobstore_key
            sample_testcase
        another-fuzzer/
            entity.proto
            blobstore_key
            sample_testcase
```

The proto file is the serialized representation of a data_types.Fuzzer entity, and the other files are the associated blobs to the entity, if present.

The entities file contains line separated fuzzer names that were last exported. It was added in this PR.

* Upload sample_testcase and blobstore_key to the target project's blobs bucket, and update the entity with new ids
* Deserialize the protobuf and persist the entity into datastore

### Testing strategy

Unit tests for:
* Correctly creating a fuzzer, from the state where it still does not exist
* Correctly deleting a fuzzer, when from the state where it is present in the target environment, in response to the export list not containing its name
* Correctly updating a fuzzer, once a newer revision with different blobs or fields is exported from the source project

### DIFF
--- a/src/clusterfuzz/_internal/cron/job_exporter.py
+++ b/src/clusterfuzz/_internal/cron/job_exporter.py
@@ -142,14 +142,16 @@ class EntityMigrator:
     entity_bucket_prefix = f'gs://{self._export_bucket}/{self._entity_type}'
     for entity in self._target_cls.query():
       entity_name = getattr(entity, 'name', None)
-      assert entity_name
+      if not entity_name:
+        raise ValueError('Expected entity name to be present, it is not.')
       self._export_entity(entity, entity_bucket_prefix, entity_name)
       entity_names.add(entity_name)
     self._export_entity_names(entity_names, entity_bucket_prefix)
 
   def _import_blobs(self, entity: ndb.Model, entity_name: str,
                     entity_location: str):
-    """Copies exported blobs to a new blob id, and updates the entity."""
+    """Copies exported blobs to a new blob id, and returns a map with
+        the new blob ids."""
     new_blob_ids = {}
     for blobstore_key in self.blobstore_keys:
       source_blob_location = f'{entity_location}/{blobstore_key}'
@@ -165,7 +167,7 @@ class EntityMigrator:
       target_blob_location = f'gs://{storage.blobs_bucket()}/{new_blob_id}'
       if not storage.copy_blob(source_blob_location, target_blob_location):
         raise ValueError(f'Failed to import blob from {source_blob_location} '
-                         'to {target_blob_location}.')
+                         f'to {target_blob_location}.')
       new_blob_ids[blobstore_key] = new_blob_id
     return new_blob_ids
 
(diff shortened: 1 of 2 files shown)


jonathanmetzman (approved):
LGTM

src/clusterfuzz/_internal/cron/job_exporter.py
```diff
@@ -119,25 +119,95 @@ def _export_data_bundle_contents_if_applicable(self, entity: ndb.Model,
...
+
   def export_entities(self):
+    """Exports individual entities of a certain type, and populates a list
+       the individual names of entities for future importing."""
+    entity_names = set()
+    entity_bucket_prefix = f'gs://{self._export_bucket}/{self._entity_type}'
     for entity in self._target_cls.query():
-      self._export_entity(entity)
+      entity_name = getattr(entity, 'name', None)
+      assert entity_name
```
ViniciustCosta: Nit: I think you should raise an explicit exception instead of doing an assert (per go/pystyle#exceptions-decision)

src/clusterfuzz/_internal/cron/job_exporter.py
```diff
@@ -119,25 +119,95 @@ def _export_data_bundle_contents_if_applicable(self, entity: ndb.Model,
...
+        continue
+      if not storage.get(source_blob_location):
+        raise ValueError(
+            f'Absent blob for {blobstore_key} in {entity_name}, it '
+            'should be present.')
+      new_blob_id = blobs.generate_new_blob_name()
+      target_blob_location = f'gs://{storage.blobs_bucket()}/{new_blob_id}'
+      if not storage.copy_blob(source_blob_location, target_blob_location):
+        raise ValueError(f'Failed to import blob from {source_blob_location} '
+                         'to {target_blob_location}.')
```
ViniciustCosta: Probably you have to add f-string after the line break.

src/clusterfuzz/_internal/cron/job_exporter.py
```diff
@@ -119,25 +119,95 @@ def _export_data_bundle_contents_if_applicable(self, entity: ndb.Model,
...
-      self._export_entity(entity)
+      entity_name = getattr(entity, 'name', None)
+      assert entity_name
+      self._export_entity(entity, entity_bucket_prefix, entity_name)
+      entity_names.add(entity_name)
+    self._export_entity_names(entity_names, entity_bucket_prefix)
+
+  def _import_blobs(self, entity: ndb.Model, entity_name: str,
+                    entity_location: str):
+    """Copies exported blobs to a new blob id, and updates the entity."""
```
ViniciustCosta: Nit: I think the docstring may be a bit misleading since IIUC it does not update the entity, returns the new blob ids.

PR #4824, opened 2025-06-12
Schedule fuzz tasks for batch on Chrome.

jonathanmetzman:
Add support for Chrome which doesn't have project weight and needs a small quantity of fuzzing jobs on batch for now.
Our config change combined with this will mean we schedule 5 fuzz tasks on batch every 10 minutes. Tasks have up to 6 hours to complete. This means a peak utilization of 180 batch tasks/instances or about 360 CPUs.

### DIFF
--- a/src/clusterfuzz/_internal/cron/schedule_fuzz.py
+++ b/src/clusterfuzz/_internal/cron/schedule_fuzz.py
@@ -206,7 +206,7 @@ class OssfuzzFuzzTaskScheduler(BaseFuzzTaskScheduler):
 
     # TODO(metzman): Handle high-end jobs correctly.
     num_instances = int(self.num_cpus / self._get_cpus_per_fuzz_job(None))
-    logs.info(f'Scheduling {num_instances} fuzz tasks.')
+    logs.info(f'Scheduling {num_instances} fuzz tasks for OSS-Fuzz.')
 
     choices = random.choices(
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -292,6 +339,14 @@ def get_available_cpus(project: str, regions: List[str]) -> int:
   return available_cpus
 
 
+def respect_project_max_cpus(num_cpus):
```
vitorguidi, line 342: this respects the number of cpus per project on each run, but what ensures that all batch tasks will be finished between every schedule_fuzz run?

jonathanmetzman: Just each run. It's a very weak guard. But assuming jobs all take 6 hours to finish (the limit we set in batch) there will be a peak usage of 360 concurrent tasks which is what we want.

vitorguidi (approved):
lgtm, just state in the PR assumptions about scheduling rate/draining rate/max task duration, so we can know how this is expected to behave over time

src/clusterfuzz/_internal/cron/schedule_fuzz.py
jonathanmetzman: Sorry I conflated tasks with CPUs. I'm going to lower it so we use a max of 360 CPUs.

PR #4827, opened 2025-06-13
[Tracing] Emiting `TestcaseRejectionEvent` during triage

vitaliset:
This PR introduces the emits of `TestcaseRejectionEvent` in the desired steps of the triage task life cycle as stated in the "Testcase rejection" section of "Lightweight Clusterfuzz Tracing" doc.

The other remaining cases where `TestcaseRejectionEvent` must be emitted (namely fuzz task and corpus pruning) will be addressed in separate PRs.

b/394056013

### DIFF
--- a/src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/analyze_task_test.py
+++ b/src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/analyze_task_test.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 """Tests for analyze task."""
 
+import datetime
 import json
 import os
 import tempfile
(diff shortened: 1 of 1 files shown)


ViniciustCosta:
I will do a full review later, but just a heads up to avoid merging this one before actually testing it, since it touches critical part of the code (triage).

src/clusterfuzz/_internal/cron/triage.py
```diff
@@ -262,10 +263,20 @@ def _check_and_update_similar_bug(testcase, issue_tracker):
     # might be caused by non-availability of latest builds. In that case,
     # don't file a new bug yet.
     if similar_testcase.open and not similar_testcase.one_time_crasher_flag:
+      events.emit(
```
vitorguidi, line 266: Now that I think of it, this event will be emited multiple times until the testcase itself is marked as closed, or something else that prevents it from entering this triage loop indefini.

 It is not as simple as it looked from the start, we need to check the cleanup code to understand if that might be correct place to emit

vitorguidi: https://github.com/google/clusterfuzz/blob/cfcc261517f89f363b9fe0446a3a031c5c6bf401/src/clusterfuzz/_internal/cron/cleanup.py#L132C5-L132C52

This seems to be the better place to emit this stuff

vitaliset:
Superseded by #4846.

PR #4829, opened 2025-06-13
[Deployment Revisited][Staging] Add import logic for Jobs in the job-exporter cronjob

vitorguidi:
### Motivation

### Implementation

Assuming that all data was exported to $SOME_BUCKET, the folder structure for jobs looks like this:

```
$SOME_BUCKET/
    job/
        entities
        some-job/
            entity.proto
            custom_binary_key
        another-job/
            entity.proto
            custom_binary_key
```

The proto file is the serialized representation of a data_types.Job entity, and the other files is the associated custom binary, if present.

The entities file contains line separated jobs names that were last exported.

* Upload the custom binary to the target project's blobs bucket, and update the entity with a new custom_binary_key from the new blob id
* Deserialize the protobuf and persist the entity into datastore, by performing an environment string substitution

As per b/422759773, some jobs override the following env vars:
* CORPUS_BUCKET
* QUARANTINE_BUCKET
* FUZZ_LOGS_BUCKET
* BACKUP_BUCKET

This implies testing environments performing mutations on a production environment, thus breaking prod isolation. To circumvent this, environment string substitutions will be performed.

The environment string substitution will be defined as a YAML map in project.yaml, over in the clusterfuzz-config repository, of the following format:

```
job_exporter:
    env_string_substitutions:
        source_value_1: target_value_1
        source_value_2: target_value_2
...
```

This allows us to create the corresponding buckets, and redirect read/writes to within the test environment.

Unit tests for:
* Correctly creating a Job, from the state where it still does not exist, and performing the desired env string substitution
* Correctly deleting a Job, when from the state where it is present in the target environment, in response to the export list not containing its name
* Correctly updating a Job, once a newer revision with different blobs or fields is exported from the source project, while also ensuring correct env string substitution

### DIFF
--- a/src/clusterfuzz/_internal/cron/job_exporter.py
+++ b/src/clusterfuzz/_internal/cron/job_exporter.py
@@ -20,6 +20,7 @@ from google.cloud import ndb
 from google.protobuf import any_pb2
 
 from clusterfuzz._internal.bot.tasks.utasks import uworker_io
+from clusterfuzz._internal.config import local_config
 from clusterfuzz._internal.datastore import data_types
 from clusterfuzz._internal.google_cloud_utils import blobs
 from clusterfuzz._internal.google_cloud_utils import gsutil
@@ -71,13 +72,20 @@ class StorageRSync(RSyncClient):
 class EntityMigrator:
   """Serializes entities to GCS, and imports them back."""
 
-  def __init__(self, target_cls: ndb.Model, blobstore_keys: list[str],
-               entity_type: str, rsync_client: RSyncClient, export_bucket: str):
+  def __init__(  # pylint: disable=dangerous-default-value
+      self,
+      target_cls: ndb.Model,
+      blobstore_keys: list[str],
+      entity_type: str,
+      rsync_client: RSyncClient,
+      export_bucket: str,
+      env_string_substitutions: dict[str, str] = {}):
     self._target_cls = target_cls
     self.blobstore_keys = blobstore_keys
     self._entity_type = entity_type
     self._rsync_client = rsync_client
     self._export_bucket = export_bucket
+    self._env_string_substitutions = env_string_substitutions
 
   def _serialize(self, entity) -> bytes:
     return uworker_io.entity_to_protobuf(entity).SerializeToString()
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/cron/job_exporter.py
```diff
@@ -71,13 +72,20 @@ def rsync(self, source: str, target: str):
...
+      entity_type: str,
+      rsync_client: RSyncClient,
+      export_bucket: str,
+      env_string_substitutions: dict[str, str] = {}):
     self._target_cls = target_cls
     self.blobstore_keys = blobstore_keys
     self._entity_type = entity_type
     self._rsync_client = rsync_client
     self._export_bucket = export_bucket
+    self._env_string_substitutions = env_string_substitutions
```
vitaliset: Using `None` as a default for mutable arguments (like lists or dictionaries) is a better approach because it prevents unintended shared state across multiple uses, ensuring each call or instance gets its own independent object. This avoids subtle, hard-to-debug issues where modifications in one part of your code unexpectedly affect others relying on the "default." go/python-tips/020

```suggestion
      env_string_substitutions: Optional[dict[str, str]] = None):
    self._target_cls = target_cls
    self.blobstore_keys = blobstore_keys
    self._entity_type = entity_type
    self._rsync_client = rsync_client
    self._export_bucket = export_bucket
    self._env_string_substitutions = env_string_substitutions or {}
```
(or `self._env_string_substitutions = env_string_substitutions if env_string_substitutions is not None else {}` to be super readable)

src/clusterfuzz/_internal/cron/job_exporter.py
```diff
@@ -183,6 +200,14 @@ def _import_entity(self, entity_name: str, entity_location: str):
     for blob_key, blob_value in new_blob_ids.items():
       setattr(entity_to_import, blob_key, blob_value)
 
+    # b/422759773 : intends to avoid testing environments from using production
+    # corpus, logs, backup or quarantine buckets, since these are hardcoded
+    # into job env strings.
```
vitaliset: Is this a TODO or just a comment? If it is a comment I would argue that using the style you chose is pretty similar to a TODO comment so it might make one confused (drop the ":" or add the bug number in the middle). If it is a TODO, then follow go/pystyle#todo and go/tott/652

src/clusterfuzz/_internal/tests/appengine/handlers/cron/job_exporter_test.py
```diff
@@ -58,11 +58,13 @@ def _job_templates_equal(template, another_template):
 
 def _sample_job(name='some-job',
                 custom_binary_key='some-key',
-                platform='some-platform'):
+                platform='some-platform',
+                environment_string='some-env-string'):
```
vitaliset, line 63: Maybe add type hints as well.

src/clusterfuzz/_internal/cron/job_exporter.py
```diff
@@ -171,6 +179,15 @@ def _import_blobs(self, entity: ndb.Model, entity_name: str,
       new_blob_ids[blobstore_key] = new_blob_id
     return new_blob_ids
 
+  def _substitute_environment_string(self, env_string: str | None):
```
vitaliset, line 181: Small docstring would be good.

vitaliset (approved):

PR #4831, opened 2025-06-16
[Deployment Revisited][Staging] Add JobTemplate import logic to the job-exporter cronjob

vitorguidi:
### Motivation

### Implementation

Assuming that all data was exported to $SOME_BUCKET, the folder structure for JobTemplates looks like this:

```
$SOME_BUCKET/
    jobtemplate/
        entities
        some-job-template/
            entity.proto
        another-job-template/
            entity.proto
```

The proto file is the serialized representation of a data_types.Job entity, and the other files is the associated custom binary, if present.

The entities file contains line separated job template names that were last exported.

* Deserialize the protobuf and persist the entity into datastore, by performing an environment string substitution

Environment string substitution behaves the same way as described in #4829 

Unit tests for:
* Correctly creating a JobTemplate, from the state where it still does not exist, and performing the desired env string substitution
* Correctly deleting a JobTemplate, when from the state where it is present in the target environment, in response to the export list not containing its name
* Correctly updating a JobTemplate, once a newer revision with different fields is exported from the source project, while also ensuring correct env string substitution

### DIFF
--- a/src/clusterfuzz/_internal/tests/appengine/handlers/cron/job_exporter_test.py
+++ b/src/clusterfuzz/_internal/tests/appengine/handlers/cron/job_exporter_test.py
@@ -861,3 +861,67 @@ class TestJobsAreCorrectlyImported(unittest.TestCase):
     self.assertTrue(
         _entity_blob_was_correctly_imported(another_custom_binary_data,
                                             imported_job.custom_binary_key))
+
+
+@test_utils.with_cloud_emulators('datastore')
+class TestJobTemplatesAreCorrectlyImported(unittest.TestCase):
+  """Test the job exporter job with Fuzzer entitites."""
+
+  def setUp(self):
+    helpers.patch_environ(self)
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/tests/appengine/handlers/cron/job_exporter_test.py
```diff
@@ -861,3 +861,162 @@ def test_jobs_are_correctly_updated(self):
     self.assertTrue(
         _entity_blob_was_correctly_imported(another_custom_binary_data,
                                             imported_job.custom_binary_key))
+
+
+@test_utils.with_cloud_emulators('datastore')
+class TestJobTemplatesAreCorrectlyImported(unittest.TestCase):
+  """Test the job exporter job with Fuzzer entitites."""
```
ViniciustCosta: Typo: "entitites". Maybe update this docstring for "... with Job Template entities" (this may be true for other classes here).

vitaliset (approved):

PR #4832, opened 2025-06-16
[WIP][Tracing] Create the events handler for sending bug notifications

ViniciustCosta:


src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -248,9 +250,109 @@ def emit(self, event: Event) -> Any:
...
+    `{<event_type> : True | list[<task_names>]}`
+  If set to True, all occurrences of the event type are disabled.
+  """
+
+  def __init__(self, disabled_events: dict | None = None):
+    if disabled_events is None:
+      disabled_events = {}
+    self.disabled_events = disabled_events
+
+  def _event_comment(self, event: Event) -> str:
```
vitorguidi: An alternative here would be to extend the event classes with a get_notification() method, which we can just consume from this handler.

I am fine with keeping it like this atm, but we might get user feedback for more informative messages.

src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -248,9 +250,109 @@ def emit(self, event: Event) -> Any:
...
+    """Creates an issue comment based on the event data."""
+    comment = f'A ClusterFuzz event was emitted: {event.event_type}'
```
vitorguidi: This could be the base implementation for the get_notification method, and we assign it to the [Event](https://github.com/google/clusterfuzz/blob/ad39a91bd2756a9e84455d6f5de1eb5d9a7ce0ac/src/clusterfuzz/_internal/metrics/events.py#L59) class. By doing this, we avoid a notifier from having to know specific messages for each type of event.

Another option is to have another class, whose responsibility is to form these messages. This way, we split the responsibilities of producing a commente, from the actual publishing.

Wdyt?

The main concern here is to allow ourselves to evolve these error messages over time. 

We might want a Task event to publish a message along the lines of :

```
ClusterFuzz executed an Analyze Task on the instance ABC-123.
Config = 1
Version=2
Outcome = Success
```
And a testcase rejection event to be something completely different:

```
ClusterFuzz rejected testcase $TESTCASE_ID because if was not leader on group $GROUP_ID
```

ViniciustCosta: Yeah, this makes a lot of sense!

I prefer your first suggestion of having this "get_notification" or "create_message" as a method in the Event dataclass, since by doing this we can attach the implementation to each event that extends this base class (like a `__repr__` method), instead of having to add conditions for event types in another class.

I will apply this change in a new commit, thanks for catching this!

vitorguidi (approved):
lgtm, run this against a real testcase with the debugger to make sure it behaves as we expect at runtime

PR #4833, opened 2025-06-17
[Deployment Revisited][Staging] Add DataBundle import logic to the job-exporter cronjob

vitorguidi:
### Motivation

### Implementation

Assuming that all data was exported to $SOME_BUCKET, the folder structure for data bundles looks like this:

```
$SOME_BUCKET/
    databundle/
        entities
        some-bundle/
            entity.proto
            contents/
              blob1
              blob2
        another-bundle/
            entity.proto
            contents/
              blob3
              blob4
```

The proto file is the serialized representation of a data_types.DataBundle entity, and the files under the content folder are the original contents of the data bundle bucket.

The entities file contains line separated data bundle names that were last exported.

* RSync the contents folder for the data bundle in question, to a new bucket in the target project. This is done through the [get_data_bundle_bucket_name](https://github.com/google/clusterfuzz/blob/ad39a91bd2756a9e84455d6f5de1eb5d9a7ce0ac/src/clusterfuzz/_internal/datastore/data_handler.py#L1026) function
* Deserialize the protobuf and persist the entity into datastore, while changing the bucket name field to the rsync target

Unit tests for:
* Correctly creating a DataBundle, from the state where it still does not exist, and copying over the data bundle contents
* Correctly deleting a DataBundle, when from the state where it is present in the target environment, in response to the export list not containing its name
* Correctly updating a DataBundle, once a newer revision with different fields is exported from the source project, while also ensuring the blobs remain up to date

### Attention points

The implementation for StorageRSync was incomplete. List_blobs returns the file names from the root bucket, and we want to use the file names starting from the contents folder. This worked for exporting the data bundles, since the blobs were on the top level folder, but not for the import:

For this reason, the storage rsync implementation was updated. It should not change the behavior in a real deployment, since it is a unit testing only class

```
gs://export-bucket/databundle/some-bundle/contents/blob

gsutil ls export-bucket:
databundle/some-bundle/contents/blob

gsutil cp gs://export-bucket/databundle/some-bundle/contents/blob gs://new-bucket/blob
```

### DIFF
--- a/src/clusterfuzz/_internal/tests/appengine/handlers/cron/job_exporter_test.py
+++ b/src/clusterfuzz/_internal/tests/appengine/handlers/cron/job_exporter_test.py
@@ -1143,3 +1143,73 @@ class TestDataBundlesAreCorrectlyImported(unittest.TestCase):
 
     remaining_data_bundles = list(data_types.DataBundle.query())
     self.assertEqual(0, len(remaining_data_bundles))
+
+  def test_data_bundles_are_correctly_modified(self):
+    """Tests if a Data Bundle and its blobs are correctly modified
+      once a newer version is exported."""
+    bundle_name = 'some-bundle'
+    bucket_name = 'some-bucket'
+    blob_data = b'some-data'
+    bundle = _sample_data_bundle(
+        name=bundle_name,
+        bucket_name=bucket_name,
+    )
+    _register_entity_and_upload_blobs(
+        entity=bundle,
+        blobstore_key_content=None,
+        sample_testcase_contents=None,
+        custom_binary_contents=None,
+        blobs_bucket=self.blobs_bucket,
+        data_bundle_blob_contents=blob_data,
+    )
+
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/tests/appengine/handlers/cron/job_exporter_test.py
```diff
@@ -172,6 +178,13 @@ def _register_entity_and_upload_blobs(
     storage.write_data(custom_binary_contents,
                        f'gs://{blobs_bucket}/{custom_binary_key}')
 
+  if data_bundle_blob_contents:
+    assert isinstance(entity, data_types.DataBundle)
+    bundle_bucket = entity.bucket_name
+    assert bundle_bucket
```
vitaliset, line 184: nit: Maybe raise or at least have some message when failing assertion. go/pystyle#exceptions

vitaliset (approved):

PR #4839, opened 2025-06-24
Stop tagging fuzz blockers as release blockers.

letitz:
This is causing issues for Chrome releases.

Fixes: https://crbug.com/399218120


pgrace-google (commented):
thank you!

PR #4794, opened 2025-05-14
Remove logs filter that adds extras to json_fields

ViniciustCosta:
In order to enable the structured logging project for cronjobs in GKE/GAE, it was necessary to add a logs filter that inserts the python logging `extras` into the `json_fields` for GCP logging (https://github.com/google/clusterfuzz/pull/4781).
However, this is breaking some of the logs ([Error log](https://pantheon.corp.google.com/errors/detail/CK2Sz6Tk9f-wrwE;time=PT1H;locations=global?e=-13802955&inv=1&invt=AbxY7Q&mods=logs_tg_prod&project=google.com:clusterfuzz)), thus this PR temporarily removes this filter until we can debug these errors.

PR #4841, opened 2025-06-25
[Fix] Fix structured logs in cronjobs

ViniciustCosta:
### Background
In [#4781](https://github.com/google/clusterfuzz/pull/4781), it was necessary to add a logs filter to propagate the structured fields into GCP Logging for cronjobs in GAE/GKE (this is an unfortunate workaround due to different log configurations set depending on the environment).

However, the way this filter was implemented broke some of the logs, when non-serializable arguments were passed to the python logging extras fields. Thus, we disabled it in [#4794](https://github.com/google/clusterfuzz/pull/4794).

### Description
This fix updates the logs filter (responsible for adding `extras` into the `json_fields` for GCP logging) by checking whether the values in logs `extras` are JSON serializable and calling the `_handle_unserializable()` function for the negative cases, which simply converts them into strings.

By doing this, we avoid the type errors that occurred in GCP (`TypeError: Object is not JSON serializable`)  while also maintaining the structure of the json for indexed queries.

### Tests
In order to test in the GKE environment, I leveraged our development environment with candidate release. I deployed a debug cronjob (`debug-logs`) calling the same log entry that broke the cleanup cronjob before. 

The test results, before and after landing the fix, were:

* Current fail state ([link to logs](https://cloudlogging.app.goo.gl/fjrBwnt5RpGs7YQy9)):
image
image

* After the fix  ([link to logs](https://cloudlogging.app.goo.gl/jc4mFe8sbiUZAJAW7)):
image
image

* Also, in order to verify if the structured json fields were correctly maintained, I added a more complex log entry:
  ```py
  logs.warning('Test structured logs.', test_dict_field={'key1':'v1', 'key2':'v2'}, test_set_field={'1', '2'}
  ```
  image

b/417782114


src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -334,12 +343,15 @@ def json_fields_filter(record):
...
-          k: truncate(v, STACKDRIVER_LOG_MESSAGE_LIMIT)
-          for k, v in getattr(record, 'extras', {}).items()
-      }
-  })
+  json_extras = {}
+  for key, val in getattr(record, 'extras', {}).items():
+    if _is_json_serializable(val):
+      valid_value = val
+    else:
+      valid_value = _handle_unserializable(val)
```
vitaliset: nit: go/pystyle#conditional-expressions

```suggestion
    valid_value = val if _is_json_serializable(val) else _handle_unserializable(val) 
```

ViniciustCosta: The line gets too long and when I try to follow the go/pytstyle guide, our formatter changes the style. Let me know what you think

vitaliset: I think that looks good. :)

vitorguidi (approved):

PR #4845, opened 2025-06-28
MNT Truncate for non string objects

vitaliset:
@ViniciustCosta realized that we don't truncate values that are not strings, even though we might want to truncate them for being too large to be logged (b/414397533).

This PR implements recursive truncation for dict-like objects (such as dictionaries and dataclasses) and lists/tuples by applying truncate to their values. We do this to keep the structure logging as smooth as possible. When that is not possible, we simply transform the object into its string representation and truncate that.

Evidence that the PR works:
- [Debugger config and script ran](https://paste.googleplex.com/6527399831011328).
- [GCP log](https://pantheon.corp.google.com/logs/query;cursorTimestamp=2025-06-28T02:14:55.125773Z;endTime=2025-06-28T02:17:12Z;query=SEARCH%2528%22aaaaatruncatedaaaa%22%2529%0A--%20jsonPayload.extras.extras.class.integer%20%3D%2010%0Atimestamp%3D%222025-06-28T02:14:55.125773Z%22%0AinsertId%3D%22rsii28efzr6u%22;startTime=2025-06-28T02:12:12Z?project=google.com:clusterfuzz).
- Prints of the log (because of TTL):[1](https://screenshot.googleplex.com/AzVYTwy8k5HPQmz), [2](https://screenshot.googleplex.com/4MviMwVMEiBb3FG).

b/414397533

### DIFF
--- a/src/clusterfuzz/_internal/metrics/logs.py
+++ b/src/clusterfuzz/_internal/metrics/logs.py
@@ -184,8 +184,33 @@ def get_logging_config_dict(name):
   }
 
 
-def truncate(msg, limit):
-  """We need to truncate the message in the middle if it gets too long."""
+def truncate(
+    msg: Any,
+    limit: int,
+) -> Any:
+  """We need to truncate the message in the middle if it gets too long.
+
+  If the message is a string and exceeds the character `limit`, it is
+  truncated in the middle, with a notice indicating how many characters
+  were removed.
+
+  For dictionaries, lists, and tuples, the function recurses over their
+  values/items. For dataclasses, it first converts them to dictionaries
+  before recursing.
+
+  Certain types defined in `NON_TRUNCATABLE_TYPES` are returned as-is.
+  All other types are coerced to strings before truncation.
+
+  Args:
+    msg: The message or collection to truncate.
+    limit: The maximum character length for strings.
+
+  Returns:
+    The truncated message, which may be a string or a collection of
+    the same type as the input. While basic types like strings, lists, and
+    dictionaries retain their type, be aware that dataclasses are converted
+    to dictionaries, and other unhandled objects are converted to strings.
+  """
   if isinstance(msg, NON_TRUNCATABLE_TYPES):
     return msg
 
--- a/src/clusterfuzz/_internal/metrics/logs.py
+++ b/src/clusterfuzz/_internal/metrics/logs.py
@@ -190,26 +190,27 @@ def truncate(
 ) -> Any:
   """We need to truncate the message in the middle if it gets too long.
 
-  If the message is a string and exceeds the character `limit`, it is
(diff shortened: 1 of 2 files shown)


javanlacerda (approved):
Very well implemented, thanks!

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -184,7 +186,24 @@ def get_logging_config_dict(name):
...
+
+  if isinstance(msg, dict):
+    return {k: truncate(v, limit) for k, v in msg.items()}
+
+  if isinstance(msg, (list, tuple)):
+    return type(msg)(truncate(item, limit) for item in msg)
+
+  # Coerce all other types to a string
+  if not isinstance(msg, str):
+    msg = str(msg)
```
ViniciustCosta, line 230: I'm a bit skeptical about converting every other object into a string, specially because this is a truncation function, so I wouldn't expect it to change the type of my object. 

I think a "smoother" workaround is to check if the len of `str(msg)` is greater than the limit and, only then, actually convert it into a string to be truncated. Event though this still changes the type, it only "penalizes" very large objects.
Also, you can explicit this in the docstring, stating that if the object is too large, it's converted into a string.

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -184,7 +186,24 @@ def get_logging_config_dict(name):
 
 def truncate(msg, limit):
   """We need to truncate the message in the middle if it gets too long."""
```
ViniciustCosta: Maybe add a brief description of this new behavior in the docstring.

src/clusterfuzz/_internal/metrics/logs.py
ViniciustCosta: This is not a blocker though, since in practice all fields will serialized into a JSON for GCP logging, and any object that's not json serializable ([types ref](https://docs.python.org/3/library/json.html#json.JSONEncoder)) is turned into a string (per `_handle_unserializable`).

vitaliset: That's a great point, and I appreciate you thinking through the implications. :) The idea of conditionally converting the object based on the size is definitely smoother. However, I believe the current, "simpler" implementation is a reasonable starting point. 

Regarding converting every other object into a string, my thinking is that for a function designed to truncate, operating on a string representation is a logical default for any non-string input. Not all object types can be intuitively truncated otherwise.

Let's log you suggestion on b/414397533 so we can revisit it if needed. :)

ViniciustCosta: That's fine by me. Thanks for adding the great docstring!

ViniciustCosta (approved):

PR #4847, opened 2025-06-30
[Tracing] Emiting `TestcaseRejectionEvent` during fuzz task

vitaliset:
This PR introduces the emits of `TestcaseRejectionEvent` in the desired steps of the fuzz task life cycle as stated in the "Testcase rejection" section of "Lightweight Clusterfuzz Tracing" doc.

The other remaining cases where `TestcaseRejectionEvent` must be emitted (namely corpus pruning and grouper) will be addressed in separate PRs.

b/394056013

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
@@ -849,6 +849,10 @@ def postprocess_process_crashes(uworker_input: uworker_msg_pb2.Input,
           uworker_output=uworker_output,
           fully_qualified_fuzzer_name=fully_qualified_fuzzer_name)
     else:
+      events.emit(
+          events.TestcaseRejectionEvent(
+              testcase=existing_testcase,
+              rejection_reason=events.RejectionReason.FUZZ_DUPLICATE))
       _update_testcase_variant_if_needed(group, existing_testcase,
                                          fuzz_task_output.crash_revision,
                                          uworker_input.job_type)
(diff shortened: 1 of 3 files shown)


vitaliset:
There are some other places where one could argue that there is a rejection of a crash being made, example:
- [NoMoreUploadUrlsError](https://github.com/google/clusterfuzz/blob/90fa2356dfc20676878800642b0c672518c6881d/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py#L1151-L1156)
- [Unable to store testcase in blobstore](https://github.com/google/clusterfuzz/blob/90fa2356dfc20676878800642b0c672518c6881d/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py#L1161-L1164)
- [Invalid crash](https://github.com/google/clusterfuzz/blob/90fa2356dfc20676878800642b0c672518c6881d/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py#L988-L997)

But I don't think it makes sense to emit anything there since there is no testcase associated with the rejection.

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -849,6 +849,10 @@ def postprocess_process_crashes(uworker_input: uworker_msg_pb2.Input,
           uworker_output=uworker_output,
           fully_qualified_fuzzer_name=fully_qualified_fuzzer_name)
     else:
+      events.emit(
```
ViniciustCosta, line 852: I don't think emitting this event type here makes a lot of sense, since the testcase has not been created yet in order to be rejected. As per the tracing design doc: "During fuzz and corpus pruning task, a crash will never be promoted to a testcase if rejected, so there is nothing to track a lifecycle for."

We could, however, create a new event type like `crash_deduplication`, if we want to see events in this case. I'm not sure if this could be a bit spammy though, since it seems that fuzz task may find a lot of duplicated crashes.

vitaliset:
Incorrectly emitting events for existing instead of the new one. The new one is never promoted to a testcase so it does not make sense to emit anything here.

PR #4848, opened 2025-06-30
[Tracing] Emiting `TestcaseRejectionEvent` during corpus pruning

vitaliset:
This PR introduces the emits of `TestcaseRejectionEvent` in the desired steps of the corpus pruning task life cycle as stated in the "Testcase rejection" section of "Lightweight Clusterfuzz Tracing" doc.

The other remaining cases where `TestcaseRejectionEvent` must be emitted (namely grouper) will be addressed in separate PRs.

b/394056013


ViniciustCosta (changes requested):
I think this is the same case as https://github.com/google/clusterfuzz/pull/4847, where testcase rejection does not seem like the correct event to be emitting here :/

vitaliset:
Incorrectly emitting events for existing instead of the new one. The new one is never promoted to a testcase so it does not make sense to emit anything here.

PR #4849, opened 2025-06-30
[Tracing] Emiting TestcaseRejectionEvent during grouper

vitaliset:
This PR introduces the emits of `TestcaseRejectionEvent` in the desired steps of the grouper life cycle as stated in b/394056013#comment9.

This is the last `TestcaseRejectionEvent` emision pull request.

b/394056013

### DIFF
--- a/src/clusterfuzz/_internal/tests/appengine/handlers/cron/grouper_test.py
+++ b/src/clusterfuzz/_internal/tests/appengine/handlers/cron/grouper_test.py
@@ -18,6 +18,7 @@ import unittest
 
 from clusterfuzz._internal.cron import grouper
 from clusterfuzz._internal.datastore import data_handler
+from clusterfuzz._internal.metrics import events
 from clusterfuzz._internal.tests.test_libs import helpers
 from clusterfuzz._internal.tests.test_libs import test_utils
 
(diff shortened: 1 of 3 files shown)


src/clusterfuzz/_internal/tests/appengine/handlers/cron/grouper_test.py
```diff
@@ -592,3 +593,58 @@ def test_group_exceed_max_testcases(self):
...
+    testcase2.crash_state = 'state'
+    testcase1.put()
+    testcase2.put()
+
+    grouper.group_testcases()
+
+    self.assertEqual(1, len(self.emitted_events))
+    self.mock.emit.assert_called_once_with(
+        events.TestcaseRejectionEvent(
+            testcase=testcase2,
```
ViniciustCosta: nit: I wonder if it's expected that the return of querying for testcases (get_open_testcase_id_iterator) is ordered by insertion? I say this because we could have a flaky test fail due to it returning testcase2 first. Just something to keep in mind, not blocker :)

ViniciustCosta (approved):
lgtm

src/clusterfuzz/_internal/tests/appengine/handlers/cron/grouper_test.py
vitaliset: Nice catch. I was already handling this in the second test. I've update this now with less brittle (and behaviour oriented) asserts.

ViniciustCosta: Great, thanks!

vitorguidi (approved):

PR #4834, opened 2025-06-23
[Tracing] Emit Testcase Creation events 

ViniciustCosta:
## Description
Adding emits for `TestcaseCreationEvent`, where new testcases are created and stored into datastore. This happens during:
* upload_testcase (manual upload)
* corpus_pruning_task
* fuzz_task

Note: For Corpus Pruning, the feature of uploading testcases for crashes found during pruning is commented out (this is being tracked and fixed at https://github.com/google/clusterfuzz/pull/4825). Thus, both the event emission and the unit tests are disabled. I enabled it locally and ran the unit tests, which completed successfully.

As referenced in our tracing design doc, "Testcase Creation" section.
b/394048114

#### Unit tests
Added unit tests to guarantee that the events are correctly emitted when a testcase is created in all these contexts. For `upload_testcase` and `corpus_pruning`, I leveraged some existing unit tests that already asserted the testcase creation steps. For `fuzz_task`, it was necessary to create a new test in order to call the testcase creation method and assert the event emission.

#### Runtime tests
For runtime tests, since it is difficult to actually trigger these code paths locally, I intend to deploy to our dev/staging environments and monitor the events. Evidence of events being emitted in dev:
* Fuzz task ([link to logs creating testcase](https://pantheon.corp.google.com/logs/query;cursorTimestamp=2025-06-25T15:19:40.084565Z;duration=PT3H;query=jsonPayload.extras.clusterfuzz_version%3D%2234abee68%22%0AjsonPayload.extras.task_name%3D%22fuzz%22%0AjsonPayload.extras.stage%3D%22postprocess%22%0Atimestamp%3D%222025-06-25T15:19:40.084565Z%22%0AinsertId%3D%22e683gef6i7ak9%22?project=clusterfuzz-development)):
Event: image
Actual testcase created: 
image

* Upload testcase event:
image
Actual testcase created:
image

PR #4850, opened 2025-06-30
Keeping event emissions test standard by mocking `_get_datetime_now`

vitaliset:
Reverts #4828.

Reading #4834 I realized the old approach is simpler.


ViniciustCosta:
Thanks! I think having to create the mock methods for all events would add some unnecessary work and complexity to the tests. Also, this approach seems to give more assurance that the event creation codepath works when called from the module being instrumented.

PR #4851, opened 2025-06-30
[Tracing] Emiting`IssueFilingEvent` during triage

vitaliset:
This PR introduces the emits of `IssueFilingEvent` in the desired steps of the triage life cycle as stated in the "Issue filing" section of "Lightweight Clusterfuzz Tracing" doc.

b/394059579

### DIFF
--- a/src/clusterfuzz/_internal/cron/triage.py
+++ b/src/clusterfuzz/_internal/cron/triage.py
@@ -32,6 +32,7 @@ from clusterfuzz._internal.issue_management import issue_tracker_utils
 from clusterfuzz._internal.metrics import crash_stats
 from clusterfuzz._internal.metrics import logs
 from clusterfuzz._internal.metrics import monitoring_metrics
+from clusterfuzz._internal.metrics import events
 
 from . import grouper
 
@@ -538,6 +539,12 @@ def _triage_testcase(testcase, excluded_jobs, all_jobs, throttler):
   _create_filed_bug_metadata(testcase)
   issue_filer.notify_issue_update(testcase, 'new')
 
+  events.emit(
+      events.IssueFilingEvent(
+          testcase=testcase,
+          issue_tracker=issue_tracker.project,
+          issue_id=str(testcase.bug_information),
+          issue_created=True))
   logs.info('Filed new issue %s for testcase %d.' % (testcase.bug_information,
                                                      testcase_id))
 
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/cron/triage.py
```diff
@@ -538,6 +539,12 @@ def _triage_testcase(testcase, excluded_jobs, all_jobs, throttler):
   _create_filed_bug_metadata(testcase)
   issue_filer.notify_issue_update(testcase, 'new')
 
+  events.emit(
+      events.IssueFilingEvent(
+          testcase=testcase,
+          issue_tracker=issue_tracker.project,
+          issue_id=str(testcase.bug_information),
+          issue_created=True))
```
ViniciustCosta, line 552: I wonder why do we need this `issue_created`  field, if it's always set to true.

If we want to emit events for failed issue filing, maybe we should change the event_type to `issue_filing` and add the emit when `_file_issue` fails.

src/clusterfuzz/_internal/cron/triage.py
```diff
@@ -538,6 +539,12 @@ def _triage_testcase(testcase, excluded_jobs, all_jobs, throttler):
   _create_filed_bug_metadata(testcase)
   issue_filer.notify_issue_update(testcase, 'new')
 
+  events.emit(
+      events.IssueFilingEvent(
+          testcase=testcase,
+          issue_tracker=issue_tracker.project,
```
ViniciustCosta: IIUC this is not returning exactly what was intended for this field based on the desing doc, which is the issue tracker type, e.g., buganizer, jira, etc. This returns a project ([example in code](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/issue_management/issue_filer.py#L360)) associated to a issue tracker. I think this is very relevant and could be another field as well, though. 

This issue tracker part is kind of confusing, so we should double check and maybe discuss offline.

src/clusterfuzz/_internal/cron/triage.py
vitaliset: As discussed offline, adding a new event emission when issue filling fails.

src/clusterfuzz/_internal/cron/triage.py
vitaliset: As discussed offline, updated the naming as the `issue_tracker.project` still gives the relevant information we want to have when tracing the testcase lifecycle.

vitorguidi (approved):

PR #4853, opened 2025-07-01
[WIP][Deployment revisited][Staging] Sample crashes during fuzz task postprocess

vitorguidi:
### Motivation

In order to get test coverage for analyze task in our testing environments (dev and staging), we need to redirect fuzz task crashes through the upload testcase handler in appengine. This will trigger analyze task, allowing this to run without people having to manually input testcases.

It also opens up the possibility of spliting the fuzzing and testcase lifecycle domains, by having testcases being created exclusively through the upload_testcase API.

### Implementation

The Crash class in fuzz task has all the data required for a testcase to be created through a manual appengine upload. The desired topic will be fetched through project.yaml (ProjectConfig), and a sample rate. For every crash group leader, if it gets sampled according to the desired probability, it will get published for replication.

The following fields will be added to project.yaml:

```
fuzz_task_sampling:
    sampling_topic: X
    sampling_rate: Y
```

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
@@ -52,7 +52,9 @@ from clusterfuzz._internal.fuzzing import fuzzer_selection
 from clusterfuzz._internal.fuzzing import leak_blacklist
 from clusterfuzz._internal.google_cloud_utils import big_query
 from clusterfuzz._internal.google_cloud_utils import blobs
+from clusterfuzz._internal.google_cloud_utils import pubsub
 from clusterfuzz._internal.google_cloud_utils import storage
+from clusterfuzz._internal.config import local_config
 from clusterfuzz._internal.metrics import events
 from clusterfuzz._internal.metrics import fuzzer_logs
 from clusterfuzz._internal.metrics import fuzzer_stats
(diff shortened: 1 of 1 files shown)


src/appengine/libs/auth.py
```diff
@@ -146,6 +171,14 @@ def get_current_user():
   if iap_email:
     return User(iap_email)
 
+  try:
+    bearer_token_email = get_email_from_bearer_token(current_request)
+    if bearer_token_email:
+      logs.info(f'bearer token email = {bearer_token_email}')
+      return User(bearer_token_email)
+  except helpers.UnauthorizedError as e:
+    logs.info(f'Exception while trying to get email from auth token: {e}')
```
ViniciustCosta: Nit: maybe this should be a warning.

src/appengine/handlers/upload_testcase.py
```diff
@@ -304,18 +304,39 @@ def get_upload(self):
...
+                     trusted_agreement_signed=False,
+                     testcase_id=None,
+                     testcase_metadata=None) -> str:
+    """Holds the logic for actually performing a testcase upload."""
     if testcase_id and not uploaded_file:
       testcase = helpers.get_testcase(testcase_id)
       if not access.can_user_access_testcase(testcase):
         raise helpers.AccessDeniedError()
 
+      # Use minimized testcase for upload (if available).
```
ViniciustCosta, line 336: Is this comment in the right place?

PR #4857, opened 2025-07-02
[Tracing] Create the Task Execution event

ViniciustCosta:
This PR adds the necessary code for emitting task execution events:
* Created the `TaskExecutionEvent` data model
* Updated the datastore entity with the new fields for this event.
* Updated the events unit test.

Instrumentation of the code to actually emit the event will be done in following PRs.

b/394059581

### DIFF
--- a/src/clusterfuzz/_internal/metrics/events.py
+++ b/src/clusterfuzz/_internal/metrics/events.py
@@ -41,6 +41,7 @@ class EventTypes:
   TESTCASE_CREATION = 'testcase_creation'
   TESTCASE_REJECTION = 'testcase_rejection'
   ISSUE_FILING = 'issue_filing'
+  TASK_EXECUTION = 'task_execution'
 
 
 class TestcaseOrigin:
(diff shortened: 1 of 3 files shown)


src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -160,11 +176,24 @@ class IssueFilingEvent(BaseTestcaseEvent, BaseTaskEvent):
...
+@dataclass(kw_only=True)
+class TaskExecutionEvent(BaseTestcaseEvent, BaseTaskEvent):
+  """Task execution event."""
+  event_type: str = field(default=EventTypes.TASK_EXECUTION, init=False)
+  # Task stage (preprocess, main or postprocess).
+  task_stage: str | None = None
+  # Task status (e.g., started, finished, exception).
+  task_status: str | None = None
+  # UTask return code (defined in the uworker protobuf error types).
+  task_return_code: int | None = None
```
vitorguidi: I am not sure if this is an int, these are defined in the uworker_msg_pb2.ErrorType class

vitorguidi (approved):

vitaliset (approved):

src/clusterfuzz/_internal/metrics/events.py
ViniciustCosta: It is an integer, as seen in the [ValueType definition](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/protos/uworker_msg_pb2.pyi#L51) and [each error being attributed this ValueType](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/protos/uworker_msg_pb2.pyi#L56).
Also, I'm using the `uworker_msg_pb2.ErrorType`  in tests.

I do think it would be better to have the literal name of the return code (e.g., ANALYZE_BUILD_SETUP), so that users wouldn't need to check the proto to understand the code, but I couldn't figure out a nice way of retrieving this without using like globals() and causing an unnecessary overhead.

ViniciustCosta: As you suggested, there is actually an easy way as it's done in https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py#L176.

I updated the field, thanks!

vitorguidi (approved):

PR #4858, opened 2025-07-02
[Tracing] Emiting `IssueFilingEvent` for frontend  

vitaliset:
This PR introduces the emits of `IssueFilingEvent` in the desired steps of the "create issue" button of the frontend. [See](https://screenshot.googleplex.com/ARuz5d4Nu3h98w2). This gap was spoted by @ViniciustCosta on b/394059579#comment6

b/394059579

### DIFF
--- a/src/appengine/handlers/testcase_detail/create_issue.py
+++ b/src/appengine/handlers/testcase_detail/create_issue.py
@@ -47,13 +47,12 @@ class Handler(base_handler.Handler):
         user_email=user_email,
         additional_ccs=additional_ccs)
 
-    event = events.IssueFilingEvent(
-        testcase=testcase,
-        issue_tracker_project=issue_tracker.project,
-        issue_id=str(issue_id) if issue_id else None,
-        issue_created=bool(issue_id))
-    print("NILVO", event.issue_id)
-    events.emit(event)
+    events.emit(
+        events.IssueFilingEvent(
+            testcase=testcase,
+            issue_tracker_project=issue_tracker.project,
+            issue_id=str(issue_id) if issue_id else None,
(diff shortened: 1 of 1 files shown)


src/appengine/handlers/testcase_detail/create_issue.py
```diff
@@ -46,6 +47,13 @@ def create_issue(testcase, severity, cc_me):
         user_email=user_email,
         additional_ccs=additional_ccs)
 
+    events.emit(
+        events.IssueFilingEvent(
+            testcase=testcase,
+            issue_tracker_project=issue_tracker.project,
+            issue_id=str(issue_id) if issue_id else None,
+            issue_created=bool(issue_id)))
```
ViniciustCosta, line 55: nit: I'd prefer the same as in the other field (to avoid unexpected behavior with bool()):
`issue_created=True if issue_id else False`

ViniciustCosta (approved):
lgtm with nit

src/appengine/handlers/testcase_detail/create_issue.py
vitaliset: Both expressions rely on the exact same underlying truthiness check. The `if issue_id` in the ternary operator effectively performs a bool() conversion under the hood, so any potential "unexpected behavior" would be present in both versions. Anyway, I can change for style if you still prefer.

ViniciustCosta: You're right, my bad! What I was going for was: `if issue_id is not None` 

But I'm fine with keeping as is :)

vitorguidi (approved):

PR #4860, opened 2025-07-04
[Tracing] Instrument code to emit Task Execution events

ViniciustCosta:
### Description
This PR instruments the code to emit `TaskExecutionEvents`  (defined at https://github.com/google/clusterfuzz/pull/4857) during utasks execution (events for trusted tasks will be added in a next PR).

The implementation leverages the `_MetricRecorder` from utasks module, since it already wraps the execution of all task stages while also storing important metadata from the task. This enables us sending events with context for each stage, even if unhandled exceptions happen during runtime.

Events are emitted in the following cases:
* After Preprocess executed:
  * If an exception occurs:   `Event: {TaskStatus=Exception, Outcome=Unhandled exception}`
  * If uworker input is NOT returned: `Event: {TaskStatus=Exception, Outcome=No preprocess return}`
  * Else preprocess is successful: `Event: {TaskStatus=Started, Outcome=None}`
* At the start of Postprocess:
  *  States that main executed: `Event: {TaskStatus=Postprocess Started, Outcome=<uworker_output.error_type>}`
* After Postprocess executed:
  * If an exception occurs: `Event: {TaskStatus=Exception, Outcome=Unhandled exception}`
  * Else postprocess finishes successfully: `Event: {TaskStatus=Postprocess Completed, Outcome=<uworker_output.error_type>}`

Notice that a known limitation is that we do not send events from utask main, since we currently don't have a workaround to access datastore from untrusted workers.

b/394059581

### Tests
Unit tests were added to verify all the instrumented codepaths and check the correctness of the events being emitted.

Runtime tests were done by deploying to dev environment.
* Task execution events created: 
image

* No new error groups related to the events:  [Error groups](https://pantheon.corp.google.com/errors;time=PT6H?project=clusterfuzz-development&e=-13802955&inv=1&invt=Ab2YGw&mods=logs_tg_prod)

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
@@ -124,7 +124,7 @@ class _MetricRecorder(contextlib.AbstractContextManager):
     self._event_data = None
     # Error type from utask main handled in postprocess.
     self.post_utask_main_failure = None
-    # Whether proprocess returned the uworker input.
+    # Whether preprocess returned the uworker input.
     self.preprocess_returned: bool | None = None
 
   def set_task_details(
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
@@ -163,9 +163,14 @@ class _MetricRecorder(contextlib.AbstractContextManager):
   def emit_task_events(self, task_status: str,
                        task_outcome: str | None = None) -> None:
     """Helper to emit task execution events during the recorder context."""
-    if self._event_data is None or not environment.is_tworker():
-      # Either missing `_event_data`, which is set during `set_task_details`.
-      # Or events can't be sent from untrusted workers for now.
+    if not environment.is_tworker():
+      # Events can't be sent from untrusted workers for now.
+      logs.warning(f'Attempted emit of task execution event from untrusted worker.')
+      return
+
+    if self._event_data is None:
+      # Missing `_event_data`, which is set during `set_task_details`.
+      logs.warning(f'Missing event data for emitting utask execution event.')
       return
 
     event_task_exec = events.TaskExecutionEvent(
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
@@ -398,9 +398,10 @@ def tworker_postprocess_no_io(utask_module, uworker_output, uworker_input):
 
     # This emit is needed because we are not sending events from utask main.
     # Thus, this confirms that main finished and postprocess will start.
-    recorder.emit_task_events(events.TaskStatus.FINISHED,
-                              uworker_msg_pb2.ErrorType.Name(  # pylint: disable=no-member
-                                  uworker_output.error_type))
+    recorder.emit_task_events(
+        events.TaskStatus.FINISHED,
+        uworker_msg_pb2.ErrorType.Name(  # pylint: disable=no-member
+            uworker_output.error_type))
 
     utask_module.utask_postprocess(uworker_output)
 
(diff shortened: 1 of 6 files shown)


vitaliset (approved):

src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
```diff
@@ -137,12 +160,58 @@ def set_task_details(self,
       # Ensure we always have a value after this method returns.
       assert self._preprocess_start_time_ns is not None
 
+  def emit_task_events(self, task_status: str,
+                       task_outcome: str | None = None) -> None:
+    """Helper to emit task execution events during the recorder context."""
+    if self._event_data is None or not environment.is_tworker():
```
vitorguidi: nit: you can add a log here stating that we attempted to file an event in a tworker codepath. We should not be doing that, but is useful to have that in case it slips through code review

src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
```diff
@@ -137,12 +160,58 @@ def set_task_details(self,
...
+      # Either missing `_event_data`, which is set during `set_task_details`.
+      # Or events can't be sent from untrusted workers for now.
+      return
+
+    event_task_exec = events.TaskExecutionEvent(
+        **self._event_data,
+        task_stage=self._subtask.value,
+        task_status=task_status,
```
vitorguidi, line 180: We need a task id here. Without it we cannot cross reference logs with events, which is the mechanism by which we allow users w/o GCP access to troubleshoot clusterfuzz (download logs filtered by such task id)

src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
ViniciustCosta: Sure, that makes sense, thanks

src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
ViniciustCosta: As we discussed offline, the `task_id` is retrieved automatically by the base class event ([BaseTaskEvent](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/metrics/events.py#L143)) from the `CF_TASK_ID` env var, which is set in `commands.py` for preprocess and propagated through the `uworker_env`.

I will add an explicit assertion in tests to make that clear, thanks!

src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
```diff
@@ -405,8 +498,19 @@ def tworker_postprocess(output_download_url) -> None:
...
         environment.platform(),
-        uworker_output.uworker_input.preprocess_start_time)
+        uworker_output.uworker_input.preprocess_start_time,
+        task_argument=uworker_output.uworker_input)
+    recorder.post_utask_main_failure = uworker_output.error_type
+
+    # This emit is needed because we are not sending events from utask main.
+    # Thus, this confirms that main finished and postprocess will start.
+    recorder.emit_task_events(
+        events.TaskStatus.FINISHED,
```
vitorguidi: Can we change this to either
1) UTASK_MAIN_FINISHED
2) POSTPROCESS_STARTED


It is hard to tell this apart from postprocess_finished, which is the actual condition for the pre -> main > post workflow to finish.

vitorguidi (approved):

src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -75,10 +75,19 @@ class TaskStage:
 class TaskStatus:
   """Task status."""
   STARTED = 'started'
-  FINISHED = 'finished'
+  MAIN_FINISHED = 'main_finished'
```
vitorguidi, line 78: Is this being used?

ViniciustCosta: Sorry, I merged this before seeing this comment. It is not being used, I'll remove in a further PR

PR #4863, opened 2025-07-10
fix: implement check and parse for namedtuple on truncate

javanlacerda:
We had an error when trying to truncate a NamedTuple, and the cause is that we are trying to deal with it as a `tuple` as tuple is a supertype of namedtuple.

The fix for it is having a checking line that deals with namedtuple properly. And the way to indentify it is checking if the type is tuple, and if the object has the attribute `_fields`.

### DIFF
--- a/src/clusterfuzz/_internal/metrics/logs.py
+++ b/src/clusterfuzz/_internal/metrics/logs.py
@@ -219,6 +219,9 @@ def truncate(
   if dataclasses.is_dataclass(msg) and not isinstance(msg, type):
     msg = dataclasses.asdict(msg)
 
+  if isinstance(msg, tuple) and hasattr(msg, '_fields'):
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -219,6 +219,9 @@ def truncate(
   if dataclasses.is_dataclass(msg) and not isinstance(msg, type):
     msg = dataclasses.asdict(msg)
 
+  if isinstance(msg, tuple) and hasattr(msg, '_fields'):
```
ViniciustCosta, line 222: Nit: Maybe check if it has the `_asdict()` method instead of `_fields` to be more generic?

ViniciustCosta (approved):

PR #4865, opened 2025-07-14
[Tracing] Create Testcase Fixed events definition

ViniciustCosta:
This PR creates the `TestcaseFixedEvent` definition, which happens when a testcase is verified as fixed against the latest build. Instrumenting the code to emit the event in the correct places will be done in a further PR.

b/394055138


javanlacerda (approved):
LGTM!

vitaliset (approved):

PR #4876, opened 2025-07-22
[Tracing] Add issue closing events to cleanup

ViniciustCosta:
This PR instruments the cleanup code to emit the `IssueClosingEvent`. These events will be emitted every time an issue linked to a testcase is marked as closed/verified by ClusterFuzz. There are three occurrences of this:
* Mark an issue fixed if all of its associated reproducible testcases are fixed.
  * Even if the job sets skip auto close issue, clusterfuzz still marks it as verified and the event is emitted.
* Mark issue as closed if testcase is unreproducible after a certain deadline.
  * In this case, we are also emitting a `TestcaseRejectionEvent`, since the testcase is also marked as invalid during the cleanup. 
* Mark issue as wontfix if testcase is invalid (testcase.fixed='NA')

**Tests**
* Added unit tests to assert that the events are correctly emitted during cleanup.
* Since the development/staging environments do not run cleanup/triage crons, it might be needed to run one time against production to verify the expected behavior.

b/394060579

### DIFF
--- a/src/clusterfuzz/_internal/cron/cleanup.py
+++ b/src/clusterfuzz/_internal/cron/cleanup.py
@@ -82,6 +82,17 @@ def _append_generic_incorrect_comment(comment, policy, issue, suffix):
       label_text=issue.issue_tracker.label_text(wrong_label)) + suffix
 
 
+def _emit_issue_closing_event(testcase, issue, closing_reason):
+  """Helper function to emit issue closing events."""
+  issue_tracker_name = data_handler.get_issue_tracker_name(testcase.job_type)
+  events.emit(
+      events.IssueClosingEvent(
+          testcase=testcase,
+          issue_tracker_project=issue_tracker_name,
+          issue_id=str(issue.id) if issue.id else None,
+          closing_reason=closing_reason))
+
+
 def job_platform_to_real_platform(job_platform):
   """Get real platform from job platform."""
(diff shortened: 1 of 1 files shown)


vitaliset (approved):

src/clusterfuzz/_internal/cron/cleanup.py
```diff
@@ -699,6 +713,13 @@ def mark_na_testcase_issues_as_wontfix(policy, testcase, issue):
 
   logs.info(
       f'Closing issue {issue.id} for invalid testcase {testcase.key.id()}.')
+  issue_tracker_name = data_handler.get_issue_tracker_name(testcase.job_type)
+  events.emit(
+      events.IssueClosingEvent(
+          testcase=testcase,
+          issue_tracker_project=issue_tracker_name,
+          issue_id=str(issue.id) if issue.id else None,
+          closing_reason=events.ClosingReason.TESTCASE_INVALID))
```
javanlacerda: You've repeated this code block several times, you think you could move it to a function that receives the `testcase`, the `issue` and the `closing_reason` as arguments? 

Then you could just call it like `emit_closing_issue_event(testcase, issue, close_reason)`

ViniciustCosta: Sure, I can add that. Thanks!

PR #4880, opened 2025-07-28
Fix async downloads

jonathanmetzman:
We were accidentally doing `.content.content` instead of `.content`.
This was particularly insidious because we catch all exceptions.

### DIFF
--- a/src/clusterfuzz/_internal/system/fast_http.py
+++ b/src/clusterfuzz/_internal/system/fast_http.py
@@ -72,6 +72,7 @@ async def _error_tolerant_download_file(session: aiohttp.ClientSession,
                                         url: str, path: str) -> bool:
   """Downloads a file asynchronosly, returns False on failure."""
   try:
+    # TODO(metzman): Catch specific exceptions.
     await _async_download_file(session, url, path)
     return True
   except:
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
For now this change is NFC because the feature is disabled.

javanlacerda:
Thanks for this! Could you provide any evidence that it's working now?

jonathanmetzman:
Evidence that the whole workflow works or that this fix (and the functions touched here) work?

jonathanmetzman:
I can confirm corpus downloading works now

javanlacerda:
> Evidence that the whole workflow works or that this fix (and the functions touched here) work?

I think that just a way to run this function and being able to download a corpus is enough.

jonathanmetzman:
> > Evidence that the whole workflow works or that this fix (and the functions touched here) work?
> 
> I think that just a way to run this function and being able to download a corpus is enough.

We have both now!

jonathanmetzman:
> lgtm. just for curiosity, how did you manage to test locally?

I ran fuzz task, and verified that it downloaded more testcases than were in the backup.

jonathanmetzman:
Maybe I'll run corpus pruning to be sure.

PR #4881, opened 2025-07-28
[Fix] Add explicit execution permission for root dir

ViniciustCosta:
After the analysis done on b/433928503, we noticed that for some machines, even though clusterfuzz is the owner of the root directory, it is created without execution permissions (needed to re-create some directories during clusterfuzz execution).

This PR tries to add this permission explicitly in order to avoid hitting these errors on different versions of the COS.

### DIFF
--- a/docker/base/setup_clusterfuzz.sh
+++ b/docker/base/setup_clusterfuzz.sh
@@ -58,3 +58,4 @@ if [[ -z "$DISABLE_MOUNTS" ]]; then
 fi
 
 chown -R $USER:$USER $INSTALL_DIRECTORY
+chmod u+x $ROOT_DIR


hunsche (approved):

ViniciustCosta:
We should test this with a custom image before merging it.

hunsche:
The PR works.

```bash
clusterfuzz-linux-zkh2$ namei -om $INSTALL_DIRECTORY
f: /mnt/scratch0
 drwxr-xr-x root        root        /
 drwxr-xr-x root        root        mnt
 drwxr-xr-x clusterfuzz clusterfuzz scratch0
```

ViniciustCosta:
> The PR works.
> 
> ```bash
> clusterfuzz-linux-zkh2$ namei -om $INSTALL_DIRECTORY
> f: /mnt/scratch0
>  drwxr-xr-x root        root        /
>  drwxr-xr-x root        root        mnt
>  drwxr-xr-x clusterfuzz clusterfuzz scratch0
> ```

I think we need to check for the $ROOT_DIR (/mnt/scratch0/clusterfuzz)

jonathanmetzman:
I'm concerned this won't fix this specific problem (though it might be useful).
I think the problem happens when ClusterFuzz is running unrar_fuzzer, unrar changes the permissions, and then ClusterFuzz tries to execute over and over again but fails because it doesnt have permission to do any file operations. I don't think this script will get a chance to run.

ViniciustCosta:
Closing this as it is not needed per the investigation done in b/433928503 and https://github.com/google/clusterfuzz/pull/4881#issuecomment-3127954295.

PR #4883, opened 2025-07-29
Allow issue filing without impact task when regression fails.

jonathanmetzman:
Regression task will fail in a number of expected scenarios such as when a bug is temporarily fixed and then broken again in a later revision. Out-of-memory and timeout are particularly vulnerable to this.

There are a number of other issues why regression task can fail. Rather than hide bugs where this happens, we should report them.

### DIFF
--- a/src/clusterfuzz/_internal/datastore/data_handler.py
+++ b/src/clusterfuzz/_internal/datastore/data_handler.py
@@ -920,8 +920,10 @@ def critical_tasks_completed(testcase):
   if not utils.is_chromium():
     return testcase.minimized_keys and testcase.regression
 
+  impact_satisfied = testcase.is_impact_set_flag or testcase.regression == 'NA'
+
   return bool(testcase.minimized_keys and testcase.regression and
-              testcase.is_impact_set_flag and not testcase.analyze_pending)
+              impact_satisfied and not testcase.analyze_pending)
 
 
 # ------------------------------------------------------------------------------


jonathanmetzman:
I was able to almost confirm this WAI on https://clusterfuzz.com/testcase-detail/6295476342030336 but I can't run issue filing locally.
There are no examples of testcases like this in staging, so I'm not so convinced it's worthwhile to deploy there first.
This change is expected to have a big impact on prod, it should cause tens, if not hundreds of (stuck) testcases to be filed.

src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -920,8 +920,12 @@ def critical_tasks_completed(testcase):
   if not utils.is_chromium():
     return testcase.minimized_keys and testcase.regression
 
+  # Only require impact if it is feasible to run it. Don't require it if not
+  # (when regression fails).
+  impact_satisfied = testcase.is_impact_set_flag or testcase.regression == 'NA'
```
ViniciustCosta, line 925: I think there's still a failure state where regression times out many times or never completes and `testcase.regression` is None, but this should be treated in the regression code itself.

ViniciustCosta (approved):
Lgmt, thanks! I think this is safe to land in prod as it is just re-defining what critical tasks pending means.

src/clusterfuzz/_internal/datastore/data_handler.py
jonathanmetzman: Yeah I think that's a seperate fix.

PR #4885, opened 2025-07-30
[Grouper] Close testcases instead of deleting them and add project level configs

ViniciustCosta:
This PR patches the grouper code in order to mark testcases as closed instead of deleting them when groups overflow. This is intended to avoid data loss when wrong grouping happens.

In addition, this also adds configuration to allow project level definitions regarding:
* Max group size
* Crash comparer similarity threshold
* Crash comparer number of equal frames threshold

b/434148261

### DIFF
--- a/src/clusterfuzz/_internal/cron/grouper.py
+++ b/src/clusterfuzz/_internal/cron/grouper.py
@@ -32,8 +32,6 @@ FORWARDED_ATTRIBUTES = ('crash_state', 'crash_type', 'group_id',
                         'one_time_crasher_flag', 'project_name',
                         'security_flag', 'timestamp', 'job_type', 'fuzzer_name')
 
-GROUP_MAX_TESTCASE_LIMIT = 25
-
 VARIANT_CRASHES_IGNORE = re.compile(
     r'^(Out-of-memory|Timeout|Missing-library|Data race|GPU failure)')
 
@@ -44,6 +42,13 @@ VARIANT_MIN_THRESHOLD = 5
 VARIANT_MAX_THRESHOLD = 10
 
 TOP_CRASHES_LIMIT = 10
+
+GROUP_MAX_TESTCASE_LIMIT = local_config.ProjectConfig().get(
+    'deduplication.group_max_limit', 25)
+CRASH_COMPARER_THRESHOLD = local_config.ProjectConfig().get(
+    'deduplication.crash_comparer_threshold')
+CRASH_COMPARER_SAME_FRAMES = local_config.ProjectConfig().get(
+    'deduplication.crash_comparer_same_frames')
 
 
 class TestcaseAttributes:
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
Maybe the threshold should be set higher. Otherwise I think we will need to add some monitoring to prevent explosions in testcases. WDYT?

jonathanmetzman:
Do you want to add a TODO to reopen testcases when the leader is closed? Maybe when this happens if nothing is reopened the ones that would have been deleted can finally be garbage collected. WDYT?

src/clusterfuzz/_internal/cron/grouper.py
```diff
@@ -413,15 +426,16 @@ def _key_func(testcase):
...
-        logs.warning(
-            ('Deleting testcase {testcase_id} due to overflowing group '
-             '{group_id}.').format(
-                 testcase_id=testcase.id, group_id=testcase.group_id))
+        logs.warning(f'Closing testcase {testcase.id} due to overflowing group '
+                     f'{testcase.group_id}.')
         events.emit(
             events.TestcaseRejectionEvent(
                 testcase=testcase_entity,
                 rejection_reason=events.RejectionReason.GROUPER_OVERFLOW))
```
vitaliset, line 459: Should we update the event type to [IssueClosingEvent](https://github.com/google/clusterfuzz/blob/cd1de2e3742dc855706fb3b1b208221c194ec4cf/src/clusterfuzz/_internal/metrics/events.py#L235-L243)?

vitaliset: And maybe have the `TestcaseRejectionEvent` being emitted during the "garbage collector TODO" Jonathan mentioned?

src/clusterfuzz/_internal/cron/grouper.py
```diff
@@ -413,15 +426,16 @@ def _key_func(testcase):
         if testcase_entity.bug_information:
           continue
 
-        logs.warning(
-            ('Deleting testcase {testcase_id} due to overflowing group '
-             '{group_id}.').format(
-                 testcase_id=testcase.id, group_id=testcase.group_id))
+        logs.warning(f'Closing testcase {testcase.id} due to overflowing group '
```
jonathanmetzman: Can we make a metric out of this?

ViniciustCosta:
> Maybe the threshold should be set higher. Otherwise I think we will need to add some monitoring to prevent explosions in testcases. WDYT?

I assume you are referring to the group max size threshold. What do you mean by explosions in testcases? Too many testcases stored in datastore? If this is the case, increasing the threshold won't help if we land this change :/ 

I kept the default threshold of 25 and we can configure this in the project config. I could also add a config to choose whether to delete or close testcases, so we can apply this change only for Chrome (another idea is having a second higher threshold to start deleting them). Wdyt?

> Do you want to add a TODO to reopen testcases when the leader is closed? Maybe when this happens if nothing is reopened the ones that would have been deleted can finally be garbage collected. WDYT?

Yes, sure! I think this should be done in another PR, but I'll add a TODO.

src/clusterfuzz/_internal/cron/grouper.py
ViniciustCosta: Not actually, this does not close any issue (in fact, testcases are only rejected here if they do not have a linked issue per: [grouper.py#L413](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/cron/grouper.py#L413)).

Issues are only closed during cleanup which is already instrumented with events. The testcase rejection events, as we implemented by now, also applies to this case where the testcase is marked as closed and invalid (`open=False, fixed=NA`), which is a recurrent way in the codebase to close/reject testcases.

jonathanmetzman:
>I assume you are referring to the group max size threshold. What do you mean by explosions in testcases? Too many testcases stored in datastore? If this is the case, increasing the threshold won't help if we land this change :/

Yes and yes, too many testcases. Why won't it help? It would save more testcases but not allow infinite testcases (that were caused by the same issue).

Your solution of project specific configs is good IMO

src/clusterfuzz/_internal/cron/grouper.py
ViniciustCosta: Good idea, I will add it

jonathanmetzman:
I'm still worried about this creating too many testcases.

ViniciustCosta:
> I'm still worried about this creating too many testcases.

We can continue discussing about this offline, I won't merge it for now :)

ViniciustCosta:
I will try to test the metrics in dev/staging before prod (maybe we can create a grouper cron that runs on staging/dev, since grouper does not open/close issues and it would increase our testing capabilities).

PR #4890, opened 2025-08-01
[batch] Support trials.

jonathanmetzman:
The old method requires db access in main, which is no longer allowed.
Trials are a feature allowing experimental flags to be used in certain chrome and v8 builds.
CC @mi-ac

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/trials.py
+++ b/src/clusterfuzz/_internal/bot/tasks/trials.py
@@ -16,10 +16,9 @@
 import json
 import os
 import random
-from typing import List
 
-from clusterfuzz._internal.bot.tasks.utasks import uworker_io
 from clusterfuzz._internal.base import utils
+from clusterfuzz._internal.bot.tasks.utasks import uworker_io
 from clusterfuzz._internal.datastore import data_types
 from clusterfuzz._internal.metrics import logs
 from clusterfuzz._internal.system import environment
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/bot/tasks/trials.py
```diff
@@ -16,6 +16,7 @@
 import json
 import os
 import random
+from typing import List
```
ViniciustCosta: Nit: from Python 3.9+, it's preferred to use built-in types instead of importing from typing (go/pystyle#typing-imports) and https://peps.python.org/pep-0585/.

jonathanmetzman:
> Could you provide more context? Is there any way to test it locally, or using development environment before deploying it in production ?

I added more context.
I think the unittests are validation, there's a lot of them.

src/clusterfuzz/_internal/bot/tasks/trials.py
```diff
@@ -32,25 +33,40 @@ def __init__(self, probability, contradicts=None):
...
+  for extension in extensions_to_strip:
+    app_name = utils.strip_from_right(app_name, extension)
+  return app_name
+
+
 class Trials:
   """Helper class for selecting app-specific extra flags."""
 
-  def __init__(self):
+  def __init__(self, db_trials):
```
ViniciustCosta, line 62: Nit: Maybe add a type hint for `db_trials`.

ViniciustCosta (approved):
lgtm with nits

src/clusterfuzz/_internal/bot/tasks/trials.py
jonathanmetzman: I think doing this made our linter or typechecker explode but I can try again.

ViniciustCosta: I did something similar here: https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/metrics/events.py#L492 and it seems to work. Not a blocker though, just nit :)

PR #4893, opened 2025-08-06
[chrome] Fix handling of queued fuzz tasks

jonathanmetzman:
Chrome clusterfuzz is weird. The schedule_fuzz workload runs in a different project than the queues. Handle this.


hunsche (approved):

ViniciustCosta (approved):
Lgtm (please fix lint).

Also, could you confirm that this WAI for the other projects besides chrome as well?

PR #4895, opened 2025-08-07
Adding is_batch label for fuzzing time metric

javanlacerda:
It adds the label is_batch for the fuzzing time metric.  It will split the metric by environment helping us to indentify environment issues.


jonathanmetzman (approved):
LGTM

PR #4904, opened 2025-08-08
Non linux utask queueing

jonathanmetzman:


src/clusterfuzz/_internal/base/tasks/__init__.py
```diff
@@ -710,12 +710,23 @@ def run(self):
 def add_utask_main(command, input_url, job_type, wait_time=None):
   """Adds the utask_main portion of a utask to the utasks queue for scheduling
   on batch. This should only be done after preprocessing."""
+  # The platform of the job is loaded into the environment when this is called.
+  platform = environment.get_value('PLATFORM')
+
+  if platform.upper() == 'LINUX':
+    queue = UTASK_MAIN_QUEUE
+  else:
+    thread_multiplier = environment.get_value('THREAD_MULTIPLIER')
```
javanlacerda, line 719: Is that still used?

javanlacerda (approved):
LGTM. Could we deploy it in staging before?

PR #4908, opened 2025-08-12
Create cloudbuild.yaml for deployment pipeline

javanlacerda:
fix b/437926994

Implement a pipeline for running deployment through butler deploy script.

The pipeline steps are:

1. Make git clone on clusterfuzz config for a given revision if the revision is provided as a `substituition`. Substituition is the GCB way to provide a kind of env vars. [documentation](https://cloud.google.com/build/docs/configuring-builds/substitute-variable-values)
If the substituition key is not provided, the latest revision is used. 

2. Install deps and run the butler deploy for a also given revision as substituition. If the revision is not provided, the lastest is used. These step expected to receive an _PROJECT substituition that points to the project that should run the deployment, eg: staging, dev, internal, external...
The way to assure that the pipeline will run against the right environment is on the trigger definition, where the substituition can be hardcoded in the trigger definition. We will have a different trigger for each environment. 

Evidence of a successful run [here](https://github.com/google/clusterfuzz/runs/47932159422)

### DIFF
--- a/deploy/cloudbuild.yaml
+++ b/deploy/cloudbuild.yaml
@@ -1,0 +1,76 @@
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+steps:
+- name: gcr.io/cloud-builders/git
+  entrypoint: bash
+  args:
+  - -c
+  - |
+    # First, clone the repository
+    git config --global credential.helper gcloud.sh
+    git clone https://clusterfuzz-config-472119376969-git.us-central1.sourcemanager.dev/clusterfuzz-testing/clusterfuzz-config.git /workspace/clusterfuzz-config
+    
+    # Move into the repository directory
+    cd /workspace/clusterfuzz-config
+    
+    # Conditionally run 'git reset'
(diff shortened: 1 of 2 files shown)


hunsche (approved):

deploy/cloudbuild.yaml
```diff
@@ -0,0 +1,76 @@
+# Copyright 2019 Google LLC
```
ViniciustCosta: Please, update the date.

deploy/cloudbuild.yaml
```diff
@@ -0,0 +1,76 @@
...
+    configpath="/workspace/clusterfuzz-config/configs/${_PROJECT}"
+
+    # Verify that the configuration directory actually exists.
+    if [ ! -d "$configpath" ]; then
+      echo "❌ Error: Configuration directory not found at $configpath"
+      exit 1
+    fi
+
+    echo "✅ Running deploy for project config: $configpath"
+    python3.11 butler.py deploy -c $configpath --prod --targets=zips --force
```
ViniciustCosta: I might be off context here, but should we be able to also config the targets to deploy?

javanlacerda: You partially right, indeed the pipeline should be able to deploy the other targets, but not as a configuration. I'll add the appengine to be deployed with zips for all executions. Terraform will be deprecated soon, then I'll not add it.

ViniciustCosta: Thanks

deploy/cloudbuild.yaml
```diff
@@ -67,7 +67,7 @@ steps:
     fi
 
     echo "✅ Running deploy for project config: $configpath"
-    python3.11 butler.py deploy -c $configpath --prod --targets=zips --force
+    python3.11 butler.py deploy -c $configpath --prod --targets=zips,appengine --force
```
ViniciustCosta: Have you tried re-running with this change? I think this does not work after this PR: https://github.com/google/clusterfuzz/pull/4891. You probably need to do `--targets zips appengine`

javanlacerda: Fixed, thanks for catching it.

ViniciustCosta (approved):

PR #4909, opened 2025-08-13
Move label is batch from boolean to str.

javanlacerda:
It updates the label is batch to string because it uses the return of the function `is_uworker` as response, but this function doesn't return always boolean. For the case where it's not a batch environment, it returns None. The label is set as a string None for this case. 

That's the reason that some datapoints are missing on the dashboard. Before running the deployment for appengine, the new label was following the "default" pattern considering all as a String, but after the deployment, the descriptor has changed the label to Boolean, what stated to make the metric with `None` to be ignored.


hunsche (approved):

ViniciustCosta (approved):
lgtm! Just for the record, the intent of not using it as a boolean field (and just fixing where the metric is emitted) is to not lose the old data points.

PR #4913, opened 2025-08-14
Install git and clone the repository when revision is present

javanlacerda:
I've added a command to install git and clone the clusterfuzz repository if the clusterfuzz revision is passed as a substituition.

This is needed because GCB doesn't pass the content of the respository as a git project, then we're not able to change the version in the build time. 

The workaround for this is to clone the repository in build time and then checkout in the given revision.


ViniciustCosta (commented):
Sgtm! Could you provide an evidence of the pipeline working with this change?

javanlacerda:
image

I managed to trigger the pipeline manually for a older commit version of CF.

ViniciustCosta (approved):

PR #4918, opened 2025-08-21
Feature/gsutil gcloud feature flag

hunsche:
Migrate gsutil commands to gcloud behind a feature flag.

This change introduces a feature flag `USE_GCLOUD_STORAGE` to allow switching between the legacy `gsutil` commands and the new `gcloud storage` commands for file operations.

The following functions in `gsutil.py` are now feature-flagged:
- `download_file`
- `upload_file`
- `upload_files_to_url`

The unit tests for these functions have been parameterized to run against both the `gsutil` and `gcloud` implementations, ensuring consistent behavior.

This change also renames the `GSUTIL_PATH` environment variable to `GCLOUD_PATH` for consistency, and includes a minor fix to the `butler.py` script to use the correct python executable, which was necessary to get the test environment working.


src/clusterfuzz/_internal/google_cloud_utils/gsutil.py
```diff
@@ -124,38 +161,53 @@ def rsync(self,
     Returns:
       A bool indicating whether or not the command succeeded.
     """
-    # Use 'gsutil -m rsync -r' to download files from GCS bucket.
-    sync_corpus_command = ['rsync', '-r']
+    # Use 'gcloud storage rsync' to download files from GCS bucket.
+    sync_corpus_command = ['storage', 'rsync', '--recursive']
     if delete:
-      sync_corpus_command.append('-d')
+      sync_corpus_command.append('--delete-unmatched-destination-objects')
```
sushantkhanna, line 167: Just wanted to bring to your attention that Gcloud does not by default follow the same behaviour for this flag in comparison to `gsutil`. We have however made some changes, currently guarded by a flag which attempts to equate that.  
We are working on enabling it by default, which should happen shortly, but if we want to enable the flag manually in the meanwhile, we can help you with that.

hunsche:
Closing this PR due to conflicts. I'll open a new one.

https://github.com/google/clusterfuzz/pull/4923

PR #4919, opened 2025-08-22
Add backend structure for event retrieval

decoNR:
This PR introduces the core backend functionality for retrieving events from the datastore.
Key changes:

- Add `get_entities()` to enable querying datastore entities with support for equality filters, ordering, and limits.
- Add `NDBEventRepository.get_events()`  to retrieve events from the datastore.
- Add `get_events()`, a public-facing function for other modules to easily and consistently retrieve events.

Related to [b/435283944](http://b/435283944).

### DIFF
--- a/src/clusterfuzz/_internal/datastore/data_handler.py
+++ b/src/clusterfuzz/_internal/datastore/data_handler.py
@@ -19,6 +19,7 @@ import os
 import re
 import shlex
 import time
+from typing import Any
 
 from google.cloud import ndb
 
@@ -1570,6 +1571,38 @@ def get_entity_by_type_and_id(entity_type, entity_id):
     return None
 
   return entity_type.get_by_id(int(entity_id))
+
+
+@retry.wrap(
+    retries=DEFAULT_FAIL_RETRIES,
+    delay=DEFAULT_FAIL_WAIT,
+    function='datastore.data_handler.get_entities')
+def get_entities(entity_kind: data_types.Model,
+                 equality_filters: dict[str, Any] | None = None,
+                 order_by: list[str] | None = None,
+                 limit: int | None = None) -> list[data_types.Model]:
+  """Queries the entity kind with equality filters, ordering, and a limit."""
+  query = entity_kind.query()
+
+  if equality_filters:
+    for property_name, value in equality_filters.items():
+      if (prop := getattr(entity_kind, property_name, None)) is None:
+        logs.warning(
+            'Query filters contain a non-existent property: %s' % property_name)
+        continue
+      query = query.filter(prop == value)
+
+  if order_by:
+    for order_field in order_by:
+      desc = order_field.startswith('-')
+      property_name = order_field[1:] if desc else order_field
+      if (prop := getattr(entity_kind, property_name, None)) is None:
+        logs.warning('Query order argument contains a non-existent property: %s'
+                     % property_name)
+        continue
+      query = query.order(-prop) if desc else query.order(prop)
+
+  return query.fetch(limit=limit)
 
 
 # ------------------------------------------------------------------------------
--- a/src/clusterfuzz/_internal/metrics/events.py
+++ b/src/clusterfuzz/_internal/metrics/events.py
@@ -284,6 +284,14 @@ class IEventRepository(ABC):
                 event_type: str | None = None) -> Event | None:
     """Retrieve an event from the underlying database and return it."""
 
+  @abstractmethod
+  def get_events(self,
+                 equality_filters: dict[str, Any] | None = None,
+                 order_by: list[str] | None = None,
+                 limit: int | None = None) -> list[Event]:
+    """Retrieve a list of events from the underlying database. The events should
+    match the specified equality filters, ordering, and limit."""
+
 
 class NDBEventRepository(IEventRepository, EventHandler):
   """Implements the event repository for Datastore.
@@ -355,6 +363,24 @@ class NDBEventRepository(IEventRepository, EventHandler):
 
     event = self._deserialize_event(event_entity)
     return event
+
+  def get_events(self,
+                 equality_filters: dict[str, Any] | None = None,
+                 order_by: list[str] | None = None,
+                 limit: int | None = None) -> list[Event]:
+    """Retrieve events from datastore matching the given equality filters,
+    ordering, and limit."""
+    entity_kind = self._default_entity
+    results = data_handler.get_entities(
+        entity_kind=entity_kind,
+        equality_filters=equality_filters,
+        order_by=order_by,
+        limit=limit)
+
+    return [
+        event for entity in results
+        if (event := self._deserialize_event(entity)) is not None
+    ]
 
   def emit(self, event: Event) -> Any:
     """Emit an event by persisting it to Datastore."""
@@ -517,3 +543,18 @@ def emit(event: Event) -> None:
 
   for handler in handlers:
     handler.emit(event)
+
+
+def get_events(equality_filters: dict[str, Any] | None = None,
+               order_by: list[str] | None = None,
+               limit: int | None = None) -> list[Event] | None:
+  """Retrieve events matching the given equality filters, ordering, and limit.
+  """
+  repository = get_repository()
+  if repository is None:
+    return None
+
+  events = repository.get_events(
+      equality_filters=equality_filters, order_by=order_by, limit=limit)
+
+  return events
--- a/src/clusterfuzz/_internal/tests/core/datastore/data_handler_test.py
+++ b/src/clusterfuzz/_internal/tests/core/datastore/data_handler_test.py
@@ -1025,3 +1025,91 @@ class TestTrustedVsUntrusted(unittest.TestCase):
         self.job.name, None, None, None, None, None, None, None,
         timeout_multiplier, None, True)
     self.assertTrue(data_handler.get_testcase_by_id(testcase_id).trusted)
+
+
+@test_utils.with_cloud_emulators('datastore')
+class GetEntitiesTest(unittest.TestCase):
+  """Tests for get_entities."""
+
+  class GetEntitiesTestModel(data_types.Model):
+    """Model for get_entities tests."""
+    name = ndb.StringProperty()
+    value = ndb.IntegerProperty()
+    timestamp = ndb.DateTimeProperty()
+
+  def setUp(self):
+    self.entity1 = self.GetEntitiesTestModel(
+        name='a', value=1, timestamp=datetime.datetime(2023, 1, 1, 0, 0, 1))
+    self.entity1.put()
+    self.entity2 = self.GetEntitiesTestModel(
+        name='b', value=2, timestamp=datetime.datetime(2023, 1, 1, 0, 0, 2))
+    self.entity2.put()
+    self.entity3 = self.GetEntitiesTestModel(
+        name='c', value=1, timestamp=datetime.datetime(2023, 1, 1, 0, 0, 3))
+    self.entity3.put()
+    self.entity4 = self.GetEntitiesTestModel(
+        name='d', value=3, timestamp=datetime.datetime(2023, 1, 1, 0, 0, 4))
+    self.entity4.put()
+
+  def test_get_all(self):
+    """Test getting all entities."""
+    result = data_handler.get_entities(self.GetEntitiesTestModel)
+    self.assertEqual(4, len(result))
+
+  def test_with_limit(self):
+    """Test with a limit."""
+    result = data_handler.get_entities(
+        self.GetEntitiesTestModel, limit=2, order_by=['timestamp'])
+    self.assertEqual(2, len(result))
+    expected_entities = [self.entity1, self.entity2]
+    self.assertTrue(test_utils.entities_list_equal(result, expected_entities))
+
+  def test_with_equality_filters(self):
+    """Test with equality filters."""
+    result = data_handler.get_entities(
+        self.GetEntitiesTestModel, equality_filters={'value': 1})
+    self.assertEqual(2, len(result))
+    expected_entities = [self.entity1, self.entity3]
+    self.assertTrue(test_utils.entities_list_equal(result, expected_entities))
+
+  def test_with_order_by_asc(self):
+    """Test with ordering (ascending)."""
+    result = data_handler.get_entities(
+        self.GetEntitiesTestModel, order_by=['value', 'name'])
+    expected_entities = [self.entity1, self.entity3, self.entity2, self.entity4]
+    self.assertTrue(test_utils.entities_list_equal(result, expected_entities))
+
+  def test_with_order_by_desc(self):
+    """Test with ordering (descending)."""
+    result = data_handler.get_entities(
+        self.GetEntitiesTestModel, order_by=['-name'])
+    expected_entities = [self.entity4, self.entity3, self.entity2, self.entity1]
+    self.assertTrue(test_utils.entities_list_equal(result, expected_entities))
+
+  def test_with_equality_filters_and_order(self):
+    """Test with equality filters and ordering."""
+    result = data_handler.get_entities(
+        self.GetEntitiesTestModel,
+        equality_filters={'value': 1},
+        order_by=['-name'])
+    self.assertEqual(2, len(result))
+    expected_entities = [self.entity3, self.entity1]
+    self.assertTrue(test_utils.entities_list_equal(result, expected_entities))
+
+  def test_non_existent_filter_property(self):
+    """Test filtering on a property that does not exist."""
+    result = data_handler.get_entities(
+        self.GetEntitiesTestModel, equality_filters={'non_existent': 1})
+    self.assertEqual(4, len(result))
+
+  def test_non_existent_order_property(self):
+    """Test ordering by a property that does not exist."""
+    result = data_handler.get_entities(
+        self.GetEntitiesTestModel, order_by=['non_existent'])
+    self.assertEqual(4, len(result))
+
+  def test_no_entites_found(self):
+    """Test no entities found."""
+    result = data_handler.get_entities(
+        self.GetEntitiesTestModel, equality_filters={'value': 5})
+    self.assertEqual([], result)
--- a/src/clusterfuzz/_internal/tests/core/metrics/events_test.py
+++ b/src/clusterfuzz/_internal/tests/core/metrics/events_test.py
@@ -16,6 +16,7 @@ import datetime
 import os
 import platform
 import unittest
+from unittest import mock
 
 from clusterfuzz._internal.datastore import data_handler
 from clusterfuzz._internal.datastore import data_types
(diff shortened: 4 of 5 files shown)


src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -1572,6 +1573,38 @@ def get_entity_by_type_and_id(entity_type, entity_id):
...
 
 
+@retry.wrap(
+    retries=DEFAULT_FAIL_RETRIES,
+    delay=DEFAULT_FAIL_WAIT,
+    function='datastore.data_handler.get_entities')
+def get_entities(entity_kind: data_types.Model,
+                 equality_filters: dict[str, Any] | None = None,
+                 order_by: list[str] | None = None,
+                 limit: int | None = None) -> list[data_types.Model]:
```
vitaliset: It's generally preferable to use "abstract types" for function argument annotations, such as `Sequence` and `Mapping`.
go/python-tips/044#type-annotating-function-arguments

This comment also applies to other functions from this PR. :)

src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -1572,6 +1573,38 @@ def get_entity_by_type_and_id(entity_type, entity_id):
...
+
+  if order_by:
+    for order_field in order_by:
+      desc = order_field.startswith('-')
+      property_name = order_field[1:] if desc else order_field
+      if (prop := getattr(entity_kind, property_name, None)) is None:
+        logs.warning('Query order argument contains a non-existent property: %s'
+                     % property_name)
+        continue
+      query = query.order(-prop) if desc else query.order(prop)
```
vitaliset: To improve readability and reduce nesting, consider extracting the logic for applying equality filters and the logic for applying ordering into separate helper functions.

This would make the main function's flow more apparent (setup, filter, order, fetch) and keep each function focused on a single responsibility.

go/tott/733

src/clusterfuzz/_internal/tests/core/datastore/data_handler_test.py
```diff
@@ -1025,3 +1025,91 @@ def test_fuzzer_created(self):
...
+        name='c', value=1, timestamp=datetime.datetime(2023, 1, 1, 0, 0, 3))
+    self.entity3.put()
+    self.entity4 = self.GetEntitiesTestModel(
+        name='d', value=3, timestamp=datetime.datetime(2023, 1, 1, 0, 0, 4))
+    self.entity4.put()
+
+  def test_get_all(self):
+    """Test getting all entities."""
+    result = data_handler.get_entities(self.GetEntitiesTestModel)
+    self.assertEqual(4, len(result))
```
vitaliset: nit: Maybe also check for `test_utils.entities_list_equal`?

src/clusterfuzz/_internal/tests/core/datastore/data_handler_test.py
```diff
@@ -1025,3 +1025,91 @@ def test_fuzzer_created(self):
...
+        self.GetEntitiesTestModel, equality_filters={'non_existent': 1})
+    self.assertEqual(4, len(result))
+
+  def test_non_existent_order_property(self):
+    """Test ordering by a property that does not exist."""
+    result = data_handler.get_entities(
+        self.GetEntitiesTestModel, order_by=['non_existent'])
+    self.assertEqual(4, len(result))
+
+  def test_no_entites_found(self):
```
vitaliset:
```suggestion
  def test_no_entities_found(self):
```

src/clusterfuzz/_internal/tests/core/datastore/data_handler_test.py
```diff
@@ -1025,3 +1025,91 @@ def test_fuzzer_created(self):
...
+        self.GetEntitiesTestModel, order_by=['value', 'name'])
+    expected_entities = [self.entity1, self.entity3, self.entity2, self.entity4]
+    self.assertTrue(test_utils.entities_list_equal(result, expected_entities))
+
+  def test_with_order_by_desc(self):
+    """Test with ordering (descending)."""
+    result = data_handler.get_entities(
+        self.GetEntitiesTestModel, order_by=['-name'])
+    expected_entities = [self.entity4, self.entity3, self.entity2, self.entity1]
+    self.assertTrue(test_utils.entities_list_equal(result, expected_entities))
```
vitaliset: Pretty similar structure. I'm okay with this, but could be a place to use `parameterized`.

go/python-tips/011#repetitive-tests

src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -284,6 +284,14 @@ def get_event(self, event_id: str | int,
                 event_type: str | None = None) -> Event | None:
     """Retrieve an event from the underlying database and return it."""
 
+  @abstractmethod
+  def get_events(self,
+                 equality_filters: dict[str, Any] | None = None,
+                 order_by: list[str] | None = None,
+                 limit: int | None = None) -> list[Event]:
+    """Retrieve a list of events from the underlying database. The events should
+    match the specified equality filters, ordering, and limit."""
```
vitaliset: 2 lines docstring don't match go/pystyle#function-docs

Also applies to `get_events` functions.

src/clusterfuzz/_internal/tests/core/datastore/data_handler_test.py
```diff
@@ -1025,3 +1025,91 @@ def test_fuzzer_created(self):
...
+        name='d', value=3, timestamp=datetime.datetime(2023, 1, 1, 0, 0, 4))
+    self.entity4.put()
+
+  def test_get_all(self):
+    """Test getting all entities."""
+    result = data_handler.get_entities(self.GetEntitiesTestModel)
+    self.assertEqual(4, len(result))
+
+  def test_with_limit(self):
+    """Test with a limit."""
```
vitaliset: Docstrings that do not provide any new information should not be used.
go/pystyle#test-docs

This might also be useful go/python-readability-advice#test_docstrings

src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -19,6 +19,7 @@
 import re
 import shlex
 import time
+from typing import Any
```
vitaliset: As a rule of thumb, I suggest avoid using `typing.Any` whenever a more specific type can be used.

src/clusterfuzz/_internal/tests/test_libs/test_utils.py
```diff
@@ -89,6 +89,19 @@ def entities_equal(entity_1, entity_2, check_key=True):
...
+                        entities_2: list[data_types.Model],
+                        check_key: bool = True) -> bool:
+  """Return a bool on whether every pair of entities are the same."""
+  if len(entities_1) != len(entities_2):
+    return False
+
+  for entity_1, entity_2 in zip(entities_1, entities_2):
+    if not entities_equal(entity_1, entity_2, check_key):
+      return False
+  return True
```
vitaliset: nit nit: Consider using `all(entities_equal(entity_1,... for zip(...`. Slightly more Pythonic.

src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -1572,6 +1573,38 @@ def get_entity_by_type_and_id(entity_type, entity_id):
...
+                 equality_filters: dict[str, Any] | None = None,
+                 order_by: list[str] | None = None,
+                 limit: int | None = None) -> list[data_types.Model]:
+  """Queries the entity kind with equality filters, ordering, and a limit."""
+  query = entity_kind.query()
+
+  if equality_filters:
+    for property_name, value in equality_filters.items():
+      if (prop := getattr(entity_kind, property_name, None)) is None:
+        logs.warning(
```
ViniciustCosta: I think it is better to use f-strings here (ditto for the other logs in this PR).

src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -1572,6 +1573,38 @@ def get_entity_by_type_and_id(entity_type, entity_id):
   return entity_type.get_by_id(int(entity_id))
 
 
+@retry.wrap(
+    retries=DEFAULT_FAIL_RETRIES,
+    delay=DEFAULT_FAIL_WAIT,
+    function='datastore.data_handler.get_entities')
+def get_entities(entity_kind: data_types.Model,
```
ViniciustCosta: What do you think about having the option to return an iterator instead of fetch for more efficient loop, if needed?

Per the NDB documentation: [...] there’s an idiom to efficiently loop over the Query results in a tasklet, properly yielding when appropriate:
```
it = query1.iter()
while (yield it.has_next_async()):
    emp = it.next()
    print(emp.name, emp.age)
```

ViniciustCosta: Also, have the option to return only the entity keys (using the `keys_only` argument)

src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -517,3 +543,18 @@ def emit(event: Event) -> None:
 
   for handler in handlers:
     handler.emit(event)
+
+
+def get_events(equality_filters: dict[str, Any] | None = None,
```
ViniciustCosta: I would suggest adding some additional more specific functions to facilitate usage (these can call the `get_events` method, but only show a more comprehensive interface to external modules).

For instance, you could have a `get_events_from_testcase` that already expects some very common fields, such as: `event_type`, `testcase_id`, `task_name`, etc. Thus avoiding having to replicate code and also make it easier for other folks  to use them.

src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -356,6 +364,24 @@ def get_event(self, event_id: str | int,
     event = self._deserialize_event(event_entity)
     return event
 
+  def get_events(self,
+                 equality_filters: dict[str, Any] | None = None,
+                 order_by: list[str] | None = None,
+                 limit: int | None = None) -> list[Event]:
+    """Retrieve events from datastore matching the given equality filters,
+    ordering, and limit."""
+    entity_kind = self._default_entity
```
ViniciustCosta: I'm trying to think of a workaround to maintain the idea of a generic ndb entity, since this class defines the `_event_to_entity_map`, which should allow different entity kinds based on event type.

The method could take a specific optional argument (`event_type`) or you could try to do `equality_filters.get('event_type')` and if it is defined, you use the mapping, otherwise you use the default and inform the user. 

The worst case we remove this logic of generic entity and force only one entity kind.

Wdyt?

src/clusterfuzz/_internal/datastore/data_handler.py
decoNR: I updated it. Thank you for the reference! I left the return type as it is based on the [NDB documentation](https://googleapis.dev/python/python-ndb/latest/query.html) regarding the `fetch` function.

src/clusterfuzz/_internal/datastore/data_handler.py
decoNR: Nice, I changed it.

src/clusterfuzz/_internal/datastore/data_handler.py
decoNR: Done.

src/clusterfuzz/_internal/tests/test_libs/test_utils.py
decoNR: Done.

src/clusterfuzz/_internal/datastore/data_handler.py
decoNR: Done.

src/clusterfuzz/_internal/tests/core/datastore/data_handler_test.py
decoNR: I updated the code using `unittest.Testcase` `assertCountEqual` in order to not depend on the order.

src/clusterfuzz/_internal/tests/core/datastore/data_handler_test.py
decoNR: Done.

src/clusterfuzz/_internal/metrics/events.py
decoNR: Done.

src/clusterfuzz/_internal/tests/core/datastore/data_handler_test.py
decoNR: Thank you for the reference. I updated it.

src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -19,7 +19,9 @@
 import re
 import shlex
 import time
-from typing import Any
+from typing import Mapping
+from typing import Sequence
+from typing import TypeAlias
```
vitaliset, line 25: go/pystyle#imports-formatting

decoNR: I did not update it as lint was not passing when I put all imports in the same line.

src/clusterfuzz/_internal/datastore/data_handler.py
decoNR: As we discussed, I updated the events retrieval logic to return a generator instead of a list. Also, I created the function `data_handler.get_entities_ids` to retrieve only the entity` s ids.

src/clusterfuzz/_internal/metrics/events.py
decoNR: Thank you for the suggestion, I added the `get_latest_events_from_testcase` function. It has all arguments that you pointed out plus ordering by `-timestamp`, which I think makes sense as we are dealing with events.

src/clusterfuzz/_internal/metrics/events.py
decoNR: I updated the code with the second suggestion but can you explain what you mean by "inform the user"?

src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -544,3 +577,31 @@ def emit(event: Event) -> None:
 
   for handler in handlers:
     handler.emit(event)
+
+
+def get_events(equality_filters: Mapping[str, FilterValue] | None = None,
+               order_by: Sequence[str] | None = None) -> Generator:
+  """Yields events matching the equality filters and ordering."""
+  repository = get_repository()
+  if repository:
```
ViniciustCosta, line 588: Please, add a warning here to log the case where repository is not set/registered for the project.

src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -544,3 +577,31 @@ def emit(event: Event) -> None:
...
+    yield from repository.get_events(
+        equality_filters=equality_filters, order_by=order_by)
+
+
+def get_latest_events_from_testcase(testcase_id: int,
```
ViniciustCosta: What about only `get_events_from_testcase()` and add an argument for the timestamp being desc or asc or none (defaulting to desc as it is the most common)?

src/clusterfuzz/_internal/tests/core/metrics/events_test.py
```diff
@@ -922,6 +924,65 @@ def test_get_event(self):
     self.assertEqual(event.timestamp, date_now)
     self._assert_common_event_fields(event)
 
+  def _get_events_patch(self):
```
ViniciustCosta: Why do we need to patch these? Since we are testing the events in datastore using the cloud emulator, I think we should try to rely on the actual methods to have more reliable tests. Wdyt?

src/clusterfuzz/_internal/metrics/events.py
ViniciustCosta: Thanks! If, during the development, you find that any more methods that would be useful to facilitate querying from other module, feel free to add them.

src/clusterfuzz/_internal/metrics/events.py
ViniciustCosta: Inform the user by adding this information in the comment (like "if no event type is specified, the default entity will be used") and/or logging when this happens (logs.info("Defaulting to entity ... for event ...")

src/clusterfuzz/_internal/metrics/events.py
decoNR: When the repository is not set/registered, `get_repository()` logs a warning for that. Do you think it is needed to add another one?

src/clusterfuzz/_internal/tests/core/metrics/events_test.py
decoNR: I choose to use a patch in order to test only the behavior of `get_events` as the patched functions already have specific tests. Also, I was thinking about adding an integration test in the next PR for testing the entire flow of retrieving the information for the testcase status machine board. What do you think?

src/clusterfuzz/_internal/metrics/events.py
ViniciustCosta: You're right! I didn't see that log, no need to add another one

src/clusterfuzz/_internal/tests/core/metrics/events_test.py
ViniciustCosta: Even though you are making an unit test for the get_events behavior, I think it's still relevant to keep the behavior as close as possible from production. If the underlying methods change and break the expected behavior (because the get events depend on this) you wouldn't catch it with this test.

[go/tott/697](https://goto.google.com/tott/697)

src/clusterfuzz/_internal/metrics/events.py
decoNR: As we discussed, I updated the function to `get_events_from_testcase()` with a flag to sort in descending order by timestamp. Adding ascending ordering would need an extra database index and we do not think it is useful.

src/clusterfuzz/_internal/metrics/events.py
decoNR: Update it by adding this information in `get_events()` docstring.

src/clusterfuzz/_internal/tests/core/metrics/events_test.py
decoNR: Thank you for your input and the reference! I updated it.

src/clusterfuzz/_internal/metrics/events.py
```diff
@@ -544,3 +579,38 @@ def emit(event: Event) -> None:
...
+      'event_type': event_type,
+      'task_name': task_name,
+  }
+  equality_filters = {
+      filter: value
+      for filter, value in potential_filters.items()
+      if value is not None
+  }
+
+  order_by = None
```
ViniciustCosta: Nit: `order_by = ['-timestamp'] if latest else None`

src/clusterfuzz/_internal/tests/core/metrics/events_test.py
```diff
@@ -1158,3 +1159,177 @@ def test_emit_datastore_event(self):
...
+        timestamp=datetime.datetime(2025, 1, 1, 12, 0))
+    self.event5.put()
+
+  def _assert_events_equal(self, result_events, entities):
+    """Asserts that each event match the associated entity"""
+    self.assertEqual(len(result_events), len(entities))
+    for event, entity in zip(result_events, entities):
+      self.assertEqual(event.event_type, entity.event_type)
+      self.assertEqual(event.timestamp, entity.timestamp)
+
```
ViniciustCosta, line 1202: Nit: Maybe also assert the testcase_id?

ViniciustCosta (approved):
Lgtm! Great work on those unit tests!

(I've left some nits, but feel free to merge.)

src/clusterfuzz/_internal/tests/core/metrics/events_test.py
```diff
@@ -1158,3 +1159,177 @@ def test_emit_datastore_event(self):
     self.assertEqual(event_entity.fuzzer, 'fuzzer1')
     self.assertEqual(event_entity.job, 'test_content_shell_drt')
     self.assertEqual(event_entity.crash_revision, 1)
+
+
+class RetrievingEventsTest(unittest.TestCase):
+  """Test retrieving events."""
```
ViniciustCosta: Nit: Add in docstring that this is a helper/base class?

src/clusterfuzz/_internal/tests/core/metrics/events_test.py
decoNR: I didn't do that because most of the retrieved entities are deserialized to an instance of `Event`, which does not have the attribute `testcase_id`. In order to cover that scenario, the test `test_retrieve_specific_event_type` deserializes the entity to an instance of `TaskExecutionEvent` and uses `assertIsInstance`. I added a `testcase_id` assertion for this one.

src/clusterfuzz/_internal/tests/core/metrics/events_test.py
decoNR: I updated it, thank you!

src/clusterfuzz/_internal/metrics/events.py
decoNR: I updated it, thank you!

decoNR:
> Lgtm! Great work on those unit tests!
> 
> (I've left some nits, but feel free to merge.)

Thank you for the awesome inputs!

PR #4920, opened 2025-08-22
Create deployment error

javanlacerda:
Contribute towards b/440042223

It create a Deployment Error class and uses it for Deployment failure.

It also fixes the issuance of the metric for failures. It seems the way that the metrics client is implemented, it's not allowed to emit the metric in a exception handling block.

### DIFF
--- a/src/local/butler/deploy.py
+++ b/src/local/butler/deploy.py
@@ -51,6 +51,10 @@ SERVICE_REGEX = re.compile(r'service\s*:\s*(.*)')
 Version = namedtuple('Version', ['id', 'deploy_time', 'traffic_split'])
 
 
+class DeploymentError(Exception):
+  """Deployment Error"""
+
+
 def now(tz=None):
   """Used for mocks."""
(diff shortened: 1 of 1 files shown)


javanlacerda:
> Could you provide an evidence of the emitted metric? :)

Sure. [Here](https://pantheon.corp.google.com/monitoring/alerting/alerts/0.nw9lm1fmffvd?channelType=mail&project=clusterfuzz-development&e=-13802955&mods=logs_tg_prod) you can see the metric was sent if success false triggered by the pipeline.

src/local/butler/deploy.py
```diff
@@ -381,6 +381,10 @@ def _staging_deployment_helper():
   print('Staging deployment finished.')
 
 
+class DeploymentError(Exception):
```
ViniciustCosta: Maybe move the class definition to the beginning of the file to help with readability.

ViniciustCosta (approved):
lgtm!

PR #4924, opened 2025-08-25
MNT Fix path traversal issue in archive extraction

vitaliset:
The `_is_attempting_path_traversal` function did not properly handle absolute paths, allowing a maliciously crafted archive to write files outside of the intended extraction directory.

This patch fixes the issue by:
- Adding a check to explicitly disallow absolute paths in archive members.
- Improving the relative path traversal check to ensure that the resolved path of an archive member is always within the intended output directory.

b/431292024

### DIFF
--- a/src/clusterfuzz/_internal/tests/core/system/archive_test.py
+++ b/src/clusterfuzz/_internal/tests/core/system/archive_test.py
@@ -12,7 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """archive tests."""
+import io
 import os
+import tarfile
 import tempfile
 import unittest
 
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/tests/core/system/archive_test.py
```diff
@@ -51,6 +53,29 @@ def test_file_list(self):
           [f.name for f in reader.list_members()],
           ["archive_dir", "archive_dir/bye", "archive_dir/hi"])
 
+  def test_unpack_absolute_path_traversal(self):
```
vitaliset, line 56: Test run with previous `_is_attempting_path_traversal` version:

```
FAIL: test_unpack_absolute_path_traversal (tests.core.system.archive_test.UnpackTest.test_unpack_absolute_path_traversal)
Test that unpacking an archive with an absolute path fails.
  File "clusterfuzz/src/clusterfuzz/_internal/tests/core/system/archive_test.py", line 75, in test_unpack_absolute_path_traversal
    self.assertFalse(result)
AssertionError: True is not false

Ran 7 tests in 0.020s

FAILED (failures=1)
````

javanlacerda:
Teria como validar em dev?

decoNR (approved):

oliverchang (approved):

PR #4930, opened 2025-08-27
Refactor: `gsutil` to `gcloud storage` Migration with Per-Command Feature Flags

hunsche:
# Refactor: `gsutil` to `gcloud storage` Migration with Per-Command Feature Flags

## Summary

This pull request fundamentally refactors our `gsutil` to `gcloud storage` migration strategy. The previous approach used a single global feature flag, `USE_GCLOUD_STORAGE`, which activated the migration for all commands simultaneously. This PR replaces that global flag with a granular, per-command flag system, allowing for a safer, more controlled, and incremental transition.

The following environment variables have been introduced to individually control their corresponding commands:

  * `USE_GCLOUD_STORAGE_CP`
  * `USE_GCLOUD_STORAGE_RSYNC`
  * `USE_GCLOUD_STORAGE_LS`
  * `USE_GCLOUD_STORAGE_MB`
  * `USE_GCLOUD_STORAGE_CORS_SET`
  * `USE_GCLOUD_STORAGE_DEFSTORAGECLASS_SET`
  * `USE_GCLOUD_STORAGE_IAM_CH`

-----

## Motivation

The migration from `gsutil` to the more modern `gcloud storage` tools is an important step to modernize our codebase and leverage the performance and feature improvements of `gcloud`. However, a "big bang" migration with a single global flag posed a significant risk. A bug or performance issue in a single command could impact all storage operations, causing widespread instability.

This new approach mitigates that risk by allowing us to enable and test the migration for each command in isolation. This makes it easier to identify issues, roll back specific changes, and ensure a smooth, stable transition to the new tooling.

-----

## Implementation Details

The main technical changes include:

  * **Dynamic Flag Function:**

      * The global `use_gcloud_storage()` function has been removed.
      * A new function, `use_gcloud_for_command(command: str)`, has been introduced in `src/clusterfuzz/_internal/google_cloud_utils/gsutil.py`. This function dynamically constructs the environment variable name (e.g., `USE_GCLOUD_STORAGE_CP`) based on the provided command and checks if it is set.

  * **`GSUtilRunner` Refactoring:**

      * The `GSUtilRunner` in `gsutil.py` no longer relies on an internal state (`self.use_gcloud_storage`).
      * Its methods (`rsync`, `download_file`, etc.) now invoke `use_gcloud_for_command` with the relevant command to decide at runtime which CLI (`gsutil` or `gcloud storage`) to use.

  * **Direct `gsutil` Call Updates:**

      * Scripts that invoked `gsutil` directly, such as `src/local/butler/common.py` and `src/clusterfuzz/_internal/scripts/copy_corpus.py`, have been refactored to use the new feature flag logic.
      * Shell scripts in `configs/` and `docker/` have also been updated to check the new command-specific environment variables.

  * **User Interface:**

      * The upload instruction message in the `corpora` handler (`src/appengine/handlers/corpora.py`) now dynamically displays the `gsutil rsync` or `gcloud storage rsync` command, depending on the state of the `USE_GCLOUD_STORAGE_RSYNC` flag, ensuring users see the correct command.

-----

## Rollout Strategy and How to Use

Activating the new functionality is done via environment variables. By default, all commands will continue to use `gsutil`.

To enable `gcloud storage` for a specific command, set the corresponding environment variable to a truthy value (e.g., `"1"` or `"True"`).

**Example:** To test the `cp` command using `gcloud storage`, set the following environment variable:

```bash
export USE_GCLOUD_STORAGE_CP=1
```

To revert to `gsutil`, simply unset the environment variable or set it to an empty string.

### Recommended Rollout Strategy:

1.  **Staging Environment**: Start by enabling the flags one by one in a staging or testing environment.
2.  **Monitoring**: After enabling each flag, closely monitor logs, performance metrics, and related functionality to ensure there are no regressions.
3.  **Incremental Rollout**: We suggest the following order for a production rollout, starting with lower-risk commands:
      * `ls`
      * `mb`, `cors set`, `defstorageclass set`
      * `cp`
      * `rsync`
4.  **Production**: Once a command has been thoroughly tested and validated in staging, the corresponding flag can be enabled in production.
5.  **Migration Completion**: The ultimate goal is to have all flags enabled in production. After a stabilization period, we can plan a future refactoring to completely remove the legacy `gsutil` code and the feature flags themselves.

-----

## Tests

  * All existing unit tests in `gsutil_test.py` and `deploy_test.py` have been updated to use the new per-command feature flags.
  * Tests have been parameterized to ensure each command functions correctly with both `gsutil` (flag disabled) and `gcloud storage` (flag enabled).
  * All tests, linters, and formatters are passing, ensuring code quality and consistency.


src/Pipfile
```diff
@@ -53,3 +53,4 @@ Jinja2 = "==3.1.4"
 PyJWT = "==2.7.0"
 requests-toolbelt = "==0.9.1"
 werkzeug = "==2.2.2"
+parameterized = "*"
```
vitaliset: Not sure we want a dependency that's not pinned. Can you update this please?

src/local/butler/common.py
```diff
@@ -57,18 +58,49 @@ class Gsutil:
...
+      elif args[0] == 'iam' and args[1] == 'ch':
+        # gsutil iam ch <role>:<member> gs://<bucket>
+        # gcloud storage buckets add-iam-policy-binding gs://<bucket> --member=<member> --role=<role>
+        role, member = args[2].split(':')
+        arguments.extend(['buckets', 'add-iam-policy-binding', args[3], f'--member={member}', f'--role={role}'])
+      else:
+        arguments.extend(args)
+    else:
+      arguments = ['gsutil']
+      arguments.extend(args)
```
vitaliset, line 93: Not sure I understand why we need to do the updates in the command.

vitaliset (approved):
I've left a few small comments on a question and an update that I think should be addressed.

I did my best to understand all the changes, but I'm essentially rubber-stamping this to unblock you since it didn't break dev. :)

src/local/butler/common.py
hunsche: This function is a wrapper for gsutil. It has a compatibility layer to use `gcloud storage` when a certain environment variable is set. This allows for a gradual migration from gsutil to gcloud storage.

src/Pipfile
hunsche: Good catch, thank you. I've updated it to the pinned version.

PR #4934, opened 2025-08-29
[Logging] Add deleted testcase data to logs from grouper

ViniciustCosta:
Related to b/434148261

### DIFF
--- a/src/clusterfuzz/_internal/cron/grouper.py
+++ b/src/clusterfuzz/_internal/cron/grouper.py
@@ -97,19 +97,26 @@ def _emit_grouping_event(moved_testcase: int,
           group_merge_reason=group_merge_reason))
 
 
+def _get_testcase_log_info(testcase: TestcaseAttributes):
+  """Return testcase data as string for logging."""
+  log_str = (f'group={testcase.group_id}, '
+             f'crash_type={testcase.crash_type}, '
+             f'crash_state={testcase.crash_state}, '
+             f'security_flag={testcase.security_flag}, '
+             f'one_time_crasher={testcase.one_time_crasher_flag}, '
+             f'fuzzer={testcase.fuzzer_name}, '
+             f'job={testcase.job_type}')
+  return log_str
+
+
 def combine_testcases_into_group(
     testcase_1: TestcaseAttributes, testcase_2: TestcaseAttributes,
     testcase_map: dict[int, TestcaseAttributes], reason: str) -> None:
   """Combine two testcases into a group."""
-  logs.info(
-      'Grouping testcase %s '
-      '(crash_type=%s, crash_state=%s, security_flag=%s, group=%s) '
-      'and testcase %s '
-      '(crash_type=%s, crash_state=%s, security_flag=%s, group=%s). Reason: %s'
-      % (testcase_1.id, testcase_1.crash_type, testcase_1.crash_state,
-         testcase_1.security_flag, testcase_1.group_id, testcase_2.id,
-         testcase_2.crash_type, testcase_2.crash_state,
-         testcase_2.security_flag, testcase_2.group_id, reason))
+  logs.info(f'Grouping testcase {testcase_1.id} '
+            f'({_get_testcase_log_info(testcase_1)}) and testcase '
+            f'{testcase_2.id} ({_get_testcase_log_info(testcase_2)}). '
+            f'Reason: {reason}.')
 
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/cron/grouper.py
```diff
@@ -97,19 +97,26 @@ def _emit_grouping_event(moved_testcase: int,
           group_merge_reason=group_merge_reason))
 
 
+def _get_testcase_log_info(testcase: TestcaseAttributes):
+  """Return testcase data as string for logging."""
+  log_str = (f'group={testcase.group_id}, '
```
javanlacerda: I can be missing the entire context, but could it be an structured log?

src/clusterfuzz/_internal/cron/grouper.py
```diff
@@ -97,19 +97,26 @@ def _emit_grouping_event(moved_testcase: int,
...
+             f'crash_type={testcase.crash_type}, '
+             f'crash_state={testcase.crash_state}, '
+             f'security_flag={testcase.security_flag}, '
+             f'one_time_crasher={testcase.one_time_crasher_flag}, '
+             f'fuzzer={testcase.fuzzer_name}, '
+             f'job={testcase.job_type}')
```
vitaliset: Maybe it will be easier to read in the logs explorer with \n?

If that's the case, maybe use `textwrap.dedent` then.

vitaliset (approved):

src/clusterfuzz/_internal/cron/grouper.py
vitaliset: Makes sense.

ViniciustCosta: That's a good point! I think the goal of the structured logs is to have some well-defined structured fields that facilitate log search, mainly for specific tasks or a testcase.

In this case, I don't really see a lot of usage for searching logs by them (besides job and fuzzer, which are already present in log structured fields). This log entry will be used to check the information from a deleted testcase, as data will not be available anymore, and not to search, for instance, for all logs from testcases with a specific security flag.

I could add them to the structured logs, though, if you prefer :)

javanlacerda: Besides the performance purposes for structure logs, I think it would make the code cleaner. Maybe to simplify you can just set a namedtuple with these fields and pass it as extras to the log call. wdyt?

ViniciustCosta: I added to the structured log fields, thanks!

ViniciustCosta:
I updated the PR to include the relevant testcase data into our structured logs rather than only for the deletion logs. PTAL

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -948,13 +957,28 @@ def get_extras(self) -> NamedTuple:
...
       try:
-        first_testcase = TestcaseLogStruct(
-            testcase_id=log_contexts.meta.get('testcase_1_id', 'null'),
-            testcase_group=log_contexts.meta.get('testcase_1_group', 'null'))
-        second_testcase = TestcaseLogStruct(
-            testcase_id=log_contexts.meta.get('testcase_2_id', 'null'),
-            testcase_group=log_contexts.meta.get('testcase_2_group', 'null'))
-        symmetric_logs = [first_testcase._asdict(), second_testcase._asdict()]
+        symmetric_logs = []
+        for i in range(1, 3):
```
javanlacerda: What is the meaning of this range? Could it be in a constant with a better defined meaning ?

ViniciustCosta:
>  I'm just wondering if it could be better creating a new struct for this, the testcase struct became complex after the changes.

Tbh I don't see why to create a new struct as all the fields added are from/related to the testcase itself. I think it is not that complex, the struct basically means we are logging the relevant info from a testcase.

src/clusterfuzz/_internal/metrics/logs.py
ViniciustCosta: For context: During the grouping step where it combines testcases, the grouper iterates over pairs of testcases. So we log for both in a symmetric way to help searching for the log entries when querying for just one of these testcases (the same log will be duplicated with distinct fields, e.g., `testcase_id=1` and `testcase_id=2`).

In this case, before the change I was using "first_testcase" and "second_testcase". Now, as we are adding more fields, I used it as a range to avoid replicating a lot of code.

PR #4938, opened 2025-09-02
feat: Add base Dockerfile for Ubuntu 24.04`

hunsche:
### **Context**

This is the foundational first step in the project to upgrade ClusterFuzz's base images from Ubuntu 20.04 to 24.04. The primary motivation is to mitigate security and maintenance risks associated with the upcoming End-of-Life (EOL) of Ubuntu 20.04.

This PR introduces the new, clean base image upon which all other specialized images (like the OSS-Fuzz image) will be built.

### **Description of Changes**

  * **New File:** Introduces a new `docker/base/Dockerfile`.
  * **Base Image:** Sets the parent image to `ubuntu:24.04`.
  * **Packages:** Installs a set of essential packages required for a basic fuzzing environment, including `build-essential`, `git`, `curl`, `wget`, `python3`, and `sudo`.
  * **Configuration:** Implements standard security best practices for the container image and sets up a non-root user.

### **How to Test**

A reviewer can verify this change by building the Docker image locally.

1.  Check out this branch.
2.  From the root of the repository, run the build command:
    ```bash
    docker build -t clusterfuzz/base:ubuntu24.04 -f docker/base/Dockerfile .
    ```
3.  The build should complete successfully. You can also run the container to inspect it:
    ```bash
    docker run -it --rm clusterfuzz/base:ubuntu24.04 bash
    ```

### **Related Issues**

  * This is part of the work for the main tracking issue: **b/395645705**
  * This PR specifically addresses the task: "Develop the base Dockerfile for Ubuntu 24.04".

### **Next Steps**

Once this base image is approved and merged, the next task will be to create the `oss-fuzz` Dockerfile, which will use this image as its foundation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
   <img src="docs/images/logo.png" width="400">
 </p>
 
-[![OpenSSF Scorecard](https://api.security.scorecards.dev/projects/github.com/google/clusterfuzz/badge)](https://api.securityscorecards.dev/projects/github.com/google/clusterfuzz)
+[![OpenSSF Scorecard](https://api.securityscorecards.dev/projects/github.com/google/clusterfuzz/badge)](https://api.securityscorecards.dev/projects/github.com/google/clusterfuzz)
 
 ClusterFuzz is a scalable [fuzzing](https://en.wikipedia.org/wiki/Fuzzing)
 infrastructure that finds security and stability issues in software.
@@ -34,59 +34,6 @@ a software project's development process:
 <p align="center">
   <img src="docs/images/overview.png">
 </p>
-
-## Docker Image Dependency Tree
-
-```mermaid
-graph TD
-    subgraph Base Images
-        A(base)
-    end
-
-    subgraph Chromium
-        F(chromium/base)
-        K(chromium/tester)
-        L(chromium/python-profiler)
-        M(chromium/high-end)
-        E(chromium/tests-syncer)
-        G(chromium/builder)
-    end
-
-    subgraph OSS-Fuzz
-        H(oss-fuzz/base)
-        N(oss-fuzz/worker)
-        O(oss-fuzz/host)
-        P(oss-fuzz/host-high-end)
-    end
-
-    subgraph General
-        B(ci)
-        C(utask-main-scheduler)
-        D(fuchsia)
-        I(tworker)
-        J(high-end)
-    end
-
-
-    A --> B;
-    A --> C;
-    A --> D;
-    A --> E;
-    A --> F;
-    A --> G;
-    A --> H;
-    A --> I;
-    A --> J;
-
-    F --> K;
-    F --> L;
-    F --> M;
-
-    H --> N;
-    H --> O;
-
-    O --> P;
-```
 
 ## Documentation
 You can find detailed documentation [here](https://google.github.io/clusterfuzz).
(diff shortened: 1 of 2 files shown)


docker/chromium/high-end/ubuntu20-04.Dockerfile
```diff
@@ -0,0 +1,17 @@
+# Copyright 2025 Google LLC
```
javanlacerda, line 1: I think this one is not necessary as highend images are deprecated from Clusterfuzz. Reference about the images that are still used here [go/bd-immutable-artifacts](http://goto.google.com/bd-immutable-artifacts)

docker/chromium/high-end/ubuntu24-04.Dockerfile
```diff
@@ -0,0 +1,17 @@
+# Copyright 2025 Google LLC
```
javanlacerda, line 1: same

docker/high-end/ubuntu20-04.Dockerfile
```diff
@@ -0,0 +1,17 @@
+# Copyright 2025 Google LLC
```
javanlacerda, line 1: Same

docker/oss-fuzz/host-high-end/ubuntu20-04.Dockerfile
```diff
@@ -0,0 +1,21 @@
+# Copyright 2025 Google LLC
```
javanlacerda, line 1: Same

docker/base/ubuntu20-04.Dockerfile
```diff
@@ -0,0 +1,156 @@
+# Copyright 2025 Google LLC
```
javanlacerda, line 1: Why do we want to create a new Dockerfile for ubuntu 20?

javanlacerda (commented):
Is there any evidence the docker build command worked for them?

docker/chromium/high-end/ubuntu20-04.Dockerfile
hunsche: Good point. Let's keep this task focused only on the image upgrade. Can you please create a new task for the high-end code cleanup so we don't forget?

docker/base/ubuntu20-04.Dockerfile
hunsche: This is part of our strategy to support multiple OS versions, not just a single latest tag. The current latest will now become ubuntu20-04, and the old Dockerfile will be cleaned up in a follow-up task.

hunsche:
> Is there any evidence the docker build command worked for them?

All builds passed successfully. The build output is extremely verbose, so I've left it out of the PR description.

Instead, I've added a CHANGELOG.md file to each Docker folder. The idea is for us to start tracking updates and recording the build status for each version going forward.

javanlacerda:
> > Is there any evidence the docker build command worked for them?
> 
> All builds passed successfully. The build output is extremely verbose, so I've left it out of the PR description.
> 
> Instead, I've added a CHANGELOG.md file to each Docker folder. The idea is for us to start tracking updates and recording the build status for each version going forward.

Nice to know it worked!

javanlacerda (approved):

PR #4939, opened 2025-09-03
Implements feature flag for update_task

javanlacerda:
It updates the task_loop function to run the update task only if the function `update_task_enabled` retuns True.

The update_task_enabled implements a logic that gets the VM template metadata called `update_task_enabled` and returns if the environment is not a Batch environment and this flag is True. We shouldn't run the update task for Batch because the task is ephemeral for Batch runs.

The flag `update_task_enabled` will be defined by instance template on the terraform repository, enabling us to rollout the update_task deprecation by instance group and gcp project. The implementation of the function will return False, disabling the update task, only if the metadata exists and the content is different of `true`.

This change was already tested in the dev environment before adding the validation for Batch environment, follow the screenshot.
image

### DIFF
--- a/src/clusterfuzz/_internal/system/environment.py
+++ b/src/clusterfuzz/_internal/system/environment.py
@@ -1196,8 +1196,7 @@ def is_tworker():
 
 
 def update_task_enabled() -> bool:
-  """ It uses the GCE VM metadata server to the the metadata 
-      `update_task_enabled` and returns it's boolean.
+  """ It uses the GCE VM metadata server `update_task_enabled` flag.
 
       This flag will be used to rollout the update_task deprecation
       by disabling it progressively for each instance group through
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/system/environment.py
```diff
@@ -1192,3 +1193,39 @@ def can_testcase_run_on_platform(testcase_platform_id, current_platform_id):
 
 def is_tworker():
   return get_value('TWORKER', False)
+
+
+def update_task_enabled() -> bool:
+  """ It uses the GCE VM metadata server to the the metadata
```
ViniciustCosta: Typos in docstring: 
* "to the the metadata" -> "get the metadata"?
* "returns it's boolean." -> "return it as boolean"?

ViniciustCosta: Also, please try to keep the initial docstring as a one-liner and the remaining description afterwards (following 2  go/pystyle#function-docs).

PR #4942, opened 2025-09-04
checkout on the same ref_name

javanlacerda:
It updates the deployment step that gets the config to perform a checkout on the same branch that is the target for the Clusterfuzz.

Currently we work with two branches: dev and master.

For the current state, when deploying for dev from Clusterfuzz pipeline, the config version will always be the master, what can override the current config that can be tested, that is in the dev branch for the config repository.

This change enforeces to sync the branches during the deployment, so dev deployment will use the dev branch for config, and the same for master.

### DIFF
--- a/deploy/cloudbuild.yaml
+++ b/deploy/cloudbuild.yaml
@@ -27,7 +27,7 @@ steps:
     
     # Enforcing that the clusterfuzz and config are using the same branch
     # It makes sure that for dev deployments the deployment overrides the
-    # current version with master
+    # current version with dev
     echo "Checking out Clusterfuzz config to the branch $REF_NAME"
     git checkout $REF_NAME
 
--- a/deploy/cloudbuild.yaml
+++ b/deploy/cloudbuild.yaml
@@ -25,7 +25,13 @@ steps:
     # Move into the repository directory
     cd /workspace/clusterfuzz-config
     
-    # Conditionally run 'git checkout'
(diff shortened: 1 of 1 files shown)


deploy/cloudbuild.yaml
```diff
@@ -25,7 +25,13 @@ steps:
     # Move into the repository directory
     cd /workspace/clusterfuzz-config
     
-    # Conditionally run 'git checkout'
+    # Enforcing that the clusterfuzz and config are using the same branch
+    # It makes sure that for dev deployments the deployment overrides the
+    # current version with master
```
decoNR: I am not sure if I understood it correctly but is this comment right?

javanlacerda: Yes, similary to Python, the comments starts with `#`. It's the continuation of the line 29.

ViniciustCosta: I think @decoNR meant that maybe this should be "overrides the current version with dev"?

decoNR: I mean the content itself. Shouldn't it be "It makes sure that for dev deployments the deployment overrides the current version with **dev**"? Meaning that it will deploy for clusterfuzz dev using clusterfuzz_config dev.

javanlacerda: Oh, I see. My bad. Its wrong, should be with dev. fixing it. Thanks for catching it @decoNR

PR #4945, opened 2025-09-08
Test stack trace parsing of Chromium equality CHECK failures

letitz:
Adds two tests exercising stack trace parsing of recent Chromium CHECK failures:

* `check_eq_multiline` is taken from https://clusterfuzz.com/testcase-detail/6572602755579904
* `check_eq_multiline2` is a simulation of what we would get once https://crrev.com/c/6923305 lands

Bug: https://crbug.com/443588668


letitz:
Thanks!

PR #4946, opened 2025-09-08
Add test for recent Chromium check failure.

letitz:
This shows the error, as the file name is incorrectly parsed. The resulting crash state's first line is:

```
atk_child in 143
```

Instead of what it should be:

```
atk_child in ax_tree_formatter_auralinux.cc
```

A subsequent PR will fix the stack trace parsing logic to extract the latter instead.

Bug: https://crbug.com/443678564


letitz:
Thanks!

PR #4947, opened 2025-09-08
Fix Chromium check failure file/line parsing.

letitz:
Builds on #4946 which introduces the test.

Fixes the regex and tests that the filename is correctly extracted now.

Fixes: https://crbug.com/443678564


letitz:
Thanks for the review! Rebased the change on top of #4946 now that it landed. Checks need to run, then this should be good to land.

letitz:
Ah, one of the tests that just landed in #4945 is failing now, because it also showed the bug :)

PR #4951, opened 2025-09-09
Fix crash grouping for "NULL" crash states

ArthurSonzogni:
The crash comparer was incorrectly grouping different crash types when their crash state was the string "NULL". This was because "NULL" was not being treated as an empty or invalid state.

This change explicitly checks for "NULL" crash states and returns False from `is_similar` if either crash state is "NULL".

A regression test has been added to verify the fix.

Bug: 443261679

### DIFF
--- a/src/clusterfuzz/_internal/cron/grouper.py
+++ b/src/clusterfuzz/_internal/cron/grouper.py
@@ -358,28 +358,28 @@ def _compare_testcases_crash_states(testcase_1, testcase_2) -> bool:
   if (testcase_1.crash_type in data_types.CRASH_TYPES_WITH_UNIQUE_STATE or
       testcase_2.crash_type in data_types.CRASH_TYPES_WITH_UNIQUE_STATE):
 
-    # For grouping, make sure that both crash type and state match.
-    if (testcase_1.crash_type != testcase_2.crash_type or
-        testcase_1.crash_state != testcase_2.crash_state):
-      return False
+    # For grouping, make sure that both crash type and state match exactly.
+    return (
+      testcase_1.crash_type == testcase_2.crash_type and
+      testcase_1.crash_state == testcase_2.crash_state
+    )
 
-  else:
-    # Rule: For functional bugs, compare for similar crash states.
-    if not testcase_1.security_flag:
-      crash_comparer = CrashComparer(
-          testcase_1.crash_type, testcase_2.crash_type,
-          CRASH_COMPARER_THRESHOLD, CRASH_COMPARER_SAME_FRAMES)
-      if not crash_comparer.is_similar():
-        return False
+  crash_comparer_type = CrashComparer(
+      testcase_1.crash_type, testcase_2.crash_type,
+      CRASH_COMPARER_THRESHOLD, CRASH_COMPARER_SAME_FRAMES)
+  crash_comparer_state = CrashComparer(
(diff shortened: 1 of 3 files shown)


src/clusterfuzz/_internal/crash_analysis/crash_comparer.py
```diff
@@ -91,6 +91,13 @@ def is_similar(self):
     if not self.crash_state_1 or not self.crash_state_2:
       return False
 
+    # The string "NULL" is used as a placeholder when a crash type is
+    # identified, but the crash state (stacktrace) is empty, and no
+    # process name or fuzz target is available. It shouldn't be used for
+    # comparison.
+    if self.crash_state_1 == 'NULL' or self.crash_state_2 == 'NULL':
```
ViniciustCosta: Are we sure that this is true for any project and not only chrome/v8?

ArthurSonzogni: Sorry, I am too unfamiliar with Clusterfuzz to understand this question.
Maybe you can try reformulate the question, and highlight why the answer would matter.

Based on `gemini-cli` the [line](https://github.com/google/clusterfuzz/blob/9b28ce87b87e9447281a0ccbe3e224446f840f51/src/clusterfuzz/stacktraces/__init__.py#L360):
```python
state.crash_state = state.process_name or self.fuzz_target or 'NULL'`
```
in `src/clusterfuzz/stacktraces/__init__.py` is the only location where the string `'NULL'` is assigned to `crash_state`.

ViniciustCosta: Clusterfuzz is used by a lot of distinct projects besides chrome, so changing this general crash similarity function will also impact them. My concern is only that we should confirm this is the expected behavior for crashes from any project, i.e., that the crash state being NULL should be considered not similar for all cases.

ArthurSonzogni: Thanks! 

I think I won't be able to help determining that.

In the lines L89-L92, we already state "If one of the crash state is empty, it can't match anything".

Given the current code, if the crash_state is "NULL", we know it means it is empty. So I assumed the comment above would apply too.

The only difference is that we know the crash has a type.

---

What about this:

Rework this patch so that we special case "NULL". We keep considering they are similar, but for two test cases with a similiar "NULL" crash state, we also check the crash_type is similar [here](https://github.com/google/clusterfuzz/blob/11b877c886361260eb7118fb003fdd4704658eaa/src/clusterfuzz/_internal/cron/grouper.py#L379-L382)

That would be enough for chromium, with a reduced impact.

ViniciustCosta: I agree with your points! I do think that the comment on crash state empty not matching anything should be applicable here. Still the `is_similar` is used in a lot of places, including checking for testcase reproducibility, so we might run into unexpected bugs.

For now, maybe let's try to reduce the impact as you suggested by only adding this as an edge case for the grouper.

ArthurSonzogni: Done. Please take another look.

ArthurSonzogni:
Thanks for the review!

Feel free to merge this patch.

PR #4962, opened 2025-09-19
[Tracing] Add event information for blame and variant tasks to the testcase status board

decoNR:
This PR adds two rows to the Task Events section of the testcase status board: one for blame task and one for variant task.

Key changes:
- Included `blame` and `variant` in `TASK_EVENTS_NAMES` to display this information on the testcase status board.

Related to b/435286563.


decoNR:
Here is how the testcase status board looks in the staging App Engine:

image

PR #4964, opened 2025-09-23
[Tracing] Add task-specific comments to task execution events

ViniciustCosta:
Currently, we have some relevant task-specific comments which are added to the testcase comments metadata and displayed on the UI. This PR adds this information to the task execution events, so that we can retrieve these comments on the testcase status board using only the events.

Related to: b/441476616

### DIFF
--- a/src/clusterfuzz/_internal/metrics/events.py
+++ b/src/clusterfuzz/_internal/metrics/events.py
@@ -643,11 +643,11 @@ def get_events_from_testcase(testcase_id: int,
   yield from get_events(equality_filters=equality_filters, order_by=order_by)
 
 
-def emit_task_event(task_command,
-                    event_data,
-                    task_status,
-                    task_stage=None,
-                    task_outcome=None):
+def emit_task_event(task_command: str,
+                    event_data: Mapping[str, Any],
+                    task_status: str,
+                    task_stage: str | None = None,
+                    task_outcome: str | None = None) -> None:
   """Helper to emit task execution events for tasks."""
 
   event_class = (


ViniciustCosta:
I still have to fix the tests.

src/clusterfuzz/_internal/bot/tasks/task_types.py
```diff
@@ -49,27 +76,24 @@ def execute(self, task_argument, job_type, uworker_env):
     # Simple tasks can just use the environment they don't need the uworker env.
     del uworker_env
     assert not environment.is_tworker()
+    os.environ.pop('TASK_COMMENTS', None)
```
decoNR: Maybe use `_reset_stage_env()` here in order to not have duplicated logic. What do you think?

src/clusterfuzz/_internal/bot/tasks/task_types.py
```diff
@@ -24,6 +27,30 @@
...
+        task_comments=environment.get_value('TASK_COMMENTS'))
+  else:
+    task_event = events.TaskExecutionEvent(
+        **event_data,
+        task_stage=events.TaskStage.NA,
+        task_status=task_status,
+        task_outcome=task_outcome,
+        task_comments=environment.get_value('TASK_COMMENTS'))
+
+  events.emit(task_event)
```
vitaliset: You can do something like
```
  event_class = (events.FuzzerTaskExecutionEvent
                 if task_utils.is_fuzzer_based_task(task_command)
                 else events.TaskExecutionEvent)

  task_event = event_class(
      **event_data,
      task_stage=events.TaskStage.NA,
      task_status=task_status,
      task_outcome=task_outcome,
      task_comments=environment.get_value('TASK_COMMENTS'))

  events.emit(task_event)
```
to avoid the duplication of the params, but this is a nit. I'll let you decide.

src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
```diff
@@ -180,13 +181,16 @@ def emit_task_events(self, task_status: str,
...
+          task_comments=environment.get_value('TASK_COMMENTS'))
     else:
       event_task_exec = events.TaskExecutionEvent(
           **self._event_data,
           task_stage=self._subtask.value,
           task_status=task_status,
-          task_outcome=task_outcome)
+          task_outcome=task_outcome,
+          task_comments=environment.get_value('TASK_COMMENTS'))
+
```
vitaliset: Maybe we should expose `_emit_trusted_task_event` as public and use it here as well?

src/clusterfuzz/_internal/bot/tasks/task_types.py
javanlacerda: +1

vitaliset (approved):

javanlacerda (approved):
LGTM. Just address @decoNR's comment :)

src/clusterfuzz/_internal/bot/tasks/task_types.py
ViniciustCosta: I think it's a bit weird to use a method defined in untrusted tasks module in a trusted task class. The other way around is not possible due to direction of imports. 

I will create a centralized method in `task_utils.py`

src/clusterfuzz/_internal/bot/tasks/task_types.py
ViniciustCosta: It makes sense! I will update it

src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
ViniciustCosta: The `_emit_trusted_task_event` is defined in `task_types.py` which is not imported here, it's the other way around.

I think I can define this helper in `task_utils.py` for this as well.

vitaliset:
LGTM!

PR #4967, opened 2025-09-23
[Tracing] Sort events by timestamp in the testcase status board

decoNR:
This PR sorts the events displayed in the testcase status board by timestamp.

Key changes:
- Sorts the event information lists, ensuring the most recent event is the last in the list.

Related to b/435286563 and b/445896035.


src/appengine/handlers/testcase_detail/testcase_status_events.py
```diff
@@ -156,7 +156,10 @@ def get_last_event_info(self,
     return {}
 
   def get_info(self) -> Mapping[str, list[EventInfo]]:
-    """Get testcase status information from events."""
+    """Get testcase status information from events.
+    
+    The lists of events are returned in chronological order.
```
vitaliset, line 161: I suggest adding a test explicitly with this intent.

decoNR: I created it, thank you!

PR #4969, opened 2025-09-25
[Tracing] Add the back-end structure for the event history table

decoNR:
This PR adds the back-end structure for the event history table. A subsequent PR will use this structure to display a testcase event history table on the App Engine testcase detail webpage.

Key changes:
- Adds `get_history()`  to retrieve all events for a given testcase. The returned event information includes a URL that points to the GCP logs associated with the event task.
- Adds `get_task_log()` which retrieves the associated task logs from GCP.

Related to b/394060581.

### DIFF
--- a/src/appengine/handlers/testcase_detail/testcase_status_events.py
+++ b/src/appengine/handlers/testcase_detail/testcase_status_events.py
@@ -24,6 +24,7 @@ from google.cloud import logging_v2
 
 from clusterfuzz._internal.base import utils
 from clusterfuzz._internal.metrics import events
+from clusterfuzz._internal.metrics import logs
 
 EventInfo: TypeAlias = dict[str, str | None]
 
@@ -215,14 +216,21 @@ class TestcaseEventHistory:
 
   def _enrich_event_info_with_gcp_log_url(self, event_info: EventInfo) -> None:
     """Adds the GCP log URL to the event info."""
-    project_id = utils.get_logging_cloud_project_id()
-    task_id = event_info.get('task_id')
-    task_name = event_info.get('task_name')
-    if project_id and task_id and task_name:
+    required_info = {
+        'project_id': utils.get_logging_cloud_project_id(),
+        'task_id': event_info.get('task_id'),
+        'task_name': event_info.get('task_name'),
+    }
+    if all(required_info.values()):
+      project_id, task_id, task_name = required_info.values()
       query = self._get_task_log_query_filter(task_id, task_name)
       encoded_query = urllib.parse.quote(query)
       event_info['gcp_log_url'] = (
           _BASE_LOGS_URL + f'?project={project_id}&query={encoded_query}')
+    else:
+      missing = [k for k, v in required_info.items() if not v]
+      logs.error('Unable to generate GCP log URL due to missing info. '
+                 f'Missing info: {missing}')
 
   def _format_event_for_history(self, event: events.Event) -> EventInfo:
     """Formats an event for display in the event history table."""
--- a/src/appengine/handlers/testcase_detail/testcase_status_events.py
+++ b/src/appengine/handlers/testcase_detail/testcase_status_events.py
@@ -15,7 +15,7 @@
 from dataclasses import asdict
 import datetime
 import json
-from typing import Generator
+from typing import Iterator
 from typing import Mapping
 from typing import TypeAlias
 import urllib.parse
@@ -229,7 +229,7 @@ class TestcaseEventHistory:
     event_info['timestamp'] = _format_timestamp(event.timestamp)
     return event_info
 
-  def get_history(self) -> Generator[Mapping, None, None]:
+  def get_history(self) -> Iterator[Mapping]:
     """Get all testcase events information in reverse chronological order."""
     event_history = events.get_events_from_testcase(self._testcase_id)
     for event in event_history:
--- a/src/appengine/handlers/testcase_detail/testcase_status_events.py
+++ b/src/appengine/handlers/testcase_detail/testcase_status_events.py
@@ -12,13 +12,25 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Helper functions for getting testcase status information from events."""
+from dataclasses import asdict
 import datetime
+import json
+from typing import Generator
 from typing import Mapping
 from typing import TypeAlias
+import urllib.parse
 
+from google.cloud import logging_v2
+
+from clusterfuzz._internal.base import utils
 from clusterfuzz._internal.metrics import events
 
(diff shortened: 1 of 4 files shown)


src/appengine/handlers/testcase_detail/testcase_status_events.py
```diff
@@ -184,6 +192,74 @@ def get_info(self) -> Mapping[str, list[EventInfo]]:
...
+class TestcaseEventHistory:
+  """Methods to retrieve the testcase events history."""
+
+  def __init__(self, testcase_id: int):
+    self._testcase_id = testcase_id
+
+  def _get_time_range_filter(self, days: int) -> str:
+    """Returns a filter string for a time range."""
+    start_time = utils.utcnow() - datetime.timedelta(days=days)
+    return f'timestamp >= "{start_time.isoformat()}Z"'
```
vitaliset, line 207: Is the "Z" intentional?

```suggestion
    return f'timestamp >= "{start_time.isoformat()}"'
```

src/appengine/handlers/testcase_detail/testcase_status_events.py
```diff
@@ -184,6 +192,74 @@ def get_info(self) -> Mapping[str, list[EventInfo]]:
...
+          f'https://console.cloud.google.com/logs/viewer'
+          f'?project={project_id}&query={encoded_query}')
+
+  def _format_event_for_history(self, event: events.Event) -> EventInfo:
+    """Formats an event for display in the event history table."""
+    event_info = {k: v for k, v in asdict(event).items() if v is not None}
+    event_info['timestamp'] = _format_timestamp(event.timestamp)
+    return event_info
+
+  def get_history(self) -> Generator[Mapping, None, None]:
```
vitaliset: Per discussion offline, maybe change to Iterator.

src/appengine/handlers/testcase_detail/testcase_status_events.py
vitaliset: And if it was not intentional, maybe add tests that touch this piece of code explicitly asserting for this string.

vitaliset (approved):

src/appengine/handlers/testcase_detail/testcase_status_events.py
```diff
@@ -184,6 +192,74 @@ def get_info(self) -> Mapping[str, list[EventInfo]]:
...
+
+  def _enrich_event_info_with_gcp_log_url(self, event_info: EventInfo) -> None:
+    """Adds the GCP log URL to the event info."""
+    project_id = utils.get_logging_cloud_project_id()
+    task_id = event_info.get('task_id')
+    task_name = event_info.get('task_name')
+    if project_id and task_id and task_name:
+      query = self._get_task_log_query_filter(task_id, task_name)
+      encoded_query = urllib.parse.quote(query)
+      event_info['gcp_log_url'] = (
```
ViniciustCosta, line 228: Nit: Maybe set a global var `_BASE_LOGS_URL` or something with the "https://console.cloud.google.com/logs/viwer" to be easier to change if we ever need?

src/appengine/handlers/testcase_detail/testcase_status_events.py
```diff
@@ -184,6 +192,74 @@ def get_info(self) -> Mapping[str, list[EventInfo]]:
...
+             f'jsonPayload.extras.task_name="{task_name}"')
+    query += f' AND {self._get_time_range_filter(days=31)}'
+    return query
+
+  def _enrich_event_info_with_gcp_log_url(self, event_info: EventInfo) -> None:
+    """Adds the GCP log URL to the event info."""
+    project_id = utils.get_logging_cloud_project_id()
+    task_id = event_info.get('task_id')
+    task_name = event_info.get('task_name')
+    if project_id and task_id and task_name:
```
ViniciustCosta: Maybe add a log if any of these are missing so we have better visibility?

src/appengine/handlers/testcase_detail/testcase_status_events.py
```diff
@@ -184,6 +192,74 @@ def get_info(self) -> Mapping[str, list[EventInfo]]:
...
+      event_info = self._format_event_for_history(event)
+      self._enrich_event_info_with_gcp_log_url(event_info)
+      yield event_info
+
+  def get_task_log(self, task_id: str, task_name: str) -> str:
+    """Returns the logs for a given task as a string."""
+    project_id = utils.get_logging_cloud_project_id()
+    client = logging_v2.Client(project=project_id)
+    filter_str = self._get_task_log_query_filter(task_id, task_name)
+    entries = client.list_entries(
```
ViniciustCosta, line 254: Is there a possible failure state here? If so, we probably should add a try-catch block when trying to get the entries.

[stretch] Maybe even add an error pop-up in the UI, but this is not a blocker at all :)

ViniciustCosta (approved):
Lgtm!

src/appengine/handlers/testcase_detail/testcase_status_events.py
decoNR: Because `utils.utcnow()` returns a "naive" Python datetime object, the `Z` (UTC timezone indicator) is required for calls to the Google Cloud Logging API. Test `test_get_time_range_filter` address that.

src/appengine/handlers/testcase_detail/testcase_status_events.py
decoNR: I did an experiment to check the type of the object returned by `get_history()` and it is indeed a `Generator`. Given that, I would leave as it is, what do you think?

decoNR: I updated it following [http://go/python-tips/044#type-annotating-function-arguments](https://engdoc.corp.google.com/eng/doc/devguide/py/totw/044.md?cl=head#type-annotating-function-arguments), thank you!

src/appengine/handlers/testcase_detail/testcase_status_events.py
decoNR: I updated it, thank you!

src/appengine/handlers/testcase_detail/testcase_status_events.py
decoNR: As discussed offline, the try catch is not needed as this flow is a separate endpoint and it will not break anything else. Also, if an error occurs, the user will be redirected to an error page. Thank you for the input!

src/appengine/handlers/testcase_detail/testcase_status_events.py
decoNR: I updated it, thank you!

PR #4970, opened 2025-09-29
[Tracing] Add event history table in the testcase detail page

decoNR:
This PR introduces an event history table on the testcase detail App Engine webpage. This table includes all events associated with a specific testcase, a download icon for task's logs related to an event, and a redirection icon linking the user to the GCP logs webpage (pre-filtered to show the task's logs).

Key changes:
- Added `testcase-event-history.html` containing the styling and logic necessary to display the event history on the testcase detail webpage.
- Added the new `testcase-detail/task-log` endpoint. This design ensures that log downloading occurs only upon a user's click on the download icon.

Related to b/394060581.


decoNR:
Here's how it looks in the staging environment. Note that it has a button to either open all the event history or collapse it.

https://screenshot.googleplex.com/4yCN5KYc24hSFEb

vitaliset (approved):

PR #4972, opened 2025-10-06
Add unicode escape minimization to JS minimization pipeline

danylo1999:
We are now fuzzing by replacing characters in JS identifiers with their Unicode escapes:
console.log(42); -> consol\u0065.log(42);

Most often Unicode escapes won't be the problem, so we want to minimize them back to the original character:
consol\u0065.log(42); -> console.log(42);

Thus we add a Unicode minimizer to JS minimization pipeline. Unicode minimizer makes a single attempt at replacing all Unicode escapes with their corresponding UTF-8 characters and seeing if the crash still reproduces after. If it doesn't reproduce after, Unicode minimizer keeps the program as it was.

The difference with the other minimizers is that they were implemented to remove tokens, while this one is replacing tokens.

Thus when we tokenize "consol\u0065.log(42);", we tokenize it into ["consol", "\u0065", "e", ".log(42);"]. Then we try to remove Unicode escapes. If unsuccessful we combine Unicode escapes with their corresponding token by keeping only Unicode escapes: ["consol", "\u0065", "e", ".log(42);"] -> ["consol", "\u0065", ".log(42);"] -> consol\u0065.log(42); 

The code was unit tested and tested from the UI with a local deployment of Clusterfuzz.

### DIFF
--- a/src/clusterfuzz/_internal/bot/minimizer/unicode_minimizer.py
+++ b/src/clusterfuzz/_internal/bot/minimizer/unicode_minimizer.py
@@ -1,0 +1,99 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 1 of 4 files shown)


javanlacerda (approved):
LGTM!

src/clusterfuzz/_internal/bot/minimizer/unicode_minimizer.py
```diff
@@ -0,0 +1,99 @@
...
+    i += 1
+
+  return b''.join(final_tokens)
+
+
+class UnicodeMinimizer(minimizer.Minimizer):
+  """Minimizer to replace \\u0041 -> a, etc. It works the following way:
+  Let's assume there's 'some text\\u0042some other text\\u0444end' string
+  Tokenizer will split it into ['some text', '\u0042', 'ф', 'some other text',
+  'A', 'end']. So for each unicode-escaped symbol, we will add it's
```
mi-ac: nit: its

PR #4971, opened 2025-10-01
Implement immutable image for chrome

javanlacerda:
fix b/448629855

PR #4973, opened 2025-10-07
Implement immutable image for the base image

javanlacerda:
Fixes b/449765246

I does use the structured implemented in #4971 for implementing the immutable images.

Here we create the Dockerfile for the immutable base, using the external config as configuration.

It also add a `push image` flag for pushing the images to the GAR. It's needed to have two different triggers, one that check if the image can be built for the pull requests and another one that builds and push the images for commits on master.

### DIFF
--- a/docker/chromium/base/immutable/Dockerfile
+++ b/docker/chromium/base/immutable/Dockerfile
@@ -19,7 +19,7 @@ ARG CLUSTERFUZZ_SOURCE_DIR
 
 COPY ${CLUSTERFUZZ_SOURCE_DIR} /data/clusterfuzz
 
-RUN bash /data/clusterfuzz/local/install_deps.bash
+RUN cd /data/clusterfuzz && bash local/install_deps.bash
 
 RUN cp -r /data/clusterfuzz/clusterfuzz-config/configs/internal /data/clusterfuzz/src/appengine/config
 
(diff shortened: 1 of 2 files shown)


docker/chromium/base/immutable/Dockerfile
```diff
@@ -19,7 +19,7 @@ ARG CLUSTERFUZZ_SOURCE_DIR
 
 COPY ${CLUSTERFUZZ_SOURCE_DIR} /data/clusterfuzz
 
-RUN bash /data/clusterfuzz/local/install_deps.bash
+RUN cd /data/clusterfuzz && bash local/install_deps.bash
```
decoNR, line 22: Just out of curiosity, why is the current approach better?

javanlacerda: Hey! It needs to be like this because the `install_deps.bash` runs a step to install the thinks from the Pipefile.lock and it expectes to have the Pipefile.lock on its relative root dir, that is on /data/clusterfuzz. Running as `bash /data/clusterfuzz/local/install_deps.bash`, the root dir is /data that doesn't have the Pipefile.

decoNR: Thank you!

PR #4976, opened 2025-10-08
Reproduce testcase locally

PauloVLB:
This PR adds a new `reproduce` command to the `butler` script, enabling local reproduction of a ClusterFuzz testcase. This functionality is a foundational component that will be migrated to a new, standalone CLI tool in the future. The current implementation requires a local clone of the ClusterFuzz repository with all dependencies configured.

---
## Changes Made

A new command was added to the `butler` script that reproduces a testcase. Given a `testcase-id` and a `config-dir`, the command performs the following actions:

1.  Fetches the specified testcase.
2.  Sets up the corresponding job and fuzzer environment.
3.  Sets up the required build and checks if it is a "bad" build.
4.  Runs `test_for_crash_with_retries` to check for a crash.
5.  If a crash is found, it then runs `test_for_reproducibility` and displays the final output.

*Note: Support for testcase exceptions (e.g., data bundles and launch scripts) is out of scope for this PR and will be addressed in follow-up work.*

---
### 1. Authentication
First, you must authenticate using Google Cloud Application Default Credentials. Execute the following command in your terminal:

`gcloud auth application-default login`

### 2. Usage
Once authenticated, run the `reproduce` command with the following format:

`python butler.py reproduce --config-dir=<path_to_config> --testcase-id=<testcase_id>`

**Important**: Please ensure the testcase you are using does not require a launch script, as this is not yet supported.

---
## Testing

Unit tests for this feature have been added in `/src/clusterfuzz/_internal/tests/core/local/butler/reproduce_test.py`.

---
## Future Work

The following improvements are planned as follow-ups to this PR:
* Validate the usage with fuzzers that have data bundles.
* Prevent the fuzzer from being downloaded again if it already exists locally.
* Add support for fuzzers with launch scripts.

### DIFF
--- a/butler.py
+++ b/butler.py
@@ -232,6 +232,14 @@ def _add_weights_subparser(toplevel_subparsers):
   _add_weights_batches_subparser(subparsers)
   _add_weights_target_subparser(subparsers)
 
+def _add_reproduce_subparser(toplevel_subparsers):
+  """Adds a parser for the `reproduce` command."""
+  parser = toplevel_subparsers.add_parser(
+      'reproduce', help='Reproduce a testcase locally.')
+  parser.add_argument(
+      '-c', '--config-dir', required=True, help='Path to application config.')
+  parser.add_argument(
+      '-t', '--testcase-id', required=True, help='The testcase ID to reproduce.')
 
 def main():
   """Parse the command-line args and invoke the right command."""
@@ -408,7 +416,7 @@ def main():
       'integration_tests', help='Run end-to-end integration tests.')
 
   _add_weights_subparser(subparsers)
-
+  _add_reproduce_subparser(subparsers)
   args = parser.parse_args()
   if not args.command:
     parser.print_help()
--- a/src/clusterfuzz/_internal/tests/core/local/butler/reproduce_test.py
+++ b/src/clusterfuzz/_internal/tests/core/local/butler/reproduce_test.py
@@ -1,0 +1,354 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import os
+import tempfile
+import unittest
+from unittest import mock
+
+from local.butler import reproduce
+
+from clusterfuzz._internal.bot import testcase_manager
+from clusterfuzz._internal.datastore import data_handler
+from clusterfuzz._internal.datastore import data_types
+from clusterfuzz._internal.datastore.data_types import Fuzzer, Job, Testcase
+from clusterfuzz._internal.google_cloud_utils import blobs
+from clusterfuzz._internal.metrics import logs
+from clusterfuzz._internal.protos import uworker_msg_pb2
+from clusterfuzz._internal.system import archive
+from clusterfuzz._internal.system import environment
+from clusterfuzz._internal.system import shell
+from clusterfuzz._internal.bot.tasks import setup
+from clusterfuzz._internal.tests.test_libs import helpers
+from clusterfuzz._internal.bot.tasks.commands import update_environment_for_job
+
+class SetupFuzzerTest(unittest.TestCase):
+
+    def setUp(self):
+        super().setUp()
+        helpers.patch_environ(self)
+        self.mock_logs = mock.patch.multiple(
+            'local.butler.reproduce.logs',
+            error=mock.DEFAULT,
+            info=mock.DEFAULT,
+            warning=mock.DEFAULT).start()
+
+        helpers.patch(self, [
+            'clusterfuzz._internal.datastore.data_types.Fuzzer.query',
+            'clusterfuzz._internal.system.environment.set_value',
+            'clusterfuzz._internal.bot.tasks.setup.get_fuzzer_directory',
+            'clusterfuzz._internal.system.shell.remove_directory',
+            'clusterfuzz._internal.google_cloud_utils.blobs.read_blob_to_disk',
+            'clusterfuzz._internal.system.archive.open',
+            'clusterfuzz._internal.system.shell.remove_file',
+            'os.path.exists',
+            'os.chmod',
+        ])
+
+        self.mock_fuzzer_query = self.mock.query
+        self.mock_get = self.mock_fuzzer_query.return_value.get
+        self.mock.get_fuzzer_directory.return_value = '/tmp/fuzzer_dir'
+        self.mock.remove_directory.return_value = True
+        self.mock.read_blob_to_disk.return_value = True
+
+        self.mock_archive_reader = mock.MagicMock()
+        self.mock.open.return_value.__enter__.return_value = self.mock_archive_reader
+
+        self.mock_fuzzer = mock.MagicMock(spec=Fuzzer)
+        self.mock_fuzzer.name = 'test_fuzzer'
+        self.mock_fuzzer.builtin = False
+        self.mock_fuzzer.data_bundle_name = 'test_bundle'
+        self.mock_fuzzer.launcher_script = None
+        self.mock_fuzzer.filename = 'fuzzer.zip'
+        self.mock_fuzzer.executable_path = 'fuzzer_exe'
+        self.mock_fuzzer.blobstore_key = 'some_key'
+        self.mock_get.return_value = self.mock_fuzzer
+
+    def test_setup_fuzzer_builtin_success(self):
+        """Test successful setup of a builtin fuzzer."""
+        self.mock_fuzzer.builtin = True
+        self.assertTrue(reproduce._setup_fuzzer('builtin_fuzzer'))
+        self.mock.set_value.assert_called_once()
+        self.mock.remove_directory.assert_not_called()
+        self.mock_logs['info'].assert_any_call(
+            'Fuzzer builtin_fuzzer is builtin, no setup required.')
+
+    def test_setup_fuzzer_external_success(self):
+        """Test successful setup of an external fuzzer."""
+        self.mock.exists.side_effect = [True, True]  # archive, executable
+        self.assertTrue(reproduce._setup_fuzzer('external_fuzzer'))
+        self.mock.remove_directory.assert_called_once_with(
+            '/tmp/fuzzer_dir', recreate=True)
+        self.mock.read_blob_to_disk.assert_called_once_with(
+            'some_key', '/tmp/fuzzer_dir/fuzzer.zip')
+        self.mock.open.assert_called_once_with('/tmp/fuzzer_dir/fuzzer.zip')
+        self.mock_archive_reader.extract_all.assert_called_once_with('/tmp/fuzzer_dir')
+        self.mock.remove_file.assert_called_once_with('/tmp/fuzzer_dir/fuzzer.zip')
+        self.mock.chmod.assert_called_once_with('/tmp/fuzzer_dir/fuzzer_exe',
+                                                reproduce._EXECUTABLE_PERMISSIONS)
+
+    def test_fuzzer_not_found(self):
+        """Test when the fuzzer is not found in the database."""
+        self.mock_get.return_value = None
+        self.assertFalse(reproduce._setup_fuzzer('nonexistent_fuzzer'))
+        self.mock_logs['error'].assert_called_with(
+            'Fuzzer nonexistent_fuzzer not found.')
+
+    def test_launcher_script_unsupported(self):
+        """Test that fuzzers with launcher scripts are not supported."""
+        self.mock_fuzzer.launcher_script = 'launcher.sh'
+        self.assertFalse(reproduce._setup_fuzzer('launcher_fuzzer'))
+        self.mock_logs['error'].assert_called_with(
+            'Fuzzers with launch script not supported yet.')
+
+    def test_remove_directory_fails(self):
+        """Test failure when clearing the fuzzer directory."""
+        self.mock.remove_directory.return_value = False
+        self.assertFalse(reproduce._setup_fuzzer('external_fuzzer'))
+        self.mock_logs['error'].assert_called_with(
+            'Failed to clear fuzzer directory: /tmp/fuzzer_dir')
+
+    def test_remove_directory_exception(self):
+        """Test exception when clearing the fuzzer directory."""
+        self.mock.remove_directory.side_effect = Exception('mock remove error')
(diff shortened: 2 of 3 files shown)


butler.py
```diff
@@ -236,6 +236,14 @@ def _add_weights_subparser(toplevel_subparsers):
   _add_weights_batches_subparser(subparsers)
   _add_weights_target_subparser(subparsers)
 
+def _add_reproduce_subparser(toplevel_subparsers):
+  """Adds a parser for the `reproduce` command."""
+  parser = toplevel_subparsers.add_parser(
+      'reproduce', help='Reproduce a testcase locally.')
+  parser.add_argument(
```
javanlacerda, line 243: I think we can set it as not required as it points to src/appengine/config by default.

PauloVLB: I think we should keep it as required since it follows the pattern of other `butler` commands that also require an explicit config path, which keeps the script consistent for users.

Also the `reproduce` command always needs a valid, working config to run and in many local development setups, the default `src/appengine/config` points to a test configuration that isn't suitable for reproduction.

javanlacerda: Make sense! thanks!

javanlacerda (approved):

src/local/butler/reproduce.py
```diff
@@ -0,0 +1,267 @@
...
+def execute(args: argparse.Namespace) -> None:
+  """Initializes the environment and reproduces a testcase locally.
+
+  Args:
+    args: Parsed command-line arguments.
+  """
+  os.environ['CONFIG_DIR_OVERRIDE'] = os.path.abspath(args.config_dir)
+  local_config.ProjectConfig().set_environment()
+  environment.set_bot_environment()
+  os.environ['LOG_TO_CONSOLE'] = 'True'
```
ViniciustCosta, line 261: It'd be better to create a method in environment module to set these two env vars (e.g., `set_local_log_only`).

Since this is not done in any script currently, I think we should set it in butler.py (or run.py) possibly with an argument to let the logs go to GCP it if the user wants to.

src/local/butler/reproduce.py
```diff
@@ -0,0 +1,267 @@
...
+    logs.info('The testcase does not reliably reproduce.')
+
+
+def execute(args: argparse.Namespace) -> None:
+  """Initializes the environment and reproduces a testcase locally.
+
+  Args:
+    args: Parsed command-line arguments.
+  """
+  os.environ['CONFIG_DIR_OVERRIDE'] = os.path.abspath(args.config_dir)
```
ViniciustCosta, line 258: Nit: it would be nice to have all these in a setup method (e.g., `_setup_reproduce`).

src/local/butler/reproduce.py
```diff
@@ -0,0 +1,267 @@
...
+from clusterfuzz._internal.bot import testcase_manager
+from clusterfuzz._internal.bot.fuzzers import init
+from clusterfuzz._internal.bot.tasks import setup
+from clusterfuzz._internal.bot.tasks.commands import update_environment_for_job
+from clusterfuzz._internal.build_management import build_manager
+from clusterfuzz._internal.config import local_config
+from clusterfuzz._internal.datastore import data_handler
+from clusterfuzz._internal.datastore import data_types
+from clusterfuzz._internal.datastore import ndb_init
+from clusterfuzz._internal.datastore.data_types import Fuzzer
```
ViniciustCosta, line 30: We usually don't import classes, just modules. (`from ...datastore import data_type`).
go/pystyle#imports

src/local/butler/reproduce.py
```diff
@@ -0,0 +1,267 @@
...
+def _setup_fuzzer(fuzzer_name: str) -> bool:
+  """Sets up the fuzzer binaries and environment.
+
+  Args:
+    fuzzer_name: The name of the fuzzer to set up.
+
+  Returns:
+    True if setup was successful, False otherwise.
+  """
+  fuzzer: Optional[Fuzzer] = data_types.Fuzzer.query(
```
ViniciustCosta, line 53: Thanks for typing the code! However, for internal variables, we try to type it only when it is very difficult to infer its type (here the type is in the query itself), as typing all vars would hinder readability and python's flexibility.

go/pystyle#typing-variables

ViniciustCosta: Also, for Python 3.10+ it is recommended to use the union type `|` instead of the `Optional` syntax.
go/pystyle#none-type

src/local/butler/reproduce.py
```diff
@@ -0,0 +1,267 @@
...
+from clusterfuzz._internal.protos import uworker_msg_pb2
+from clusterfuzz._internal.system import archive
+from clusterfuzz._internal.system import environment
+from clusterfuzz._internal.system import shell
+
+_DEFAULT_TEST_TIMEOUT = 60
+_EXECUTABLE_PERMISSIONS = 0o750
+
+
+def _setup_fuzzer(fuzzer_name: str) -> bool:
```
ViniciustCosta, line 44: I might be oversimplifying things, but I wonder if we could have used methods already available for this. It seems that we are "reinventing the wheel" having to create another setup_fuzzer method, which I imagine copies a lot of the work done in other setup methods.

Maybe, as this is a local thing, we could create a centralized `setup_local_fuzzer` in some other utils/fuzzer module?

src/local/butler/reproduce.py
```diff
@@ -0,0 +1,267 @@
...
+    os.chmod(fuzzer_path, _EXECUTABLE_PERMISSIONS)
+  except OSError as e:
+    logs.error(
+        f'Failed to set permissions on fuzzer executable {fuzzer_path}: {e}')
+    return False
+
+  return True
+
+
+def _setup_testcase_locally(testcase: Testcase) -> Tuple[bool, Optional[str]]:
```
ViniciustCosta, line 124: Ditto for the `_setup_fuzzer()`, I imagine this could also be in another utils/testcase/setup module.

src/local/butler/reproduce.py
```diff
@@ -0,0 +1,267 @@
...
+        - Optional[str]: The local file path to the testcase, or None on failure.
+  """
+  try:
+    shell.clear_testcase_directories()
+  except Exception as e:
+    logs.error(f'Error clearing testcase directories: {e}')
+    return False, None
+
+  try:
+    _, testcase_file_path = setup._get_testcase_file_and_path(testcase)
```
ViniciustCosta, line 142: Should we maybe make the `_get_testcase...` method public? Or, as I commented previously, this method should also live in the setup module.

src/local/butler/reproduce.py
```diff
@@ -0,0 +1,267 @@
...
+    if not blobs.read_blob_to_disk(testcase.fuzzed_keys, testcase_file_path):
```
ViniciustCosta, line 143: I might be missing something, but what about trying to use the `minimized_keys` instead of the `fuzzed_keys`? How does this work for other tasks, such as progression task?

src/local/butler/reproduce.py
```diff
@@ -0,0 +1,267 @@
...
+  return True, testcase_file_path
+
+
+def _reproduce_testcase(args: argparse.Namespace) -> None:
+  """Reproduces a testcase locally based on the provided arguments.
+
+  Args:
+    args: Parsed command-line arguments.
+  """
+  testcase: Optional[Testcase] = data_handler.get_testcase_by_id(
```
ViniciustCosta, line 162: Ditto.

src/local/butler/reproduce.py
```diff
@@ -0,0 +1,267 @@
...
+      args.testcase_id)
+  if not testcase:
+    logs.error(f'Testcase with ID {args.testcase_id} not found.')
+    return
+
+  job: Optional[Job] = data_types.Job.query(
```
ViniciustCosta, line 168: Ditto.

src/local/butler/reproduce.py
```diff
@@ -0,0 +1,267 @@
...
+  # The job name is not set in update_environment_for_job,
+  # so it was needed to manually set it here.
+  environment.set_value('JOB_NAME', job.name)
+  update_environment_for_job(job.get_environment_string())
+
+  if not _setup_fuzzer(testcase.fuzzer_name):
+    logs.error(f'Failed to setup fuzzer {testcase.fuzzer_name}. Exiting.')
+    return
+
+  ok, testcase_file_path = _setup_testcase_locally(testcase)
```
ViniciustCosta, line 183: I think `ok` is not a very insightful variable name.

src/local/butler/reproduce.py
```diff
@@ -0,0 +1,267 @@
...
+  Returns:
+    True if setup was successful, False otherwise.
+  """
+  fuzzer: Optional[Fuzzer] = data_types.Fuzzer.query(
+      data_types.Fuzzer.name == fuzzer_name).get()
+  if not fuzzer:
+    logs.error(f'Fuzzer {fuzzer_name} not found.')
+    return False
+
+  environment.set_value('UNTRUSTED_CONTENT', fuzzer.untrusted_content)
```
ViniciustCosta, line 59: We should add a warning to users about running untrusted code on their machines.

src/local/butler/reproduce.py
PauloVLB: You're right, there is significant overlap with the existing logic in `setup.py`.

The key reason for not reusing it directly is that its flow is designed for untrusted tasks and ultimately downloads things from a signed URL: https://github.com/google/clusterfuzz/blob/fb94beb289f001841f8bd5dd05cdad605c7d4323/src/clusterfuzz/_internal/bot/tasks/setup.py#L624 which we don't want to do locally, right?

My approach instead follows the pattern used by trusted tasks (like `unpack_task`), which use the `blobs` module for direct downloads. 

I agree that centralizing this logic into a `setup_local_fuzzer` is a great idea, and I can refactor it into a shared utils module.

src/local/butler/reproduce.py
PauloVLB: It uses https://github.com/google/clusterfuzz/blob/fb94beb289f001841f8bd5dd05cdad605c7d4323/src/clusterfuzz/_internal/bot/tasks/setup.py#L342
which checks whether the testcase has minimized keys or not, if so, returns the minimized key. I will reproduce this behavior here. Thanks!

src/local/butler/reproduce.py
PauloVLB: Thanks! Addressing in https://github.com/google/clusterfuzz/pull/4988

src/local/butler/reproduce.py
PauloVLB: Thanks! Addressing in https://github.com/google/clusterfuzz/pull/4988

src/local/butler/reproduce.py
PauloVLB: Thanks! Addressing in https://github.com/google/clusterfuzz/pull/4988

src/local/butler/reproduce.py
PauloVLB: Thanks! Addressing in https://github.com/google/clusterfuzz/pull/4988

src/local/butler/reproduce.py
PauloVLB: Agreed. Addressing in https://github.com/google/clusterfuzz/pull/4988

src/local/butler/reproduce.py
PauloVLB: We could pass an argument to https://github.com/google/clusterfuzz/blob/fb94beb289f001841f8bd5dd05cdad605c7d4323/butler.py#L432
And set it there, what do you think? Something like
```python
def _setup(args):
  """Set up configs and import paths."""
  # ...
  if args.local_logging:
    environment.set_local_log_only()
```

I've addressed this in https://github.com/google/clusterfuzz/pull/4988

src/local/butler/reproduce.py
```diff
@@ -0,0 +1,267 @@
...
+  Returns:
+    A tuple containing:
+        - bool: True if the testcase was downloaded successfully.
+        - Optional[str]: The local file path to the testcase, or None on failure.
+  """
+  try:
+    shell.clear_testcase_directories()
+  except Exception as e:
+    logs.error(f'Error clearing testcase directories: {e}')
+    return False, None
```
vitaliset, line 139: As discussed offline, using a separate `ok` boolean in a return, such as `(ok, value)`, is generally unneeded. It complicates the return type hint and is redundant. Returning `None` to signal failure is idiomatic in Python. If value can be `None` to indicate an issue, the `ok` boolean is superfluous because `value is None` already conveys the failure state. For truly exceptional failure conditions, raising custom exceptions is often the most Pythonic approach, as this allows the caller to handle different failure types distinctly.

Also, we don't use this "status return pattern" in the codebase, so this would make it non-standard. Please fix this in your follow-up refactor. :)

src/clusterfuzz/_internal/tests/core/local/butler/reproduce_test.py
```diff
@@ -0,0 +1,391 @@
...
+    # Default return values for success paths
+    self.mock.get_fuzzer_directory.return_value = '/tmp/fuzzer_dir'
+    self.mock.remove_directory.return_value = True
+    self.mock.read_blob_to_disk.return_value = True
+
+    self.mock_archive_reader = mock.MagicMock()
+    self.mock.open.return_value.__enter__.return_value = self.mock_archive_reader
+
+    # Common mock fuzzer object
+    self.mock_fuzzer = mock.MagicMock(spec=Fuzzer)
```
vitaliset, line 77: Always use `mock.create_autospec` with `instance=True` and `spec_set=True`. This makes mocks safer by ensuring they mimic the original object's API, catching errors if methods are misspelled or called with incorrect signatures. This applies generally in the file.

go/python-tips/049

src/clusterfuzz/_internal/tests/core/local/butler/reproduce_test.py
```diff
@@ -0,0 +1,391 @@
+# Copyright 2025 Google LLC
```
vitaliset, line 1: _Replacing your code’s dependencies with mocks can make unit tests easier to write and faster to run. However, among other problems, using mocks can lead to tests that are less effective at catching bugs._ go/tott/697

It's okay to use mocks, but always reflect on whether there is a better option: go/choose-test-double

Writing this comment as a general advice! :)

src/clusterfuzz/_internal/tests/core/local/butler/reproduce_test.py
PauloVLB: Huge thanks for the tip! Applying that on https://github.com/google/clusterfuzz/pull/4988

src/local/butler/reproduce.py
PauloVLB: Thanks! Fixed in https://github.com/google/clusterfuzz/pull/4988

src/local/butler/reproduce.py
PauloVLB: As discussed offline, I will add `setup_local_fuzzer` and `setup_local_testcase` in the existing [setup.py](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/tasks/setup.py). Doing that in https://github.com/google/clusterfuzz/pull/4988

PR #4979, opened 2025-10-09
fix: Copy the config properly on Dockerfiles.

javanlacerda:
We noticed the way the config was being copied to the Clusterfuzz were wrong. The project config should be replacing the directory on src/appengine/config. But actually it was being copied inside of the src/appengine/config, having something like: src/appengine/config/chrome-development.

This change to copy and override the src/appengine/config, having this config with the content of the projects configs.

### DIFF
--- a/docker/base/immutable/external/Dockerfile
+++ b/docker/base/immutable/external/Dockerfile
@@ -19,8 +19,8 @@ ARG CLUSTERFUZZ_SOURCE_DIR
 
 COPY ${CLUSTERFUZZ_SOURCE_DIR} /data/clusterfuzz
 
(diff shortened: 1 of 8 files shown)


decoNR:
Please fix "[build-immutable-images-no-push (clusterfuzz-images)](https://github.com/google/clusterfuzz/pull/4979/checks?check_run_id=52363280873)" check.

javanlacerda:
> Please fix "[build-immutable-images-no-push (clusterfuzz-images)](https://github.com/google/clusterfuzz/pull/4979/checks?check_run_id=52363280873)" check.

I did push the changes to fix, but it seems github is unstable. :/

PR #4988, opened 2025-10-10
[Butler] Enhance reproduce

PauloVLB:
This PR is a follow-up to #4976, addressing the feedback and discussions from the original review. The primary focus is on refactoring and code quality improvements.

*(Note: This PR involves a significant amount of code being moved to more appropriate modules.)*

---
Quick summary:
  - Moved local setup functions into the `setup.py` module 
  - Updated imports to be module-level, adhering to style guidelines.
  - Added a `--local-logging` flag to `butler` to enable convenient local log output during development.
  - Reduced the reliance on mocks in tests
  - Implemented logic to correctly handle `minimized_keys` during setup.
  - Improved error handling to be more Pythonic.
  - Removed unnecessary type hints and enhanced overall test reliability.

### DIFF
--- a/src/clusterfuzz/_internal/tests/core/local/butler/reproduce_test.py
+++ b/src/clusterfuzz/_internal/tests/core/local/butler/reproduce_test.py
@@ -11,7 +11,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Tests for the reproduce butler."""
+"""Tests for the reproduce butler.
+  Run with
+  python butler.py py_unittest -t core -p reproduce_test.py
+"""
 
 import argparse
 import unittest


src/clusterfuzz/_internal/bot/tasks/setup.py
```diff
@@ -651,7 +654,7 @@ def _update_fuzzer(
     return False
 
   # Make fuzzer executable.
-  os.chmod(fuzzer_path, 0o750)
+  os.chmod(fuzzer_path, _EXECUTABLE_PERMISSIONS)
```
javanlacerda, line 657: nice

src/clusterfuzz/_internal/bot/tasks/setup.py
```diff
@@ -799,8 +802,10 @@ def get_fuzzer_directory(fuzzer_name):
   return fuzzer_directory
 
 
-def archive_testcase_and_dependencies_in_gcs(resource_list, testcase_path: str,
-                                             upload_url: str):
+def archive_testcase_and_dependencies_in_gcs(
+    resource_list,
+    testcase_path: str,  # pylint: disable=line-too-long
```
javanlacerda: I think this line-too-long can be removed. ?

javanlacerda (approved):
lgtm

src/clusterfuzz/_internal/bot/tasks/setup.py
PauloVLB: You're right. I've removed and runned lint again. Thanks!

ViniciustCosta:
Could you provide a quick evidence of these changes working?

src/clusterfuzz/_internal/bot/tasks/setup.py
```diff
@@ -890,3 +893,119 @@ def archive_testcase_and_dependencies_in_gcs(resource_list, testcase_path: str,
...
+  """
+  try:
+    shell.clear_testcase_directories()
+  except Exception as e:
+    logs.error(f'Error clearing testcase directories: {e}')
+    return None
+
+  try:
+    _, testcase_file_path = _get_testcase_file_and_path(testcase)
+    if testcase.minimized_keys and testcase.minimized_keys != 'NA':
```
ViniciustCosta: Why not use the `_is_testcase_minimized` method here?

src/clusterfuzz/_internal/system/environment.py
```diff
@@ -1001,6 +1001,12 @@ def set_bot_environment():
   return True
 
 
+def set_local_log_only():
+  """Force logs to be local-only."""
+  set_value('LOG_TO_GCP', '')
```
ViniciustCosta, line 1010: Please, add a short comment explaining that we set this to an empty string because currently the logs module does not correctly evaluate env vars (i.e., 'false' is evaluated to true).

src/clusterfuzz/_internal/bot/tasks/setup.py
```diff
@@ -890,3 +893,119 @@ def archive_testcase_and_dependencies_in_gcs(resource_list, testcase_path: str,
     shell.remove_file(zip_path)
 
   return archived, absolute_filename, zip_filename
+
+
+def setup_local_testcase(testcase: data_types.Testcase) -> str | None:
```
ViniciustCosta, line 898: There is a `preprocess_setup_testcase` method which you don't define a local version. Is there a reason? 

Also, I noticed that in this function, it sets the `APP_ARGS` based on the return of `_get_application_arguments`, which you do only in the reproduce module. Maybe you could try to do it here instead and using this helper method?

src/clusterfuzz/_internal/tests/core/local/butler/reproduce_test.py
```diff
@@ -11,243 +11,19 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+"""Tests for the reproduce butler."""
```
vitaliset: go/pystyle#test-docs

vitaliset (approved):

src/clusterfuzz/_internal/bot/tasks/setup.py
PauloVLB: For the local workflow, I merged the logic of `preprocess_setup_testcase` (which is primarily setting environment variables) directly into setup_local_testcase.

Regarding the use of `_get_application_arguments`, it only returns a value different from `testcase.minimized_arguments` if it's called in a variant task. For that reason, I just set `APP_ARGS` to `testcase.minimized_arguments`.

src/clusterfuzz/_internal/tests/core/local/butler/reproduce_test.py
PauloVLB: I added a module docstring with instructions on how to run this test, as required by the linter. Thanks!

PauloVLB:
> Could you provide a quick evidence of these changes working?

I ran the `reproduce` command with the same testcase ID and config on both the current version and the refactored one. The output is identical.

Note that the refactored version requires the new `--local-logging` flag.

Before Refactor
image

After Refactor
image

src/clusterfuzz/_internal/bot/tasks/setup.py
ViniciustCosta: I see, thanks for the explanation!

Regarding the `preprocess_setup_testcas` and `setup_testcase` I see there are some codepaths there that are not replicated in this local method (e.g., setting the user defined timeout). But I imagine you decided to not include these codepaths as they might be not needed locally. Lgtm

src/clusterfuzz/_internal/bot/tasks/setup.py
```diff
@@ -890,3 +893,119 @@ def archive_testcase_and_dependencies_in_gcs(resource_list, testcase_path: str,
...
+  Returns:
+    True if setup was successful, False otherwise.
+  """
+  fuzzer = data_types.Fuzzer.query(data_types.Fuzzer.name == fuzzer_name).get()
+  if not fuzzer:
+    logs.error(f'Fuzzer {fuzzer_name} not found.')
+    return False
+
+  if fuzzer.untrusted_content:
+    logs.warning('You are about to run an untrusted fuzzer locally. '
```
ViniciustCosta, line 946: Nice

PR #4992, opened 2025-10-13
Support google groups for privileged users

ViniciustCosta:
Add support for google groups to manage privileged users on Clusterfuzz UI. 

### Changes
* Added a new field on the configuration form to allow for admins to add privileged groups (this was done to maintain the support for adding direct emails on the privileged users field).
* Added methods based on the Cloud Identity Groups API in order to retrieve the group ID and to check membership of a logged user. If the user is not a member of the group or the appengine service account doesn't have access to the group, the access will be denied.

### Tests
* Unit tests
* Testing in development by deploying this changes and checking the access with help of the team.
  * Tested admin access still working.
  * Tested that members of a group (test-acl-clusterfuz) had privileged access.
  * Tested that member directly added to the privileged users field had access.

### Limitations
Notice that, for now, this support is limited to groups that are accessible by the Clusterfuzz appengine service account. I.e., groups must either be public or have the service account as a member. Due to internal security policies, the appengine service account does not have domain-wide delegations needed to check membership within any google group. There is a possible workaround by leveraging the user's OAuth credentials when calling the API that might be implemented in the future.

Bug: b/429657295

### DIFF
--- a/src/appengine/libs/auth.py
+++ b/src/appengine/libs/auth.py
@@ -21,6 +21,7 @@ from google.auth.transport import requests as google_requests
 from google.cloud import ndb
 from google.oauth2 import id_token
 from googleapiclient import discovery
+from googleapiclient.errors import HttpError
 import jwt
 import requests
 
@@ -270,8 +271,8 @@ def get_google_group_id(group_email: str,
     request = identity_service.groups().lookup(groupKey_id=group_email)
     response = request.execute()
     return response.get('name')
-  except Exception:
-    logs.info(f"Unable to look up group {group_email}.")
+  except HttpError:
+    logs.warning(f"Unable to look up group {group_email}.")
     return None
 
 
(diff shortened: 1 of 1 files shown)


vitaliset (approved):
Nice work Vini! :)

src/appengine/libs/auth.py
```diff
@@ -249,3 +251,47 @@ def decode_claims(session_cookie):
...
+  try:
+    query_params = parse.urlencode({
+        "query": "member_key_id == '{}'".format(member)
+    })
+    request = identity_service.groups().memberships().checkTransitiveMembership(
+        parent=group_id)
+    request.uri += "&" + query_params
+    response = request.execute()
+    return response.get('hasMembership', False)
+  except Exception:
```
vitaliset: Can we be more specific on the exceptions expected to avoid the general exception? Or maybe we should pass some of the exception information into the logs? Same for line 273.

ViniciustCosta: Sure, I will use an `HttpError` as this is expected when the service account doesn't have permission to access the group.

PR #4995, opened 2025-10-15
feat(tasks): Implement OS-based task routing logic

hunsche:
This change implements the core application logic to support routing ClusterFuzz tasks to different worker pools based on the target operating system.

This is achieved via two main changes:

1.  **Task Publishing (`tasks.add_task`):**
    The `add_task` function now determines the required `base_os_version` for a task by checking the `Job` and `OssFuzzProject` entities. This OS version is then added as a `base_os_version` attribute to the message published on Pub/Sub, effectively "tagging" the task for consumption by the correct worker fleet.

2.  **Task Consuming (`get_*_task` functions):**
    The consumer functions (`get_preprocess_task`, `get_postprocess_task`, `get_utask_mains`) have been updated to be OS-aware. They now read a `BASE_OS_VERSION` environment variable (expected to be set in the Docker image). If this variable is present, they dynamically construct the name of the Pub/Sub subscription to pull from (e.g., `preprocess` becomes `preprocess-ubuntu-24-04`).

Unit tests have been added to cover both the publishing and consuming logic.

**Dependencies:**
This change depends on corresponding infrastructure changes in the `clusterfuzz-config` repository, where the filtered Pub/Sub subscriptions (e.g., `preprocess-ubuntu-24-04`) are created. The code will fail to connect if the subscriptions do not exist.

Fixes: [b/441792657](b/441792657)

### DIFF
--- a/src/clusterfuzz/_internal/tests/core/base/tasks/tasks_test.py
+++ b/src/clusterfuzz/_internal/tests/core/base/tasks/tasks_test.py
@@ -17,7 +17,6 @@ from unittest import mock
 
 from clusterfuzz._internal.base import tasks
 from clusterfuzz._internal.datastore import data_types
-from clusterfuzz._internal.google_cloud_utils import pubsub
 from clusterfuzz._internal.tests.test_libs import test_utils
 
 
@@ -312,9 +311,8 @@ class AddTaskTest(unittest.TestCase):
     self.oss_fuzz_project.put()
 
   @mock.patch('clusterfuzz._internal.base.tasks.data_types.Job.query')
-  def test_add_task_internal_job_with_os_version(self, mock_job_query,
-                                                   mock_add_external,
-                                                   mock_bulk_add):
+  def test_add_task_internal_job_with_os_version(
+      self, mock_job_query, mock_add_external, mock_bulk_add):
     """Test add_task with an internal job and an OS version."""
     mock_job = mock.MagicMock()
     mock_job.base_os_version = 'ubuntu-20-04'
@@ -330,9 +328,8 @@ class AddTaskTest(unittest.TestCase):
     self.assertEqual(task_payload.extra_info['base_os_version'], 'ubuntu-20-04')
 
   @mock.patch('clusterfuzz._internal.base.tasks.data_types.Job.query')
-  def test_add_task_external_job_with_os_version(self, mock_job_query,
-                                                   mock_add_external,
-                                                   mock_bulk_add):
+  def test_add_task_external_job_with_os_version(
+      self, mock_job_query, mock_add_external, mock_bulk_add):
     """Test add_task with an external (OSS-Fuzz) job and an OS version."""
     mock_job = mock.MagicMock()
     mock_job.base_os_version = 'ubuntu-20-04'
@@ -370,7 +367,7 @@ class GetTaskQueueSelectionTest(unittest.TestCase):
       'clusterfuzz._internal.base.tasks.task_utils.is_remotely_executing_utasks'
   )
   def test_get_postprocess_task_without_os_version(self, mock_is_remote,
-                                                    mock_env_get, mock_puller):
+                                                   mock_env_get, mock_puller):
     """Tests that get_postprocess_task selects the default queue."""
     mock_is_remote.return_value = True
     mock_puller.return_value.get_messages.return_value = []
@@ -385,7 +382,7 @@ class GetTaskQueueSelectionTest(unittest.TestCase):
       'clusterfuzz._internal.base.tasks.task_utils.is_remotely_executing_utasks'
   )
   def test_get_postprocess_task_with_os_version(self, mock_is_remote,
-                                                 mock_env_get, mock_puller):
+                                                mock_env_get, mock_puller):
     """Tests that get_postprocess_task selects the suffixed queue."""
     mock_is_remote.return_value = True
     mock_puller.return_value.get_messages.return_value = []


src/clusterfuzz/_internal/base/tasks/__init__.py
```diff
@@ -296,7 +296,13 @@ def get_postprocess_task():
   # wasting our precious non-linux bots on generic postprocess tasks.
   if not environment.platform().lower() == 'linux':
     return None
-  pubsub_puller = PubSubPuller(POSTPROCESS_QUEUE)
+
+  queue_name = POSTPROCESS_QUEUE
+  base_os_version = environment.get_value('BASE_OS_VERSION')
+  if base_os_version:
+    queue_name = f'{queue_name}-{base_os_version}'
```
javanlacerda, line 310: I think it could be an static configuration.

src/clusterfuzz/_internal/base/tasks/__init__.py
```diff
@@ -312,7 +318,12 @@ def allow_all_tasks():
 
 
 def get_preprocess_task():
-  pubsub_puller = PubSubPuller(PREPROCESS_QUEUE)
+  queue_name = PREPROCESS_QUEUE
+  base_os_version = environment.get_value('BASE_OS_VERSION')
+  if base_os_version:
+    queue_name = f'{queue_name}-{base_os_version}'
```
javanlacerda, line 331: Same

src/clusterfuzz/_internal/base/tasks/__init__.py
```diff
@@ -758,6 +774,18 @@ def add_task(command,
     if not job:
       raise Error(f'Job {job_type} not found.')
 
+    # Determine base_os_version.
+    base_os_version = job.base_os_version
+    if environment.get_value('PROJECT_NAME') == 'oss-fuzz':
```
javanlacerda: I think we could move this `oss-fuzz` to a constant to avoid this `magic value`

hunsche: Good point. I've updated the code to use the existing `job.is_external()` method, which removes the magic value.

src/clusterfuzz/_internal/base/tasks/__init__.py
hunsche: That's a good point. In this case, the dynamic approach is needed so the bot can be aware of its own OS environment. The BASE_OS_VERSION is set at the infrastructure level (in the Docker image), allowing each worker to correctly identify which filtered queue it should pull from.

ViniciustCosta:
Did we test this in development?

hunsche:
> Did we test this in development?

Yes, for two weeks.

ViniciustCosta:
> > Did we test this in development?
> 
> Yes, for two weeks.

Nice! Could you provide evidence of it working on dev?

src/clusterfuzz/_internal/tests/core/base/tasks/tasks_test.py
```diff
@@ -296,3 +298,145 @@ def test_set_queue(self):
...
+class AddTaskTest(unittest.TestCase):
+  """Tests for add_task."""
+
+  def setUp(self):
+    self.oss_fuzz_project = data_types.OssFuzzProject(
+        name='d8', base_os_version='ubuntu-24-04')
+    self.oss_fuzz_project.put()
+
+  @mock.patch('clusterfuzz._internal.base.tasks.data_types.Job.query')
+  def test_add_task_with_os_version(self, mock_job_query, mock_add_external,
```
ViniciustCosta: Could you add a test for the case without the os version?

src/clusterfuzz/_internal/tests/core/base/tasks/tasks_test.py
```diff
@@ -296,3 +298,145 @@ def test_set_queue(self):
...
+    tasks.add_task('regression', '123', 'linux_asan_d8_dbg')
+    mock_add_external.assert_not_called()
+    mock_bulk_add.assert_called()
+    task_payload = mock_bulk_add.call_args[0][0][0]
+    self.assertEqual(task_payload.extra_info['base_os_version'], 'ubuntu-20-04')
+
+    mock_bulk_add.reset_mock()
+    mock_add_external.reset_mock()
+
+    # Scenario 2: External (OSS-Fuzz) job. Should use the project's OS version
```
ViniciustCosta: It is better to split these into separate tests, as they are testing distinct behaviors.

[go/tott/751](https://goto.google.com/tott/751) [go/tott/726](https://goto.google.com/tott/726)

src/clusterfuzz/_internal/tests/core/base/tasks/tasks_test.py
```diff
@@ -296,3 +298,145 @@ def test_set_queue(self):
...
+    mock_puller.return_value.get_messages.return_value = []
+
+    # Scenario 1: No OS version ENV. Should use the default queue.
+    mock_env_get.return_value = None
+    tasks.get_preprocess_task()
+    mock_puller.assert_called_with('preprocess')
+
+    mock_puller.reset_mock()
+
+    # Scenario 2: With OS version ENV. Should use the suffixed queue.
```
ViniciustCosta: Ditto.

src/clusterfuzz/_internal/tests/core/base/tasks/tasks_test.py
```diff
@@ -296,3 +298,145 @@ def test_set_queue(self):
...
+    mock_env_get.side_effect = lambda key, default='': {
+        'BASE_OS_VERSION': '',
+        'QUEUE_OVERRIDE': ''
+    }.get(key, default)
+
+    # Scenario 1: Linux platform, no OS version.
+    mock_platform.return_value = 'LINUX'
+    self.assertEqual(tasks.default_queue_suffix(), '-linux')
+
+    # Scenario 2: Linux platform, with OS version.
```
ViniciustCosta: Ditto.

javanlacerda:
> > > Did we test this in development?
> > 
> > 
> > Yes, for two weeks.
> 
> Nice! Could you provide evidence of it working on dev?

+1

hunsche:
> > > Did we test this in development?
> > 
> > 
> > Yes, for two weeks.
> 
> Nice! Could you provide evidence of it working on dev?

As you can see in the attached images, the dev Linux workflows have run normally for the past two weeks. This is by design, as we are not changing the behavior for any instance where `base_os_version` is `ubuntu-24-04`.

https://screenshot.googleplex.com/5euaPuX5H7npv7u.png
https://screenshot.googleplex.com/72nnQjEvKR8nR5z.png

src/clusterfuzz/_internal/base/tasks/__init__.py
javanlacerda: I see. Couldn't we set the POSTPROCESS_QUEUE = <current-postprocess>-<base_os_version> on the start bot? Then we can just use the POSTPROCESS_QUEUE instead of creating it everytime?

src/clusterfuzz/_internal/tests/core/base/tasks/tasks_test.py
javanlacerda: +1

src/clusterfuzz/_internal/base/tasks/__init__.py
hunsche: Good question. I considered that, but I prefer keeping the logic inside the function for clarity.

This way, it's explicit that the queue name is conditional, and we avoid modifying a global constant at startup, which could be confusing to debug later. The performance cost is negligible, so I think this approach is cleaner and easier to maintain.

src/clusterfuzz/_internal/tests/core/base/tasks/tasks_test.py
hunsche: Dome

src/clusterfuzz/_internal/tests/core/base/tasks/tasks_test.py
hunsche: Done

src/clusterfuzz/_internal/tests/core/base/tasks/tasks_test.py
hunsche: Done

src/clusterfuzz/_internal/tests/core/base/tasks/tasks_test.py
hunsche: Done

ViniciustCosta: Thanks!

ViniciustCosta (approved):
lgtm

AFAICT the lint errors are related to other files and it's due to the recently corrected bug in the CI #5009

src/clusterfuzz/_internal/base/tasks/__init__.py
javanlacerda: I disagree this approach improves the clarity, and I think that following the pattern of having a constant would be easier to debug, but it's your call

hunsche: I see your point. For me, it's a trade-off between having the logic in one place at startup vs. having it be explicit where the value is used.

I personally prefer keeping the conditional logic self-contained in the function, as I think it's clearer for future readers.

Thanks for the discussion, I appreciate the feedback.

PR #4996, opened 2025-10-15
Make the immutable instance work properly as a bot.

javanlacerda:
Validate and performs a fine tunning on the changes for b/407497080.

This changes were needed to allow the immutable images to run bots properly. 

I managed to validade it by creating a new instance group using an immutable image with this code version.

[there's the logs of an sucessfull instance](https://pantheon.corp.google.com/logs/query;query=resource.type%3D%22gce_instance%22%0Aresource.labels.instance_id%3D%228195657829940611415%22;cursorTimestamp=2025-10-15T19:42:39.022580Z;duration=PT30M?e=-13802955&mods=logs_tg_prod&project=clusterfuzz-development)

The changes:

* On the setup clusterfuzz scprit, the clusterfuzz source code is copied to the $INSTALL_DIRECTORY if it's a immutable image. This is required because in the current approach, the code is downloaded in the docker volume in /mnt/sractch0 during the startup script. As we copy the code in the build time now, we're not able to copy in the expected path, because the volume doesn't exist in build time.  
* Persist the manifest on src/appengine/resources/clusterfuzz-source.manifest. The immutable images pipeline generates it's own revision, and this change saves it on the expected file for providing this information in runtime (for strcutured logs, for instance)
* Add a guard flag for the function `update_source_code_if_needed` that verifies if it's a immutable instance, and block updating the source code.

### DIFF
--- a/docker/base/setup_clusterfuzz.sh
+++ b/docker/base/setup_clusterfuzz.sh
@@ -28,7 +28,7 @@ fi
 
 if [[ "$IMMUTABLE_IMAGE" == "true" ]]; then
   echo "Not downloading Clusterfuzz source code as it is an immutable image"
-  cp -r /data/clusterfuzz $INSTALL_DIRECTORY/.
+  ln -s /data/clusterfuzz $INSTALL_DIRECTORY/
 else
   CLUSTERFUZZ_FILE=clusterfuzz_package.zip
   # When $LOCAL_SRC is set, use source zip on mounted volume for local testing.


ViniciustCosta:
The logs you linked as evidence are empty, probably since they were shared with relative time. Could you retrieve them and share with absolute time?

docker/build-immutable.sh
```diff
@@ -50,9 +50,9 @@ rm -rf src/appengine/config
 
 # Loop through each image name in the IMAGES array.
 for image_name in "${IMAGES[@]}"; do
-  # Read the ClusterFuzz revision from the revision.txt file. This is used to
+  # Read the ClusterFuzz revision from the clusterfuzz-source.manifest file. This is used to
   # tag the Docker images.
-  CURRENT_CLUSTERFUZZ_REVISION="$(cat /workspace/revision.txt)"
+  CURRENT_CLUSTERFUZZ_REVISION="$(cat src/appengine/resources/clusterfuzz-source.manifest)"
```
ViniciustCosta, line 55: nit: Maybe use a global var to set the manifest filename?

ViniciustCosta (approved):

PR #5009, opened 2025-10-29
fix: Update environment variable check for GitHub Actions in lint script

hunsche:
This PR addresses an issue where the linting script failed to reliably detect the GitHub Actions CI environment.

The script previously checked for the `GITHUB_ACTIONS` environment variable, which is not always propagated to the script's `os.environ`.

To fix this, the following changes were made:
- The GitHub Actions workflow (`.github/workflows/tests.yaml`) was updated to explicitly define a new environment variable, `IS_GITHUB_ACTIONS`, for the test execution step.
- The linting script (`src/local/butler/lint.py`) was modified to check for this new `IS_GITHUB_ACTIONS` variable instead of the default one.

This change ensures a more robust and explicit mechanism for detecting the CI environment, preventing the script from using incorrect git diff commands.

### DIFF
--- a/docker/base/Dockerfile
+++ b/docker/base/Dockerfile
@@ -112,12 +112,14 @@ RUN curl -sS https://www.python.org/ftp/python/3.11.4/Python-3.11.4.tgz | tar -C
     ./configure --enable-optimizations --enable-loadable-sqlite-extensions && make altinstall && \
     rm -rf /tmp/Python-3.11.4 /tmp/Python-3.11.4.tar.xz
 RUN pip3.11 --no-cache-dir install pipenv==2022.8.5
+RUN ln -s /usr/local/bin/python3.11 /usr/bin/python3.11
 
 # Install Node.js
 COPY setup_19.x /data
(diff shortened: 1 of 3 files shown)


local/tests/ci_tests.bash
```diff
@@ -18,6 +18,7 @@ IMAGE=gcr.io/clusterfuzz-images/ci:bc696af-202407021946
 
 docker run -i --rm \
   -e PIPENV_VENV_IN_PROJECT=1 \
+  -e IS_GITHUB_ACTIONS=true \
```
javanlacerda, line 21: I think we should add to all of the docker runs?

PauloVLB: Actually, to solve the current problem, it should be in the third docker run

hunsche: We should refactor our workflows to use GitHub Actions' native support for containers, rather than the current `workflow -> bash -> docker` approach. It would make our CIs much cleaner and easier to maintain.

For this PR, however, my main focus is to fix the currently broken lint issue. I believe the most direct solution right now is to add the environment variable to all the docker run commands.

Issue #4914, opened 2025-08-16
https://clusterfuzz.com/testcase-detail/ page for a Chromium bug doesn't have a "GN Config" section


asvitkine-chromium:
For example:

clusterfuzz.com/testcase-detail/5468644226039808

That page just says:

"Reproduce bug
See https://chromium.googlesource.com/chromium/src/+/master/testing/libfuzzer/reproducing.md for instructions on reproducing this bug locally.
"

And the linked docs say "This will open up an editor. Copy the gn configuration parameters from the values provided in GN Config section in the ClusterFuzz testcase report."

But that "GN Config" section isn't there!

letitz:
I've filed https://crbug.com/441128474 to investigate.

ViniciustCosta:
I added the missing build url and GN config to `clusterfuzz.com/testcase-detail/5468644226039808`.

The root cause was fixed by #5013

PR #5013, opened 2025-10-31
[Fix] Correctly set testcase build url and gn args for fuzzing on batch

ViniciustCosta:
Fix missing testcase metadata when fuzz task is running on batch (or any non-local execution).

#### Context
Some build-related environment variables (`build_url`, `build_key` and `gn_args_path`) are set during the build setup on fuzz task main stage. These should be retrieved during postprocess to set the testcase's metadata, which is used to display the build url and the gn args config section on the clusterfuzz UI.

However, for non-local execution, these variables are not propagated through the `uworker_output` proto and are missing in postprocess, so any testcase created during fuzzing on batch is not displaying these values.

#### Changes
* Added the `build_url`, `build_key` and `gn_args` fields to the fuzz task output message.
* Added the logic to store these fields and use them to set the initial testcase metadata.

#### Tests
* Testing on development environment is ineffective as it is not currently running fuzz tasks on batch. I deployed to dev anyway and confirmed that, at least, the local execution of fuzz tasks is still working.
* Testing on prod (deployed this PR to internal)
   * Fuzzing hours stable after deploy: https://screenshot.googleplex.com/7JfregzBR8F6MRA.png
   * Logs for fuzz tasks executing on batch: https://cloudlogging.app.goo.gl/5xXpLjUdwzK6wTeR9 (main) and https://cloudlogging.app.goo.gl/nX832G1782asdMDXA (postprocess)
   * Logs for fuzz tasks executing on GCE VMs: https://cloudlogging.app.goo.gl/61nvxR1BwQrQnvXr8 (main) and https://cloudlogging.app.goo.gl/6yoQSvU8r1mUVxsN7 (postprocess)

*  Evidence of testcases creates after the deploy containing the build url and GN config args:
    * GCE VMs (local exec): https://clusterfuzz.com/testcase-detail/6746169883426816
    * Batch: https://clusterfuzz.com/testcase-detail/6451774202249216
        * Build URL: https://screenshot.googleplex.com/45n7M75hEPoEsNZ.png
        * GN Configs: https://screenshot.googleplex.com/9r3SxRPRQRyC4tv.png

Bug: b/441128474
Related issue: https://github.com/google/clusterfuzz/issues/4914

### DIFF
--- a/cli/casp/README.md
+++ b/cli/casp/README.md
@@ -63,3 +63,11 @@ For example, to add a `my-command` command, follow these steps:
 
 Once you have completed these steps, the new command will be available as
 `casp my-command`.
+
+## Running Tests
+
+To run all unit tests for the `casp` CLI, use the following command from the root of the project:
+
+```bash
+python -m unittest discover -s cli/casp/src/casp/tests -v
+```
--- a/cli/casp/src/casp/tests/__init__.py
+++ b/cli/casp/src/casp/tests/__init__.py
@@ -1,0 +1,13 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
--- a/cli/casp/src/casp/tests/utils/__init__.py
+++ b/cli/casp/src/casp/tests/utils/__init__.py
@@ -1,0 +1,13 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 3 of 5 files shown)


src/clusterfuzz/_internal/tests/core/datastore/data_handler_test.py
```diff
@@ -40,12 +40,11 @@ def setUp(self):
     test_utils.set_up_pyfakefs(self)
     helpers.patch_environ(self)
 
-  def test_set(self):
-    """Test set everything."""
+  def test_testcase_metadata_from_env(self):
+    """Tests the initial testcase metadata set from the env vars."""
```
decoNR, line 44: Nice improvement :)

ViniciustCosta:
> Could you provide any evidence this changes will not break anything in prod?

I added to the PR description. PTAL

src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -856,8 +848,41 @@ def set_initial_testcase_metadata(testcase):
...
+
+def set_build_metadata_to_testcase(testcase: data_types.Testcase,
+                                   build_key: str | None = None,
+                                   build_url: str | None = None,
+                                   gn_args: str | None = None,
+                                   update: bool = False):
+  """Set testcase metadata fields related to the build metadata."""
+  dirty_flag = False
+  build_key = (
+      environment.get_value('BUILD_KEY') if build_key is None else build_key)
```
javanlacerda, line 863:
```suggestion
  build_key = build_key if build_key else environment.get_value('BUILD_KEY')
```
I think like this is cleaner and easier to understand.

src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -856,8 +848,41 @@ def set_initial_testcase_metadata(testcase):
...
+  if build_key and not testcase.get_metadata('build_key'):
+    dirty_flag = True
+    testcase.set_metadata('build_key', build_key, update_testcase=False)
+
+  build_url = (
```
javanlacerda, line 868: Same here.

src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -856,8 +848,41 @@ def set_initial_testcase_metadata(testcase):
...
+      environment.get_value('BUILD_URL') if build_url is None else build_url)
+  if build_url and not testcase.get_metadata('build_url'):
+    dirty_flag = True
+    testcase.set_metadata('build_url', build_url, update_testcase=False)
+
+  gn_args = get_filtered_gn_args() if gn_args is None else gn_args
```
javanlacerda, line 874: Same

src/clusterfuzz/_internal/datastore/data_handler.py
ViniciustCosta: Thanks for the suggestion, but in this case I do want to explicitly compare with `None` as an empty string may be a valid argument.

PR #5016, opened 2025-11-05
feat: Add USE_GCLOUD_STORAGE_RSYNC=1 to batch fallback init script

hunsche:
This commit adds the USE_GCLOUD_STORAGE_RSYNC=1 environment variable to the linux-init.yaml fallback script used by Cloud Batch jobs.

This ensures that all Cloud Batch containers (bots and batch jobs) will have this environment variable set, enabling rsync for GCS operations.

File modified:
- configs/test/gce/linux-init.yaml
  - Added '-e USE_GCLOUD_STORAGE_RSYNC=1' to the 'docker run' command in the ExecStart section.

### DIFF
--- a/configs/test/gce/linux-init.yaml
+++ b/configs/test/gce/linux-init.yaml
@@ -46,7 +46,7 @@ write_files:
       [Service]
       Environment="HOME=/home/root"
       ExecStartPre=/usr/bin/docker-credential-gcr configure-docker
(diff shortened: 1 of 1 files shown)


decoNR:
For a better understanding on my side, why was it not needed to add `USE_GCLOUD_STORAGE_CP=1` in the `CP` migration?

hunsche:
> For a better understanding on my side, why was it not needed to add `USE_GCLOUD_STORAGE_CP=1` in the `CP` migration?


It might need to. It looks like Cloud Batch is hitting a fallback condition that uses this file.

I'm waiting for a CP usage report. If we don't zero it out for gsutil, this is likely one of the causes.

decoNR:
> It might need to. It looks like Cloud Batch is hitting a fallback condition that uses this file.

> I'm waiting for a CP usage report. If we don't zero it out for gsutil, this is likely one of the causes.

I see. Do we have evidence to confirm that `cloud storage cp` is being used in some cases?

decoNR (approved):

hunsche:
Upon further investigation, the correct PR to resolve this is https://github.com/google/clusterfuzz/pull/5017. I'm closing this one.

PR #5017, opened 2025-11-06
feat: Add USE_GCLOUD_STORAGE_RSYNC=1 to Cloud Batch Jobs

hunsche:
### Descrição:

This PR enables the `USE_GCLOUD_STORAGE_RSYNC=1` environment variable for all containers running as part of a Cloud Batch job. This change is intended to improve the performance and reliability of GCS operations within these jobs by enabling `gcloud storage rsync`.

#### Context

An investigation into the Cloud Batch infrastructure revealed that the environment variables for containers are not set via `user-data` scripts or instance templates. Instead, they are hardcoded in the `docker run` options string that the ClusterFuzz application builds when creating a Batch job.

The `user-data` field in `batch.yaml` files is ignored, and the GCE instance templates defined in Terraform are only used for persistent bots, not for ephemeral Batch VMs.

#### Changes

This PR makes a single, targeted change to correctly inject the environment variable:

-   **Modified `src/clusterfuzz/_internal/google_cloud_utils/batch.py`**:
    -   The `-e USE_GCLOUD_STORAGE_RSYNC=1` flag has been added to the `runnable.container.options` string within the `_get_task_spec` function.

### DIFF
--- a/configs/test/gce/linux-init.yaml
+++ b/configs/test/gce/linux-init.yaml
@@ -46,7 +46,7 @@ write_files:
       [Service]
       Environment="HOME=/home/root"
       ExecStartPre=/usr/bin/docker-credential-gcr configure-docker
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/google_cloud_utils/batch.py
```diff
@@ -133,6 +133,7 @@ def _get_task_spec(batch_workload_spec):
       '-e HOST_UID=1337 -P --privileged --cap-add=all '
       f'-e CLUSTERFUZZ_RELEASE={clusterfuzz_release} '
       '--name=clusterfuzz -e UNTRUSTED_WORKER=False -e UWORKER=True '
+      '-e USE_GCLOUD_STORAGE_RSYNC=1 '
```
decoNR, line 136: As discussed offline, please create a note to potentially add `USE_GCLOUD_STORAGE_CP` here in the future.

PR #5020, opened 2025-11-10
Add 'trunk_staging' release

AashutoshMurthy:
For the purpose of automotive fuzzing, append `trunk_staging` instead of `next` to the `seahawk_hwasan` target to prevent inaccurate querying of AB/ API.


javanlacerda (approved):
LGTM. It will be deployed on next Monday deployment window. Thanks!

PR #5021, opened 2025-11-10
[Fix] Fix find testcase when dedup only same target is set

ViniciustCosta:
As described in b/459451185#comment2, there is a failure state where minimize keeps failing with an exception when the `DEDUP_ONLY_SAME_TARGET` env is set for a job.

This is happening due to two codepaths that were fixed here:
* Minimimize calls `data_handler.handle_duplicate_entry()` which uses the `testcase.fuzzer_name` instead of the `testcase.actual_fuzzer_name()` for the `fuzz_target`.
* `data_handler.find_testcase()` had a misplaced assert that would fail if the first engine from the list was not the one used by the testcase.

Also, note that this PR changes the behavior to avoid the assert. Instead, we log a more descriptive error and return None, since failing to parse the fuzz_target is an expected issue (could be result of misconfiguration) and should be handled gracefully.

b/459451185

### DIFF
--- a/src/clusterfuzz/_internal/tests/core/datastore/data_handler_test.py
+++ b/src/clusterfuzz/_internal/tests/core/datastore/data_handler_test.py
@@ -288,8 +288,14 @@ class DataHandlerTest(unittest.TestCase):
     self.assertTrue(
         test_utils.entities_equal(exclude_result, reproducible_no_bug))
 
+  @parameterized.parameterized.expand([
+      ('engine1_target_test1', True),  # Same fuzz_target.
+      ('engine2_target_test1', True),  # Same fuzz_target with diff engine.
+      ('engine1_target_test3', False)  # Not matching any fuzz_target.
+  ])
   @mock.patch('clusterfuzz._internal.fuzzing.ENGINES', ['engine1', 'engine2'])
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/tests/core/datastore/data_handler_test.py
```diff
@@ -288,6 +288,66 @@ def test_find_testcase(self):
...
+
+    # Case 3: Engine in fuzz_target not found.
+    result = data_handler.find_testcase(
+        'project', 'type', 'state', True, fuzz_target='unknown_target_test2')
+    self.assertIsNone(result)
+
+    # Case 4: Fuzz_target not correctly formatted.
+    result = data_handler.find_testcase(
+        'project', 'type', 'state', True, fuzz_target='engine1')
+    self.assertIsNone(result)
```
vitaliset: Maybe we can parametrize this into two tests, one for case 1/2 and the second for case 3/4.

if you do parametrize it, then maybe the setup (testcase creation) can be moved to a helper function.

ViniciustCosta: Done, thanks

PR #5024, opened 2025-11-12
[CLI] feat: add image selection and fix tests

PauloVLB:
### Context

This PR introduces a way to choose between `internal`, `external` and `dev` immutable images when choosing the project in the `init` command of the CLI and update the tests accordingly.

For doing that, the old `DOCKER_IMAGE` constant is now a dictionary that maps the name of the project to the gcr path of the correspondent image.  It was also added the option `--project` (or `-p`) to specify the project that the user wants to use and pull the correspondent image in the initialization. The user can select multiple projects, if he wants, ex.:

```shell
$ casp init -p internal -p dev
```

### Further work

The gcr path to the images are still hardcoded, this can be improved in future work.

### DIFF
--- a/cli/casp/src/casp/commands/init.py
+++ b/cli/casp/src/casp/commands/init.py
@@ -17,7 +17,6 @@ import os
 import sys
 from typing import Any
 from typing import Dict
-from typing import Tuple
 
 from casp.utils import config
 from casp.utils import docker_utils
@@ -106,7 +105,7 @@ def _pull_image_for_project(project: str = 'internal'):
     type=click.Choice(
         docker_utils.PROJECT_TO_IMAGE.keys(), case_sensitive=False),
     multiple=True)
-def cli(projects: Tuple[str, ...]):
+def cli(projects: tuple[str, ...]):
   """Initializes the CASP CLI.
 
   This is done by:


cli/casp/src/casp/commands/init.py
```diff
@@ -117,5 +127,7 @@ def cli():
   config.save_config(cfg)
   click.secho(f'Configuration saved to {config.CONFIG_FILE}.', fg='green')
 
-  _pull_image()
+  for img in image:
```
javanlacerda: Could you write it clearer, like: for image in project_images?

cli/casp/src/casp/utils/docker_utils.py
```diff
@@ -52,16 +59,17 @@ def check_docker_setup() -> docker.client.DockerClient | None:
     return None
 
 
-def pull_image() -> bool:
+def pull_image(image: str = 'internal') -> bool:
```
javanlacerda: Here `internal` is more a project then an image. You could rename the function to pull_image_for_project, and keep the behavior or actually receive the `image` here instead of the project name, and extracting the step of doing `DOCKER_IMAGES[image]` out of this function.

cli/casp/src/casp/commands/init.py
PauloVLB: Sure! Done.

cli/casp/src/casp/utils/docker_utils.py
PauloVLB: You're right. I've renamed the `project` parameter and the `pull_image_for_project` function as you suggested. The concepts should be used correctly now.

ViniciustCosta:
You can address this in a separate PR, but it would be nice to rename the test files to follow the codebase pattern: `<filename>_test.py`

javanlacerda:
> You can address this in a separate PR, but it would be nice to rename the test files to follow the codebase pattern: `<filename>_test.py`

+1

cli/casp/src/casp/commands/init.py
```diff
@@ -82,19 +83,30 @@ def _setup_custom_config(cfg: Dict[str, Any]):
...
+    '--project',
+    '-p',
+    help=('The ClusterFuzz project to use. You can specify multiple projects.'
+          'Ex.: -p dev -p internal'),
+    required=False,
+    default=('internal',),
+    type=click.Choice(
+        docker_utils.PROJECT_TO_IMAGE.keys(), case_sensitive=False),
+    multiple=True)
+def cli(projects: Tuple[str, ...]):
```
ViniciustCosta: For this kind of typing, prefer using either abstract container types (like `Sequence`) or built-in type like `tuple` itself over the type alias. (go/pystyle#typing-imports)

cli/casp/src/casp/commands/init.py
```diff
@@ -82,19 +83,30 @@ def _setup_custom_config(cfg: Dict[str, Any]):
   click.secho(f'Custom config path set to: {custom_config_path}', fg='green')
 
 
-def _pull_image():
-  """Pulls the docker image."""
-  click.echo(f'Pulling Docker image: {docker_utils.DOCKER_IMAGE}...')
-  if not docker_utils.pull_image():
+def _pull_image_for_project(project: str = 'internal'):
```
ViniciustCosta, line 85: IMO this method should not have a default value for `project`.  It should always expect it as an arg.

cli/casp/src/casp/commands/init.py
```diff
@@ -82,19 +83,30 @@ def _setup_custom_config(cfg: Dict[str, Any]):
...
+         f'{docker_utils.PROJECT_TO_IMAGE[project]}.'),
         fg='red')
     click.secho('Initialization failed.', fg='red')
     sys.exit(1)
 
 
 @click.command(name='init', help='Initializes the CLI')
-def cli():
+@click.option(
+    '--projects',
```
ViniciustCosta, line 98: Suggestion: Maybe the name project is not very clear for external users. I think `config` could be a bit more clear, wdyt?

cli/casp/src/casp/commands/init.py
PauloVLB: Done

cli/casp/src/casp/commands/init.py
PauloVLB: Addressing in future PR.

cli/casp/src/casp/commands/init.py
PauloVLB: Addressing in future PR.

PR #5025, opened 2025-11-12
[fix] Move the place to set base os version to bulk add tasks

javanlacerda:
Moves the logic to set the task base_os_version from add_task to bulk_add_tasks, because add_task is not the final place where tasks are added to the queue. add_task calls bulk_add_tasks, and bulk_add_tasks is called from other places, for instance, the schedule fuzz.

Without this fix, the tasks cannot be routed properly to the right bots, as the base_os_version is omitted by the tasks scheduled by the schedule fuzz cronjob.

### DIFF
--- a/src/clusterfuzz/_internal/base/tasks/__init__.py
+++ b/src/clusterfuzz/_internal/base/tasks/__init__.py
@@ -817,7 +817,7 @@ def bulk_add_tasks(tasks, queue=None, eta_now=False):
     if job.base_os_version:
       task.extra_info['base_os_version'] = job.base_os_version
 
-    if job.is_external():
+    if utils.is_oss_fuzz():
       oss_fuzz_project = data_types.OssFuzzProject.query(
           data_types.OssFuzzProject.name == job.project).get()
       if oss_fuzz_project and oss_fuzz_project.base_os_version:
--- a/src/clusterfuzz/_internal/google_cloud_utils/batch.py
+++ b/src/clusterfuzz/_internal/google_cloud_utils/batch.py
@@ -274,7 +274,7 @@ def _get_config_names(
 
     # If we are running in the oss-fuzz context, the project-specific config
     # is more specific and overrides the job-level one.
-    if job.is_external():
+    if utils.is_oss_fuzz():
       oss_fuzz_project = data_types.OssFuzzProject.query(
           data_types.OssFuzzProject.name == job.project).get()
       if oss_fuzz_project and oss_fuzz_project.base_os_version:
--- a/src/clusterfuzz/_internal/tests/core/base/tasks/tasks_test.py
+++ b/src/clusterfuzz/_internal/tests/core/base/tasks/tasks_test.py
@@ -358,13 +358,14 @@ class AddTasksTest(unittest.TestCase):
 
   @mock.patch('clusterfuzz._internal.google_cloud_utils.pubsub.PubSubClient')
   @mock.patch('clusterfuzz._internal.base.tasks.data_types.Job.query')
-  def test_bulk_add_tasks_internal_job_with_os_version(self, mock_job_query,
-                                                       mock_pubsub_client):
+  @mock.patch('clusterfuzz._internal.base.tasks.utils.is_oss_fuzz')
+  def test_bulk_add_tasks_internal_job_with_os_version(
+      self, mock_is_oss_fuzz, mock_job_query, mock_pubsub_client):
     """Test bulk_add_tasks with an internal job and an OS version."""
+    mock_is_oss_fuzz.return_value = False
     mock_job = mock.MagicMock()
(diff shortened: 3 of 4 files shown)


src/clusterfuzz/_internal/base/tasks/__init__.py
```diff
@@ -806,6 +806,17 @@ def bulk_add_tasks(tasks, queue=None, eta_now=False):
     for task in tasks:
       task.eta = now
 
+  for task in tasks:
+    # Determine base_os_version.
+    job = data_types.Job.query(data_types.Job.name == task.job).get()
```
ViniciustCosta, line 811: Maybe we should check if the job exists in Datastore and handle it if it doesn't.

src/clusterfuzz/_internal/google_cloud_utils/batch.py
```diff
@@ -133,7 +133,6 @@ def _get_task_spec(batch_workload_spec):
       '-e HOST_UID=1337 -P --privileged --cap-add=all '
       f'-e CLUSTERFUZZ_RELEASE={clusterfuzz_release} '
       '--name=clusterfuzz -e UNTRUSTED_WORKER=False -e UWORKER=True '
-      '-e USE_GCLOUD_STORAGE_RSYNC=1 '
```
ViniciustCosta, line 136: I think this was not meant for this PR, right?

javanlacerda: Definetly. I added back

src/clusterfuzz/_internal/base/tasks/__init__.py
javanlacerda: Make sense. I added just a warning and skipped the task to avoid breaking the entire bulk.

src/clusterfuzz/_internal/google_cloud_utils/batch.py
```diff
@@ -275,8 +275,8 @@ def _get_config_names(
     # If we are running in the oss-fuzz context, the project-specific config
     # is more specific and overrides the job-level one.
     if environment.get_value('PROJECT_NAME') == 'oss-fuzz':
-      project_name = job.project
-      oss_fuzz_project = data_types.OssFuzzProject.get_by_id(project_name)
+      oss_fuzz_project = data_types.OssFuzzProject.query(
+          data_types.OssFuzzProject.name == job.project).get()
```
hunsche, line 279: Why are we changing this?

javanlacerda: the entity `OssFuzzProject` doesn't have an `id`, and I couldn't find any other usage of `get_by_id` on the codebase, then I'm not sure with `get_by_id` it would work.

src/clusterfuzz/_internal/google_cloud_utils/batch.py
```diff
@@ -275,8 +275,8 @@ def _get_config_names(
     # If we are running in the oss-fuzz context, the project-specific config
     # is more specific and overrides the job-level one.
     if environment.get_value('PROJECT_NAME') == 'oss-fuzz':
```
hunsche: Can we change this condition to use is_external?

hunsche (approved):

src/clusterfuzz/_internal/base/tasks/__init__.py
```diff
@@ -806,6 +806,23 @@ def bulk_add_tasks(tasks, queue=None, eta_now=False):
...
+    if not job:
```
ViniciustCosta, line 812: Maybe we should still schedule the task instead of skipping. Based on line 826 `if job_type != 'none'` it seems that there may be cases where the job is deliberately set to none.

javanlacerda: But here I'm just skipping the part of adding the extra_info with base_os_version, but the task will be scheduled anyways.

src/clusterfuzz/_internal/google_cloud_utils/batch.py
javanlacerda: Sounds good. Even this wasn't touched by my changes, I applied this refactor :)

src/clusterfuzz/_internal/base/tasks/__init__.py
ViniciustCosta: Right, my bad!

ViniciustCosta (approved):
lgtm, please fix the tests before merging :)

src/clusterfuzz/_internal/google_cloud_utils/batch.py
javanlacerda: Actually @hunsche looking better on the logic, the `is_external` doesn't mean the job if for the `oss-fuzz` project. I'll move back to the  initial logic.

PR #5023, opened 2025-11-11
feat: Add client-side OS version filtering for Pub/Sub tasks

hunsche:
## Description

This PR introduces a client-side filtering mechanism to make ClusterFuzz bots more resilient to misrouted Pub/Sub messages. It ensures that bots only process tasks intended for their specific operating system version, preventing errors and wasted resources when tasks are sent to legacy, unfiltered subscriptions.

### The Problem

When introducing new OS versions, such as Ubuntu 24.04, we create new, filtered Pub/Sub subscriptions (e.g., `my-queue-ubuntu-24-04`). However, existing bots subscribed to legacy, unfiltered queues (e.g., `my-queue`) could still pull messages intended for the new OS. Since Pub/Sub subscription filters are immutable, we cannot simply update old subscriptions. This can lead to bots attempting to execute incompatible tasks, causing errors and inefficiencies.

### The Solution

This change implements a robust, centralized check within the task-pulling logic.

1.  **Centralized Filtering Function:** A new private function, `_filter_task_for_os_mismatch`, has been created in `src/clusterfuzz/_internal/base/tasks/__init__.py`. This function encapsulates all the logic for OS version validation.

2.  **Behavior:**
    - When a bot pulls a message, this function compares the `base_os_version` attribute on the message with the bot's `BASE_OS_VERSION` environment variable.
    - If a mismatch is detected, the function logs a warning and immediately **acknowledges (`ack()`)** the message.
    - Acknowledging the message permanently removes it from that subscription, effectively skipping it for the current bot. This assumes the message was also correctly delivered to another, properly filtered subscription for processing.
    - If the OS versions match, or if either the bot or the message does not have an OS version specified, the task is processed as usual.

3.  **Integration:** This check is performed within `get_task_from_message`, ensuring it is applied to all types of tasks pulled from Pub/Sub (`regular`, `preprocess`, `postprocess`, etc.) without code duplication.

### Benefits

- **Resilience:** Bots are now resilient to misrouted messages and will not fail on incompatible tasks.
- **Cleanliness:** The logic is centralized in a single, well-documented function, improving code maintainability.
- **Forward-Compatibility:** This provides a safety net for future OS migrations and ensures that legacy bots can coexist with newer ones without issue.

### Testing

- Added comprehensive unit tests in `src/clusterfuzz/_internal/tests/core/base/tasks/tasks_test.py` to validate the filtering logic.
- Tests cover all scenarios:
  - OS mismatch (message is skipped and acked).
  - OS match (message is processed).
[...]

PR #5026, opened 2025-11-13
feat: Improve OS version filtering for Pub/Sub tasks

hunsche:
### Description:

**Summary:**

This PR addresses a flaw in the client-side OS version filtering for Pub/Sub tasks. Legacy bots, which do not have the `BASE_OS_VERSION` environment variable set, were not correctly skipping tasks intended for specific, newer OS versions. This could lead to bots attempting to execute incompatible tasks, causing errors and wasting resources, particularly during OS version migrations.

This change makes the task handling more resilient by ensuring that bots only process tasks compatible with their environment.

**Changes:**

- **Modified `_filter_task_for_os_mismatch`:** The core filtering logic in `src/clusterfuzz/_internal/base/tasks/__init__.py` has been updated. It now correctly skips a task if the incoming message has a `base_os_version` attribute and the bot's own OS version is either different or not set (`None`).
- **Updated Unit Tests:** The corresponding unit tests in `src/clusterfuzz/_internal/tests/core/base/tasks/tasks_test.py` have been updated to validate the corrected logic and ensure there are no regressions.
- **Improved Documentation:** The docstring for the filtering function has been refined to be more explicit and conform to the Google Python Style Guide. Redundant inline comments were removed to improve code clarity.

**Testing:**

All changes are covered by unit tests. The relevant test suite passes successfully.

### DIFF
--- a/src/clusterfuzz/_internal/base/tasks/__init__.py
+++ b/src/clusterfuzz/_internal/base/tasks/__init__.py
@@ -586,40 +586,43 @@ class PubSubTTask(PubSubTask):
 
 
 def _filter_task_for_os_mismatch(message, queue) -> bool:
-  """Filters a Pub/Sub message if its OS version does not match the bot's OS.
+  """Filters a Pub/Sub message if its OS version mismatches the bot's OS.
 
-  This function checks the `base_os_version` attribute in the incoming message
-  against the bot's `BASE_OS_VERSION` environment variable. This handles cases
-  where a message is misrouted or received from a legacy subscription without
-  OS-specific filters.
+  This function compares the `base_os_version` attribute from the message
+  against the bot's `BASE_OS_VERSION` environment variable.
 
-  If an OS version mismatch is detected, the function logs a warning and
-  acknowledges (`ack()`) the message. Acknowledging the message permanently
-  removes it from the current subscription, effectively skipping it for this
-  bot. This assumes the message was also correctly delivered to another,
-  properly filtered subscription for processing.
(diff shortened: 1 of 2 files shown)


ViniciustCosta:
I wonder if we could have better tested this in our dev/staging environment for #5023 to avoid having these constant hotfixes.

hunsche:
> I wonder if we could have better tested this in our dev/staging environment for #5023 to avoid having these constant hotfixes.

You're right. In this case, it is genuinely difficult to test since we don't have all behaviors mapped in these environments. The issue actually slipped through an unmapped edge case.

ViniciustCosta:
> > I wonder if we could have better tested this in our dev/staging environment for #5023 to avoid having these constant hotfixes.
> 
> You're right. In this case, it is genuinely difficult to test since we don't have all behaviors mapped in these environments. The issue actually slipped through an unmapped edge case.

What kind of legacy bots don't have the variable set and prevented us from catching this case in dev/staging?

hunsche:
> > > I wonder if we could have better tested this in our dev/staging environment for #5023 to avoid having these constant hotfixes.
> > 
> > 
> > You're right. In this case, it is genuinely difficult to test since we don't have all behaviors mapped in these environments. The issue actually slipped through an unmapped edge case.
> 
> What kind of legacy bots don't have the variable set and prevented us from catching this case in dev/staging?

All bots not running on Ubuntu 24.04 are missing this environment variable. As I mentioned, we should have tested more thoroughly, but we missed catching this in our executed tests.

PR #5028, opened 2025-11-13
[CLI] Run command inside docker container

PauloVLB:
This PR introduces a utility function that executes a command within a Docker container. It's a crucial function for the CLI, as most of its commands run inside a container.

Here's the function's description, which can also be found in its docstring:
```
Runs a command in a docker container and streams logs.
Args:
  command: The command to run.
  volumes: A dictionary of volumes to mount.
  image: The docker image to use.
  privileged: Whether to run the container as privileged.
Returns:
  True on success, False otherwise.
```

Here are two printouts of it in action. The first one simply prints "hello" in the container, and the second one runs `reproduce` for an `internal` testcase.

* `echo hello`
image

* `reproduce` (truncated)
image
image

### DIFF
--- a/cli/casp/src/casp/utils/docker_utils.py
+++ b/cli/casp/src/casp/utils/docker_utils.py
@@ -60,7 +60,7 @@ def check_docker_setup() -> docker.client.DockerClient | None:
     return None
 
 
-def pull_image(image: str = PROJECT_TO_IMAGE['internal']) -> bool:
+def pull_image(image: str) -> bool:
   """Pulls the docker image."""
   client = check_docker_setup()
   if not client:
--- a/cli/casp/src/casp/tests/utils/test_docker.py
+++ b/cli/casp/src/casp/tests/utils/test_docker.py
@@ -129,7 +129,7 @@ class PullImageTest(unittest.TestCase):
         docker.models.images.ImageCollection, instance=True, spec_set=True)
     mock_client.images = mock_images_collection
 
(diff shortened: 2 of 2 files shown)


cli/casp/src/casp/utils/docker_utils.py
```diff
@@ -72,3 +73,71 @@ def pull_image(image: str = PROJECT_TO_IMAGE['internal']) -> bool:
   except docker.errors.DockerException:
     click.secho(f'Error: Docker image {image} not found.', fg='red')
     return False
+
+
+def run_command(command: list[str],
+                volumes: dict,
+                privileged: bool = False,
+                image: str = PROJECT_TO_IMAGE['internal']) -> bool:
```
ViniciustCosta: It's probably not a very good practice to set a default arg to a value from a dict getter (go/pystyle#default-arguments-decision). It's probably better to set as None and retrieve it within the method if None. 

IMO, similar to the previous PR, all methods here probably shouldn't have a default arg for the image/config. I don't see the 'internal' as the default case, it really depends on which user is calling it.

PauloVLB: Thanks for the tip about default arguments! I see now why this method shouldn't have a default value. The `run_command` should be generic and expect the image. I've updated it, thanks!

PauloVLB:
> It would be nice to have some prints of this working :)

I just added some prints to the description

PR #5030, opened 2025-11-14
[FIX] Improve datastore query performance for jobs in bulk tasks

javanlacerda:
Follow up #5025 

It improves the performance of the function bulk_add_tasks. With this the time to perform it should decrease by:

* Make job_names unique, then can be used by all tasks that has the same job.
* Leverage the `.IN` for performing the query in batch instead of single queries for each task.
   (the `.IN` is optimized to be performant)

Evidence of performance improvement. Before the changes, the schedule-fuzz took 10 minutes to schedule 25k tasks, and now as showed in the image, the schedule-fuzz spent 53 seconds to schedule 9.3k tasks, which means that for 25k tasks it would spend 143 seconds, or 2 minutes and 23 seconds.
image

### DIFF
--- a/src/clusterfuzz/_internal/base/tasks/__init__.py
+++ b/src/clusterfuzz/_internal/base/tasks/__init__.py
@@ -824,7 +824,7 @@ def bulk_add_tasks(tasks, queue=None, eta_now=False):
   oss_fuzz_projects_map = {}
   if utils.is_oss_fuzz():
     # Fetch all unique project names from the jobs.
-    project_names = list({job.project for job in jobs if job and job.project})
+    project_names = list({job.project for job in jobs_map.values()})
     logs.info("Project names", project_names=project_names)
     if project_names:
       # Query all OssFuzzProject entities in a single batch.


decoNR:
LGTM, but as discussed offline, perhaps we should have a stress test and analyze other solutions before merging.

src/clusterfuzz/_internal/base/tasks/__init__.py
```diff
@@ -809,9 +809,29 @@ def bulk_add_tasks(tasks, queue=None, eta_now=False):
...
+  job_names = list({task.job for task in tasks})
+  # Query all jobs in a single batch.
+  jobs_query = data_types.Job.query(data_types.Job.name.IN(job_names))
+  jobs = ndb_utils.get_all_from_query(jobs_query)
+  jobs_map = {job.name: job for job in jobs}
+
+  oss_fuzz_projects_map = {}
+  if utils.is_oss_fuzz():
+    # Fetch all unique project names from the jobs.
+    project_names = list({job.project for job in jobs if job})
```
ViniciustCosta: Since you built the `jobs_map` by iterating through all jobs, you could use that same "for" to build this and avoid this second iteration.

ViniciustCosta:
I wonder if every time a object `Task` is created, it already needs to query for the job, so we could add this information a priori to the task itself instead of here.

javanlacerda:
> I wonder if every time a object `Task` is created, it already needs to query for the job, so we could add this information a priori to the task itself instead of here.

It seems a great approach, but I'm afraid it can be more disruptive as it would require modifying other different places. My current approach only improves the performance of a logic that is already tested.

javanlacerda:
Giving visibility, I've added evidence about the performance improvement in the PR description.

src/clusterfuzz/_internal/base/tasks/__init__.py
javanlacerda: I see, but it's a loop in memory, using the list comprehension sounds better to have a more idiomatic approach against avoiding one more iteration. Note that the `list()` function will iterate over the set once again as well, so in this case the performance improvement is not so high.

PR #5031, opened 2025-11-14
Refactor base_os_version Logic for Task Scheduling Performance

hunsche:
## Motivation

The previous implementation for determining the `base_os_version` for new tasks introduced a significant performance bottleneck. The `add_task` function (and its underlying `bulk_add_tasks` wrapper) queried the Datastore for the `Job` and `OssFuzzProject` entities *for each individual task* being created.

In high-throughput scenarios, such as the `schedule_fuzz.py` cron job which can schedule upwards of 300,000 tasks at once, this behavior results in an equivalent number of Datastore queries. This "N+1" query problem leads to extreme slowness, high operational costs, and a significant risk of timeouts and failed task creation.

An alternative approach using a single batch query with an `IN` clause was considered. However, this is also not scalable for a very large number of entities and could hit Datastore limits or result in an unacceptably slow query.

This PR refactors the logic to be far more efficient and scalable.

## Solution

The core idea of this change is to move the responsibility of determining the `base_os_version` to the point where the necessary information is already available, thus eliminating redundant Datastore lookups.

1.  **Logic moved to `schedule_fuzz.py`:**
    The `schedule_fuzz.py` cron job already queries for all `Job` and `OssFuzzProject` entities to perform its scheduling calculations. We now leverage these in-memory entities to determine the correct `base_os_version` *before* the `Task` object is created.

2.  **`base_os_version` Precedence:**
    The logic for selecting the OS version is now explicitly handled within the schedulers (`OssfuzzFuzzTaskScheduler` and `ChromeFuzzTaskScheduler`) with the following precedence:
    - Use `OssFuzzProject.base_os_version` if it exists.
    - Otherwise, use `Job.base_os_version` if it exists.
    - Otherwise, the value is `None`.

3.  **Simplified Task Creation:**
    The determined `base_os_version` is passed directly into the `Task` constructor via the `extra_info` dictionary. This makes the `add_task` and `bulk_add_tasks` functions in `tasks/__init__.py` "dumb" in this regard; they no longer perform any Datastore queries for this purpose and simply publish the tasks they are given.

4.  **Reverting to `add_task`:**
    The logic has been consolidated back into the `add_task` function, removing the `bulk_add_tasks` implementation to simplify the call chain. The `add_task` function now correctly handles the `base_os_version` logic and uses `job.is_external()` for dispatching to `external_tasks`.

## Benefits

*   **Drastic Performance Improvement:** Reduces the number of Datastore queries during the scheduling of fuzz tasks from potentially hundreds of thousands to zero.
*   **Enhanced Scalability:** The system can now schedule extremely large batches of tasks efficiently without overwhelming the Datastore or risking timeouts.
*   **Improved Code Cohesion:** The logic for determining task properties now resides within the scheduler, where the necessary context is already present. This makes the `add_task` function simpler and more focused on its core responsibility of enqueuing a task.

### DIFF
--- a/src/clusterfuzz/_internal/tests/appengine/handlers/cron/schedule_fuzz_test.py
+++ b/src/clusterfuzz/_internal/tests/appengine/handlers/cron/schedule_fuzz_test.py
@@ -46,7 +46,6 @@ class OssfuzzFuzzTaskScheduler(unittest.TestCase):
         name=job_name,
         environment_string=f'PROJECT_NAME = {project_name}',
         platform='LINUX',
-        base_os_version='job-os-version',
     )
     job.put()
 
@@ -66,7 +65,7 @@ class OssfuzzFuzzTaskScheduler(unittest.TestCase):
     dead_project_job = data_types.FuzzerJob(
         job='dead_project_job', platform='LINUX', fuzzer='libFuzzer')
     dead_project_job.put()
-    data_types.OssFuzzProject(name=project_name, base_os_version='project-os-version').put()
+    data_types.OssFuzzProject(name=project_name).put()
     data_types.OssFuzzProject(name='dead_project', cpu_weight=0.0).put()
 
     num_cpus = 10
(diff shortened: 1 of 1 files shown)


hunsche:
# Local Fuzz Scheduling Test Harness: Evidence Report

This document provides a detailed summary of the local testing performed on the fuzz scheduling mechanism, including setup, execution, and results of a high-load test.

## 1. Objective

The primary goal was to create a local, reproducible environment to test and validate the `schedule_fuzz.py` cron job, particularly its ability to handle a large volume of tasks and correctly assign `base_os_version` attributes. This was done to ensure the performance and correctness of the logic in PR #5031 before deployment, without impacting production systems.

## 2. Test Environment Setup

This local test harness was developed and is available on the `feature/local-pubsub-test-harness` git branch. It was created using a Pub/Sub emulator and a set of scripts to automate the workflow.

### Pub/Sub Emulator

A Docker Compose file (`docker-compose.yml`) was created to easily spin up a local Pub/Sub emulator:

```yaml
version: '3.8'
services:
  pubsub-emulator:
    image: gcr.io/google.com/cloudsdktool/cloud-sdk:latest
    command: gcloud beta emulators pubsub start --project=testes-locais --host-port=0.0.0.0:8085
    ports:
      - "8085:8085"
```

### Automation Script (`local/run_fuzz_schedule_test.sh`)

A bash script was created to orchestrate the entire test flow:
1.  Set up environment variables (`PUBSUB_EMULATOR_HOST`, `CLOUDSDK_API_ENDPOINT_OVERRIDES_PUBSUB`).
2.  Automatically create the necessary Pub/Sub topic (`topic-testes`) and subscription (`topic-testes-sub`) on the emulator if they don't exist.
3.  Execute the fuzz scheduler (`run_schedule_fuzz_directly.py`).
4.  Execute an evidence collection script (`local/collect_evidence.py`) to pull messages and generate a report.

## 3. Code Modifications for Local Testing

To make the production code compatible with the local emulator environment, several temporary modifications were made:

*   **`src/clusterfuzz/_internal/cron/schedule_fuzz.py`**:
    *   The `get_available_cpus` function was hardcoded to return `300000` to simulate a high-load scenario.
    *   The `main` function was modified to force the use of local Application Default Credentials (ADC) and to initialize the `ndb` client for the test project.
    *   A timer was added to measure the execution time of the `schedule_fuzz_tasks()` function.

*   **`src/clusterfuzz/_internal/base/tasks/__init__.py`**:
    *   The `PubSubClient` was explicitly initialized with `emulator_host='localhost:8085'`.
    *   The project ID and topic name were hardcoded to point to the local test setup.

*   **`src/clusterfuzz/_internal/google_cloud_utils/pubsub.py`**:
    *   Several patches were applied to the `publish` method to ensure compatibility with the `pubsub_v1` client library used by the emulator:
        *   Added `import os` to resolve a `NameError`.
        *   Ensured message `data` defaults to an empty string (`''`) instead of `None` to prevent an `AttributeError`.
        *   Made the publish call synchronous by waiting for `future.result()` on each message. This was crucial to prevent the test script from finishing before messages were actually published, which would cause a race condition.

## 4. High-Load Test Execution and Results

A high-load test was performed to generate a large number of tasks.

### Execution Summary

*   **Target Tasks**: 150,000 (based on `get_available_cpus` returning 300,000 and `CPUS_PER_FUZZ_JOB` being 2).
*   **Actual Tasks Scheduled**: 150,000
*   **Scheduling Time**: The `schedule_fuzz_tasks()` function successfully generated and published **150,000 tasks** in approximately **10.84 seconds**.

### Evidence of Filtering (`filtered_evidence.txt`)

The `collect_evidence.py` script was modified to pull all 150,000 messages and filter for tasks that had a specific `base_os_version` assigned (i.e., not `None`).

The script produced the following output in `filtered_evidence.txt`:

```
--- Filtered Pub/Sub Test Evidence ---

Total tasks with a specific 'base_os_version': 172

--- Tasks ---

Job: libfuzzer_asan_bios-bmc-smm-error-logger, Base OS Version: ubuntu-24-04
Job: libfuzzer_asan_bios-bmc-smm-error-logger, Base OS Version: ubuntu-24-04
Job: libfuzzer_asan_bios-bmc-smm-error-logger, Base OS Version: ubuntu-24-04
... (and 169 more entries) ...
```

This result demonstrates:
1.  The system successfully processed all 150,000 messages from the queue.
2.  The filtering logic correctly identified **172 tasks** that were assigned a specific `base_os_version`.
3.  The assigned version (`ubuntu-24-04`) was correctly passed through the scheduling and Pub/Sub pipeline.

## 5. Conclusion

The local test harness was successful. It proved that the refactored logic in PR #5031 is highly performant, capable of scheduling 150,000 tasks in under 11 seconds. Furthermore, the end-to-end test confirmed that the `base_os_version` attribute is correctly determined and propagated through the system. The production code in the PR is sound and ready for deployment.

hunsche:
Following the oss-fuzz deployment, the schedule-fuzz cronjob continues to schedule tasks normally and is correctly populating the base_os_version.

We can confirm that the "Skipping task for different OS" message is being logged as expected. Please find the screenshots attached below to assist with the review.



https://screenshot.googleplex.com/AJGFKFo2pWh5Kzp.png

https://screenshot.googleplex.com/5bn9AKmQqZAnmfc.png

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -166,13 +173,22 @@ def get_fuzz_tasks(self) -> Dict[str, tasks.Task]:
...
     # Then get FuzzTaskCandidate weights.
     logs.info('Getting jobs.')
     # TODO(metzman): Handle cases where jobs are fuzzed by multiple fuzzers.
     candidates_by_job = {}
     for job in ndb_utils.get_all_from_query(data_types.Job.query()):
+      project = projects_by_name.get(job.project)
+      base_os_version = None
+      if project and project.base_os_version:
+        base_os_version = project.base_os_version
+      elif job.base_os_version:
```
javanlacerda, line 187: Should we add a else here with a error or warning giving visibility that this task doesn't have any `project.base_os_version` or `job.base_os_version`?

hunsche: If implemented, this log will capture all current behavior. Essentially, any instance not running on Ubuntu 24.04 will trigger it.

PR #5032, opened 2025-11-14
[CLI] Butler commands TBI

PauloVLB:
This PR adds the `butler.py` commands to the `casp` CLI, but leaves it marked as "To be implemented".

These changes are intended to serve as a roadmap for further implementation. Since the commands in `butler.py` need to be validated before migrating to the new CLI, it's crucial to establish a structure and expectations around which commands should be included in `casp`. This will serve as a reminder that these commands should be incorporated into the CLI.

Those commands are:
- [x] bootstrap           
- [x] py_unittest         
- [x] js_unittest         
- [x] format              
- [x] lint                
- [x] package             
- [ ] deploy              
- [ ] run_server          
- [x] run                 
- [ ] run_bot             
- [ ] remote              
- [ ] clean_indexes       
- [ ] create_config       
- [ ] integration_tests   
- [ ] weights             
- [x] reproduce

### DIFF
--- a/docker/base/Dockerfile
+++ b/docker/base/Dockerfile
@@ -128,8 +128,9 @@ RUN ln -s /usr/local/bin/python3.11 /usr/bin/python3.11
 COPY setup_19.x /data
 RUN bash setup_19.x && apt-get update -y && apt-get install -y nodejs
 
-RUN export CLOUDSDK_PYTHON=python3.11 && \
-    echo "deb https://packages.cloud.google.com/apt cloud-sdk main" \
+ENV CLOUDSDK_PYTHON=python3.11
+
+RUN echo "deb https://packages.cloud.google.com/apt cloud-sdk main" \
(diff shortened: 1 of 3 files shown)


javanlacerda:
> Not sure I understand why we are creating all of these as not implemented instead of just creating each one as we migrate them? What is the benefit here?

I also think that without a description, it's a bit vague. I think it's useful to define the structure to be followed in next PRs, but you need to make it explicit in the description.

PauloVLB:
> > Not sure I understand why we are creating all of these as not implemented instead of just creating each one as we migrate them? What is the benefit here?
> 
> I also think that without a description, it's a bit vague. I think it's useful to define the structure to be followed in next PRs, but you need to make it explicit in the description.

I agree with you all. Without a clear description of the intent of this PR, it comes across as vague and unhelpful. I've updated the description to include a reminder that the features intended for `casp` should be implemented. This will help ensure that these important migrations and features aren't forgotten.

PR #5039, opened 2025-11-18
fix python sdk version

javanlacerda:
The base image is not setting the python3.11 and the CLOUD_SDK version because it's doing it by using `export` in a RUN command and not using the `ENV`. 

It fix it by moving the CLOUD_SDK to ENV. 
image


hunsche:
Why is this necessary? This was addressed in PR https://github.com/google/clusterfuzz/pull/5007. It's strange that it stopped working.

javanlacerda:
> Why is this necessary? This was addressed in PR #5007. It's strange that it stopped working.

The way its addressed in #5007 is wrong, it's running `RUN export ...`, and should be with `ENV`.

PR #5047, opened 2025-11-24
Resolve 'TypeError' in _setup_split_targets_build()

9640-Akhil:
This change addresses a `TypeError` in `_setup_split_targets_build()`. [The error](https://pantheon.corp.google.com/errors/detail/CNbfppvCxoWiyAE;locations=global?project=google.com:cluster-fuzz) arises when `_get_targets_list()` returns `None`, because there is an attempt to check for `fuzz_target` in a `NoneType` object. To fix this, a separate check has been introduced to verify that `targets_list` is not `None` before it is iterated over.

### DIFF
--- a/src/clusterfuzz/_internal/tests/core/build_management/build_manager_test.py
+++ b/src/clusterfuzz/_internal/tests/core/build_management/build_manager_test.py
@@ -1731,6 +1731,12 @@ class SplitFuzzTargetsBuildTest(fake_filesystem_unittest.TestCase):
         os.environ['FUZZ_TARGET_BUILD_BUCKET_PATH'])
     self.assertCountEqual(['target1', 'target3'], targets_list)
 
+  def test_setup_split_build_no_targets_list(self):
+    """Test that BuildNotFoundError is raised when the targets list is missing."""
+    self.mock.read_data.return_value = None
+    with self.assertRaises(errors.BuildNotFoundError):
+      build_manager.setup_build(fuzz_target='target3')
+
   def test_target_no_longer_built(self):
     """Test a target that's not longer listed in target.list."""
     test_helpers.patch(self, [


9640-Akhil:
Hello @marktefftech, thank you for the feedback. I've addressed your points below.

1. The `_get_targets_list` function returns `None`, when the `targets.list` file in the build's storage bucket is either missing or empty. As of today, we are seeing 3 jobs (`libfuzzer_asan_android_host_nugget_os_emulator`, `libfuzzer_asan_android_host_sliderGSA`, `libfuzzer_asan_android_host_udfps`) that are giving this error. When we checked the`FUZZ_TARGET_BUILD_BUCKET_PATH` for said jobs, the folder either ⁠[doesn't exist](https://pantheon.corp.google.com/storage/browser/android-haiku?project=google.com:asdl-haiku&pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%257B_22k_22_3A_22_22_2C_22t_22_3A10_2C_22v_22_3A_22_5C_22host-udfps_5C_22_22%257D%255D%22))&prefix=host-udfps&forceOnObjectsSortingFiltering=false) or ⁠[is empty](https://pantheon.corp.google.com/storage/browser/android-haiku/target-nugget-os?project=google.com:asdl-haiku&pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))).
2. The [TypeError](https://pantheon.corp.google.com/errors/detail/CNbfppvCxoWiyAE;locations=global?project=google.com:cluster-fuzz) occurs because the original code attempts to check for the presence of `fuzz_target` within `targets_list` even when `targets_list` is `None` . The change adds a preliminary check/condition for this case, the first part of the condition. If it is indeed found to be `None`, then there is no iteration over `targets_list` successfully avoiding the `TypeError` and raising a `BuildNotFoundError` inside the check instead.
3. Thank you. I have added the testcase.

PR #5048, opened 2025-11-24
Support distinct handling and configuration for DCHECK failures

ArthurSonzogni:
Separates `DCHECK` failures from standard `CHECK` failures to enable granular severity assessment and issue tracking policies.

In Chromium, `DCHECK` failures often carry different security and priority implications than production `CHECK` failures. While they may not always be treated as immediate security vulnerabilities, they present information disclosure risks if filed publicly. Current logic groups them together, preventing distinct visibility rules.

Detailed changes:
- **Stack Parsing:** Updates `stacktraces` regex constants to explicitly distinguish "DCHECK failed" from "Check failed/NOTREACHED", assigning the distinct crash type `DCHECK failure`.
- **Security Implications:** Introduces the `DCHECKS_HAVE_SECURITY_IMPLICATION` environment variable to control whether DCHECKs are flagged as security issues per-fuzzer.
- **Policy Engine:** Refactors `IssueTrackerPolicy` to support recursive configuration application. This allows nested conditions (e.g., `all` -> `non_security` -> `dcheck`) to apply specific labels, access limits, or priority levels based on the intersection of crash traits. This decouple the configuration depth from the code, enabling arbitrary nesting or rules and simplifying the addition of future condition types.

Bug: https://issues.chromium.org/issues/406667202

### DIFF
--- a/src/clusterfuzz/_internal/crash_analysis/crash_analyzer.py
+++ b/src/clusterfuzz/_internal/crash_analysis/crash_analyzer.py
@@ -320,15 +320,20 @@ def is_security_issue(crash_stacktrace, crash_type, crash_address):
     return True
 
   if crash_type == 'CHECK failure':
-    # TODO(ochang): Remove this once we pick up newer builds that distinguish
-    # DCHECKs from CHECKs.
     checks_have_security_implication = environment.get_value(
         'CHECKS_HAVE_SECURITY_IMPLICATION', False)
     return checks_have_security_implication
 
-  # Debug CHECK failure should be marked with security implications.
-  if crash_type in ('Security DCHECK failure', 'DCHECK failure'):
+  # Security DCHECKs are always security issues.
+  if crash_type == 'Security DCHECK failure':
     return True
+
+  # For regular DCHECKs, projects can choose whether or not a particular fuzzer
+  # job treats them as security issues.
+  if crash_type == 'DCHECK failure':
+    dchecks_have_security_implication = environment.get_value(
+        'DCHECKS_HAVE_SECURITY_IMPLICATION', True)
+    return dchecks_have_security_implication
 
   # Hard crash, explicitly enforced in code.
   if (crash_type == 'Fatal error' or crash_type == 'Unreachable code' or
--- a/src/clusterfuzz/_internal/issue_management/issue_tracker_policy.py
+++ b/src/clusterfuzz/_internal/issue_management/issue_tracker_policy.py
@@ -146,62 +146,80 @@ class IssueTrackerPolicy:
     """Get the fallback policy message, if it exists."""
     return self._data.get('fallback_policy_message')
 
-  def get_new_issue_properties(self, is_security, is_crash):
+  def get_new_issue_properties(self, is_security, is_crash, crash_type):
     """Get the properties to apply to a new issue."""
     policy = NewIssuePolicy()
 
-    if 'all' in self._data:
-      self._apply_new_issue_properties(policy, self._data['all'], is_crash)
+    properties = {
+        'is_security': is_security,
+        'is_crash': is_crash,
+        'is_dcheck': 'DCHECK failure' in crash_type
+    }
 
+    self._apply_new_issue_properties(policy, self._data.get('all'),
+                                     **properties)
+
+    # Legacy support: support top-level 'security' and 'non_security' keys. New
+    # configs should use the 'all' section and add 'security' and
+    # 'non_security' subsections within that.
     if is_security:
-      if 'security' in self._data:
-        self._apply_new_issue_properties(policy, self._data['security'],
-                                         is_crash)
+      self._apply_new_issue_properties(policy, self._data.get('security'),
+                                       **properties)
     else:
-      if 'non_security' in self._data:
-        self._apply_new_issue_properties(policy, self._data['non_security'],
-                                         is_crash)
+      self._apply_new_issue_properties(policy, self._data.get('non_security'),
+                                       **properties)
 
     return policy
 
-  def _apply_new_issue_properties(self, policy, issue_type, is_crash):
+  def _apply_new_issue_properties(self, policy, data, **properties):
     """Apply issue policies."""
-    if not issue_type:
+    if not data:
       return
 
-    if 'status' in issue_type:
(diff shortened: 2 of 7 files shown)


ArthurSonzogni:
Hey @letitz 
I would be happy to get some feedback about this patch? Does the allow Chrome to properly adjust visibility of dcheck failures according to you? (Via a new config and/or via the environment flag for specific fuzzers)

The review is a low priority. I can wait as long as needed.

src/clusterfuzz/_internal/crash_analysis/crash_analyzer.py
```diff
@@ -320,16 +320,21 @@ def is_security_issue(crash_stacktrace, crash_type, crash_address):
...
     return checks_have_security_implication
 
-  # Debug CHECK failure should be marked with security implications.
-  if crash_type in ('Security DCHECK failure', 'DCHECK failure'):
+  # Security DCHECKs are always security issues.
+  if crash_type == 'Security DCHECK failure':
     return True
 
+  # For regular DCHECKs, projects can choose whether or not a particular fuzzer
+  # job treats them as security issues.
```
letitz, line 332: The terms "fuzzer" and "job" have specific meanings in ClusterFuzz (and `FuzzerJob` is also a thing), so let's reword:

```suggestion
  # Regular DCHECKs are considered security issues by default, but users
  # can invert that if they so desire.
```

src/clusterfuzz/_internal/issue_management/issue_tracker_policy.py
```diff
@@ -146,62 +146,80 @@ def fallback_policy_message(self):
...
-      if 'non_security' in self._data:
-        self._apply_new_issue_properties(policy, self._data['non_security'],
-                                         is_crash)
+      self._apply_new_issue_properties(policy, self._data.get('non_security'),
+                                       **properties)
 
     return policy
 
-  def _apply_new_issue_properties(self, policy, issue_type, is_crash):
+  def _apply_new_issue_properties(self, policy, data, **properties):
```
letitz, line 174: I would rather keep explicit arguments (named or not, perhaps even wrapped in a dataclass if the list of arguments is getting unwieldy) than this, we lose type information here. AFAICT, properties always has the same 3 keys?

src/clusterfuzz/_internal/issue_management/issue_tracker_policy.py
```diff
@@ -146,62 +146,80 @@ def fallback_policy_message(self):
...
     """Apply issue policies."""
```
letitz, line 175: While we're here and understand how this code works probably better than anyone else in the world, let's document it:

```suggestion
  def _apply_new_issue_properties(self, policy: NewIssuePolicy, data, **properties):
    """Fill in `policy` with properties that an issue linked to a given testcase should possess.
    
    policy: Output argument.
    data: configuration that applies to the testcase linked to the issue. 
    properties: Properties of the testcase linked to the issue.
    """
```

src/clusterfuzz/_internal/issue_management/issue_tracker_policy.py
```diff
@@ -146,62 +146,80 @@ def fallback_policy_message(self):
...
 
-    if 'status' in issue_type:
-      policy.status = self._data['status'][issue_type['status']]
+    if 'status' in data:
+      policy.status = self._data['status'][data['status']]
 
-    if 'ccs' in issue_type:
-      policy.ccs.extend(issue_type['ccs'])
+    if 'ccs' in data:
+      policy.ccs.extend(data['ccs'])
```
letitz, line 183: Optional:

```suggestion
    policy.ccs.extend(data.get('ccs', []))
```

src/clusterfuzz/_internal/issue_management/issue_tracker_policy.py
```diff
@@ -146,62 +146,80 @@ def fallback_policy_message(self):
...
+        'security': properties.get('is_security', None) is True,
+        'non_security': properties.get('is_security', None) is False,
+        'crash': properties.get('is_crash', None) is True,
+        'non_crash': properties.get('is_crash', None) is False,
+        'dcheck': properties.get('is_dcheck', None) is True,
+        'non_dcheck': properties.get('is_dcheck', None) is False,
+    }
+    for child, condition in children_conditions.items():
+      if condition:
+        self._apply_new_issue_properties(policy, data.get(child), **properties)
```
letitz, line 207: I think spelling this out imperatively would be easier to read:

```py
if properties.get('is_security'):
  self._apply_new_issue_properties(policy, data.get('security'), **properties)
# etc.

File: src/clusterfuzz/_internal/issue_management/issue_tracker_policy.py [hunk 6]
```diff
@@ -146,62 +146,80 @@ def fallback_policy_message(self):
...
+
+    # Legacy support: support "crash_label" and "non_crash_label" keys.
+    # New configs should use the "crash" and "non_crash" sections instead and
+    # a "labels" list within those sections.
+    if properties.get('is_crash', None) is False and 'crash_label' in data:
+      policy.labels.append(str(data['crash_label']))
+    if properties.get('is_crash', None) is True and 'non_crash_label' in data:
```
letitz, line 214: Why do we do `is False` and `is True` explicitly instead of relying on None being falsey?

src/clusterfuzz/_internal/issue_management/issue_tracker_policy.py
```diff
@@ -146,62 +146,80 @@ def fallback_policy_message(self):
...
+      policy.labels.append(str(data['non_crash_label']))
+
   def get_existing_issue_properties(self):
     """Get the properties to apply to a new issue."""
     policy = NewIssuePolicy()
-
-    if 'existing' in self._data:
-      self._apply_new_issue_properties(policy, self._data['existing'], False)
-
+    properties = {}  # No recursive properties for existing issues.
```
letitz, line 220: Is this really the right comment here? And are we replicating previous behavior? We used to pass `is_crash = False` explicitly.

letitz (commented):
Looking good overall, a couple comments about specifics.

letitz:
ping @ArthurSonzogni

PR #5049, opened 2025-11-24
[CLI] Match test name pattern of Clusterfuzz

PauloVLB:
This PR renames all the `casp` test files to match Clusterfuzz's `'*_test.py` pattern. It also adds instructions for writing tests for CLI features, suggesting the use of `click.testing.CliRunner`.

Here's a screenshot of the unit tests running with the new naming convention.

image

### DIFF
--- a/cli/casp/README.md
+++ b/cli/casp/README.md
@@ -48,9 +48,9 @@ For example, to add a `my-command` command, follow these steps:
 
     import click
 
-    from .commands import hi
-    from .commands import version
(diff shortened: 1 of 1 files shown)


PauloVLB:
> lgtm
> 
> Suggestion: not sure how the click library works, but it would be better to not print the logs/echo on the terminal when running unit tests.

Agreed! This won't be an issue once those tests are added to the `py_unittest` script. Thanks!

PR #5057, opened 2025-11-26
[CLI] Migrate run

PauloVLB:
### Changes

This PR adds two new commands to the `casp` CLI: `run local` and ` run container`. Both of these commands run a one-off script against the datastore, with the capability to use Clusterfuzz internal code. The difference between them is **where** they run, as explained below.

#### Local
This version runs the script on the host machine, using the local code of clusterfuzz. Basically, users need to have a cloned and configured clusterfuzz and clusterfuzz-config repositories locally. The script they want to run must be in the path `clusterfuzz/src/local/butler/scripts`, just like the `butler.py` command.

#### Container
This version runs inside a Docker container. Choose an image using the `-p` argument: `internal`, `external`, or `dev`. Then, provide the path to your script, which will be mounted inside the container.

### How it works:

- `casp run container <script_name> -p <project>` runs the specified script inside a Docker container (requires a project target: `dev`, `internal`, or `external`).
- `casp run local <script_name> -c <config_dir>` runs the script locally on the host machine (requires the path to the application config).
- Both commands default to a dry run; use `--non-dry-run` to enable actual datastore writes.
- Use `--script_args <text>` to pass specific arguments to the script being executed.

### Changes in butler

The only change in butler was adding a more descriptive print when there are `script_args` in the `run` call. Here are screenshots from both master and this branch showing that the behavior is the same:

image
image

### Demo prints

Here are some screenshots of `casp run` in action:

1. Running in local mode:
- `casp run local list_testcases_by_job --script_args centipede_v8_asan -c ../clusterfuzz-config/configs/internal`
image

2. Running the same script in container mode (note that the script was placed in the root directory here):
- `casp run container list_testcases_by_job.py --script_args centipede_v8_asan -p internal`
image

Here's the script used in the demos for reproduction purposes:
```python
"""One-off script that lists all testcases of a given job"""

def execute(args):
  """Executes the script"""
  try:
    query = ndb_utils.get_all_from_query(
      data_types.Testcase.query(data_types.Testcase.job_type == args.script_args[0]))

    print(f'Those are the testcase ids of job {args.script_args[0]}')
    for testcase in query:
      print(testcase.key.id())
  except Exception as e:
    print(e)
    
```

### DIFF
--- a/cli/casp/src/casp/commands/run.py
+++ b/cli/casp/src/casp/commands/run.py
@@ -38,8 +38,10 @@ def common_options(func):
       '--non-dry-run',
       is_flag=True,
       default=False,
-      help='Run with actual datastore writes. Default to dry-run.')(
-          func)
+      help=('Run with actual datastore writes. '
+            'Default to dry-run. The dry-run behavior needs to be implemented'
+            ' by the script itself, it is not enforced globally'))(
+                func)
   func = click.option(
       '--script_args', multiple=True, help='Script specific arguments')(
           func)


cli/casp/src/casp/commands/run.py
```diff
@@ -7,18 +7,159 @@
...
+from casp.utils import container
+from casp.utils import docker_utils
+from casp.utils import local_butler
 import click
 
 
-@click.command(
+def common_options(func):
+  """
+    Decorator to add common options to both
```
ViniciustCosta: Please, fix docstring style.

src/local/butler/run.py
```diff
@@ -26,7 +26,8 @@
 def execute(args):
   """Run Python unit tests under v2. For unittests involved appengine, sys.path
      needs certain modification."""
-  print(args.script_args)
+  if args.script_args is not None:
```
ViniciustCosta, line 29: Nice hehe

cli/casp/src/casp/utils/docker_utils.py
```diff
@@ -35,26 +35,30 @@
 _DEFAULT_WORKING_DIR = '/data/clusterfuzz'
 
 
+def add_volume(volumes: dict, container_path: str, host_path: str) -> None:
+  """ 
+    Adds a volume in `volumes` dictionary from host to container.
```
ViniciustCosta: Ditto.

cli/casp/src/casp/commands/run.py
```diff
@@ -7,18 +7,159 @@
...
+    local and container modes
+  """
+  func = click.option(
+      '--local',
+      is_flag=True,
+      default=False,
+      help='Run against local server instance.')(
+          func)
+  func = click.option(
+      '--non-dry-run',
```
ViniciustCosta, line 38: nit: IIRC the dry-run flag does not work properly, in the sense that it does not set a global config to avoid writes to the db. As is, the script being executed has to implement its "dry-run" mode. Maybe add that this has to be implemented by the script itself and is not enforced globally.

cli/casp/src/casp/commands/run.py
```diff
@@ -7,18 +7,159 @@
...
+  if local:
+    butler_args['local'] = None
+  if config_dir:
+    butler_args['config_dir'] = config_dir
+
+  return butler_args
+
+
+def _get_script_args_list(script_args: list[str]) -> list[str]:
+  script_args_list = []
```
ViniciustCosta, line 71: Missing docstring (even though it is a private method, it would be nice to have).

cli/casp/src/casp/commands/run.py
PauloVLB: Done

cli/casp/src/casp/utils/docker_utils.py
PauloVLB: Done

cli/casp/src/casp/commands/run.py
PauloVLB: Done. I've enforced this in the help message for the argument.

cli/casp/src/casp/commands/run.py
PauloVLB: Done.

PR #5065, opened 2025-12-01
Assign predator bug component id even when missing suspected components

ViniciustCosta:
Recently, a possible issue in predator was making testcases miss the `suspected_components` field in predator result, even though the correct `suspected_buganizer_component_id` was set. However, the logic in cleanup was preventing the bugs to be assigned with this component ID. This PR fixes by allowing this case.

Context: b/464899796

### DIFF
--- a/src/clusterfuzz/_internal/cron/cleanup.py
+++ b/src/clusterfuzz/_internal/cron/cleanup.py
@@ -1100,6 +1100,14 @@ def update_fuzz_blocker_label(policy, testcase, issue,
   issue.save(new_comment=update_message, notify=True)
 
 
+def _should_update_component_id(issue, component_id):
+  """Check whether updating the issue component id is needed."""
+  if not hasattr(issue, 'component_id'):
+    return False
+
+  return component_id and issue.component_id != component_id
+
+
 def update_component_labels_and_id(policy, testcase, issue):
   """Add components to the issue if needed."""
   if not issue:
@@ -1132,9 +1140,8 @@ def update_component_labels_and_id(policy, testcase, issue):
     if not found_component_in_issue:
       filtered_components.append(component)
 
-  issue_component_id = getattr(issue, 'component_id', None)
   if not (filtered_components or
-          (component_id and issue_component_id != component_id)):
+          _should_update_component_id(issue, component_id)):
     # If there are no new components to add and neither a component ID to
     # update, then we shouldn't make any changes to issue.
     return


ViniciustCosta:
Actually, we should check whether we need to update the component_id as well to avoid updating the issue every time cleanup runs. I will update the PR.

src/clusterfuzz/_internal/cron/cleanup.py
```diff
@@ -1132,9 +1132,11 @@ def update_component_labels_and_id(policy, testcase, issue):
     if not found_component_in_issue:
       filtered_components.append(component)
 
-  if not filtered_components:
-    # If there are no new components to add, then we shouldn't make any changes
-    # to issue.
+  issue_component_id = getattr(issue, 'component_id', None)
```
letitz: Should we unconditionally set `issue.component_id` to `None` in `__init__`, so that we don't have to work around the attribute maybe not existing?

src/clusterfuzz/_internal/tests/appengine/handlers/cron/cleanup_test.py
```diff
@@ -1450,22 +1450,53 @@ def test_components_added(self):
...
+            'suspected_buganizer_component_id': '123456'
+        }})
+
+    cleanup.update_component_labels_and_id(self.policy, self.testcase,
+                                           self.issue)
+    self.assertEqual('123456', self.issue.component_id)
+    self.assertIn('Test-Predator-Auto-Components', self.issue.labels)
+
+  def test_component_id_already_set(self):
+    """Ensure we do not set a component_id again."""
```
letitz:
```suggestion
    """Ensure we do not make an API call if the component ID has not changed."""
```

letitz: I wanted to highlight that if the component ID is already set to a different value, it should be modified. Can we add a test for that case?

src/clusterfuzz/_internal/cron/cleanup.py
letitz: Ah,  I see the comment on line 1165 about this only existing on issue tracker issues. I still think it should be None for all other kinds of issues, this attr detection is awkward. Anyway, it might not be the right idea to update it in this PR, and you may decide it's not worth doing at all. Up to you!

letitz (approved):
I think we just need one more test for the case where the issue component ID should be changed (different value, no auto-components label). LGTM otherwise, thanks!

src/clusterfuzz/_internal/tests/appengine/handlers/cron/cleanup_test.py
ViniciustCosta: Sure, I'll add it

src/clusterfuzz/_internal/cron/cleanup.py
ViniciustCosta: I think a possible problem is that the issue trackers that don't implement a `component_id` will always return None and (I'm not sure if this is possible) if the predator component_id is set, it will try to update the component_id every cleanup call since this will not be propagated.
In fact, I added this sanity check to see if the issue has the `component_id` attr to avoid this.

I would rather improve this in another PR so we resolve the P0 issue for now :)

PR #5066, opened 2025-12-02
Correct crash classification for MTE SIGABRT and SIGTRAP traces

AYUSHMAAN-B:
Previously, MTE crashes involving `SIGABRT` and `SIGTRAP` were not explicitly handled, leading to the crash_type being misclassified as `UNKNOWN`.

This change adds explicit checks for `SIGABRT` and `SIGTRAP`, updates the crash_type to `Abort` and `Trap` respectively, and extracts the `crash_address` for `SIGTRAP` crashes.

The update ensures that crashes involving these MTE faults are parsed correctly and filed under the appropriate crash type.


marktefftech (approved):
Thanks for adding the test cases!

PR #5079, opened 2025-12-15
Refactoring the fuzzing hours labels by moving from is_batch to runtime

javanlacerda:
Fix b/469019068

It does update the fuzzing hours metrics to add KATA CONTAINER as a runtime.
It removes the label `is_batch` and move it to the `runtime` label as `BATCH`.


jonathanmetzman:
OOOF I think we lose all of our old stats if we land this. How can we preserve them while changing to this?

javanlacerda:
> OOOF I think we lose all of our old stats if we land this. How can we preserve them while changing to this?

Done [here](https://github.com/google/clusterfuzz/pull/5080), could you take a look?

PR #5080, opened 2025-12-15
Add back the is_batch label

javanlacerda:
Add back the is_batch label to avoid missing the metric.


javanlacerda:
> Can you leave a comment to say why we leaving it.

Done :)

PR #5087, opened 2025-12-15
Update AGENTS.md with running code instructions

jonathanmetzman:
Added instructions for running code with ClusterFuzz library.


jonathanmetzman:
This is probably a bad idea.

PR #5090, opened 2025-12-16
[batch] Add congestion jobs.

jonathanmetzman:
The idea is if we schedule a bunch of NOOP jobs and we don't see
many in the last hour, that things are congested.
This might not react fast enough to congestion, but can potentially
pause things for long enough to drain the queue.
In the future we should stop hardcoding and infer how many we should
see completed in the last hour.
Other future improvements can be testing congestion by subconfig
so we can tell if some other resource is saturated.

### DIFF
--- a/src/clusterfuzz/_internal/cron/schedule_fuzz.py
+++ b/src/clusterfuzz/_internal/cron/schedule_fuzz.py
@@ -26,10 +26,10 @@ from googleapiclient import discovery
 
 from clusterfuzz._internal.base import tasks
 from clusterfuzz._internal.base import utils
-from clusterfuzz._internal.batch import gcp as batch
 from clusterfuzz._internal.config import local_config
 from clusterfuzz._internal.datastore import data_types
 from clusterfuzz._internal.datastore import ndb_utils
+from clusterfuzz._internal.google_cloud_utils import batch
 from clusterfuzz._internal.google_cloud_utils import credentials
 from clusterfuzz._internal.metrics import logs
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -93,3 +93,11 @@ python butler.py format
 
 This will format the changed code in your current branch.
 It's possible to get into a state where linting and formatting contradict each other. In this case, STOP, the human will fix it.
+
+## Codebase Notes
+
+### Batch Logic
+
+- `src/clusterfuzz/_internal/batch/gcp.py` contains low-level GCP Batch client logic. `check_congestion_jobs` is placed here as it directly queries job status using the client.
+- `src/clusterfuzz/_internal/batch/service.py` contains high-level batch service logic, including configuration management. `create_congestion_job` is placed here because it depends on configuration logic (`_get_specs_from_config`, etc.).
+- `src/clusterfuzz/_internal/google_cloud_utils/batch.py` acts as a facade/wrapper for backward compatibility or convenience, delegating to `gcp.py` and `service.py`.
--- a/src/clusterfuzz/_internal/batch/gcp.py
+++ b/src/clusterfuzz/_internal/batch/gcp.py
@@ -154,6 +154,29 @@ def count_queued_or_scheduled_tasks(project: str,
   return (queued, scheduled)
 
 
+def check_congestion_jobs(job_ids: List[str]) -> int:
+  """Checks the status of the congestion jobs."""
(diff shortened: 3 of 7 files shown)


src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -377,26 +378,88 @@ def respect_project_max_cpus(num_cpus):
...
+      except Exception:
+        logs.error(f'Failed to create congestion job in {region}.')
+
+
 def schedule_fuzz_tasks() -> bool:
   """Schedules fuzz tasks."""
-  multiprocessing.set_start_method('spawn')
+  try:
+    multiprocessing.set_start_method('spawn')
+  except RuntimeError:
```
javanlacerda: Do we want to ignore this?

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -377,26 +378,88 @@ def respect_project_max_cpus(num_cpus):
   return min(max_cpus_per_schedule, num_cpus)
 
 
+def get_congested_regions() -> List[str]:
+  """Returns a list of congested regions."""
```
javanlacerda: Could you describe better in the docstring the strategy to detect when a region is congested?

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -377,26 +378,88 @@ def respect_project_max_cpus(num_cpus):
...
+    jobs.sort(key=lambda j: j.timestamp, reverse=True)
+    # Check the last 3 jobs.
+    recent_jobs = jobs[:3]
+    if len(recent_jobs) < 3:
+      continue
+
+    completed_count = batch.check_congestion_jobs(
+        [job.job_id for job in recent_jobs])
+    if completed_count < 3:
+      logs.error(f'Congestion detected in {region}: {completed_count}/3 '
```
javanlacerda, line 417: Should this be logged as an error?

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -377,26 +378,88 @@ def respect_project_max_cpus(num_cpus):
...
+                 'congestion jobs completed in the last hour.')
+      congested_regions.append(region)
+  return congested_regions
+
+
+def schedule_congestion_jobs(fuzz_tasks, all_regions):
+  """Schedules congestion jobs for all regions."""
+  # Run a hello world task that finishes very quickly. We need job, pick any.
```
javanlacerda: Not clear this comment

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -377,26 +378,88 @@ def respect_project_max_cpus(num_cpus):
...
-  if not available_cpus:
-    return False
 
-  fuzz_tasks = get_fuzz_tasks(available_cpus)
-  if not fuzz_tasks:
-    logs.error('No fuzz tasks found to schedule.')
-    return False
+  fuzz_tasks = []
+  if available_cpus > 0:
+    fuzz_tasks = get_fuzz_tasks(available_cpus)
```
javanlacerda: I think the get_fuzz_tasks function should be the one to validate the `available_cpus`

src/clusterfuzz/_internal/cron/schedule_fuzz.py
jonathanmetzman: Yes. I've left a comment explaining.

src/clusterfuzz/_internal/cron/schedule_fuzz.py
jonathanmetzman: Yes. It means something needs to be fixed. In face I've left a comment saying there should be monitioring here even.

src/clusterfuzz/_internal/cron/schedule_fuzz.py
jonathanmetzman: I've clarified.

jonathanmetzman:
Done. PTAL again

jonathanmetzman:
Actually it's not reay yet. Still dealing with the merge conflict

PR #5107, opened 2026-01-05
Fix GitHub rate limit in project_setup by using recursive tree fetch

hunsche:
## Description
This PR addresses the GitHub API rate limiting issues encountered during the `project_setup` cron job execution for OSS-Fuzz projects.

### Problem
The previous implementation of `get_oss_fuzz_projects` made multiple API calls per project (fetching directory trees individually), resulting in O(N) requests where N is the number of projects. This frequently triggered GitHub's rate limits (403 Forbidden).

### Solution
Optimized the project fetching logic to use the `recursive=1` parameter when fetching the OSS-Fuzz repository tree.
- Fetches the entire directory tree in a single API call (O(1)).
- Uses SHA from the tree to cache `project.yaml` contents, skipping unchanged files.
- Disables retries specifically for 403 errors to prevent retry storms.
- Parses the flattened tree structure to identify projects with `project.yaml` and `Dockerfile`.
- Significantly reduces the API request count, mitigating rate limit issues.

### Verification
- **Regression Testing**: Ran existing tests in `src/clusterfuzz/_internal/tests/appengine/handlers/cron/project_setup_test.py`. All passed.
- **New Tests**: Added `test_get_oss_fuzz_projects_api_error` to verify error handling when the API call fails.
- **Mock Data**: Updated `url_results.txt` to include a mock response for the recursive tree fetch, covering edge cases like:
    - Invalid/Malformed YAML.
    - Nested files (ignored).
    - Files at root (ignored).
    - Projects without YAML (ignored).

### Impact
- **Performance**: Drastic reduction in GitHub API calls.
- **Reliability**: Reduced likelihood of cron job failures due to rate limiting.
- **Scope**: Changes are isolated to `project_setup.py` and only affect the OSS-Fuzz project source flow.

### DIFF
--- a/src/clusterfuzz/_internal/cron/project_setup.py
+++ b/src/clusterfuzz/_internal/cron/project_setup.py
@@ -81,6 +81,14 @@ class ProjectSetupError(Exception):
   """Exception."""
 
 
+class GitHubGenericError(Exception):
+  """GitHub generic error."""
+
+
+class GitHubRateLimitError(Exception):
+  """GitHub rate limit error."""
+
+
 class JobInfo:
   """Job information."""
 
@@ -219,7 +227,11 @@ def _to_experimental_job(job_info):
   return job_info
 
 
-@retry.wrap(retries=3, delay=2, function='cron.project_setup.get_github_url')
+@retry.wrap(
+    retries=3,
+    delay=2,
+    function='cron.project_setup.get_github_url',
+    exception_types=[GitHubGenericError])
 def get_github_url(url):
   """Return contents of URL."""
   github_credentials = db_config.get_value('github_credentials')
@@ -229,10 +241,14 @@ def get_github_url(url):
   client_id, client_secret = github_credentials.strip().split(';')
   response = requests.get(
       url, auth=(client_id, client_secret), timeout=HTTP_TIMEOUT_SECONDS)
+
+  if response.status_code == 403:
+    raise GitHubRateLimitError(f'GitHub rate limit exceeded for {url}.')
+
   if response.status_code != 200:
     logs.error(
         f'Failed to get github url: {url}.', status_code=response.status_code)
-    response.raise_for_status()
+    raise GitHubGenericError(f'Failed to get github url: {url}.')
 
   return json.loads(response.text)
 
@@ -259,15 +275,30 @@ def get_oss_fuzz_projects():
     filename = parts[2]
 
     if project_name not in project_map:
-      project_map[project_name] = {'yaml_url': None, 'has_dockerfile': False}
+      project_map[project_name] = {
+          'yaml_url': None,
+          'has_dockerfile': False,
+          'yaml_sha': None
+      }
 
     if filename == 'project.yaml':
       project_map[project_name]['yaml_url'] = item['url']
+      project_map[project_name]['yaml_sha'] = item['sha']
     elif filename == 'Dockerfile':
       project_map[project_name]['has_dockerfile'] = True
 
+  # Get all existing projects to check for cache hits.
+  existing_projects = {p.name: p for p in data_types.OssFuzzProject.query()}
+
   for project_name, details in project_map.items():
     if not details['yaml_url']:
+      continue
+
+    # Check if we have a cached version of project.yaml.
+    existing_project = existing_projects.get(project_name)
+    if (existing_project and existing_project.project_yaml_sha and
+        existing_project.project_yaml_sha == details['yaml_sha']):
+      # Cache hit.
       continue
 
     try:
@@ -282,7 +313,7 @@ def get_oss_fuzz_projects():
     if not has_dockerfile:
       continue
 
-    projects.append((project_name, info))
+    projects.append((project_name, info, details['yaml_sha']))
 
   return projects
 
@@ -294,7 +325,7 @@ def get_projects_from_gcs(gcs_url):
   except json.decoder.JSONDecodeError as e:
     raise ProjectSetupError(f'Error loading json file from {gcs_url}: {e}')
 
-  return [(project['name'], project) for project in data['projects']]
+  return [(project['name'], project, None) for project in data['projects']]
 
 
 def _process_sanitizers_field(sanitizers):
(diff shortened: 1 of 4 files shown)


src/clusterfuzz/_internal/cron/project_setup.py
```diff
@@ -247,34 +247,53 @@ def find_github_item_url(github_json, name):
...
     return []
 
-  tree = get_github_url(projects_url)
+  projects = []
+  project_map = {}
+
   for item in tree['tree']:
-    if item['type'] != 'tree':
+    path = item['path']
+    if not path.startswith('projects/'):
```
decoNR, line 267: According to `url_results.txt` (e.g., line 54), the path contains neither a `/` nor the full path, it simply has the word `projects`. I am flagging this but maybe the examples in `url_results.txt` are incomplete. Do you have more information regarding the required format?

jonathanmetzman:
Could you check doc or verify that the API doesn't paginate or cut off our results? The UI on github hates our project because we have more 1000 subdirs, and we should be sure the API isn't similarly bad

jonathanmetzman:
https://api.github.com/repos/google/oss-fuzz/git/trees/master?recursive=1 looks fine to me

src/clusterfuzz/_internal/tests/appengine/handlers/cron/project_setup_test.py
```diff
@@ -1680,6 +1680,12 @@ def test_get_oss_fuzz_projects(self):
             }
         })])
 
+  def test_get_oss_fuzz_projects_api_error(self):
```
jonathanmetzman, line 1688: I'm not sure what this test is proving for us?

src/clusterfuzz/_internal/cron/project_setup.py
hunsche: The line 54 in [url_results.txt](cci:7://file:///usr/local/google/home/matheushunsche/projects/clusterfuzz/src/clusterfuzz/_internal/tests/appengine/handlers/cron/project_setup_data/url_results.txt:0:0-0:0) corresponds to the standard (non-recursive) tree fetch, which was used in the previous implementation.

In this PR, I changed the API call to use `?recursive=1` (see [project_setup.py](cci:7://file:///usr/local/google/home/matheushunsche/projects/clusterfuzz/src/clusterfuzz/_internal/cron/project_setup.py:0:0-0:0)). When `recursive=1` is used, the GitHub API returns a flat list of all files with their full paths relative to the repository root (e.g., `projects/boringssl/project.yaml`), rather than a nested tree structure.

I added a new mock entry at the end of [url_results.txt](cci:7://file:///usr/local/google/home/matheushunsche/projects/clusterfuzz/src/clusterfuzz/_internal/tests/appengine/handlers/cron/project_setup_data/url_results.txt:0:0-0:0) (key: `.../git/trees/master?recursive=1`) that reflects this flattened structure with full paths, which matches what the new code expects. The existing data at line 54 is for the old non-recursive call and is not used by the new logic.

src/clusterfuzz/_internal/tests/appengine/handlers/cron/project_setup_test.py
hunsche: It verifies that the function handles API failures gracefully by returning an empty list instead of raising an exception, ensuring the cron job doesn't crash on transient GitHub errors.

hunsche:
> Could you check doc or verify that the API doesn't paginate or cut off our results? The UI on github hates our project because we have more 1000 subdirs, and we should be sure the API isn't similarly bad

The GitHub API documentation confirms that the limit for recursive tree requests is **100,000 entries** (files and directories) or **7MB** of data, which is significantly higher than the 1,000 limit often seen in the UI.

I verified this manually by running the script against the production repo, and it successfully retrieved **1293 projects**, confirming that we are not being truncated at 1000 entries. The API response also includes a `truncated` boolean field that we can check if we ever approach the 100k limit.

src/clusterfuzz/_internal/tests/appengine/handlers/cron/project_setup_test.py
jonathanmetzman: I think it's actually testing that the system fails quietly when it would be better to fail loudly. Can you change the code so that there is a retry and the exception bubbles up?

src/clusterfuzz/_internal/cron/project_setup.py
```diff
@@ -247,34 +247,53 @@ def find_github_item_url(github_json, name):
 def get_oss_fuzz_projects():
   """Return list of projects for oss-fuzz."""
   ossfuzz_tree_url = ('https://api.github.com/repos/google/oss-fuzz/'
-                      'git/trees/master')
-  tree = get_github_url(ossfuzz_tree_url)
-  projects = []
-
-  projects_url = find_github_item_url(tree, 'projects')
```
jonathanmetzman, line 254: find_github_item_url is dead code and can be removed right?

jonathanmetzman:
Good find. I see we are already at 2.4 MB of data in this JSON. It's not inconceivable that in a few years the new limit is hit. Maybe leave a TODO to an issue?

hunsche:
> Good find. I see we are already at 2.4 MB of data in this JSON. It's not inconceivable that in a few years the new limit is hit. Maybe leave a TODO to an issue?

That's a fair point. Given the current rate limit blocking us, this seems like the best tradeoff for now. I'll create a new issue to track this potential future limit (100k entries or 7MB) and explore alternatives if we get close to it.

src/clusterfuzz/_internal/cron/project_setup.py
hunsche: Done

src/clusterfuzz/_internal/cron/project_setup.py
decoNR: Thank you for the explanation.

src/clusterfuzz/_internal/tests/appengine/handlers/cron/project_setup_test.py
hunsche: Done

hunsche: Actually, I had to revert the retry logic for 403 errors because it was causing a 'retry storm' and triggering secondary rate limits. Instead, I implemented SHA-based caching for `project.yaml` to significantly reduce API calls (O(1) fetch) and disabled retries specifically for 403s to fail fast and safe.

jonathanmetzman (approved):
lgtm

PR #5108, opened 2026-01-06
[TMP] Enable gcloud storage on mac and android

ViniciustCosta:
Directly enables the gcloud storage flag for MAC and Android bots.

This is done because, for internal/chrome deployment, the init scripts that start ClusterFuzz are stored at chrome infra's repositories, so setting/reverting these env vars in these scripts would have a much higher overhead.

### DIFF
--- a/src/clusterfuzz/_internal/system/environment.py
+++ b/src/clusterfuzz/_internal/system/environment.py
@@ -1020,6 +1020,12 @@ def set_bot_environment():
   from clusterfuzz._internal.config import local_config
   local_config.ProjectConfig().set_environment()
 
+  # Tmp: set gcloud_storage flag for android/MAC bots
+  # to validate the gsutil migration on these platforms.
+  if is_android() or platform() == 'MAC':
+    os.environ['USE_GCLOUD_STORAGE_CP'] = 'True'
+    os.environ['USE_GCLOUD_STORAGE_RSYNC'] = '1'
+
   # Success.
   return True
 


src/clusterfuzz/_internal/system/environment.py
```diff
@@ -1020,6 +1020,12 @@ def set_bot_environment():
   from clusterfuzz._internal.config import local_config
   local_config.ProjectConfig().set_environment()
 
+  # Tmp: set gcloud_storage flag for android/MAC bots
+  # to validate the gsutil migration on these platforms.
+  if is_android() or platform() == 'MAC':
+    os.environ['USE_GCLOUD_STORAGE_CP'] = 'True'
```
hunsche, line 1026: I'm almost certain that using `1` is the correct approach for all of them. The patterns currently set to `True` likely need to be changed to `1`.

ViniciustCosta: Yeah, I used the same pattern as it was for these flags in other parts of the codebase. Actually, they had set this flag to 'true' with a lowercase 't', but to be strictly correct, it should have been with a capital 'T'. This is true for several other ClusterFuzz environment variables (something we need to fix).

If you look at the `environment.get_value()` method, it calls `ast.literal_eval`, which tries to cast it to the correct type, so 'True' becomes the boolean True. It keeps 'true' as a string; so when you call `bool()` on that string, anything that isn't empty returns True.

PR #5113, opened 2026-01-08
Kubernetes Job service

javanlacerda:
This PR introduces full support for scheduling and managing fuzzing tasks on Kubernetes clusters,
  specifically targeting GKE. It implements a new KubernetesService to
  handle batch job creation, supports Kata Containers for isolation, and includes robust testing
  and configuration mechanisms.

  Key Features:
   * Kubernetes Service: A new backend for RemoteTaskInterface that schedules tasks as Kubernetes
     Jobs. It supports both standard and Kata Container runtimes, automatic Service Account
     creation with Workload Identity, and intelligent job limiting to prevent cluster overload.
   * Traffic Shifting (RemoteTaskGate): A new gating mechanism (RemoteTaskGate) that intelligently
     routes tasks between the legacy GCP Batch service and the new Kubernetes service based on
     configurable probabilities, allowing for a gradual, controlled migration.
   * Feature Flags: A new dynamic configuration system backed by Datastore to control runtime
     behaviors like job concurrency limits.

  Detailed Changes by Module:

   * Kubernetes Integration (`src/clusterfuzz/_internal/k8s/`):
       * service.py: Implemented KubernetesService for job lifecycle management (creation,
         monitoring, limiting). Includes GKE credential loading, Kata Container spec generation,
         and Service Account provisioning.
       * Tests: Added k8s_service_test.py (unit), k8s_service_limit_test.py (limits), and
         k8s_service_e2e_test.py (integration on Kind).

   * Remote Task Management (`src/clusterfuzz/_internal/remote_task/`):
       * __init__.py: Introduced RemoteTaskGate, a smart router that implements
         RemoteTaskInterface. It initializes both GcpBatchService and KubernetesService and
         distributes tasks between them based on probabilities defined in job_frequency.py. This
         enables traffic splitting (e.g., 10% to K8s, 90% to Batch) for safe rollout.
       * job_frequency.py: Added logic to manage task scheduling frequency and split ratios.
       * Refactored core task logic to use the generic RemoteTask and RemoteTaskInterface
         abstractions.

   * Datastore & Configuration (`src/clusterfuzz/_internal/datastore/`):
       * data_types.py: Added FeatureFlag model to store configuration dynamically.
       * feature_flags.py: Added FeatureFlags enum/helper for type-safe access to flags (e.g.,
         K8S_PENDING_JOBS_LIMITER).

   * Batch & Legacy Refactoring (`src/clusterfuzz/_internal/batch/`):
       * service.py: Updated to align with the new RemoteTask interface.
       * Removed obsolete gcp.py and google_cloud_utils/batch.py utilities in favor of the new
         structure.

   * Infrastructure & CI:
       * .github/workflows/kubernetes-e2e-tests.yaml: New workflow for running E2E tests on a Kind
         cluster.
       * Pipfile / src/Pipfile: Added kubernetes client and updated Google Cloud dependencies.

   * Bot & Metrics:
       * src/python/bot/startup/run_bot.py: Updates to support K8s-based bot execution via the new
         gate.
       * src/clusterfuzz/_internal/metrics/: Enhanced logging and monitoring for remote tasks.

Evidences:

image
Batch and Kata containers fuzzing hours, proving the Remote Gate, the Batch and Kubernetes services are working properly. 
The Feature Flag is used to set the job_frequency, then it proves the feature flag and its usage is working as well.

### DIFF
--- a/src/clusterfuzz/_internal/k8s/service.py
+++ b/src/clusterfuzz/_internal/k8s/service.py
@@ -35,7 +35,8 @@ from clusterfuzz._internal.metrics import logs
 from clusterfuzz._internal.remote_task import remote_task_types
 from clusterfuzz._internal.system import environment
 
-CLUSTER_NAME = 'clusterfuzz-cronjobs-gke'
+CLUSTER_NAME = project_config = local_config.ProjectConfig().get(
+    'cluster_name', 'clusterfuzz-cronjobs-gke')
 
 KubernetesJobConfig = collections.namedtuple('KubernetesJobConfig', [
     'job_type',


Pipfile
```diff
@@ -10,6 +10,19 @@ future = "==0.17.1"
...
+google-cloud-storage = "==2.10.0"
+google-cloud-secret-manager = "==2.17.0"
+google-cloud-logging = "==3.6.0"
+google-cloud-monitoring = "==2.15.1"
+google-cloud-datastore = "==2.16.1"
+oauth2client = "==4.1.3"
+requests = "==2.21.0"
+PyYAML = "==6.0"
+httplib2 = "==0.19.0"
+google-auth-oauthlib = "==0.4.1"
```
jonathanmetzman: I think the dependencies that get installed on the images generally come from src/Pipfile and these are more the dev ones? I'm pretty sure we don't need to be adding requests here for example.

butler.py
```diff
@@ -435,7 +435,6 @@ def main():
       'clean_indexes', help='Clean up undefined indexes (in index.yaml).')
   parser_clean_indexes.add_argument(
       '-c', '--config-dir', required=True, help='Path to application config.')
-
   parser_create_config = subparsers.add_parser(
```
jonathanmetzman, line 439: Undo this please.

src/clusterfuzz/_internal/tests/core/platforms/__init__.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2019 Google LLC
+# Copyright 2025 Google LLC
```
jonathanmetzman: Let's leave out changes to this file.

src/python/bot/startup/run_bot.py
```diff
@@ -17,6 +17,8 @@
 # to be able to import dependencies directly, but we must store these in
 # subdirectories of common so that they are shared with App Engine.
 from clusterfuzz._internal.base import modules
+from clusterfuzz._internal.remote_task import RemoteTask
```
jonathanmetzman: We generally don't import classes and import modules instead. Per the google python style guide.

src/setup.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020 Google LLC
+# Copyright 2026 Google LLC
```
jonathanmetzman: leave this.

jonathanmetzman (commented):
left some very surface level comments.

butler.py
javanlacerda: Done

src/clusterfuzz/_internal/tests/core/platforms/__init__.py
javanlacerda: Done

src/python/bot/startup/run_bot.py
javanlacerda: Done

src/setup.py
javanlacerda: Done

Pipfile
javanlacerda: Some of them are needed for running the tests, the K8s end to end for example. I'll remove the requests one and check its required for any test

jonathanmetzman: I think the tests will use both, I think this file is for deps that are only used locally.

javanlacerda: You mean like using butler? If so we need all of them for running the `run_remote_task.py` for instance.

src/clusterfuzz/_internal/remote_task/__init__.py
```diff
@@ -30,11 +51,86 @@ class RemoteTaskInterface(abc.ABC):
...
+    
+    The choice of backend is determined by the `_should_use_kubernetes` method.
+    """
+    if self._should_use_kubernetes():
+      return self._kubernetes_service.create_utask_main_job(
+          module, job_type, input_download_url)
+    return self._gcp_batch_service.create_utask_main_job(
+        module, job_type, input_download_url)
+
+  def create_utask_main_jobs(self, remote_tasks: List[RemoteTask]):
```
ViniciustCosta: I feel like this code became very coupled to use only batch and k8s.
I think the idea should be to generalize this to facilitate extending it to any "remote executer". The gate should be able to dispatch tasks to any implemented backend (batch, k8s, gcp run, etc) using the defined remote task interface. As discussed offline, maybe a factory pattern could be useful here, having a factory method construct and return the object for each remote service.

This should also be improved in the job_frequency module, maybe mapping each percentage to the remote service that should receive the task.

src/local/butler/format.py
```diff
@@ -32,10 +32,7 @@ def execute(args):
   """Format changed code."""
   if args.path:
     diff_command = f'git ls-files {os.path.abspath(args.path)}'
-  elif os.path.exists('.git/FETCH_HEAD'):
```
ViniciustCosta, line 35: Maybe not do this in this PR to avoid side effects :)

javanlacerda: Yes, I included it by mistake.

Pipfile
```diff
@@ -10,6 +10,20 @@ future = "==0.17.1"
 protobuf = "==4.23.4"
 psutil = "==5.9.4"
 google-cloud-ndb = "==2.3.4"
+kubernetes = "==34.1.0"
```
jonathanmetzman: I don't know where our thread went but I think this file is the wrong place for these packages and it should be in src/Pipe

src/clusterfuzz/_internal/remote_task/__init__.py
javanlacerda: That's a nice call. I've implemented something similar to it :)

Pipfile
javanlacerda: The thread is here https://github.com/google/clusterfuzz/pull/5113#discussion_r2692355654. I commented that if this is the requirements for local, we need all of this for running the end to end tests, so we should probably keep it.

local/tests/kubernetes_e2e_test.bash
```diff
@@ -11,17 +13,17 @@
...
+pipenv --python 3.11
+pipenv install
 
-class KubernetesJobClient(RemoteTaskInterface):
-  """A remote task execution client for Kubernetes.
-  
-  This class is a placeholder for a future implementation of a remote task
-  execution client that uses Kubernetes. It is not yet implemented.
-  """
+./local/install_deps.bash
```
jonathanmetzman, line 25: This is only intended to be used in CI?

src/clusterfuzz/_internal/base/tasks/__init__.py
```diff
@@ -550,7 +557,8 @@ def lease(self, _event=None):  # pylint: disable=arguments-differ
       leaser_thread.join()
 
     # If we get here the task succeeded in running. Acknowledge the message.
-    self._pubsub_message.ack()
+    if not self.do_not_ack:
```
jonathanmetzman: What is this change?

src/clusterfuzz/_internal/datastore/feature_flags.py
```diff
@@ -0,0 +1,70 @@
+# Copyright 2024 Google LLC
```
jonathanmetzman: Year

src/clusterfuzz/_internal/k8s/__init__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2025 Google LLC
```
jonathanmetzman: year

src/clusterfuzz/_internal/k8s/service.py
```diff
@@ -0,0 +1,352 @@
...
+          task.pubsub_task.do_not_ack = True
+      return []
+
+    job_specs = collections.defaultdict(list)
+    configs = _get_k8s_job_configs(remote_tasks)
+    for remote_task in remote_tasks:
+      logs.info(f'Scheduling {remote_task.command}, {remote_task.job_type}.')
+      config = configs[(remote_task.command, remote_task.job_type)]
+      job_specs[config].append(remote_task.input_download_url)
+    logs.info('Creating batch jobs.')
```
jonathanmetzman, line 302: Shouldn't this be k8s?

src/clusterfuzz/_internal/k8s/service.py
```diff
@@ -0,0 +1,352 @@
...
+    return job_body['metadata']['name']
+
+  def _get_pending_jobs_count(self) -> int:
+    """Returns the number of pending jobs."""
+    try:
+      pods = self._core_api.list_namespaced_pod(
+          namespace='default',
+          label_selector='app.kubernetes.io/name=clusterfuzz-kata-job',
+          field_selector='status.phase=Pending')
+      logs.info(f"Found {len(pods.items)} pending jobs.")
```
jonathanmetzman: single quotes.

src/clusterfuzz/_internal/k8s/service.py
```diff
@@ -0,0 +1,352 @@
...
+    configs[(task.command, task.job_type)] = config
+
+  return configs
+
+
+def _create_job_body(config: KubernetesJobConfig, input_url: str,
+                     service_account_name: str) -> dict:
+  """Creates the body of a Kubernetes job."""
+
+  job_name = config.job_type.replace('_', '-') + '-' + str(uuid.uuid4()).split(
```
jonathanmetzman: I think at OSS-Fuzz scale there is actually a high chance of a collision in a single day (>50%, https://en.wikipedia.org/wiki/Birthday_problem).
How bad is that? Why can't we use the full uuid?
If there's a length problem maybe we want to truncate the job name?

src/clusterfuzz/_internal/k8s/service.py
```diff
@@ -0,0 +1,352 @@
...
+      '-', maxsplit=1)[0]
```
jonathanmetzman: Why do maxsplit? It seems unnatural and unnecessary

src/clusterfuzz/_internal/k8s/service.py
```diff
@@ -0,0 +1,352 @@
...
+      clusters = response.get('clusters', [])
+      cluster = next((c for c in clusters if c['name'] == CLUSTER_NAME), None)
+
+      if not cluster:
+        logs.error(f"Cluster {CLUSTER_NAME} not found in project {project}.")
+        print(f"DEBUG: Cluster {CLUSTER_NAME} not found in project {project}.")
+        return
+
+    except Exception as e:
+      logs.error(f"Failed to list clusters in {project}: {e}")
```
jonathanmetzman: Single quotes.

src/clusterfuzz/_internal/batch/service.py
```diff
@@ -18,24 +18,168 @@
...
 from clusterfuzz._internal.base.tasks import task_utils
-from clusterfuzz._internal.batch.data_structures import BatchTask
-from clusterfuzz._internal.batch.data_structures import BatchWorkloadSpec
-from clusterfuzz._internal.batch.gcp import GcpBatchClient
 from clusterfuzz._internal.config import local_config
 from clusterfuzz._internal.datastore import data_types
 from clusterfuzz._internal.datastore import ndb_utils
+from clusterfuzz._internal.google_cloud_utils import credentials
 from clusterfuzz._internal.metrics import logs
+from clusterfuzz._internal.remote_task import types
```
jonathanmetzman: This shadows a python builtin module. Can you rename it to remote_task_types.

Pipfile
jonathanmetzman: It's the wrong place for this because this is local only deps, the deps in src/Pipfile goes into both prod and local installs?

src/clusterfuzz/_internal/k8s/service.py
```diff
@@ -0,0 +1,352 @@
...
+    return result[0]
+
+  def create_utask_main_jobs(self, remote_tasks: typing.List[types.RemoteTask]):
+    """Creates a batch job for a list of uworker main tasks.
+
+    This method groups the tasks by their workload specification and creates a
+    separate batch job for each group. This allows tasks with similar
+    requirements to be processed together, which can improve efficiency.
+    """
+    if feature_flags.FeatureFlags.K8S_PENDING_JOBS_LIMITER.enabled and \
```
jonathanmetzman: This condition is very hard to read. Can you break it up?

src/clusterfuzz/_internal/k8s/service.py
```diff
@@ -0,0 +1,352 @@
...
+    configs = _get_k8s_job_configs(remote_tasks)
+    for remote_task in remote_tasks:
+      logs.info(f'Scheduling {remote_task.command}, {remote_task.job_type}.')
+      config = configs[(remote_task.command, remote_task.job_type)]
+      job_specs[config].append(remote_task.input_download_url)
+    logs.info('Creating batch jobs.')
+    jobs = []
+    logs.info('Batching utask_mains.')
+    for config, input_urls in job_specs.items():
+      # TODO(javanlacerda): Batch multiple tasks into a single job.
```
jonathanmetzman: Is that actually need for kata?

src/clusterfuzz/_internal/k8s/service.py
```diff
@@ -0,0 +1,352 @@
...
+
+
+def _get_k8s_job_configs(
+    remote_tasks: typing.List[types.RemoteTask]
+) -> typing.Dict[typing.Tuple[str, str], KubernetesJobConfig]:
+  """Gets the configured specifications for a batch workload."""
+
+  if not remote_tasks:
+    return {}
+  #TODO(javanlacerda): Create remote task config
```
jonathanmetzman: nit: Space after `#`

.github/workflows/kubernetes-e2e-tests.yaml
```diff
@@ -0,0 +1,42 @@
...
+on: [pull_request]
+
+permissions: read-all
+
+jobs:
+  build:
+    runs-on: ubuntu-24.04
+
+    steps:
+      - uses: actions/checkout@v3
```
jonathanmetzman: Why not use v4?

.github/workflows/kubernetes-e2e-tests.yaml
```diff
@@ -0,0 +1,42 @@
...
+          git fetch origin master --depth 1
+          git symbolic-ref refs/remotes/origin/HEAD refs/remotes/origin/master
+
+      - name: Setup python environment
+        uses: actions/setup-python@b55428b1882923874294fa556849718a1d7f2ca5
+        with:
+          python-version: 3.11
+
+      - name: Set up JDK 21
+        uses: actions/setup-java@v3
```
jonathanmetzman: Why not use v4?

src/clusterfuzz/_internal/k8s/service.py
```diff
@@ -0,0 +1,352 @@
...
+    except Exception as e:
+      logs.error(f"Failed to list clusters in {project}: {e}")
+      return
+
+    endpoint = cluster['endpoint']
+    # ca_cert is base64 encoded.
+    ca_cert = base64.b64decode(cluster['masterAuth']['clusterCaCertificate'])
+
+    # Write CA cert to a temporary file.
+    fd, ca_cert_path = tempfile.mkstemp()
```
jonathanmetzman: This should live for the lifetime of the tworker right? Maybe it should be in a known location in the bot directory instead of a temp path?

src/clusterfuzz/_internal/k8s/service.py
```diff
@@ -0,0 +1,352 @@
...
+    'command',
+    'disk_size_gb',
+    'service_account_email',
+    'clusterfuzz_release',
+    'is_kata',
+])
+
+
+def _get_config_names(remote_tasks: typing.List[types.RemoteTask]):
+  """"Gets the name of the configs for each batch_task. Returns a dict
```
jonathanmetzman: Extra quote!

src/clusterfuzz/_internal/k8s/service.py
jonathanmetzman: Also it seems like this is duplicated from create_kata_container_job. Let's try to share code.

src/Pipfile
```diff
@@ -49,8 +51,7 @@ charset-normalizer = "==3.3.2"
 firebase-admin = "==6.2.0"
 Flask = "==2.2.2"
 itsdangerous = "==2.0.1"
-Jinja2 = "==3.1.4"
-PyJWT = "==2.7.0"
+pyjwt = "==2.7.0"
```
jonathanmetzman: Why change this?

Pipfile
```diff
@@ -10,6 +10,20 @@ future = "==0.17.1"
...
+aiohttp = "==3.10.5"
+google-cloud-storage = "==2.10.0"
+google-cloud-secret-manager = "==2.17.0"
+google-cloud-logging = "==3.6.0"
+google-cloud-monitoring = "==2.15.1"
+google-cloud-datastore = "==2.16.1"
+Jinja2 = "==3.1.2"
+oauth2client = "==4.1.3"
+requests = "==2.21.0"
+PyYAML = "==6.0"
```
jonathanmetzman: Why are we using a different version here and in src/Pipfile?

Pipfile
```diff
@@ -10,6 +10,20 @@ future = "==0.17.1"
...
+google-api-python-client = "==2.93.0"
+aiohttp = "==3.10.5"
+google-cloud-storage = "==2.10.0"
+google-cloud-secret-manager = "==2.17.0"
+google-cloud-logging = "==3.6.0"
+google-cloud-monitoring = "==2.15.1"
+google-cloud-datastore = "==2.16.1"
+Jinja2 = "==3.1.2"
+oauth2client = "==4.1.3"
+requests = "==2.21.0"
```
jonathanmetzman: Why are we adding such old versions of requests and httplib2?

src/clusterfuzz/_internal/k8s/job_template.yaml
```diff
@@ -0,0 +1,61 @@
+# Copyright 2026 Google LLC
```
jonathanmetzman, line 1: What's difference between thsi and the next template?

jonathanmetzman: It might have been a good idea to consider knative instead of rebuilding batch.

src/clusterfuzz/_internal/k8s/job_template.yaml
```diff
@@ -0,0 +1,61 @@
...
+            - ALL
+        volumeMounts:
+        - mountPath: /dev/shm
+          name: dshm
+      volumes:
+      - name: dshm
+        emptyDir:
+          medium: Memory
+          sizeLimit: 1.9Gi
+      restartPolicy: Never
```
jonathanmetzman: FYI I do restart for non fuzz jobs.

src/clusterfuzz/_internal/k8s/kata_job_template.yaml
```diff
@@ -0,0 +1,79 @@
...
+        - name: UWORKER_INPUT_DOWNLOAD_URL
+          value: "{{input_url}}"
+        - name: IS_K8S_ENV
+          value: 'true'
+        - name: DISABLE_MOUNTS
+          value: 'true'
+        - name: UPDATE_WEB_TESTS
+          value: 'False'
+      restartPolicy: Never
+      volumes:
```
jonathanmetzman: Why don't we mount dev/shm here?

jonathanmetzman:
This is cool. I maybe would tried cloud run before kata because 1. It is probably less management? 2. It might be more performant because as far as I know doesn't use nested virt.

jonathanmetzman:
Are we using preemptibles btw?

Pipfile
javanlacerda: I see your point, I'll keep it only in the src/Pipefile.

src/Pipfile
javanlacerda: Not needed actually, reverting.

.github/workflows/kubernetes-e2e-tests.yaml
javanlacerda: We can definetly use, updating it.

local/tests/kubernetes_e2e_test.bash
javanlacerda: Yes!

src/clusterfuzz/_internal/tests/core/k8s/k8s_service_e2e_test.py
```diff
@@ -0,0 +1,371 @@
...
+  """End-to-end tests for the Kubernetes service."""
+
+  @classmethod
+  def setUpClass(cls):
+    """Set up the test environment."""
+    if not os.getenv('K8S_E2E'):
+      raise unittest.SkipTest('K8S_E2E environment variable not set.')
+
+    cls.mock_batch_config = mock.Mock()
+    cls.mock_batch_config.get.return_value = 'test-project'
```
decoNR: It appears this is overwritten on line 114 by `cls.mock_batch_config.get.side_effect = get_batch_config`.

src/clusterfuzz/_internal/remote_task/remote_task_adapters.py
```diff
@@ -0,0 +1,44 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Remote task adapters."""
+
+from enum import Enum
+
+from clusterfuzz._internal.batch.service import GcpBatchService
```
ViniciustCosta: We usually import the module, not the class, per: go/pystyle#imports

In this case, as all backends would expose a service module, I guess you could do `from clusturefuzz...batch import service as batch_service`

src/clusterfuzz/_internal/remote_task/remote_task_adapters.py
```diff
@@ -0,0 +1,44 @@
...
+from clusterfuzz._internal.datastore import feature_flags
+from clusterfuzz._internal.k8s.service import KubernetesService
+
+
+class RemoteTaskAdapters(Enum):
```
ViniciustCosta, line 23: Nice!

src/clusterfuzz/_internal/remote_task/remote_task_types.py
```diff
@@ -11,40 +11,41 @@
...
+
+  @abc.abstractmethod
+  def create_utask_main_job(self, module: str, job_type: str,
+                            input_download_url: str):
+    """Creates a single remote task for a uworker main task."""
+    raise NotImplementedError
+
+  @abc.abstractmethod
+  def create_utask_main_jobs(self, remote_tasks: list[RemoteTask]):
+    """Creates a many remote tasks for uworker main tasks."""
```
ViniciustCosta: nit: "Creates many remote jobs ..."

src/clusterfuzz/_internal/remote_task/__init__.py
```diff
@@ -19,22 +19,142 @@
...
-  @abc.abstractmethod
-  def create_job(self, spec, input_urls):
-    """Creates a remote job.
-    
-    This method is responsible for creating a new job in the remote execution
-    environment. It takes a workload specification and a list of input URLs,
-    and returns a representation of the created job.
+  def __init__(self):
+    # Avoiding circular import
+    from clusterfuzz._internal.remote_task import remote_task_adapters
```
ViniciustCosta: What is the circular import avoided here? Is it the case where other module imports both this and the remote_task_adapters?

If that's the case, I think we can expect that other modules won't need to import remote_task_adapter as it should be an abstraction only used by the remote task gate, so maybe we don't need this local import.

src/clusterfuzz/_internal/base/tasks/__init__.py
decoNR: IMO readability would be improved by using `ack` instead of `do_not_ack` (go/tott/764).

javanlacerda: Its part of the job limiter for the Kubernetes service, we can probably use this for implement the job limiter for Batch as well, using the new feature they implemented for us. The rationale behind is if the task cannot be scheduled for Kubernetes because it already reached the limit of jobs, the message should not be acked, allowing the other adapter, such as Batch, to process the message.

src/clusterfuzz/_internal/k8s/service.py
javanlacerda: Yes. fixing

src/clusterfuzz/_internal/batch/service.py
```diff
@@ -93,11 +237,6 @@ def _get_config_names(batch_tasks: List[BatchTask]):
   return config_map
 
 
-def _get_task_duration(command):
-  return tasks.TASK_LEASE_SECONDS_BY_COMMAND.get(command,
-                                                 tasks.TASK_LEASE_SECONDS)
-
-
 WeightedSubconfig = collections.namedtuple('WeightedSubconfig',
```
decoNR: nit: Maybe this can be moved to the top of the file, below `BatchWorkloadSpec`.

src/clusterfuzz/_internal/k8s/service.py
javanlacerda: I refactored it. we're using the entire uuid now. nice catch on that.

src/clusterfuzz/_internal/k8s/service.py
javanlacerda: removed.

src/clusterfuzz/_internal/k8s/service.py
javanlacerda: Its not needed. I don't recall why I added it here

src/clusterfuzz/_internal/k8s/service.py
javanlacerda: Sure

src/clusterfuzz/_internal/datastore/feature_flags.py
```diff
@@ -0,0 +1,70 @@
...
+  def enabled(self):
+    """Check if a feature flag is enabled."""
+    flag = self.flag
+    if not flag:
+      return False
+    return flag.enabled
+
+  @property
+  def content(self):
+    """Get a feature flag value."""
```
decoNR: nit: The description of this function is repeated in the functions below.

src/clusterfuzz/_internal/batch/service.py
javanlacerda: nice call!

src/clusterfuzz/_internal/batch/service.py
javanlacerda: nice call.

Pipfile
javanlacerda: We was, but I rolled it back to update only the src/Pipfile.

Pipfile
javanlacerda: fixed

src/clusterfuzz/_internal/k8s/job_template.yaml
javanlacerda: Initially I created different templates for raw kubernetes jobs and for Jobs over Kata, but I updated it for having a single template with conditionals.

src/clusterfuzz/_internal/k8s/kata_job_template.yaml
javanlacerda: we should, I centralized it in a single template that does mount it.

src/clusterfuzz/_internal/k8s/job_template.yaml
javanlacerda: Great call, I'll update the template to set it conditionally on the task.

src/clusterfuzz/_internal/k8s/job_template.yaml
javanlacerda: About the Knative, it seems good, but I wound't like to tackle it in this PR as it's working fine as is, but we should definetly explore it.

src/clusterfuzz/_internal/datastore/feature_flags.py
javanlacerda: yeppp. fixing it

src/clusterfuzz/_internal/remote_task/remote_task_adapters.py
javanlacerda: make sense! thanks for that.

src/clusterfuzz/_internal/remote_task/__init__.py
javanlacerda: nice catch on that, even the problem is not the adapters being imported in a different place, but I managed to avoid the circular import keeping the usual import pattern.

src/clusterfuzz/_internal/tests/core/k8s/k8s_service_e2e_test.py
javanlacerda: Its not because this one set the return_value and in the line 114 set the side_effect.

decoNR: But `side_effect` also relates to the return value since it overrides the function, right? And `side_effect` takes precedence over `return_value` (https://docs.python.org/3/library/unittest.mock.html#order-of-precedence-of-side-effect-return-value-and-wraps).

javanlacerda: Intresting! indeed I managed to remove the  `cls.mock_batch_config.get.return_value = 'test-project'` and continues working. Thanks for that!

javanlacerda:
> Are we using preemptibles btw?

We're not, and will it's not part of the plan using it for the clusters. You can see more details on [go/clusterfuzz-to-kubernetes](http://goto.google.com/clusterfuzz-to-kubernetes)

src/clusterfuzz/_internal/base/tasks/__init__.py
```diff
@@ -548,9 +554,9 @@ def lease(self, _event=None):  # pylint: disable=arguments-differ
     finally:
       _event.set()
       leaser_thread.join()
-
+    
     # If we get here the task succeeded in running. Acknowledge the message.
-    self._pubsub_message.ack()
+    self._pubsub_message.ack()
```
jonathanmetzman: Trailing spaces.

decoNR: nit: It seems that you added unintended spaces.

src/clusterfuzz/_internal/base/feature_flags.py
```diff
@@ -0,0 +1,69 @@
+# Copyright 2026 Google LLC
```
jonathanmetzman, line 1: Hmmm...I'm not sure feature_flags module should be in datastore/

src/clusterfuzz/_internal/base/feature_flags.py
```diff
@@ -0,0 +1,69 @@
...
+"""FeatureFlags."""
+
+from enum import Enum
+
+from google.cloud import ndb
+
+from clusterfuzz._internal.datastore import data_types
+
+
+class FeatureFlags(Enum):
```
jonathanmetzman, line 23: Should we be caching any of this?

src/clusterfuzz/_internal/k8s/service.py
jonathanmetzman: `""""Gets`
should be `"""Gets`

src/clusterfuzz/_internal/k8s/service.py
```diff
@@ -0,0 +1,310 @@
...
+      clusters = response.get('clusters', [])
+      cluster = next((c for c in clusters if c['name'] == CLUSTER_NAME), None)
+
+      if not cluster:
+        logs.error(f"Cluster {CLUSTER_NAME} not found in project {project}.")
+        print(f"DEBUG: Cluster {CLUSTER_NAME} not found in project {project}.")
+        return
+
+    except Exception as e:
+      logs.error(f"Failed to list clusters in {project}: {e}")
```
jonathanmetzman: nit: single quotes here and rest of the file (other than docstrings)

src/clusterfuzz/_internal/k8s/service.py
jonathanmetzman: Ping

src/clusterfuzz/_internal/base/tasks/__init__.py
jonathanmetzman: ACK

src/clusterfuzz/_internal/k8s/service.py
```diff
@@ -0,0 +1,310 @@
...
+from clusterfuzz._internal.base import utils
+from clusterfuzz._internal.base.tasks import task_utils
+from clusterfuzz._internal.config import local_config
+from clusterfuzz._internal.datastore import data_types
+from clusterfuzz._internal.datastore import ndb_utils
+from clusterfuzz._internal.metrics import logs
+from clusterfuzz._internal.remote_task import remote_task_types
+from clusterfuzz._internal.system import environment
+
+CLUSTER_NAME = 'clusterfuzz-cronjobs-gke'
```
jonathanmetzman: Nit: Should this be loaded from a config? I think it's fine. WDYT?

jonathanmetzman (approved):

src/clusterfuzz/_internal/k8s/service.py
ViniciustCosta: +1 for setting it in config

src/python/bot/startup/run_bot.py
```diff
@@ -17,6 +17,8 @@
 # to be able to import dependencies directly, but we must store these in
 # subdirectories of common so that they are shared with App Engine.
 from clusterfuzz._internal.base import modules
+from clusterfuzz._internal.remote_task import remote_task_gate
```
ViniciustCosta: Probably should import these alongside the other modules after the `modules.fix_module_search_paths()`

src/clusterfuzz/_internal/base/tasks/__init__.py
javanlacerda: yep. fixed

javanlacerda: fixed

src/clusterfuzz/_internal/base/feature_flags.py
javanlacerda: hmmm. maybe under base? I'll move it

src/clusterfuzz/_internal/base/feature_flags.py
javanlacerda: What do you mean?

src/python/bot/startup/run_bot.py
javanlacerda: definetly.

src/clusterfuzz/_internal/k8s/service.py
javanlacerda: Done!

PR #5122, opened 2026-01-14
Fully migrate gsutil module to gcloud storage

ViniciustCosta:
Fully migrate the gsutil module to use gcloud storage by:
* Removing the feature flags used during the migration;
* Migrating the name for the gcloud storage runner;
* Fixing some inconsistencies in the gcloud storage runner to match the behavior used by gsutil (e.g., add --recursive to rsync, add quiet flag `-q`, etc);
* Adjusting the unit tests

Temporarily retained the `gsutil.py` filename to facilitate revision, as changing the filename would highlight all the code and not only the actual changes (to be done in a follow-up PR). Temporarily retained the `GSUTIL_PATH` as a fallback, since changing this to `GCLOUD_PATH` would depend on landing changes on infra init/setup scripts.

Tests
* All calls from the `gsutil.py` module are already using `gcloud storage` in our workflows in prod (including mac and android bots). 
  * Internal: https://screenshot.googleplex.com/8UjftM5DFHcUcWH.png
  * External and google3 have no usage from this module (https://screenshot.googleplex.com/5cEeEu9p7RQ7squ.png and https://screenshot.googleplex.com/BMoHPdUuMxxrpe8.png)
  
* Testing this in staging before deploying (as dev is not currently synce with master).


src/clusterfuzz/_internal/tests/core/fuzzing/corpus_manager_test.py
```diff
@@ -65,16 +65,18 @@ def test_rsync_to_disk(self, _):
     self.assertTrue(corpus.rsync_to_disk('/dir', timeout=60))
 
     self.assertEqual(self.mock.Popen.call_args[0][0], [
-        '/gsutil_path/gsutil', '-m', '-o', 'GSUtil:parallel_thread_count=16',
-        '-q', 'rsync', '-r', '-d', 'gs://bucket/', '/dir'
+        '/gcloud_path/gcloud', 'storage', '--no-user-output-enabled', '-q',
```
sushantkhanna, line 68: For the `-q` flag, was this placement tested ?  
Also, the original command does not seem to have the equivalent of `--no-user-output-enabled` ?

src/clusterfuzz/_internal/tests/core/fuzzing/corpus_manager_test.py
```diff
@@ -105,17 +109,19 @@ def test_upload_files(self):
     self.assertTrue(corpus.upload_files(['/dir/a', '/dir/b']))
 
     self.assertEqual(self.mock.Popen.call_args[0][0], [
-        '/gsutil_path/gsutil', '-m', '-o', 'GSUtil:parallel_thread_count=16',
-        'cp', '-I', 'gs://bucket/'
+        '/gcloud_path/gcloud', 'storage', '--user-output-enabled', '-q', 'cp',
```
sushantkhanna, line 112: Just curious since previous commands were `--(no)-user-output-enabled`, but this one is enabling it and agai with a `-q` option when the original `gsutil` command did not seem to specify that.  
This is intentional and tested ?

src/clusterfuzz/_internal/tests/core/google_cloud_utils/gsutil_test.py
```diff
@@ -25,378 +23,306 @@
...
+        '--no-user-output-enabled'
+    ]
+    additional_args += ['-q'] if quiet else []
+    return additional_args
+
+  def test_rsync(self):
+    """Test remote rsync."""
+    self.gcloud_runner_obj.rsync('gs://source_bucket/source_path',
+                                 'gs://target_bucket/target_path')
+    expected_args = [
```
sushantkhanna, line 48: The original `gsutil` command had the `-q` option as well.

src/clusterfuzz/_internal/google_cloud_utils/gsutil.py
```diff
@@ -11,86 +11,44 @@
...
 def get_gcloud_path():
   """Get path to gcloud executable."""
   gcloud_executable = 'gcloud'
   if environment.platform() == 'WINDOWS':
     gcloud_executable += '.cmd'
 
+  gcloud_storage_dir = environment.get_value('GCLOUD_PATH')
+  if not gcloud_storage_dir:
+    # Fallback to older GSUTIL_PATH, if needed.
+    gcloud_storage_dir = environment.get_value('GSUTIL_PATH')
```
sushantkhanna, line 37: But would this not mean that you also fallback to the older `gsutil` commands ?

src/clusterfuzz/_internal/google_cloud_utils/gsutil.py
```diff
@@ -194,39 +140,48 @@ def rsync(self,
...
+    rsync_command = [
+        'rsync',
         _filter_path(source, write=True),
-        _filter_path(destination, write=True),
-    ])
+        _filter_path(destination, write=True)
+    ]
 
-    return self.run_gsutil(command, use_gcloud, timeout=timeout, quiet=True)
+    if recursive:
```
sushantkhanna, line 166: Since I am not entirely sure of all the possible scenarios, why do we want to make this flag dependent on an input, when for `gsutil` this was always set ?

src/clusterfuzz/_internal/google_cloud_utils/gsutil.py
```diff
@@ -235,100 +190,78 @@ def upload_file(self,
...
-        cp_command.append('--gzip-in-flight-all')
-
-      cp_command.extend([file_path, _filter_path(gcs_url, write=True)])
-      result = self.run_gsutil(cp_command, use_gcloud, timeout=timeout)
+    cp_command = ['cp']
+    if gzip:
+      cp_command.append('--gzip-local-all')
+    cp_command.extend([file_path, _filter_path(gcs_url, write=True)])
+
+    result = self.run_gcloud_storage(cp_command, timeout=timeout)
```
sushantkhanna, line 204: You might want to put some checks on the stdout/stderr prints as well.

src/clusterfuzz/_internal/google_cloud_utils/gsutil.py
```diff
@@ -235,100 +190,78 @@ def upload_file(self,
...
+    if result.return_code != 0:
+      logs.error('GCloudStorageRunner.upload_file (cp step) failed:\n'
+                 f'Command: {result.command}\n'
+                 f'Filename: {file_path}\n'
+                 f'Output: {result.output}')
+      return False
 
+    # For gcloud, setting metadata is a separate step after uploading.
```
sushantkhanna, line 212: I am not perfectly sure of the use-case here, but that is not necessarily true.  
You could set customer metadata along with the `cp` command - https://docs.cloud.google.com/sdk/gcloud/reference/storage/cp#--custom-metadata

src/clusterfuzz/_internal/google_cloud_utils/gsutil.py
ViniciustCosta: I think it's always better to have it configurable to address possible future use cases. But, as you can check, it is set as true for all current usages.

src/clusterfuzz/_internal/google_cloud_utils/gsutil.py
ViniciustCosta: Not actually, the GSUTIL_PATH points to the gcloud sdk bin directory (for instance, `GSUTIL_PATH="$INSTALL_DIRECTORY/$GOOGLE_CLOUD_SDK/bin"`). This fallback is needed so we don't have to wait for folks to update their init/setup scripts with the new GCLOUD_PATH name, but it will still call gcloud storage.

PR #5125, opened 2026-01-14
Add 'yapf' style parameter to butler.py yapf calls

mverde:
When using butler.py's format and lint commands it seems the default style config currently used for yapf calls is 'chromium', which has been renamed to 'yapf': https://github.com/google/yapf/blob/12005095296072751e3e4c1f33a047d41b0ce18d/CHANGELOG.md?plain=1#L168. This was giving me errors related to the 'chromium' style not being found.

### DIFF
--- a/src/local/butler/format.py
+++ b/src/local/butler/format.py
@@ -17,6 +17,7 @@ import os
 
 from local.butler import common
 
+
 FIRST_PARTY_MODULES = [
     'handlers',
     'libs',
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
The CI failure makes me think you ran the formatting command while not in our virtualenv/using a different version of yapf. After installing our dependencies and doing `python3 -m pipenv shell` do you still get the error?
I think the change is from 3 years ago which is probably more recent than the last time we have upgraded yapf.

PR #5126, opened 2026-01-15
Fix empty testcase upload.

jonathanmetzman:
If you look at the fuzzer logs linked to from stats. There is a testcase uploaded after every fuzzer run. This is a mistake.

### DIFF
--- a/src/clusterfuzz/_internal/bot/testcase_manager.py
+++ b/src/clusterfuzz/_internal/bot/testcase_manager.py
@@ -437,6 +437,9 @@ def upload_testcase(testcase_path, testcase_data, log_time):
     with open(testcase_path, 'rb') as file_handle:
       testcase_data = file_handle.read()
 
+  if not testcase_data:
+    return
+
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
It's not totally correct to assume the testcase can't be empty but nothing depends on this, it's for humans. So I think it's OK not to do the perfectly correct behavior.

ViniciustCosta:
> It's not totally correct to assume the testcase can't be empty but nothing depends on this, it's for humans. So I think it's OK not to do the perfectly correct behavior.

I mean, this seems correct, why would we want to upload an empty file? Not sure what would be the perfectly correct behavior

jonathanmetzman:
> > It's not totally correct to assume the testcase can't be empty but nothing depends on this, it's for humans. So I think it's OK not to do the perfectly correct behavior.
> 
> I mean, this seems correct, why would we want to upload an empty file? Not sure what would be the perfectly correct behavior

There can be empty crashes. But my thinking is these are easy to recover without the testcase ;-)

PR #5133, opened 2026-01-20
Fix unknown ubsan crash types

ViniciustCosta:
This should fix the following unknown ubsan crash types (https://screenshot.googleplex.com/C4sGA6UKssV4Gms.png):
* `assumption is violated during execution`
* `upcast of null pointer of type <type>`

These two are actually related to golang specific crashes:
* `slice bounds out of range [:-1]`
*  `index out of range [-1]`
It should be fixed by checking the golanc specific crash types before the generic UBSan, since if it is a golang crash, it will set the `crash_type` and it will not enter the generic UBSan block.

Added unit test for the new ubsan types.

b/473559705

### DIFF
--- a/src/clusterfuzz/stacktraces/__init__.py
+++ b/src/clusterfuzz/stacktraces/__init__.py
@@ -615,6 +615,17 @@ class StackParser:
             new_type='Bad-cast',
             new_frame_count=0)
 
+      # Golang stacktraces. Needs to be done before the other UBSan crash as
+      # it uses the same pattern of "runtime error:".
+      if state.is_golang:
+        for golang_crash_regex, golang_crash_type in GOLANG_CRASH_TYPES_MAP:
+          if self.update_state_on_match(
+              golang_crash_regex, line, state, new_type=golang_crash_type):
+            state.found_golang_crash = True
+            state.crash_state = ''
+            state.frame_count = 0
+            continue
+
       # Other UndefinedBehavior Sanitizer crash.
       ubsan_runtime_match = UBSAN_RUNTIME_ERROR_REGEX.match(line)
       if ubsan_runtime_match and not state.crash_type and self.include_ubsan:
@@ -641,16 +652,6 @@ class StackParser:
           new_type='Memcpy-param-overlap',
           reset=True,
           address_from_group=2)
-
-      # Golang stacktraces.
-      if state.is_golang:
-        for golang_crash_regex, golang_crash_type in GOLANG_CRASH_TYPES_MAP:
-          if self.update_state_on_match(
-              golang_crash_regex, line, state, new_type=golang_crash_type):
-            state.found_golang_crash = True
-            state.crash_state = ''
-            state.frame_count = 0
-            continue
 
       # Python stacktraces.
       if state.is_python:


jonathanmetzman:
Can you do a query and check if there are others?

ViniciustCosta:
> Can you do a query and check if there are others?

Sure, here it is (probably not the best way to do it, but it confirms there aren't others): https://cloudlogging.app.goo.gl/Ff7E3wkP3VnrGdt17

ViniciustCosta:
Actually, just found another one: `Unknown UBSan crash type: upcast of null pointer of type` (https://cloudlogging.app.goo.gl/xaZ8UXKAGnV62ofP8)

ViniciustCosta:
Also, I think the `out of range` crashes are not actually UBSan, but rather an expected golang specific crash type that is not being correctly treated.

AFAICT all projects reporting these as unknown crash types are go project (jobs: `libfuzzer_asan_mtail`, `libfuzzer_asan_gitea`, `libfuzzer_asan_openkruise`) and even though the constants for these crash types exist (https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/stacktraces/constants.py#L338), they are reaching the UBSan codepath (probably due to `runtime error:` being the ubsan expected regex).

ViniciustCosta:
PTAL. I added the real two new types as non security issues, please let me know if you think otherwise.

src/clusterfuzz/stacktraces/__init__.py
```diff
@@ -615,6 +615,17 @@ def parse(self, stacktrace: str) -> CrashInfo:
...
+      # it uses the same pattern of "runtime error:".
+      if state.is_golang:
+        for golang_crash_regex, golang_crash_type in GOLANG_CRASH_TYPES_MAP:
+          if self.update_state_on_match(
+              golang_crash_regex, line, state, new_type=golang_crash_type):
+            state.found_golang_crash = True
+            state.crash_state = ''
+            state.frame_count = 0
+            continue
+
```
jonathanmetzman, line 628: I hope this doesn't break anything.

jonathanmetzman:
Thanks!

src/clusterfuzz/stacktraces/__init__.py
```diff
@@ -615,6 +615,17 @@ def parse(self, stacktrace: str) -> CrashInfo:
             new_type='Bad-cast',
             new_frame_count=0)
 
+      # Golang stacktraces. Needs to be done before the other UBSan crash as
```
jonathanmetzman, line 618: Actually. I'm not sure what's the right thing to do here.

jonathanmetzman (approved):

src/clusterfuzz/stacktraces/__init__.py
ViniciustCosta: As we discussed offline, other errors shouldn't be treated as golang errors since there is this `is_golang` check before entering the code, that basically checks if ".go:" exists anywhere in the stack trace (which I'd assume works correctly).

Also, that "other ubsan crash" code block is only entered if the crash_type is `None` (`if ubsan_runtime_match and not state.crash_type and self.include_ubsan`). So, AFAICT the only difference of moving up is that, for these golang cases, it will set the crash_type before the ubsan check, avoiding to log the error.

PR #5136, opened 2026-01-22
Add recursive flag to rsync in gcloud storage

ViniciustCosta:
Based on [gcloud documentation](https://docs.cloud.google.com/sdk/gcloud/reference/storage/rsync), rsync is not recursive by default for `gcloud storage`. This hotfix is intended to add back that flag as it was used for `gsutil`.


ViniciustCosta:
Forgot to fix the tests, my bad. Will do

PR #5137, opened 2026-01-22
Revert "[ossfuzz][project_setup] Improve weighting of memory safe vs unsafe projects"

vitaliset:
Reverts google/clusterfuzz#5135


vitaliset:
Wrong revert PR. :)

PR #5140, opened 2026-01-23
Use private batch API to avoid infinite queing

jonathanmetzman:


src/clusterfuzz/_internal/batch/service.py
```diff
@@ -1,8 +1,8 @@
 # Copyright 2025 Google LLC
 #
-# Licensed under the Apache License, Version 2.0
```
jonathanmetzman, line 3: This was done by gemini but I think I'm reverting it to the correct form.

src/clusterfuzz/_internal/batch/service.py
```diff
@@ -65,6 +67,13 @@
 # See https://cloud.google.com/batch/quotas#job_limits
 MAX_CONCURRENT_VMS_PER_JOB = 1000
 
+MAX_QUEUE_SIZE = 50
```
ViniciustCosta: Maybe this could be a config per project?

src/clusterfuzz/_internal/batch/service.py
```diff
@@ -242,15 +261,46 @@ def _get_config_names(batch_tasks: List[remote_task_types.RemoteTask]):
...
+  project = batch_config.get('project')
+
+  for subconfig in instance_subconfigs:
+    name = subconfig['name']
+    conf = all_subconfigs[name]
+    region = conf['region']
+
+    if region in queue_check_regions:
+      load = get_region_load(project, region)
+      logs.info(f'Region {region} has {load} queued/scheduled jobs.')
```
ViniciustCosta: nit: maybe return like -1 when the `get_region_load()` fails and check it before logging to avoid confusion by saying that the region has 0 queued.

PR #5145, opened 2026-01-27
Add handling for reformatted archives conditional on archive format version

letitz:
Opening on behalf of @mverde to start reviewing early.

### DIFF
--- a/src/clusterfuzz/_internal/build_management/build_archive.py
+++ b/src/clusterfuzz/_internal/build_management/build_archive.py
@@ -313,6 +313,7 @@ class ChromeBuildArchive(DefaultBuildArchive):
     return self._root_dir
 
   def archive_schema_version(self) -> int:
+    """Returns the schema version number for this archive."""
     return self._archive_schema_version
 
   def get_dependency_path(self, path: str, deps_file_path: str) -> str:


src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -14,16 +14,14 @@
...
 
 import abc
+import json
 import os
-from typing import BinaryIO
-from typing import Callable
-from typing import List
-from typing import Optional
-from typing import Union
+from typing import BinaryIO, Callable, List, Optional, Union
```
letitz: I think you might run afoul of the linter with this change.

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -224,12 +222,24 @@ class ChromeBuildArchive(DefaultBuildArchive):
   In case something goes wrong, this defaults to using the default unpacker.
   """
 
+  def __init__(self, reader: archive.ArchiveReader):
+    super().__init__(reader)
+    manifest_path = os.path.join(self.root_dir(), 'clusterfuzz_manifest.json')
+    """The manifest may not exist for earlier versions of archives. In this
+    case, default to schema version 0.
+    """
```
letitz: A docstring here surprises me. This should be a comment.

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -224,12 +222,24 @@ class ChromeBuildArchive(DefaultBuildArchive):
   In case something goes wrong, this defaults to using the default unpacker.
   """
 
+  def __init__(self, reader: archive.ArchiveReader):
+    super().__init__(reader)
+    manifest_path = os.path.join(self.root_dir(), 'clusterfuzz_manifest.json')
```
letitz: I don't think we need to handle archives where `clusterfuzz_manifest.json` is within a nested directory. We can declare that for simplicity, manifests must appear at the root of the zip.

```suggestion
    manifest_path = 'clusterfuzz_manifest.json'
```

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -224,12 +222,24 @@ class ChromeBuildArchive(DefaultBuildArchive):
...
+    """The manifest may not exist for earlier versions of archives. In this
+    case, default to schema version 0.
+    """
+    if self.file_exists(manifest_path):
+      archive_schema_json = json.loads(self.open(manifest_path).read().decode())
+      self._archive_schema_version = archive_schema_json.get('version', 0)
```
letitz: We should log a warning when a manifest is present but the version field is missing - that's an unexpected case.

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -224,12 +222,24 @@ class ChromeBuildArchive(DefaultBuildArchive):
...
   """
 
+  def __init__(self, reader: archive.ArchiveReader):
+    super().__init__(reader)
+    manifest_path = os.path.join(self.root_dir(), 'clusterfuzz_manifest.json')
+    """The manifest may not exist for earlier versions of archives. In this
+    case, default to schema version 0.
+    """
+    if self.file_exists(manifest_path):
+      archive_schema_json = json.loads(self.open(manifest_path).read().decode())
```
letitz: Nit: I would just call this `manifest`.

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -224,12 +222,24 @@ class ChromeBuildArchive(DefaultBuildArchive):
   In case something goes wrong, this defaults to using the default unpacker.
   """
 
+  def __init__(self, reader: archive.ArchiveReader):
```
letitz: (I would leave a comment above but GitHub doesn't support that)

We should amend the class docstring to mention the new handling of manifests.

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -224,12 +222,24 @@ class ChromeBuildArchive(DefaultBuildArchive):
...
+    else:
+      self._archive_schema_version = 0
+
   def root_dir(self) -> str:
     if not hasattr(self, '_root_dir'):
       self._root_dir = super().root_dir()  # pylint: disable=attribute-defined-outside-init
     return self._root_dir
 
-  def to_archive_path(self, path: str) -> str:
+  def to_archive_path(self, path: str, archive_schema_version: int = 0) -> str:
```
letitz: Let's use `self._archive_schema_version` here instead of passing an argument.

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -242,8 +252,18 @@ def to_archive_path(self, path: str) -> str:
...
-    if path.startswith('../../'):
+    if archive_schema_version == 0 and path.startswith('../../'):
       path = path.replace('../../', 'src_root/')
+    elif archive_schema_version > 0:
+      if path.startswith('../../'):
+        """For newer archive versions, runtime_deps that were formerly stored
+        under {self.root_dir()}/src_root/ are now stored in the root directory,
+        while the build artifacts formerly stored in the root directory are now
+        stored under {self.root_dir()}/out/msan/.
+        """
```
letitz: Same comment re: docstrings vs comments.

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -242,8 +252,18 @@ def to_archive_path(self, path: str) -> str:
...
 
-    if path.startswith('../../'):
+    if archive_schema_version == 0 and path.startswith('../../'):
       path = path.replace('../../', 'src_root/')
+    elif archive_schema_version > 0:
+      if path.startswith('../../'):
+        """For newer archive versions, runtime_deps that were formerly stored
+        under {self.root_dir()}/src_root/ are now stored in the root directory,
+        while the build artifacts formerly stored in the root directory are now
+        stored under {self.root_dir()}/out/msan/.
```
letitz: The exact path under `out` depends on the build.

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -242,8 +252,18 @@ def to_archive_path(self, path: str) -> str:
...
+        """
+        path = path.replace('../../', '')
+      elif path.startswith('./'):
+        path = path.replace('./', 'out/msan/')
```
letitz: I think we should instead rework this logic to interpret the deps path *relative to a `.runtime_deps` file*, which is really its intended meaning.

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -271,7 +292,8 @@ def _get_common_files(self) -> List[str]:
 
   def get_target_dependencies(
       self, fuzz_target: str) -> List[archive.ArchiveMemberInfo]:
-    target_path = self.to_archive_path(fuzz_target)
+    target_path = self.to_archive_path(fuzz_target,
```
letitz: We're using `to_archive_path` for both this fuzz target name and its dependencies' paths (from `.runtime_deps`) below, but the behavior we want is really quite different in each case:

* for fuzz target names (these are not always paths), we really want to use `get_path_for_target()` (which is as yet conspicuously unused)
* for dependency paths, we want to do the `../..` conditional remapping, relative to `target_path` / `deps_file`

All that to say, I think we want to do:

```py
target_path = self.get_path_for_target(fuzz_target)
```

And rename `to_archive_path()` to `get_dependency_path()`.

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -271,7 +292,8 @@ def _get_common_files(self) -> List[str]:
...
+                                       self._archive_schema_version)
```
letitz: I think there is a bug with the current code, if I'm reasoning through it correctly: the fuzz target name passed as an argument to this method comes from `list_fuzz_targets()`. That method [1] uses `normalize_target_name()` [2], which keeps only the base name of the fuzz target path to derive a fuzz target name: `foo/bar/baz_fuzzer.exe` -> `baz_fuzzer`.

This is compensated for by `to_archive_path()` which prepends `root_dir()` to the fuzz target name again to derive the full path. It only works because all fuzz targets are assumed to belong to the root dir. With our new archive structure, the root dir will be `.` yet the fuzz targets will live under `out/$BUILD_DIR`. So when we try to derive the path to the fuzzer here, we will fail to find a fuzzer executable and its runtime deps file.

We can fix this by using `get_path_for_target()` instead, which will do the right thing and recall the original fuzz target path.

Now, this brings me to an additional improvement I would love to see land, and which is unlocked by manifests. If we directly list the fuzz targets in the manifest:

```js
{
  archive_schema_version: 1,
  fuzz_targets: [
    'out/msan/foo_fuzzer',
    'out/msan/bar_fuzzer',
  ]
}
```

Then we can avoid guessing which files in the archive are fuzz targets using heuristics (see `list_fuzz_targets()` again - it looks at filenames and even sometimes inspects the symbols defined in the binaries - this has long irked me). Better yet, it would fix crbug.com/445765265.

The manifest provides the perfect place to do this. At construction time we can read the list from JSON and populate `self._fuzz_targets` directly, or extract some common functionality:

```py
def set_fuzz_target_paths(paths: Sequence[str]):
  self._fuzz_targets = {
    fuzzer_utils.normalize_target_name(path): path
    for path in paths
  }
```

Then `list_fuzz_targets()`'s lazy initialization logic will work in our favor, and we will never fall back to heuristics.

It would likely make sense to land this as a separate CL - I initially thought we would have to do this to fix this CL, but we don't.

[1]: https://github.com/google/clusterfuzz/blob/e139717456403d5a332b9fe2058fd519714baafc/src/clusterfuzz/_internal/build_management/build_archive.py#L194
[2]: https://github.com/google/clusterfuzz/blob/e139717456403d5a332b9fe2058fd519714baafc/src/clusterfuzz/_internal/bot/fuzzers/utils.py#L194

src/clusterfuzz/_internal/build_management/build_archive.py
mverde: Let's chalk it up to me not being so used to writing Python? :)

src/clusterfuzz/_internal/build_management/build_archive.py
mverde: Done.

src/clusterfuzz/_internal/build_management/build_archive.py
mverde: Done.

src/clusterfuzz/_internal/build_management/build_archive.py
mverde: Done.

src/clusterfuzz/_internal/build_management/build_archive.py
mverde: Done.

src/clusterfuzz/_internal/build_management/build_archive.py
mverde: Done.

src/clusterfuzz/_internal/build_management/build_archive.py
mverde: How does this look?

src/clusterfuzz/_internal/build_management/build_archive.py
mverde: Done.

src/clusterfuzz/_internal/build_management/build_archive.py
mverde: Ack. This sounds good as a separate CL.

src/clusterfuzz/_internal/build_management/build_archive.py
mverde: Done.

src/clusterfuzz/_internal/build_management/build_archive.py
mverde: Done.

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -217,53 +217,78 @@ def unpack(self,
...
-    """The manifest may not exist for earlier versions of archives. In this
-    case, default to schema version 0.
-    """
-    if self.file_exists(manifest_path):
-      archive_schema_json = json.loads(self.open(manifest_path).read().decode())
-      self._archive_schema_version = archive_schema_json.get('version', 0)
-    else:
-      self._archive_schema_version = 0
+  def archive_schema_version(self) -> int:
+    if not hasattr(self, '_archive_schema_version'):
```
letitz: I'd rather see this field initialized at construction time than lazily like this. It's easier to reason about the former and avoids hasattr shenanigans.

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -217,53 +217,78 @@ def unpack(self,
...
-        path = path.replace('../../', '')
-      elif path.startswith('./'):
-        path = path.replace('./', 'out/msan/')
+    elif self.archive_schema_version() > 0:
+      # Assumes the dependency path is relative to the deps file and
+      # transforms it into into a full path relative to the archive root. For
+      # example:
+      #
+      # deps_file_path: "/root/A/B/fuzz_target.runtime_deps"
+      # os.path.dirname(deps_file_path) => "/A/B/" (call this DEPS_DIR)
```
letitz, line 344: `dirname` here would likely be `/root/A/B`?

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -217,53 +217,78 @@ def unpack(self,
...
-      if path.startswith('../../'):
-        """For newer archive versions, runtime_deps that were formerly stored
-        under {self.root_dir()}/src_root/ are now stored in the root directory,
-        while the build artifacts formerly stored in the root directory are now
-        stored under {self.root_dir()}/out/msan/.
-        """
-        path = path.replace('../../', '')
-      elif path.startswith('./'):
-        path = path.replace('./', 'out/msan/')
+    elif self.archive_schema_version() > 0:
```
letitz: I think we can branch earlier and avoid the initial `normpath` call on `path` in this case:

```py
if self.archive_schema_version() > 0:
  # Just interpret `path` as relative to `deps_file_path`. 
  deps_dir = os.path.dirname(deps_file_path)
  return os.path.normpath(os.path.join(deps_dir), path)

# Legacy behavior. Remap `../../` to `src_root/`.
...
```

src/clusterfuzz/_internal/tests/core/build_management/build_archive_test.py
```diff
@@ -133,9 +135,11 @@ def setUp(self):
         'clusterfuzz._internal.system.archive.ArchiveReader',
         'clusterfuzz._internal.system.archive.open',
         'clusterfuzz._internal.bot.fuzzers.utils.is_fuzz_target',
+        'clusterfuzz._internal.build_management.build_archive.ChromeBuildArchive.archive_schema_version',
     ])
     self.mock.open.return_value.list_members.return_value = []
     self.mock.is_fuzz_target.side_effect = self._mock_is_fuzz_target
+    self.mock.archive_schema_version.return_value = 0
     self.build = build_archive.ChromeBuildArchive(self.mock.open.return_value)
```
letitz, line 140: Let's allow passing an archive schema version as a constructor argument to `ChromeBuildArchive` instead of mocking the method. I think it's easier to reason about.

src/clusterfuzz/_internal/tests/core/build_management/build_archive_test.py
```diff
@@ -201,24 +211,7 @@ def test_possible_dependencies(self, dir_prefix):
     self._declare_fuzzers(['my_fuzzer'])
     to_extract = self.build.get_target_dependencies('my_fuzzer')
     to_extract = [f.name for f in to_extract]
-    self.assertCountEqual(to_extract, needed_files)
-
-  @parameterized.parameterized.expand(['/b/build/', 'build/', ''])
```
letitz, line 206: Why delete this test?

src/clusterfuzz/_internal/tests/core/build_management/build_archive_test.py
```diff
@@ -189,10 +193,16 @@ def _mock_open(_):
   def _declare_fuzzers(self, fuzzers):
     self._declared_fuzzers = fuzzers
 
+  def _set_archive_schema_version(self, version):
```
letitz, line 237: I assume you are planning this, but adding a comment so as not to lose track of it: we need to add some tests for the new behavior.

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -219,31 +217,78 @@ def unpack(self,
...
+  Defaults to using the default unpacker in case something goes wrong.
   """
 
+  def archive_schema_version(self) -> int:
+    if not hasattr(self, '_archive_schema_version'):
+      manifest_path = 'clusterfuzz_manifest.json'
+      # The manifest may not exist for earlier versions of archives. In this
+      # case, default to schema version 0.
+      if self.file_exists(manifest_path):
+        manifest = json.loads(self.open(manifest_path).read().decode())
```
letitz: You can just use `json.load()` with a file argument, and we should probably use `with` to ensure the file is closed if necessary:

```py
with self.open(manifest_path) as f:
  manifest = json.load(f)
```

src/clusterfuzz/_internal/build_management/build_archive.py
letitz: Much better! One more comment elsewhere.

src/clusterfuzz/_internal/tests/core/build_management/build_archive_test.py
mverde: Ah sorry, this is actually something I meant to ask about but forgot over the weekend... this test seems exactly the same as the test that proceeds it other than in name. Do you happen to know if the test case is still relevant?

src/clusterfuzz/_internal/build_management/build_archive.py
mverde: This happens when I run butler.py format?

src/clusterfuzz/_internal/build_management/build_archive.py
mverde: Done.

src/clusterfuzz/_internal/build_management/build_archive.py
mverde: My bad, /root was left over from a previous version of the comment; meant to remove it to make the dir names more concise.

src/clusterfuzz/_internal/build_management/build_archive.py
mverde: Done.

src/clusterfuzz/_internal/tests/core/build_management/build_archive_test.py
mverde: Done.

src/clusterfuzz/_internal/build_management/build_archive.py
mverde: Done.

src/clusterfuzz/_internal/build_management/build_archive.py
letitz: Interesting. Are you running from within `pipenv shell`, as @jonathanmetzman asked in #5125?

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -227,21 +227,23 @@ class ChromeBuildArchive(DefaultBuildArchive):
...
+    # The manifest may not exist for earlier versions of archives. In this
+    # case, default to schema version 0.
+    manifest_path = 'clusterfuzz_manifest.json'
+    if self.file_exists(manifest_path):
+      with self.open(manifest_path) as f:
+        manifest = json.load(f)
+      self._archive_schema_version = manifest.get('archive_schema_version', 0)
+      if self._archive_schema_version == 0:
+        logs.warning(
+            'clusterfuzz_manifest.json was incorrectly formatted or missing an archive_schema_version field'
```
letitz: This will also be logged if the manifest contained:

```js
{
  'archive_schema_version': 0
}
```

How about doing:

```py
self._archive_schema_version = manifest.get('archive_schema_version')
if self._archive_schema_version is None:
  log.warning(...)
  self._archive_schema_version = 0
```

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -227,21 +227,23 @@ class ChromeBuildArchive(DefaultBuildArchive):
...
-        if self._archive_schema_version == 0:
-          logs.warning(
-              'clusterfuzz_manifest.json was incorrectly formatted or missing an archive_schema_version field'
-          )
-      else:
-        self._archive_schema_version = 0
-    return self._archive_schema_version
+  def __init__(self,
+               reader: archive.ArchiveReader,
+               archive_schema_version: int = 0):
```
letitz: Given that this is only used if the archive is missing, let's call it `default_archive_schema_version`.

Also, let's document this method:

```py
def __init__(self, ...):
  """Initializes a `ChromiumBuildArchive` with the given reader.

  Arguments:
    reader: See `DefaultBuildArchive`.
    default_archive_schema_version: Specifies which version of a build archive to
      expect if `clusterfuzz_manifest.json` is missing or badly formatted.
  """
```

src/clusterfuzz/_internal/build_management/build_archive.py
letitz: We should document what the legacy schema is, and what version 1 looks like.

src/clusterfuzz/_internal/tests/core/build_management/build_archive_test.py
```diff
@@ -245,11 +290,50 @@ def test_other_fuzzer_not_extracted(self, dir_prefix):
     to_extract = [f.name for f in to_extract]
     self.assertCountEqual(to_extract, needed_files)
 
+  @parameterized.parameterized.expand(['/b/build/', 'build/', ''])
+  def test_possible_dependencies(self, dir_prefix):
+    """Tests that all the necessary dependencies are correctly extracted from
+    the runtime_deps file.
+
+    Under the current archive schema, dependency paths in `runtime_deps` files
```
letitz: Let's say "Under archive schema version 1", it'll age better.

src/clusterfuzz/_internal/tests/core/build_management/build_archive_test.py
```diff
@@ -179,6 +187,41 @@ def _generate_possible_fuzzer_dependencies(self, dir_prefix, fuzz_target):
...
+        f'{fuzz_target}.options',
+        f'{fuzz_target}.runtime_deps',
+        f'{fuzz_target}.par',
+        f'{fuzz_target}.dSYM/Contents/Resources/DWARF/some_dependency',
+        'shared.dll',
+        'shared.dll.pdb',
+        './llvm-symbolizer',
+        'icudtl.dat',
+        'swiftshader/libGLESv2.so',
+        'instrumented_libraries/msan/lib/libgcrypt.so.11.8.2',
```
letitz: In real `.runtime_deps` files, I would expect the reference to e.g. `libgcrypt.so` to be prefixed with `../../`. Can you look at such a file in a ClusterFuzz build to see what it looks like?

src/clusterfuzz/_internal/build_management/build_archive.py
mverde: Ah I see, yeah it's probably because of that. I am running from a shell and AFAICT followed the instructions for setting up the dependencies; haven't figured out yet what went wrong. In any case, reverted this change.

src/clusterfuzz/_internal/tests/core/build_management/build_archive_test.py
mverde: Here's an example:

./zucchini_patch_fuzzer
zucchini_patch_fuzzer.owners
../../testing/test_env.py
../../testing/xvfb.py
../../testing/scripts/common.py
../../.vpython3
../../build/util/lib/__init__.py
../../build/util/lib/proto/
../../.vpython3
../../build/util/lib/__init__.py
../../build/util/lib/results/
./libbase.so
../../tools/valgrind/asan/
../../third_party/llvm-build/Release+Asserts/bin/llvm-symbolizer
./libbase_allocator_partition_allocator_src_partition_alloc_raw_ptr.so
./libbase_allocator_partition_allocator_src_partition_alloc_allocator_base.so
../../tools/valgrind/asan/
../../third_party/llvm-build/Release+Asserts/bin/llvm-symbolizer
../../tools/memory/sanitizer/escalate_sanitizer_warnings.py
../../third_party/instrumented_libs/binaries/msan-chained-origins-noble-lib/lib
third_party/instrumented_libs/binaries/msan-chained-origins-noble-lib/lib/ld-linux-x86-64.so.2
./libc++.so
./libatomic.so
./libbase_allocator_partition_allocator_src_partition_alloc_allocator_core.so
./libbase_allocator_partition_allocator_src_partition_alloc_allocator_shim.so
./libthird_party_abseil-cpp_absl.so
./libthird_party_boringssl.so
./libthird_party_perfetto_libperfetto.so
./libchrome_zlib.so
./libthird_party_icu_icui18n.so
./libicuuc.so
icudtl.dat
bin/run_zucchini_patch_fuzzer
../../.vpython3
../../testing/location_tags.json

The answer seems to be that some are prefixed with relative separators and some are not.

src/clusterfuzz/_internal/build_management/build_archive.py
mverde: Done.

src/clusterfuzz/_internal/build_management/build_archive.py
mverde: Done. Does this look good to you?

src/clusterfuzz/_internal/tests/core/build_management/build_archive_test.py
mverde: Done.

src/clusterfuzz/_internal/build_management/build_archive.py
mverde: Done.

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -219,29 +221,97 @@ def unpack(self,
...
+    manifest_path = 'clusterfuzz_manifest.json'
+    if self.file_exists(manifest_path):
+      with self.open(manifest_path) as f:
+        manifest = json.load(f)
+      self._archive_schema_version = manifest.get('archive_schema_version')
+      if self._archive_schema_version is None:
+        logs.warning(
+            'clusterfuzz_manifest.json was incorrectly formatted or missing an archive_schema_version field'
+        )
+        self._archive_schema_version = 0
```
letitz: We should use `default_archive_schema_version` here.

src/clusterfuzz/_internal/build_management/build_archive.py
letitz: It's a great start, thanks! I think giving a couple example directory trees would speak volumes, and convey the schema more clearly.

src/clusterfuzz/_internal/tests/core/build_management/build_archive_test.py
letitz: Got it, thanks! Then can you amend this list of fake dependencies to look like a real one?

In particular, I note that there are only 3 cases:

* unprefixed paths:
   * `zucchini_patch_fuzzer.owners`
   *  `icudtl.dat`
   * `third_party/instrumented_libs/...` (we should be able to get rid of this eventually as part of fixing MSAN)
 * paths prefixed with `./`:
   * `./zucchini_patch_fuzzer`
   * `./libbase.so`
* paths prefixed with `../../`:
   * `../../testing/xvfb.py`
   * `../../third_party/instrumented_lib/...`

src/clusterfuzz/_internal/tests/core/build_management/build_archive_test.py
```diff
@@ -204,59 +257,83 @@ def test_possible_dependencies(self, dir_prefix):
...
+    other_fuzzer = self._generate_possible_fuzzer_dependencies_legacy(
+        dir_prefix, 'other_fuzzer')
+    self._add_files_to_archive(list(set(needed_files + other_fuzzer)))
     self._generate_runtime_deps(deps_files)
+    self._declare_fuzzers(['my_fuzzer', 'other_fuzzer'])
+    to_extract = self.build.get_target_dependencies('my_fuzzer')
+    to_extract = [f.name for f in to_extract]
+    self.assertCountEqual(to_extract, needed_files)
+
+  @parameterized.parameterized.expand(['/b/build/', 'build/', ''])
```
letitz, line 292: I don't think we need to parameterize these tests. The fuzzers should live under `out/foo` and we can test for that scenario only.

src/clusterfuzz/_internal/tests/core/build_management/build_archive_test.py
```diff
@@ -204,59 +257,83 @@ def test_possible_dependencies(self, dir_prefix):
...
+    the runtime_deps file.
+
+    Under archive schema version 1, dependency paths in `runtime_deps` files
+    are interpreted as being relative to the file itself, meaning that they must
+    be normalized to the equivalent path relative to the archive root before
+    they can be extracted.
+    """
+    self._set_archive_schema_version(1)
+    deps_entries = self._generate_possible_fuzzer_dependencies('my_fuzzer')
+    deps_files = self._generate_normalized_dependency_filenames(
```
letitz: This function calls `_generate_possible_fuzzer_dependencies()` again even though we just did on the line above. How about passing in a list of relative paths directly, and calling this something like `_resolve_relative_dependency_paths(fuzz_target_path, dependency_paths)`?

src/clusterfuzz/_internal/tests/core/build_management/build_archive_test.py
```diff
@@ -204,59 +257,83 @@ def test_possible_dependencies(self, dir_prefix):
...
     self._add_files_to_archive(list(set(needed_files + other_fuzzer)))
-    self._generate_runtime_deps(deps_files)
+    self._generate_runtime_deps(deps_entries)
     self._declare_fuzzers(['my_fuzzer', 'other_fuzzer'])
     to_extract = self.build.get_target_dependencies('my_fuzzer')
     to_extract = [f.name for f in to_extract]
     self.assertCountEqual(to_extract, needed_files)
 
   @parameterized.parameterized.expand(['/b/build/', 'build/', ''])
   def test_dsyms_are_correctly_unpacked(self, dir_prefix):
```
letitz: We should leave the test for the legacy archive format in place, and duplicate it for schema version 1.

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -219,29 +221,97 @@ def unpack(self,
...
+    Arguments:
+      reader: See `DefaultBuildArchive`.
+      default_archive_schema_version: Specifies which version of a build archive
+        to expect if `clusterfuzz_manifest.json` is missing or badly formatted.
+    """
+    super().__init__(reader)
+    # The manifest may not exist for earlier versions of archives. In this
+    # case, default to schema version 0.
+    manifest_path = 'clusterfuzz_manifest.json'
+    if self.file_exists(manifest_path):
```
letitz, line 298: This branch is untested. How hard would it be to test it?

letitz (commented):
Nearly there!

src/clusterfuzz/_internal/build_management/build_archive.py
mverde: Done.

src/clusterfuzz/_internal/tests/core/build_management/build_archive_test.py
mverde: Done.

src/clusterfuzz/_internal/build_management/build_archive.py
mverde: Done.

src/clusterfuzz/_internal/tests/core/build_management/build_archive_test.py
mverde: Done.

src/clusterfuzz/_internal/tests/core/build_management/build_archive_test.py
mverde: I did something similar... how does this look?

src/clusterfuzz/_internal/tests/core/build_management/build_archive_test.py
mverde: Done.

src/clusterfuzz/_internal/build_management/build_archive.py
mverde: It was hard to avoid issues with mocks stepping over each other until I added a new test class.

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -219,29 +221,145 @@ def unpack(self,
...
+  src_root/
+    .vpython3
+    third_party/instrumented_libs/binaries/msan-chained-origins-noble-lib/lib/
+      # all instrumented libs
+    # etc. for other deps
+
+  my_fuzzer.runtime_deps:
+  ==========
+  ./my_fuzzer
+  my_fuzzer.options
```
letitz: In the end both .runtime_deps examples are identical - one mentions `libatomic.so` and such, but that presumably is just an omission from the other one. I think we can list the runtime deps a single time.

letitz (commented):
LGTM, thanks! I apparently cannot approve my own PR, so @decoNR for review.

decoNR:
It seems that lint is failing. Please fix it, thanks!

src/clusterfuzz/_internal/build_management/build_archive.py
mverde: Done.

mverde:
> It seems that lint is failing. Please fix it, thanks!

Fixed! @decoNR can you take one more look?

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -219,29 +220,143 @@ def unpack(self,
...
   def root_dir(self) -> str:
     if not hasattr(self, '_root_dir'):
       self._root_dir = super().root_dir()  # pylint: disable=attribute-defined-outside-init
     return self._root_dir
 
-  def to_archive_path(self, path: str) -> str:
-    """Deps are relative to the Chrome root directory. However, there might be
-    a common root directory in the archive, which means we need to make sure
-    the file path is correct.
+  def archive_schema_version(self) -> int:
```
decoNR, line 315: This function does not appear to be used anywhere. If I am mistaken and it should be retained, please add a doc string, thanks!

mverde: Added a doc string; it's used in tests to verify expectations!

PR #5147, opened 2026-01-29
Amending broken link in Heartbleed example.

guy-with-a-why:
A small change to the `Building a libFuzzer target for OpenSSL` section of the Heartbleed Example page where the hyperlink to the OpenSSL tar no longer exists and a wget is required in place of a curl request.


cemon721-a11y:
I think you can remove my all existed system what I made and kindly
resolved all

guy-with-a-why:
Apologies. I don't understand your comment.

> I think you can remove my all existed system what I made and kindly
> resolved all
>

PR #5149, opened 2026-01-29
Add module for using cloud identity api to manage google groups

ViniciustCosta:
This is some groundwork in order to use google groups for CCs in OSS-Fuzz issues.
Context: b/477964128, b/388390041 and https://github.com/google/oss-fuzz/issues/12945

Since we used google groups for the auth/access in appengine, I updated it to centralize in the same module under google cloud utils.

Tests: 
* Used a local script (gpaste/4919004936404992) to execute all methods using a test group. Output: gpaste/6567249169219584

### DIFF
--- a/src/clusterfuzz/_internal/google_cloud_utils/google_groups.py
+++ b/src/clusterfuzz/_internal/google_cloud_utils/google_groups.py
@@ -17,13 +17,15 @@ import threading
 from urllib import parse
 
 from googleapiclient import discovery
-from googleapiclient.errors import HttpError
+from googleapiclient import errors
 
 from clusterfuzz._internal.config import local_config
 from clusterfuzz._internal.google_cloud_utils import credentials
 from clusterfuzz._internal.metrics import logs
 
 # pylint: disable=no-member
+
+_FAIL_RETRIES = 3
 
 _local = threading.local()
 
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/google_cloud_utils/google_groups.py
```diff
@@ -0,0 +1,174 @@
+# Copyright 2026 Google LLC
```
vitaliset, line 1: Are you adding a test for this file?

javanlacerda: +1

ViniciustCosta: Not actually tbh. I did not find any way to "emulate" the cloud identity API, so adding unit tests to this module seems ineffective, as it would need to mock all the api calls and responses without actually verifying if it works.

I think adding tests for the future modules that use this one will be more useful, as they will expect a defined response to do some task.

ViniciustCosta:
> Is there a way to test it in dev ?

Not actually as this will only be used by OSS-Fuzz (and dev simulates chrome environment).

The plan here is to create a separate cronjob (already did in https://github.com/google/clusterfuzz/pull/5151) to create and sync the OSS-Fuzz projects' groups initially. Afterwards, I will adapt the issue tracker to CC the groups instead of the actual emails. Once everything is working fine, then we can integrate the group syncing step into the project-setup cron.

PR #5150, opened 2026-01-29
Implement Kubernetes job limiter

javanlacerda:
This PR enhances the reliability of remote task scheduling by introducing queue size limits and handling uncreated tasks via lease cancellation. It refactors RemoteTaskGate, GcpBatchService, and KubernetesService to return tasks that fail scheduling, ensuring they are redelivered by Pub/Sub.

If the task cannot be scheduled to the runtime, it returns to utask-main scheduler queue. 

Also, the UTask execute function raises an exception if the utask-main schedule queue size is higher than a threshold. This exception makes the preprocess task to not be acked.

Finally, this follow up PR https://github.com/google/clusterfuzz/pull/5153 addresses the schedule-fuzz implementation to avoid scheduling more tasks if the preprocess queue size is also bigger then a threshold. 

This entire logic described was tested in the dev environment. Follow the evidences:

Then the Kubernetes job limiter was reached, the utask-main queue started to grow due to the not acked tasks. It stopped close to 10k messages, because the UTask execute function only runs the preprocess if the utask-main queue is under 10k. 
image

As the tworkers stopped run preprocess, the preprocess queue started to grow. Note that the preprocess queue size stick close to 10K as well. It happens because the schedule-fuzz only creates the tasks based on preprocess queue size, aiming to have 10k messages there. It is being addressed in this PR https://github.com/google/clusterfuzz/pull/5153.
image

### DIFF
--- a/src/clusterfuzz/_internal/base/errors.py
+++ b/src/clusterfuzz/_internal/base/errors.py
@@ -115,5 +115,5 @@ class InvalidConfigKey(Error):
 class QueueLimitReachedError(Error):
   """Error thrown when the queue limit is reached."""
 
-  def __init__(self, size):
-    super().__init__(f'Queue limit reached {size}.')
+  def __init__(self, size, queue):
+    super().__init__(f'Queue {queue} has reached the limit of {size}.')
--- a/src/clusterfuzz/_internal/base/tasks/__init__.py
+++ b/src/clusterfuzz/_internal/base/tasks/__init__.py
@@ -109,6 +109,12 @@ UTASK_QUEUE_PULL_SECONDS = 150
 # scheduling on batch.
 MAX_UTASKS = 3000
 
+# Time window to get the metrics.
+# We should look for metrics in
+# start = now - _QUEUE_LIMIT_INTERVAL
+# end = now
+_QUEUE_LIMIT_INTERVAL = 5 * 60  # 5 minutes.
+
 
 class Error(Exception):
   """Base exception class."""
@@ -707,7 +713,7 @@ def get_utask_main_queue_size():
     now = time.time()
     interval = monitoring_v3.TimeInterval(
         end_time={'seconds': int(now)},
-        start_time={'seconds': int(now - 5 * 60)},
+        start_time={'seconds': int(now - _QUEUE_LIMIT_INTERVAL)},
     )
 
     results = client.list_time_series(
--- a/src/clusterfuzz/_internal/bot/tasks/task_types.py
+++ b/src/clusterfuzz/_internal/bot/tasks/task_types.py
@@ -20,6 +20,7 @@ from clusterfuzz._internal.base import errors
 from clusterfuzz._internal.base import feature_flags
 from clusterfuzz._internal.base import tasks
 from clusterfuzz._internal.base.tasks import task_utils
+from clusterfuzz._internal.base.tasks import UTASK_MAIN_QUEUE
 from clusterfuzz._internal.batch import service as batch_service
 from clusterfuzz._internal.bot.tasks import utasks
 from clusterfuzz._internal.metrics import events
@@ -165,7 +166,10 @@ class UTask(BaseUTask):
     if utask_flag and utask_flag.enabled:
       utask_main_queue_limit = utask_flag.content
     if utask_main_queue_size > utask_main_queue_limit:
-      raise errors.QueueLimitReachedError(utask_main_queue_size)
+      base_os_version = environment.get_value('BASE_OS_VERSION')
+      queue_name = UTASK_MAIN_QUEUE if not base_os_version else \
+        f'{UTASK_MAIN_QUEUE}-{base_os_version}'
+      raise errors.QueueLimitReachedError(utask_main_queue_size, queue_name)
 
     logs.info('Preprocessing utask.')
     download_url = self.preprocess(task_argument, job_type, uworker_env)
--- a/src/clusterfuzz/_internal/base/feature_flags.py
+++ b/src/clusterfuzz/_internal/base/feature_flags.py
@@ -31,6 +31,8 @@ class FeatureFlags(Enum):
 
   UTASK_MAIN_QUEUE_LIMIT = 'utask_main_queue_limit'
 
+  GCP_BATCH_JOBS_FREQUENCY = 'gcp_batch_jobs_frequency'
+
   @property
   def flag(self):
     """Get the feature flag."""
--- a/src/clusterfuzz/_internal/remote_task/remote_task_adapters.py
+++ b/src/clusterfuzz/_internal/remote_task/remote_task_adapters.py
@@ -35,7 +35,8 @@ class RemoteTaskAdapters(Enum):
   """
   KUBERNETES = ('kubernetes', k8s_service.KubernetesService,
                 feature_flags.FeatureFlags.K8S_JOBS_FREQUENCY, 0.0)
-  GCP_BATCH = ('gcp_batch', batch_service.GcpBatchService, None, 1.0)
+  GCP_BATCH = ('gcp_batch', batch_service.GcpBatchService,
+               feature_flags.FeatureFlags.GCP_BATCH_JOBS_FREQUENCY, 1.0)
 
   def __init__(self, adapter_id, service, feature_flag, default_weight):
     self.id = adapter_id


src/clusterfuzz/_internal/remote_task/remote_task_gate.py
```diff
@@ -105,13 +107,43 @@ def get_job_frequency(self):
     logs.info('Job frequencies', frequencies=frequencies)
     return frequencies
 
+  def prepare_unscheduled_tasks(
```
ViniciustCosta: Not all tasks are fuzzer-based, i.e., the task argument might not be the fuzzer. For minimize, regression, etc., the task argument is actually the testcase ID. We might need to add a logic to retrieve that information from the signed URL. I think it should be as simple as:
```py
uworker_input = uworker_io.download_and_deserialize_uworker_input(
          input_download_url)
if fuzzer_based_task: task_argument = uworker_input.fuzzer_name
if testcase_based_task: task_argument = uworker_input.testcase_id
```

Check https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/base/tasks/task_utils.py#L67

src/clusterfuzz/_internal/remote_task/remote_task_types.py
```diff
@@ -42,10 +46,15 @@ class RemoteTaskInterface(abc.ABC):
...
+    """
     raise NotImplementedError
 
   @abc.abstractmethod
-  def create_utask_main_jobs(self, remote_tasks: list[RemoteTask]):
-    """Creates many remote tasks for uworker main tasks."""
+  def create_utask_main_jobs(
+      self, remote_tasks: typing.List[RemoteTask]) -> typing.List[RemoteTask]:
+    """Creates many remote tasks for uworker main tasks.
+       Returns the tasks that couldn't be created.
```
decoNR, line 57: This is not True yet for GCP Batch, right?

src/clusterfuzz/_internal/remote_task/remote_task_gate.py
javanlacerda: Nice catch on that! I updated it to use the `fuzzer_name` or the `testcase_id` based on its testcase based or not.

src/clusterfuzz/_internal/remote_task/remote_task_types.py
javanlacerda: Its kind of true, because for batch we don't have job limiter yet, then all tasks are considered created and then the Batch service returns a empty list. The next step is implementing the job limiter for batch and then use this pattern. 
[This PR](https://github.com/google/clusterfuzz/pull/5140) drafts the job limiter implementation for batch

javanlacerda:
> Just a clarification: is there a chance the `preprocess queue` could become flooded?

Great question. There is a chance actually, but I'm addressing to handle this in [this PR](https://github.com/google/clusterfuzz/pull/5153). Basically the schedule fuzz will push messages there only if the queue is not flooded.

src/clusterfuzz/_internal/remote_task/remote_task_types.py
decoNR: Nice, thank you.

decoNR:
> Great question. There is a chance actually, but I'm addressing to handle this in https://github.com/google/clusterfuzz/pull/5153. Basically the schedule fuzz will push messages there only if the queue is not flooded.

This approach seems nice! Thanks. Maybe we should deploy them together.

src/clusterfuzz/_internal/batch/service.py
```diff
@@ -396,12 +396,10 @@ def create_utask_main_jobs(self,
...
-
     logs.info('Batching utask_mains.')
     for spec, input_urls in job_specs.items():
       for input_urls_portion in utils.batched(input_urls,
                                               MAX_CONCURRENT_VMS_PER_JOB - 1):
-        jobs.append(self.create_job(spec, input_urls_portion).name)
+        self.create_job(spec, input_urls_portion).name
 
-    return jobs
+    return []
```
ViniciustCosta, line 405: It's kind of weird always returning an empty list. IIUC this change is to avoid enabling the rescheduling logic for unacked tasks on batch?

src/clusterfuzz/_internal/k8s/service.py
```diff
@@ -303,17 +303,26 @@ def create_utask_main_jobs(
     separate batch job for each group. This allows tasks with similar
     requirements to be processed together, which can improve efficiency.
     """
+
+    flag = feature_flags.FeatureFlags.K8S_JOBS_PENDING_LIMIT.flag
```
ViniciustCosta: Nit: why not maintain syle by actually using the feature flag class? Like it is done here https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/remote_task/remote_task_gate.py#L80

src/clusterfuzz/_internal/k8s/service.py
```diff
@@ -303,17 +303,26 @@ def create_utask_main_jobs(
...
+    limit = int(
+        flag.value) if flag and flag.enabled else K8S_JOBS_PENDING_LIMIT_DEFAULT
+
+    pending_jobs_count = self._get_pending_jobs_count()
+    if pending_jobs_count >= limit:
+      logs.warning(
+          f'Pending jobs count {pending_jobs_count} reached limit {limit}.')
```
ViniciustCosta: nit: add `for k8s`

src/clusterfuzz/_internal/k8s/service.py
```diff
@@ -303,17 +303,26 @@ def create_utask_main_jobs(
...
     for remote_task in remote_tasks:
-      logs.info(f'Scheduling {remote_task.command}, {remote_task.job_type}.')
+      logs.info(
+          f'Scheduling {remote_task.command}, {remote_task.job_type} in K8s.')
       config = configs[(remote_task.command, remote_task.job_type)]
       job_specs[config].append(remote_task.input_download_url)
-    logs.info('Creating batch jobs.')
-    jobs = []
-    logs.info('Batching utask_mains.')
+    logs.info('Creating GCP batch jobs.')
```
ViniciustCosta: Maybe "Creating utask main jobs in K8S" instead?

src/clusterfuzz/_internal/remote_task/remote_task_gate.py
```diff
@@ -21,8 +21,10 @@
 
 import collections
 import random
-from typing import List
+import typing
```
ViniciustCosta: Nit: Prefer using the built-in types (like `list`) over the parametric type aliases from typing lib.
go/pystyle#typing-imports

src/clusterfuzz/_internal/remote_task/remote_task_gate.py
```diff
@@ -105,13 +107,44 @@ def get_job_frequency(self):
...
     return frequencies
 
+  def prepare_unscheduled_tasks(
+      self, unscheduled_remote_tasks: typing.List[remote_task_types.RemoteTask]
+  ) -> typing.List[remote_task_types.RemoteTask]:
+    """Prepares the unscheduled remote tasks to be sent back to the 
+      preprocess queue.
+
+      The messages in the preprocess queue expects tasks with 
+      the properties followng task_name, fuzzer, job, eta.
```
ViniciustCosta: typo: followng -> following

src/clusterfuzz/_internal/remote_task/remote_task_gate.py
```diff
@@ -105,13 +107,44 @@ def get_job_frequency(self):
...
 
+  def prepare_unscheduled_tasks(
+      self, unscheduled_remote_tasks: typing.List[remote_task_types.RemoteTask]
+  ) -> typing.List[remote_task_types.RemoteTask]:
+    """Prepares the unscheduled remote tasks to be sent back to the 
+      preprocess queue.
+
+      The messages in the preprocess queue expects tasks with 
+      the properties followng task_name, fuzzer, job, eta. 
+      The utasks has the signed url as argument instead of the fuzzer,
```
ViniciustCosta: Please update the docstring to match the actual implementation.

src/clusterfuzz/_internal/remote_task/remote_task_gate.py
```diff
@@ -146,13 +179,16 @@ def create_utask_main_jobs(self,
...
           service = self._service_map[adapter_id]
-          results.extend(service.create_utask_main_jobs(tasks))
+          unscheduled_tasks.extend(service.create_utask_main_jobs(tasks))
         except Exception:  # pylint: disable=broad-except
           logs.error(f'Failed to send {len(tasks)} tasks to {adapter_id}.')
-    return results
+
+    prepared_unscheduled_tasks = self.prepare_unscheduled_tasks(
+        unscheduled_tasks)
+    return prepared_unscheduled_tasks
```
ViniciustCosta: Is there any code that used the jobs information that was returned before this change?

src/python/bot/startup/run_bot.py
```diff
@@ -103,7 +103,17 @@ def schedule_utask_mains():
...
+    # We send back to the preprocess the tasks that couldn't be scheduled to
+    # be executed. This is necessary because the signed url can expire.
+    # We also need to send it back and not discard it because it would break
+    # the coreography avoiding to execute follow up tasks, such as
+    # progression, minimize...
+    uncreated_tasks = remote_task_gate.RemoteTaskGate().create_utask_main_jobs(
+        batch_tasks)
+    logs.info(f'Sending back {len(uncreated_tasks)} uncreated tasks to '
+              f'{tasks.PREPROCESS_QUEUE} queue.')
+    tasks.bulk_add_tasks(
```
ViniciustCosta: I think we need to make sure that the task data that is going back to the preprocess queue is as expected including `extra_info` (i.e., the same as it was in preprocess). Could you try to test this and confirm it in dev?

src/clusterfuzz/_internal/batch/service.py
javanlacerda: Yes. Basically the new contract for the function `create_utask_main_jobs` defines to return the tasks that couldn't be created. As we don't have job limiter for Batch yet, all tasks are created, then we should return an empty list.

src/clusterfuzz/_internal/k8s/service.py
javanlacerda: Getting the feature flag from the adapter tuple inside the service would cause circular import

src/clusterfuzz/_internal/k8s/service.py
javanlacerda: yep. my bad

src/clusterfuzz/_internal/remote_task/remote_task_gate.py
javanlacerda: Resolving as we moved the strategy to just not ack the task.

src/clusterfuzz/_internal/remote_task/remote_task_gate.py
javanlacerda: nice call

src/clusterfuzz/_internal/remote_task/remote_task_gate.py
javanlacerda: removed.

src/clusterfuzz/_internal/remote_task/remote_task_gate.py
javanlacerda: It wasn't used actually. [here](https://github.com/google/clusterfuzz/pull/5113/changes#diff-73d2bdc88e870909471db564257e42c3f2bc7e28879cc00559f5fc0d41758c05L104) I started to use it.

src/python/bot/startup/run_bot.py
javanlacerda: resolving as we're just not acking the message now.

src/clusterfuzz/_internal/batch/service.py
ViniciustCosta: I see, thanks

src/clusterfuzz/_internal/k8s/service.py
ViniciustCosta: No, not getting it from the adapter tuple. I meant that we should standardize how to use feature flags throughout the code. Instead of doing `flag = feature_flag...PENDING_LIMIT.flag`, one should use `flag = feature_flag...PENDING_LIMIT` and then use the properties from the Enum class (e.g., `flag.content`).

Not a blocker though, just nit :)

src/clusterfuzz/_internal/remote_task/remote_task_gate.py
ViniciustCosta: Gotcha, thanks

src/clusterfuzz/_internal/base/errors.py
```diff
@@ -110,3 +110,10 @@ def __init__(self, key_name):
 
   def __str__(self):
     return 'Invalid config key %s.' % self.key_name
+
+
+class QueueLimitReachedError(Error):
+  """Error thrown when the queue limit is reached."""
+
+  def __init__(self, size):
+    super().__init__(f'Queue limit reached {size}.')
```
ViniciustCosta: Nit: Maybe it would be nice to have the queue name here also `(self, queue_name, size) -> f'Queue {queue_name} limit reached {size} '`

src/clusterfuzz/_internal/base/tasks/__init__.py
```diff
@@ -678,6 +688,46 @@ def get_utask_mains() -> List[PubSubTask]:
   return handle_multiple_utask_main_messages(messages, UTASK_MAIN_QUEUE)
 
 
+def get_utask_main_queue_size():
+  """Returns the size of the utask main queue."""
+  queue_name = UTASK_MAIN_QUEUE
```
ViniciustCosta, line 699: Would be nice to have a method to get the queue_name with the optional base_os_version, to centralize this.

src/clusterfuzz/_internal/base/tasks/__init__.py
```diff
@@ -678,6 +688,46 @@ def get_utask_mains() -> List[PubSubTask]:
...
+                  f'resource.labels.subscription_id="{queue_name}"')
+
+  try:
+    creds = credentials.get_default()[0]
+    client = monitoring_v3.MetricServiceClient(credentials=creds)
+
+    now = time.time()
+    interval = monitoring_v3.TimeInterval(
+        end_time={'seconds': int(now)},
+        start_time={'seconds': int(now - 5 * 60)},
```
ViniciustCosta: Maybe better to have a global var defining the interval `_QUEUE_LIMIT_INTERVAL`?

src/clusterfuzz/_internal/k8s/service.py
javanlacerda: I see your point, I'll update it

src/clusterfuzz/_internal/remote_task/remote_task_gate.py
```diff
@@ -139,20 +140,25 @@ def create_utask_main_jobs(self,
             remote_tasks[start_index:start_index + count])
         start_index += count
 
-      # Distribute any remainder tasks (due to rounding) one by one. This
-      # ensures that all tasks are assigned to a backend.
       remaining_tasks = remote_tasks[start_index:]
-      for i, task in enumerate(remaining_tasks):
-        adapter_id = list(frequencies.keys())[i % len(frequencies)]
-        tasks_by_adapter[adapter_id].append(task)
+      if sum(frequencies.values()) >= 0.999:
```
ViniciustCosta, line 144: Not sure I understand this logic, if the frequencies don't add up to 100%, we are simply not scheduling all tasks?

src/clusterfuzz/_internal/bot/tasks/task_types.py
```diff
@@ -154,6 +158,15 @@ def execute(self, task_argument, job_type, uworker_env):
       self.execute_locally(task_argument, job_type, uworker_env)
       return
 
+    utask_main_queue_size = tasks.get_utask_main_queue_size()
```
ViniciustCosta, line 162: Not sure if this is in the correct place, shouldn't this be on utask main? Here the task would be doing preprocess.

ViniciustCosta: Correcting myself here, think I got it, the idea is to not even execute preprocess if the utask main queue is full, right?

src/clusterfuzz/_internal/base/errors.py
javanlacerda: sounds good.

src/clusterfuzz/_internal/bot/tasks/task_types.py
javanlacerda: Exactly!

src/clusterfuzz/_internal/remote_task/remote_task_gate.py
javanlacerda: Yes, that's the idea. Because as the scheduling of the tasks has a fixed flow, we should be able to use the gate not only to set a proportion for scheduling all task, but also creating a bottleneck for any adapter.

For instance, considering the fleet of utask-main-schedulers is able to schedule 1000 tasks per minute, we should be able to set 10% for Batch and 5% for K8s, and then, we know that we'll have 100 tasks per minute being scheduled to Batch and 50 for K8s.

src/clusterfuzz/_internal/base/tasks/__init__.py
javanlacerda: I agree, but this is the pattern that is being used in the code, building the queue name in the same place that uses it. 

Ideally we should create this centralized function and to refactor the code for using it.

PR #5151, opened 2026-01-30
Add cron for creating OSS-Fuzz CC groups

ViniciustCosta:
Cronjob to create and sync the OSS-Fuzz projects' CCs with their respective google groups, which will be used in the issue tracker.

After running this for a couple of times (which will create all the groups for the existing projects and sync them), my plan is to add this step in the project setup and remove this cronjob.

Context: b/477964128, b/388390041, https://github.com/google/oss-fuzz/issues/12945

### DIFF
--- a/src/clusterfuzz/_internal/tests/appengine/handlers/cron/oss_fuzz_cc_groups_test.py
+++ b/src/clusterfuzz/_internal/tests/appengine/handlers/cron/oss_fuzz_cc_groups_test.py
@@ -1,0 +1,106 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


ViniciustCosta:
> It seems great! Could you please deploy it in dev and collect evidences? You would need to create the cronjob entity in the dev config as well.

This is intended only for oss-fuzz, so deploying it to dev won't work

Also, this does not change any behavior from CF yet. It only creates the groups and add members, so the worst that can happen is the cronjob not finishing

ViniciustCosta:
> Approving to unblock, but I think we should be able to test this kind of change before landing. Probably implementing in further PRs an end to end tests for project setup?

Thanks!

In this case, I created the feature as a separate cronjob precisely for testing before adding it as a step in the project-setup, so that if this breaks, it does not interfere with any other code.

I agree that it would be great to have an e2e or integration test for project setup in the future, but that should be another task.

PR #5153, opened 2026-02-01
schedule fuzz based on preprocess queue size

javanlacerda:
SHOULD BE MERGED AFTER https://github.com/google/clusterfuzz/pull/5150.

It updates the logic for scheduling fuzz tasks. Instead of aiming to load the GCP Batch infrastructure based on the available CPUs, it looks only to the preprocess queue size. 

By default, it aims to keep the preprocess queue with 10k messages, creating the number of tasks based on the difference between the aim value and the current value. 

image

### DIFF
--- a/src/clusterfuzz/_internal/cron/schedule_fuzz.py
+++ b/src/clusterfuzz/_internal/cron/schedule_fuzz.py
@@ -14,40 +14,24 @@
 """Cron job to schedule fuzz tasks that run on batch."""
 
 import collections
-import multiprocessing
 import random
 import time
-from typing import Dict
-from typing import List
 
 from google.cloud import monitoring_v3
-from googleapiclient import discovery
 
+from clusterfuzz._internal.base import feature_flags
 from clusterfuzz._internal.base import tasks
 from clusterfuzz._internal.base import utils
-from clusterfuzz._internal.batch import service as batch
-from clusterfuzz._internal.config import local_config
 from clusterfuzz._internal.datastore import data_types
 from clusterfuzz._internal.datastore import ndb_utils
 from clusterfuzz._internal.google_cloud_utils import credentials
 from clusterfuzz._internal.metrics import logs
 
-# TODO(metzman): Actually implement this.
-CPUS_PER_FUZZ_JOB = 2
-
-# Pretend like our CPU limit is 3% higher than it actually is so that we use the
-# full CPU capacity even when scheduling is slow.
-CPU_BUFFER_MULTIPLIER = 1.03
+PREPROCESS_TARGET_SIZE_DEFAULT = 10000
 
 
-def _get_quotas(creds, project, region):
-  compute = discovery.build('compute', 'v1', credentials=creds)
-  return compute.regions().get(  # pylint: disable=no-member
-      region=region, project=project).execute()['quotas']
-
-
-def count_unacked(creds, project_id, subscription_id):
-  """Counts the unacked messages in |subscription_id|."""
+def get_queue_size(creds, project_id, subscription_id):
+  """Returns the size of the queue (unacked messages)."""
   # TODO(metzman): Not all of these are fuzz_tasks. Deal with that.
   metric = 'pubsub.googleapis.com/subscription/num_undelivered_messages'
   query_filter = (f'metric.type="{metric}" AND '
@@ -76,58 +60,14 @@ def count_unacked(creds, project_id, subscription_id):
   return 0
 
 
-def get_cpu_usage(creds, project: str, region: str) -> int:
-  """Returns the number of available CPUs in the current GCE region."""
-
-  quotas = _get_quotas(creds, project, region)
-
-  # Sometimes, the preemptible quota is 0, which means the number of preemptible
-  # CPUs is actually limited by the CPU quota.
-  # If preemptible quota is not defined, we need to use CPU quota instead.
-  cpu_quota = None
-  preemptible_quota = None
-
-  # Get preemptible_quota and cpu_quota from the list of quotas.
-  for quota in quotas:
-    if preemptible_quota and cpu_quota:
-      break
-    if quota['metric'] == 'CPUS':
-      cpu_quota = quota
-      continue
-    if quota['metric'] == 'PREEMPTIBLE_CPUS':
-      preemptible_quota = quota
-      continue
-  assert preemptible_quota or cpu_quota
-
-  if not preemptible_quota['limit']:
-    # Preemptible quota is not set. Obey the CPU quota since that limits us.
-    quota = cpu_quota
-  else:
-    quota = preemptible_quota
-  assert quota['limit'], quota
-
-  # TODO(metzman): Do this in a more configurable way.
-  # We need this because us-central1 and us-east4 have different numbers of
-  # cores alloted to us in their quota. Treat them the same to simplify things.
-  limit = min(quota['limit'], 100_000)
-  project_config = local_config.ProjectConfig()
-  # On OSS-Fuzz there is a limit to the number of CPUs we can use.
-  limit = min(limit, project_config.get('schedule_fuzz.cpu_limit', limit))
-  return limit, quota['usage']
-
-
 class BaseFuzzTaskScheduler:
   """Base fuzz task scheduler for any deployment of ClusterFuzz."""
 
-  def __init__(self, num_cpus):
-    self.num_cpus = num_cpus
+  def __init__(self, num_tasks):
+    self.num_tasks = num_tasks
 
   def get_fuzz_tasks(self):
     raise NotImplementedError('Child class must implement.')
-
-  def _get_cpus_per_fuzz_job(self, job_name):
-    del job_name
-    return CPUS_PER_FUZZ_JOB
 
 
 class FuzzTaskCandidate:
@@ -159,7 +99,7 @@ class FuzzTaskCandidate:
 class OssfuzzFuzzTaskScheduler(BaseFuzzTaskScheduler):
   """Fuzz task scheduler for OSS-Fuzz."""
 
-  def get_fuzz_tasks(self) -> Dict[str, tasks.Task]:
+  def get_fuzz_tasks(self) -> list[tasks.Task]:
     # TODO(metzman): Handle high end.
     # A job's weight is determined by its own weight and the weight of the
     # project is a part of. First get project weights.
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
I'm not going to do a thorough review but will share context to avoid a production incident.
I think it's likely there will be infinite queuing if https://github.com/google/clusterfuzz/pull/5140 is not landed before this.

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -35,19 +33,9 @@
...
-
-def _get_quotas(creds, project, region):
-  compute = discovery.build('compute', 'v1', credentials=creds)
-  return compute.regions().get(  # pylint: disable=no-member
-      region=region, project=project).execute()['quotas']
-
-
-def count_unacked(creds, project_id, subscription_id):
-  """Counts the unacked messages in |subscription_id|."""
+def get_queue_size(creds, project_id, subscription_id):
```
jonathanmetzman, line 33: It's probably worth notiing somewhere that the queue size metric is delayed by about 5 minutes.

jonathanmetzman:
I'd run this locally using butler scripts to make sure it behaves nicely.

javanlacerda:
> I'd run this locally using butler scripts to make sure it behaves nicely.

Its running in dev for 4 days already :)

javanlacerda:
> I'm not going to do a thorough review but will share context to avoid a production incident. I think it's likely there will be infinite queuing if #5140 is not landed before this.

Will not because this should be landed only after https://github.com/google/clusterfuzz/pull/5150. And even we don't have the job limiter for batch yet, we can control how many tasks will be forward to there by the RemoteTaskGate frequencies.

src/clusterfuzz/_internal/cron/schedule_fuzz.py
javanlacerda: Not sure I got your point.

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -223,7 +167,7 @@ def get_fuzz_tasks(self) -> Dict[str, tasks.Task]:
       weights.append(fuzz_task_candidate.weight)
 
     # TODO(metzman): Handle high-end jobs correctly.
```
decoNR: Just for clarification, how do these new changes relate to this comment?

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -276,7 +227,7 @@ def get_fuzz_tasks(self) -> List[tasks.Task]:
       fuzz_task_candidates.append(fuzz_task_candidate)
 
     weights = [candidate.weight for candidate in fuzz_task_candidates]
-    num_instances = int(self.num_cpus / self._get_cpus_per_fuzz_job(None))
+    num_instances = self.num_tasks
```
decoNR: Maybe change this variable name?

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -295,101 +246,41 @@ def get_fuzz_tasks(self) -> List[tasks.Task]:
...
   fuzz_tasks = scheduler.get_fuzz_tasks()
   return fuzz_tasks
 
 
-def get_batch_regions(batch_config):
-  fuzz_subconf_names = {
-      subconf['name'] for subconf in batch_config.get(
-          'mapping.LINUX-PREEMPTIBLE-UNPRIVILEGED.subconfigs')
-  }
+PREPROCESS_TARGET_SIZE_DEFAULT = 10000
```
decoNR: I think it would be easier to read if this is at the top of the file. Is there a specific reason for it to be here?

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -223,7 +167,7 @@ def get_fuzz_tasks(self) -> Dict[str, tasks.Task]:
       weights.append(fuzz_task_candidate.weight)
 
     # TODO(metzman): Handle high-end jobs correctly.
-    num_instances = int(self.num_cpus / self._get_cpus_per_fuzz_job(None))
+    num_instances = self.num_tasks
```
decoNR: ditto.

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -247,7 +191,14 @@ class ChromeFuzzTaskScheduler(BaseFuzzTaskScheduler):
...
   def __init__(self, *args, **kwargs):
     super().__init__(*args, **kwargs)
-    self.num_cpus = respect_project_max_cpus(self.num_cpus)
+    self._apply_project_limit()
+
+  def _apply_project_limit(self):
+    conf = local_config.ProjectConfig()
+    max_cpus_per_schedule = conf.get('max_cpus_per_schedule')
+    if max_cpus_per_schedule:
+      max_tasks = int(max_cpus_per_schedule / CPUS_PER_FUZZ_JOB)
```
decoNR: Can this be deprecated? It appears to be used only here.

src/clusterfuzz/_internal/cron/schedule_fuzz.py
javanlacerda: I don't know what @jonathanmetzman meant here, but AFAIK we doesn't support high-end jobs anymore.

cemon721-a11y:
​"Hi Javan,
​As a Research Analysis Researcher, I've been following this thread. I'm concerned about how the 'infinite queuing' risk or the manual control of 'RemoteTaskGate frequencies' might affect the consistency of our data processing and overall system stability.
​Could you please confirm if I should pause any ongoing research tasks or data extraction until PR #5140 and #5150 are fully landed? I want to ensure that our research outputs aren't compromised by these technical dependencies.
​Thanks for the context!"

src/clusterfuzz/_internal/cron/schedule_fuzz.py
javanlacerda: I think so. I'll remove it

src/clusterfuzz/_internal/cron/schedule_fuzz.py
javanlacerda: It sounds good. I'll move it to `fuzz_tasks`

src/clusterfuzz/_internal/cron/schedule_fuzz.py
javanlacerda: Moving to `fuzz_tasks`

src/clusterfuzz/_internal/cron/schedule_fuzz.py
javanlacerda: It should be in the top. Moving.

jonathanmetzman:
> > I'm not going to do a thorough review but will share context to avoid a production incident. I think it's likely there will be infinite queuing if #5140 is not landed before this.
> 
> Will not because this should be landed only after #5150. And even we don't have the job limiter for batch yet, we can control how many tasks will be forward to there by the RemoteTaskGate frequencies.

Great. I haven't had a chance to review/understand that PR, but are you relying sending everything to k8s to avoide infinite queuing in batch? Otherwise I don't see how we ever know if batch is full/queueing without querying the batch API. I trust you!

src/clusterfuzz/_internal/cron/schedule_fuzz.py
jonathanmetzman: Sorry, I mean that we should mention the need for delays. If the cron runs too often, it might check the queue size before the metric has actually updated to reflect the jobs it just added. Does this make sense?

javanlacerda: That's a good point. But the idea is to tweak this by using the feature flags to have the balance on the size of the queue and the frequency of the cron execution.

cemon721-a11y:
​Hi Javan,

​Thank you for your guidance and help over the past 4 days regarding the
technical issues in this thread.

​I am writing to let you know that I am closing my requests and will be
leaving GitHub entirely. As the founder of *Hill Tope Europe*, I have
realized that this platform is not the right fit for my current business
needs. I am concerned about the privacy of my data and the public exposure
of my ideas, so I have decided to reset my repositories and deactivate my
account to protect my startup’s content.

​I apologize for any inconvenience this quick departure causes, but I must
prioritize the security of my intellectual property. Thank you for your
expertise and for your efforts to prevent production incidents during our
discussion.

​Best regards,

​*Shakib Sadman* Founder, Hill Tope Europe

(GitHub: cemon721-a11y)

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -14,40 +14,24 @@
...
-from clusterfuzz._internal.batch import service as batch
-from clusterfuzz._internal.config import local_config
 from clusterfuzz._internal.datastore import data_types
 from clusterfuzz._internal.datastore import ndb_utils
 from clusterfuzz._internal.google_cloud_utils import credentials
 from clusterfuzz._internal.metrics import logs
 
-# TODO(metzman): Actually implement this.
-CPUS_PER_FUZZ_JOB = 2
+PREPROCESS_TARGET_SIZE_DEFAULT = 10000
```
ViniciustCosta, line 30: How did you reach this number? Shouldn't the default depend on the project config(external, internal, ...)? 

Also, maybe we should create a rationale to define this number based on the amount of tworkers + batch/k8s quota

src/clusterfuzz/_internal/cron/schedule_fuzz.py
decoNR: Not sure if I understood it correctly, the point is to not run `schedule_fuzz` too often, right?

src/clusterfuzz/_internal/cron/schedule_fuzz.py
javanlacerda: Its an arbitrary default number, it will be controlled through the feature flag. And indeed we will tweak each project based on on the fleet.

ViniciustCosta: I still think that regardless of the feature flag, we should have a default defined on the project config (similar to what was done with the cpu count)

javanlacerda:
> Approving this to unblock the architecture changes so we can monitor them in prod.
> 
> I'm still not sure how this will not overload batch if batch does not have the job limiter implemented yet (and scheduling based on the preprocess queue size seems to have this as an assumption to work).

I've described the rationale here https://github.com/google/clusterfuzz/pull/5150#discussion_r2766649870

PR #5074, opened 2025-12-10
Update Component ID Logic

AashutoshMurthy:
Some issues are not getting filed into the desired component of their code owners. This component is provided as a metadata.

The current change aims at emphasising this metadata component by removing the automatic component.

Please refer to [*b/438511288*](https://b.corp.google.com/issues/438511288) for additional details.

PR #5154, opened 2026-02-03
Resolve AttributeError in issue_filer.py

AashutoshMurthy:
Safeguard the process of issue filing by ensuring the presence of attributes within `Issue` variants before accessing them.

This issue occurs due to the changes made in #5074. #5074 adds debugging logs, and one of the logs tries to access an attribute (`component_id`) of a class (`Issue`), resulting in the following error:
**'Issue' object has no attribute 'component_id'**

This issue has blocked the issue filing process.

The current changes aim to resolve this issue using `getattr()`, which checks if an attribute exists and only then accesses it.

### DIFF
--- a/src/clusterfuzz/_internal/issue_management/issue_filer.py
+++ b/src/clusterfuzz/_internal/issue_management/issue_filer.py
@@ -327,7 +327,8 @@ def file_issue(testcase,
 
   issue = issue_tracker.new_issue()
   if issue_tracker.project == 'google-buganizer':
-    logs.info(f"New issue's default component id = {issue.component_id}")
+    default_component_id = getattr(issue, 'component_id', None)
+    logs.info(f"New issue's default component id = {default_component_id}")
   issue.title = data_handler.get_issue_summary(testcase)
   issue.body = data_handler.get_issue_description(
       testcase, reporter=user_email, show_reporter=True)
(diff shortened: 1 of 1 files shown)


cemon721-a11y:
Thanks for the quick fix! Using getattr() to handle the missing attribute looks like the right approach. This will unblock our workflow."

cemon721-a11y:
​*Hi Aashutosh,*

​I hope you are doing well.

​I saw your recent Pull Request (*#5154*) where you addressed the
AttributeError in issue_filer.py. Thank you for working on this fix!

​I wanted to reach out because I am still learning the ropes of open-source
contribution and reviewing issues. To be honest, I don't have extensive
knowledge about all the rules and technical complexities of this project
yet.

​Could you please help me understand if there are specific things I should
look for when reviewing such PRs? Since I want to make sure the changes are
perfectly fine before they move forward, your guidance on how to verify
this fix would be very helpful for my learning process.

​Looking forward to hearing from you.

src/clusterfuzz/_internal/issue_management/issue_filer.py
```diff
@@ -492,8 +493,10 @@ def file_issue(testcase,
     if issue_tracker.project == 'google-buganizer':
       logs.info('The values of Component IDs:')
       logs.info(f'1. Backing: {list(issue.components)}')
-      logs.info(f'2. Removed: {list(issue.components.removed)}')
-      logs.info(f'3. Added: {list(issue.components.added)}')
+      removed = list(getattr(issue.components, 'removed', None))
```
ViniciustCosta: You should check if it is `None` before casting it into a list, as trying to do `list(None)` will raise a `TypeError`.

AashutoshMurthy: Thank you for catching this. I have updated the code so that `getattr()` returns an empty list instead of `None`.

PR #5155, opened 2026-02-06
Fix group settings to allow external member into oss-fuzz cc groups

ViniciustCosta:
Creating groups through the cloud identity API uses a default set of access settings. We need to change these in order for them to work as CC groups for OSS-Fuzz projects. Mainly, allowing external members to be added to the group.  These changes cannot be done through the identity API, so we need to use the groups settings API (which I enabled for the clusterfuzz-external deployment).

I added some other settings as safeguard to avoid issues with these groups (as they will be in the public tracker), such as who can view/post/contact/manage the group to only owners/managers. Also, fixed some single/double quote style.

**Tests**
* Added unit test.
* Run locally impersonating the clusterfuzz-external service account.
  * Output: https://paste.googleplex.com/4898709642018816 (errors in the first run are expected as googleapiclient logs an error when the checking for the group and it does not exist).
  * Group settings after executing it: https://screenshot.googleplex.com/5ymqVPoFMwHYraN.png

Context: b/477964128

### DIFF
--- a/src/clusterfuzz/_internal/tests/appengine/handlers/cron/oss_fuzz_cc_groups_test.py
+++ b/src/clusterfuzz/_internal/tests/appengine/handlers/cron/oss_fuzz_cc_groups_test.py
@@ -75,8 +75,8 @@ class OssFuzzCcGroupsTest(unittest.TestCase):
     self.mock.delete_google_group_membership.assert_called_with(
         'group2_id', 'member1@example.com', 'membership1')
 
-  def test_set_access_settings_to_group(self):
-    """Test sets oss-fuzz cc group access settings for a new group."""
+  def test_main_exec_for_new_group(self):
+    """Test main execution for a newly created group."""
     self.mock.get_oss_fuzz_projects.return_value = [('project1', {'info': 1})]
     self.mock.get_group_id.return_value = '1'
     self.mock.get_google_group_memberships.return_value = {


src/clusterfuzz/_internal/google_cloud_utils/google_groups.py
```diff
@@ -49,7 +59,7 @@ def get_group_id(group_name: str, exists_check: bool = False) -> str | None:
     return response.get('name')
   except errors.HttpError:
     if not exists_check:
-      logs.warning(f"Unable to look up group {group_name}.")
+      logs.warning(f'Unable to look up group {group_name}.')
```
decoNR, line 62: Thanks :)

src/clusterfuzz/_internal/tests/appengine/handlers/cron/oss_fuzz_cc_groups_test.py
```diff
@@ -74,6 +75,26 @@ def test_main(self):
     self.mock.delete_google_group_membership.assert_called_with(
         'group2_id', 'member1@example.com', 'membership1')
 
+  def test_set_access_settings_to_group(self):
+    """Test sets oss-fuzz cc group access settings for a new group."""
```
decoNR: As the code mocks most of the functions, this test is mainly testing the workflow of `oss_fuzz_cc_groups.main()', right? Maybe change the description?

decoNR (approved):

src/clusterfuzz/_internal/tests/appengine/handlers/cron/oss_fuzz_cc_groups_test.py
ViniciustCosta: Sure, done!

PR #5158, opened 2026-02-09
Implement job limiter for GCP Batch

javanlacerda:
It implements the a job limiter for the GCP Batch adapter for remote tasks.

It uses a private API for checking the availability of the regions for scheduling jobs, if all of them are loaded, the tasks are returned as unscheduled tasks and sent back to the queue.

### DIFF
--- a/src/clusterfuzz/_internal/batch/service.py
+++ b/src/clusterfuzz/_internal/batch/service.py
@@ -1,8 +1,8 @@
 # Copyright 2025 Google LLC
 #
-# Licensed under the Apache License, Version 2.0
-# (the "License"); you may not use this file except in compliance with
-# the License. You may obtain a copy of the License at
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
 #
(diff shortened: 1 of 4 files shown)


src/clusterfuzz/_internal/batch/service.py
```diff
@@ -1,8 +1,8 @@
 # Copyright 2025 Google LLC
 #
-# Licensed under the Apache License, Version 2.0
-# (the "License"); you may not use this file except in compliance with
-# the License. You may obtain a copy of the License at
+# Licensed under the Apache License, Version 2.0 (the "License");
```
ViniciustCosta, line 3: Is this change needed?

src/clusterfuzz/_internal/batch/service.py
```diff
@@ -65,6 +70,13 @@
 # See https://cloud.google.com/batch/quotas#job_limits
 MAX_CONCURRENT_VMS_PER_JOB = 1000
 
+MAX_QUEUE_SIZE = 100
```
ViniciustCosta, line 73: Nit: could this be a project config?

PR #5163, opened 2026-02-11
Fix handle of missing customer id in oss-fuzz groups

ViniciustCosta:
Doing `str()` for the value of customer ID from config was preventing the code from checking if it is missing, hindering our debug capability.

Also, added a log for the start of the cronjob.

### DIFF
--- a/src/clusterfuzz/_internal/google_cloud_utils/google_groups.py
+++ b/src/clusterfuzz/_internal/google_cloud_utils/google_groups.py
@@ -117,8 +117,8 @@ def create_google_group(group_name: str,
   """Create a google group."""
   identity_service = get_identity_api()
 
-  customer_id = customer_id or str(
-      local_config.ProjectConfig().get('groups_customer_id'))
+  customer_id = customer_id or local_config.ProjectConfig().get(
+      'groups_customer_id')
   if not customer_id:
     logs.error('No customer ID set. Unable to create a new google group.')
     return None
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/google_cloud_utils/google_groups.py
```diff
@@ -117,8 +117,8 @@ def create_google_group(group_name: str,
   """Create a google group."""
   identity_service = get_identity_api()
 
-  customer_id = customer_id or str(
-      local_config.ProjectConfig().get('groups_customer_id'))
+  customer_id = customer_id or local_config.ProjectConfig().get(
```
decoNR, line 120: Not sure, but maybe it is needed to cast `customer_id` to `str` after line 122 due to line 128.

ViniciustCosta: Not needed tbh (and it is unusual to do it in cf codebase), the getter for project config already deals with evaluating the value from the yaml. If the value is not a string, it should break as a genuine unexpected exception due to misconfiguration.

decoNR: Thanks! I pointed that out because the value was being cast before this change, so I wanted to be sure.

decoNR (approved):

src/clusterfuzz/_internal/google_cloud_utils/google_groups.py
ViniciustCosta: Yeah, it was my bad, I had done it to avoid the vscode linter from complaining.

Like probably the "rightest" thing to do is cast it to str within a try-catch block, but it seems unnecessary extra work to do this every time we get a value from project config.

PR #5164, opened 2026-02-11
[tests-syncer] Rename fuzz files in the whole archive

mi-ac:
We already renamed "fuzz-" test cases from crash tests, so that clusterfuzz doesn't confuse them with output cases.

This change expands this functionality to all synced folders, also external repos.

BUG=http://b/379684065

### DIFF
--- a/src/clusterfuzz/_internal/cron/oss_fuzz_cc_groups.py
+++ b/src/clusterfuzz/_internal/cron/oss_fuzz_cc_groups.py
@@ -72,6 +72,7 @@ def sync_project_cc_group(project_name, info):
 
 def main():
   """Sync OSS-Fuzz projects groups used to CC owners in the issue tracker."""
+  logs.info('OSS-Fuzz CC groups sync started.')
   projects = project_setup.get_oss_fuzz_projects()
 
   for project, info in projects:
--- a/src/clusterfuzz/_internal/google_cloud_utils/google_groups.py
+++ b/src/clusterfuzz/_internal/google_cloud_utils/google_groups.py
@@ -117,8 +117,8 @@ def create_google_group(group_name: str,
   """Create a google group."""
   identity_service = get_identity_api()
 
-  customer_id = customer_id or str(
-      local_config.ProjectConfig().get('groups_customer_id'))
+  customer_id = customer_id or local_config.ProjectConfig().get(
+      'groups_customer_id')
   if not customer_id:
     logs.error('No customer ID set. Unable to create a new google group.')
     return None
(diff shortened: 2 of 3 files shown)


src/clusterfuzz/_internal/cron/chrome_tests_syncer.py
```diff
@@ -253,6 +236,30 @@ def filter_members(member, path):
   shell.remove_file(local_archive)
 
 
+def rename_testcase_files(directory):
+  # Rename all fuzzed testcase files as regular files.
+  for root, _, files in os.walk(directory):
```
mi-ac, line 244: This moved from above 1:1.

src/clusterfuzz/_internal/cron/chrome_tests_syncer.py
```diff
@@ -300,23 +307,27 @@ def sync_tests(tests_archive_bucket: str, tests_archive_name: str,
   # compatibility with older testcases.
   create_symbolic_link(tests_directory, 'src/third_party/blink/web_tests',
                        'LayoutTests')
+  
+  test_folders = [
+      'CrashTests',
```
mi-ac, line 316: The crash tests are included here, so this preserves the functionality we had before, where we called the renaming loop within the unpack_crash_testcases function.

javanlacerda:
Could you run `yapf -p -i src/clusterfuzz/_internal/cron/chrome_tests_syncer.py` for fixing the lint.

src/clusterfuzz/_internal/cron/chrome_tests_syncer.py
```diff
@@ -253,6 +236,30 @@ def filter_members(member, path):
...
+          testcase_manager.FUZZ_PREFIX):]
+      stripped_file_path = os.path.join(
+          os.path.dirname(file_path), stripped_file_name)
+      try:
+        os.rename(file_path, stripped_file_path)
+      except Exception as e:
+        raise RuntimeError(f'Failed to rename testcase {file_path}') from e
+
+
+def clean_up_filenames(tests_directory, test_folders):
```
javanlacerda, line 260: Where is this function being called?

mi-ac: Good catch :) - it's called now

mi-ac:
> Is there any unit tests that could be created for checking this change?

Not that I know of. If there were existing testing, I'd be happy to add something. But I'm not enthusiastic to be the first to set this up.

However, breaking the test-syncer is somewhat low risk. If something doesn't work, we can revert. Then we don't have an archive for a day.

The biggest risk is I guess to not rename the files at all (just like missing calling the function, which you caught in review). If more fuzz- prefixed tests are in the archive, we'd get even more garbage runs. But also that would not be a big deal for one day.

mi-ac:
Finally the linter is happy...

letitz:
Hey, this is clever! Thanks for doing this @mi-ac !

PR #5170, opened 2026-02-25
Swarming: Implement JSON env var encapsulation and v2 API protos

IvanBM18:
## Finishes the last details in the base of clusterfuzz-swarming integration.

This PR finalizes the base integration for Clusterfuzz-Swarming. The primary focus is ensuring environment variable persistence across Docker/ swarming bots and laying the groundwork for the Swarming v2 API.

- Encapsulates all of the env vars in a single JSON env var, so that this JSON is be processed by a startup script. This is done so that the env vars are not lost in swarming bots that launch clusterfuzz in a docker container.
- Simplified the way we check if a task is a swarming task to avoid creating unused newTaskRequest objects and too many quries into the db, resulting in a refactor of various related methods
- Adds new protos for the swarming v2 api
- Also adds some unit tests.

This integration will later be refactored to use remote_tasks once the remote task gate allows to send tasks to specific backends always that some conditions are given.

Bug: b/479476219

### DIFF
--- a/src/clusterfuzz/_internal/swarming/__init__.py
+++ b/src/clusterfuzz/_internal/swarming/__init__.py
@@ -65,13 +65,11 @@ def create_new_task_request(command: str, job_name: str, download_url: str
   Returns None if the task should'nt be executed on swarming"""
   job = data_types.Job.query(data_types.Job.name == job_name).get()
   if job is None:
-    print("Job is None")
     return None
 
   swarming_config = _get_swarming_config()
   instance_spec = _get_instance_spec(swarming_config, job)
   if instance_spec is None:
-    print("Instance spec is None")
     return None
 
   swarming_pool = swarming_config.get('swarming_pool')
@@ -111,11 +109,13 @@ def create_new_task_request(command: str, job_name: str, download_url: str
       task_environment.append(
           swarming_pb2.StringPair(key=var['key'], value=var['value']))  # pylint: disable=no-member
 
+  swarming_bot_environment = []
   if instance_spec.get('docker_image'):
-    task_environment.append(
+    swarming_bot_environment.append(
         swarming_pb2.StringPair(  # pylint: disable=no-member
             key='DOCKER_IMAGE',
             value=instance_spec['docker_image']))
+  swarming_bot_environment.append(_env_vars_to_json(task_environment))
 
   task_dimensions = [
       swarming_pb2.StringPair(key='os', value=job.platform),  # pylint: disable=no-member
@@ -146,7 +146,7 @@ def create_new_task_request(command: str, job_name: str, download_url: str
                   cipd_input=cipd_input,
                   cas_input_root=cas_input_root,
                   execution_timeout_secs=execution_timeout_secs,
-                  env=[_env_vars_to_json(task_environment)],
+                  env=swarming_bot_environment,
                   env_prefixes=env_prefixes,
                   secret_bytes=base64.b64encode(download_url.encode('utf-8'))))
       ])
--- a/src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
+++ b/src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
@@ -71,9 +71,14 @@ class SwarmingTest(unittest.TestCase):
                     execution_timeout_secs=86400,
                     env=[
                         swarming_pb2.StringPair(
+                            key='DOCKER_IMAGE',
+                            value=
+                            'gcr.io/clusterfuzz-images/base:a2f4dd6-202202070654'
+                        ),
+                        swarming_pb2.StringPair(
                             key='DOCKER_ENV_VARS',
                             value=
-                            '{"UWORKER": "True", "SWARMING_BOT": "True", "LOG_TO_GCP": "True", "LOGGING_CLOUD_PROJECT_ID": "project_id", "DOCKER_IMAGE": "gcr.io/clusterfuzz-images/base:a2f4dd6-202202070654"}'
+                            '{"UWORKER": "True", "SWARMING_BOT": "True", "LOG_TO_GCP": "True", "LOGGING_CLOUD_PROJECT_ID": "project_id"}'
                         ),
                     ],
                     secret_bytes=base64.b64encode(
@@ -183,9 +188,14 @@ class SwarmingTest(unittest.TestCase):
                     execution_timeout_secs=12345,
                     env=[
                         swarming_pb2.StringPair(
+                            key='DOCKER_IMAGE',
+                            value=
+                            'gcr.io/clusterfuzz-images/base:a2f4dd6-202202070654'
+                        ),
+                        swarming_pb2.StringPair(
                             key='DOCKER_ENV_VARS',
                             value=
-                            '{"UWORKER": "True", "SWARMING_BOT": "True", "LOG_TO_GCP": "True", "LOGGING_CLOUD_PROJECT_ID": "project_id", "DOCKER_IMAGE": "gcr.io/clusterfuzz-images/base:a2f4dd6-202202070654"}'
+                            '{"UWORKER": "True", "SWARMING_BOT": "True", "LOG_TO_GCP": "True", "LOGGING_CLOUD_PROJECT_ID": "project_id"}'
                         ),
                     ],
                     secret_bytes=base64.b64encode(
@@ -227,9 +237,14 @@ class SwarmingTest(unittest.TestCase):
                     execution_timeout_secs=12345,
                     env=[
                         swarming_pb2.StringPair(
+                            key='DOCKER_IMAGE',
+                            value=
+                            'gcr.io/clusterfuzz-images/base:a2f4dd6-202202070654'
+                        ),
+                        swarming_pb2.StringPair(
                             key='DOCKER_ENV_VARS',
                             value=
-                            '{"UWORKER": "True", "SWARMING_BOT": "True", "LOG_TO_GCP": "True", "LOGGING_CLOUD_PROJECT_ID": "project_id", "DOCKER_IMAGE": "gcr.io/clusterfuzz-images/base:a2f4dd6-202202070654"}'
+                            '{"UWORKER": "True", "SWARMING_BOT": "True", "LOG_TO_GCP": "True", "LOGGING_CLOUD_PROJECT_ID": "project_id"}'
                         ),
                     ],
                     secret_bytes=base64.b64encode(


src/clusterfuzz/_internal/swarming/__init__.py
```diff
@@ -34,7 +35,8 @@ def _requires_gpu() -> bool:
 
 
 def is_swarming_task(command: str, job_name: str):
-  """Returns True if the task is supposed to run on swarming."""
+  """Returns True if the task is supposed to run on swarming.
+  Currently swarming only supports Linux tasks."""
```
jardondiego: Let's link a bug on this comment line.
See go/todo-style

jardondiego:
I think it's important to be more specific with both the title and description.

src/clusterfuzz/_internal/protos/swarming_pb2_grpc.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2023 Google LLC
```
jardondiego: Sorry I am not sure if these gRPC should be on this PR. Are they needed for changes in `swarming/__init__.py`?

IvanBM18:
> I think it's important to be more specific with both the title and description.

Yeah i guess it was a little bit to generic, thanks for the feedback

src/clusterfuzz/_internal/protos/swarming_pb2_grpc.py
IvanBM18: All of the used protos in this repo live under `src/_internal/protos`, i just added the ones that were missing but maybe @letitz can give more context here

src/clusterfuzz/_internal/swarming/__init__.py
IvanBM18: Done :D thanks!

src/clusterfuzz/_internal/swarming/__init__.py
```diff
@@ -143,6 +148,17 @@ def _get_new_task_spec(command: str, job_name: str,
   return new_task_request
 
 
+def _compress_env_vars(
```
jardondiego: nit: We could also name it `_env_vars_to_json` to be more clear.

src/clusterfuzz/_internal/protos/swarming_pb2.pyi
```diff
@@ -0,0 +1,689 @@
+# Copyright 2023 Google LLC
```
javanlacerda: Could we fix the date here? (Even being a generated code from the proto)

src/clusterfuzz/_internal/swarming/__init__.py
```diff
@@ -35,8 +36,10 @@ def _requires_gpu() -> bool:
 
 def is_swarming_task(command: str, job_name: str):
```
javanlacerda: Are moving this file logic to be under a swarming service that implements the RemoteTaskInterface?

jardondiego: Yes. The plan is to eventually migrate it to use the Remote Task interface.

src/clusterfuzz/_internal/swarming/__init__.py
IvanBM18: Done :D thanks!

src/clusterfuzz/_internal/protos/swarming_pb2.pyi
IvanBM18: Sure, thanks for pointing it out

src/clusterfuzz/_internal/protos/swarming_pb2_grpc.py
```diff
@@ -0,0 +1,18 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Generated by the gRPC Python protocol compiler plugin. DO NOT EDIT!
+"""Client and server classes corresponding to protobuf-defined services."""
+import grpc
```
letitz, line 17: Is this file missing some content? I see lots more stuff in `swarming_pb2.pyi` that seems to be missing here. What am I missing?

src/clusterfuzz/_internal/swarming/__init__.py
```diff
@@ -35,8 +36,10 @@ def _requires_gpu() -> bool:
...
   """Returns True if the task is supposed to run on swarming."""
+  # TODO: b/487716733 - Trigger swarming tasks for MAC and Windows
   job = data_types.Job.query(data_types.Job.name == job_name).get()
-  if not job or not _requires_gpu():
+  if not job or not _requires_gpu() or str(
+      job.platform).upper() == 'MAC' or str(job.platform).upper() == 'WINDOWS':
```
letitz: How about we drop `_requires_gpu()` and directly move to checking that the job specifies a `SWARMING_DIMENSIONS` env var? `_get_new_task_spec()` will do the platform check implicitly by looking up an instance spec by `job.platform`.

Going further, we're computing the task spec twice. Here we compute it and throw it away, then if this function returns true we will end up calling `_get_new_task_spec()` again to use it. We could instead refactor this to export `get_new_task_spec()` (perhaps renaming it to `create_new_task_request()`) and document that the function returns `None` iff the task should not be executed on swarming. Then we can make `push_swarming_task()` take a `NewTaskRequest` argument instead and pass the result of `get_new_task_spec()`?

The caller would then look like this:

```py
  request = swarming.create_new_task_request()
  if request is not None:
    swarming.push_task(request)
  else:
    # Whatever else
```

src/clusterfuzz/_internal/swarming/__init__.py
```diff
@@ -106,6 +109,8 @@ def _get_new_task_spec(command: str, job_name: str,
             key='DOCKER_IMAGE',
             value=instance_spec['docker_image']))
 
+  task_environment.append(_env_vars_to_json(task_environment))
```
letitz: IIUC we're putting env vars twice:

* once "naked" in `task_environment`
* and then again JSON-encoded as `CF_BOT_VARS`

It seems to me there are two sets of environment variables, which in my mind are disjoint but may have overlap if I'm missing something:

* variables used by the startup script outside docker
* variables used inside the docker container

I would expect the first set to be passed "naked" in `task_environment`, and the second set to be passed inside `CF_BOT_VARS`. Is that understanding correct?

src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
```diff
@@ -264,3 +284,40 @@ def test_push_swarming_task(self):
...
+
+    expected_result = swarming_pb2.StringPair(
+        key='CF_BOT_VARS', value='{"ENV_VAR1": "VALUE1", "ENV_VAR2": "VALUE2"}')
+    self.assertEqual(result, expected_result)
+
+  def test_env_vars_to_json_empty(self):
+    """Tests that _env_vars_to_json handles an empty list of environment variables safely."""
+    env_vars = []
+
+    result = swarming._env_vars_to_json(env_vars)  # pylint: disable=protected-access
```
letitz:
```suggestion
    result = swarming._env_vars_to_json([])  # pylint: disable=protected-access
```

src/clusterfuzz/_internal/protos/swarming_pb2_grpc.py
jardondiego: I actually believe these two files should not be on this pull request. I think we should have them in another.

IvanBM18: mmm since the proto files are auto generated im not sure if this file is missing something, i see other proto files being kinda empty too, for example: `src/clusterfuzz/i_nternal/protos/process_state_pb2_grpc.py`

So i guess this "empty" files only look like this for protos that end in _grpc.py

src/clusterfuzz/_internal/swarming/__init__.py
IvanBM18: Yeah they are kind of passed twice, the ones in `task_environment` are directly injected in the bots envs, we expecting to use them in future bot platforms(windows & mac) were we run clusterfuzz by pulling the code instead of running a docker container.

Were as` CF_BOT_VARS` are the same ones, but grouped and prepared to be the processed by a  bot startup script and passed down to a docker container. Does that make sense?

src/clusterfuzz/_internal/swarming/__init__.py
IvanBM18: Diego is already tweaking the required_gpu conditional in this PR: https://github.com/google/clusterfuzz/pull/5173
So do we want to also include some additonal chages here?

Regarding your other point... i also noticed that we were creating a newTaskRequest object twice, ill refactor the code to avoid this, thanks for the feedback :D

src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
IvanBM18: Done :D

src/clusterfuzz/_internal/protos/swarming_pb2_grpc.py
letitz: I don't understand how this works at all, but fair enough if this is how to do things for grpc.

src/clusterfuzz/_internal/swarming/__init__.py
letitz: Oh good, then never mind the `REQUIRES_GPU` stuff.

src/clusterfuzz/_internal/swarming/__init__.py
letitz: Got it. Then I think we should:

* rename `CF_BOT_VARS` to `DOCKER_ENV_VARS`
* remove the code to put most of them directly in `task_environment`
* add that back when we need it later, likely conditional on whether the config is missing `DOCKER_IMAGE`

src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
```diff
@@ -63,10 +63,9 @@ def _timestamp_now() -> Timestamp:
   return ts
 
 
-def _get_execution_mode(utask_module, job_type):
+def _get_execution_mode(job_type):
```
letitz, line 66: This change should be mentioned in the PR description.

src/clusterfuzz/_internal/bot/tasks/task_types.py
```diff
@@ -181,8 +181,10 @@ def execute(self, task_argument, job_type, uworker_env):
     if batch_service.is_remote_task(command, job_type):
```
letitz: We should invert the swarming and batch case. Currently we execute on swarming *if the task cannot run on batch*. We should execute on swarming *if we can run on swarming*.

Also, we lost the assert in the refactor.

In code:

```py
if request := swarming.create_new_task_request(...):
  swarming.push_swarming_task(request)
else:
  assert batch_service.is_remote_task(...)
  ...
```

src/clusterfuzz/_internal/swarming/__init__.py
```diff
@@ -33,16 +34,15 @@ def _requires_gpu() -> bool:
...
-    return False
-  try:
-    _get_new_task_spec(command, job_name, '')
-    return True
-  except ValueError:
+  swarming_config = _get_swarming_config()
+  platform_config = swarming_config.get('mapping').get(job.platform, None)
+  if not job or not _requires_gpu() or platform_config is None:
     return False
+  return True
```
letitz: Just return the boolean:


```suggestion
  return job and platform_config and _requires_gpu()
```

src/clusterfuzz/_internal/swarming/__init__.py
```diff
@@ -54,10 +54,15 @@ def _get_swarming_config():
...
-def _get_new_task_spec(command: str, job_name: str,
-                       download_url: str) -> swarming_pb2.NewTaskRequest:  # pylint: disable=no-member
-  """Gets the configured specifications for a swarming task."""
+def create_new_task_request(command: str, job_name: str, download_url: str
+                           ) -> swarming_pb2.NewTaskRequest | None:  # pylint: disable=no-member
+  """Gets the configured specifications for a swarming task. 
+  Returns None if the task should'nt be executed on swarming"""
   job = data_types.Job.query(data_types.Job.name == job_name).get()
+
+  if not is_swarming_task(job_name):
```
letitz: We're still looking up `job` twice in the DB, and finding `swarming_config` and `instance_spec` twice.

How about:

```py
def _get_instance_spec(swarming_config: ..., job: data_types.Job) -> ... | None:
  if not _requires_gpu():
    return None
  return swarming_config.get('mapping').get(job.platform, None)

def is_swarming_task(job_name: str) -> bool:
  if job is None:
    return False
  
  return _get_instance_spec(_get_swarming_config(), job) is not None

def create_new_task_request(...) -> ...:
  if job is None:
    return None

  swarming_config = _get_swarming_config()
  instance_spec = _get_instance_spec(swarming_config, job)
  if instance_spec is None:
    return None

  ...
```

I think it maximizes reuse and minimizes the risk of `is_swarming_task()` and `create_task_request()` diverging.

src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
```diff
@@ -32,16 +32,18 @@ class SwarmingTest(unittest.TestCase):
...
     self.mock._get_task_name.return_value = 'task_name'  # pylint: disable=protected-access
+    self.mock._requires_gpu.return_value = True  # pylint: disable=protected-access
     self.maxDiff = None
 
   def test_get_spec_from_config_with_docker_image(self):
     """Tests that _get_new_task_spec works as expected."""
     job = data_types.Job(name='libfuzzer_chrome_asan', platform='LINUX')
     job.put()
-    spec = swarming._get_new_task_spec(  # pylint: disable=protected-access
+    spec = swarming.create_new_task_request(  # pylint: disable=protected-access
```
letitz: No need for `# pylint: disable=protected-access` anymore.

src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
```diff
@@ -91,19 +93,19 @@ def test_get_spec_from_config_with_docker_image(self):
...
-  def test_get_spec_from_config_raises_error_on_unknown_config(self):
-    """Tests that _get_new_task_spec raises error when there's no mapping for the config."""
+  def test_get_spec_from_config_returns_none_on_unknown_config(self):
+    """Tests that _get_new_task_spec returns None when there's no mapping for the config."""
     job = data_types.Job(name='some_job_name', platform='UNKNOWN-PLATFORM')
     job.put()
-    with self.assertRaises(ValueError):
-      swarming._get_new_task_spec(  # pylint: disable=protected-access
-          'corpus_pruning', job.name, 'https://download_url')
+    spec = swarming.create_new_task_request(  # pylint: disable=protected-access
```
letitz: Ditto

src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
```diff
@@ -91,19 +93,19 @@ def test_get_spec_from_config_with_docker_image(self):
...
+        'corpus_pruning', job.name, 'https://download_url')
+    self.assertIsNone(spec)
 
   def test_get_spec_from_config_without_docker_image(self):
     """Tests that _get_new_task_spec works as expected (without a docker image)."""
     job = data_types.Job(name='libfuzzer_chrome_asan', platform='MAC')
     job.put()
-    spec = swarming._get_new_task_spec(  # pylint: disable=protected-access
+    spec = swarming.create_new_task_request(  # pylint: disable=protected-access
```
letitz: Ditto

src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
```diff
@@ -172,7 +174,7 @@ def test_get_spec_from_config_for_fuzz_task(self):
     """Tests that _get_new_task_spec works as expected for fuzz commands."""
     job = data_types.Job(name='libfuzzer_chrome_asan', platform='LINUX')
     job.put()
-    spec = swarming._get_new_task_spec(  # pylint: disable=protected-access
+    spec = swarming.create_new_task_request(  # pylint: disable=protected-access
```
letitz: Ditto

src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
IvanBM18: Done :D

src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
IvanBM18: Good one, i didn't payed attention to the comments when renaming things, thanks!

src/clusterfuzz/_internal/swarming/__init__.py
IvanBM18: Sure I'll refactor it, thanks!

src/clusterfuzz/_internal/swarming/__init__.py
IvanBM18: Note: With the refactor proposed below to avoid querying the DB twice this  no longer applies, still thanks for the feedback

src/clusterfuzz/_internal/bot/tasks/task_types.py
IvanBM18: Done!

src/clusterfuzz/_internal/swarming/__init__.py
IvanBM18: Ok, then, ill also update the startup script to look out for the correct env var name

sadmanshakib396-png:
Ok 💭 but liitle bit time need to understand about the theme

src/clusterfuzz/_internal/swarming/__init__.py
sadmanshakib396-png: > Yes. The plan is to eventually migrate it to use the Remote Task interface.

How it's not me...?

src/clusterfuzz/_internal/swarming/__init__.py
```diff
@@ -59,15 +64,16 @@ def create_new_task_request(command: str, job_name: str, download_url: str
   """Gets the configured specifications for a swarming task. 
   Returns None if the task should'nt be executed on swarming"""
   job = data_types.Job.query(data_types.Job.name == job_name).get()
-
-  if not is_swarming_task(job_name):
+  if job is None:
+    print("Job is None")
```
letitz: Leftover from testing I presume!

src/clusterfuzz/_internal/swarming/__init__.py
```diff
@@ -59,15 +64,16 @@ def create_new_task_request(command: str, job_name: str, download_url: str
...
     return None
 
-  config_name = job.platform
   swarming_config = _get_swarming_config()
-  instance_spec = swarming_config.get('mapping').get(config_name, None)
+  instance_spec = _get_instance_spec(swarming_config, job)
   if instance_spec is None:
-    raise ValueError(f'No mapping for {config_name}')
+    print("Instance spec is None")
```
letitz: Ditto

src/clusterfuzz/_internal/swarming/__init__.py
```diff
@@ -111,8 +117,6 @@ def create_new_task_request(command: str, job_name: str, download_url: str
             key='DOCKER_IMAGE',
```
letitz, line 116: I don't think we want `DOCKER_IMAGE` to be encoded in `DOCKER_ENV_VARS`, right? This is a value that is read only by the startup script? Does it need to be defined inside the docker container?

src/clusterfuzz/_internal/swarming/__init__.py
IvanBM18: yeah my bad 😅

src/clusterfuzz/_internal/swarming/__init__.py
IvanBM18: you are absolutely right, `DOCKER_IMAGE` should be available to the swarming bot without needing to "unpack it" from the `DOCKER_ENV_VARS` JSON, ill change this, thanks :D

javanlacerda:
It would be good testing it in the dev environment.

PR #5130, opened 2026-01-16
Update jdk version in ci image

ViniciustCosta:
Latest CI images use an updated gcloud version that requires at least JRE 21 to execute the datastore emulator.

PR #5171, opened 2026-02-25
Upgrade OpenJDK from 11 to 17

jardondiego:
Running `./local/install_deps.bash` is now broken due to the fact that `openjdk-11-jdk` is no longer present in default `apt` repositories.

### DIFF
--- a/local/install_deps_linux.bash
+++ b/local/install_deps_linux.bash
@@ -115,7 +115,7 @@ sudo apt-get update
 sudo apt-get install -y \
     docker-ce \
     google-cloud-cli \
-    openjdk-11-jdk \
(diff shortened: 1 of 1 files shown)


javanlacerda:
@ViniciustCosta could you take a look on it? I remember we had a problem when upgrading the JDK.

ViniciustCosta:
> @ViniciustCosta could you take a look on it? I remember we had a problem when upgrading the JDK.

I updated the jdk version in our CI image: https://github.com/google/clusterfuzz/pull/5130

This PR is updating the local development env. The only issue I can think is this hindering any local clusterfuzz server testing or having any issues with the butler deploy. But I think it should be safe to land

PR #5173, opened 2026-02-26
Update logic to identify Swarming task

jardondiego:
In order to actually identify Swarming tasks, we will look at the job configuration for a flag `IS_SWARMING_TASK`.


jardondiego:
> Question: This is the logic that runs inside the swarming machine it self right? Not the logic that schedules a swarming job.

Yes, but it is also used in several places, when dispatching the task to Swarming it first checks for that condition, as well as used to identify it as a remote task in `src/clusterfuzz/_internal/bot/tasks/task_types.py`.

letitz:
> > Question: This is the logic that runs inside the swarming machine it self right? Not the logic that schedules a swarming job.
> 
> Yes, but it is also used in several places, when dispatching the task to Swarming it first checks for that condition, as well as used to identify it as a remote task in `src/clusterfuzz/_internal/bot/tasks/task_types.py`.

I think the answer to @javanlacerda's question is largely no. This is mostly used to determine where to schedule `utask_main` [1]. It's somewhat anecdotally used during the execution of tasks, e.g. to annotate metrics with an `execution_mode` label [2].

[1]: https://github.com/google/clusterfuzz/blob/7bc23f5fb08a589218cadf0edb761423f28852ea/src/clusterfuzz/_internal/bot/tasks/task_types.py#L184
[2]: https://github.com/google/clusterfuzz/blob/7bc23f5fb08a589218cadf0edb761423f28852ea/src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py#L504

PR #5175, opened 2026-02-27
memoize queue size for utask-main and preprocess

javanlacerda:
I does a memorization of 60 seconds for the functions that queries the `utask-main` and `preprocess` queues size. It's required to stop reaching the queries per minute quota. 

It intends to fix [this error group](https://pantheon.corp.google.com/errors/detail/CPWHwu6d_4SnLg;locations=global?project=clusterfuzz-external&e=-13802955&mods=logs_tg_prod)

### DIFF
--- a/src/clusterfuzz/_internal/cron/schedule_fuzz.py
+++ b/src/clusterfuzz/_internal/cron/schedule_fuzz.py
@@ -20,6 +20,7 @@ import time
 from google.cloud import monitoring_v3
 
 from clusterfuzz._internal.base import feature_flags
+from clusterfuzz._internal.base import memoize
 from clusterfuzz._internal.base import tasks
 from clusterfuzz._internal.base import utils
 from clusterfuzz._internal.datastore import data_types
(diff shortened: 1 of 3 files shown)


src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -30,6 +31,7 @@
 PREPROCESS_TARGET_SIZE_DEFAULT = 10000
 
 
+@memoize.wrap(memoize.InMemory(60))
```
ViniciustCosta, line 34: I'm ok with adding it, but just that this might not be very relevant for schedule_fuzz, as each job execution queries this only once, right? 

I guess the utask_main one is more useful, as tworkers can keep trying to execute preprocess in task loop.

javanlacerda: Totally agree! I can remove it, but keeping could be good to have the patern of `caching metrics query`. Any strong thoughts ?

ViniciustCosta: Your call, feel free to merge :)

PR #5176, opened 2026-02-27
Normalize emails for oss-fuzz cc groups

ViniciustCosta:
**Motivation**
The cronjob is hitting some errors during group membership management that are mainly related to normalizing emails to match how Cloud Identity Groups API treats them. For instance:
* User emails containing invalid character '+' (usually used for tags, but unsupported by groups API):  `"Error(1005): Entity Id must not contain the '+' character."}]}]"`
* User emails that are normalized/canonicalized when added to google groups. Thus, when the membership is retrieved to compare with the email from the project yaml, the same email is considered distinct (which creates issues with trying incorrectly add/remove members):  `"Error(4003): Cannot create membership '****' in 'groups/***' because it already exists."`

**Changes**
Added a method to normalize the emails retrived from the ccs info in project yaml and from the membership query for the google groups. The normalization consists on:
* lowercasing.
* converting googlemail to gmail.
* removing pluses aliases/tags not supported by groups.
* removing dots (only when the domain is gmail, as dots are ignored in this domain).

These were loosely based on [go/email-address-normalization](http://go/email-address-normalization)

Context: b/477964128

### DIFF
--- a/src/clusterfuzz/_internal/cron/oss_fuzz_cc_groups.py
+++ b/src/clusterfuzz/_internal/cron/oss_fuzz_cc_groups.py
@@ -95,7 +95,7 @@ def sync_project_cc_group(project_name: str, ccs: list[str]):
     # Ignore the SA that created the group from members to delete.
     if not member or utils.is_service_account(member):
       continue
-    memebership_name = group_memberships[member]
+    memebership_name = group_memberships_norm[member]
     google_groups.delete_google_group_membership(group_id, member,
                                                  memebership_name)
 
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/cron/oss_fuzz_cc_groups.py
```diff
@@ -23,6 +23,30 @@
 _CC_GROUP_DESC = 'External CCs in OSS-Fuzz issue tracker for project'
 
 
+def normalize_email_for_group(email):
```
javanlacerda, line 26: You think you could leverage the https://github.com/JoshData/python-email-validator for this? It seems theres exactly the `normalize` functionality.

It is a validated package https://pypi.org/project/email-validator/

ViniciustCosta: I don't think so. This lib seems good for checking if an email address is syntactically valid, but it does not implement specific normalization rules like ignoring dots or plus aliases in gmail.

For example:
```py
>>> test = "test+alias@example.org"
>>> test2 = "Test@gmail.com"
>>> validate_email(test, check_deliverability=False).normalized
'test+alias@example.org'
>>> validate_email(test2, check_deliverability=False).normalized
'Test@gmail.com'
```

PR #5178, opened 2026-03-01
Add devcontainer configuration with Ubuntu and Python 3.11

jardondiego:
Devcontainers provide a consistent and reproducible development environment for all contributors. By containerizing the setup, we eliminate 'it works on my machine' issues and significantly simplify the onboarding process for new developers.

With this change, running `local/install_deps.bash` results in consistent outcomes. After this change lands, you should be able to spawn a fresh development environment where you can simply start playing around with ClusterFuzz.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -18,6 +18,6 @@
       }
     }
   },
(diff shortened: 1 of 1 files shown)


jardondiego:
I will open a new pull request with just one commit.

PR #5179, opened 2026-03-02
Add devcontainer configuration for Python 3.11

jardondiego:
Devcontainers provide a consistent and reproducible development environment for all contributors. By containerizing the setup, we eliminate 'it works on my machine' issues and significantly simplify the onboarding process for new developers. We only need to install the [Dev Containers extension for VS Code].(https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers)

With this change, running `local/install_deps.bash` results in consistent outcomes. After this change lands, you should be able to spawn a fresh development environment where you can simply start playing around with ClusterFuzz.

image

After that finishes, we can simply run `python butler.py run_server --skip-install-deps` to see.

Screenshot 2026-03-02 at 9 42 23 a m


IvanBM18:
LGTM, but can we also update the documentation to mention this new way of setting up a dev environment? Maybe somewhere under [docs/getting-started/local_instance.md](https://github.com/google/clusterfuzz/blob/master/docs/getting-started/local_instance.md)?

javanlacerda:
> LGTM, but can we also update the documentation to mention this new way of setting up a dev environment? Maybe somewhere under [docs/getting-started/local_instance.md](https://github.com/google/clusterfuzz/blob/master/docs/getting-started/local_instance.md)?

+1.

jardondiego:
> That's pretty nice! Thanks for that. LGTM, just adding a documentation would be great.
I will add the corresponding documentation in a follow-up PR later today. Thanks you guys for taking a look.

PR #5184, opened 2026-03-02
Add old state and diff logging on fuzzer update

tbantikyan:
Logs currently include only the fuzzer name. Previous state and diff information can help identify issues with any updates to fuzzers.

### DIFF
--- a/src/clusterfuzz/_internal/tests/appengine/handlers/fuzzers_test.py
+++ b/src/clusterfuzz/_internal/tests/appengine/handlers/fuzzers_test.py
@@ -1,0 +1,58 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tests for fuzzers handler."""
+# pylint: disable=protected-access
+
+import datetime
+import unittest
+
+from clusterfuzz._internal.datastore import data_types
+from handlers import fuzzers
+
+
+class BaseEditHandlerTest(unittest.TestCase):
+  """Test BaseEditHandler."""
+
+  def setUp(self):
+    self.handler = fuzzers.BaseEditHandler()
+
+  def test_get_fuzzer_state_str(self):
+    """Test that fuzzer state str excludes specific fields."""
+    fuzzer = data_types.Fuzzer(
+        name='test_fuzzer',
+        revision=1,
+        timeout=10,
+        result='bad',
+        console_output='some output',
+        result_timestamp=datetime.datetime(2021, 1, 1),
+        return_code=1,
+        sample_testcase='testcase',
+        stats_columns='cols',
+        stats_column_descriptions='desc',
+    )
+
+    state_str = self.handler._get_fuzzer_state_str(fuzzer)
+
+    self.assertIn('name: test_fuzzer', state_str)
+    self.assertIn('revision: 1', state_str)
+    self.assertIn('timeout: 10', state_str)
+
+    # Explicitly excluded fields
+    self.assertNotIn('result:', state_str)
+    self.assertNotIn('result_timestamp', state_str)
+    self.assertNotIn('console_output:', state_str)
+    self.assertNotIn('return_code:', state_str)
+    self.assertNotIn('sample_testcase:', state_str)
+    self.assertNotIn('stats_columns:', state_str)
+    self.assertNotIn('stats_column_descriptions:', state_str)
--- a/src/clusterfuzz/_internal/tests/appengine/libs/helpers_test.py
+++ b/src/clusterfuzz/_internal/tests/appengine/libs/helpers_test.py
@@ -159,3 +159,27 @@ class LogTest(unittest.TestCase):
     helpers.log('message', helpers.VIEW_OPERATION)
     self.mock.info.assert_called_once_with(
         'ClusterFuzz: %s (%s): %s.', helpers.VIEW_OPERATION, 'email', 'message')
+
+
+class DiffTest(unittest.TestCase):
+  """Test diff."""
+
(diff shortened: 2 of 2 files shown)


src/appengine/handlers/fuzzers.py
```diff
@@ -139,6 +139,23 @@ def _get_integer_value(self, key):
 
     return value
 
+  def _get_fuzzer_state_str(self, fuzzer):
```
letitz: Can we rely on `ndb.Model.to_dict()` [1] instead? I think `fuzzer` is of type `Fuzzer` [2]. It would avoid hardcoding the list of properties, and would allow us to compute a diff that is not just textual?

[1]: http://googleapis.dev/python/python-ndb/latest/model.html#google.cloud.ndb.model.Model.to_dict
[2]: https://github.com/google/clusterfuzz/blob/a96e146b165d49d81797c080b2719ef06ff4797c/src/clusterfuzz/_internal/datastore/data_types.py#L261

tbantikyan: Great suggestion. I also added a [blacklist](https://github.com/google/clusterfuzz/pull/5184/changes/a1ebe1321ce8796d8aa890a6eb1bc423f1015ed0#diff-3c1aa419095182eb8691389ad9b3e0cb04f407a0b2c5447240ef863dd9912461R37-R40) to cut down on noise.

tbantikyan: Although I am curious what the advantage is of computing non-textual diffs (presumably diff between two dict objects).

src/appengine/handlers/fuzzers.py
```diff
@@ -139,6 +143,13 @@ def _get_integer_value(self, key):
...
+    fuzzer_dict = fuzzer.to_dict(exclude=FUZZER_FIELDS_EXCLUDED_FROM_LOG)
+    state_str = ""
+    for key, val in fuzzer_dict.items():
+      state_str += f"{key}: {val}\n"
+    return state_str
```
letitz: I would just farm out to `json.dumps()`, but if you like this formatting how about:

```suggestion
    return '\n'.join(f"{key}: {val}" for key, val in fuzzer_dict.items())
```

src/appengine/handlers/fuzzers.py
letitz: Thanks, good idea!

Non-textual diffs is maybe not the right idea indeed given that we're not sure all fields can be compared for equality. Turning the values into strings is useful. I guess I just find that going for `difflib` is a bit of a big hammer. We could just do:

```py
  def to_str_dict(fuzzer: Fuzzer) -> dict[str, str]:
    fuzzer_dict = fuzzer.to_dict(exclude=...)
    return { key: str(val) for key, val in fuzzer_dict.items() }

  def key_difference(lhs: dict[str, str], rhs: dict[str, str]) -> list[str]:
    """Returns the list of keys whose values in `lhs` and `rhs` differ.
    
    Assumes `lhs` and `rhs` have the same set of keys.
    """
    return list(k for k, v in lhs.items() if rhs.get(k) != v)
    

  # later
  old_state = fuzzer.to_str_dict()

  # even later
  new_state = fuzzer.to_str_dict()
  changed_keys = key_difference(old_state, new_state)
```

Now that I've typed it out, I find it meh. It's maybe a bit simpler, but the logs will look worse. Do as you prefer!

src/appengine/libs/helpers.py
```diff
@@ -165,3 +166,19 @@ def log(message, operation_type):
   """Logs operation being carried by current logged-in user."""
   logging.info('ClusterFuzz: %s (%s): %s.', operation_type, get_user_email(),
                message)
+
+
+def diff(old_str, new_str):
+  """Generates the diff between the two provided strings."""
+  old_lines = old_str.splitlines(keepends=True)
```
letitz, line 173: Why do we `keepends`? Can we instead `return '\n'.join(clean_diff)` below?

src/appengine/libs/helpers.py
```diff
@@ -165,3 +166,19 @@ def log(message, operation_type):
...
+  new_lines = new_str.splitlines(keepends=True)
+
+  diff_generator = difflib.ndiff(old_lines, new_lines)
+  clean_diff = [
+      line for line in diff_generator
+      if line.startswith('- ') or line.startswith('+ ')
+  ]
+  if not clean_diff:
+    return "No changes detected."
```
letitz: I find this a bit surprising from the user's point of view. I would expect the diff to be the empty string, or `None`, when the inputs are equal. If you want to keep this, do document it in the docstring.

src/appengine/libs/helpers.py
```diff
@@ -165,3 +166,19 @@ def log(message, operation_type):
   """Logs operation being carried by current logged-in user."""
   logging.info('ClusterFuzz: %s (%s): %s.', operation_type, get_user_email(),
                message)
+
+
+def diff(old_str, new_str):
```
letitz: Let's annotate the types here, it's easy:


```suggestion
def diff(old_str: str, new_str: str) -> str:
```

src/appengine/handlers/fuzzers.py
```diff
@@ -139,6 +143,13 @@ def _get_integer_value(self, key):
 
     return value
 
+  def _get_fuzzer_state_str(self, fuzzer):
```
letitz:
```suggestion
  def _get_fuzzer_state_str(self, fuzzer: data_types: Fuzzer) -> str:
```

src/appengine/handlers/fuzzers.py
tbantikyan: I do like the simpler code, but I imagine it will be a worse user experience to parse differences by jumping between the "key differences" and the logged key-value pairs.

src/appengine/libs/helpers.py
tbantikyan: Without `keepends` the newline characters are stripped by `splitlines`. My thinking was that this could result in inaccurate diffs.

src/appengine/libs/helpers.py
tbantikyan: Good callout. I'll change it to empty string.

src/appengine/libs/helpers.py
letitz: I don't follow, but it really does not matter :)

tbantikyan:
> Could you add unit tests for this change?

Thank you for the suggestion. Added them in: [a73fc33](https://github.com/google/clusterfuzz/pull/5184/commits/a73fc33a9d78212353ab05f2e1463ae27a36fb1b)

PR #5185, opened 2026-03-02
Adds Feature Flag for Swarming tasks

IvanBM18:
Adds a Feature flag so that we are able to control wether tasks in swarming will be able to be triggered or not trough an external value.

### DIFF
--- a/src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
+++ b/src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
@@ -33,10 +33,12 @@ class SwarmingTest(unittest.TestCase):
     helpers.patch(self, [
         'clusterfuzz._internal.base.utils.post_url',
         'clusterfuzz._internal.swarming._get_task_name',
+        'clusterfuzz._internal.swarming._requires_gpu',
         'clusterfuzz._internal.base.feature_flags.FeatureFlags'
     ])
     self.mock._get_task_name.return_value = 'task_name'  # pylint: disable=protected-access
     self.mock.FeatureFlags.SWARMING_TASKS.enabled = True
+    self.mock._requires_gpu.return_value = True  # pylint: disable=protected-access
     self.maxDiff = None
 
   def test_is_swarming_task_true(self):
@@ -48,13 +50,11 @@ class SwarmingTest(unittest.TestCase):
     job.put()
     self.assertTrue(swarming.is_swarming_task('libfuzzer_chrome_asan'))
 
-  def test_is_swarming_task_false_not_swarming_job(self):
-    """Tests that is_swarming_task returns False if IS_SWARMING_JOB is not True."""
-    job = data_types.Job(
-        name='libfuzzer_chrome_asan',
-        platform='LINUX',
-        environment_string='IS_SWARMING_JOB = False')
+  def test_is_swarming_task_false_not_requires_gpu(self):
+    """Tests that is_swarming_task returns False if REQUIRES_GPU is False."""
+    job = data_types.Job(name='libfuzzer_chrome_asan', platform='LINUX')
     job.put()
+    self.mock._requires_gpu.return_value = False  # pylint: disable=protected-access
     self.assertFalse(swarming.is_swarming_task('libfuzzer_chrome_asan'))
 
   def test_is_swarming_task_false_no_mapping(self):


src/clusterfuzz/_internal/base/feature_flags.py
```diff
@@ -35,6 +35,8 @@ class FeatureFlags(Enum):
 
   PREPROCESS_QUEUE_SIZE_LIMIT = 'preprocess_queue_size_limit'
 
+  SWARMING_TASKS = 'swarming_tasks'
```
IvanBM18, line 38: We want this feature flag to be disabled(to be enabled in the future), Is it needed to have the feature flag also in the datastore so that `FeatureFlag.SWARMING_TASKS.enabled` returns false? @javanlacerda

javanlacerda: Yes, you can check the implementation in the line 49 of this file.

IvanBM18: Done :D 

Image

IvanBM18:
> Could you provide evidence that it's working in dev? As dev is not sync with master, you will probably need to have a commit that combines your implementation with this fix, and then cherry-pick it over dev.

[Added some logs to dev env](https://github.com/google/clusterfuzz/commit/3c4ec59bcafc1993f761ccd5f2f9d59bd8d066b2) And here's a screenshot of some logs of tasks in dev that show the feature flag as false:
image

src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
```diff
@@ -33,8 +33,10 @@ def setUp(self):
     helpers.patch(self, [
         'clusterfuzz._internal.base.utils.post_url',
         'clusterfuzz._internal.swarming._get_task_name',
+        'clusterfuzz._internal.base.feature_flags.FeatureFlags'
     ])
     self.mock._get_task_name.return_value = 'task_name'  # pylint: disable=protected-access
+    self.mock.FeatureFlags.SWARMING_TASKS.enabled = True
```
letitz, line 40: We should have a test for `create_new_task_request()` when the feature flag is false.

src/clusterfuzz/_internal/swarming/__init__.py
javanlacerda:
```suggestion
  if job is None or not feature_flags.FeatureFlags.SWARMING_TASKS.enabled:
```
The error we had in prod was when this `is_swarming_task` is called and the swarming environment is not set. You should check the feature flag here as well to avoid a blast radious error, as the `is_swarming_task` is being called for every task.

javanlacerda (changes requested):
You need to test the entire implementation that regards swarming tasks with feature flag in dev and then provide evidence that even with the new code, we don't have any new error group.

src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
```diff
@@ -33,8 +33,10 @@ def setUp(self):
     helpers.patch(self, [
         'clusterfuzz._internal.base.utils.post_url',
         'clusterfuzz._internal.swarming._get_task_name',
+        'clusterfuzz._internal.base.feature_flags.FeatureFlags'
```
jardondiego, line 37:
```suggestion
        'clusterfuzz._internal.base.feature_flags.FeatureFlagsss'
```

jardondiego: This is a test. Please disregard.

src/clusterfuzz/_internal/swarming/__init__.py
IvanBM18: Oh i thought the issue was only in the other method(since that's what the log shown), nevertheless i just added a feature flag validation to this method too, thanks for pointing it out :D

src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
IvanBM18: Sure i added that one and another additional for the changes javan requested, thanks!

IvanBM18:
Closing this one since we are doing a full rollback on the swarming changes ill send it back again later when we are sure the changes are healthy and ready

jardondiego:
> Closing this one since we are doing a full rollback on the swarming changes ill send it back again later when we are sure the changes are healthy and ready.

I think it would make sense to keep this to help protect future releases. I think we should re-open it.

PR #5190, opened 2026-03-05
Swarming: Creates Feature Flag to control its usage

IvanBM18:
Continuing the swarming efforts, we add a new Feature Flag to control wether swarming tasks are allowed or not to be scheduled. so for this:
- Add a new filter in the `is_swarming_task` to check for the feature flag
- Implemented a UT to verify that when a feature flag is not found in the DB we return false or empty values

-----------------------------------------
## `dev` environment test proof
This feature flag has been present in the `dev` env  DB [since last monday](https://github.com/google/clusterfuzz/commit/bb86465d75d16e79662859f6e65c791283246027), but this changes were present in `dev` under the previous swarming state (see this [PR](https://github.com/google/clusterfuzz/pull/5185)), still the metrics seemed healthy from monday to wednesday , worth nothing that it has been set on false:

Image

So today i send them back again to `dev` to test this changes after the rollback and metrics continue to be healthy and no regressions seem to be introduced cause of this error. The only noticable error i saw in the logs is:

image
Which seems unrelated.

### DIFF
--- a/src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
+++ b/src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
@@ -32,9 +32,11 @@ class SwarmingTest(unittest.TestCase):
   def setUp(self):
     helpers.patch(self, [
         'clusterfuzz._internal.base.utils.post_url',
-        'clusterfuzz._internal.swarming._get_task_name'
+        'clusterfuzz._internal.swarming._get_task_name',
+        'clusterfuzz._internal.swarming.FeatureFlags',
     ])
     self.mock._get_task_name.return_value = 'task_name'  # pylint: disable=protected-access
+    self.mock.FeatureFlags.SWARMING_REMOTE_EXECUTION.enabled = True
     self.maxDiff = None
 
   def test_get_spec_from_config_with_docker_image(self):


src/clusterfuzz/_internal/swarming/__init__.py
```diff
@@ -33,10 +34,10 @@ def _requires_gpu() -> bool:
   return bool(utils.string_is_true(requires_gpu))
 
 
-def is_swarming_task(command: str, job_name: str):
+def is_swarming_task(command: str, job_name: str) -> bool:
   """Returns True if the task is supposed to run on swarming."""
   job = data_types.Job.query(data_types.Job.name == job_name).get()
-  if not job or not _requires_gpu():
+  if not job or not _requires_gpu() or not FeatureFlags.SWARMING_TASKS.enabled:
```
letitz: The feature flag check should occur first, before we query the db or inspect anything else. The point is that we want to reduce the blast radius of swarming-specific code when swarming is disabled.

IvanBM18: Agree, just moved it to be checked before making a call to the DB, thanks :D

src/clusterfuzz/_internal/base/feature_flags.py
```diff
@@ -35,6 +35,8 @@ class FeatureFlags(Enum):
 
   PREPROCESS_QUEUE_SIZE_LIMIT = 'preprocess_queue_size_limit'
 
+  SWARMING_TASKS = 'swarming_tasks'
```
jardondiego: nit: this might be a clearer name
```suggestion
  SWARMING_REMOTE_EXECUTION = 'swarming_remote_execution'
```

src/clusterfuzz/_internal/swarming/__init__.py
```diff
@@ -33,8 +34,10 @@ def _requires_gpu() -> bool:
   return bool(utils.string_is_true(requires_gpu))
 
 
-def is_swarming_task(command: str, job_name: str):
+def is_swarming_task(command: str, job_name: str) -> bool:
   """Returns True if the task is supposed to run on swarming."""
+  if not not FeatureFlags.SWARMING_TASKS.enabled:
```
jardondiego: I suppose this is a typo

```suggestion
  if not FeatureFlags.SWARMING_TASKS.enabled:
```

src/clusterfuzz/_internal/base/feature_flags.py
IvanBM18: Sure, just renamed them

src/clusterfuzz/_internal/swarming/__init__.py
IvanBM18: Yes, good catch!

decoNR:
> If it were me, I would put this flag in the yaml config at the root instead of in the database. Seems to me that even if we want to Turn It Off Now, we could do a rollout with a new config file / rollback to the previous version, instead of flipping a bit in the database (which leaves no audit trail, requires no review, and makes an instant global config change).

> I defer to ClusterFuzz maintainers however on their method of choice for gating new functionality.

Hi, I think @javanlacerda would be the best person to provide details on the chosen approach.

Your suggestion seems good to me.

src/clusterfuzz/_internal/tests/core/datastore/feature_flag_test.py
```diff
@@ -53,3 +53,11 @@ def test_get_and_set(self):
     self.assertEqual(feature_flags.FeatureFlags.TEST_FLOAT_FLAG.content, 1.23)
     self.assertEqual(feature_flags.FeatureFlags.TEST_FLOAT_FLAG.string_value,
                      'checker')
+
+  def test_not_found_in_db(self):
```
decoNR, line 57: Thanks for adding this!

src/clusterfuzz/_internal/base/feature_flags.py
```diff
@@ -35,6 +35,8 @@ class FeatureFlags(Enum):
 
   PREPROCESS_QUEUE_SIZE_LIMIT = 'preprocess_queue_size_limit'
 
+  SWARMING_REMOTE_EXECUTION = 'swarming_remote_execution'
```
decoNR, line 38: I note that the production database needs an update as it contains the feature_flag `swarming_tasks` instead of `swarming_remote_execution`.

IvanBM18: Sure i added it to the `cluster-fuzz` project.

Image

decoNR: Thanks!

cemon721-a11y:
Controls pull

PR #5191, opened 2026-03-05
Fix check for Swarming task

jardondiego:
Now checks job configuration to determine if it's a Swarming task


jardondiego:
Should we re-open this pull request to now target `dev`? @javanlacerda

cemon721-a11y:
Perfectly moved into unsubscribe from this kind of email

PR #5194, opened 2026-03-06
WIP: Swarming remote task interface service

jardondiego:
Also includes updated logic for Remote Task Gate


src/clusterfuzz/_internal/remote_task/remote_task_gate.py
```diff
@@ -44,6 +45,11 @@ def __init__(self):
     }
     self._adapters = remote_task_adapters.RemoteTaskAdapters
 
+  def _is_swarming_job(self, job_type):
+    """Returns true if the job is a swarming job."""
+    job = data_types.Job.query(data_types.Job.name == job_type).get()
+    return job and job.get_environment().get('IS_SWARMING_JOB')
```
jardondiego, line 51: We should be able to reuse `swarming.is_swarming_task`.

PR #5199, opened 2026-03-09
Swarming: Adds dynamic Job specific bot dimensions

IvanBM18:
In an effort to gradually add swarming integration changes back. This PR adds the ability to set dynamic job dimensions in addition to the static config platform specific dimensions for swarming tasks.

The dimensions will be added trough the job env_string with the following syntax:
- SWARMING_DIMENSIONS={‘cpu’:’x86’,’os’:’windows’}
The Job specific dimensions will have more presedence than that the static config ones, which will allows us to have more flexibility when sending swarming tasks.

Also adds two unit tests to handle the new test cases

Bug: b/483772015

### DIFF
--- a/src/clusterfuzz/_internal/swarming/__init__.py
+++ b/src/clusterfuzz/_internal/swarming/__init__.py
@@ -14,8 +14,10 @@
 """Swarming helpers."""
 
 import base64
+import json
 import uuid
 
+from google.auth.transport import requests
 from google.protobuf import json_format
 
 from clusterfuzz._internal.base import utils
@@ -25,6 +27,11 @@ from clusterfuzz._internal.datastore import data_types
 from clusterfuzz._internal.google_cloud_utils import credentials
 from clusterfuzz._internal.protos import swarming_pb2
 from clusterfuzz._internal.system import environment
+
+_SWARMING_SCOPES = [
+    'https://www.googleapis.com/auth/cloud-platform',
+    'https://www.googleapis.com/auth/userinfo.email'
+]
 
 
 def is_swarming_task(command: str, job_name: str):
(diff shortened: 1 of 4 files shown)


src/clusterfuzz/_internal/swarming/__init__.py
```diff
@@ -55,6 +55,30 @@ def _get_swarming_config():
   return local_config.SwarmingConfig()
 
 
+def _get_task_dimensions(job: data_types.Job, platform_specific_dimensions: list
```
jardondiego, line 65: It's great to have type hints!

src/clusterfuzz/_internal/swarming/__init__.py
```diff
@@ -55,6 +55,30 @@ def _get_swarming_config():
...
+  Job dimensions have more precedence than static dimensions"""
+  unique_dimensions = {}
+  unique_dimensions['os'] = job.platform
+  unique_dimensions['pool'] = _get_swarming_config().get('swarming_pool')
+
+  for dimension in platform_specific_dimensions:
+    unique_dimensions[dimension['key'].lower()] = dimension['value']
+
+  swarming_dimensions = environment.get_value('SWARMING_DIMENSIONS')
+  if isinstance(swarming_dimensions, dict):
```
jardondiego, line 77: qq: Where does the parsing of such a variable take place? Is it a string or a `dict`?

IvanBM18: The get_value function doesn't necessarily return a string, the value you see comes from:
[this function](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/system/environment.py#L67) Which basically casts a string to numbers, tuples, lists, dicts, etc.

I personally tested this by creating small script to see how this worked for dicts:
```python
import ast

def _eval_value(value_string):
    """Returns evaluated value."""
    try:
        return ast.literal_eval(value_string)
    except Exception as e:
        # String fallback.
        print(f"Failed to eval: {e}")
        return value_string

def test_eval():
    string_dict = '{"key1": "value1", "key2": "value2"}'
    print(f"Original string: {string_dict} (type: {type(string_dict)})")
    
    evaluated = _eval_value(string_dict)
    print(f"Evaluated value: {evaluated} (type: {type(evaluated)})")
    
    if isinstance(evaluated, dict):
        print("Iterating through .items():")
        for k, v in evaluated.items():
            print(f"  {k}: {v}")

if __name__ == "__main__":
    test_eval()
```

And it worked perfectly:
```
Original string: {"key1": "value1", "key2": "value2"} (type: <class 'str'>)                                                                                                               
Evaluated value: {'key1': 'value1', 'key2': 'value2'} (type: <class 'dict'>)                                                                                                              
Iterating through .items():                                                                                                                                                                              key1: value1                                                                                                                                                                           
key2: value2  
```

javanlacerda:
There are some conflicts.

PR #5200, opened 2026-03-09
Enable manual job configuration setting to select OS for Swarming task execution

jardondiego:
By default, os selection for swarming tasks should be the `job.platform` field, but we should also be able to override in case it's needed since not all possible values of `job.platform` map to an OS field in swarming.

After testing the change in the development environment, there are spikes in errors and overall reliability metrics look fine as well.
Dashboard http://shortn/_dqBN6VW9hf
Logs http://shortn/_2CAdQoldae

### DIFF
--- a/src/clusterfuzz/_internal/swarming/__init__.py
+++ b/src/clusterfuzz/_internal/swarming/__init__.py
@@ -69,7 +69,7 @@ def _get_task_dimensions(job: data_types.Job, platform_specific_dimensions: list
   unique_dimensions = {}
   unique_dimensions['os'] = job.platform
   unique_dimensions['pool'] = _get_swarming_config().get('swarming_pool')
-  
+
   os_task_dimension = job.get_environment().get('SWARMING_OS_DIMENSION')
   if os_task_dimension:
     unique_dimensions['os'] = os_task_dimension
--- a/src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
+++ b/src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
@@ -97,21 +97,6 @@ class SwarmingTest(unittest.TestCase):
 
     self.assertEqual(spec, expected_spec)
 
-  def test_get_spec_with_swarming_os_dimension(self):
-    """Tests that _get_new_task_spec uses SWARMING_OS_DIMENSION if present."""
-    job = data_types.Job(
-        name='libfuzzer_chrome_asan',
-        platform='LINUX',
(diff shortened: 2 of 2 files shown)


src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
```diff
@@ -97,6 +97,21 @@ def test_get_spec_from_config_with_docker_image(self):
...
+        environment_string='SWARMING_OS_DIMENSION = custom_os')
+    job.put()
+    spec = swarming._get_new_task_spec(  # pylint: disable=protected-access
+        'corpus_pruning', job.name, 'https://download_url')
+
+    # Verify that the 'os' dimension is 'custom_os' instead of 'LINUX'.
+    dimensions = spec.task_slices[0].properties.dimensions
+    os_dimension = [d.value for d in dimensions if d.key == 'os'][0]
+    self.assertEqual(os_dimension, 'custom_os')
+
```
jardondiego:
```suggestion
```

jardondiego:
After testing the change in the development environment, there are spikes in errors and overall reliability metrics look fine as well.
Dashboard http://shortn/_dqBN6VW9hf
Logs http://shortn/_2CAdQoldae

src/clusterfuzz/_internal/swarming/__init__.py
```diff
@@ -70,6 +70,10 @@ def _get_task_dimensions(job: data_types.Job, platform_specific_dimensions: list
   unique_dimensions['os'] = job.platform
   unique_dimensions['pool'] = _get_swarming_config().get('swarming_pool')
 
+  os_task_dimension = job.get_environment().get('SWARMING_OS_DIMENSION')
```
IvanBM18, line 73: Is this change required? currently you can specify the os to use in swarming from either:
- the job platform
- the  SWARMING_DIMENSIONS env var
  - e.g. SWARMING_DIMENSIONS={'os':'Linux'}

Where the OS from the the env var has more precedence than the os from the job. So why would we need to another source for the specific OS selection?

jardondiego: Just a note, this change here was conceived before the PR with the dynamic dimensions was merged with the assumption that it being a simple change it could go through more quickly. Thanks for raising this. I do see that this functionality is covered with that. I did not notice that PR went through. In upcoming efforts, let's make sure we're aligned on topics like this to avoid cannibalizing each other's work.

PR #5243, opened 2026-04-14
Protect testcase task-log downloads

M0nd0R:
This PR is a clean replacement for #5214 with only the task-log fix and its focused tests.

It adds testcase access checks before returning task logs and quotes task log filter values before building the Cloud Logging query so user-controlled task parameters cannot alter the intended filter constraints.

Verification:
- targeted smoke check: TaskLogHandler now calls access.check_access_and_get_testcase(123) before fetching logs
- targeted smoke check: injected filter content remains inside the quoted task_id string literal
- local pytest selection reaches the new tests, but full execution is blocked here because this environment does not have gcloud for the datastore emulator

### DIFF
--- a/src/appengine/handlers/testcase_detail/show.py
+++ b/src/appengine/handlers/testcase_detail/show.py
@@ -640,8 +640,9 @@ class TaskLogHandler(base_handler.Handler):
   @handler.get(handler.TEXT)
   def get(self):
     """Serve the task log."""
-    testcase_id = helpers.cast(flask.request.args.get('testcase_id'), int,
-                               "The param 'testcase_id' is not a number.")
+    testcase_id = helpers.cast(
+        flask.request.args.get('testcase_id'), int,
+        "The param 'testcase_id' is not a number.")
     access.check_access_and_get_testcase(testcase_id)
 
     task_id = flask.request.args.get('task_id')
--- a/src/appengine/handlers/testcase_detail/show.py
+++ b/src/appengine/handlers/testcase_detail/show.py
@@ -640,9 +640,18 @@ class TaskLogHandler(base_handler.Handler):
   @handler.get(handler.TEXT)
   def get(self):
     """Serve the task log."""
+    testcase_id = helpers.cast(flask.request.args.get('testcase_id'), int,
+                               "The param 'testcase_id' is not a number.")
+    access.check_access_and_get_testcase(testcase_id)
+
     task_id = flask.request.args.get('task_id')
+    if not task_id:
+      raise helpers.EarlyExitError('No task ID provided.', 400)
+
     task_name = flask.request.args.get('task_name')
-    testcase_id = flask.request.args.get('testcase_id')
+    if not task_name:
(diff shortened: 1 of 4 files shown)


src/appengine/handlers/testcase_detail/show.py
```diff
@@ -640,9 +640,18 @@ class TaskLogHandler(base_handler.Handler):
   @handler.get(handler.TEXT)
   def get(self):
     """Serve the task log."""
+    testcase_id = helpers.cast(flask.request.args.get('testcase_id'), int,
```
decoNR: This cast is unnecessary, as `data_handler.get_testcase_by_id` in `check_access_and_get_testcase` already handles it. Please remove it.

decoNR (changes requested):

decoNR:
Please fix lint errors (ref: [Run basic tests / build (pull_request)](https://github.com/google/clusterfuzz/actions/runs/24408546095/job/71299435220?pr=5243)).

src/appengine/handlers/testcase_detail/show.py
M0nd0R: Removed the explicit cast and now rely on `check_access_and_get_testcase`/`data_handler.get_testcase_by_id` for testcase ID validation. The handler test was updated to assert that raw `testcase_id` is delegated there and that the resolved testcase key ID is used for the log lookup.

M0nd0R:
Addressed the review about the unnecessary cast. The handler now passes the raw `testcase_id` string into `check_access_and_get_testcase`, relies on `data_handler.get_testcase_by_id` for validation, and uses the returned testcase key ID for the log lookup. I also updated `TaskLogHandlerTest` to cover that flow explicitly.

src/clusterfuzz/_internal/tests/appengine/handlers/testcase_detail/show_test.py
```diff
@@ -18,13 +18,23 @@
...
+import webtest
+
 from clusterfuzz._internal.datastore import data_types
 from clusterfuzz._internal.tests.test_libs import helpers as test_helpers
 from clusterfuzz._internal.tests.test_libs import test_utils
 from handlers.testcase_detail import show
 from libs import helpers
 
 
+def assert_dict_contains_subset(test_case, expected_subset, actual):
```
decoNR: Please remove this function as this is not related to the tests you added.

decoNR (changes requested):

src/clusterfuzz/_internal/tests/appengine/handlers/testcase_detail/show_test.py
M0nd0R: Removed the unrelated helper and restored the existing `assertDictContainsSubset` calls. The remaining diff in `show_test.py` is limited to the `TaskLogHandlerTest` coverage for the task-log access change.

decoNR (approved):
LGTM, thanks! We will probably deploy it in the next release

ViniciustCosta (approved):
lgtm

PR #5244, opened 2026-04-14
[BlackboxAudit] Add butler script to download a fuzzer config

dylanjew:
Adds script to download a JSON config with the parameters required to upload a fuzzer.

Along with #5236 allows us to edit configs without going through the UI.

We want to store these configs in VCS so that we have a history of the jobs that a fuzzer is enabled for

### Testing
I ran this with my local server and the  prod config with `python butler.py  run --non-dry-run -c ../clusterfuzz-config/configs/internal download_fuzzer_config --script_args ifratric-browserfuzzer-v3`
and
`python butler.py  run --non-dry-run --local -c configs/test download_fuzzer_config --script_args dylanj_test_fuzzer`

Then uploaded a test fuzzer locally with `python butler.py  run --non-dry-run -c configs/test/ --local upload_fuzzer --script_args --fuzzer-name=dylanj_test_fuzzer --fuzzer-archive-path=../dylanj_b0ring_webidl_fuzzer.zip --config-path=test_config.json`

### DIFF
--- a/src/clusterfuzz/_internal/datastore/data_types.py
+++ b/src/clusterfuzz/_internal/datastore/data_types.py
@@ -351,6 +351,20 @@ class Fuzzer(Model):
   # on disk |FUZZ_INPUTS_DISK|, rather than smaller tmpfs one (FUZZ_INPUTS).
   has_large_testcases = ndb.BooleanProperty(default=False)
 
+  def get_config_dict(self):
+    """Returns a dict containing the required config to upload a fuzzer."""
+
+    return {
+        'jobs': self.jobs,
+        'data_bundle_name': self.data_bundle_name,
+        'timeout': self.timeout,
+        'max_testcases': self.max_testcases,
+        'external_contribution': self.external_contribution,
+        'additional_environment_string': self.additional_environment_string,
+        'executable_path': self.executable_path,
+        'launcher_script': self.launcher_script,
+    }
+
 
 class BuildCrashStatsJobHistory(Model):
   """Represents the record of build_crash_stats run."""
--- a/src/clusterfuzz/_internal/datastore/data_types.py
+++ b/src/clusterfuzz/_internal/datastore/data_types.py
@@ -261,6 +261,27 @@ class Blacklist(Model):
 class Fuzzer(Model):
   """Represents a fuzzer."""
 
+  _EXCLUDED_FIELDS_FOR_CONFIG = (
+      'created_at',
+      'timestamp',
+      'name',
+      'filename',
+      'blobstore_key',
+      'file_size',
+      'supported_platforms',
+      'result',
+      'result_timestamp',
+      'console_output',
+      'return_code',
+      'sample_testcase',
+      'untrusted_content',
+      'stats_columns',
+      'stats_column_descriptions',
+      'builtin',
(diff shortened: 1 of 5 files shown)


aakallam (approved):

src/clusterfuzz/_internal/datastore/data_types.py
```diff
@@ -351,6 +351,20 @@ class Fuzzer(Model):
   # on disk |FUZZ_INPUTS_DISK|, rather than smaller tmpfs one (FUZZ_INPUTS).
   has_large_testcases = ndb.BooleanProperty(default=False)
 
+  def get_config_dict(self):
```
ViniciustCosta, line 354: Could you maybe use `to_dict()` and filter out the unneeded fields? Similar to what is done here: https://github.com/google/clusterfuzz/blob/master/src/appengine/handlers/fuzzers.py#L147

Maybe you can even centralize this as the same function, if fields are the same.

dylanjew: Initially I didn't want to prematurely DRY anything since this was for a local script. But good point, so I took a deeper look.

My current thinking is that if we add a new field, it's more clear to also explicitly add it to the `config.json` that we will store in version control.

Unless the field is also added to the [edit fuzzer request](https://github.com/google/clusterfuzz/blob/c7308075a573ef1157abc223ed5e079e64396e6d/src/appengine/handlers/fuzzers.py#L173-L179), it doesn't *need* to be part of the config.

My initial worry was that adding a new field and forgetting to add it to an `EXCLUDED_FIELDS_FOR_CONFIG` would break the upload script. But I just verified that additional fields are ignored in the request, so it's safe to add anything else we want in there. 

Maybe some of those non required fields would also be helpful to also track in the config, like `source`, which changes after edits.

I think the ideal state would be to have an `Edit`/`CreateFuzzerRequest` object and this config would be the serialized version of that. Then you'd just write a converter between the Fuzzer and that request object. But that seems like a larger change.

What are your thoughts? I'm fine with using `to_dict` to make it implicit if you'd prefer. I suspect the list would include a few more fields on top of the ones in `FUZZER_FIELDS_EXCLUDED_FROM_LOG`

ViniciustCosta: I didn't quite follow everything you said, since I'm not very familiar with this fuzzer request/config part.  I do think that it should be the same effort (or even smaller) to add a new field to an explicit list (`FUZZER_CONFIG_FIELDS`) than to add it to the mapping you created.

> new field and forgetting to add it to an EXCLUDED_FIELDS_FOR_CONFIG would break...

I see that it wouldn't break in this case, but a suggestion for cases like this is trying to add unit tests that fails if different parts of the code that should be in sync are not updated together.

I will approve anyway, feel free to decide if this change is worth it :)

ViniciustCosta (approved):

src/clusterfuzz/_internal/datastore/data_types.py
dylanjew: Sure, that makes sense. I switched to an exclusion list and added a test.

ViniciustCosta (approved):

PR #5245, opened 2026-04-15
Swarming: Patches Auth default credentials at startup

IvanBM18:
Fuzzing tasks in swarming have an intermittent error in which startup of any gcp library or module fails with:
```
  File "/mnt/scratch0/clusterfuzz/src/clusterfuzz/_internal/datastore/ndb_init.py", line 39, in _client
    _ndb_client = ndb.Client(project=utils.get_application_id())
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/scratch0/clusterfuzz/src/third_party/google/cloud/ndb/client.py", line 145, in __init__
    super(Client, self).__init__(
  File "/mnt/scratch0/clusterfuzz/src/third_party/google/cloud/client/__init__.py", line 320, in __init__
    _ClientProjectMixin.__init__(self, project=project, credentials=credentials)
  File "/mnt/scratch0/clusterfuzz/src/third_party/google/cloud/client/__init__.py", line 268, in __init__
    project = self._determine_default(project)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/scratch0/clusterfuzz/src/third_party/google/cloud/ndb/client.py", line 249, in _determine_default
    return _determine_default_project(project)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/scratch0/clusterfuzz/src/third_party/google/cloud/ndb/client.py", line 70, in _determine_default_project
    project = _helpers._determine_default_project(project=project)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/scratch0/clusterfuzz/src/third_party/google/cloud/_helpers/__init__.py", line 152, in _determine_default_project
    _, project = google.auth.default()
                 ^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/scratch0/clusterfuzz/src/third_party/google/auth/_default.py", line 692, in default
    raise exceptions.DefaultCredentialsError(_CLOUD_SDK_MISSING_CREDENTIALS)
google.auth.exceptions.DefaultCredentialsError: Your default credentials were not found. To set up Application Default Credentials, see https://cloud.google.com/docs/authentication/external/set-up-adc for more information.
```
This happens because the timeout window for looking for the default credentials(trough the `google.auth.credentials.default()` method) is very short, this timeout window cannot be increased and since we are using a `luci-auth` server the requests tend to take a little bit more time than the server in an actual GCE bot.

To fix this, we are manually patching the `google.auth.default` module to return credentials that do come from the swarming auth & that doesn't time out.

## Tests performed
This changes have been present in the `dev` env since April 15th, which allowed us to perform some fuzzing sessions in swarming like [this one](https://chrome-swarming.appspot.com/task?d=true&id=77c28b2dbab10510)


src/clusterfuzz/_internal/base/modules.py
```diff
@@ -44,6 +44,50 @@ def _patch_appengine_modules_for_bots():
...
+    return
+
+  try:
+    import google.auth
+    from google.auth import compute_engine
+    import requests
+
+    def patched_default_credentials(
+        *args, **kwargs) -> tuple[compute_engine.Credentials, str] | None:
+      # pylint: disable=unused-argument
```
IvanBM18, line 65: Note: We have the args and kwargs in the method signature because otherwise if we call the default() method with any args an exception would be thrown because the patch didn't matched the expected signature.

src/clusterfuzz/_internal/base/modules.py
```diff
@@ -44,6 +44,50 @@ def _patch_appengine_modules_for_bots():
...
+        try:
+          project_id = requests.get(
+              url, headers={
+                  "Metadata-Flavor": "Google"
+              }, timeout=2).text.strip()
+        except Exception:
+          pass
+
+      if not project_id:
+        print('''[Swarming] [Error] Failed to patch google.auth.default.
```
IvanBM18, line 80: This will be logged into swarming only, not GCP, since when this executes we still haven't setup GCP logging

letitz:
I'm confused and need some more details. I was assuming that `luci-auth` would set up some credentials on disk somewhere, that GCP libs would then pick up. You're saying that it runs a local server that GCP libs talk to instead?

In general, do you have a bug link for this that we can discuss more on?

IvanBM18:
> I'm confused and need some more details. I was assuming that `luci-auth` would set up some credentials on disk somewhere, that GCP libs would then pick up. You're saying that it runs a local server that GCP libs talk to instead?
> 
> In general, do you have a bug link for this that we can discuss more on?

Sure heres the bug im using:
b/502273525

Regarding `luci-auth`....
This doc explains in more detail how service account and auth works in a swarming task:
[go/swarming-service-accounts](http://go/swarming-service-accounts)
TL;DR
It setups a "mock" metadata server in the swarming host(thats why we need to pass a bunch of env vars and setup the `--network=host` at the container) said server then uses the callers account(that being me or the appengine account) to impersonate the service account, and then it generates short lived Oauth tokens.

You can also think of it like its a kind of a proxy running in your swarming bot.

src/clusterfuzz/_internal/base/modules.py
jardondiego: Changing the function signature to this should fix it without having this pylint directive.
```
def patched_default_credentials(*_args, **_kwargs) -> tuple[compute_engine.Credentials, str] | None:
```

IvanBM18: Done!

Thanks :D

src/clusterfuzz/_internal/base/modules.py
```diff
@@ -44,6 +44,49 @@ def _patch_appengine_modules_for_bots():
     pass
 
 
+def _patch_google_auth_for_bots() -> None:
```
jardondiego:
```suggestion
def _patch_google_auth_for_swarming_bots() -> None:
```

src/clusterfuzz/_internal/base/modules.py
```diff
@@ -75,3 +118,4 @@ def fix_module_search_paths():
 
   # TODO(ochang): Remove this once SDK is removed from images.
   _patch_appengine_modules_for_bots()
+  _patch_google_auth_for_bots()
```
jardondiego:
```suggestion
  _patch_google_auth_for_swarming_bots()
```

jardondiego (approved):

letitz:
Got it, thanks for the reply and sorry for my own delayed response.

I'm still super wary of patching production code from a third-party library - that seems like an absolute last resort. The `google.auth` library docs [1] lists extensive ways of configuring `google.auth` so I'm surprised we need to resort to monkey patching instead.

If I'm reading the stack trace correctly, the issue arises when `google.auth` libs try to determine the cloud project ID. Let's just specify that using `GOOGLE_CLOUD_PROJECT` and be done with it? From the docs again [2]:

> You can override the project detection by setting the GOOGLE_CLOUD_PROJECT environment variable.

[1]: https://googleapis.dev/python/google-auth/latest/user-guide.html
[2]: https://googleapis.dev/python/google-auth/latest/user-guide.html#using-external-identities

IvanBM18:
> Got it, thanks for the reply and sorry for my own delayed response.
> 
> I'm still super wary of patching production code from a third-party library - that seems like an absolute last resort. The `google.auth` library docs [1](https://googleapis.dev/python/google-auth/latest/user-guide.html) lists extensive ways of configuring `google.auth` so I'm surprised we need to resort to monkey patching instead.
> 
> If I'm reading the stack trace correctly, the issue arises when `google.auth` libs try to determine the cloud project ID. Let's just specify that using `GOOGLE_CLOUD_PROJECT` and be done with it? From the docs again [2](https://googleapis.dev/python/google-auth/latest/user-guide.html#using-external-identities):
> 
> > You can override the project detection by setting the GOOGLE_CLOUD_PROJECT environment variable.

Although passing down the GOOGLE_CLOUD_PROJECT would solve one problem, it uncovers another one:
```
  File "/mnt/scratch0/clusterfuzz/src/third_party/google/cloud/ndb/client.py", line 145, in __init__
    super(Client, self).__init__(
  File "/mnt/scratch0/clusterfuzz/src/third_party/google/cloud/client/__init__.py", line 321, in __init__
    Client.__init__(
  File "/mnt/scratch0/clusterfuzz/src/third_party/google/cloud/client/__init__.py", line 178, in __init__
    credentials, _ = google.auth.default(scopes=scopes)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/scratch0/clusterfuzz/src/third_party/google/auth/_default.py", line 692, in default
    raise exceptions.DefaultCredentialsError(_CLOUD_SDK_MISSING_CREDENTIALS)
google.auth.exceptions.DefaultCredentialsError: Your default credentials were not found. To set up Application Default Credentials, see https://cloud.google.com/docs/authentication/external/set-up-adc for more information.
  File "/b/s/w/ir/./setup_docker_env.py", line 103, in <module>
```
[Full logs](https://chrome-swarming.appspot.com/task?d=true&id=77eb52876891b510)
Im investigating now the google_auth code for more details hoping to find another solution

letitz:
Ok thanks! Let's discuss on b/502273525.

IvanBM18:
Closing this PR.

After further discussions and investigations(at b/506127465) we found a cleaner solution to avoid patching this method. 

Thanks!

PR #5246, opened 2026-04-15
Swarming: Avoids mounting directories at startup

IvanBM18:
We previously worked in avoiding mounting some directories into the docker container from swarming, [see this pr](https://github.com/google/clusterfuzz/pull/5213)

After fixing other issues related to auth, we discovered that later in the bot execution, other startup scripts [try to remount](https://github.com/google/clusterfuzz/blob/master/docker/base/setup_clusterfuzz.sh) some more directories:
```
if [[ -z "$DISABLE_MOUNTS" ]]; then
  # Setup Tmpfs dirs for frequently accessed files to save disk I/O.
  .....
```
This can't be done in swarming, so this change aims to avoid said mount at startup by adding the env variable as one of the default required env vars for a swarming task request.

Note: @javanlacerda is the e2e github CI check still failing by default?


jardondiego (approved):
LGTM

javanlacerda (approved):

PR #5247, opened 2026-04-17
Custom logger for Swarming & Removes circular dependency at logs.py

IvanBM18:
Currently swarming bots use the K8s default logging configurations(which work fine :D), but when we try to find logs specific to a given swarming bot host or a swarming task, things start to get complicated, its hard to pinpoint the execution workflow.

In this PR we changed the configuration so the logger can displays different information if its running on swarming. To achieve this:
- Added a logger config if running in swarming and a new method in the `environment` module
- Removed a circular dependency between `environment` and `logs` module
  - Causing a method to be moved from one module to the other
- Simplified a bunch of conditionals in the `logs` module

Note: This PR is complimentary to this [other pr](https://clusterfuzz-config-472119376969.us-central1.sourcemanager.dev/clusterfuzz-testing/clusterfuzz-config/pulls/291)

## Tests performed
I launched some fuzzing jobs that executed in swarming, thanks to the new fields i was able to easily find the logs for a given bot or a given task. [See example](https://cloudlogging.app.goo.gl/kNdC9cXeGEmF5hdRA)


src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -99,9 +99,10 @@ def _file_logging_enabled():
   This is disabled if we are running in app engine or kubernetes as these have
     their dedicated loggers, see configure_appengine() and configure_k8s().
```
letitz: Comment needs an update.

src/clusterfuzz/_internal/system/environment.py
```diff
@@ -749,9 +749,10 @@ def get_runtime() -> UtaskMainRuntime:
   return UtaskMainRuntime.INSTANCE_GROUP
 
 
-def is_swarming_bot():
+def is_running_on_swarming():
   """Return whether or not the current bot is a swarming bot."""
-  return get_value('SWARMING_BOT')
+  value = get_value('SWARMING_BOT')
+  return value is True or value == 'true'
```
letitz: Can we fix the environment variable wherever we pass it instead of adding this workaround for lowercase?

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -554,12 +557,42 @@ def cloud_label_filter(record):
...
+  sys.excepthook = uncaught_exception_handler
+
+
 def configure(name, extras=None):
   """Set logger. See the list of loggers in bot/config/logging.yaml.
   Also configures the process to log any uncaught exceptions as an error.
   |extras| will be included by emit() in log messages."""
   suppress_unwanted_warnings()
 
+  from clusterfuzz._internal.system import environment
```
letitz: I see this import has to live here to avoid a circular dependency. I think we can fix this.

It appears that `environment.py` only uses `logs.info()`, `logs.warning()` and `logs.error()`, and then only in `update_task_enabled()`. If we extract that function to `run_bot.py` (the only place it's called), then we can remove the dependency from `environment` on `logs`.

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -554,12 +557,42 @@ def cloud_label_filter(record):
   logging.getLogger().addHandler(handler)
 
 
+def configure_swarming(name, extras=None):
+  """Configure logging for swarming bots."""
+  if extras is None:
+    extras = {}
```
letitz, line 554: You can instead take `extras={}` as a function argument. If someone passes `None` explicitly then it will explode, but we can annotate the type as `dict[str, str]` to discourage that.

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -110,9 +111,11 @@ def _cloud_logging_enabled():
   This is disabled for local development and if we are running in a app engine
     or kubernetes as these have their dedicated loggers, see
     configure_appengine() and configure_k8s()."""
+  from clusterfuzz._internal.system import environment
   return (bool(os.getenv('LOG_TO_GCP', 'True')) and
```
letitz: A note in passing: `bool(os.getenv('LOG_TO_GCP', 'True'))` will return True if `LOG_TO_GCP = False`, because any non-empty string is truthy, which is very counterintuitive in this setting. It's worth a comment here and in `_file_logging_enabled()` while we're at it.

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -110,9 +111,11 @@ def _cloud_logging_enabled():
...
           not os.getenv("PY_UNITTESTS") and not _is_local() and
-          not _is_running_on_app_engine() and not _is_running_on_k8s())
+          not _is_running_on_app_engine() and
+          (not _is_running_on_k8s() or environment.is_running_on_swarming()))
```
letitz: I don't think we need any of these checks for appengine, k8s and swarming here.

AFAICT the only place this is called before your PR is in `configure()` down on line 604. At that point, we have already checked if we're running on app engine, k8s and swarming, and the answer is no - otherwise we would have diverted execution into `_configure_k8s()` etc. These two lines then must evalute to True.

You've added a new call for this in `_configure_swarming()`, but similarly by that point we know that we are not on app engine, not on k8s because we are on swarming - so these two lines evaluate to True.

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -99,9 +99,10 @@ def _file_logging_enabled():
   This is disabled if we are running in app engine or kubernetes as these have
     their dedicated loggers, see configure_appengine() and configure_k8s().
   """
+  from clusterfuzz._internal.system import environment
   return bool(os.getenv(
-      'LOG_TO_FILE',
-      'True')) and not _is_running_on_app_engine() and not _is_running_on_k8s()
+      'LOG_TO_FILE', 'True')) and not _is_running_on_app_engine() and (
+          not _is_running_on_k8s() or environment.is_running_on_swarming())
```
letitz: Same comment as below in `_cloud_logging_enabled()`: I think we only need to check `LOG_TO_FILE` because callers have already checked which platform we're running on.

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -99,9 +99,10 @@ def _file_logging_enabled():
   This is disabled if we are running in app engine or kubernetes as these have
     their dedicated loggers, see configure_appengine() and configure_k8s().
   """
+  from clusterfuzz._internal.system import environment
   return bool(os.getenv(
```
letitz: Same comment as below re: `bool(os.getenv())` returning True in surprising cases.

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -554,12 +557,42 @@ def cloud_label_filter(record):
...
+  if extras is None:
+    extras = {}
+  extras['task_id'] = os.getenv('TASK_ID')
+  extras['instance_id'] = os.getenv('BOT_NAME')
+  extras['platform'] = 'swarming'
+
+  global _default_extras
+  _default_extras = extras
+
+  if _console_logging_enabled():
```
letitz: I don't think we terribly need to support logging to console or files in swarming, especially since swarming logs are not very locked down. Let's just support cloud logging.

src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/utasks_test.py
```diff
@@ -208,12 +208,13 @@ def setUp(self):
...
     """Tests that uworker_main works as intended."""
+    # pylint: disable=protected-access
     start_time_ns = time.time_ns()
 
     if execution_mode == utasks.Mode.SWARMING:
-      self.mock.is_swarming_bot.return_value = True  # pylint: disable=protected-access
+      self.mock.is_running_on_swarming.return_value = True
     else:
-      self.mock.is_swarming_bot.return_value = False
+      self.mock.is_running_on_swarming.return_value = False
```
letitz:
```suggestion
  self.mock.is_running_on_swarming.return_value = (execution_mode == utasks.Mode.SWARMING)
```

src/clusterfuzz/_internal/tests/core/metrics/logs_test.py
```diff
@@ -521,6 +521,85 @@ def test_configure_appengine(self):
...
+    """Test configure for swarming bot."""
+    # pylint: disable=protected-access
+    os.environ['SWARMING_BOT'] = 'true'
+    os.environ['TASK_ID'] = 'task-123'
+    os.environ['BOT_NAME'] = 'bot-123'
+
+    helpers.patch(
+        self,
+        ['clusterfuzz._internal.system.environment.is_running_on_swarming'])
+    self.mock.is_running_on_swarming.return_value = True
```
letitz: I don't think we need to mock this if we're setting `SWARMING_BOT` to `True` in the environment.

src/clusterfuzz/_internal/tests/core/metrics/logs_test.py
```diff
@@ -521,6 +521,85 @@ def test_configure_appengine(self):
     logs.configure('test')
     self.assertEqual(0, self.mock.dictConfig.call_count)
 
+  def test_configure_swarming(self):
+    """Test configure for swarming bot."""
+    # pylint: disable=protected-access
+    os.environ['SWARMING_BOT'] = 'true'
```
letitz:
```suggestion
    os.environ['SWARMING_BOT'] = 'True'
```

letitz (changes requested):

src/clusterfuzz/_internal/metrics/logs.py
IvanBM18: I updated the method signature to add typing, but i don't think setting {} as default would be the best call here.
Using a mutable object(like the dict here) as a default argument would mean that the default dict instance would be shared across all calls to this function, this would mean that modifications from a previous call to configure_swarming() would leak into consequent subsequent calls.
```
# --- Scenario ---
# Call 1: doesn't pass extras
configure_swarming('logger1')
# Output: {'task_id': 'default_task', 'instance_id': 'default_bot', 'platform': 'swarming'}

# Then for some reason we to add a custom flag to the extras global dictionary
_default_extras['custom_debug_flag'] = True


# Call 2: Later, another part of the code calls it, also without passing extras
configure_swarming('logger2')
# Output:  {'task_id': 'default_task', 'instance_id': 'default_bot', 'platform': 'swarming', 'custom_debug_flag': True}
```
I know right now this edge case wouldn't happen but i thinks its best to plan for it in case in the future something changes.

src/clusterfuzz/_internal/system/environment.py
```diff
@@ -1238,44 +1236,3 @@ def can_testcase_run_on_platform(testcase_platform_id, current_platform_id):
 
 def is_tworker():
   return get_value('TWORKER', False)
-
```
IvanBM18, line 1241: Note for reviews:
To remove a circular dependency between `logs.py` and `environment.py` we moved this method into `run_bot.py`, this method required the logs module hence causing a circular dependency. Its safe to remove it from here as this was really only used at `run_bot.py`

src/python/bot/startup/run_bot.py
```diff
@@ -193,6 +195,47 @@ def task_loop():
   return stacktrace, clean_exit, task_payload
 
 
+def update_task_enabled() -> bool:
```
IvanBM18, line 198: See context at context for this change at [this comment](https://github.com/google/clusterfuzz/pull/5247/changes#r3102621854) and this [other comment](https://github.com/google/clusterfuzz/pull/5247#discussion_r3098845369)

src/clusterfuzz/_internal/metrics/logs.py
letitz: Ah you're right, I'd forgotten about that footgun. Thanks for pushing back!

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -93,26 +95,18 @@ def _console_logging_enabled():
...
-  This is enabled by default.
-  This is disabled for local development and if we are running in a app engine
-    or kubernetes as these have their dedicated loggers, see
-    configure_appengine() and configure_k8s()."""
-  return (bool(os.getenv('LOG_TO_GCP', 'True')) and
-          not os.getenv("PY_UNITTESTS") and not _is_local() and
-          not _is_running_on_app_engine() and not _is_running_on_k8s())
+  This is enabled by default but disabled for local development."""
+  return environment.get_value('LOG_TO_GCP',
+                               True) and (not os.getenv('PY_UNITTESTS') and
```
letitz: Optional: I would move this PY_UNITTESTS check to its own helper method in `environment.py`: `is_running_unit_tests()`.

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -554,12 +548,38 @@ def cloud_label_filter(record):
   logging.getLogger().addHandler(handler)
 
 
+def configure_swarming(name: str, extras: dict[str, str] = None) -> None:
```
letitz:
```suggestion
def configure_swarming(name: str, extras: dict[str, str] | None = None) -> None:
```

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -93,26 +95,18 @@ def _console_logging_enabled():
...
   """Return bool True when logging to files (bot/logs/*.log) is enabled.
-  This is enabled by default.
-  This is disabled if we are running in app engine or kubernetes as these have
-    their dedicated loggers, see configure_appengine() and configure_k8s().
-  """
-  return bool(os.getenv(
-      'LOG_TO_FILE',
-      'True')) and not _is_running_on_app_engine() and not _is_running_on_k8s()
+  This is enabled by default."""
+  return environment.get_value('LOG_TO_FILE', True)
```
letitz, line 101: This is certainly more correct, but it's also a behavior change (same goes for `LOG_TO_GCP` below). We should call it out in the PR description. Alternatively consider splitting this specific change out to its own PR.

src/python/bot/startup/run_bot.py
letitz: We should mention this change in the PR description (which will become the commit message).

letitz (approved):
> Currently swarming bots use the K8s default logging configurations(which work fine :D), but when we try to find logs specific to a given swarming bot host or a swarming task, things start to get complicated, its hard to pinpoint the execution workflow.
> 
> In this PR we changed the configuration so the logger can displays different information if its running on swarming.

This is vague. Please mention what is actually changing:

* new metadata fields on logs
* dropped metadata fields
* any other behavioral changes from the PR
* any refactorings done along the way

LGTM otherwise to avoid blocking on me further, but please resolve comments before landing.

IvanBM18:
> > Currently swarming bots use the K8s default logging configurations(which work fine :D), but when we try to find logs specific to a given swarming bot host or a swarming task, things start to get complicated, its hard to pinpoint the execution workflow.
> > In this PR we changed the configuration so the logger can displays different information if its running on swarming.
> 
> This is vague. Please mention what is actually changing:
> 
> * new metadata fields on logs
> * dropped metadata fields
> * any other behavioral changes from the PR
> * any refactorings done along the way
> 
> LGTM otherwise to avoid blocking on me further, but please resolve comments before landing.

Sure i'll do that, thanks for the feedback :D

src/clusterfuzz/_internal/metrics/logs.py
IvanBM18: I'll keep them in this PR, so that the GOSST team can have the full context on why we made this small   business logic change. So I'll update the description accordingly🫡

jardondiego (approved):

javanlacerda (commented):
As it touches in code that can have misbehavior for the other environments, it would be good to have it running in dev for a while. Could you deploy this change there and then provide evidence that there's no any side effect?

IvanBM18:
> As it touches in code that can have misbehavior for the other environments, it would be good to have it running in dev for a while. Could you deploy this change there and then provide evidence that there's no any side effect?

Hi @javanlacerda this changes have been in `dev` since April 16th. I have been using them since that day to easily get info from various jobs running on swarming. There was only one issue:
```
  File "/mnt/scratch0/clusterfuzz/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py", line 429, in __exit__
    'runtime': environment.get_runtime().value,
               ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/scratch0/clusterfuzz/src/clusterfuzz/_internal/system/environment.py", line 747, in get_runtime
    if is_running_on_swarming():
       ^^^^^^^^^^^^^^^^^^^^^^
NameError: name 'is_running_on_swarming' is not defined
```
But this one only happened because of merge conflicts when cherry-picking the commit(which i just fixed)
Another errogroup i see a lot is this one:
```
Code: 403. Reason: <?xml version='1.0' encoding='UTF-8'?><Error><Code>AccessDenied</Code><Message>Access denied.</Message><Details>gcs-signer@clusterfuzz-development.iam.gserviceaccount.com does not have storage.objects.get access to the Google Cloud Storage object. Permission 'storage.objects.get' denied on resource (or it may not exist).</Details></Error>. Raise.
```
But this it's happening because no target was found for the given job, which doesn't seem related to my changes.

Apart from that everything looks healthy :D

javanlacerda (approved):

PR #5248, opened 2026-04-20
Swarming: Tag metrics with swarming runtime.

IvanBM18:
Currently if we run fuzzing sessions on swarming, the metrics dont seem to register those fuzzing hours correctly:
image
Note that the swarming fuzzing hours appear as empty: `runtime=`

- This PR adds swarming as a possible runtime
- Adds swarming in the error count metric
- Now that theres no circular dependency between `logs` <-> `environment` modules, i removed some methods that were previously declared twice to avoid said circular dependency, now they are correctly imported from a single module
- Also adds unit tests

## Tests performed
Changes present in dev since before the last `dev` branch reset(which is around 22th of April).

#### Worth noting error logs in `dev` env:
Heres a list of logs that came to my attention:

There is a recurring PERMISSION_DENIED exception in the testcase_manager.py. It specifically points to "Missing or insufficient permissions" when calling the Datastore API via gRPC.
```
method: "_do_run_testcase_and_return_result_in_queue"
path: "/mnt/scratch0/clusterfuzz/src/clusterfuzz/_internal/bot/testcase_manager.py"
}
message: "Exception occurred while running run_testcase_and_return_result_in_queue.
  File "/mnt/scratch0/clusterfuzz/src/third_party/google/cloud/ndb/_datastore_api.py", line 98, in rpc_call
    result = yield rpc
             ^^^^^^^^^
	status = StatusCode.PERMISSION_DENIED
	details = "Missing or insufficient permissions."
	debug_error_string = "UNKNOWN:Error received "
>
```
Not related since we are not making any changes nor to auth or to the data store api.

```
raise RuntimeError('Request to %s failed. Code: %d. Reason: %s' %
RuntimeError: Request to https://storage.googleapis.com....
<Error><Code>SignatureDoesNotMatch</Code><Message>Access denied.</Message><Details>The request signature we calculated does not match the signature you provided. Check your Google secret key and signing method.</Details>
```
A weird error group, seems related to missing permissions or mismatch across service accounts, may be worth looking, or may come from other tests that someone is performing in dev


src/clusterfuzz/_internal/bot/tasks/utasks/__init__.py
```diff
@@ -454,7 +454,7 @@ def uworker_main(input_download_url) -> None:
     _start_web_server_if_needed(uworker_input.job_type)
 
     utask_module = get_utask_module(uworker_input.module_name)
-    execution_mode = Mode.SWARMING if environment.is_swarming_bot(
+    execution_mode = Mode.SWARMING if environment.is_running_on_swarming(
```
jardondiego, line 457: Any reason in particular for this change in name? Not sure it improves descriptiveness.

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -93,26 +95,18 @@ def _console_logging_enabled():
...
-  This is disabled for local development and if we are running in a app engine
-    or kubernetes as these have their dedicated loggers, see
-    configure_appengine() and configure_k8s()."""
-  return (bool(os.getenv('LOG_TO_GCP', 'True')) and
-          not os.getenv("PY_UNITTESTS") and not _is_local() and
-          not _is_running_on_app_engine() and not _is_running_on_k8s())
+  This is enabled by default but disabled for local development."""
+  return environment.get_value('LOG_TO_GCP',
+                               True) and (not os.getenv('PY_UNITTESTS') and
+                                          not _is_local())
```
jardondiego: I think there are cleaner ways to write this expression.

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -93,26 +95,18 @@ def _console_logging_enabled():
...
-def _file_logging_enabled():
+def _file_logging_enabled() -> bool:
   """Return bool True when logging to files (bot/logs/*.log) is enabled.
-  This is enabled by default.
-  This is disabled if we are running in app engine or kubernetes as these have
-    their dedicated loggers, see configure_appengine() and configure_k8s().
-  """
-  return bool(os.getenv(
-      'LOG_TO_FILE',
-      'True')) and not _is_running_on_app_engine() and not _is_running_on_k8s()
```
jardondiego: Do we actually want to remove `not _is_running_on_app_engine() and not _is_running_on_k8s()` from here?

src/clusterfuzz/_internal/system/environment.py
```diff
@@ -785,9 +789,9 @@ def parse_environment_definition(environment_string):
   return values
 
 
-def is_running_on_k8s():
+def is_running_on_k8s() -> bool:
   """Returns whether or not we're running on K8s."""
-  return os.getenv('IS_K8S_ENV') == 'true'
+  return get_value('IS_K8S_ENV', False)
```
jardondiego: This is not guaranteed to return a boolean, since `get_value` will return a Python value for what it receives as value. We should either remote the return type hint or add the necessary checks here. What do you think?

IvanBM18: Adding the necessary checks would make sense it will have discrepancy with a bunch of other validations in CF, but i guess is this is a right step moving forward.

I also eliminated the one we have in the `logs.py` to remove repetead code/validations.

Thanks!

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -54,17 +54,11 @@
 _default_extras = {}
 
 
-def _is_running_on_k8s():
```
IvanBM18, line 57: Now that we dont have a circular dependency we can start using the same method that its declared on `environment.py`

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -79,15 +73,6 @@ def _is_local():
           os.getenv('SERVER_SOFTWARE', '').startswith('Development/'))
 
 
-def _is_running_on_app_engine():
```
IvanBM18, line 82: Same as above:
Now that we dont have a circular dependency we can start using the same method that its declared on `environment.py`

jardondiego:
It looks like we have failing tests. Can you take a look?

jardondiego (approved):
LGTM

src/clusterfuzz/_internal/tests/core/system/environment_test.py
```diff
@@ -51,6 +51,35 @@ def test_set_bot_environment_default_variables(self):
...
+  def test_get_runtime_instance_group(self):
+    """Tests that get_runtime() returns INSTANCE_GROUP when not a uworker."""
+    environment.set_value('UWORKER', False)
+    self.assertEqual(environment.get_runtime(),
+                     environment.UtaskMainRuntime.INSTANCE_GROUP)
+
+  def test_get_runtime_kata_container(self):
+    """Tests that get_runtime() returns KATA_CONTAINER when running on k8s."""
+    environment.set_value('UWORKER', True)
+    environment.set_value('IS_K8S_ENV', 'True')
```
letitz: Is there a reason for the inconsistent use of quotes around environment values?

letitz (approved):
LGTM % nits.

Can you adjust the PR title? It does not describe what the change fixes. How about "Swarming: tag fuzzing hours metrics with correct runtime."?

src/clusterfuzz/_internal/tests/core/system/environment_test.py
IvanBM18: Not really, I added them to triple check in my local. I guess it looks misleading here, i'll remove so that they are consistent

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -558,7 +545,6 @@ def configure_swarming(name: str, extras: dict[str, str] | None = None) -> None:
   global _default_extras
   _default_extras = extras
 
-  logging.basicConfig(level=logging.INFO)
```
javanlacerda, line 561: Why we need to delete this?

javanlacerda (approved):
LGTM! Single question I have is if with this changes, the fuzzing hours for swarming are with the swarming label?

src/clusterfuzz/_internal/metrics/logs.py
IvanBM18: Allowed the logs to be printed in the swarming console, which is a behaviour we dont want to, we want the logs to be only visible at GCP, hence thats why i removed it.

PR #5249, opened 2026-04-20
Swarming: Fix double base64 encoding of secret_bytes in swarming tasks

IvanBM18:
This small Pr fixes a misconception we had since the beginning, we were manually encoding the `input_dowload_url` to then give it to the `secret_bytes` at the swarming request, which in paper is correct. BUT the swarming request proto implicitly encodes the given bytes, hence we were encoding them twice.

So we remove the additional encoding from the request.

## Test performed
This changes have been present in `dev` since April 16th
We are able to [successfully download the input](https://cloudlogging.app.goo.gl/n84Sk58pjRwzNssVA) using the signed url.


letitz (approved):
LGTM

jardondiego (approved):

javanlacerda (approved):
LGTM!

PR #5250, opened 2026-04-21
Handle broken symlinks in devcontainer set up

dylanjew:
I was running into an issue running the server in the devcontainer when I had initialized the config on my host machine.

This PR checks and cleans up broken symlinks correctly

After this change I can run the server in the devcontainer and the symlink is fixed

### DIFF
--- a/src/local/butler/common.py
+++ b/src/local/butler/common.py
@@ -351,7 +351,7 @@ def install_dependencies(platform_name=None):
 
 def remove_symlink(target):
   """Removes a symlink."""
-  if not os.path.exists(target):
+  if not os.path.lexists(target):
     return
 
   if os.path.isdir(target) and get_platform() == 'windows':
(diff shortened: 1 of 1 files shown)


src/local/butler/common.py
```diff
@@ -372,7 +372,7 @@ def symlink(src, target):
   else:
     os.symlink(src, target)
 
-  assert os.path.exists(target), f'Failed to create {target} symlink for {src}.'
+  assert os.path.lexists(target), f'Failed to create {target} symlink for {src}.'
```
aakallam: I get changing to `lexists` for the cleanup method but shouldn't we catch here if the symlink is broken?

aakallam (approved):

src/local/butler/common.py
dylanjew: makes sense! done

vitaliset (approved):

src/local/butler/common.py
```diff
@@ -351,7 +351,7 @@ def install_dependencies(platform_name=None):
 
 def remove_symlink(target):
   """Removes a symlink."""
-  if not os.path.exists(target):
+  if not os.path.lexists(target):
```
ViniciustCosta, line 354: IIRC this is also used by the deployment code. It would be nice to confirm that this does not break the deploy by pushing this to dev

dylanjew: Sure thing! I deployed this to dev and verified things look good.

ViniciustCosta: Nice, thanks!

PR #5251, opened 2026-04-21
Allow bots with RUN_TIMEOUT flash devices

Xeicker:
## Description

There is an assumption that id `RUN_TIMEOUT` is defined then flashing android devices is not needed, however, that's not always true.

This opens the possibility to allow flashing even if `RUN_TIMEOUT` is defined

- Add log entries for future easier debugging

## Testing

- Tested in local that the code follows the expected path, didn't have a flashable device on hand to demonstrate flashing actually occurs, but that might be out of the scope of this PR
- Test on dev was not possible since there are no Android bots available

### DIFF
--- a/src/clusterfuzz/_internal/platforms/android/flash.py
+++ b/src/clusterfuzz/_internal/platforms/android/flash.py
@@ -109,7 +109,12 @@ def flash_to_latest_build_if_needed():
     return
 
   run_timeout = environment.get_value('RUN_TIMEOUT')
-  if run_timeout:
+  force_flashing = utils.string_is_true(
+      environment.get_value('FORCE_ANDROID_FLASHING'))
+
+  if run_timeout and not force_flashing:
+    logs.info('Skipping reimage: RUN_TIMEOUT is set. '
+              'And FORCE_ANDROID_FLASHING is false')
     # If we have a run timeout, then we are already scheduled to bail out and
     # will be probably get re-imaged. E.g. using frameworks like Tradefed.
     return
@@ -121,6 +126,9 @@ def flash_to_latest_build_if_needed():
   needs_flash = last_flash_time is None or dates.time_has_expired(
       last_flash_time, seconds=FLASH_INTERVAL)
   if not needs_flash:
+    logs.info(
+        f'Skipping reimage: FLASH_INTERVAL ({FLASH_INTERVAL}s) has not expired '
+        f'since last flash ({last_flash_time:%Y-%m-%d %H:%M:%S}).')
     return
 
   is_google_device = settings.is_google_device()


src/clusterfuzz/_internal/platforms/android/flash.py
```diff
@@ -109,7 +109,10 @@ def flash_to_latest_build_if_needed():
     return
 
   run_timeout = environment.get_value('RUN_TIMEOUT')
-  if run_timeout:
+  force_flashing = environment.get_value('FORCE_ANDROID_FLASHING')
```
jardondiego: Is this checking for existence or truthiness? We would want to check for the latter, right?

jardondiego (commented):
Looks good, overall. :)
Left a small question there.

src/clusterfuzz/_internal/platforms/android/flash.py
Xeicker: Ah yes you are completely right

src/clusterfuzz/_internal/platforms/android/flash.py
```diff
@@ -109,7 +109,10 @@ def flash_to_latest_build_if_needed():
     return
 
   run_timeout = environment.get_value('RUN_TIMEOUT')
-  if run_timeout:
+  force_flashing = str(environment.get_value('FORCE_ANDROID_FLASHING', 'false')).lower() == 'true'
```
jardondiego: If I recall correctly, there is a utility function `utils.string_is_true` that already does this. I guess my next question would be, if we think of this envvar as boolean, I believe `environment.get_value` as is would return `True` directly.

jardondiego: nit: should we name this as `SHOULD_FORCE_ANDROID_FLASHING`?

jardondiego:
It seems we have failing tests.

src/clusterfuzz/_internal/platforms/android/flash.py
Xeicker: I updated the code to use string is true, looks a lot better.

About the name if it is not too bad it would be great to keep it as is. I already, landed the change for it in the script that setups that env var

Xeicker:
> It seems we have failing tests.

It is not failing, it is pending waiting for a collaborator to give the command `/gcbrun` . If you could help me with that while I obtain that title it would be great

IvanBM18:
> > It seems we have failing tests.
> 
> It is not failing, it is pending waiting for a collaborator to give the command `/gcbrun` . If you could help me with that while I obtain that title it would be great

I just relaunched the CI jobs. Overall changes LGTM

src/clusterfuzz/_internal/platforms/android/flash.py
```diff
@@ -109,7 +109,11 @@ def flash_to_latest_build_if_needed():
     return
 
   run_timeout = environment.get_value('RUN_TIMEOUT')
-  if run_timeout:
+  force_flashing = utils.string_is_true(
+      environment.get_value('FORCE_ANDROID_FLASHING'))
```
jardondiego, line 113:
```suggestion
  force_flashing = environment.get_value('FORCE_ANDROID_FLASHING')
```

jardondiego (approved):

src/clusterfuzz/_internal/platforms/android/flash.py
```diff
@@ -109,7 +109,11 @@ def flash_to_latest_build_if_needed():
...
+
+  if run_timeout and not force_flashing:
+    logs.info('Skipping reimage: RUN_TIMEOUT is set.')
```
letitz: Let's mention `FORCE_ANDROID_FLASHING` too.

letitz (approved):
LGTM % two things:

* One comment in the code about a log
* You misspelled `RUN_TIMEOUT` as `RUN_TIME` in the PR title and description

IvanBM18 (approved):

src/clusterfuzz/_internal/platforms/android/flash.py
jardondiego: Good point. +1

src/clusterfuzz/_internal/platforms/android/flash.py
Xeicker: with that approach there is thie issue with lower case letters:
```
$ FORCE_ANDROID_FLASHING=false python3.11
Python 3.11.9 (main, May  9 2025, 07:23:21) [GCC 14.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from clusterfuzz._internal.system import environment
>>> environment.get_value('FORCE_ANDROID_FLASHING')
'false'
>>> bool(environment.get_value('FORCE_ANDROID_FLASHING'))
True
>>>
```

letitz:
At this point, now that you've gotten LGTM from your teammates, you want to request a review from the maintainers. Added @javanlacerda as poc for you.

ViniciustCosta (approved):

src/clusterfuzz/_internal/platforms/android/flash.py
```diff
@@ -121,6 +126,9 @@ def flash_to_latest_build_if_needed():
   needs_flash = last_flash_time is None or dates.time_has_expired(
       last_flash_time, seconds=FLASH_INTERVAL)
   if not needs_flash:
+    logs.info('Skipping reimage: FLASH_INTERVAL (%ds) has not expired '
```
ViniciustCosta: Please, try to use f-strings

ViniciustCosta (changes requested):

src/clusterfuzz/_internal/platforms/android/flash.py
Xeicker: It is usually not recommended to use f-strings in logging https://google.github.io/styleguide/pyguide.html#3101-logging

ViniciustCosta: In this case it is preferred, since the logging lib is our internal implementation and not the default lib (which expects string literals). The string is evaluated before reaching the logger anyway, so it's best for readability. You can check this usage in most of newer logs (some are older than python 3, so they still use the % operator) :)

Xeicker: Got it. Updated to use f strings

ViniciustCosta: Thanks

ViniciustCosta (approved):
Lgtm

Please fix the linter (try running python butler.py format) and I'll merge it

PR #5252, opened 2026-04-23
[BlackboxAudit] Only update fuzzer.source when creating a fuzzer

dylanjew:
The `Fuzzer.source` is the point of contact or owner of a fuzzer. When the fuzzer is created, the source is the author who uploads the fuzzer. Currently, when we edit the jobs, configuration, or upload a new version of the fuzzer, the `Fuzzer.source` updates to the person who is making that change.

This was fine when the only one maintaining a fuzzer was the current owner of the fuzzer, but breaks down when making LSC type changes to our fuzzer suite. This PR changes the behavior to only set the source when creating a fuzzer and introduces a `last_edited_by` to capture the paper trail of changes. Now, if we want to change the owner we can do so in the datastore directly, and small edits to disable/enable a Fuzzer won't change the source, which is the Fuzzer's owner

The user who is making the edit is also logged as part of the [fuzzer_update_message](https://github.com/google/clusterfuzz/blob/0b9e68ae390d76ae9f475165b1957ca10689f50b/src/appengine/handlers/fuzzers.py#L221).

Tested this locally by creating and editing a fuzzer.

### DIFF
--- a/src/appengine/handlers/fuzzers.py
+++ b/src/appengine/handlers/fuzzers.py
@@ -186,7 +186,7 @@ class BaseEditHandler(base_handler.GcsUploadHandler):
 
     fuzzer.jobs = jobs
     fuzzer.revision = fuzzer.revision + 1
-    fuzzer.source = helpers.get_user_email()
+    fuzzer.last_edited_by = helpers.get_user_email()
     fuzzer.timeout = timeout
     fuzzer.max_testcases = max_testcases
     fuzzer.result = None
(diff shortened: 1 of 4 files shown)


aakallam (approved):

ViniciustCosta:
Maybe it would be nice to create a last edited field for blame reasons?  Logs will get lost after their TTL.

dylanjew:
> Maybe it would be nice to create a last edited field for blame reasons? Logs will get lost after their TTL.

I've added a `last_edited_by` field and verified its set locally

ViniciustCosta (approved):
lgtm, thanks!

PR #5253, opened 2026-04-24
Update CHROME_CHECK_FAILURE_REGEX to match new format.

letitz:
In fuzzing mode, Chromium now ends check failure messages with a semicolon to separate the static condition from runtime variables. This change updates the regex to ignore everything after the semicolon to calculate crash states.

Bug: crbug.com/443588668


dylanjew (approved):

letitz:
@javanlacerda friendly ping. ISTM the k8s tests are broken, unrelated to this change.

ViniciustCosta (approved):

PR #5207, opened 2026-03-13
Set rss_limit_mb to 0 by default for chromium/google centipede targets

aakallam:
This matches the behavior for libFuzzer targets: https://github.com/google/clusterfuzz/blob/9f0c774d414c1d9be2926ed0a8a1ef69d44dd8bd/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/fuzzer.py#L48-L51

PR #5255, opened 2026-04-27
Override centipede default rss_limit_mb for Google/chromium targets

aakallam:
#5207 updated this default only for reproduce() but changes the override location to also affect the the fuzz() and minimize_corpus() methods.

### DIFF
--- a/src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
@@ -179,6 +179,10 @@ class Engine(engine.Engine):
     if options:
       arguments = options.get_engine_arguments('centipede')
 
+    if constants.RSS_LIMIT_MB_FLAGNAME not in arguments and (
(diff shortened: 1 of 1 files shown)


g-ortuno:
How is this done for libfuzzer? I'm curious why the libfuzzer change was a single line but we need to duplicate this for all task types. Seems a bit brittle :/

aakallam:
This matches what libFuzzer is doing by overriding the flag whenever the fuzzer arguments are built from the fuzzer-specific config and the default config: https://github.com/google/clusterfuzz/blob/9f0c774d414c1d9be2926ed0a8a1ef69d44dd8bd/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/fuzzer.py#L32-L65

g-ortuno:
Oh! I see, sorry, the commit message made me think that we changed the specific tasks to use the new value. LGTM!

g-ortuno (approved):
Thanks!

ViniciustCosta (approved):

PR #5263, opened 2026-04-30
[BlackboxBenchmarking] Index the Fuzzer.builtin field

dylanjew:
We use this field to determine whether a fuzzer is blackbox or an engine guided fuzzer. In order to aggregate stats for blackbox fuzzers, we need to be able to query by the `builtin` field, which requires it to be indexed.

https://docs.cloud.google.com/datastore/docs/concepts/indexes#unindexed_properties

The migration to index the field for existing fuzzers just requires rewriting the Fuzzers in the datastore without actually making any changes. Then the index will be created under the hood. 

Testing
I ran this in dev and verified I can filter by `builtin == False`


aakallam (approved):
LGTM!

dylanjew:
Fixed lint error for the unused import

ViniciustCosta (approved):
lgtm

PR #5264, opened 2026-05-01
[BlackboxBenchmarking] Add script to update all fuzzer stats tables

dylanjew:
Some of the old fuzzer stats tables are not updated because the fuzzers are currently disabled. 

This script updates their stats tables to add the new fields so we can query them safely without having to add additional checks on whether the table has the new fields.

This is a one-off migration that needs to be run and does not need to be merged.

I tested this in dev and verified the tables have the new fields

### DIFF
--- a/src/local/butler/scripts/update_bigquery_stats_schema.py
+++ b/src/local/butler/scripts/update_bigquery_stats_schema.py
@@ -1,0 +1,88 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Script to migrate BigQuery JobRun table schemas for old fuzzers."""
+
+from clusterfuzz._internal.base import utils
+from clusterfuzz._internal.datastore import data_types
+from clusterfuzz._internal.datastore import ndb_utils
+from clusterfuzz._internal.google_cloud_utils import big_query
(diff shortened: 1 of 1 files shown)


aakallam (approved):

ViniciustCosta (approved):
Lgtm!

Do you think that this will be used again in the future? Or is it an one-off script?

Historically, we don't actually merge every local script, we just add them for review, run when approved, add the result to the PR or associated bug, and then close the PR.

I don't mind merging it if you think this migration can be recurrent :)

dylanjew:
> Lgtm!
> 
> Do you think that this will be used again in the future? Or is it an one-off script?
> 
> Historically, we don't actually merge every local script, we just add them for review, run when approved, add the result to the PR or associated bug, and then close the PR.
> 
> I don't mind merging it if you think this migration can be recurrent :)

This is a one off script, so I don't think it needs to be merged. Thanks for taking a look, I'll close out this PR. 
I can add whether it is a one off migration to the descriptions going forward as well.

dylanjew:
I need to run the script again once https://github.com/google/clusterfuzz/pull/5263 is deployed to prod and the builtin field is completely indexed so that this script can filter correctly on that field.  I tried running it just now but it misses some fuzzers if any of them are updated in prod without that PR deployed

dylanjew:
Ran it again just now after yesterday's release was deployed and it picked up the last 2 fuzzers that weren't updated yetesrday.

```
python butler.py run -c ../clusterfuzz-config/configs/internal  --non-dry-run update_bigquery_stats_schema
/usr/local/google/home/dylanj/clusterfuzz/src/third_party/google/cloud/ndb/__init__.py:24: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  from pkg_resources import get_distribution
empty google cloud project
app id cluster-fuzz
Starting BigQuery stats tables schema migration.
Checking 100 fuzzer stats tables.
Successfully updated schema for fuzzer: ochang_domfuzzer
Successfully updated schema for fuzzer: b0ring_webidl_fuzzer
BigQuery schema migration complete. Updated 2 fuzzer table.

Please remember to run the migration individually on all projects
```

PR #5265, opened 2026-05-04
[BlackboxBenchmarking] Add daily cron job to aggregate fuzzer stats

dylanjew:
Adds cron job to aggregate fuzzer stats into a daily bigquery table `fuzzer_stats.daily_stats`. 

## Context

We will use this to benchmark our blackbox fuzzers, previously we couldn't easily join the fuzzing hours from BigQuery with the bugs filed by clusterfuzz in our dashboards. We need a separate aggregated table because the `fuzzer_stats` `JobRun` tables are all in separate datasets per fuzzer, and we can't simply query across all of those datasets in BigQuery or Plx.

The cron job defaults to yesterdays stats so we can run it after the stats are loaded into bigquery, but takes a date flag so we can backfill days as necessary. 

### Idempotency
Whenever a date is inserted, the schema uses `WRITE_TRUNCATE` with a date partition to overwrite all of the rows for that date. So if the job runs multiple times for the same day, it will not add additional rows but overwrite any previous rows for that date.

This simplifies edge cases where the job fails or runs multiple times. We can just make sure the last run of the job succeeds and the data will be correct. It will just pull in the latest data on the JobRun tables for the fuzzers.

#### Example query:
```
select fuzzer_name,
SUM(fuzzing_duration) as fuzzing_duration,
SUM(testcases_executed) as testcases_executed,
from `your-project.fuzzer_stats.daily_stats`
group by fuzzer_name
order by fuzzing_duration desc
limit 1000;
```

The remaining work here is to set up the cron job configuration. This PR only adds the logic for the job. [crbug.com/501066151](https://crbug.com/501066151)

### Related PRs:
These migrate the bigquery and datastore schemas to support the new fields
https://github.com/google/clusterfuzz/pull/5264
https://github.com/google/clusterfuzz/pull/5263

### Testing
Ran this against the dev data and verified that the fuzzer stats bigquery table is populated.
Logs from dev: https://paste.googleplex.com/4884361662038016

After the job inserted the aggregated rows into BigQuery, I was able to compare the aggregated testcase stats and fuzzing hours between fuzzers for a given date range.


aakallam (approved):

src/clusterfuzz/_internal/cron/aggregate_fuzzer_stats.py
```diff
@@ -0,0 +1,337 @@
...
+        'name': 'fuzzing_duration',
+        'type': 'INTERVAL',
+        'mode': 'NULLABLE'
+    }]
+}
+
+
+def _execute_insert_request(request):
+  """Executes a table/dataset insert request, retrying on transport errors."""
+  for i in range(NUM_RETRIES + 1):
```
ViniciustCosta: Not sure if applicable here, but we do have a retry module which may be better to reuse (https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/base/retry.py)

ViniciustCosta (approved):
lgtm

you also need to create a new helm template in clusterfuzz-config to setup this cronjob.

src/clusterfuzz/_internal/cron/aggregate_fuzzer_stats.py
dylanjew: Thanks, I've added the retry wrapper! much cleaner 😄

dylanjew:
> lgtm
> 
> you also need to create a new helm template in clusterfuzz-config to setup this cronjob.

Thanks for the review! 

I opened go/clusterfuzz-config-pr/299 and go/clusterfuzz-config-pr/300 for the helm config. I've verified that the cron jobs are running with the right command in dev, but I found that it always pulls in the code from master, rather than what I've pushed to dev. So once this is deployed I can verify the job is working e2e in dev. [logs](https://cloudlogging.app.goo.gl/LyJ8aHLLSQkGNQLB7)

PR #5266, opened 2026-05-05
Bulter.py: Adds `preprocess` butler script for local preprocess

IvanBM18:
# Adds `preprocess` butler script
This butler script allows developers to trigger the preprocess portion of a `fuzz` task and in consecuence generate the serialized and compressed `uworker_input` payload, upload it to real GCS, and get the signed download URL, exactly as it happens remotely. We can then use the resulting url to trigger a task in any backend that we want:
- In swarming trough a prpc request
- In batch trough manually posting the task to the `utask_main` queue

This accelerates local debugging of the `tworker` preprocessing phase without relying on remote execution queues, which has proven to take multiple hours to "ACK" a task request.  

Note: To use this command you need the `Secret Manager Secret Accessor` for Dev or setup a service account key in your local(by using the `gcloud auth` cli) that has said role and any other role required for a tworker's `preprocess`.

#### Example:
With an input like this
```bash
pipenv run python butler.py run preprocess --config-dir ~/clusterfuzz-config/configs/ --script_args fuzzer linux_asan_fuzzer
```
You should get an output like this:
```bash
.......
Preprocess successful!
Input Download URL: https://storage.googleapis.com/uworker-input.project/.....
```

## Changes

- Added the preprocess script.
  - It interacts with the actual Datastore and GCS based on the provided configuration.
  - It fetches and populates the uworker_env with:
    - Job-specific environment variables from the Datastore.
    - Fuzzer-specific environment variables (for blackbox fuzzers).
    - Required logging metadata (CF_TASK_NAME, CF_TASK_ARGUMENT, CF_TASK_JOB_NAME, CF_TASK_ID) to ensure logs in  the subsequent uworker_main step have the correct context as if we executed it using the usual `commands.py` module workflow.

## Tests performed
Executed the following command in dev:
```bash
pipenv run python butler.py run preprocess -c <config_dir>  --script_args <fuzzer> <job>
```
Successfully creates and uploads the payload and returns a valid signed URL. This signed url was later used to trigger a swarming task trough prpc, here are [the logs](https://cloudlogging.app.goo.gl/umtgtBzd6MYVraas6)

### DIFF
--- a/src/local/butler/preprocess.py
+++ b/src/local/butler/preprocess.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""uworker_preprocess.py runs the preprocess step of a fuzz task locally."""
+"""preprocess.py runs the preprocess step of a fuzz task locally."""
 
 import os
 import sys
@@ -43,21 +43,19 @@ def _get_fuzzer_environment(fuzzer_name, job_name):
     
   fuzzer = data_types.Fuzzer.query(data_types.Fuzzer.name == fuzzer_name).get()
   if not fuzzer:
-    print(f"Error: Fuzzer {fuzzer_name} not found in Datastore.")
-    sys.exit(1)
-
+    return {}
     
   additional_default_variables = ''
   additional_variables_for_job = ''
   
-  if fuzzer.additional_environment_string:
-    for env_variable_line in fuzzer.additional_environment_string.splitlines():
-      if '=' in env_variable_line and ':' in env_variable_line.split('=', 1)[0]:
-        fuzzer_job_name, environment_definition = env_variable_line.split(':', 1)
+  if hasattr(fuzzer, 'additional_environment_string') and fuzzer.additional_environment_string:
+    for line in fuzzer.additional_environment_string.splitlines():
+      if '=' in line and ':' in line.split('=', 1)[0]:
+        fuzzer_job_name, environment_definition = line.split(':', 1)
         if fuzzer_job_name == job_name:
           additional_variables_for_job += '\n%s' % environment_definition
         continue
-      additional_default_variables += '\n%s' % env_variable
+      additional_default_variables += '\n%s' % line
       
   env_string = additional_default_variables + additional_variables_for_job
   return environment.parse_environment_definition(env_string)
@@ -95,7 +93,7 @@ def _early_setup(args):
 
 
 def execute(args):
-  """Executes the uworker_preprocess command."""
+  """Executes the preprocess command."""
   _early_setup(args)
   
   print(f'Running preprocess for fuzzer: {args.fuzzer}, job: {args.job}')
--- a/src/local/butler/preprocess.py
+++ b/src/local/butler/preprocess.py
@@ -22,8 +22,8 @@ from clusterfuzz._internal.bot.tasks.utasks import fuzz_task
 from clusterfuzz._internal.config import local_config
 from clusterfuzz._internal.datastore import data_types
 from clusterfuzz._internal.datastore import ndb_init
-from clusterfuzz._internal.system import environment
 from clusterfuzz._internal.metrics import logs
+from clusterfuzz._internal.system import environment
 
 
 def _get_job_environment(job_name):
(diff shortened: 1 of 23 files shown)


src/local/butler/preprocess.py
```diff
@@ -0,0 +1,115 @@
...
+  # Add logging metadata to be carried over to uworker_main
+  uworker_env['CF_TASK_NAME'] = 'fuzz'
+  uworker_env['CF_TASK_JOB_NAME'] = args.job
+  uworker_env['CF_TASK_ARGUMENT'] = args.fuzzer
+  uworker_env['CF_TASK_ID'] = str(uuid.uuid4())
+
+  return uworker_env
+
+
+def _early_setup(args):
```
jardondiego: nit: any reason not to name it just `setup`?

jardondiego:
> Note: To use this command you need the Secret Manager Secret Accessor for Dev or setup a service account in your local that has those permissions.

Did you use a `.json` service account for this? It would be nice to have a few sentences/links to understand it better.

jardondiego:
Does it need to be a subcommand of `butler` itself? Why not make it a standalone script? I think it's a good idea to not over-populate butler with subcommands. What do you think?

src/local/butler/preprocess.py
```diff
@@ -0,0 +1,115 @@
...
+  env_string = additional_default_variables + additional_variables_for_job
+  return environment.parse_environment_definition(env_string)
+
+
+def _get_uworker_env(args):
+  """Prepares the complete environment variables for the payload."""
+  uworker_env = _get_job_environment(args.job)
+  uworker_env.update(_get_fuzzer_environment(args.fuzzer, args.job))
+
+  # Replicate what process_command_impl does in a real tworker
```
letitz, line 65: Could we use `process_command_impl()` then instead?

IvanBM18:
@jardondiego 
> > Note: To use this command you need the Secret Manager Secret Accessor for Dev or setup a service account in your local that has those permissions.
> 
> Did you use a `.json` service account for this? It would be nice to have a few sentences/links to understand it better.

Not in this case, but its possible to use a service account, you just need to generate a key, save it in your local and set it up as the default credentials for any gcloud library and cli operation. This is done using the `gcloud auth` subcommand.

Added more context in the description so future reviewers can easily understand this

IvanBM18:
@jardondiego 
> Does it need to be a subcommand of `butler` itself? Why not make it a standalone script? I think it's a good idea to not over-populate butler with subcommands. What do you think?

Yes, its need to as butler already handles a lot of bootstrapping operations for the same purpose, for example if we didn't use `butler`, we would need to add methods to read, parse, and populate configurations based on the yaml files

src/local/butler/preprocess.py
IvanBM18: Since most of this script is just env setup i thought it was easier for me to understand what made this method different or what is purpose is if i called it early_setup

src/local/butler/preprocess.py
IvanBM18: At the end of said method we call run_command():
https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/tasks/commands.py#L482
Which in turns triggers a workflow in which the preprocess step immediately queues the main task for remote execution when finished or just straight ups executes all 3 steps in the same machine(depending on setup), but we don't want that, we want to stop just after finishing the preprocess so we could manually trigger the main portion wherever and whenever we need to

jardondiego:
> Yes, its need to as butler already handles a lot of bootstrapping operations for the same purpose, for example if we didn't use butler, we would need to add methods to read, parse, and populate configurations based on the yaml files

I think we are referring to different things, what I mean is that I think we could have as a standalone butler script so that we can run it with as
```bash
python butler.py run <name_of_script> --non-dry-run --config $MY_DIR
```

That way we don't have to handle all of that by ourselves. Does it make sense?

jardondiego:
> Added more context in the description so future reviewers can easily understand this

Thanks!

src/local/butler/preprocess.py
jardondiego: I think `setup` by itself makes more sense and it is also more common to find throughout the codebase but I am also okay if we go through with it as is.

IvanBM18: done :D

jardondiego (approved):
LGTM

src/local/butler/preprocess.py
letitz: I see, thanks. Then this works for me. It will inevitably drift apart from the prod implementation, especially since this codepath is not exercised in production, but it's good enough to run for now and commit for others to reuse :)

letitz (approved):
LGTM as a testing tool, not production code.

ViniciustCosta (approved):

PR #5271, opened 2026-05-12
[BlackboxBenchmarking] Remove dry run arg from aggregate_fuzzer_stats cron

dylanjew:
removes dry run args and makes args optional in `aggregate_fuzzer_stats` cron job

The bot version of [run_cron.py](https://github.com/google/clusterfuzz/blob/47ed79626d0fb4fc873f3e4c4646f5fdda6a9630/src/python/bot/startup/run_cron.py#L68) does not accept any args, while the [local run_cron.py](https://github.com/google/clusterfuzz/blob/47ed79626d0fb4fc873f3e4c4646f5fdda6a9630/src/local/butler/scripts/run_cron.py#L96) does accept args.

I previously used required args and this caused the cron job to fail in dev after I pushed the helm config.

```
Unhandled exception in cron: main() missing 1 required positional argument: 'argv'
  File "/mnt/scratch0/clusterfuzz/src/python/bot/startup/run_cron.py", line 68, in main
    return 0 if task_module.main() else 1
                ^^^^^^^^^^^^^^^^^^
TypeError: main() missing 1 required positional argument: 'argv'
```

### DIFF
--- a/src/clusterfuzz/_internal/cron/aggregate_fuzzer_stats.py
+++ b/src/clusterfuzz/_internal/cron/aggregate_fuzzer_stats.py
@@ -226,15 +226,10 @@ def _gather_all_stats(fuzzers, project_id, target_date_str):
 
 
 def _persist_daily_stats(all_rows, bigquery_client, project_id,
-                         date_partition_str, non_dry_run):
+                         date_partition_str):
   """Writes gathered row statistics to destination table."""
   if not all_rows:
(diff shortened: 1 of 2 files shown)


aakallam (approved):

ViniciustCosta (approved):
Yeah, the local run_cron.py is different from the prod one. I think this was a recent change unfortunatelly.

PR #5272, opened 2026-05-12
Add MAX_EXECUTIONS environment variable limit

jardondiego:
This PR introduces a `MAX_EXECUTIONS` environment variable that limits the number of tasks the bot will execute before exiting cleanly. It will facilitate testing locally moving forward.

This is particularly useful when running a local bot with a `COMMAND_OVERRIDE`, allowing developers to restrict the bot to a specific number of iterations rather than looping infinitely.

**Changes:**
* Added an `execution_count` tracker to the main `task_loop` in `run_bot.py`.
* The bot will break out of the loop with a clean exit if `MAX_EXECUTIONS` is set and reached.
* Added corresponding unit tests in `run_bot_test.py`.


src/python/bot/startup/run_bot.py
```diff
@@ -191,6 +192,13 @@ def task_loop():
       time.sleep(utils.random_number(1, failure_wait_interval))
       break
 
+    execution_count += 1
+    max_executions = environment.get_value('MAX_EXECUTIONS')
+    if max_executions and execution_count >= int(max_executions):
+      logs.info('Reached MAX_EXECUTIONS limit (%s). Exiting.', max_executions)
```
IvanBM18: Nit: Can you format using f strings?

```suggestion
      logs.info(f'Reached MAX_EXECUTIONS limit {max_executions}. Exiting.')
```
We don't really format strings that way in the project.
Besides i think that using the `logs.info()` method that way would result in the following log:
```
....
extras :{
  max_executions: 5
},
message: "Reached MAX_EXECUTIONS limit (%s). Exiting."
....
```

IvanBM18 (commented):
LGTM

IvanBM18 (approved):

src/python/bot/startup/run_bot.py
jardondiego: Good catch! Thanks!

src/clusterfuzz/_internal/tests/core/bot/startup/run_bot_test.py
```diff
@@ -107,6 +109,36 @@ def test_exception(self):
     self.assertFalse(clean_exit)
     self.assertEqual('payload', payload)
 
+  def test_max_executions(self):
+    """Test that the loop breaks after MAX_EXECUTIONS iterations."""
+    from clusterfuzz._internal.system import environment
+    environment._initial_environment = None
+    os.environ['MAX_EXECUTIONS'] = '3'
+    # Use side_effect to avoid raising an exception so it loops cleanly.
+    self.mock.process_command.side_effect = None
```
letitz: I guess this works, but do we even need it? And why not use `return_value` instead?

src/python/bot/startup/run_bot.py
```diff
@@ -114,12 +114,26 @@ def schedule_utask_mains():
...
 
 
+def _get_max_executions():
+  """Returns the MAX_EXECUTIONS limit as an int, or None if invalid/unset."""
+  val = environment.get_value('MAX_EXECUTIONS')
+  if not val:
+    return None
+  try:
+    return int(val)
+  except ValueError:
```
letitz, line 125: I think we should instead loudly fail and exit if the environment variable is wrong. It would surface configuration issues immediately, right before we start the task loop, whereas logging an error gets a bit lost in the noise.

src/python/bot/startup/run_bot.py
```diff
@@ -191,6 +205,13 @@ def task_loop():
       time.sleep(utils.random_number(1, failure_wait_interval))
       break
 
+    execution_count += 1
+    max_executions = _get_max_executions()
```
letitz: This should move to above the while loop. There is no reason to re-check the environment variable for a new value on every iteration - it should have the same value for the lifetime of the whole process.

This also ties in with loudly failing on invalid values. If we check the value before we even enter the loop, we will immediately surface configuration errors instead of waiting until after we run a task to notice the error.

Finally, consider switching to a for loop instead. It looks like `itertools.repeat()` would work well here:

```py
max_executions = _get_max_executions()
for _ in itertools.repeat(None, times=max_executions):
  ...
```

See https://docs.python.org/3/library/itertools.html#itertools.repeat

I guess the downside of the for loop is that it's harder to log the clean exit message.

letitz (changes requested):

src/python/bot/startup/run_bot.py
```diff
@@ -114,12 +114,26 @@ def schedule_utask_mains():
         task.pubsub_task.cancel_lease_ack()
 
 
+def _get_max_executions():
+  """Returns the MAX_EXECUTIONS limit as an int, or None if invalid/unset."""
+  val = environment.get_value('MAX_EXECUTIONS')
```
letitz: Ah, and one final comment: `max_executions` is a bit vague - clusterfuzz also runs fuzzers that have their own concept of executions, execution speed, etc.. Consider `max_task_executions` to disambiguate.

jardondiego: Will do!

src/python/bot/startup/run_bot.py
jardondiego: Agreed. I will update this.

src/clusterfuzz/_internal/tests/core/bot/startup/run_bot_test.py
jardondiego: Good catch! This is redundant. I've removed this line in the last commit. Thanks!

src/python/bot/startup/run_bot.py
jardondiego: Fixed!

ViniciustCosta (commented):
Seems good! Could you just add an evidence of it not blocking execution in dev environment?

jardondiego:
> Seems good! Could you just add an evidence of it not blocking execution in dev environment?

Will do!

javanlacerda:
> Seems good! Could you just add an evidence of it not blocking execution in dev environment?

+1

jardondiego:
Changes are now present in `dev`, will monitor a post about results back here.

jardondiego:
> Changes are now present in dev, will monitor a post about results back here.

These changes have now been present in `dev` for a few hours, and logs are looking consistent as well as general metrics.

Metrics
https://pantheon.corp.google.com/monitoring/dashboards/builder/9835e451-d912-4a4f-8c67-8165fdde7061;duration=PT12H?project=clusterfuzz-development&e=-13802955&mods=logs_tg_prod&pageState=(%22eventTypes%22:(%22selected%22:%5B%22CLOUD_ALERTING_ALERT%22,%22VM_TERMINATION%22,%22GKE_WORKLOAD_DEPLOYMENT%22%5D))

Logs
https://pantheon.corp.google.com/logs/query;query=severity%3D%22ERROR%22;cursorTimestamp=2026-05-21T05:06:57.868814025Z;histogramBreakdownField=severity;duration=P1D?referrer=search&e=-13802955&mods=logs_tg_prod&project=clusterfuzz-development

src/python/bot/startup/run_bot.py
```diff
@@ -114,17 +114,16 @@ def schedule_utask_mains():
...
-  """Returns the MAX_EXECUTIONS limit as an int, or None if invalid/unset."""
-  val = environment.get_value('MAX_EXECUTIONS')
+def _get_max_task_executions():  # pylint: disable=inconsistent-return-statements
+  """Returns the MAX_TASK_EXECUTIONS limit as an int, or None if
+  invalid/unset."""
+  val = environment.get_value('MAX_TASK_EXECUTIONS')
   if not val:
     return None
   try:
     return int(val)
```
letitz, line 124: [Optional] I think you could just remove the try/catch and let `ValueError` be raised instead.

src/python/bot/startup/run_bot.py
```diff
@@ -206,9 +207,10 @@ def task_loop():
       break
 
     execution_count += 1
-    max_executions = _get_max_executions()
-    if max_executions and execution_count >= max_executions:
-      logs.info(f'Reached MAX_EXECUTIONS limit ({max_executions}). Exiting.')
+    if max_task_executions and execution_count >= max_task_executions:
+      logs.info(
+          'Reached MAX_TASK_EXECUTIONS limit. Exiting.',
```
letitz: Consider inlining the limit value in the log message body instead of as metadata. It's small anyway.

letitz (approved):
Two nits below. LGTM % updating the PR description to match the variable rename.

ViniciustCosta (approved):

PR #5274, opened 2026-05-13
[BlackboxBenchmarking] Adds return True for aggregate fuzzer stats cron

dylanjew:
The bots `run_cron` script expects the cron job's `main()` to return `True`, otherwise it assumes the script failed if it returns None ([code](https://github.com/google/clusterfuzz/blob/0ac00ac0508df394073ae3805363a13b85498b1a/src/python/bot/startup/run_cron.py#L68)). This was causing the job to retry multiple times in dev.

[logs](https://cloudlogging.app.goo.gl/rGCGx5Nkc251ZuVh8)

### DIFF
--- a/src/clusterfuzz/_internal/tests/appengine/handlers/cron/aggregate_fuzzer_stats_test.py
+++ b/src/clusterfuzz/_internal/tests/appengine/handlers/cron/aggregate_fuzzer_stats_test.py
@@ -72,7 +72,7 @@ class AggregateFuzzerStatsTest(unittest.TestCase):
         }],
         total_count=1)
 
-    aggregate_fuzzer_stats.main()
+    self.assertTrue(aggregate_fuzzer_stats.main())
 
     self.mock_api_client.datasets().insert.assert_called_with(
         projectId='test-clusterfuzz',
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/tests/appengine/handlers/cron/aggregate_fuzzer_stats_test.py
```diff
@@ -72,7 +72,8 @@ def test_aggregate_fuzzer_stats(self):
         }],
         total_count=1)
 
-    aggregate_fuzzer_stats.main()
+    result = aggregate_fuzzer_stats.main()
+    self.assertTrue(result)
```
aakallam: Nit: combine these lines so that an assertion failure is more descriptive.

```suggestion
    self.assertTrue(aggregate_fuzzer_stats.main())
```

Same with the new `self.assertTrue` below.

aakallam (approved):

src/clusterfuzz/_internal/tests/appengine/handlers/cron/aggregate_fuzzer_stats_test.py
dylanjew: Thanks, done!

ViniciustCosta (approved):

PR #5276, opened 2026-05-14
Swarming: Add TasksCountRequest and TasksCount to swarming.proto

IvanBM18:
As part of the initiative of creating a backpressure logic for the swarming integration we are adding first the required protobuf definitions for a upcoming(in another pr) `/prpc/swarmig.v2.tasks/CountTasks` prpc endpoint.

### Background
For more context on the backpressure initative see more info in internal fuzz docs

## Changes
- Adds the following protobuf definitions to the `swarming.proto` file:
  -  CountTask message
  - TasksCountRequest message
  - StateQuery enum
- Updated the reference to the chromium repo in the swarming.proto file
- Ran the `src/clusterfuzz/_internal/protos/generate.sh` script, which added a bunch of changes to the generated files.

## Tests
No unit tests nor actual tests in the CF code are  required for this one, as we are just adding more protobuf definitions not doing nothing new with them(yet),

PR #5277, opened 2026-05-14
Swarming: Adds CountTasks endpoint & Refactors swarming Prpc request

IvanBM18:
As part of the Swarming backpressure initiative, we needed to launch request to the prpc's `swarming.v2.Task/CountTasks` endpoint so we can calculate the amount of fuzz tasks needed to schedule, to achieve this i moved the prpc request logic to its own module so that we can reuse code and keep responsibilities in separate files.

## Changes
- Created a new `api.py` file in the swarming module.
  - Moved the request logic from `__init__.py` to  `api.py`
  - Removed push_swarming_task from `init.py`
  - Updated `service.py` to use `SwarmingAPI` directly.
  - Created a new class `SwarmingAPI` to handle the prpc request logic such as token/auth logic, swarming config handling, and any other request logic
- Created new unit tests for this api and Updated existing unit tests.

## Tests
- Created unit tests for the new api & endpoint
- Scheduled swarming jobs to verify that the refactor didn't broke the existing swarming logic
  - See [example logs](https://cloudlogging.app.goo.gl/JurSX9HsEpuZP1686)


src/clusterfuzz/_internal/swarming/api.py
```diff
@@ -0,0 +1,118 @@
...
+_SWARMING_SCOPES = [
+    'https://www.googleapis.com/auth/cloud-platform',
+    'https://www.googleapis.com/auth/userinfo.email'
+]
+
+
+class SwarmingAPI:
+  """Client for Swarming pRPC API."""
+
+  def __init__(self):
```
Xeicker: This constructor is duplicating the functionality of `_get_swarming_config` you could give the config as an argument to this class. Or if it makes sense obtain the config from this class and delete that other function

src/clusterfuzz/_internal/swarming/api.py
```diff
@@ -0,0 +1,118 @@
...
+    try:
+      self.config = local_config.SwarmingConfig()
+    except (BadConfigError, ValueError) as e:
+      logs.error(f'[Swarming] Failed to retrieve config: {e}')
+      self.config = None
+
+  def _get_headers(self) -> dict:
```
Xeicker: nit
`-> dict[str,str]`

src/clusterfuzz/_internal/swarming/api.py
```diff
@@ -0,0 +1,118 @@
...
+    if not creds.token:
+      creds.refresh(requests.Request())
+
+    return {
+        'Accept': 'application/json',
+        'Content-Type': 'application/json',
+        'Authorization': f'Bearer {creds.token}'
+    }
+
+  def push_task(self, task_request: swarming_pb2.NewTaskRequest) -> str | None:  # pylint: disable=no-member
```
Xeicker, line 120: `push_task` and `count_task` look so similar that I would count them as duplicated code. Can you consider creating a third function that handles the small differences between these 2?

src/clusterfuzz/_internal/swarming/api.py
```diff
@@ -0,0 +1,118 @@
...
+        f"[Swarming] Pushing task {task_request.name}",
+        url=url,
+        body=message_body)
+
+    response = utils.post_url(url=url, data=message_body, headers=headers)
+    logs.info(
+        f'[Swarming] Response from {task_request.name}', response=response)
+    return response
+
+  def count_tasks(self,
```
Xeicker, line 139: I don't see where this function is called, is it expected?

Xeicker (commented):
For future PRs, when you move code and then add functionality it is a good practice to do each of those steps in separate commits for an easier review

src/clusterfuzz/_internal/swarming/api.py
IvanBM18: Yes, this prs focuses on adding the new request logic and i'll use this new method in future PRs

src/clusterfuzz/_internal/swarming/api.py
IvanBM18: You are right i forgot to do that... thanks for pointing it out!

src/clusterfuzz/_internal/swarming/api.py
IvanBM18: Sure, justcreated a new method `_make_request()` with almost all the common code

src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
```diff
@@ -252,109 +250,6 @@ def test_get_spec_from_config_for_fuzz_task(self):
         ])
     self.assertEqual(spec, expected_spec)
 
-  def test_push_swarming_task(self):
```
IvanBM18, line 255: Note: this test method is no longer needed here, in its place we added unit tests at the `api_test` that checks for the same functionality

src/clusterfuzz/_internal/swarming/api.py
```diff
@@ -103,16 +119,12 @@ def count_tasks(self,
...
-    headers = self._get_headers()
-    if not headers:
-      return None
-
-    swarming_server = self.config.get('swarming_server')
-    url = f'https://{swarming_server}/prpc/swarming.v2.Tasks/CountTasks'
     message_body = json_format.MessageToJson(count_request)
+    logs.info(
+        "[Swarming] Counting tasks in queue",
+        url=self._base_url,
```
Xeicker: nit

Might be useful to have the complete url in the log

src/clusterfuzz/_internal/swarming/api.py
```diff
@@ -103,16 +119,12 @@ def count_tasks(self,
...
+        body=message_body)
 
-    logs.info("[Swarming] Counting tasks", url=url, body=message_body)
-
-    response = utils.post_url(url=url, data=message_body, headers=headers)
+    response = self._make_request('swarming.v2.Tasks/CountTasks', message_body)
```
Xeicker: nit

It sounds to me that endpoints could be module constants

Xeicker (approved):
Other than some nits changes LGTM

src/clusterfuzz/_internal/swarming/api.py
IvanBM18: Done :D

src/clusterfuzz/_internal/swarming/api.py
IvanBM18: fair point, just update it, thanks :D

Xeicker (approved):

src/clusterfuzz/_internal/protos/swarming.proto
jardondiego, line 1: I think these changes should be a in separate PR. I think we should aim at smaller pull requests.

IvanBM18: These changes are indeed in a separate PR:
https://github.com/google/clusterfuzz/pull/5276
In this Pr description i noted that this branch is based off that one, because the changes on that pr are needed in this one

jardondiego: In that case, this should be a draft until that one is merged, right?

jardondiego: I think it's confusing to have the two in active simultaneously.

jardondiego: What we could do is have this one point to the branch associated with #5276 and make this a draft, once that other PR is merged we can point this PR to`master`. What do you think?

IvanBM18: Seems fair, just did so with this pr :D

letitz: The base PR is a commit in this PR, so you can view the diff using this link: https://github.com/google/clusterfuzz/pull/5277/changes/fc9445964181d134b58f17a405fcbfc17ed62f09..5299ee9d3c229b335f3bf3fb86f10bc9c3653e0e

I used this in past PRs I sent that were stacked, by commenting on the PR with the link to review only the relevant diff, in a bid to make reviewers' lives easier. Still, GitHub really doesn't do a great job of supporting this workflow.

src/clusterfuzz/_internal/swarming/__init__.py
```diff
@@ -87,15 +79,15 @@ def _get_task_dimensions(job: data_types.Job, platform_specific_dimensions: list
...
+  swarming_config = get_swarming_config()
   if not swarming_config:
     logs.error(
         '[Swarming] No dimensions set. Reason: failed to retrieve config')
     return []
 
   unique_dimensions = {}
   unique_dimensions['os'] = str(job.platform).capitalize()
-  unique_dimensions['pool'] = _get_swarming_config().get('swarming_pool')
+  unique_dimensions['pool'] = get_swarming_config().get('swarming_pool')
```
letitz: We already looked up the config on line 82. Let's use it:

```suggestion
  unique_dimensions['pool'] = swarming_config.get('swarming_pool')
```

src/clusterfuzz/_internal/swarming/api.py
```diff
@@ -0,0 +1,135 @@
...
+]
+
+_COUNT_TASKS_ENDPOINT = 'swarming.v2.Tasks/CountTasks'
+_NEW_TASK_ENDPOINT = 'swarming.v2.Tasks/NewTask'
+
+
+class SwarmingAPI:
+  """Client for Swarming pRPC API."""
+
+  _config: SwarmingConfig = None
```
letitz:
```suggestion
  _config: SwarmingConfig | None = None
```

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -17,11 +17,20 @@
...
 
 class SwarmingService(remote_task_types.RemoteTaskInterface):
   """Remote task service implementation for Swarming."""
 
+  _api: SwarmingAPI = None
+
+  def _get_api(self) -> SwarmingAPI:
+    """Returns the Swarming API instance."""
+    if not self._api:
+      self._api = SwarmingAPI()
```
letitz: Why instantiate this lazily? Let's just instantiate it at construction time, it's less surprising.

src/clusterfuzz/_internal/swarming/api.py
```diff
@@ -0,0 +1,135 @@
...
+      self._base_url = f"https://{self._config.get('swarming_server')}/prpc/"
+
+  def _get_headers(self) -> dict[str, str]:
+    """Checks config and returns headers for pRPC request.
+    
+    Returns:
+      A dict containing headers, or empty dict if config is missing or
+      auth fails.
+    """
+    if not self._config:
```
letitz: If `self._config` is None, then this instance is useless. All its methods will just do nothing, and just log errors when invoked. It's a valid design choice, but it's a bit surprising: this is an API client, unless it's not and then nothing works!

I suggest giving this class a factory function that can fail if the config is missing, and exposing that failure to callers:

```py
@staticmethod
def create() -> 'SwarmingAPI' | None:
  config = get_swarming_config()
  if config is None:
    return None

  return SwarmingAPI(config)

def __init__(self, config: SwarmingConfig):
  self._config = config
  self._base_url = ...
```

Then you can use that in `SwarmingService` to create an optional `_api = SwarmingAPI | None` member. Alternatively, you can inline the contents of `create()` in `SwarmingService.__init__()`.

At that point, you've just moved the problem up one layer (now `SwarmingService` might be useless) but to my sense it's still cleaner.

Now, I think that to make this truly clean you'd want to go even one layer higher and only instantiate `SwarmingService` in `RemoteTaskGate` if swarming is enabled, and fail loudly if the config is missing (presumably if swarming is enabled, we can expect to have a `SwarmingConfig`).

I see that `SwarmingService` is unconditionally instantiated by `RemoteTaskGate` as part of its dict of adapters [1]. I suggest only instantiating adapters if the relevant feature flag is enabled:

```py
    self._service_map = {
        adapter.id: adapter.service()
        for adapter in remote_task_adapters.RemoteTaskAdapters
        if adapter.feature_flag.enabled  # new
    }
```

Then you only create instances of classes at runtime that you actually expect to use. This further limits the blast radius of changes to swarming code: if there is a bug in `SwarmingService.__init__()`, it won't be triggered in ClusterFuzz instances where Swarming is disables.

Finally, you can assert/crash if service instances cannot be constructed properly. In our case, you could assert in `SwarmingService.__init__()` that `SwarmingAPI.create()` succeeded, and make `SwarmingService._api` a non-optional member of type `SwarmingAPI`.

[1]: https://github.com/google/clusterfuzz/blob/e82914eac6a550ea178ed85b07e37cb8987d76ea/src/clusterfuzz/_internal/remote_task/remote_task_gate.py#L45

src/clusterfuzz/_internal/swarming/api.py
```diff
@@ -0,0 +1,135 @@
...
+    """
+    message_body = json_format.MessageToJson(count_request)
+    logs.info(
+        "[Swarming] Counting tasks in queue",
+        url=self._base_url,
+        endpoint=_COUNT_TASKS_ENDPOINT,
+        body=message_body)
+
+    response = self._make_request(_COUNT_TASKS_ENDPOINT, message_body)
+    logs.info('[Swarming] Response from CountTasks', response=response)
```
letitz: This feels like debug-level logs to me, but I'll let ClusterFuzz owners decide here.

src/clusterfuzz/_internal/swarming/api.py
```diff
@@ -0,0 +1,135 @@
...
+    
+    Args:
+      count_request: The TasksCountRequest proto message.
+      
+    Returns:
+      The raw JSON response string from the server, or None if the request
+      could not be made (e.g. missing config, auth failure) or failed.
+    """
+    message_body = json_format.MessageToJson(count_request)
+    logs.info(
```
letitz: Consider moving this log to `_make_request()` since it's basically the same in each case.

src/clusterfuzz/_internal/swarming/api.py
```diff
@@ -0,0 +1,135 @@
...
+_SWARMING_SCOPES = [
+    'https://www.googleapis.com/auth/cloud-platform',
+    'https://www.googleapis.com/auth/userinfo.email'
+]
+
+_COUNT_TASKS_ENDPOINT = 'swarming.v2.Tasks/CountTasks'
+_NEW_TASK_ENDPOINT = 'swarming.v2.Tasks/NewTask'
+
+
+class SwarmingAPI:
```
letitz: Nit: Prefer `Api` over `API`, this is what google c++ [1] and java [2] style guides recommend - though the Python style guide is silent on the matter [3].

[1]: https://google.github.io/styleguide/cppguide.html#General_Naming_Rules
[2]: https://google.github.io/styleguide/javaguide.html#s5.3-camel-case
[3]: https://google.github.io/styleguide/pyguide.html#316-naming

src/clusterfuzz/_internal/tests/core/swarming/api_test.py
```diff
@@ -0,0 +1,98 @@
...
+    self.mock_creds.token = 'fake_token'
+    self.mock.get_scoped_service_account_credentials.return_value = self.mock_creds
+
+    self.api = SwarmingAPI()
+
+  def test_push_task(self):
+    """Tests that push_task works as expected."""
+    task_request = swarming_pb2.NewTaskRequest(name='test_task')
+    self.api.push_task(task_request)
+
```
letitz, line 68: Let's mock a response from `post_url` and check `push_task()`'s return value.

src/clusterfuzz/_internal/tests/core/swarming/api_test.py
```diff
@@ -0,0 +1,98 @@
...
+    self.assertEqual(response, '{"count": 42}')
+
+  def test_push_task_no_config(self):
+    """Tests that push_task fails when config is missing."""
+    with mock.patch('clusterfuzz._internal.config.local_config.SwarmingConfig'
+                   ) as mock_config:
+      mock_config.side_effect = ValueError('Failed to load')
+      api = SwarmingAPI()
+      response = api.push_task(swarming_pb2.NewTaskRequest())
+      self.assertIsNone(response)
```
letitz: If you keep this design, you should check that no call was made to `post_url()`.

src/clusterfuzz/_internal/tests/core/swarming/api_test.py
```diff
@@ -0,0 +1,98 @@
...
+
+  def test_push_task_no_credentials(self):
+    """Tests that push_task fails when credentials are missing."""
+    self.mock.get_scoped_service_account_credentials.return_value = None
+    response = self.api.push_task(swarming_pb2.NewTaskRequest())
```
letitz: We should check that `post_url()` was not called.

src/clusterfuzz/_internal/tests/core/swarming/api_test.py
```diff
@@ -0,0 +1,98 @@
...
+    self.assertIsNone(response)
+
+  def test_count_tasks_no_credentials(self):
+    """Tests that count_tasks fails when credentials are missing."""
+    self.mock.get_scoped_service_account_credentials.return_value = None
+    response = self.api.count_tasks(swarming_pb2.TasksCountRequest())
```
letitz: Ditto

src/clusterfuzz/_internal/tests/core/swarming/service_test.py
```diff
@@ -27,7 +27,7 @@ class SwarmingServiceTest(unittest.TestCase):
   def setUp(self):
     helpers.patch(self, [
         'clusterfuzz._internal.swarming.is_swarming_task',
-        'clusterfuzz._internal.swarming.push_swarming_task',
+        'clusterfuzz._internal.swarming.service.SwarmingService._get_api',
```
letitz: If you define `SwarmingApi.create()`, you can mock that instead of a private helper.

src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
letitz: This test exercised a much more complex `NewTaskRequest`, consider adding some of that back to `api_test.py`.

letitz (changes requested):

src/clusterfuzz/_internal/swarming/__init__.py
IvanBM18: oops my bad, thanks for pointing it out :D

src/clusterfuzz/_internal/swarming/api.py
IvanBM18: Seems fair :D

src/clusterfuzz/_internal/swarming/api.py
IvanBM18: Sadly we dont have a debug log level in CF, only info+

For context: I tried adding it in once(in the `logs.py` module), but for some reason the debug logs were not shown in the GCP console, thought that's why we don't have it, seems that adding them requires a bigger effort than just adding a method that does `logs.(msg, level=Logs.DEBUG)`

src/clusterfuzz/_internal/swarming/api.py
IvanBM18: Thanks a ton for this review, its pretty insightful and i actually learned a ton from it :D. 
It makes sense for me to go ahead and "propagate" the problem to be exposed by the service.

But now that we mean to propagate issues, then i think its best if we revisit the exceptions too, so that  we: 
- Log and catch for specific exceptions in the service
- Unify the error catching for auth logic by letting the errors to be raised by the `requests` module instead

Also now the service will crash horribly with a ValueError (makes more sense than assertionError) if no config was found for envs that enable swarming

src/clusterfuzz/_internal/tests/core/swarming/api_test.py
IvanBM18: Now it will be called when no credentials were found this way HTTP error is raised and we can more easily debug the root cause for this from just reading the logs(we would see a 401 error)

src/clusterfuzz/_internal/tests/core/swarming/api_test.py
IvanBM18: Ditto +1 hahah
Now i will be called when no credentials were found this way HTTP error is raised and we can more easily debug the root cause for this from just reading the logs(we would see a HTTP 401 error)

src/clusterfuzz/_internal/tests/core/swarming/api_test.py
IvanBM18: No longer applies to code with changes

src/clusterfuzz/_internal/tests/core/swarming/api_test.py
IvanBM18: Sure :D

src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
IvanBM18: Sure just did that thanks for point it out :D

src/clusterfuzz/_internal/swarming/api.py
IvanBM18: Adding the last bit of changes(making the service crash at constructor) made the pr way bigger with changes to multiple test files, hence i created a separate feature branch for those specific changes to be reviewed in this [PR](https://github.com/google/clusterfuzz/pull/5285)

src/clusterfuzz/_internal/swarming/api.py
```diff
@@ -32,40 +35,59 @@
...
+      creds = credentials.get_scoped_service_account_credentials(
+          _SWARMING_SCOPES)
+      if not creds:
+        logs.error('[Swarming] Failed to get credentials. None found.')
+        return ""
+
+      if not creds.token:
+        creds.refresh(requests.Request())
+
+      return creds.token or ""
```
letitz: Can `creds.token` still be None after we call `refresh()`?

src/clusterfuzz/_internal/swarming/api.py
```diff
@@ -32,40 +35,59 @@
...
-      return {}
-
-    creds = credentials.get_scoped_service_account_credentials(_SWARMING_SCOPES)
-    if not creds:
-      logs.error('[Swarming] Failed to get credentials.')
-      return {}
-
-    if not creds.token:
-      creds.refresh(requests.Request())
+    token = self._get_token()
```
letitz, line 82: IIUC this means we'll make requests with no tokens if we fail to get a token, which is bound to fail? ISTM we should bail out and not make a request if we cannot get an auth token for it.

I would either expose to callers that `SwarmingApi` methods might raise either `GoogleAuthError` or `RequestException`, or wrap those in a new exception type here: `SwarmingError` and let the caller deal with that.

Edit: I see your comment on the test, that this will show up as 401 errors. I still find it a bit surprising to ask the system to do something we know will fail, but I must admit it works. I'll let @javanlacerda decide how he feels about it.

src/clusterfuzz/_internal/swarming/api.py
```diff
@@ -97,15 +127,14 @@ def push_task(self, task_request: swarming_pb2.NewTaskRequest) -> str | None:  #
...
     """
     message_body = json_format.MessageToJson(task_request)
-    logs.info(
-        f"[Swarming] Pushing task {task_request.name}",
-        url=self._base_url,
-        endpoint=_NEW_TASK_ENDPOINT,
-        body=message_body)
 
     response = self._make_request(_NEW_TASK_ENDPOINT, message_body)
     logs.info(
```
letitz: Nit: These logs can also go in `_make_request()`.

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -12,24 +12,23 @@
...
 class SwarmingService(remote_task_types.RemoteTaskInterface):
   """Remote task service implementation for Swarming."""
 
-  _api: SwarmingAPI = None
+  _api: SwarmingApi | None = None
 
-  def _get_api(self) -> SwarmingAPI:
-    """Returns the Swarming API instance."""
+  def __init__(self):
     if not self._api:
```
letitz: `self._api` is undefined at initialization time, surely? I think we can unconditionally call `SwarmingApi.create()`.

src/clusterfuzz/_internal/swarming/api.py
```diff
@@ -32,40 +35,59 @@
...
+    """Gets a valid token for the Swarming API.  Returns "" if it fails."""
+    try:
+      creds = credentials.get_scoped_service_account_credentials(
+          _SWARMING_SCOPES)
+      if not creds:
+        logs.error('[Swarming] Failed to get credentials. None found.')
+        return ""
+
+      if not creds.token:
+        creds.refresh(requests.Request())
```
letitz: I actually think this can be removed. `get_scoped_service_account_credentials()` already calls `refresh()` right before returning: https://github.com/google/clusterfuzz/blob/03cb7cd76842ccabd55c90802382dc0b677719b8/src/clusterfuzz/_internal/google_cloud_utils/credentials.py#L129

src/clusterfuzz/_internal/swarming/api.py
letitz: I would expect that debug-level logs are not logged in production, and only enabled if requested specifically somehow (environment variable?) during e.g. local testing. Anyway, thanks for considering it!

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -60,10 +59,17 @@ def create_utask_main_jobs(self,
           continue
         if request := swarming.create_new_task_request(
             task.command, task.job_type, task.argument):
-          self._get_api().push_task(request)
-      except Exception:  # pylint: disable=broad-except
+          self._api.push_task(request)
+      except HTTPError as api_failure:
         logs.error(
```
letitz: Should this be a warning? My mental model is usually:

* warning: an issue we can work around
* error: a bug that someone should look into and fix

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -60,10 +59,17 @@ def create_utask_main_jobs(self,
...
-            f'Failed to push task to Swarming: {task.command}, {task.job_type}.'
-        )
+            f'''Failed to push task to Swarming: {task.command}, {task.job_type}
+            . Reason: {api_failure}.
+            ''')
+        unscheduled_tasks.append(task)
+      except Exception as e:  # pylint: disable=broad-except
```
letitz: Do we still need this?

src/clusterfuzz/_internal/tests/core/swarming/api_test.py
```diff
@@ -36,12 +36,36 @@ def setUp(self):
     self.mock_creds.token = 'fake_token'
     self.mock.get_scoped_service_account_credentials.return_value = self.mock_creds
 
-    self.api = SwarmingAPI()
+    self.mock_config = mock.MagicMock()
+    self.mock_config.get.return_value = 'server-name'
+    self.api = SwarmingApi(self.mock_config)
```
letitz: Let's pass the "real" (testing yaml) swarming config? Or better yet, use `SwarmingApi.create()` here?

src/clusterfuzz/_internal/swarming/api.py
```diff
@@ -120,15 +149,14 @@ def count_tasks(self,
...
     """
     message_body = json_format.MessageToJson(count_request)
-    logs.info(
-        "[Swarming] Counting tasks in queue",
-        url=self._base_url,
-        endpoint=_COUNT_TASKS_ENDPOINT,
-        body=message_body)
 
     response = self._make_request(_COUNT_TASKS_ENDPOINT, message_body)
     logs.info('[Swarming] Response from CountTasks', response=response)
```
letitz: We should log the returned task ID, either here or in `SwarmingService`, in the main log message body. It will make debugging across both systems easier.

src/clusterfuzz/_internal/tests/core/swarming/api_test.py
IvanBM18: Sure i think not setting it up for quicker tests but i guess it makes sense for more robust tests :D

src/clusterfuzz/_internal/swarming/service.py
IvanBM18: MM no, its not needed i left it there because "the more we catch the merrier" hahah but i just read our internal readability doc and seems this is not recommended.. so ill delete it, thanks :D

src/clusterfuzz/_internal/swarming/service.py
IvanBM18: Mmm if the swarming requests start to fail i guess they are errors because a config is broken(probably) but, since they are not breaking CF i guess we can think of it as a warning

src/clusterfuzz/_internal/swarming/api.py
IvanBM18: Sure :D, 
i noticed we don't have the proto's for the swarming response in here, which are not needed but they indeed makes it easier for us to do operations with the json response, so for this purpose i opened yet another [PR](https://github.com/google/clusterfuzz/pull/5289) that adds the new proto and use it to log the taskId and for other edge case i just thought of.

src/clusterfuzz/_internal/tests/core/swarming/service_test.py
letitz, line 124: We should have a test for what happens when the API raises an exception we can catch.

src/clusterfuzz/_internal/swarming/api.py
letitz: Ok thanks! Fine with me to leave logging that field to a followup PR.

letitz (approved):
LGTM % adding one last test. Outstanding comments are left for @javanlacerda's consideration :)

javanlacerda (approved):

PR #5278, opened 2026-05-15
[BlackboxBenchmarking] Replace INTERVAL duration fields with DOUBLE duration_seconds

dylanjew:
F1 doesn't support INTERVAL types natively (b/213368390). This PR replaces the `INTERVAL` duration fields in `cluster-fuzz.fuzzer_stats.daily_stats` with `DOUBLE` duration_seconds.

### Rollout
I plan to just drop the existing `cluster-fuzz.fuzzer_stats.daily_stats` table after this is deployed and rerun the job to backfill the data. Since there are no consumers of this data yet, this seems simpler than doing a dual write + alter table drop column, then changing the job load to stop including the old rows.

The concern I had is that a BigQuery load without the old rows would fail, so dropping the table after this is deployed and then backfilling seems like a simpler approach.

### Testing
I verified the SQL query by running it in BigQuery against the `cluster-fuzz.fuzzer_stats.daily_stats` table.

I dropped the table in dev and ran the  cron with the new fields.  https://paste.googleplex.com/5110869211086848
https://screenshot.googleplex.com/7iTMbm3Ry6J22ZQ

### DIFF
--- a/src/clusterfuzz/_internal/cron/aggregate_fuzzer_stats.py
+++ b/src/clusterfuzz/_internal/cron/aggregate_fuzzer_stats.py
@@ -55,20 +55,20 @@ DAILY_STATS_SCHEMA = {
         'type': 'INTEGER',
         'mode': 'NULLABLE'
     }, {
-        'name': 'testcase_execution_duration',
-        'type': 'INTERVAL',
+        'name': 'testcase_execution_duration_seconds',
+        'type': 'FLOAT',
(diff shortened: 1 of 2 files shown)


aakallam (approved):

ViniciustCosta (approved):

dylanjew:
Thanks for the reviews! I dropped the staging and prod tables, so once this is deployed we should start writing the new schema

PR #5280, opened 2026-05-16
Optimize reboots with Android emulator initialization

jardondiego:
### Goal
Significantly improve the startup efficiency and fuzzing throughput of Android bots by eliminating redundant, time-consuming device reboots during the initialization and sanitizer setup phases.

### Background
Historically, Android bots suffered from excessive hard reboots (~60s penalty each) during task execution:
  1. Unconditional Final Reboots: initialize_device() unconditionally triggered a hard reboot at the end of its sequence to ensure a "good state," even if setup_asan_if_needed() had just performed a clean shell restart seconds prior.
  2. Sanitizer Option Reboots: Setting ASan options (sanitizer.set_options) writes to the /system partition. The helper function adb.write_data_to_file() contained a hardcoded rule forcing a reboot anytime a file in /system was modified.

### DIFF
--- a/src/clusterfuzz/_internal/platforms/android/adb.py
+++ b/src/clusterfuzz/_internal/platforms/android/adb.py
@@ -886,8 +886,9 @@ def write_data_to_file(contents, file_path, should_reboot=True):
   Args:
     contents: The string content to write.
     file_path: The path to the file on the Android device.
-    should_reboot: Whether to reboot the device if a system file is modified.
-        Defaults to True.
+    should_reboot: Whether to reboot the device after writing the file.
+        Only applies if `file_path` is under `/system`, since we need to
+        remount that partition as read-write first.
   """
   # If this is a file in /system, we need to remount /system as read-write and
   # after file is written, revert it back to read-only.
--- a/src/clusterfuzz/_internal/platforms/android/adb.py
+++ b/src/clusterfuzz/_internal/platforms/android/adb.py
@@ -881,7 +881,14 @@ def write_command_line_file(command_line, app_path):
 
 
 def write_data_to_file(contents, file_path, should_reboot=True):
-  """Writes content to file."""
+  """Writes content to file.
+  
+  Args:
+    contents: The string content to write.
+    file_path: The path to the file on the Android device.
+    should_reboot: Whether to reboot the device if a system file is modified.
+        Defaults to True.
+  """
   # If this is a file in /system, we need to remount /system as read-write and
   # after file is written, revert it back to read-only.
   is_system_file = file_path.startswith('/system')
--- a/src/clusterfuzz/_internal/platforms/android/adb.py
+++ b/src/clusterfuzz/_internal/platforms/android/adb.py
@@ -880,7 +880,7 @@ def write_command_line_file(command_line, app_path):
   write_data_to_file(command_line_file_contents, command_line_path)
 
 
-def write_data_to_file(contents, file_path):
+def write_data_to_file(contents, file_path, wait_for_reboot=True):
   """Writes content to file."""
   # If this is a file in /system, we need to remount /system as read-write and
   # after file is written, revert it back to read-only.
@@ -895,6 +895,6 @@ def write_data_to_file(contents, file_path):
   # Make command line file is readable.
   run_shell_command('chmod 0644 %s' % file_path, root=True)
 
-  if is_system_file:
+  if is_system_file and wait_for_reboot:
     reboot()
     wait_until_fully_booted()
--- a/src/clusterfuzz/_internal/platforms/android/adb.py
+++ b/src/clusterfuzz/_internal/platforms/android/adb.py
@@ -880,7 +880,7 @@ def write_command_line_file(command_line, app_path):
   write_data_to_file(command_line_file_contents, command_line_path)
 
 
-def write_data_to_file(contents, file_path, wait_for_reboot=True):
+def write_data_to_file(contents, file_path, should_reboot=True):
   """Writes content to file."""
   # If this is a file in /system, we need to remount /system as read-write and
   # after file is written, revert it back to read-only.
@@ -895,6 +895,6 @@ def write_data_to_file(contents, file_path, wait_for_reboot=True):
   # Make command line file is readable.
   run_shell_command('chmod 0644 %s' % file_path, root=True)
 
-  if is_system_file and wait_for_reboot:
+  if is_system_file and should_reboot:
     reboot()
     wait_until_fully_booted()
--- a/src/clusterfuzz/_internal/platforms/android/adb.py
+++ b/src/clusterfuzz/_internal/platforms/android/adb.py
@@ -903,6 +903,9 @@ def write_data_to_file(contents, file_path, should_reboot=True):
   run_shell_command('chmod 0644 %s' % file_path, root=True)
 
   if is_system_file:
-    if should_reboot or not environment.is_android_emulator():
+    if should_reboot:
       reboot()
       wait_until_fully_booted()
+    else:
+      # Manually revert /system to read-only since we aren't rebooting.
+      run_shell_command('mount -o ro,remount /system', root=True)
--- a/src/clusterfuzz/_internal/platforms/android/adb.py
+++ b/src/clusterfuzz/_internal/platforms/android/adb.py
@@ -902,6 +902,7 @@ def write_data_to_file(contents, file_path, should_reboot=True):
   # Make command line file is readable.
   run_shell_command('chmod 0644 %s' % file_path, root=True)
 
-  if is_system_file and should_reboot:
-    reboot()
-    wait_until_fully_booted()
+  if is_system_file:
+    if should_reboot or not environment.is_android_emulator():
+      reboot()
+      wait_until_fully_booted()
--- a/src/clusterfuzz/_internal/platforms/android/device.py
+++ b/src/clusterfuzz/_internal/platforms/android/device.py
@@ -183,9 +183,13 @@ def configure_device_settings():
   adb.write_data_to_file(local_properties_file_contents, LOCAL_PROP_PATH)
 
 
-def configure_system_build_properties():
+def configure_system_build_properties() -> bool:
   """Modifies system build properties in /system/build.prop for better boot
-  speed and power use."""
+  speed and power use.
+  
+  Returns:
+    True if the device was rebooted during configuration, False otherwise.
+  """
   adb.run_as_root()
 
   # Check md5 checksum of build.prop to see if already updated,
--- a/src/clusterfuzz/_internal/platforms/android/device.py
+++ b/src/clusterfuzz/_internal/platforms/android/device.py
@@ -195,9 +195,9 @@ def configure_system_build_properties():
   current_md5 = adb.get_file_checksum(BUILD_PROP_PATH)
   if current_md5 is None:
     logs.error('Unable to find %s on device.' % BUILD_PROP_PATH)
-    return
+    return False
   if old_md5 == current_md5:
-    return
+    return False
 
   # Pull to tmp file.
   bot_tmp_directory = environment.get_value('BOT_TMPDIR')
@@ -205,7 +205,7 @@ def configure_system_build_properties():
   adb.run_command(['pull', BUILD_PROP_PATH, old_build_prop_path])
   if not os.path.exists(old_build_prop_path):
     logs.error('Unable to fetch %s from device.' % BUILD_PROP_PATH)
-    return
+    return False
 
   # Write new build.prop.
   new_build_prop_path = os.path.join(bot_tmp_directory, 'new.prop')
@@ -228,11 +228,13 @@ def configure_system_build_properties():
 
   # Keep verified boot disabled for M and higher releases. This makes it easy
   # to modify system's app_process to load asan libraries.
+  reboot_done = False
   build_version = settings.get_build_version()
   if is_build_at_least(build_version, 'M'):
     adb.run_as_root()
     adb.run_command('disable-verity')
     reboot()
+    reboot_done = True
 
   # Make /system writable.
   adb.run_as_root()
@@ -254,6 +256,8 @@ def configure_system_build_properties():
   # Set persistent cache key containing and md5sum.
   current_md5 = adb.get_file_checksum(BUILD_PROP_PATH)
   persistent_cache.set_value(constants.BUILD_PROP_MD5_KEY, current_md5)
+
+  return reboot_done
 
 
 def get_debug_props_and_values():
@@ -306,16 +310,17 @@ def initialize_device():
   adb.setup_adb()
 
   # General device configuration settings.
-  configure_system_build_properties()
+  reboot_done = configure_system_build_properties()
   configure_device_settings()
   add_test_accounts_if_needed()
 
   # Setup AddressSanitizer if needed.
-  sanitizer.setup_asan_if_needed()
+  reboot_done = sanitizer.setup_asan_if_needed() or reboot_done
 
   # Reboot device as above steps would need it and also it brings device in a
   # good state.
-  reboot()
+  if not reboot_done:
+    reboot()
 
   # Make sure we are running as root after restart.
   adb.run_as_root()
--- a/src/clusterfuzz/_internal/platforms/android/device.py
+++ b/src/clusterfuzz/_internal/platforms/android/device.py
@@ -228,13 +228,11 @@ def configure_system_build_properties():
 
   # Keep verified boot disabled for M and higher releases. This makes it easy
   # to modify system's app_process to load asan libraries.
-  reboot_done = False
   build_version = settings.get_build_version()
   if is_build_at_least(build_version, 'M'):
     adb.run_as_root()
     adb.run_command('disable-verity')
     reboot()
-    reboot_done = True
 
   # Make /system writable.
   adb.run_as_root()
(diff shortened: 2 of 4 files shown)


src/clusterfuzz/_internal/platforms/android/device.py
IvanBM18, line 191: Nit:
Add typing hints and update the string doc for this method so that other devs reading this know what the bool means

src/clusterfuzz/_internal/platforms/android/device.py
```diff
@@ -255,6 +255,8 @@ def configure_system_build_properties():
   current_md5 = adb.get_file_checksum(BUILD_PROP_PATH)
   persistent_cache.set_value(constants.BUILD_PROP_MD5_KEY, current_md5)
 
+  return True
+
 
 def get_debug_props_and_values():
```
IvanBM18, line 264: Nit x2:
Add typing hints and update the string doc for this method so that other devs reading this know what the bool means

src/clusterfuzz/_internal/platforms/android/sanitizer.py
```diff
@@ -69,7 +69,9 @@ def set_options(sanitizer_tool_name, sanitizer_options):
   if not sanitizer_options_file_path:
     return
 
-  adb.write_data_to_file(sanitizer_options, sanitizer_options_file_path)
+  # Skip reboot as the app will pick up the options file on restart.
+  adb.write_data_to_file(
+      sanitizer_options, sanitizer_options_file_path, wait_for_reboot=False)
 
 
 def setup_asan_if_needed():
```
IvanBM18: Nit x3:
Add typing hints and update the string doc for this method so that other devs reading this know what the bool means

IvanBM18 (approved):
LGTM, minus some NITs

src/clusterfuzz/_internal/platforms/android/device.py
fernandofloresg: Agree with Ivan

fernandofloresg (approved):
lgtm, would be nice to add those docstrings

src/clusterfuzz/_internal/platforms/android/adb.py
```diff
@@ -880,7 +880,7 @@ def write_command_line_file(command_line, app_path):
   write_data_to_file(command_line_file_contents, command_line_path)
 
 
-def write_data_to_file(contents, file_path):
+def write_data_to_file(contents, file_path, wait_for_reboot=True):
```
letitz: We should document the new argument at least. While we're at it, we can document the whole function: args, return value, any exceptions it raises.

src/clusterfuzz/_internal/platforms/android/device.py
```diff
@@ -255,6 +259,8 @@ def configure_system_build_properties():
   current_md5 = adb.get_file_checksum(BUILD_PROP_PATH)
   persistent_cache.set_value(constants.BUILD_PROP_MD5_KEY, current_md5)
 
+  return True
```
letitz: If I'm reading this right, we only rebooted on line 239 if the android version was as least M?

src/clusterfuzz/_internal/platforms/android/device.py
```diff
@@ -306,16 +312,17 @@ def initialize_device():
   adb.setup_adb()
 
   # General device configuration settings.
-  configure_system_build_properties()
+  needs_full_reboot = configure_system_build_properties()
```
letitz: The docstring above states that this function returns whether the device was already rebooted, not whether it needed to be rebooted. Which is it?

src/clusterfuzz/_internal/platforms/android/device.py
```diff
@@ -306,16 +312,17 @@ def initialize_device():
...
   add_test_accounts_if_needed()
 
   # Setup AddressSanitizer if needed.
-  sanitizer.setup_asan_if_needed()
+  asan_reboot_done = sanitizer.setup_asan_if_needed()
 
   # Reboot device as above steps would need it and also it brings device in a
   # good state.
-  reboot()
+  if needs_full_reboot or not asan_reboot_done:
```
letitz: I'm having a hard time understanding the logic here, regardless of whether this boolean really means `needs_full_reboot` or if it means `already_rebooted`, per the above comment.

In the `needs_full_reboot` case: if we needed to reboot after configuring the device, and the asan setup step already rebooted it, then why do we reboot once more here? In other words, why is this not:

```suggestion
  if needs_full_reboot and not asan_reboot_done:
```

In the `already_rebooted` case: why do avoid rebooting if `already_rebooted` is true? I would expect instead:

```suggestion
  if not already_rebooted and not asan_reboot_done:
```

src/clusterfuzz/_internal/platforms/android/sanitizer.py
```diff
@@ -69,25 +69,31 @@ def set_options(sanitizer_tool_name, sanitizer_options):
   if not sanitizer_options_file_path:
     return
 
-  adb.write_data_to_file(sanitizer_options, sanitizer_options_file_path)
+  # Skip reboot as the app will pick up the options file on restart.
```
letitz, line 72: Just to make sure I understand, this means that the next time the app is started, it will read from the options file?

src/clusterfuzz/_internal/platforms/android/adb.py
letitz: Nit: `wait_for_reboot=False` suggests to me that we're rebooting the device but not waiting around for it to be full booted. Instead, we wish to express that the device should not be rebooted at all. How about `reboot=False`, or `should_reboot=False`?

src/clusterfuzz/_internal/tests/core/platforms/android/device_test.py
letitz: lol

src/clusterfuzz/_internal/tests/core/platforms/android/device_test.py
```diff
@@ -23,3 +26,79 @@ class InitializeEnvironmentTest(android_helpers.AndroidTest):
...
     """Ensure that initialize_environment throws no exceptions."""
     device.initialize_environment()
+
+
+class InitializeDeviceRebootLogicTest(unittest.TestCase):
+  """Tests the reboot batching logic in initialize_device."""
+
+  def setUp(self):
+    # Mock environment to bypass the engine fuzzer check
+    mock.patch(
```
letitz: This is the first time I see mocks written like this in ClusterFuzz. Historically I've seen use of `helpers.patch` [1] and `mock.patch` decorators [2]. In this case, I would expect the former.

[1]: https://github.com/google/clusterfuzz/blob/e82914eac6a550ea178ed85b07e37cb8987d76ea/src/clusterfuzz/_internal/tests/core/base/retry_test.py#L29
[2]: https://github.com/google/clusterfuzz/blob/e82914eac6a550ea178ed85b07e37cb8987d76ea/src/clusterfuzz/_internal/tests/core/k8s/k8s_integration_test.py#L32

src/clusterfuzz/_internal/tests/core/platforms/android/device_test.py
```diff
@@ -23,3 +26,79 @@ class InitializeEnvironmentTest(android_helpers.AndroidTest):
...
+    device.initialize_device()
+    self.mock_reboot.assert_not_called()
+
+  def test_reboot_if_clean_slate_needed(self):
+    """Test that it still reboots to ensure a clean state if no other steps restarted it."""
+    self.mock_config_props.return_value = False
+    self.mock_setup_asan.return_value = False
+
+    device.initialize_device()
+    self.mock_reboot.assert_called_once()
```
letitz: We test False/True, True/False and False/False. What about True/True? This goes back to my comment about `or not` vs `and not`.

src/clusterfuzz/_internal/tests/core/platforms/android/device_test.py
```diff
@@ -23,3 +26,79 @@ class InitializeEnvironmentTest(android_helpers.AndroidTest):
...
+    mock.patch('clusterfuzz._internal.platforms.android.ui.clear_notifications'
+              ).start()
+    mock.patch(
+        'clusterfuzz._internal.platforms.android.ui.unlock_screen').start()
+
+  def tearDown(self):
+    mock.patch.stopall()
+
+  def test_reboot_if_props_changed(self):
+    """Test that it reboots if build.prop changed, even if ASan didn't run."""
```
letitz:
```suggestion
    """Test that `initialize_device()` reboots the device if
    `configure_system_build_properties()` did, and the ASan setup script did not.
```

src/clusterfuzz/_internal/tests/core/platforms/android/device_test.py
```diff
@@ -23,3 +26,79 @@ class InitializeEnvironmentTest(android_helpers.AndroidTest):
...
+    self.mock_config_props.return_value = True
+    self.mock_setup_asan.return_value = False
+
+    device.initialize_device()
+    self.mock_reboot.assert_called_once()
+
+  def test_no_reboot_if_asan_ran(self):
+    """Test that the final reboot is skipped if ASan setup performed a shell restart."""
```
letitz:
```suggestion
    """Test that `initialize_device()` skips calling `reboot()` if
    `configure_system_build_properties()` did not cause a reboot but the ASan
    setup script did.
    """
```

src/clusterfuzz/_internal/tests/core/platforms/android/device_test.py
```diff
@@ -23,3 +26,79 @@ class InitializeEnvironmentTest(android_helpers.AndroidTest):
...
+    # If build.prop didn't change, the ASan restart handles the clean state.
+    self.mock_config_props.return_value = False
+    self.mock_setup_asan.return_value = True
+
+    device.initialize_device()
+    self.mock_reboot.assert_not_called()
+
+  def test_reboot_if_clean_slate_needed(self):
+    """Test that it still reboots to ensure a clean state if no other steps restarted it."""
```
letitz:
```suggestion
    """Test that `initialize_device()` calls `reboot()` if neither
    `configure_system_build_properties()` nor the ASan setup script did.
    """
```

letitz (changes requested):
I'm a bit confused about the main logic change in `initialize_device()`, see comment below.

src/clusterfuzz/_internal/platforms/android/adb.py
jardondiego: Fixed.

src/clusterfuzz/_internal/tests/core/platforms/android/device_test.py
jardondiego: Fixed.

src/clusterfuzz/_internal/tests/core/platforms/android/device_test.py
jardondiego: Fixed.

src/clusterfuzz/_internal/tests/core/platforms/android/device_test.py
jardondiego: Fixed.

src/clusterfuzz/_internal/tests/core/platforms/android/device_test.py
jardondiego: Updated the docstring.

src/clusterfuzz/_internal/platforms/android/adb.py
jardondiego: Fixed!

src/clusterfuzz/_internal/platforms/android/device.py
jardondiego: The code has changed substantially. Please review latest changes.

src/clusterfuzz/_internal/platforms/android/sanitizer.py
jardondiego: Yes. We write the ASan options to the system file. When the target fuzzer application is launched in a new process shortly after this, the new options file are read from disk during its initialization. A full device reboot isn't required for this step, so we can safely skip it to save time. Forcing a ~60s kernel reboot to update a text string that a process is going to read 2 seconds later is redundant.

src/clusterfuzz/_internal/platforms/android/device.py
jardondiego: Sorry for the confusion. I think the code wasn't in great shape, I changed it significantly. Thanks for taking another look.

src/clusterfuzz/_internal/platforms/android/adb.py
```diff
@@ -880,8 +880,15 @@ def write_command_line_file(command_line, app_path):
...
-def write_data_to_file(contents, file_path):
-  """Writes content to file."""
+def write_data_to_file(contents, file_path, should_reboot=True):
+  """Writes content to file.
+  
+  Args:
+    contents: The string content to write.
+    file_path: The path to the file on the Android device.
+    should_reboot: Whether to reboot the device if a system file is modified.
+        Defaults to True.
```
letitz: Let's explain what a system file is, and no need to repeat the default argument value since that's self-documenting:


```suggestion
    should_reboot: Whether to reboot the device after writing the file.
        Only applies if `file_path` is under `/system`, since we need to
        remount that partition as read-write first.
```

src/clusterfuzz/_internal/platforms/android/sanitizer.py
letitz: I see, thanks. IIUC the logic in `write_data_to_file()`, that means we'll leave `/system` in read-write mode. Is that intentional?

src/clusterfuzz/_internal/tests/core/platforms/android/device_test.py
```diff
@@ -13,13 +13,55 @@
...
   def test(self):
     """Ensure that initialize_environment throws no exceptions."""
     device.initialize_environment()
+
+
+class InitializeDeviceRebootLogicTest(unittest.TestCase):
+  """Tests the reboot batching logic in initialize_device."""
+
+  def setUp(self):
+    from clusterfuzz._internal.tests.test_libs import helpers
```
letitz: Let's move this to the top level, no need to make this import conditional on the test being run?

letitz (commented):
One substantive comment around skipping reboots after writing ASan options, otherwise just two small nits.

src/clusterfuzz/_internal/platforms/android/adb.py
jardondiego: Done!

src/clusterfuzz/_internal/tests/core/platforms/android/device_test.py
jardondiego: Good point. It should be above.

src/clusterfuzz/_internal/platforms/android/sanitizer.py
jardondiego: Yes. I don't think it would be a problem for emulated devices since they would be launched and destroyed in every task bit it might be an issue for non-virtual devices since a fuzzing job could break a system file and we would have to address that issue in a physical device, and that could mean having to physically check such device. We could either skip this optimization for virtual devices or manually trigger a read-only re-mounting. What do you suggest? 

- Option A: https://github.com/google/clusterfuzz/commit/ea78515760b799328966ba14a328350a1fe190ed
- Option B: https://github.com/google/clusterfuzz/commit/ccffa147ea91615052e512aa83900d8f3ec99276

jardondiego: I applied Option A for now, but I'm all ears for any suggestions from you. Thanks!

src/clusterfuzz/_internal/platforms/android/adb.py
letitz: I don't see the edit! I think it got lost.

src/clusterfuzz/_internal/platforms/android/sanitizer.py
```diff
@@ -69,25 +69,31 @@ def set_options(sanitizer_tool_name, sanitizer_options):
...
+  adb.write_data_to_file(
+      sanitizer_options, sanitizer_options_file_path, should_reboot=False)
```
letitz, line 74: I don't see the change you mentioned as option A. This seems to me like it will always skip rebooting.

src/clusterfuzz/_internal/platforms/android/sanitizer.py
letitz: See other comment, I don't think option A was applied in the end.

In any case, I don't think it's a big deal to lose 1m per fuzzing session - they last 3 hours anyway. I would rather go for option B and avoid the risk of weird issues where /system files get corrupted. I agree with you that for emulated devices that should be safe, but I'm not 100% confident :)

letitz (approved):
LGTM % option b and actually fixing the docstring for `write_data_to_file()`.

src/clusterfuzz/_internal/platforms/android/sanitizer.py
jardondiego: I think something weird happened with my workspace, I was moving around multiple branches and it somehow got lost. I will fix it!

src/clusterfuzz/_internal/platforms/android/sanitizer.py
jardondiego: I think something weird happened in my workspace, I was moving around several branches and got confused.   Agreed! I will adjust as suggested!

src/clusterfuzz/_internal/platforms/android/adb.py
jardondiego: I somehow messed up when moving around several branches, is is now fixed.

javanlacerda (approved):

PR #5282, opened 2026-05-18
[Swarming] Push preprocess tasks to swarming queue

IvanBM18:
## Overview

This change adds support for scheduling tasks to the new Swarming backend. Because Swarming uses a different execution model and to be able to later account for `backpressure` it requires its own separate `preprocess` queue and a much lower default target size (5) to prevent unbounded task queuing. 

By refactoring the cron scheduling logic, we can now simultaneously feed both the Swarming and Batch environments at their respective ideal rates.

## Changes

- Add a distinct `SWARMING_PREPROCESS_TARGET_SIZE_DEFAULT` set to `5` to support the Swarming backend's task capacity needs.
- Define Swarming-specific queue mappings (`SWARMING_QUEUES`).
- Refactored the `scheduler_fuzz` so that the schedulers are composed of multiple helper functions to avoid complex inheritance:
  - Renamed them to Providers, as they no longer schedule tasks, they only look for them
  - Create a DTO called `Queue` that contains information regarding the queues
- Update `ChromeFuzzTaskScheduler` to independently schedule Swarming tasks  alongside standard Batch tasks.
  - Now it also looks for Android jobs as well
- Update `schedule_fuzz_test.py` to match the updated scheduler class signatures.

## Tests performed
This changes have been living for 2 days in dev and seems the cron job is scheduling fuzz tasks successfully:
image

Also the fuzz task hours seem healthy as well
image

Note:
The spike in fuzz hours corresponds to the spike in more scheduling being performed

The only thing thats weird to me(not related to this PR btw) is that it seems that the cron job runs once a day in dev, when the config in dev says it should run every 10 min, maybe this is because of the limited quota?

#### TODO

- `src/clusterfuzz/_internal/base/feature_flags.py`: Update this value based off dev & stage metrics and tests.

### DIFF
--- a/src/clusterfuzz/_internal/base/feature_flags.py
+++ b/src/clusterfuzz/_internal/base/feature_flags.py
@@ -36,6 +36,8 @@ class FeatureFlags(Enum):
   PREPROCESS_QUEUE_SIZE_LIMIT = 'preprocess_queue_size_limit'
 
   SWARMING_REMOTE_EXECUTION = 'swarming_remote_execution'
+  # TODO(ibarba): Set this value based off dev & stage metrics and tests.
+  SWARMING_PREPROCESS_QUEUE_SIZE_LIMIT = 'swarming_preprocess_queue_size_limit'
 
   @property
   def flag(self):
--- a/src/clusterfuzz/_internal/base/feature_flags.py
+++ b/src/clusterfuzz/_internal/base/feature_flags.py
@@ -36,7 +36,7 @@ class FeatureFlags(Enum):
   PREPROCESS_QUEUE_SIZE_LIMIT = 'preprocess_queue_size_limit'
 
   SWARMING_REMOTE_EXECUTION = 'swarming_remote_execution'
-  # TODO(ibarba): Set this value based off dev & stage metrics and tests.
+  # TODO(b/516630567): Set this value based off dev & stage metrics and tests.
   SWARMING_PREPROCESS_QUEUE_SIZE_LIMIT = 'swarming_preprocess_queue_size_limit'
 
   @property
--- a/src/clusterfuzz/_internal/cron/schedule_fuzz.py
+++ b/src/clusterfuzz/_internal/cron/schedule_fuzz.py
@@ -13,22 +13,25 @@
 # limitations under the License.
 """Cron job to schedule fuzz tasks that run on batch."""
 
+from abc import ABC
+from abc import abstractmethod
 import collections
 import random
 import time
 
 from google.cloud import monitoring_v3
 
-from clusterfuzz._internal.base import feature_flags
 from clusterfuzz._internal.base import memoize
 from clusterfuzz._internal.base import tasks
 from clusterfuzz._internal.base import utils
+from clusterfuzz._internal.base.feature_flags import FeatureFlags
 from clusterfuzz._internal.datastore import data_types
 from clusterfuzz._internal.datastore import ndb_utils
 from clusterfuzz._internal.google_cloud_utils import credentials
 from clusterfuzz._internal.metrics import logs
 
 PREPROCESS_TARGET_SIZE_DEFAULT = 10000
+SWARMING_PREPROCESS_TARGET_SIZE_DEFAULT = 5
 
 
 @memoize.wrap(memoize.InMemory(60))
@@ -62,14 +65,54 @@ def get_queue_size(creds, project_id, subscription_id):
   return 0
 
 
-class BaseFuzzTaskScheduler:
+class BaseFuzzTaskScheduler(ABC):
   """Base fuzz task scheduler for any deployment of ClusterFuzz."""
 
-  def __init__(self, num_tasks):
-    self.num_tasks = num_tasks
-
-  def get_fuzz_tasks(self):
+  @abstractmethod
+  def get_fuzz_tasks(self, num_tasks: int) -> list[tasks.Task]:
     raise NotImplementedError('Child class must implement.')
+
+  def schedule_fuzz_tasks(self) -> bool:
+    """Schedules fuzz tasks."""
+    return self._schedule_fuzz_tasks()
+
+  def _schedule_fuzz_tasks(
+      self,
+      queue: str = tasks.PREPROCESS_QUEUE,
+      default_target_size: int = PREPROCESS_TARGET_SIZE_DEFAULT,
+      target_size_flag: FeatureFlags = FeatureFlags.PREPROCESS_QUEUE_SIZE_LIMIT
+  ) -> bool:
+    """Internal method to schedule fuzz tasks."""
+    project = utils.get_application_id()
+    start = time.time()
+    creds = credentials.get_default()[0]
+    preprocess_queue_size = get_queue_size(creds, project, queue)
+
+    target_size = default_target_size
+    if target_size_flag.enabled and target_size_flag.content:
+      target_size = int(target_size_flag.content)
+
+    num_tasks = target_size - preprocess_queue_size
+    logs.info(f'Queue {queue} size: {preprocess_queue_size}. '
+              f'Target: {target_size}. Needed: {num_tasks}.')
+
+    if num_tasks <= 0:
+      logs.info('Queue size met or exceeded. Not scheduling tasks.')
+      return False
+
+    fuzz_tasks = self.get_fuzz_tasks(num_tasks)
+    if not fuzz_tasks:
+      logs.error('No fuzz tasks found to schedule.')
+      return False
+
+    logs.info(f'Adding {len(fuzz_tasks)} tasks to queue {queue}.')
+    tasks.bulk_add_tasks(fuzz_tasks, queue=queue, eta_now=True)
+    logs.info(f'Scheduled {len(fuzz_tasks)} tasks on queue {queue}.')
+
+    end = time.time()
+    total = end - start
+    logs.info(f'Task scheduling took {total} seconds.')
+    return True
 
 
 class FuzzTaskCandidate:
@@ -101,7 +144,7 @@ class FuzzTaskCandidate:
 class OssfuzzFuzzTaskScheduler(BaseFuzzTaskScheduler):
   """Fuzz task scheduler for OSS-Fuzz."""
 
-  def get_fuzz_tasks(self) -> list[tasks.Task]:
+  def get_fuzz_tasks(self, num_tasks: int) -> list[tasks.Task]:
     # TODO(metzman): Handle high end.
     # A job's weight is determined by its own weight and the weight of the
     # project is a part of. First get project weights.
@@ -164,11 +207,9 @@ class OssfuzzFuzzTaskScheduler(BaseFuzzTaskScheduler):
     for fuzz_task_candidate in fuzz_task_candidates:
       weights.append(fuzz_task_candidate.weight)
 
-    fuzz_tasks_count = self.num_tasks
-    logs.info(f'Scheduling {fuzz_tasks_count} fuzz tasks for OSS-Fuzz.')
+    logs.info(f'Scheduling {num_tasks} fuzz tasks for OSS-Fuzz.')
 
-    choices = random.choices(
-        fuzz_task_candidates, weights=weights, k=fuzz_tasks_count)
+    choices = random.choices(fuzz_task_candidates, weights=weights, k=num_tasks)
     fuzz_tasks = [
         tasks.Task(
             'fuzz',
@@ -186,7 +227,7 @@ class OssfuzzFuzzTaskScheduler(BaseFuzzTaskScheduler):
 class ChromeFuzzTaskScheduler(BaseFuzzTaskScheduler):
   """Fuzz task scheduler for Chrome."""
 
-  def get_fuzz_tasks(self) -> list[tasks.Task]:
+  def get_fuzz_tasks(self, num_tasks: int) -> list[tasks.Task]:
     """Returns fuzz tasks for chrome, weighted by job weight."""
     logs.info('Getting jobs for Chrome.')
 
@@ -214,14 +255,12 @@ class ChromeFuzzTaskScheduler(BaseFuzzTaskScheduler):
       fuzz_task_candidates.append(fuzz_task_candidate)
 
     weights = [candidate.weight for candidate in fuzz_task_candidates]
-    fuzz_tasks_count = self.num_tasks
-    logs.info(f'Scheduling {fuzz_tasks_count} fuzz tasks for Chrome.')
+    logs.info(f'Scheduling {num_tasks} fuzz tasks for Chrome.')
 
     if not fuzz_task_candidates:
       return []
 
-    choices = random.choices(
-        fuzz_task_candidates, weights=weights, k=fuzz_tasks_count)
+    choices = random.choices(fuzz_task_candidates, weights=weights, k=num_tasks)
     fuzz_tasks = [
         tasks.Task(
             'fuzz',
@@ -232,51 +271,28 @@ class ChromeFuzzTaskScheduler(BaseFuzzTaskScheduler):
     ]
     return fuzz_tasks
 
+  def _schedule_swarming_fuzz_tasks(self) -> bool:
+    if not FeatureFlags.SWARMING_REMOTE_EXECUTION.enabled:
+      return False
 
-def get_fuzz_tasks(num_tasks: int) -> list[tasks.Task]:
-  if utils.is_oss_fuzz():
-    scheduler = OssfuzzFuzzTaskScheduler(num_tasks)
-  else:
-    scheduler = ChromeFuzzTaskScheduler(num_tasks)
-  fuzz_tasks = scheduler.get_fuzz_tasks()
-  return fuzz_tasks
+    swarming_preprocess_queue = tasks.SWARMING_QUEUES[tasks.PREPROCESS_QUEUE]
+    return self._schedule_fuzz_tasks(
+        queue=swarming_preprocess_queue,
+        default_target_size=SWARMING_PREPROCESS_TARGET_SIZE_DEFAULT,
+        target_size_flag=FeatureFlags.SWARMING_PREPROCESS_QUEUE_SIZE_LIMIT)
 
+  def _schedule_batch_fuzz_tasks(self) -> bool:
+    return self._schedule_fuzz_tasks()
 
-def schedule_fuzz_tasks() -> bool:
-  """Schedules fuzz tasks."""
-
-  project = utils.get_application_id()
-  start = time.time()
-  creds = credentials.get_default()[0]
-  preprocess_queue_size = get_queue_size(creds, project, tasks.PREPROCESS_QUEUE)
-
-  target_size = PREPROCESS_TARGET_SIZE_DEFAULT
-  target_size_flag = feature_flags.FeatureFlags.PREPROCESS_QUEUE_SIZE_LIMIT
-  if target_size_flag.enabled and target_size_flag.content:
-    target_size = int(target_size_flag.content)
-
-  num_tasks = target_size - preprocess_queue_size
-  logs.info(f'Preprocess queue size: {preprocess_queue_size}. '
-            f'Target: {target_size}. Needed: {num_tasks}.')
-
-  if num_tasks <= 0:
-    logs.info('Queue size met or exceeded. Not scheduling tasks.')
-    return False
-
-  fuzz_tasks = get_fuzz_tasks(num_tasks)
-  if not fuzz_tasks:
-    logs.error('No fuzz tasks found to schedule.')
-    return False
-
-  logs.info(f'Adding {fuzz_tasks} to preprocess queue.')
-  tasks.bulk_add_tasks(fuzz_tasks, queue=tasks.PREPROCESS_QUEUE, eta_now=True)
-  logs.info(f'Scheduled {len(fuzz_tasks)} fuzz tasks.')
-
-  end = time.time()
-  total = end - start
-  logs.info(f'Task scheduling took {total} seconds.')
-  return True
+  def schedule_fuzz_tasks(self) -> bool:
+    self._schedule_swarming_fuzz_tasks()
+    self._schedule_batch_fuzz_tasks()
+    return True
 
 
 def main():
-  return schedule_fuzz_tasks()
+  if utils.is_oss_fuzz():
+    scheduler = OssfuzzFuzzTaskScheduler()
+  else:
+    scheduler = ChromeFuzzTaskScheduler()
+  return scheduler.schedule_fuzz_tasks()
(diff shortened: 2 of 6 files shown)


src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -13,22 +13,25 @@
...
 from clusterfuzz._internal.base import tasks
 from clusterfuzz._internal.base import utils
+from clusterfuzz._internal.base.feature_flags import FeatureFlags
 from clusterfuzz._internal.datastore import data_types
 from clusterfuzz._internal.datastore import ndb_utils
 from clusterfuzz._internal.google_cloud_utils import credentials
 from clusterfuzz._internal.metrics import logs
 
 PREPROCESS_TARGET_SIZE_DEFAULT = 10000
+SWARMING_PREPROCESS_TARGET_SIZE_DEFAULT = 5
```
IvanBM18: The Swarming pool has a hard limit of 25 (LINUX) bots running 1 task each.
- At average 1 hour per fuzzing/swarming task, those 25 bots can finish 4 to 5 tasks every 10 minutes (the interval the cron job runs)
-  Because the 2,000(in prod) `preprocess` tworkers almost instantly process the preprocess queue, the target size acts as an injection rate more than a buffer.

So, Injecting 5 tasks every 10 minutes matches the expected Swarming rate, preventing an infinitely growing backlog of stale tasks. This is still the default value, the real value is managed trough a feature flag, we will later tweak this feature flag based on metrics & how swarming handled this workload, so that we have a more acqurate value

src/clusterfuzz/_internal/base/feature_flags.py
```diff
@@ -36,6 +36,8 @@ class FeatureFlags(Enum):
   PREPROCESS_QUEUE_SIZE_LIMIT = 'preprocess_queue_size_limit'
 
   SWARMING_REMOTE_EXECUTION = 'swarming_remote_execution'
+  # TODO(ibarba): Set this value based off dev & stage metrics and tests.
```
fernandofloresg: is this still true? if this is going to master then is going to stage and prod ?

fernandofloresg (approved):
lgtm just had one question

src/clusterfuzz/_internal/base/feature_flags.py
IvanBM18: Yes this is still true, right now no swarming related code executes outside of dev, we have a ton of featureFlags in place for this reason. So when this changes reach stage/prod they are going to be safe.

letitz: I would rather tweak this kind of knob in the yaml config rather than in live datastore configs, since it is less prone to mutating unexpectedly without leaving an audit trail. I see however how this must be much more convenient for use in dev and staging. Up to you!

letitz: TODO should reference a bug link instead.

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -186,13 +227,14 @@ def get_fuzz_tasks(self) -> list[tasks.Task]:
...
+  def get_fuzz_tasks(self, num_tasks: int) -> list[tasks.Task]:
     """Returns fuzz tasks for chrome, weighted by job weight."""
     logs.info('Getting jobs for Chrome.')
 
     candidates_by_job = {}
-    # Only consider linux jobs for chrome fuzzing.
-    job_query = data_types.Job.query(data_types.Job.platform == 'LINUX')
+    # Only consider LINUX or ANDROID jobs
+    job_query = data_types.Job.query(
+        data_types.Job.platform.IN(['LINUX', 'ANDROID']))
```
letitz: This will not work. It will randomly (and mostly) return `LINUX` jobs for swarming, and also randomly (though rarely) `ANDROID` jobs for Batch.

Also, not all `ANDROID` jobs can run on Swarming - only those that define `SWARMING_DIMENSIONS` can. 

Combined with my comment above, here is a suggestion for how to restructure this:

1. Package queue details in a new dataclass:

```py
@dataclass
class Queue:
  name: str
  default_target_size: int
  target_size_flag: FeatureFlags

_DEFAULT_QUEUE = Queue(
    name = tasks.PREPROCESS_QUEUE,
    default_target_size = PREPROCESS_TARGET_SIZE_DEFAULT,
    target_size_flag = FeatureFlags.PREPROCESS_QUEUE_SIZE_LIMIT,
)

_SWARMING_QUEUE = Queue(
   name = tasks.SWARMING_QUEUES[tasks.PREPROCESS_QUEUE]
   default_target_size = SWARMING_PREPROCESS_TARGET_SIZE_DEFAULT,
   target_size_flag=FeatureFlags.SWARMING_PREPROCESS_QUEUE_SIZE_LIMIT,
)
```

2. Move `BaseFuzzTaskScheduler._schedule_fuzz_tasks` to a private helper (optional: break it up), and give it a couple new arguments:

```py
def _remaining_queue_capacity(queue: _Queue) -> int:
  # query the queue size, target size
  return target_size - preprocess_queue_size

def _fill_queue(queue: Queue, scheduler: BaseFuzzTaskScheduler):
  # measure start time

  num_tasks = _remaining_queue_capacity(queue)
  if num_tasks <= 0:
    return

  fuzz_tasks = scheduler.get_fuzz_tasks(num_tasks)
  # enqueue the tasks, log as needed
```

3. Allow injecting the set of jobs considered for execution into `ChromeFuzzTaskScheduler`:

```py
def _get_jobs_for_platform(platform: str):
    return ndb_utils.get_all_from_query(
      data_types.Job.query(data_types.Job.platform == self._platform))

def _get_swarming_jobs(self):
  android_jobs = _get_jobs_for_platform('ANDROID')
  return [
    job for job in android_jobs
    if 'SWARMING_DIMENSIONS' in job.get_environment()
  ]

class ChromeFuzzTaskScheduler(BaseFuzzTaskScheduler):
  _candidates: list[FuzzTaskCandidate]

  def __init__(self, jobs: list[data_types.Job]):
    self._candidates = _jobs_to_candidates(jobs)  # calculates weights etc.

  def get_fuzz_tasks(self):
    # perform random choice over `self._candidates`
```

4. Reintroduce the top-level `schedule_fuzz_tasks()` function:

```py
def schedule_chrome_fuzz_tasks():
  default_scheduler = ChromeFuzzTaskScheduler(_get_jobs_for_platform('LINUX'))
  _fill_queue(_DEFAULT_QUEUE, default_scheduler)
  
  if not swarming_enabled:
    return

  swarming_scheduler = ChromeFuzzTaskScheduler(_get_swarming_jobs())
  _fill_queue(_SWARMING_QUEUE, swarming_scheduler)

def schedule_fuzz_tasks():
  if utils.is_oss_fuzz():
    _fill_queue(_DEFAULT_QUEUE, OssfuzzFuzzTaskScheduler())
  else:
    schedule_chrome_fuzz_tasks()
```

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -62,15 +65,55 @@ def get_queue_size(creds, project_id, subscription_id):
...
 
-  def __init__(self, num_tasks):
-    self.num_tasks = num_tasks
-
-  def get_fuzz_tasks(self):
+  @abstractmethod
+  def get_fuzz_tasks(self, num_tasks: int) -> list[tasks.Task]:
     raise NotImplementedError('Child class must implement.')
 
+  def schedule_fuzz_tasks(self) -> bool:
```
letitz: You're going from a pure abstract base class (an interface, in Java terms) to a partially-concrete base class, which is worse. Prefer composition over inheritance as a rule, as it makes things usually easier to work with and understand.

In any case, `get_fuzz_tasks` needs an additional argument that is only useful for Chrome, so I don't think this is the right interface. See comment below for a suggestion on how to restructure this.

src/clusterfuzz/_internal/cron/schedule_fuzz.py
letitz: I think we should prevent the infinitely-growing queue of tasks using `CountTasks` and a pretty low limit on the `utask_main` queue size, such that we can aim for a bit more saturation here (e.g. 10 tasks, such that we never expect the queue to be empty). This works for now though.

letitz (changes requested):
One big comment about reworking the main logic here.

src/clusterfuzz/_internal/base/feature_flags.py
IvanBM18: I also agree its best to link to a bug, i previously asked gemini if we can add any go link or bug references in open source code/reviews in github, basically it said:
```
Throughout the repos you maintain (READMEs, code comments, etc.): Don't put any Google internal info, including go/ and b/ links. If you're sure the referenced content can be public, move it to GitHub issues or README or Wiki
```
Although im aware ai can allucinate and im not a open source expert, but as far as i know, thats the reason for this repo be full of `#TODO(metzman):` comments

So your call, can we include bug links in here?

IvanBM18: I like the idea of having a record/history of how the value for this remote flag has changed, lets come back to this once we are using swarming for prod jobs :D

letitz: @javanlacerda or @ViniciustCosta would know I'm sure!

src/clusterfuzz/_internal/cron/schedule_fuzz.py
IvanBM18: Uff this is a big refactor, I chose inheritance because it looked cleaner but agree  that in the long run composition is king. 

Also i renamed the classes since now more of schedulers they feel like generators/providers

Thanks a ton for your feedback :D

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -14,21 +14,101 @@
...
+  """Returns all jobs for the given platform."""
+  return ndb_utils.get_all_from_query(
+      data_types.Job.query(data_types.Job.platform == platform))
+
+
+def _get_swarming_jobs():
+  """Returns all jobs that have swarming environment variables."""
+  jobs = []
+  jobs.extend(_get_jobs_for_platform('ANDROID'))
+  jobs.extend(_get_jobs_for_platform('LINUX'))
```
IvanBM18: Note: We can also fuzz linux jobs in swarming, hence adding them as a possible option here

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -14,21 +14,101 @@
...
 from clusterfuzz._internal.datastore import data_types
 from clusterfuzz._internal.datastore import ndb_utils
 from clusterfuzz._internal.google_cloud_utils import credentials
 from clusterfuzz._internal.metrics import logs
 
 PREPROCESS_TARGET_SIZE_DEFAULT = 10000
+SWARMING_PREPROCESS_TARGET_SIZE_DEFAULT = 5
+
+
+@dataclass
```
letitz, line 39: These need comments, if minimal.

src/clusterfuzz/_internal/cron/schedule_fuzz.py
letitz: We don't need it now, so I would be inclined to leave a TODO instead. Fine to include if you want. In that case, we can query for both platforms at once:

```py
def _get_jobs_for_platforms(platforms: list[str]) -> list[data_types.Job]:
  """Returns all jobs for the given platforms."""
  return ndb_utils.get_all_from_query(
      data_types.Job.query(data_types.Job.platform.IN(platforms)))


def _get_swarming_jobs():
  """Returns all jobs that have swarming environment variables."""
  jobs = _get_jobs_for_platforms(['ANDROID', 'LINUX'])
```

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -14,21 +14,101 @@
...
+    target_size = int(queue.target_size_flag.content)
+
+  num_tasks = target_size - preprocess_queue_size
+  logs.info(f'Queue {queue.name} size: {preprocess_queue_size}. '
+            f'Target: {target_size}. Needed: {num_tasks}.')
+
+  return num_tasks
+
+
+def _fill_queue(queue: Queue, provider: 'BaseFuzzTaskProvider'):
```
letitz: Let's define this below `BaseFuzzTaskProvider` so we can reference the type without using a string.

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -62,13 +142,10 @@ def get_queue_size(creds, project_id, subscription_id):
   return 0
 
 
-class BaseFuzzTaskScheduler:
-  """Base fuzz task scheduler for any deployment of ClusterFuzz."""
+class BaseFuzzTaskProvider():
```
letitz: I liked using `ABC` and `@abstractmethod`, it removes the need to raise `NotImplementedError` by hand and catches partial implementations at instantiation time instead of when the method is called. Let's restore it.

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -183,23 +258,26 @@ def get_fuzz_tasks(self) -> list[tasks.Task]:
...
+  _candidates: list[FuzzTaskCandidate]
+
+  def __init__(self, jobs: list[data_types.Job]):
+    self._candidates = [
+        FuzzTaskCandidate(
+            job=job.name,
+            project=job.project,
+            base_os_version=job.base_os_version)
+        for job in jobs
+        if job.platform in ['LINUX', 'ANDROID']
```
letitz: I don't think we should filter the jobs here. The instance should consider all jobs that it's given as fair game.

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -183,23 +258,26 @@ def get_fuzz_tasks(self) -> list[tasks.Task]:
...
-class ChromeFuzzTaskScheduler(BaseFuzzTaskScheduler):
-  """Fuzz task scheduler for Chrome."""
+class ChromeFuzzTaskProvider(BaseFuzzTaskProvider):
+  """Fuzz task provider for Chrome."""
+
+  _candidates: list[FuzzTaskCandidate]
+
+  def __init__(self, jobs: list[data_types.Job]):
+    self._candidates = [
+        FuzzTaskCandidate(
```
letitz: These are incomplete candidates with missing `fuzzer`, `weight` fields, which is not well reflected in the type system. Let's move the logic to query the weights from the database here (or in a helper method, more likely) and stick a fully-formed list of candidates in `self._candidates`.

src/clusterfuzz/_internal/swarming/__init__.py
```diff
@@ -37,6 +37,14 @@
 ]
 
 
+def has_swarming_env_vars(job_environment: dict) -> bool:
+  """Returns True if the job environment contains swarming env vars."""
+  return bool(
+      job_environment and
```
letitz: The function signature says that `job_environment` will always be a dict, so I don't think we need to check for `None`. Is this a check for emptiness?

src/clusterfuzz/_internal/swarming/__init__.py
```diff
@@ -37,6 +37,14 @@
...
+      (utils.string_is_true(job_environment.get('IS_SWARMING_JOB')) or
```
letitz: Do we use `IS_SWARMING_JOB` anywhere? Can we remove it for simplicity? (this can be in a followup)

src/clusterfuzz/_internal/tests/appengine/handlers/cron/schedule_fuzz_test.py
```diff
@@ -232,3 +233,20 @@ def test_os_version_job_without_version(self):
     self._setup_chrome_entities()
     task = self._run_and_get_task()
     self.assertIsNone(task.extra_info.get('base_os_version'))
+
+  def test_job_filtering(self):
+    """Tests that jobs are filtered by platform"""
```
letitz: Mentioned above, I don't think this should be `ChromeFuzzTaskProvider`'s job.

src/clusterfuzz/_internal/tests/core/swarming/swarming_test.py
```diff
@@ -50,6 +50,40 @@ def setUp(self):
...
+
+  def test_has_swarming_env_vars_swarming_dimensions(self):
+    """Tests that SWARMING_DIMENSIONS returns True."""
+    self.assertTrue(
+        swarming.has_swarming_env_vars({
+            'SWARMING_DIMENSIONS': 'os:linux'
+        }))
+
+  def test_has_swarming_env_vars_both(self):
+    """Tests that both variables return True."""
```
letitz, line 87: What about `IS_SWARMING_JOB = False`, `SWARMING_DIMENSIONS = {}`?

letitz (commented):
Looking good, a bunch of small comments left to address.

src/clusterfuzz/_internal/swarming/__init__.py
IvanBM18: Yes we do, each time we define a new job we don't necessarily need specific swarming dimensions for it, so for this cases we have been using `IS_SWARMING_JOB` instead.

src/clusterfuzz/_internal/swarming/__init__.py
IvanBM18: Yes, this was a check for the rare case in which the job_environment is None

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -36,6 +38,7 @@
 
 @dataclass
 class Queue:
+  """Data class that holds information about a pub/subqueue."""
   name: str
```
letitz, line 49: I was more meaning that the fields should be documented - what does each store? Apart from the "pub/sub" reference, this docstring tells me nothing I didn't know by reading the class definition.

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -309,15 +235,101 @@ def get_fuzz_tasks(self, num_tasks: int) -> list[tasks.Task]:
...
+
+
+def _get_swarming_jobs():
+  """Returns all jobs that have swarming environment variables."""
+  jobs = []
+  jobs.extend(_get_jobs_for_platforms(['ANDROID', 'LINUX']))
+  return [
+      job for job in jobs
+      if swarming.has_swarming_env_vars(job.get_environment())
+  ]
```
letitz, line 256: I was curious why you do `jobs = [], jobs.extend(...)`, so I went looking at the type signature for `get_all_from_query()`. Indeed it returns an iterator (so our type signature for `_get_job_for_platforms()` is wrong, we don't return a list right now), and in fact there's a TODO in there stating that it's become useless and callers should be migrated off of it.

Let's do:

```suggestion
def _get_jobs_for_platforms(platforms: list[str]) -> list[data_types.Job]:
  """Returns all jobs for the given platforms."""
  return list(data_types.Job.query(data_types.Job.platform.IN(platforms)))


def _get_swarming_jobs():
  """Returns all jobs that have swarming environment variables."""
  jobs = _get_jobs_for_platforms(['ANDROID', 'LINUX'])
  return [
      job for job in jobs
      if swarming.has_swarming_env_vars(job.get_environment())
  ]
```

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -309,15 +235,101 @@ def get_fuzz_tasks(self, num_tasks: int) -> list[tasks.Task]:
...
+
+  fuzzer_job_query = ndb_utils.get_all_from_query(
+      data_types.FuzzerJob.query(
+          data_types.FuzzerJob.job.IN(list(candidates_by_job.keys()))))
+  fuzz_task_candidates = []
+  for fuzzer_job in fuzzer_job_query:
+    candidate = candidates_by_job[fuzzer_job.job].copy()
+    candidate.fuzzer = fuzzer_job.fuzzer
+    candidate.weight = fuzzer_job.actual_weight
+    fuzz_task_candidates.append(candidate)
```
letitz: [Very optional] Consider not building partial FuzzTaskCandidates only to copy and fill them out later.

```suggestion
  if not jobs:
    return []
    
  jobs_by_name = { job.name: job for job in jobs }

  fuzzer_job_query = ndb_utils.get_all_from_query(
      data_types.FuzzerJob.query(
          data_types.FuzzerJob.job.IN(list(jobs_by_name.keys()))))
          
  fuzz_task_candidates = []
  for fuzzer_job in fuzzer_job_query:
    job = jobs_by_name[fuzzer_job.job]
    fuzz_task_candidates.append(FuzzTaskCandidate(
      job=job.name,
      project=job.project,
      base_os_version=job.base_os_version,
      fuzzer=fuzzer_job.fuzzer,
      candidate.weight=fuzzer_job.actual_weight,
    ))
```

src/clusterfuzz/_internal/tests/appengine/handlers/cron/schedule_fuzz_test.py
```diff
@@ -217,7 +217,8 @@ def _setup_chrome_entities(self, job_os_version=None):
   def _run_and_get_task(self):
     """Runs the scheduler and returns the single task created."""
     jobs = list(data_types.Job.query())
-    provider = schedule_fuzz.ChromeFuzzTaskProvider(jobs)
+    candidates = schedule_fuzz._create_candidates_from_jobs(jobs)
```
letitz: [Optional] Consider calling `_create_candidates_from_jobs()` in `ChromeFuzzTaskProvider.__init__()` since every call site needs it, then you won't need to reach into a protected module member here.

src/clusterfuzz/_internal/swarming/__init__.py
letitz: Ok thanks!

letitz (approved):
LGTM with a couple tiny comments left. Thanks for working with me through this!

src/clusterfuzz/_internal/cron/schedule_fuzz.py
IvanBM18: Sure thing :D

src/clusterfuzz/_internal/tests/appengine/handlers/cron/schedule_fuzz_test.py
IvanBM18: Sure thing makes sense

src/clusterfuzz/_internal/cron/schedule_fuzz.py
IvanBM18: Sure!

IvanBM18:
> LGTM with a couple tiny comments left. Thanks for working with me through this!

Thanks to you! :D for your patience and for guiding me trough this

src/clusterfuzz/_internal/cron/schedule_fuzz.py
javanlacerda: I agree with @letitz

IvanBM18: Sure, just doubled the preprocess queue default size to 10

javanlacerda (approved):

PR #5283, opened 2026-05-18
[Swarming]Pull and push utask_main tasks to Swarming queues

IvanBM18:
## Description
This change updates `run_bot.py` to fetch tasks from both the default `UTASK_MAIN_QUEUE` and the corresponding Swarming queue if the `SWARMING_REMOTE_EXECUTION` feature flag is enabled. This allows bots to process tasks intended for Swarming execution.

## Changes
- refactored the method signatures under `tasks/__init__.py` to push to to the given utask_main queue name 
- now `run_bot.py` Fetchs from both default and swarming queues if feature flag is enabled.
 -  Added unit tests for `run_bot` behavior with both queues.

## Test performed
These changes focuses on changing how we push and pull tasks to utask_main queues, so for:

#### Pushing
This happens in the preprocess step, here are example logs showing we can still push tasks into swarming
image
And we can see that reflected in the utask main queue
image

#### Pulling
At the utask_main scheduler:
image


src/python/bot/startup/run_bot.py
```diff
@@ -92,7 +93,13 @@ def schedule_utask_mains():
   """Schedules utask_mains from preprocessed utasks on remote backends."""
 
   logs.info('Attempting to combine batch tasks.')
-  utask_mains = tasks.get_utask_mains()
+  utask_mains = tasks.get_utask_mains(tasks.UTASK_MAIN_QUEUE)
+
+  if feature_flags.FeatureFlags.SWARMING_REMOTE_EXECUTION.enabled:
+    swarming_queue = tasks.SWARMING_QUEUES[tasks.UTASK_MAIN_QUEUE]
```
jardondiego: As it stands now, `tasks.UTASK_MAIN_QUEUE` will resolve to `utask_main`, which does not exist in `tasks.SWARMING_QUEUES`, and this would fail, I believe what we want to do is:

```python
#...
swarming_queue = tasks.SWARMING_QUEUES[tasks.SWARMING_QUEUES.UTASK_MAIN_QUEUE]
# ...
```

or alternatively

```python
#...
swarming_queue = tasks.SWARMING_QUEUES['utask_main-swarming']
#...
```

jardondiego (changes requested):

src/python/bot/startup/run_bot.py
jardondiego: As per file `tasks/__init__.py` on [line 101](https://github.com/google/clusterfuzz/blob/feature/swarming/utask_main_queue/src/clusterfuzz/_internal/base/tasks/__init__.py#L101), tasks.UTASK_MAIN_QUEUE would resolve as follows:  
`UTASK_MAIN_QUEUE = 'utask_main'`

jardondiego: And `'utask_main'` is not a key in the `SWARMING_QUEUES` dictionary definition on [line 104](https://github.com/google/clusterfuzz/blob/feature/swarming/utask_main_queue/src/clusterfuzz/_internal/base/tasks/__init__.py#L104) in that same file.

IvanBM18: Mmm are you sure? because: 
Since `tasks.UTASK_MAIN_QUEUE` evaluates to 'utask_main', this line:
```
swarming_queue = tasks.SWARMING_QUEUES[tasks.UTASK_MAIN_QUEUE]
```
Is equivalent to:
```
swarming_queue = tasks.SWARMING_QUEUES['utask_main']
```
The same applies to the SWARMING_QUEUES dictionary definition on [line 104](https://github.com/google/clusterfuzz/blob/feature/swarming/utask_main_queue/src/clusterfuzz/_internal/base/tasks/__init__.py#L104)
its equivalent to:
```
SWARMING_QUEUES = {
    'preprocess': 'preprocess-swarming',
    'utask_main': 'utask_main-swarming',
}
```

IvanBM18: So following this logic, the dictionary definition and usage works

IvanBM18: Its the same thing that we do in the `schedule_fuzz.py` 
https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/cron/schedule_fuzz.py#L61
And we have proved that it works

jardondiego: I think it's a confusion on my end. I see what you say. Thanks for break it down.

src/clusterfuzz/_internal/base/tasks/__init__.py
```diff
@@ -855,15 +854,19 @@ def run(self):
         logs.error('Leaser thread failed.')
 
 
-def add_utask_main(command, input_url, job_type, wait_time=None):
+def add_utask_main(command: str,
+                   input_url: str,
+                   job_type: str,
+                   wait_time: Optional[int] = None,
```
jardondiego: No need for `| None` when it's `Optional`.
```suggestion
                   wait_time: Optional[int],
```

jardondiego (approved):
LGTM

jardondiego:
Please include evidence of tests conducted in our development environment.

jardondiego:
It seems we have linting errors, please fix.

src/clusterfuzz/_internal/base/queues/__init__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2026 Google LLC
```
IvanBM18: Changes from this PR: https://github.com/google/clusterfuzz/pull/5296

src/clusterfuzz/_internal/base/queues/definitions.py
```diff
@@ -0,0 +1,47 @@
+# Copyright 2026 Google LLC
```
IvanBM18, line 1: Changes from this PR: https://github.com/google/clusterfuzz/pull/5296

src/clusterfuzz/_internal/base/queues/queue.py
```diff
@@ -0,0 +1,33 @@
+# Copyright 2026 Google LLC
```
IvanBM18: Changes from this PR: https://github.com/google/clusterfuzz/pull/5296

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -16,7 +16,6 @@
 from abc import ABC
```
IvanBM18, line 16: Changes from this PR: https://github.com/google/clusterfuzz/pull/5296

src/clusterfuzz/_internal/base/tasks/__init__.py
```diff
@@ -691,10 +691,9 @@ def get_task_from_message(message, queue=None, can_defer=True,
   return task
 
 
-def get_utask_mains() -> List[PubSubTask]:
+def get_utask_mains(queue_name: str = UTASK_MAIN_QUEUE) -> List[PubSubTask]:  # pylint: disable=dangerous-default-value
```
letitz: This function is called 4 times in the codebase before this change - only twice outside tests. Let's remove the default value and update its existing callers instead, and then we can remove the lint suppression.

src/clusterfuzz/_internal/base/tasks/__init__.py
```diff
@@ -706,9 +705,8 @@ def get_utask_mains() -> List[PubSubTask]:
 
 
 @memoize.wrap(memoize.InMemory(60))
-def get_utask_main_queue_size():
+def get_utask_main_queue_size(queue_name: str = UTASK_MAIN_QUEUE) -> int:
```
letitz: Same comment here.

src/clusterfuzz/_internal/base/tasks/__init__.py
```diff
@@ -860,15 +858,19 @@ def run(self):
         logs.error('Leaser thread failed.')
 
 
-def add_utask_main(command, input_url, job_type, wait_time=None):
+def add_utask_main(command: str,
+                   input_url: str,
+                   job_type: str,
+                   wait_time: Optional[int],
+                   queue_name: str = UTASK_MAIN_QUEUE) -> None:
```
letitz: Same comment here, this function is only called once.

src/clusterfuzz/_internal/tests/core/bot/startup/run_bot_test.py
```diff
@@ -165,6 +167,12 @@ def setUp(self):
         'clusterfuzz._internal.base.tasks.get_utask_mains',
         'clusterfuzz._internal.remote_task.remote_task_gate.RemoteTaskGate',
     ])
+    patcher = mock.patch(
```
letitz: If we have a datastore emulator, can we instead write a featureflags entry to datastore? Alternatively if you'd rather patch, then:

1. Add this mock to the list of mocks in `helpers.patch()` just above
2. Remove the datastore emulator?

src/clusterfuzz/_internal/tests/core/bot/startup/run_bot_test.py
```diff
@@ -197,14 +205,36 @@ def test_schedule_tasks_success(self):
...
     )
 
+  def test_schedule_tasks_both_queues(self):
+    """Test that schedule_utask_mains picks up tasks from both queues."""
+    regular_task = mock.MagicMock(
+        command='command1', job='job1', argument='argument1')
+    swarming_task = mock.MagicMock(
+        command='command2', job='job2', argument='argument2')
+
+    self.mock.get_utask_mains.side_effect = [[regular_task], [swarming_task]]
```
letitz, line 223: We should check that `get_utask_mains` was called with different queue names, otherwise this test could pass if the code just pulled from the default queue twice.

letitz (commented):
A few requests for small changes, but the high-level direction is good! Excited to see this land.

src/clusterfuzz/_internal/tests/core/bot/startup/run_bot_test.py
IvanBM18: I followed option 3 :D since we patch and have it disabled(but enabled in only one test) i just removed the patch and add in the data store emulator on the required test

src/clusterfuzz/_internal/base/tasks/__init__.py
IvanBM18: Done!

src/clusterfuzz/_internal/base/tasks/__init__.py
IvanBM18: Done :D

src/clusterfuzz/_internal/base/tasks/__init__.py
IvanBM18: makes sense, i tried to avoid it because it would lead to having a ton of file changes for just a small signature change, but seems a fair trade to avoid this linting error

letitz (approved):
LGTM

javanlacerda (approved):

PR #5284, opened 2026-05-19
[Swarming] Add backpressure mechanism to SwarmingService

IvanBM18:
## Overview
This PR implements a backpressure mechanism in `SwarmingService` to prevent overloading the Swarming pool with too many tasks.

## Changes
- Added `MAX_PENDING_TASKS = 25` to limit the number of pending tasks.
  - This threshold was chosen considering a pool of at most 25 bots and that a single scheduler takes ~3 min to schedule all of its tasks. And since we have 1000 schedulers in prod.. the best choice is to instead look for a stable backlog.
  - 25 ensures that the 25 bots always have a buffer of work  without building up a massive backlog 
- Updated `create_utask_main_jobs` to check the queue size before pushing each task using Swarming's `CountTasks` RPC.
- Implemented a **Fail Closed** strategy: if the `CountTasks` rpc call fails, we assume the queue is full and stop scheduling further tasks in the batch.
- Added unit tests to verify backpressure and failure handling.

## Tests performed
In this image we can appreciate that the utask main schedulers picks tasks from both the `utask_main` queue and the new `utask_main-swarming` queue
image

Heres another logs showing us that non swarming tasks are still being scheduled:
image

Here are some logs of the scheduler pushing tasks into swarming:
image


src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -51,14 +93,35 @@ def create_utask_main_jobs(self,
...
+              f'[Swarming] Failed to find OS dimension for job {task.job_type}.'
+          )
+          unscheduled_tasks.append(task)
+          continue
+
+        count_request = swarming_pb2.TasksCountRequest(  # pylint: disable=no-member
+            tags=['pool:chrome-sec-clusterfuzz', f'os:{os_val}'],
+            state=swarming_pb2.QUERY_PENDING)  # pylint: disable=no-member
+
+        if self._is_backpressure_applied(count_request):
```
IvanBM18: Note: if the queue is full then theres no sense to continue scheduling tasks in this "scheduling session" hence we just return the tasks as unscheduled and continue.

Since there are a ton of schedulers and each of them takes around 3mins to perform a scheduling session, this should not harm the scheduling of swarming tasks

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -51,14 +93,35 @@ def create_utask_main_jobs(self,
...
+
+        os_val = self._get_os_dimension(task_req)
+        if os_val == "":
+          logs.error(
+              f'[Swarming] Failed to find OS dimension for job {task.job_type}.'
+          )
+          unscheduled_tasks.append(task)
+          continue
+
+        count_request = swarming_pb2.TasksCountRequest(  # pylint: disable=no-member
```
IvanBM18, line 125: Note:
Since there are multiple concurrent scheduling sessions/bots, then its important to always make a req to swarming to account for the possibility that in the first iteration for this scheduler the queue size is 0, and then on the next one is 15 because of the multiple schedulers

src/clusterfuzz/_internal/base/feature_flags.py
```diff
@@ -37,6 +37,9 @@ class FeatureFlags(Enum):
 
   SWARMING_REMOTE_EXECUTION = 'swarming_remote_execution'
 
+  # TODO(b/516836939): Migrate feature flag to static config value.
```
jardondiego, line 41: It's great to have a bug created for this!

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -51,14 +94,35 @@ def create_utask_main_jobs(self,
...
+            tags=['pool:chrome-sec-clusterfuzz', f'os:{os_val}'],
+            state=swarming_pb2.QUERY_PENDING)  # pylint: disable=no-member
```
jardondiego, line 127: Shouldn't we also count tasks with a `RUNNING` state?

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -12,14 +12,19 @@
...
 
 from clusterfuzz._internal import swarming
+from clusterfuzz._internal.base.feature_flags import FeatureFlags
 from clusterfuzz._internal.base.tasks import task_utils
 from clusterfuzz._internal.metrics import logs
+from clusterfuzz._internal.protos import swarming_pb2
 from clusterfuzz._internal.remote_task import remote_task_types
 from clusterfuzz._internal.swarming.api import SwarmingApi
 
+_DEFAULT_MAX_PENDING_TASKS = 25
```
jardondiego: How did we come up with this number? It would be great to have a comment explaining how we determined this was the right number.

jardondiego: Why name it `_DEFAULT_MAX_PENDING_TASKS` and not `DEFAULT_MAX_PENDING_TASKS`?
Should this be considered local to this module?

IvanBM18: You can find more info in this PR description :D, but ill add a separate comment here as well

IvanBM18: Yes :D, theres no reason to make it public atm and i dont think it will be

src/clusterfuzz/_internal/swarming/service.py
IvanBM18: mmm good question, AFAIK backpressure is calculated over queue sizes not over working bots + queue sizes.

We then tweak the max amount of pending tasks in queues(trough feature flags or default values) depending on the bot fleet size so backpressure has a good balance.

So no we shouldn't be interested in knowing how many bots are busy to calculate backpressure

src/clusterfuzz/_internal/swarming/service.py
IvanBM18: This threshold was chosen considering a pool of at most 25 bots and that a single scheduler takes ~3 min to schedule all of its tasks. And since we have 1000 schedulers in prod.. the best choice is to instead look for a stable backlog.
- 25 ensures that the 25 bots always have a buffer of work without building up a massive backlog.

And as comments in other prs says:
```
I think we should prevent the infinitely-growing queue of tasks using CountTasks and a pretty low limit on the utask_main queue size
```

jardondiego: Makes sense. Thanks :)

jardondiego: Makes sense, thanks!

src/clusterfuzz/_internal/swarming/service.py
jardondiego: Got it. Thanks!

src/clusterfuzz/_internal/swarming/service.py
jardondiego: Please mark as resolved once the comment is added.

src/clusterfuzz/_internal/swarming/api.py
```diff
@@ -136,22 +136,30 @@ def push_task(self, task_request: swarming_pb2.NewTaskRequest) -> str | None:  #
     response = self._make_request(_NEW_TASK_ENDPOINT, message_body)
     return response
 
-  def count_tasks(self,
-                  count_request: swarming_pb2.TasksCountRequest) -> str | None:  # pylint: disable=no-member
+  def count_tasks(
+      self,
+      count_request: swarming_pb2.TasksCountRequest,  # pylint: disable=no-member
+  ) -> swarming_pb2.TasksCount | None:  # pylint: disable=no-member
```
jardondiego: Isn't better to use `Optional` here instead of `| None`?

src/clusterfuzz/_internal/base/queues/__init__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2026 Google LLC
```
IvanBM18: Changes from this PR: https://github.com/google/clusterfuzz/pull/5296

src/clusterfuzz/_internal/base/queues/definitions.py
```diff
@@ -0,0 +1,47 @@
+# Copyright 2026 Google LLC
```
IvanBM18, line 1: Changes from this PR: https://github.com/google/clusterfuzz/pull/5296

src/clusterfuzz/_internal/base/queues/queue.py
```diff
@@ -0,0 +1,33 @@
+# Copyright 2026 Google LLC
```
IvanBM18: Changes from this PR: https://github.com/google/clusterfuzz/pull/5296

src/clusterfuzz/_internal/base/feature_flags.py
```diff
@@ -38,6 +38,10 @@ class FeatureFlags(Enum):
   SWARMING_REMOTE_EXECUTION = 'swarming_remote_execution'
```
IvanBM18, line 38: Changes from this PR: https://github.com/google/clusterfuzz/pull/5296

src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -16,7 +16,6 @@
 from abc import ABC
```
IvanBM18, line 16: Changes from this PR: https://github.com/google/clusterfuzz/pull/5296

src/clusterfuzz/_internal/swarming/api.py
letitz: ISTM `| None` is preferable now as Python supports the syntax and it removes the need for one more `typing` import. This seems to be the stance of Python typing folks: https://typing.python.org/en/latest/reference/best_practices.html#shorthand-syntax. Internally it is also encouraged: go/python-tips/035#fundamental-building-blocks

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -29,6 +32,44 @@ class SwarmingService(remote_task_types.RemoteTaskInterface):
...
+      if dimension.key == 'os':
+        return dimension.value
+    return ""
+
+  def _get_max_pending_tasks(self) -> int:
+    """Returns the maximum number of pending tasks allowed in Swarming queue"""
+    flag = SWARMING_UTASK_MAIN_QUEUE.target_size_flag
+    if flag.enabled and flag.content is not None:
+      return int(flag.content)
+    return SWARMING_UTASK_MAIN_QUEUE.default_target_size
```
letitz: This logic is now duplicated with `schedule_fuzz.py`. Let's extract it to a member of `Queue` in #5296?

In fact, even the logic for `_count_remaining_tasks()` is now duplicated, we should extract that.

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -29,6 +32,44 @@ class SwarmingService(remote_task_types.RemoteTaskInterface):
...
+
+      count = int(response.count)
+      max_pending_tasks = self._get_max_pending_tasks()
+      if count >= max_pending_tasks:
+        logs.info(f'[Swarming] Backpressure applied. Queue size: {count}. '
+                  'Stopping scheduling.')
+        return True
+
+      return False
+    except HTTPError as api_failure:
```
letitz: AFAICT this should only wrap the `count_tasks()` call on line 56, so let's tighten the try/catch block.

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -29,6 +32,44 @@ class SwarmingService(remote_task_types.RemoteTaskInterface):
...
+    if flag.enabled and flag.content is not None:
+      return int(flag.content)
+    return SWARMING_UTASK_MAIN_QUEUE.default_target_size
+
+  def _is_backpressure_applied(
+      self, count_request: swarming_pb2.TasksCountRequest) -> bool:  # pylint: disable=no-member
+    """Checks if backpressure should be applied based on pending tasks count.
+    
+    Returns True if backpressure is applied or if the check fails (Fail Closed).
+    """
```
letitz, line 58: We should document that it can raise a `RuntimeError`.

src/clusterfuzz/_internal/swarming/api.py
```diff
@@ -136,22 +136,30 @@ def push_task(self, task_request: swarming_pb2.NewTaskRequest) -> str | None:  #
...
     """Counts tasks on swarming.
     
     Args:
       count_request: The TasksCountRequest proto message.
       
     Returns:
-      The raw JSON response string from the server, or None if the response is
-      empty.
+      The TasksCount parsed proto message from the server, or None
```
letitz: Given that the only caller raises an error in the empty case, can we instead expose the Parse error? I assume `json_format.Parse()` raises some kind of exception if given an empty string, but that may be wrong.

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -29,6 +32,44 @@ class SwarmingService(remote_task_types.RemoteTaskInterface):
...
+    try:
+      response = self._api.count_tasks(count_request)
+      if not response:
+        raise RuntimeError("Empty response from CountTasks")
```
letitz: `ValueError` seems like a slightly more precise type? See also the comment in `api.py` about exposing a parse exception instead.

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -29,6 +32,44 @@ class SwarmingService(remote_task_types.RemoteTaskInterface):
...
+    return ""
+
+  def _get_max_pending_tasks(self) -> int:
+    """Returns the maximum number of pending tasks allowed in Swarming queue"""
+    flag = SWARMING_UTASK_MAIN_QUEUE.target_size_flag
+    if flag.enabled and flag.content is not None:
+      return int(flag.content)
+    return SWARMING_UTASK_MAIN_QUEUE.default_target_size
+
+  def _is_backpressure_applied(
```
letitz: Naming nit: The function is named "is applied", but the docstring immediately says "should apply". We should pick one and stick with it.

In fact, I would invert the meaning of the return value and go for something simpler: `_is_queue_full()`, or return an int and name it `_remaining_queue_capacity()`.

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -29,6 +32,44 @@ class SwarmingService(remote_task_types.RemoteTaskInterface):
...
+      if count >= max_pending_tasks:
+        logs.info(f'[Swarming] Backpressure applied. Queue size: {count}. '
+                  'Stopping scheduling.')
+        return True
+
+      return False
+    except HTTPError as api_failure:
+      logs.error('[Swarming] Failed to check backpressure (Fail Closed): '
+                 f'{api_failure}')
+      return True  #Always fail if swarming request fails
```
letitz: FWIW the comment is duplicative with the error log above.

src/clusterfuzz/_internal/swarming/service.py
letitz: Fair enough! Thanks for pointing that out. This is typically something you should write as a comment in the code. If it's important and subtle enough to call out to your reviewer, it almost certainly is useful to call it out to future readers of the code :)

src/clusterfuzz/_internal/swarming/service.py
letitz: Also a good point, let's write that in a comment in the code.

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -29,6 +32,44 @@ class SwarmingService(remote_task_types.RemoteTaskInterface):
...
+  def _get_os_dimension(self, request: swarming_pb2.NewTaskRequest) -> str:  # pylint: disable=no-member
+    """Extracts the OS dimension from the task request."""
+    for dimension in request.task_slices[0].properties.dimensions:
+      if dimension.key == 'os':
+        return dimension.value
+    return ""
+
+  def _get_max_pending_tasks(self) -> int:
+    """Returns the maximum number of pending tasks allowed in Swarming queue"""
+    flag = SWARMING_UTASK_MAIN_QUEUE.target_size_flag
```
letitz: The target size specified here is actually quite subtly allocated independently for each OS, so there can be e.g. 25 android tasks and 25 linux tasks. I'm not convinced we want to have the same limit for each OS. Since we only have android for now, I would be inclined to drop the OS dimension checking entirely from the code and replace it with a TODO linking to a relevant bug. If you want to keep it, then let's add a TODO here linking to a bug tracking what will need to be improved before we can spin up different kinds of bots in swarming.

src/clusterfuzz/_internal/tests/core/swarming/api_test.py
```diff
@@ -117,6 +117,7 @@ def test_push_task_no_credentials(self):
   def test_count_tasks_no_credentials(self):
```
letitz, line 162: We should still have a test for `CountTasks` returning an empty string.

src/clusterfuzz/_internal/tests/core/swarming/service_test.py
```diff
@@ -38,8 +41,18 @@ def setUp(self):
     self.mock_api = mock.MagicMock()
     self.mock.create.return_value = self.mock_api
     self.service = service.SwarmingService()
-    self.mock.create_new_task_request.return_value = 'fake_request'
+
+    self.mock_request = mock.MagicMock()
+    mock_dimension = mock.MagicMock()
```
letitz: Why do we need a mock here, can't we use a real proto?

src/clusterfuzz/_internal/tests/core/swarming/service_test.py
```diff
@@ -38,8 +41,18 @@ def setUp(self):
...
+    mock_dimension.key = 'os'
+    mock_dimension.value = 'Linux'
+    self.mock_request.task_slices[0].properties.dimensions = [mock_dimension]
+    self.mock.create_new_task_request.return_value = self.mock_request
+
     self.mock.get.return_value = None
+    self.mock_api = mock.MagicMock()
+    self.service._api = self.mock_api  # pylint: disable=protected-access
+    self.mock_api.count_tasks.return_value = mock.MagicMock(count=0)
```
letitz: Same question

src/clusterfuzz/_internal/tests/core/swarming/service_test.py
```diff
@@ -146,3 +159,60 @@ def test_create_utask_main_jobs_handles_http_error(self):
...
+    self.assertEqual(unscheduled[0].job_type, 'job1')
+    self.assertEqual(unscheduled[1].job_type, 'job2')
+
+    self.mock_api.push_task.assert_not_called()
+
+  def test_get_max_pending_tasks_with_feature_flag(self):
+    """Test that the max pending tasks priority is feature flag > default"""
+    data_types.FeatureFlag(
+        id='swarming_max_pending_tasks', enabled=True, value=50.0).put()
+    self.assertEqual(self.service._get_max_pending_tasks(), 50)  # pylint: disable=protected-access
```
letitz: If you extract that logic to the queue module/class, you can test it there without reaching for a protected member.

letitz (changes requested):
One substantive comment about the queue size limit being per-OS, otherwise a bunch of small comments.

src/clusterfuzz/_internal/tests/core/swarming/service_test.py
IvanBM18: Sure, just created some tests in its pr https://github.com/google/clusterfuzz/pull/5296

IvanBM18: Ill proceed to remove them from this pr

src/clusterfuzz/_internal/swarming/service.py
IvanBM18: Just updated the other PR https://github.com/google/clusterfuzz/pull/5296
So that this method exists there, ill proceed and remove this method from here

src/clusterfuzz/_internal/swarming/api.py
IvanBM18: It raises AttributeError or ParserError, Do we want to expose this?
I feel like exposing 3 exceptions is way too much and if we do in the end we will end up catching all of those exceptions(if used elsewhere).

So whats the python way here?

src/clusterfuzz/_internal/swarming/service.py
IvanBM18: Fair enough just done so :D thanks

src/clusterfuzz/_internal/swarming/service.py
IvanBM18: Sure, thanks :D

src/clusterfuzz/_internal/swarming/service.py
IvanBM18: Sure thing!

src/clusterfuzz/_internal/swarming/service.py
IvanBM18: Done :D

src/clusterfuzz/_internal/swarming/service.py
IvanBM18: Fair enough i liked _is_queue_full() better here :D

src/clusterfuzz/_internal/swarming/service.py
IvanBM18: Done!

src/clusterfuzz/_internal/tests/core/swarming/service_test.py
IvanBM18: We can!

src/clusterfuzz/_internal/tests/core/swarming/service_test.py
IvanBM18: We can!

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -39,36 +40,34 @@ def _get_os_dimension(self, request: swarming_pb2.NewTaskRequest) -> str:  # pyl
...
+                     count_request: swarming_pb2.TasksCountRequest) -> bool:  # pylint: disable=no-member
+    """Checks if the queue is full based on pending tasks count.
     
-    Returns True if backpressure is applied or if the check fails (Fail Closed).
+    Returns True if the queue is full or if the check fails (Fail Closed).
+
+    Raises:
+      ValueError: If failed to get pending tasks count.
     """
+    #TODO(b/517517107): Calculate backpressure for multiple target OS
```
letitz: We already calculate backpressure per OS, so it's hard for me to parse this TODO. How about:

```suggestion
    # TODO(b/517517107): Improve backpressure calculation to account for
    # differently-sized bot groups in swarming.
```

I'll comment on the bug to give more details there too.

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -39,36 +40,34 @@ def _get_os_dimension(self, request: swarming_pb2.NewTaskRequest) -> str:  # pyl
...
     except HTTPError as api_failure:
       logs.error('[Swarming] Failed to check backpressure (Fail Closed): '
                  f'{api_failure}')
-      return True  #Always fail if swarming request fails
+      return True
+
+    if not response:
+      raise ValueError("Empty response from CountTasks")
+
+    count = int(response.count)
```
letitz: No need to convert the field to an int, it already is one: https://source.chromium.org/chromium/infra/infra_superproject/+/main:infra/luci/appengine/swarming/proto/api_v2/swarming.proto;l=976;drc=52e221fe2ac985548bfd535943a17bcedef40dbd

src/clusterfuzz/_internal/swarming/api.py
letitz: Fair enough! I think the Python way here would be to write a `SwarmingApiError` class that can represent any of these failures, and let callers deal with a single exception type. Returning `None` is fine too since there is little information we could stick in the exception that the caller would not have access to also.

I'll note that technically, `json_format.Parse()` can still fail if e.g. our proto definition is wrong, or the server simply misbehaves, and we're just assuming that won't happen - which I think is relatively reasonable :)

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -51,14 +91,42 @@ def create_utask_main_jobs(self,
...
+          )
+          unscheduled_tasks.append(task)
+          continue
+
+        # Since there are multiple concurrent scheduling sessions/bots, it is
+        # important to always check the queue size with Swarming to account for
+        # the possibility that the queue was empty in the first iteration but
+        # filled up on subsequent iterations due to other schedulers.
+        count_request = swarming_pb2.TasksCountRequest(  # pylint: disable=no-member
+            tags=['pool:chrome-sec-clusterfuzz', f'os:{os_val}'],
```
letitz: Can we glean the pool from the task request instead of hardcoding it here?

src/clusterfuzz/_internal/tests/core/swarming/service_test.py
letitz: Great, let's do it then!

letitz (approved):
LGTM % a couple remaining comments to avoid blocking on me further.

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -29,6 +33,42 @@ class SwarmingService(remote_task_types.RemoteTaskInterface):
...
+    except HTTPError as api_failure:
+      logs.error('[Swarming] Failed to check backpressure (Fail Closed): '
+                 f'{api_failure}')
+      return True
+
+    if not response:
+      raise ValueError("Empty response from CountTasks")
+
+    count = int(response.count)
+    max_pending_tasks = SWARMING_UTASK_MAIN_QUEUE.get_max_pending_size()
```
letitz: Wait a minute. Is this the right queue? I think not. This is the queue that links the preprocess bots to the utask_main schedulers (that run `RemoteTaskGate` code). Here we are considering a different queue (not a pub/sub queue, in fact): the swarming server's internal task queue. We should probably have a different knob to tune that?

letitz (changes requested):

src/clusterfuzz/_internal/swarming/service.py
IvanBM18: sure :D

src/clusterfuzz/_internal/swarming/service.py
IvanBM18: Thanks a ton for pointing it out, i must have gotten confused with the introduction of the queues module

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -29,6 +37,45 @@ class SwarmingService(remote_task_types.RemoteTaskInterface):
...
+      logs.error('[Swarming] Failed to check backpressure (Fail Closed): '
+                 f'{e}')
+      return True
+
+    if not response:
+      logs.warning('[Swarming] Got an empty response from Swarming CountTasks.')
+      return True
+
+    count = response.count
+    max_pending_tasks = get_max_size_for_queue(
```
IvanBM18, line 69: Note for reviewer: 
Since the swarming queue is not a pub/sub queue i extracted some of the logic from the data class to the public to reuse it here.

The other alternative to this is create a PubSubTaskQueue instance for the swarming queue(im not against this!) but if i do so i would also like to rename the PubSubTaskQueue module, since it would be confusing to have a object instance that its not really pub/sub queue.

src/clusterfuzz/_internal/tests/core/swarming/api_test.py
IvanBM18: Done! thanks

src/clusterfuzz/_internal/swarming/api.py
```diff
@@ -136,22 +141,34 @@ def push_task(self, task_request: swarming_pb2.NewTaskRequest) -> str | None:  #
...
+      if the response is empty.
 
     Raises:
       requests.exceptions.HTTPError: If the request fails with a 4xx or 5xx
```
letitz: Ah, if you want to introduce a new exception type then I think it should be the only exception you can raise here. I don't think it's useful to callers to know if it's an HTTP error vs another kind of error. Plus, you could expose that info from `SwarmingApiError` if need be.

src/clusterfuzz/_internal/base/tasks/pub_sub_task_queue.py
```diff
@@ -80,3 +78,11 @@ def get_max_target_size(self) -> int:
     default_target_size=SWARMING_UTASK_MAIN_QUEUE_LIMIT_DEFAULT,
     target_size_flag=FeatureFlags.SWARMING_MAX_PENDING_TASKS,
 )
+
+
+def get_max_size_for_queue(default_size: int,
```
letitz, line 83: Should be documented.

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -29,6 +37,45 @@ class SwarmingService(remote_task_types.RemoteTaskInterface):
   def __init__(self):
     self._api = SwarmingApi.create()
 
+  # pylint: disable=no-member
+  def _get_dimension(self, request: swarming_pb2.NewTaskRequest,
+                     key: str) -> str:
+    """Extracts a dimension value from the task request."""
```
letitz: We should document the return value when not found.

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -29,6 +37,45 @@ class SwarmingService(remote_task_types.RemoteTaskInterface):
...
+    # differently-sized bot groups in swarming.
+    try:
+      response = self._api.count_tasks(count_request)
+    except (HTTPError, SwarmingApiError) as e:
+      logs.error('[Swarming] Failed to check backpressure (Fail Closed): '
+                 f'{e}')
+      return True
+
+    if not response:
+      logs.warning('[Swarming] Got an empty response from Swarming CountTasks.')
```
letitz: My 2c: I think you should avoid returning `None` for empty responses and instead raise an API error.

Reasoning:

* there is no difference in caller behavior here for exception vs None, we return True in both cases
* an empty HTTP response body is a breach of the RPC contract, so I think it should be handled like a parse error which raises an exception

src/clusterfuzz/_internal/swarming/service.py
letitz: That makes sense, and it's good enough for me to land this PR. I think ideally you would have something like:

```py
# task_queue.py

class TaskQueueLimiter:
  default_limit: int
  feature_flag: FeatureFlags

  def max_pending_tasks(self) -> int:
    ...


class PubSubTaskQueue:
  name: str
  limiter: TaskQueueLimiter
```

Since the module name here is still awkward. I don't think you should actually do it though, this is good enough and we have other things to do :)

letitz (commented):
Thanks, we're nearly there!

src/clusterfuzz/_internal/swarming/api.py
IvanBM18: Done :D

src/clusterfuzz/_internal/base/tasks/pub_sub_task_queue.py
IvanBM18: Done!

src/clusterfuzz/_internal/swarming/service.py
IvanBM18: Done :D

src/clusterfuzz/_internal/swarming/service.py
IvanBM18: Done!

src/clusterfuzz/_internal/swarming/api.py
```diff
@@ -155,20 +156,24 @@ def count_tasks(
       if the response is empty.
```
letitz: Docstring needs a fix since we never return None now.

src/clusterfuzz/_internal/swarming/api.py
letitz: We never return `None` now, type annotation should be fixed.

letitz (approved):
LGTM % tiny fixes to `SwarmingApi.count_tasks()`.

javanlacerda (approved):
lgtm!

PR #5285, opened 2026-05-20
[Swarming] Crash service on startup 

IvanBM18:
## Overview
`SwarmingService` and other remote task services will now be instantiated in `RemoteTaskGate` based on the their feature flag. 

Now the `SwarmingService` will crash horribly at startup, this is done to avoid happing an api instances that does nothing, and a swarming service that just fails when called, with this new logic we make sure that all parts work as expected and if they dont.... well we just wont be able to fully schedule on envs where swarming is enabled.

### Background
This PR comes from a discussion & agreement to a required change from [this review](https://github.com/google/clusterfuzz/pull/5277#discussion_r3266259807)

## Changes
- Raises an exception at `SwarmingService` constructor if the swarming config was not found
- Updated `RemoteTaskGate` to conditionally instantiate services based on the linked feature flag.
  - This safeguards envs where the swarming feature flag is not enabled.
- Added unit tests & updated a lot of them for the remote task gate changes.

### DIFF
--- a/src/clusterfuzz/_internal/remote_task/remote_task_gate.py
+++ b/src/clusterfuzz/_internal/remote_task/remote_task_gate.py
@@ -43,6 +43,7 @@ class RemoteTaskGate(remote_task_types.RemoteTaskInterface):
     self._service_map = {
         adapter.id: adapter.service()
         for adapter in remote_task_adapters.RemoteTaskAdapters
+        if adapter.feature_flag.enabled
     }
     self._adapters = remote_task_adapters.RemoteTaskAdapters
 
--- a/src/clusterfuzz/_internal/tests/core/remote_task/remote_task_gate_test.py
+++ b/src/clusterfuzz/_internal/tests/core/remote_task/remote_task_gate_test.py
@@ -46,19 +46,19 @@ class RemoteTaskGateTest(unittest.TestCase):
                 mock.Mock(
                     id='kubernetes',
                     service=mock.Mock(return_value=self.mock_k8s_service),
-                    feature_flag=None,
+                    feature_flag=mock.Mock(enabled=True),
                     default_weight=0.0),
             'GCP_BATCH':
                 mock.Mock(
                     id='gcp_batch',
                     service=mock.Mock(return_value=self.mock_gcp_batch_service),
-                    feature_flag=None,
+                    feature_flag=mock.Mock(enabled=True),
                     default_weight=1.0),
             'SWARMING':
                 mock.Mock(
                     id='swarming',
                     service=mock.Mock(return_value=self.mock_swarming_service),
-                    feature_flag=None,
+                    feature_flag=mock.Mock(enabled=True),
                     default_weight=0.0),
         })
     self.patcher.start()
@@ -488,19 +488,19 @@ class RemoteTaskGateProcessingTest(unittest.TestCase):
                 mock.Mock(
                     id='kubernetes',
                     service=mock.Mock(),
-                    feature_flag=None,
+                    feature_flag=mock.Mock(enabled=True),
                     default_weight=0.0),
             'GCP_BATCH':
                 mock.Mock(
                     id='gcp_batch',
                     service=mock.Mock(),
-                    feature_flag=None,
+                    feature_flag=mock.Mock(enabled=True),
                     default_weight=1.0),
             'SWARMING':
                 mock.Mock(
                     id='swarming',
                     service=mock.Mock(return_value=self.mock_swarming_service),
-                    feature_flag=None,
+                    feature_flag=mock.Mock(enabled=True),
                     default_weight=0.0),
         })
     self.patcher.start()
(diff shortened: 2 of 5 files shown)


src/clusterfuzz/_internal/tests/core/remote_task/remote_task_test.py
```diff
@@ -47,6 +48,34 @@ def setUp(self):
     self.addCleanup(patcher.stop)
     patcher.start()
 
+    # Patch RemoteTaskAdapters to enable feature flags in tests.
```
letitz: [Optional] This works and aligns with other tests, so it's good enough.

AFAICT, you are only overriding `enabled` on the feature flags. If you are curious to try, you could see if you could instead just set the feature flag values in datastore for the duration of the tests.

See the FeatureFlags test for inspiration: https://github.com/google/clusterfuzz/blob/03cb7cd76842ccabd55c90802382dc0b677719b8/src/clusterfuzz/_internal/tests/core/datastore/feature_flag_test.py#L29

Something like:

```py
data_types.FeatureFlag(
    id=RemoteTaskAdapters.SWARMING.feature_flag.value,
    enabled=True,
).put()
```

You'll need the datastore emulator in that case.

src/clusterfuzz/_internal/tests/core/remote_task/remote_task_gate_test.py
letitz, line 1: We should likely have a test that inspects what kinds of adapters are constructed when feature flags are disabled.

letitz (commented):
IIUC the only diff from PR #5277 is 5fbc97d3043816bf8b75f0b3f0d74b3ce4487f77. At least that's what I reviewed :)

In that case, this LGTM other than one optional suggestion and one request for more tests.

src/clusterfuzz/_internal/tests/core/remote_task/remote_task_test.py
IvanBM18: Good catch! :D 
thanks

src/clusterfuzz/_internal/remote_task/remote_task_gate.py
```diff
@@ -43,6 +43,7 @@ def __init__(self):
     self._service_map = {
         adapter.id: adapter.service()
         for adapter in remote_task_adapters.RemoteTaskAdapters
+        if adapter.feature_flag.enabled
```
jardondiego, line 46: Are you sure this makes it fail more loudly? If the feature flag is not marked as enabled, this dictionary of services would not include it, right? Wouldn't a loud failure imply raising some sort of exception?

IvanBM18: Answering your questions...
1- Yes now when the service is not correctly created a ValueError is raised, causing it to fail more loudly, thats where we raise the exception you ask
2- If the feature flag is disabled, then the dictionary will not include an entry for the given service(see unit test at `remote_task_test` as proof)

Xeicker:
I'd suggest rebase your branch on top of master so we don't merge many commits like `Merge branch 'master' into feature/swarming/crash_service_at_startup`

Other than that, LGTM

fernandofloresg (approved):
LGTM

IvanBM18:
> I'd suggest rebase your branch on top of master so we don't merge many commits like `Merge branch 'master' into feature/swarming/crash_service_at_startup`
> 
> Other than that, LGTM

Done :D

src/clusterfuzz/_internal/tests/core/remote_task/remote_task_gate_test.py
```diff
@@ -488,19 +488,19 @@ def setUp(self):
...
                     service=mock.Mock(),
-                    feature_flag=None,
+                    feature_flag=mock.Mock(enabled=True),
                     default_weight=1.0),
             'SWARMING':
                 mock.Mock(
                     id='swarming',
                     service=mock.Mock(return_value=self.mock_swarming_service),
-                    feature_flag=None,
+                    feature_flag=mock.Mock(enabled=True),
```
javanlacerda, line 503: Should we have a test case for enabled=False?

javanlacerda (approved):

src/clusterfuzz/_internal/tests/core/remote_task/remote_task_gate_test.py
IvanBM18: Sure, i didn't read this before merging(sorry about that) but ill add it in [this open pr](https://github.com/google/clusterfuzz/pull/5289)

IvanBM18: Wait nvm, i did add a case for it at [‎src/clusterfuzz/_internal/tests/core/remote_task/remote_task_test.py‎](https://github.com/google/clusterfuzz/pull/5285/changes/BASE..f537ad75b7148ab564200e40a671ee630ef126ce#diff-7b891a065911b780a8f4e5aade6728f3cb399980056d04cfab6f309d26e87f95)

PR #5286, opened 2026-05-20
Provide context on how to correctly call butler in AGENTS.md

IvanBM18:
## Overview
This PR updates `AGENTS.md` to provide clear instructions on how to correctly use `butler.py` with `pipenv`, ensuring agents operate within the virtual environment.

Currently, agents always fail when calling butler commands because they don't know(yet) that butler commands fail if out of a pipenv environment, hence you are always correcting it, or waiting for it till it reruns commands in `pipenv` .

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -54,6 +54,20 @@ Now you are ready to install the project's dependencies. Run the following comma
 ./local/install_deps.bash
 ```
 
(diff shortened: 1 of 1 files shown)


fernandofloresg (approved):
lgtm

javanlacerda (approved):

PR #5287, opened 2026-05-20
Ignore rows which are missing stats during daily aggregation

dylanjew:
There were a couple old versions of CF running in the clusterfuzz-candidate2 bots. This PR ignores the stats which were generated by those bots. It's unfortunate that the stats are missing these, but IMO it's better to completely ignore these rows rather than writing partial data. There were 10 of these instances running, which resulted in some fuzzers having a fuzzing session unaccounted for.

Testing: Ran this SQL against the BigQuery table and verified it correctly filters out the rows and resulted in the `testcases_generated == testcases_executed`

### Rollout

we will have to rerun the cron job for the past month again after this deploys to properly ignore those rows.


aakallam (approved):
LGTM

ViniciustCosta (approved):

PR #5288, opened 2026-05-20
Fix Android UWORKER initialization crash bypassing test account provisioning

jardondiego:
When executing Android fuzzing tasks within the Untrusted Worker (UWORKER) environment on Swarming, the initialization sequence consistently crashes with a fatal gRPC StatusCode.PERMISSION_DENIED (403) error.

### Root Cause
During the Android device setup, device.initialize_device() calls add_test_accounts_if_needed(). This legacy function attempts to retrieve test account credentials by executing db_config.get(), which triggers a query to Google Cloud Datastore. Because UWORKER Docker containers are intentionally stripped of IAM privileges (to prevent malicious payloads from
compromising the database), the Datastore query fails and crashes the entire fuzzing pipeline before the fuzzer can even launch.

### Solution
This PR implements a tactical bypass to short-circuit the test account provisioning logic when operating inside an untrusted environment.

  * Added a check (if environment.is_uworker(): return) at the top of add_test_accounts_if_needed().
  * Added unit tests to device_test.py to ensure the bypass functions correctly and prevents the downstream Datastore queries.

### Rationale
While the "architecturally correct" fix would involve querying these credentials during the Trusted Worker preprocessing phase and serializing them into the uworker_msg.proto payload, that represents a significant engineering effort.

The vast majority of Android fuzzing targets (especially native binaries and blackbox fuzzers) do not strictly require a logged-in Google account to achieve coverage. This bypass is the least invasive way to immediately unblock Android fuzzing on the Swarming fleet without compromising the UWORKER security boundaries.


src/clusterfuzz/_internal/tests/core/platforms/android/device_test.py
```diff
@@ -23,3 +25,20 @@ class InitializeEnvironmentTest(android_helpers.AndroidTest):
   def test(self):
     """Ensure that initialize_environment throws no exceptions."""
     device.initialize_environment()
+
+
+class AddTestAccountsIfNeededTest(unittest.TestCase):
+  """Tests for add_test_accounts_if_needed."""
+
+  def setUp(self):
```
fernandofloresg, line 33: is it needed to be camelCase? usually python uses snake_case for function names?

fernandofloresg (approved):
approved only had one question

src/clusterfuzz/_internal/tests/core/platforms/android/device_test.py
jardondiego: It is part of Python's built-in [unittest framework](https://docs.python.org/3/library/unittest.html), it's indeed `setUp` in camel case.

ViniciustCosta (approved):
Lgtm

PR #5289, opened 2026-05-21
[Swarming] Adds Response proto & discards tasks with missing dimensions

IvanBM18:
## Overview
It was request for me to add a log to easily debug when swarming scheduled a task, but i noticed we didnt had the protos for parsing the response, hence i added them.

Also this PR handles an edge case in which swarming will not return a task_id if the tasks wasn't scheduled due to swarming related issues(not request, nor pc contract).

#### Background
This PR comes from a discussion & agreement to a required change from this [PR](https://github.com/google/clusterfuzz/pull/5277)

## Changes
- Adds the missing swarming_task_response proto
  - Ran `protos/generate.sh` which updated a ton of files
-  The swarming service now marks tasks as not scheduled when no task_id is found, task_id cannot be found if swarming didn't queue the task, because of wrong dimensions or other swarming internal reasons
- As the `count_tasks()` api method, now the `push_task()` will also only raise `SwarmingAPIErrors`


src/clusterfuzz/_internal/protos/swarming_pb2.py
```diff
@@ -29,15 +29,15 @@
 from google.protobuf import timestamp_pb2 as google_dot_protobuf_dot_timestamp__pb2
```
IvanBM18, line 29: Changes autogenerated by the `protos/genreate.sh` script

src/clusterfuzz/_internal/protos/swarming_pb2.pyi
```diff
@@ -797,6 +797,90 @@ class NewTaskRequest(google.protobuf.message.Message):
 
 global___NewTaskRequest = NewTaskRequest
 
+@typing_extensions.final
```
IvanBM18, line 800: Changes autogenerated by the `protos/genreate.sh` script

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -99,42 +97,53 @@ def create_utask_main_jobs(self,
     unscheduled_tasks = []
     logs.info(f'[Swarming] Pushing {len(remote_tasks)} tasks trough service.')
     for i, task in enumerate(remote_tasks):
-      try:
-        if not swarming.is_swarming_task(task.job_type):
-          unscheduled_tasks.append(task)
-          continue
+      if not swarming.is_swarming_task(task.job_type):
```
IvanBM18, line 103: Note, this is marked as new lines because i moved the `try` downwards to line 133, so that we avoid unnecesay nesting,since the only method that really throws exceptions is located at line 134

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -99,42 +97,53 @@ def create_utask_main_jobs(self,
...
+        continue
 
-        os_val = self._get_dimension(task_req, 'os')
-        pool_val = self._get_dimension(task_req, 'pool')
-        if not os_val or not pool_val:
-          logs.error(f'[Swarming] Failed to find required dimension for job '
+      os_val = self._get_dimension(task_req, 'os')
+      pool_val = self._get_dimension(task_req, 'pool')
+      if not os_val or not pool_val:
+        logs.warning(f'[Swarming] Failed to find required dimension for job '
```
Xeicker: why did you change to warning level? this looks like error level to me

fernandofloresg (approved):
Other than Edgar's comment LGTM

src/clusterfuzz/_internal/swarming/service.py
fernandofloresg: I think what Ivan is doing is this flow is whenever an error is detected he raises an Exception, rather than logging it as an error, apparently what happens after this part is that the job is added to "unscheduled" tasks, my question would be, what happens with unscheduled tasks later on the flow?

IvanBM18: When the task is unscheduled it just returns to the queue.
i moved this to a warning because:
- Errors should be things that we should immediately fix, warnings are just to keep note of them but they don't brake anything, in this case a non scheduled job doesn't break anything, but its worth noting 
- We already have the logs polluted with to many errors

cc: @Xeicker @fernandofloresg

letitz: In this case we don't expect to ever be able to push this task if we're missing required information. So once the task returns to the queue, it will be stuck in a never-ending loop of getting dequeued and re-enqueued.

I'm not sure how @javanlacerda would like to handle such cases.

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -99,42 +97,53 @@ def create_utask_main_jobs(self,
...
+      # filled up on subsequent iterations due to other schedulers.
+      count_request = swarming_pb2.TasksCountRequest(  # pylint: disable=no-member
+          tags=[f'pool:{pool_val}', f'os:{os_val}'],
+          state=swarming_pb2.QUERY_PENDING)  # pylint: disable=no-member
+
+      # If the queue is full, there is no sense in continuing to schedule
+      # tasks in this session, so we return the remaining tasks as
+      # unscheduled.
+      if self._is_queue_full(count_request):
+        unscheduled_tasks.extend(remote_tasks[i:])
```
letitz, line 133: A passing thought: we will likely want to have some kind of exponential backoff here to avoid hammering Swarming with repeated CountTasks requests when all bots are busy. Let's see how it behaves in production.

src/clusterfuzz/_internal/swarming/service.py
```diff
@@ -99,42 +97,53 @@ def create_utask_main_jobs(self,
...
+        break
+
+      try:
+        response = self._api.push_task(task_req)
+        if not response or not response.task_id:
```
letitz: Even this `if` does not have to be in the try block.

letitz (approved):
LGTM overall, a couple comments below.

src/clusterfuzz/_internal/swarming/service.py
IvanBM18: Adding a TODO, for Memoize the back pressure calculation

src/clusterfuzz/_internal/swarming/service.py
javanlacerda: I think in this case we can log it as an error and discard the task to avoid the infinite loop Tituoan mentioned.

javanlacerda (approved):

PR #5290, opened 2026-05-22
Fix UWORKER initialization crash by bypassing Wi-Fi Datastore queries in Android

jardondiego:
### Overview
This PR fixes a critical initialization crash that occurs when untrusted workers (UWORKERs) attempt to set up Android emulators.

### The Bug
During the Android device initialization sequence, wifi.configure() is called to ensure the device is connected to the network. This function attempts to retrieve Wi-Fi credentials (SSID/password) by calling db_config.get(), which queries Google Cloud Datastore.

However, UWORKER containers run under a restricted security boundary and are explicitly denied IAM permissions to access Datastore (to prevent malicious fuzzers from escalating privileges or stealing project secrets). When the Datastore SDK intercepts the request, it returns a 403 Missing or insufficient permissions exception. Because this exception is
unhandled, it violently crashes the entire setup process, preventing the UWORKER from ever reaching the fuzzing loop.

### The Fix
This PR introduces an environment check at the top of wifi.configure():
```
if environment.is_uworker():
   return
```
This gracefully short-circuits the function for untrusted workers, bypassing the illegal Datastore query.

### Notes
  1. Networking works without it: Android emulators executed via Swarming in this architecture share the host network namespace (--network=host). They inherit outbound internet access directly from the Swarming bot's host machine and do not require explicit Wi-Fi credentials to be injected.
  2. Established Pattern: This mirrors the exact same security bypass we already use in device.add_test_accounts_if_needed(), which also short-circuits to avoid Datastore queries in UWORKERs.

### DIFF
--- a/src/clusterfuzz/_internal/tests/core/platforms/android/wifi_test.py
+++ b/src/clusterfuzz/_internal/tests/core/platforms/android/wifi_test.py
@@ -1,4 +1,4 @@
-# Copyright 2024 Google LLC
+# Copyright 2026 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/src/clusterfuzz/_internal/tests/core/platforms/android/wifi_test.py
+++ b/src/clusterfuzz/_internal/tests/core/platforms/android/wifi_test.py
@@ -1,0 +1,40 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/tests/core/platforms/android/wifi_test.py
```diff
@@ -0,0 +1,40 @@
+# Copyright 2024 Google LLC
```
fernandofloresg:
```suggestion
# Copyright 2026 Google LLC
```

fernandofloresg (approved):
lgtm

Xeicker:
Question, from your notes I see that in swarming we are sure this wouldn't cause an issue. But aren't there uworkers out of swarming that could still need to set up wifi?

IvanBM18 (approved):
LGTM

jardondiego:
Chnages have now been in dev for several hours. Metrics and logs are ok.

javanlacerda (approved):

PR #5293, opened 2026-05-25
Fix bypass Datastore access in uworker for job definition

jardondiego:
### Problem

The function `get_value_from_job_definition` in data_handler.py can cause crashes when executed within a UWORKER (untrusted worker) environment.

The function originally executed a query against Google Cloud Datastore:

`job = data_types.Job.query(data_types.Job.name == job_type).get()`

Because the UWORKER Docker container uses a restricted IAM Service Account ( untrusted-worker@... ), it has zero permissions to read from Datastore. This caused a fatal  403 PermissionDenied  exception, crashing processes like the log upload.

### Fix

When a task runs as a UWORKER, the Trusted Worker (TWORKER) that scheduled the task has already queried Datastore, extracted all the Job's environment variables, and injected them directly into the UWORKER's local environment. Therefore, querying Datastore from the UWORKER is both forbidden and redundant.

### DIFF
--- a/src/clusterfuzz/_internal/datastore/data_handler.py
+++ b/src/clusterfuzz/_internal/datastore/data_handler.py
@@ -1262,6 +1262,11 @@ def get_value_from_job_definition(job_type, variable_pattern, default=None):
   if not job_type:
     return default
 
+  # Short-circuit: UWORKERs do not have Datastore access. The TWORKER has
+  # already injected all job variables into the local environment.
+  if environment.is_uworker():
+    return environment.get_value(variable_pattern, default)
+
   job = data_types.Job.query(data_types.Job.name == job_type).get()
   if not job:
     return default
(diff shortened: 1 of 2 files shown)


fernandofloresg (approved):
lgtm

Xeicker (approved):
LGTM

IvanBM18 (approved):

IvanBM18:
Note: Please provide evidences that your changes work fine in dev and don't break anything so that the GOSST team feels confident

jardondiego:
> Note: Please provide evidences that your changes work fine in dev and don't break anything so that the GOSST team feels confident

Sounds good! Changes will stay in dev for a while and I will attach evidences here.

jardondiego:
Metrics and logs in `dev` are stable. These changes have been present for almost 20 hours.

javanlacerda (approved):

PR #5295, opened 2026-05-26
Adds option to override the default queue for tworkers

IvanBM18:
We needed a way of making some tworker bots to always pull from the same queue, so this way we can boot up some tworkers that will only pull from swarming's preprocess queue, to achieve this we added a way to override the tworkers default behavior by forcing it to always pull from the given queue.

## Changes
- Adds an option at `tworker_get_task()` of always selecting a queue when:
  - the `OVERRIDE_TWORKER_QUEUE` env var is set for the given bot.
  - the metadata key 'override_tworker_queue' is present in this instance metadata
  - the new method is memoized because we don't expect this value to change mid flight, it must change trough terraform config files
- Create some unit tests for this new behavior

## Tests performed
Currently theres one tworker in `dev` that its running with this new code, and it works perfectly fine its readding tasks from the given queue
image

Note: the error you see comes from the fact that the `utask_main` queue is currently full in dev which is an expected behavior right now.

Heres a log of what happens when the `utask_main` queue is not full:
image

### DIFF
--- a/src/clusterfuzz/_internal/base/tasks/__init__.py
+++ b/src/clusterfuzz/_internal/base/tasks/__init__.py
@@ -24,6 +24,7 @@ from typing import List
 from typing import Optional
 
 from google.cloud import monitoring_v3
+from requests import exceptions
 
 from clusterfuzz._internal.base import external_tasks
 from clusterfuzz._internal.base import memoize
@@ -34,6 +35,7 @@ from clusterfuzz._internal.config import local_config
 from clusterfuzz._internal.datastore import data_types
 from clusterfuzz._internal.datastore import ndb_utils
 from clusterfuzz._internal.fuzzing import fuzzer_selection
+from clusterfuzz._internal.google_cloud_utils import compute_metadata
 from clusterfuzz._internal.google_cloud_utils import credentials
 from clusterfuzz._internal.google_cloud_utils import pubsub
 from clusterfuzz._internal.google_cloud_utils import storage
@@ -364,6 +366,28 @@ def get_preprocess_task():
   return task
 
 
+@memoize.wrap(memoize.FifoInMemory(1))
+def _get_tworker_queue_override() -> str:
+  """Gets the tworker queue override from environment or metadata."""
+  queue_override = environment.get_value('OVERRIDE_QUEUE')
+  if queue_override:
+    return queue_override
+
+  if not compute_metadata.is_gce():
+    return ""
+
+  try:
+    queue_override = compute_metadata.get('instance/attributes/override_queue')
+    if queue_override:
+      return queue_override.strip()
+  except exceptions.RequestException as e:
+    if not (isinstance(e, exceptions.HTTPError) and
+            e.response.status_code == 404):
+      logs.warning(f'Error fetching override_queue metadata: {e}')
+
+  return ""
+
+
 def tworker_get_task():
   """Gets a task for a tworker to do."""
   assert environment.is_tworker()
@@ -372,8 +396,9 @@ def tworker_get_task():
   # queue that is probably empty) to do a single preprocess. Investigate
   # combining preprocess and postprocess queues and allowing pulling of
   # multiple messages.
-  queue_override = environment.get_value('OVERRIDE_QUEUE')
+  queue_override = _get_tworker_queue_override()
   if queue_override:
+    logs.info(f'Overriding default queue to {queue_override}')
     return get_regular_task(queue=queue_override)
 
   if random.random() < .5:
--- a/src/clusterfuzz/_internal/base/tasks/__init__.py
+++ b/src/clusterfuzz/_internal/base/tasks/__init__.py
@@ -24,7 +24,6 @@ from typing import List
 from typing import Optional
 
 from google.cloud import monitoring_v3
-from requests import exceptions
 
 from clusterfuzz._internal.base import external_tasks
 from clusterfuzz._internal.base import memoize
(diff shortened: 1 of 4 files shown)


fernandofloresg (approved):
lgtm, will we need to update documentation somewhere about this new env var?

IvanBM18:
> lgtm, will we need to update documentation somewhere about this new env var?

@javanlacerda any suggestions? i can add python docs but im not sure if this is enough, maybe [here](https://google.github.io/clusterfuzz/architecture/#fuzzing-bots) under non-preemtible bots?

src/clusterfuzz/_internal/base/tasks/__init__.py
```diff
@@ -364,6 +366,28 @@ def get_preprocess_task():
...
+      return queue_override.strip()
+  except exceptions.RequestException as e:
+    if not (isinstance(e, exceptions.HTTPError) and
+            e.response.status_code == 404):
+      logs.warning(f'Error fetching override_queue metadata: {e}')
+
+  return ""
+
+
 def tworker_get_task():
```
jardondiego: Did you consider defining it as an argument for this function instead? It could default to `None` in case `OVERRIDE_QUEUE` is not set. What do you think?

src/clusterfuzz/_internal/base/tasks/__init__.py
```diff
@@ -364,6 +366,28 @@ def get_preprocess_task():
   return task
 
 
+@memoize.wrap(memoize.FifoInMemory(1))
+def _get_tworker_queue_override() -> str:
+  """Gets the tworker queue override from environment or metadata."""
+  queue_override = environment.get_value('OVERRIDE_QUEUE')
```
jardondiego: Let's make this envvar's name more descriptive, what about `OVERRIDE_TWORKER_QUEUE`?

IvanBM18: Sure thing :D, ill do the same for the metadata key name

src/clusterfuzz/_internal/base/tasks/__init__.py
IvanBM18: aaa sure we can do that

jardondiego (approved):
LGTM

src/python/bot/startup/run_bot.py
```diff
@@ -126,6 +129,29 @@ def _get_max_task_executions():  # pylint: disable=inconsistent-return-statement
...
+  """Gets the tworker queue override from environment or metadata."""
+  queue_override = environment.get_value('OVERRIDE_TWORKER_QUEUE')
+  if queue_override:
+    return queue_override
+
+  if not compute_metadata.is_gce():
+    return ""
+
+  try:
+    queue_override = compute_metadata.get(
```
letitz, line 143: [Optional] The pattern I've seen (and used) in the past for this stuff is to do the metadata lookup in the startup script that wraps python code, and set the environment variable based on that: https://github.com/google/clusterfuzz/blob/027bee49c54318d73972e5730bc8c1a56a246224/docker/base/setup_clusterfuzz.sh#L24

Then the python code can just use the environment variable and avoid doing the lookup.

src/python/bot/startup/run_bot.py
```diff
@@ -126,6 +129,29 @@ def _get_max_task_executions():  # pylint: disable=inconsistent-return-statement
     logs.log_fatal_and_exit(f'Invalid value for MAX_TASK_EXECUTIONS: {val}')
 
 
+@memoize.wrap(memoize.FifoInMemory(1))
```
letitz, line 132: [Optional] This totally works, even though I find it a bit non-obvious. Consider setting and reading a global variable instead.

src/clusterfuzz/_internal/base/tasks/__init__.py
```diff
@@ -369,14 +370,18 @@ def get_preprocess_task():
...
   """Gets a task for a tworker to do."""
   assert environment.is_tworker()
   # TODO(metzman): Pulling tasks is relatively expensive compared to
   # preprocessing. It's too expensive to pull twice (once from the postproces
   # queue that is probably empty) to do a single preprocess. Investigate
   # combining preprocess and postprocess queues and allowing pulling of
   # multiple messages.
+  if override_queue:
+    logs.info(f'Overriding default queue to {override_queue}')
+    return get_regular_task(queue=override_queue)
```
letitz, line 383: This should have a matching test in `tasks_test.py`.

letitz (approved):
LGTM  % adding a test, two other optional comments.

src/python/bot/startup/run_bot.py
IvanBM18: Agree is not that obvious,  but i prefer this method as its the one commonly used in all of clusterfuzz code and easier to maintain if at any point GOSST decides to improve/remove it.

src/clusterfuzz/_internal/base/tasks/__init__.py
IvanBM18: Sure :D

src/python/bot/startup/run_bot.py
IvanBM18: Good suggestion, but making changes to the setup scripts in the docker images are not reflected until we update the docker image, hence moving it there will require additional effort.

But ill consideer it for future prs thanks :D

javanlacerda (approved):
lgtm

PR #5297, opened 2026-05-29
Enforce testcase-level access control for testcase blob downloads in /download

evilgensec:
### Summary

`handlers/download.py` serves testcase-scoped blobs (a testcase's minimized/fuzzed reproducer) after only the general `access.has_access()` check, consulting the testcase-level `access.can_user_access_testcase()` check only as a fallback:

```python
if access.has_access():
    return self._send_blob(...)
if not testcase:
if access.can_user_access_testcase(testcase):   # only reached if has_access() was false
    return self._send_blob(...)
```

The sibling handlers that serve the same testcase blobs gate testcase content on `can_user_access_testcase()` first:

- `handlers/viewer.py` — `if testcase_id: if not access.can_user_access_testcase(testcase): raise AccessDeniedError()`; `has_access()` is used only for non-testcase content.
- `handlers/testcase_detail/download_testcase.py` — `access.check_access_and_get_testcase(testcase_id)`.

`can_user_access_testcase()` escalates to privileged access for security testcases (`need_privileged_access = testcase.security_flag and not config.relax_security_bug_restrictions`), whereas `has_access()` (`get_access(need_privileged_access=False)`) also returns `Allowed` for any `whitelisted_domains` user. As a result, on `/download` a user with general access but without testcase-level access can download a testcase reproducer they are not permitted to view via `/viewer`.

### Change

Gate testcase-scoped blobs on `can_user_access_testcase()` before the general `has_access()` check, matching `viewer.py`; retain `has_access()` for non-testcase blobs. Privileged users, the testcase uploader, and associated-bug participants are unaffected (they pass `can_user_access_testcase()`); domain-allowed users keep access to non-security testcases. The public OSS-Fuzz path (`check_public_testcase`) is unchanged and runs earlier.

`handlers/download_test.py`'s `test_download_user` is updated to assert the corrected behavior: general access alone grants non-testcase blobs, and testcase blobs require testcase-level access.

Note: the full handler suite was not run locally (it requires the datastore emulator); the change is a structural alignment with `viewer.py` and the test mirrors existing patterns in the file.

### DIFF
--- a/src/appengine/handlers/crash_query.py
+++ b/src/appengine/handlers/crash_query.py
@@ -19,6 +19,7 @@ from clusterfuzz._internal.crash_analysis import crash_analyzer
 from clusterfuzz._internal.crash_analysis.stack_parsing import stack_analyzer
 from clusterfuzz._internal.datastore import data_handler
 from handlers import base_handler
+from libs import access
 from libs import auth
 from libs import handler
 from libs import helpers
@@ -55,6 +56,11 @@ class Handler(base_handler.Handler):
 
     duplicate_testcase = data_handler.find_testcase(
         project, state.crash_type, state.crash_state, security_flag)
+    if (duplicate_testcase and duplicate_testcase.security_flag and
+        not access.can_user_access_testcase(duplicate_testcase)):
+      # Do not disclose the existence or identifiers of a security-confidential
+      # testcase to a user who is not allowed to access it.
+      duplicate_testcase = None
     if duplicate_testcase:
       result['result'] = 'duplicate'
       result['duplicate_id'] = duplicate_testcase.key.id()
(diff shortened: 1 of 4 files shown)


evilgensec:
Added a second commit that applies the same `security_flag`-aware scoping to the `/crash-query` duplicate-detection handler (`crash_query.py`): it returned a matching testcase's id and bug id to any authenticated user, without the `crash_access.add_scope`/`can_user_access_testcase` check the sibling query and per-testcase handlers use. The handler now suppresses a security-confidential duplicate when the caller cannot access it, with a regression test.

vitaliset (approved):
LGTM, thanks @evilgensec!

cc: @PauloVLB

PauloVLB (approved):

PR #5302, opened 2026-06-01
Enforce testcase access control in the crash-query endpoint

adilburaksen:
## Problem

The `/crash-query` endpoint (`crash_query.py`) is gated only by a logged-in check (`auth.get_current_user()`). It takes a caller-supplied stacktrace, computes the crash signature and `security_flag`, calls `data_handler.find_testcase(project, crash_type, crash_state, security_flag)` (which matches security testcases) and returns the matched testcase's `duplicate_id` and `bug_id`.

Every other crash-data path applies `crash_access.add_scope()`, which hides `security_flag=True` testcases from non-privileged users (including domain-allowed "everything-but-security" users): `testcase_list.py`, `crash_stats.py`, `commit_range.py`. `crash_query` applies no per-result access check, so a non-privileged user can confirm the existence of, and obtain the testcase id and linked bug id of, a restricted security testcase that the rest of the UI hides from them.

## Fix

Gate the duplicate disclosure behind `access.can_user_access_testcase()` (the same check used by `check_access_and_get_testcase`, which enforces `need_privileged_access` for `security_flag` testcases plus uploader/issue-owner relations). If the current user cannot access the matched testcase, respond as if no duplicate exists, consistent with that user's view everywhere else.

The computed `security` boolean (derived from the caller's own submitted stacktrace) is unchanged.

### DIFF
--- a/src/appengine/handlers/crash_query.py
+++ b/src/appengine/handlers/crash_query.py
@@ -19,6 +19,7 @@ from clusterfuzz._internal.crash_analysis import crash_analyzer
 from clusterfuzz._internal.crash_analysis.stack_parsing import stack_analyzer
 from clusterfuzz._internal.datastore import data_handler
 from handlers import base_handler
+from libs import access
 from libs import auth
 from libs import handler
 from libs import helpers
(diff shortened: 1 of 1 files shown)


adilburaksen:
Closing in favor of #5297, which already adds the same testcase-level access enforcement to crash_query.py (and the related /download fix). Deferring to that PR.

PR #5303, opened 2026-06-01
Anchor CORS Origin allowlist matching (re.fullmatch)

adilburaksen:
## Problem

`allowed_cors` (`libs/handler.py`) matches the request `Origin` against each configured `whitelisted_cors_urls` regex with `re.match`, which only anchors at the start of the string. On a match it reflects the origin into `Access-Control-Allow-Origin` together with `Access-Control-Allow-Credentials: true`.

With the shipped example pattern `https?://(.*-dot-)?test-client-site.appspot.com`, `re.match` also matches malicious origins such as:

- `https://test-client-site.appspot.com.evil.com` (trailing suffix)
- `https://anything.evil.com-dot-test-client-site.appspot.com` (the `.*` prefix)
- `https://test-client-siteXappspot.com` (unescaped `.`)

A page on such an origin can then make credentialed cross-origin requests to the CORS-enabled endpoint and read the victim's authorized crash/testcase data.

## Fix

- Use `re.fullmatch` so the pattern must match the entire `Origin` (anchors both ends), closing the trailing-suffix and prefix bypasses.
- Update the example regexes in `configs/test/gae/auth.yaml` to escape literal dots and document the requirement, so copied configs are safe by default.

### DIFF
--- a/configs/test/gae/auth.yaml
+++ b/configs/test/gae/auth.yaml
@@ -27,9 +27,12 @@ whitelisted_oauth_emails:
 #  - test-client-email1@developer.gserviceaccount.com
 
 # CORS urls that have access to some pages in ClusterFuzz (e.g. Crashes by range).
+# These are full-match regexes against the request Origin, so escape literal
+# dots and avoid trailing wildcards to prevent origins such as
+# 'https://test-client-site.appspot.com.evil.com' from being whitelisted.
 whitelisted_cors_urls:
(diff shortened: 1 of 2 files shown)


adilburaksen:
@ViniciustCosta gentle nudge on this one — the CORS Origin allowlist uses a loose match, so origins like evil-allowed.com slip through. Switching to re.fullmatch anchors it. One-liner plus a test. Let me know if you'd prefer a different matching approach.

Issue #5258, opened 2026-04-28
Two handlers missing access checks: `issue_redirector` and `coverage_report`


sammiee5311:
Two web handlers don't have access checks. Same class as #5196 (task-log, fixed in #5243).

### `/issue/<testcase_id>`

[src/appengine/handlers/issue_redirector.py:22-33](https://github.com/google/clusterfuzz/blob/master/src/appengine/handlers/issue_redirector.py#L22-#L33)

```python
class Handler(base_handler.Handler):
  """Handler that redirects user to the issue URL."""

  def get(self, testcase_id=None):
    """Redirect user to the correct URL."""
    testcase = helpers.get_testcase(testcase_id)
    issue_url = helpers.get_or_exit(
        lambda: issue_tracker_utils.get_issue_url(testcase),
        'Issue tracker for testcase (id=%s) is not found.' % testcase_id,
        'Failed to get the issue tracker URL.')

    return self.redirect(issue_url)
```

`helpers.get_testcase()` fetches by ID without any access control. Every other testcase handler uses `access.check_access_and_get_testcase()`. See [testcase_detail/show.py:303](https://github.com/google/clusterfuzz/blob/master/src/appengine/handlers/testcase_detail/show.py#L303).

So an unauthenticated caller can iterate testcase IDs and pull the bug tracker URL for any testcase, including security-sensitive ones. On IAP-protected deployments, any IAP-authenticated user can read URLs for testcases scoped to projects they don't have access to. There's no per-testcase check.

### `/coverage-report/`

[src/appengine/handlers/coverage_report.py:69-80](https://github.com/google/clusterfuzz/blob/master/src/appengine/handlers/coverage_report.py#L69-#L80)

```python
class Handler(base_handler.Handler):
  """Coverage Report Handler."""

  # pylint: disable=unused-argument
  @handler.get(handler.HTML)
  @handler.oauth
  def get(self, report_type=None, argument=None, date=None, extra=None):
    """Handle a get request."""
    report_url = get_report_url(report_type, argument, date)
    if report_url:
      return self.redirect(report_url)
    raise helpers.EarlyExitError('Failed to get coverage report.', 400)
```

[...]

PR #5304, opened 2026-06-01
Security: Add missing authorization checks to web handlers

Ashutosh0x:
## Summary

Add missing authorization checks to three web handlers in ClusterFuzz to prevent unauthorized access to testcase data, coverage reports, and cross-object issue manipulation.

### 1. issue_redirector.py - Missing Access Control
The handler uses `helpers.get_testcase()` which only checks if a testcase exists, NOT if the current user has permission to view it. This leaks private issue tracker URLs to unauthorized users.

**Fix:** Replaced with `access.check_access_and_get_testcase()` which enforces proper permission checks.

### 2. coverage_report.py - Missing Job-Level Authorization
Any authenticated user can request coverage reports for any fuzzer job across any project, regardless of their access level.

**Fix:** Added `access.has_access(job_type=job)` check before serving coverage data.

### 3. update_issue.py - Cross-Object Authorization Bypass (IDOR)
The handler checks if the user has access to the testcase, but NOT to the target issue. Since it runs as a high-privilege service account, it can post comments and modify metadata on private issues in other projects.

**Fix:** Added user-association check verifying the requesting user is an admin, privileged user, or explicitly associated (reporter/assignee/CC) with the target issue.

## Impact

Without these fixes:
- Unauthenticated users can enumerate testcase IDs and discover private issue URLs
- Any user can access coverage data for restricted projects
- Users with testcase access can modify unrelated private issues

## Related

This PR extends the scope of #5258 by also fixing the cross-object authorization bypass in `update_issue.py`, which is not covered by PR #5259.

### DIFF
--- a/src/appengine/handlers/coverage_report.py
+++ b/src/appengine/handlers/coverage_report.py
@@ -20,6 +20,7 @@ from clusterfuzz._internal.datastore import data_handler
 from clusterfuzz._internal.datastore import data_types
 from clusterfuzz._internal.metrics import fuzzer_stats
 from handlers import base_handler
+from libs import access
 from libs import handler
 from libs import helpers
 
@@ -74,6 +75,10 @@ class Handler(base_handler.Handler):
   @handler.oauth
   def get(self, report_type=None, argument=None, date=None, extra=None):
     """Handle a get request."""
+    job = argument
+    if job and not access.has_access(job_type=job):
+      raise helpers.AccessDeniedError()
+
     report_url = get_report_url(report_type, argument, date)
     if report_url:
       return self.redirect(report_url)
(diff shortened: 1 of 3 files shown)


Ashutosh0x:
@oliverchang @jonathanmetzman Hi! This PR adds missing authorization checks to three web handlers, extending the scope of #5258 / PR #5259:

1. **issue_redirector.py** - Replaced `helpers.get_testcase()` with `access.check_access_and_get_testcase()` (same fix as PR #5259)
2. **coverage_report.py** - Added `access.has_access(job_type=job)` check (same fix as PR #5259)
3. **update_issue.py** - **NEW FIX** not covered by PR #5259: Added user-association check to prevent cross-object IDOR (testcase access allowing modification of unrelated private issues)

The `update_issue.py` fix is the key addition here - it prevents a cross-object authorization bypass where testcase-scoped access incorrectly allows updating unrelated issue IDs via the privileged service account.

CLA is signed. All basic checks passed. Could you please review and trigger the CI test suite?

PR #5305, opened 2026-06-01
Support fuzz_targets property in clusterfuzz_manifest.json

notvictorl:
Adds support for `fuzz_targets` property in `clusterfuzz_manifest.json`. This allows coverage-guided fuzzing jobs with `ChromeBuildArchive`s to skip the unpacking and fuzz target discovery phases.

**Problem:**
Currently, ClusterFuzz wastes time unpacking complete archives to then traverse the directory to discover fuzz targets. Furthermore, on certain large build types, this unpacking process frequently triggers out-of-disk-space errors due to the massive size of the zip files.

**Solution:**
By introducing `fuzz_targets` — a list of fuzz target paths relative to `clusterfuzz_manifest.json` — we can bypass the discovery step entirely. ClusterFuzz will now map this list so that each fuzz target has a normalized name (key) and its corresponding path (value).

More Info: crbug.com/508214240

### DIFF
--- a/src/clusterfuzz/_internal/build_management/build_archive.py
+++ b/src/clusterfuzz/_internal/build_management/build_archive.py
@@ -62,6 +62,7 @@ class BuildArchive(archive.ArchiveReader):
 
   def __init__(self, reader: archive.ArchiveReader):
     self._reader = reader
+    self._fuzz_targets = None
 
   def list_members(self) -> List[archive.ArchiveMemberInfo]:
     return self._reader.list_members()
@@ -78,12 +79,33 @@ class BuildArchive(archive.ArchiveReader):
   def close(self) -> None:
     self._reader.close()
 
-  @abc.abstractmethod
   def list_fuzz_targets(self) -> List[str]:
     """Lists fuzzing targets in the archive.
 
+    Guarantees that self._fuzz_targets is populated with a dict mapping
+    normalized target names to their paths in the archive.
+
     Returns:
         The list of fuzz targets.
+    """
+    if self._fuzz_targets is not None:
+      return list(self._fuzz_targets.keys())
+
+    # Import here as this path is not available in App Engine context.
+    from clusterfuzz._internal.bot.fuzzers import utils as fuzzer_utils
+
+    self._fuzz_targets = {
+        fuzzer_utils.normalize_target_name(path): path
+        for path in self.find_fuzz_targets()
+    }
+    return list(self._fuzz_targets.keys())
+
+  @abc.abstractmethod
+  def find_fuzz_targets(self) -> List[str]:
+    """Finds fuzzing targets in the archive.
+
+    Returns:
+        The list of paths to fuzz targets in the archive.
     """
     raise NotImplementedError
 
@@ -139,10 +161,6 @@ class DefaultBuildArchive(BuildArchive):
   """Default class for handling builds. This should work with everything.
   """
 
-  def __init__(self, reader: archive.ArchiveReader):
-    super().__init__(reader)
-    self._fuzz_targets = None
-
   def get_path_for_target(self, fuzz_target: str) -> str:
     """Returns the path in the archive of the fuzz_target if found.
     This is needed because target name normalization means we're losing
@@ -151,10 +169,7 @@ class DefaultBuildArchive(BuildArchive):
     if self._fuzz_targets is None:
       self.list_fuzz_targets()
 
-    if not self._fuzz_targets or fuzz_target not in self._fuzz_targets:
-      return None
-
-    return self._fuzz_targets[fuzz_target]
+    return self._fuzz_targets.get(fuzz_target)
 
   def get_target_dependencies(
       self, fuzz_target: str) -> List[archive.ArchiveMemberInfo]:
@@ -184,20 +199,20 @@ class DefaultBuildArchive(BuildArchive):
 
     return to_extract
 
-  def list_fuzz_targets(self) -> List[str]:
-    if self._fuzz_targets is not None:
-      return list(self._fuzz_targets.keys())
+  def find_fuzz_targets(self) -> List[str]:
+    """Finds fuzzing targets in the archive.
 
-    self._fuzz_targets = {}
+    Returns:
+        The list of paths to fuzz targets in the archive.
+    """
     # Import here as this path is not available in App Engine context.
     from clusterfuzz._internal.bot.fuzzers import utils as fuzzer_utils
 
-    for archive_file in self.list_members():
-      if fuzzer_utils.is_fuzz_target(archive_file.name, self.open):
-        fuzz_target = fuzzer_utils.normalize_target_name(archive_file.name)
-        self._fuzz_targets[fuzz_target] = archive_file.name
-
-    return list(self._fuzz_targets.keys())
+    return [
+        member.name
+        for member in self.list_members()
+        if fuzzer_utils.is_fuzz_target(member.name, self.open)
+    ]
 
   def unpacked_size(self, fuzz_target: Optional[str] = None) -> int:
     if not fuzz_target:
@@ -294,35 +309,39 @@ class ChromeBuildArchive(DefaultBuildArchive):
         to expect if `clusterfuzz_manifest.json` is missing or badly formatted.
     """
     super().__init__(reader)
+    self._manifest_fuzz_targets = None
     # The manifest may not exist for earlier versions of archives. In this
     # case, default to schema version 0.
     manifest_path = 'clusterfuzz_manifest.json'
-    if self.file_exists(manifest_path):
-      with self.open(manifest_path) as f:
-        manifest = json.load(f)
-      self._archive_schema_version = manifest.get('archive_schema_version')
-      if self._archive_schema_version is None:
-        logs.warning(
-            'clusterfuzz_manifest.json was incorrectly formatted or missing an '
-            'archive_schema_version field')
-        self._archive_schema_version = default_archive_schema_version
-
-      fuzz_targets_data = manifest.get('fuzz_targets')
-      if isinstance(fuzz_targets_data, list):
-        self._fuzz_targets = {}
-        from clusterfuzz._internal.bot.fuzzers import utils as fuzzer_utils
-        for target_path in fuzz_targets_data:
-          if isinstance(target_path, str):
-            fuzz_target = fuzzer_utils.normalize_target_name(target_path)
-            self._fuzz_targets[fuzz_target] = target_path
-          else:
-            logs.error(
-                f'Entry in fuzz_targets (clusterfuzz_manifest.json) is not a '
-                f'string: {target_path}')
-      elif fuzz_targets_data is not None:
-        logs.error('fuzz_targets in clusterfuzz_manifest.json is not a list')
-    else:
+    if not self.file_exists(manifest_path):
       self._archive_schema_version = default_archive_schema_version
+      return
+
+    with self.open(manifest_path) as f:
+      manifest = json.load(f)
+    self._archive_schema_version = manifest.get('archive_schema_version')
+    if self._archive_schema_version is None:
+      logs.warning(
+          'clusterfuzz_manifest.json was incorrectly formatted or missing an '
+          'archive_schema_version field')
+      self._archive_schema_version = default_archive_schema_version
+
+    fuzz_target_paths = manifest.get('fuzz_targets')
+    if fuzz_target_paths is None:
+      return
+
+    if not isinstance(fuzz_target_paths, list):
+      logs.error('fuzz_targets in clusterfuzz_manifest.json is not a list')
+      return
+
+    self._manifest_fuzz_targets = []
+    for target_path in fuzz_target_paths:
+      if isinstance(target_path, str):
+        self._manifest_fuzz_targets.append(target_path)
+      else:
+        logs.error(
+            f'Entry in fuzz_targets (clusterfuzz_manifest.json) is not a '
+            f'string: {target_path}')
 
   def root_dir(self) -> str:
     if not hasattr(self, '_root_dir'):
@@ -332,6 +351,11 @@ class ChromeBuildArchive(DefaultBuildArchive):
   def archive_schema_version(self) -> int:
     """Returns the schema version number for this archive."""
     return self._archive_schema_version
+
+  def find_fuzz_targets(self) -> List[str]:
+    if self._manifest_fuzz_targets:
+      return self._manifest_fuzz_targets
+    return super().find_fuzz_targets()
 
   def get_dependency_path(self, path: str, deps_file_path: str) -> str:
     """Deps are given as paths relative to the deps file where they are listed,
--- a/src/clusterfuzz/_internal/build_management/build_archive.py
+++ b/src/clusterfuzz/_internal/build_management/build_archive.py
@@ -88,16 +88,15 @@ class BuildArchive(archive.ArchiveReader):
     Returns:
         The list of fuzz targets.
     """
-    if self._fuzz_targets is not None:
-      return list(self._fuzz_targets.keys())
+    if self._fuzz_targets is None:
+      # Import here as this path is not available in App Engine context.
+      from clusterfuzz._internal.bot.fuzzers import utils as fuzzer_utils
 
-    # Import here as this path is not available in App Engine context.
-    from clusterfuzz._internal.bot.fuzzers import utils as fuzzer_utils
+      self._fuzz_targets = {
+          fuzzer_utils.normalize_target_name(path): path
+          for path in self.find_fuzz_targets()
+      }
 
-    self._fuzz_targets = {
-        fuzzer_utils.normalize_target_name(path): path
-        for path in self.find_fuzz_targets()
-    }
     return list(self._fuzz_targets.keys())
 
   @abc.abstractmethod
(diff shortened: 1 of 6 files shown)


src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -141,17 +141,17 @@ class DefaultBuildArchive(BuildArchive):
...
     """Returns the path in the archive of the fuzz_target if found.
     This is needed because target name normalization means we're losing
     information about the actual file reprensenting the fuzz_target.
     """
-    if not self._fuzz_targets:
+    if self._fuzz_targets is None:
       self.list_fuzz_targets()
 
-    if not fuzz_target in self._fuzz_targets:
+    if not self._fuzz_targets or fuzz_target not in self._fuzz_targets:
```
letitz: I think we should assume and document in DefaultBuildArchive that list_fuzz_targets() always sets self._fuzz_targets to a dict value. Then we can skip this second check for not self._fuzz_targets. This ties in to my comment below about making `list_fuzz_targets()` a concrete method.

Also, we should use `return self._fuzz_targets.get(fuzz_target)` to avoid the double lookup.

src/clusterfuzz/_internal/build_management/build_archive.py
letitz: This branch has grown enough, let's invert the branch with an early return for the schema v0 case to reduce nesting.

```py
  if not self.file_exists(manifest_path):
    self._archive_schema_version = default_archive_schema_version
    return

  with self.open(manifest_path) as f:
    ...
```

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -304,6 +306,21 @@ def __init__(self,
             'clusterfuzz_manifest.json was incorrectly formatted or missing an '
             'archive_schema_version field')
         self._archive_schema_version = default_archive_schema_version
+
+      fuzz_targets_data = manifest.get('fuzz_targets')
+      if isinstance(fuzz_targets_data, list):
```
letitz: Again we can reduce nesting with an early return instead in the error case. Also a nit, let's name the variable `fuzz_target_paths`:

```py
if not isinstance(fuzz_targets_paths, list):
  logs.error("not a list")
  return
```

src/clusterfuzz/_internal/build_management/build_archive.py
letitz, line 59: Thinking aloud here. If you enjoy refactoring code and feel motivated to take this on, do feel free to, but it's straying off the path a fair bit so please also feel free to just ignore this.

As I review this, I'm coming to the conclusion that this abstraction is a bit wonky.

1. It really does not make much sense for it to inherit from `ArchiveReader`, at most it could expose `self._reader` through a getter but from skimming the code that instantiates `BuildArchive` objects [1], it does not seem we need to reach down into the reader anyway.
2. `unpacked_size()` and `unpack()` don't need to be abstract, they are only overridden once in the class hierarchy by `DefaultBuildArchive`. All that's left apart from that is `list_fuzz_targets()` and `get_target_dependencies()`.

The design would make more sense to me if we had:

* `ArchiveReader`: interface that abstracts over different file formats (tar vs zip), allows operations on files
* `FuzzTargetFinder`: interface that abstracts over how to find fuzz targets and their dependencies, given an archive reader
* `BuildArchive`: concrete class that abstracts over different directory layouts inside the archive, allows operations on the whole build and fuzz targets, using an injected `FuzzTargetFinder`.

Something like:

```py
class FuzzTargetFinder(abc.ABC):
  @abc.abstractmethod
  def find_targets(self, reader: ArchiveReader) -> list[os.PathLike]:
    pass

  @abc.abstractmethod
  def get_target_dependencies(self, reader: ArchiveReader, target: os.PathLike) -> list[os.PathLike]:
    pass

class DefaultFuzzTargetFinder(FuzzTargetFinder):
  pass

class ChromeFuzzTargetFinder(FuzzTargetFinder):
  pass

def make_fuzz_target_finder(archive_reader: ArchiveReader) -> FuzzTargetFinder:
  # check for args.gn, return either Finder class

class BuildArchive:
  def __init__(self, fuzz_target_finder: FuzzTargetFinder):
    self._fuzz_target_finder = fuzz_target_finder

  def archive_reader(self) -> ArchiveReader:
    return self._fuzz_target_finder.archive_reader
```

[1]: https://github.com/google/clusterfuzz/blob/a226d62be0683b57cea49feb75e42222bf5f0f59/src/clusterfuzz/_internal/build_management/build_manager.py#L522

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -185,8 +185,10 @@ def get_target_dependencies(
     return to_extract
 
   def list_fuzz_targets(self) -> List[str]:
```
letitz: We're using inheritance here and below to override the initialization behavior of `self._fuzz_targets`, which works but is a bit messy and duplicates a bit of logic. See my topmost comment for a completely optional suggestion on a refactor.

A simpler way to make things a bit better here without reworking the class hierarchy entirely is maybe to define an abstract pure method that just handles finding fuzz targets, and share the implementation of `list_fuzz_targets()` in `BuildArchive` above.

```py
class BuildArchive:
  def _init__(self, reader):
    self._reader = reader
    self._fuzz_targets = None

  def list_fuzz_targets():
    if self._fuzz_targets is None:
      self._fuzz_targets = {
        fuzzer_utils.normalize_target_name(path): path
        for path in self.find_fuzz_targets()
      }

    return list(self._fuzz_targets.keys())

  @abc.abstractmethod
  def find_fuzz_targets(self) -> list[os.PathLike]:
    pass

class DefaultBuildArchive(BuildArchive):
  def find_fuzz_targets(self):
    # use `is_fuzz_target()` to filter files in the archive

class ChromeBuildArchive(DefaultBuildArchive):
  def __init__(self, reader):
    self._fuzz_target_paths = # from the manifest

  def find_fuzz_targets(self):
    if self._fuzz_target_paths:
      return self._fuzz_target_paths

    return super().find_fuzz_targets()
```

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -304,6 +306,21 @@ def __init__(self,
...
+      fuzz_targets_data = manifest.get('fuzz_targets')
+      if isinstance(fuzz_targets_data, list):
+        self._fuzz_targets = {}
+        from clusterfuzz._internal.bot.fuzzers import utils as fuzzer_utils
+        for target_path in fuzz_targets_data:
+          if isinstance(target_path, str):
+            fuzz_target = fuzzer_utils.normalize_target_name(target_path)
+            self._fuzz_targets[fuzz_target] = target_path
+          else:
+            logs.error(
```
letitz: If the entire list consists of invalid values, the resulting dict will be empty. In that case, I think we should fall back to `DefaultBuildArchive`'s logic. Also if the list in the manifest is itself empty.

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -522,6 +522,19 @@ def _unpack_build(self,
       with self._open_build_archive(base_build_dir, build_dir, build_url,
                                     http_build_url) as build:
         unpack_start_time = time.time()
+        if environment.is_engine_fuzzer_job() and not self.fuzz_target:
+          # If no fuzz target is selected (e.g. during initial fuzz target
+          # discovery), we only need the list of fuzz targets. We get them
+          # from the build archive (which instantly reads
+          # clusterfuzz_manifest.json if present) and skip unpacking the
+          # archive to disk entirely to prevent out of disk space errors.
```
letitz: Can we share this logic with the branch below?

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -1351,7 +1364,8 @@ def setup_regular_build(revision,
         build.build_dir,  # Store inside the main build.
         revision,
         extra_build_url,
-        build_prefix=fuzzer_utils.EXTRA_BUILD_DIR)
+        build_prefix=fuzzer_utils.EXTRA_BUILD_DIR,
+        fuzz_target=fuzz_target)
```
letitz: I don't know about this. I don't  think we use `EXTRA_BUILD_BUCKET_PATH` at all in Chrome?

src/clusterfuzz/_internal/tests/core/build_management/build_archive_test.py
```diff
@@ -409,3 +409,58 @@ def test_manifest_is_correctly_read(self):
     self._generate_manifest(1)
     test_archive = build_archive.ChromeBuildArchive(self.mock.open.return_value)
     self.assertEqual(test_archive.archive_schema_version(), 1)
+
+  def _generate_manifest_with_fuzz_targets(self, archive_schema_version,
```
letitz: Let's collapse all these `_generate_manifest()` functions into a single one that expects a dict. The format is simple enough that we can repeat it in most tests:

```py
def _generate_manifest(self, contents: any):
     json_contents = json.dumps(contents).encode()
     def _mock_open(_):
      buffer = io.BytesIO(b'')
      buffer.write(json_contents)
      buffer.seek(0)
      return buffer

    self.mock.open.return_value.open.side_effect = _mock_open

# and call it like
def test_foo(self):
  self._generate_manifest({ 'archive_schema_version': 1 })
```

src/clusterfuzz/_internal/tests/core/build_management/build_archive_test.py
```diff
@@ -409,3 +409,58 @@ def test_manifest_is_correctly_read(self):
...
+    self.mock.file_exists.return_value = True
+    self._generate_manifest_with_fuzz_targets(
+        1, {'my_fuzzer': 'out/build/my_fuzzer'})
+    self.mock.open.return_value.list_members.return_value = []
+    test_archive = build_archive.ChromeBuildArchive(self.mock.open.return_value)
+    self.assertEqual(test_archive.archive_schema_version(), 1)
+    test_archive.list_fuzz_targets()
+    self.mock.open.return_value.list_members.assert_called_once()
+
+  def test_manifest_fuzz_targets_missing(self):
```
letitz, line 439: We should have a test for what happens when only *some* of the entries are invalid.

src/clusterfuzz/_internal/tests/core/build_management/build_archive_test.py
```diff
@@ -409,3 +409,58 @@ def test_manifest_is_correctly_read(self):
...
+
+    self.mock.open.return_value.open.side_effect = _mock_open
+
+  def test_manifest_fuzz_targets_list(self):
+    """Tests that fuzz_targets specified as a list in the manifest correctly
+    populates self._fuzz_targets and bypasses discovery."""
+    self.mock.file_exists.return_value = True
+    self._generate_manifest_with_fuzz_targets(
+        1, ['out/build/my_fuzzer', 'out/build/other_fuzzer'])
+    test_archive = build_archive.ChromeBuildArchive(self.mock.open.return_value)
```
letitz: [Optional] `self.mock.open.return_value` is mentioned many times, and its name is quite opaque. It may be worth defining either a local variable `mock_archive_reader` or even a test fixture class member `self.mock_archive_reader`.

src/clusterfuzz/_internal/tests/core/build_management/build_manager_test.py
```diff
@@ -741,7 +761,9 @@ def test_setup_fuzz_over_http(self):
     os.environ['ALLOW_UNPACK_OVER_HTTP'] = "True"
     self.mock.unzip_over_http_compatible.return_value = True
     self.mock.time.return_value = 1000.0
-    build = build_manager.setup_regular_build(2)
+    fuzz_target = build_manager.pick_random_fuzz_target(
```
letitz, line 767: This test may have been exercising the code path for blackbox fuzzing, and no longer is. Is there another test that checks what happens when `fuzz_target` is None? If not, we should duplicate the old test to conserve coverage.

src/clusterfuzz/_internal/tests/core/build_management/build_manager_test.py
```diff
@@ -778,7 +800,9 @@ def test_setup_fuzz_over_http_unpack_all(self):
         'https://example.com/path/file-release-([0-9]+).zip')
     self.mock.unzip_over_http_compatible.return_value = True
     self.mock.time.return_value = 1000.0
-    build = build_manager.setup_regular_build(2)
+    fuzz_target = build_manager.pick_random_fuzz_target(
```
letitz, line 807: Ditto

src/clusterfuzz/_internal/tests/core/build_management/build_archive_test.py
notvictorl: Refactored `self.mock.open.return_value` to `self.mock_archive_reader`.

src/clusterfuzz/_internal/build_management/build_archive.py
notvictorl: I agree a refactor on the abstraction would make a lot more sense, although I don't think it's best to implement it in this PR. I can set up a follow-up PR to clean up the abstraction.

letitz: Only if you feel like it and have the time, it's by no means necessary.

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -184,20 +198,20 @@ def get_target_dependencies(
...
-  def list_fuzz_targets(self) -> List[str]:
-    if self._fuzz_targets is not None:
-      return list(self._fuzz_targets.keys())
+  def find_fuzz_targets(self) -> List[str]:
+    """Finds fuzzing targets in the archive.
 
-    self._fuzz_targets = {}
+    Returns:
+        The list of paths to fuzz targets in the archive.
+    """
```
letitz: No need to document the overridden method again, except to note any important differences in behavior (usually there are none). Also, apparently this can and should be marked with `@override`: https://google.github.io/styleguide/pyguide.html#3831-overridden-methods

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -522,35 +522,31 @@ def _unpack_build(self,
...
-          # from the build archive (which instantly reads
-          # clusterfuzz_manifest.json if present) and skip unpacking the
-          # archive to disk entirely to prevent out of disk space errors.
+        if self.is_discovery or not self._unpack_everything:
+          # We need the list of fuzz targets, either because we are performing
+          # initial fuzz target discovery, or because we are only unpacking
+          # selected fuzz targets.
           list_fuzz_target_start_time = time.time()
           self._fuzz_targets = list(build.list_fuzz_targets())
           _emit_job_build_retrieval_metric(list_fuzz_target_start_time,
```
letitz, line 531: This is convenient, we'll have a metric to look at when this rolls out that should show us this step taking a lot less time.

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -607,6 +603,12 @@ def build_dir(self):
     """The build directory. Usually a subdirectory of base_build_dir."""
     raise NotImplementedError
 
+  @property
+  def is_discovery(self):
+    """Return True if this is a discovery run."""
+    return (environment.is_engine_fuzzer_job() and not self.fuzz_target and
+            not self.build_prefix)
```
letitz, line 614: What's up with `build_prefix`? Can you add a comment here explaining the reason behind this whole expression?

letitz (commented):
LGTM % understanding `build_prefix`.

src/clusterfuzz/_internal/build_management/build_manager.py
notvictorl: Added explanation of `is_discovery` to the docstring. Essentially, I'm using `build_prefix` to check if its an extra build. Since Chrome doesn't use extra builds and `fuzz_target` isn't set for extra builds, we want those to proceed to the normal unpack step. Let me know if that makes sense or if it can be improved

letitz: Got it, thanks! The new docstring makes sense.

letitz (approved):
I expect clusterfuzz maintainers will want to see evidence of this working in `dev` before they land this. Can you try to test it in `dev`?

notvictorl:
Have it running on [dev](https://github.com/google/clusterfuzz/commit/a039693007e9e2da4fe5b0974ea2fa5ab252d011) and it seems to be working  fine on variant tasks (uploaded testcases) and `libfuzzer_chrome_asan_schemav1`. `libfuzzer_chrome_msan_schemav1` doesn't seem to be scheduled but `asan` and `msan` should interact similarly for this change.

notvictorl:
@vitaliset friendly nudge to review. thanks!

PauloVLB (approved):
LGTM 

Thank you for working on this improvement!

For the record, this also directly affects (and fixes) b/509600495, as the issue stems from [those](https://pantheon.corp.google.com/storage/browser/chromium-browser-libfuzzer/win32-release_x64-asan;tab=objects?e=-13802955&mods=logs_tg_prod&project=google.com:clusterfuzz&prefix=&forceOnObjectsSortingFiltering=true&pageState=\(%22StorageObjectListTable%22:\(%22s%22:[\(%22i%22:%22objectListDisplayFields%2FtimeCreated%22,%22s%22:%221%22\),\(%22i%22:%22displayName%22,%22s%22:%220%22\)%5D\)\)) huge builds causing Windows jobs to terminate prematurely due to lack of space.

PR #5314, opened 2026-06-08
Require verified email for whitelisted-domain trust

vitaliset:
Carry the real email_verified state on the User and require it before trusting an email's domain, closing the GitHub/EMU email-verification bypass. GitHub login and own-testcase access are unaffected.

b/516412711


PauloVLB (approved):
LGTM

decoNR (approved):
Rubberstamp since paulo's LGTM is not enough since he is not a member of clusterfuzz on github.

PR #5315, opened 2026-06-08
Do not file public GitHub issues for security testcases

vitaliset:
oss-fuzz-robot filed a public issue containing the testcase id and report URL with no security_flag gate, disclosing embargoed testcase identifiers. Skip filing entirely for security testcases.


PauloVLB (approved):
LGTM

decoNR (approved):
Rubberstamp given that paulo's LGTM is not enough since he is not a member of clusterfuzz on github.

PR #5318, opened 2026-06-11
Skip RPATH patching on Schema v1 builds

notvictorl:
Currently, blackbox MSan jobs like `linux_msan_chrome` attempt to patch the RPATHs of binaries to point to instrumented libs provided in the [Dockerfile](https://github.com/google/clusterfuzz/blob/471584471f4b111f861c255be9978adcd308caa5/docker/chromium/base/Dockerfile#L53) since schema v0 archives did not include the instrumented libs in the archives. Schema v1 archives do include the instrumented libs that are needed by the binaries, but ClusterFuzz still tries to patch the RPATHs leading to mismatches causing `Bad build` errors.

ClusterFuzz can safely skip RPATH patching for MSan Schema v1 jobs so that the binaries can use the instrumented libs included in the same archive. Adding `clusterfuzz_manifest.json` to always be extracted, and using the file to read for `archive_schema_version` on new and re-used builds to know if RPATH patching can be skipped. Decoupling `_patch_rpaths()` from `_post_setup_success()` to allow us to handle Release and Debug separately, and helps if schema versions are different.

More info: https://crbug.com/516755275


src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -563,6 +616,25 @@ def _unpack_build(self,
...
+
+        # Sync in-memory schema version with disk if we haven't read it yet.
+        # This helps detect mismatches in shared build dirs (e.g.,
+        # SymbolizedBuild).
+        if self._schema_version is None:
+          self._schema_version = self._read_schema_version_from_disk()
+
+        if self._schema_version is None:
+          # First time unpacking: record the schema version.
+          self._write_schema_version(schema_version)
```
letitz: I don't understand why we write to the directory and re-read from it. Is the intention to handle cases where a build is re-used without being unpacked?

letitz (commented):
One high-level question before a more thorough review.

src/clusterfuzz/_internal/build_management/build_manager.py
notvictorl: Yes, that's correct. The idea is in [`setup()`](https://github.com/google/clusterfuzz/blob/b9081eb617f048cdd7f3a132067dbc054d4fcca9/src/clusterfuzz/_internal/build_management/build_manager.py#L746), we don't always unpack but always do `_post_setup_success` so I am accounting for that case with a file written with the version number.

letitz: Thanks, that makes sense.

In that case, I suggest hooking into the `BuildManager.setup()` branch [1] where we skip unpacking the archive to read the schema version from the manifest file. We can even do something clever for symbolized builds since the code path is different [2], and read both manifest files from both release and debug build dirs.

Then we don't need to lazily get the schema version.we can get it in both cases and thread it to `_post_setup_success()`. In fact, we should likely break the rpath patching logic out of `_post_setup_success()` so that we can call it twice for symbolized builds.

[1]:https://github.com/google/clusterfuzz/blob/3095c816fd2e184b701970eb842e22645be1c265/src/clusterfuzz/_internal/build_management/build_manager.py#L759
[2]: https://github.com/google/clusterfuzz/blob/3095c816fd2e184b701970eb842e22645be1c265/src/clusterfuzz/_internal/build_management/build_manager.py#L876

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -390,22 +404,39 @@ def _patch_rpath(self, binary_path, instrumented_library_paths):
...
+      return
+
     if environment.is_engine_fuzzer_job():
       # Import here as this path is not available in App Engine context.
       from clusterfuzz._internal.bot.fuzzers import utils as fuzzer_utils
 
-      for target_path in fuzzer_utils.get_fuzz_targets(self.build_dir):
+      for target_path in fuzzer_utils.get_fuzz_targets(build_dir):
         self._patch_rpath(target_path, instrumented_library_paths)
     else:
```
letitz: Consider an early return here to reduce nesting.

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -769,6 +795,7 @@ def setup(self):
 
     self._setup_application_path(build_update=build_update)
     self._post_setup_success(update_revision=build_update)
+    self._patch_rpaths()
```
letitz: `setup_application_path` sets up `APP_PATH` by default [1], so I don't think we should try to patch `APP_PATH` and `APP_PATH_DEBUG` in this case. Combined with my comment above, let's do:

```py
_patch_rpaths(self.build_dir, 'APP_PATH')
```

[1]: https://github.com/google/clusterfuzz/blob/936471386478f593079cc2ccec71e308e5105ff6/src/clusterfuzz/_internal/build_management/build_manager.py#L656

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -378,6 +379,19 @@ def _pre_setup(self):
     environment.set_value(self.env_prefix + 'APP_PATH', '')
     environment.set_value(self.env_prefix + 'APP_PATH_DEBUG', '')
 
+  def _read_schema_version_from_manifest(self, build_dir: str) -> int:
```
letitz: AFAICT this method does not use `self`. Move it to a  helper outside the class, or annotate it as `@staticmethod`.

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -390,22 +404,39 @@ def _patch_rpath(self, binary_path, instrumented_library_paths):
 
     set_rpaths(binary_path, rpaths)
 
-  def _patch_rpaths(self, instrumented_library_paths):
+  def _patch_rpaths(self,
+                    build_dir: str | None = None,
+                    app_path_env: str | None = None):
```
letitz: I think it would be a bit easier to follow if `build_dir` was always supplied (and we did not have a fallback for `app_path_env = None`). Let's move this method (and `_patch_rpath()`) outside this class or annotate them as `@staticmethod`, then call it with `self.build_dir`, `self.release_build_dir`, or `self.debug_build_dir`:

```py
def _patch_rpaths(build_dir: str, app_path_env: str):
  ...
  if environment.is_engine_fuzzer_job():
    ...

  assert app_path_env
  ...
```

src/clusterfuzz/_internal/build_management/build_manager.py
```diff
@@ -390,22 +404,39 @@ def _patch_rpath(self, binary_path, instrumented_library_paths):
...
-      app_path_debug = environment.get_value('APP_PATH_DEBUG')
-      if app_path_debug:
-        self._patch_rpath(app_path_debug, instrumented_library_paths)
+      if app_path_env:
+        app_paths = [environment.get_value(app_path_env)]
+      else:
+        app_paths = [
+            environment.get_value('APP_PATH'),
+            environment.get_value('APP_PATH_DEBUG')
+        ]
```
letitz: I don't think we ever want to fix both at once now that we're calling this method twice for symbolized builds. See also my comment above.

src/clusterfuzz/_internal/build_management/build_manager.py
notvictorl: Thanks! (forgot to send comment earlier) this makes it a lot simpler. I extracted `_patch_rpaths()` from `_post_setup_success()` allowing us to separate release and debug builds, and had `clusterfuzz_manifest.json` always be extracted so its available to read on re-used runs.

notvictorl:
@letitz friendly nudge for re-review. thanks in advance!

letitz (commented):
Oh man, I was sitting on a pending review for days. Sorry about that. When I checked it I saw the comments and did not see answers, but I missed that all my comments were marked as pending 🤦 .

I think we're nearly there, this is looking good.

src/clusterfuzz/_internal/build_management/build_manager.py
notvictorl: gotcha, making them explicit and only target release does make it easier to follow for `RegularBuild` and `CustomBuild`.

letitz (approved):
LGTM, thanks!

notvictorl:
@PauloVLB friendly nudge for review. thanks!

PauloVLB (approved):

PR #5323, opened 2026-06-15
[VRP] Ensure untrusted fuzzers do not fuzz in long-lived bots

PauloVLB:
### Problem
Untrusted fuzzers (e.g., external VRP fuzzers) shouldn't run on long-lived bots because they could be malicious or leak sensitive host info. They should only run in isolated, ephemeral environments like Google Cloud Batch.

### Proposed Solution
Filter out fuzzers that aren't marked as `trusted` from being selected by long-lived bots in the task leasing logic. 

Added runtime assertions (`log_fatal_and_exit`) to ensure that if a long-lived bot somehow ends up running an untrusted fuzzer, it will crash immediately before executing any code.

### Chaining Note
This PR is chained on top of `vrp-untrusted-fuzzer-ui` which introduces the `trusted` field. Note that `trusted` defaults to `True` for now to maintain compatibility with existing fuzzers without requiring a backfill, but it will be flipped to `False` by default in a future update.

### Testing
Validated by checking task assignment logic to confirm long-lived bots do not pick up tasks for non-trusted fuzzers.


PauloVLB:
This PR has been pending for some time due to competing priorities and the difficulty of validating changes in dev. Given that we don't have "truly untrusted" fuzzers in production yet and I would like to make progress, I believe it is reasonable to test with a fake untrusted fuzzer in the production environment. That is why I am opening this for review without full validation in dev.

src/clusterfuzz/_internal/base/tasks/__init__.py
```diff
@@ -234,8 +234,18 @@ def get_regular_task(queue=None):
...
     task = get_task_from_message(messages[0], queue)
-    if task:
-      return task
+    if not task:
+      continue
+
+    if task.command == 'fuzz' and not environment.is_tworker():
+      fuzzer = data_types.Fuzzer.query(
+          data_types.Fuzzer.name == task.argument).get()
+      if fuzzer and not fuzzer.trusted:
```
dylanjew: should we assert the fuzzer exists or throw `InvalidFuzzerError` (and handle elsewhere). Or just log and throw out the invalid task here?

src/clusterfuzz/_internal/base/tasks/__init__.py
```diff
@@ -234,8 +234,18 @@ def get_regular_task(queue=None):
...
+        logs.warning(
```
dylanjew: nit: This sounds more like an info log. Unless our queues are already split such that we only expect tasks for trusted fuzzers here and this is line unexpected.

src/clusterfuzz/_internal/fuzzing/fuzzer_selection.py
```diff
@@ -170,6 +170,19 @@ def get_fuzz_task_payload(platform=None):
     jobs = get_job_list(jobs_selection)
     selected_mappings = [entity for entity in mappings if entity.job in jobs]
 
+  if not environment.is_uworker():
```
dylanjew, line 173: optional: I've mentioned this on another PR, but it would be very nice to have helper functions that capture what we mean in a more descriptive way. We have separate `is_tworker` and `is_uworker` and there is a nuance to using `not is_uworker` and `not is_tworker` because it's possible for both `is_uworker` and `is_tworker` to be false for our "Trusted" Windows and Mac bots.  Perhaps we link the enum of `RunTime` with their separate capabilities. 

fwiw your logic here seems correct :)

src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/uworker_io_test.py
```diff
@@ -387,3 +387,40 @@ def test_issue_metadata_field(self):
...
+        'clusterfuzz._internal.metrics.logs.log_fatal_and_exit',
+    ])
+    self.fuzzer = mock.MagicMock(spec=data_types.Fuzzer)
+    self.fuzzer.name = 'test_fuzzer'
+
+  def test_trusted_fuzzer(self):
+    """Test that trusted fuzzer passes without checks."""
+    self.fuzzer.trusted = True
+    uworker_io.check_running_fuzzer_safe(self.fuzzer)
+    self.mock.is_uworker.assert_not_called()
```
dylanjew: it would be better to have the function return `True` rather than `None`, or just assert that a fatal exception/exit happens, rather than relying on mocks here

src/clusterfuzz/_internal/tests/core/fuzzing/fuzzer_selection_test.py
```diff
@@ -325,6 +325,37 @@ def test_platform_restriction(self, local_development):
...
+    trusted_mapping.put()
+
+    data_types.FuzzerJobs(
+        platform='linux', fuzzer_jobs=[untrusted_mapping,
+                                       trusted_mapping]).put()
+
+    argument, job = fuzzer_selection.get_fuzz_task_payload('linux')
+
+    if is_uworker == 'True':
+      self.assertEqual(('untrusted_fuzzer', 'job_1'), (argument, job))
```
dylanjew: nit: it would be cleaner to pass the expected fuzzer into parameterized alongside the `is_uworker` state as a tuple rather than using an if statement and changing the assertion.

src/clusterfuzz/_internal/base/tasks/__init__.py
PauloVLB: I've chosen to log an error and call `task.dont_retry()`. Thanks!

src/clusterfuzz/_internal/base/tasks/__init__.py
PauloVLB: I agree. Changed to `logs.info`.

src/clusterfuzz/_internal/fuzzing/fuzzer_selection.py
PauloVLB: That’s a good point. I agree that it’s a helpful cleanup, I can make a follow-up PR to address it.

src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/uworker_io_test.py
PauloVLB: Done.

src/clusterfuzz/_internal/tests/core/fuzzing/fuzzer_selection_test.py
PauloVLB: Done.

dylanjew (approved):

src/clusterfuzz/_internal/base/tasks/__init__.py
```diff
@@ -234,8 +234,23 @@ def get_regular_task(queue=None):
       return None
 
     task = get_task_from_message(messages[0], queue)
-    if task:
-      return task
+    if not task:
+      continue
+
+    if task.command == 'fuzz' and not environment.is_tworker():
```
ViniciustCosta, line 240: Since other tasks's main stage also run untrusted code, maybe we should avoid any task from untrusted fuzzers to run on trusted bots?

ViniciustCosta, line 240: Also, can a tworker reach this if in run_bot.py we have this logic?
```
      if environment.is_tworker():
        task = tasks.tworker_get_task(
            override_queue=_get_tworker_queue_override())
```

ViniciustCosta (commented):
I had a few questions that we discussed offline and I will drop here:
* Are long-lived bots only running fuzz tasks or is the logic still "get task from queue, if idle, fuzz"? 
* Is this also mapped for mac/android/windows bots?

PauloVLB:
@ViniciustCosta 

> * Are long-lived bots only running fuzz tasks or is the logic still "get task from queue, if idle, fuzz"?

After https://github.com/google/clusterfuzz/pull/5411, only bots under the `ENABLE_FUZZ_FOR_BOTS` feature flag can fuzz when idle (which currently includes only `windows` and `mac`)

> * Is this also mapped for mac/android/windows bots?

Yes. But since untrusted fuzzers cannot be created with non-Linux jobs (https://github.com/google/clusterfuzz/pull/5329), they should not be an issue.

src/clusterfuzz/_internal/base/tasks/__init__.py
PauloVLB: For blackbox fuzzers (the only use case for untrusted fuzzers), only the fuzz task runs the fuzzer binary. 

I will address testcase hardening in a future PR (already in draft: https://github.com/google/clusterfuzz/pull/5368), specifically to prevent running untrusted testcases on long-lived bots.

PauloVLB: In an extreme scenario, yes: if `OVERRIDE_TWORKER_QUEUE` is set to, for example, `jobs-linux`, `tworker_get_task` redirects to `get_regular_task`, allowing a tworker to reach this.

ViniciustCosta: I see, thanks!

ViniciustCosta: Hmm not sure the use case for this flag, but ok, no harm in checking it again

ViniciustCosta:
> @ViniciustCosta
> 
> > * Are long-lived bots only running fuzz tasks or is the logic still "get task from queue, if idle, fuzz"?
> 
> After #5411, only bots under the `ENABLE_FUZZ_FOR_BOTS` feature flag can fuzz when idle (which currently includes only `windows` and `mac`)
> 
> > * Is this also mapped for mac/android/windows bots?
> 
> Yes. But since untrusted fuzzers cannot be created with non-Linux jobs (#5329), they should not be an issue.

Sgtm, thanks Paulo!

ViniciustCosta (approved):

PR #5327, opened 2026-06-16
Migrate Android Build API usage to V4 OnePlatform under feature flag

hunsche:
This PR migrates the Android Build API client calls from legacy V3 (androidbuildinternal.googleapis.com) to the new V4 Private API (androidbuild-pa.googleapis.com) using standard OnePlatform discovery.

### Changes:
- Added a new Datastore Config property 'use_android_build_api_v4' to the Config model class to support runtime toggle of the API version.
- Updated 'fetch_artifact.py' to conditionally initialize and use either the V3 or V4 client/endpoints depending on the Datastore feature flag.
- Refactored logs to use structured dictionary-based entries with the '[AndroidBuildAPI]' tag to facilitate filtering and diagnosis in Cloud Logging.

### DIFF
--- a/src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
+++ b/src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
@@ -48,6 +48,23 @@ DEFAULT_STABLE_CUTTLEFISH_BUILD_INFO = (
     "gs://android-haiku/target-cuttlefish/stable_build_info.json")
 
 
+def _use_v4():
+  """Return True if we should use V4 Android Build API."""
+  try:
+    use_v4 = db_config.get_value('use_android_build_api_v4') or False
+    logs.info(
+        'AndroidBuildAPI feature flag status read.',
+        use_android_build_api_v4=use_v4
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
```diff
@@ -68,24 +83,67 @@ def download_artifact(client, bid, target, attempt_id, name, output_directory,
...
+      'AndroidBuildAPI download_artifact started.',
+      api_version=version_tag,
+      operation='download_artifact',
+      build_id=bid,
+      target=target,
+      attempt_id=attempt_id,
+      artifact_name=name)
+
+  if _use_v4():
+    artifact_query = client.buildartifacts().get(
```
dylanjew, line 98: Just to confirm: Is the difference between the v4 path and the v3 path `buildartifacts` vs `buildartifact`?

dylanjew:
Any concerns with landing this so we can test the API and uncover any issues? @hunsche

PR #5329, opened 2026-06-17
[VRP] Untrusted fuzzer UI

PauloVLB:
### Problem
We need a way to identify fuzzers that are "untrusted" (e.g., external fuzzers provided by VRP researchers) so we can apply different security policies to them (like restricting where they can run). Currently, there is no field in the UI or Datastore to manage this state.

### Proposed Solution
We added a `trusted` boolean field to the `Fuzzer` model in Datastore (defaulting to `False` for safety) and exposed it as a "Trusted" checkbox in the Fuzzer management UI. This ensures all new fuzzers are treated as untrusted by default, maximizing security.

Key changes:
- Added `trusted` boolean property to `Fuzzer` model (default `False`).
- Factored out job validation logic into a centralized helper `check_job_supports_untrusted_workloads` in `data_handler.py` for consistent enforcement between fuzzers and testcases.
- Added a migration script to set `trusted=True` for existing fuzzers to maintain backward compatibility.

### UI Change
image

Trying to run an untrusted fuzzer (i.e. `trusted` is unchecked) on a non-Linux job gives a clear error:

image

### Testing
Validated by deploying to dev and verifying that the checkbox state is correctly persisted to and loaded from Datastore. Added/updated unit tests to cover the `trusted` property and validation helper.

### Validation

I have uploaded a fuzzer with `trusted` set to false; here is a screenshot.

image


src/appengine/handlers/fuzzers.py
```diff
@@ -208,6 +210,16 @@ def apply_fuzzer_changes(self, fuzzer, upload_info):
...
 
+    if fuzzer.untrusted:
+      for job_name in jobs:
+        job = data_types.Job.query(data_types.Job.name == job_name).get()
+        if not job:
+          raise helpers.EarlyExitError(f'Job {job_name} not found.', 400)
+        if job.platform.lower() != 'linux':
+          raise helpers.EarlyExitError(
+              f'Untrusted fuzzers can only be run on Linux jobs. '
+              f'Job "{job_name}" has platform "{job.platform}".', 400)
```
letitz: There is a similar check for untrusted testcases: https://github.com/google/clusterfuzz/blob/5f7e9d7a1dcec2d99bbce8f526d5f3b09699f250/src/appengine/handlers/upload_testcase.py#L416-L423

We should factor the check out into a single method somewhere to ensure there is no accidental divergence in behavior as the code evolves:

```py
def check_job_supports_untrusted_workloads(job_name: str) -> None:
  """Checks that the named job can run untrusted testcases or fuzzers.
  
  Raises:
    ValueError: if the job can run on privileged bots and should never run
      an untrusted workload.
  """
  if not ...:
    raise ValueError(f'Job "{job_name}" does not support running untrusted workloads. $reason")
```

Structuring it with an exception allows this helper to define the error message to show to users, which can evolve over time as the condition changes (maybe one day we won't support untrusted execution on all linux jobs, or maybe we'll support it for windows jobs etc.).

src/appengine/private/components/fuzzers-page/edit-form.html
```diff
@@ -220,11 +220,16 @@
       </div>
       <div class="inline narrow">
         <paper-checkbox checked="{{fuzzer.external_contribution}}">External contribution</paper-checkbox>
+      </div>
+      <div class="inline narrow">
+        <paper-checkbox checked="{{fuzzer.untrusted}}">Untrusted</paper-checkbox>
```
letitz: We should default to safe behavior: if someone forgets to check the box, we should be conservative and assume the fuzzer is not trusted. Let's invert the meaning of the box to "trusted" instead.

src/clusterfuzz/_internal/datastore/data_types.py
```diff
@@ -353,6 +353,9 @@ class Fuzzer(Model):
   # Does it run un-trusted content ? Examples including running live sites.
   untrusted_content = ndb.BooleanProperty(default=False)
```
letitz, line 354: This is interesting! Have you looked into what this does?

src/clusterfuzz/_internal/datastore/data_types.py
```diff
@@ -353,6 +353,9 @@ class Fuzzer(Model):
...
 
+  # Is this fuzzer untrusted? (VRP support)
+  untrusted = ndb.BooleanProperty(default=False)
```
letitz: Same deal here, let's default to safe behavior and assume jobs are untrusted if not specifically told otherwise. Let's invert this boolean to "trusted" (default false) just like for `Testcase`: https://github.com/google/clusterfuzz/blob/5f7e9d7a1dcec2d99bbce8f526d5f3b09699f250/src/clusterfuzz/_internal/datastore/data_types.py#L615

We can run a migration script to set the property to true for all existing fuzzers in the database.

letitz (changes requested):
Thanks for sending this! Overall I think we should invert the trusted/untrusted bit to make trust explicit instead of implicit, and default to safety in the absence of an explicit trust signal.

src/appengine/handlers/fuzzers.py
PauloVLB: Thanks for pointing that out! I have extracted the check into the new method as you suggested. I chose to place it in `clusterfuzz._internal.datastore.data_handler`, as it already contains some Job-related checks.

src/appengine/private/components/fuzzers-page/edit-form.html
PauloVLB: Done

src/clusterfuzz/_internal/datastore/data_types.py
PauloVLB: It appears to relate to the data the fuzzer consumes (e.g., live sites). Inverting it to `trusted` helps make this distinction clearer

src/clusterfuzz/_internal/datastore/data_types.py
PauloVLB: Done. I will also come up with the migration script

PauloVLB:
I have updated the PR description to reflect the changes requested by @letitz and added the `migrate_fuzzer_trusted.py` migration to set all existing fuzzers as trusted.

src/appengine/handlers/fuzzers.py
```diff
@@ -193,6 +194,7 @@ def apply_fuzzer_changes(self, fuzzer, upload_info):
     fuzzer.sample_testcase = None
     fuzzer.console_output = None
     fuzzer.external_contribution = bool(external_contribution)
+    fuzzer.trusted = bool(trusted)
     fuzzer.differential = bool(differential)
```
letitz: [Optional] Consider moving the `bool()` call to the definition of `trusted`, `external_contribution` and `differential`. The types are a bit simpler then - these variables are always booleans.

src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -1211,6 +1212,29 @@ def bot_run_timed_out():
...
+    ValueError: if the job can run on privileged bots and should never run
+      an untrusted workload.
+  """
+  job = data_types.Job.query(data_types.Job.name == job_name).get()
+  if not job:
+    raise ValueError(f'Job "{job_name}" not found.')
+
+  # Chrome is the only ClusterFuzz deployment where there are trusted bots
+  # running utasks. This check also fails on oss-fuzz because of the way it
+  # abuses platform.
```
letitz: This check definitely does not fail on oss-fuzz, since we only raise `ValueError` if `utils.is_chromium()` returns True. Either the comment or the code should be amended. I think the comment was wrong already.

src/appengine/handlers/upload_testcase.py
```diff
@@ -413,14 +412,12 @@ def _handle_upload(self,
     # Chrome is the only ClusterFuzz deployment where there are trusted bots
     # running utasks. This check also fails on oss-fuzz because of the way it
     # abuses platform.
```
letitz: See comment below, I don't think this check ever fails (and ever failed) on oss-fuzz.

src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -1211,6 +1212,29 @@ def bot_run_timed_out():
...
+  if utils.is_chromium() and task_utils.is_remotely_executing_utasks():
+    if ((platform_id and platform_id != 'Linux') or
+        job.platform.lower() != 'linux'):
```
letitz: Consider doing something like:

```py
if (platform_id or job.platform).lower() != 'linux':
```

src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -1211,6 +1212,29 @@ def bot_run_timed_out():
...
+      an untrusted workload.
+  """
+  job = data_types.Job.query(data_types.Job.name == job_name).get()
+  if not job:
+    raise ValueError(f'Job "{job_name}" not found.')
+
+  # Chrome is the only ClusterFuzz deployment where there are trusted bots
+  # running utasks. This check also fails on oss-fuzz because of the way it
+  # abuses platform.
+  if utils.is_chromium() and task_utils.is_remotely_executing_utasks():
```
letitz: Consider inverting the conditional and using an early return to reduce nesting:

```py
# Chrome is the only deployment where there are privileged bots executing utasks.
if not utils.is_chromium():
  return

if not task_utils.is_remotely_executing_utasks():
  return
```

src/clusterfuzz/_internal/datastore/data_types.py
```diff
@@ -353,6 +353,9 @@ class Fuzzer(Model):
   # Does it run un-trusted content ? Examples including running live sites.
   untrusted_content = ndb.BooleanProperty(default=False)
 
+  # Is this fuzzer trusted? (VRP support)
```
letitz:
```suggestion
  # Whether this fuzzer is trusted or not.
  # Untrusted fuzzers can only execute on unprivileged bots, as they might produce
  # malicious outputs. All the testcases they produce are also treated as untrusted.
  # See also `data_handler.check_job_supports_untrusted_workloads()`.
```

src/clusterfuzz/_internal/tests/appengine/handlers/fuzzers_test.py
```diff
@@ -188,3 +193,68 @@ def test_update_fuzzer(self):
...
+        source='original@example.com',
+        timeout=10,
+    )
+    fuzzer.put()
+
+    request_payload = {
+        'csrf_token': form.generate_csrf_token(),
+        'key': fuzzer.key.id(),
+        'name': fuzzer_name,
+        'trusted': False,  # It is untrusted by default (trusted=False)
```
letitz: Let's check that by default, if we omit `trusted`, then we cannot associate the fuzzer with a non-linux job.

src/clusterfuzz/_internal/datastore/data_handler.py
letitz: Why do we check `is_remotely_executing_utasks()`? Can you add a comment explaining that? It's not clear to me.

letitz (commented):
LGTM % responding to comments below and to avoid blocking on me further as I go OOO.

letitz (approved):

src/clusterfuzz/_internal/datastore/data_handler.py
PauloVLB: Fixed, thanks!

src/clusterfuzz/_internal/datastore/data_handler.py
PauloVLB: I have added the explanation to the commentary.

ViniciustCosta (approved):

src/appengine/handlers/fuzzers.py
```diff
@@ -173,8 +173,9 @@ def apply_fuzzer_changes(self, fuzzer, upload_info):
     jobs = request.get('jobs', [])
     timeout = self._get_integer_value('timeout')
     max_testcases = self._get_integer_value('max_testcases')
-    external_contribution = request.get('external_contribution', False)
-    differential = request.get('differential', False)
+    external_contribution = bool(request.get('external_contribution', False))
+    trusted = bool(request.get('trusted', False))
```
javanlacerda, line 177: nit: Having this default value as True would avoid the need to run the backfill script you created. Maybe we could create this property with default True and then move to false?

javanlacerda (approved):

src/appengine/handlers/fuzzers.py
PauloVLB: That makes sense! I will change it to `True` and then submit another PR to set it back to `False`.

PauloVLB: Just a quick comment: this line does not reflect the default value on datastore, rather it is a fallback for HTML checkboxes (which are omitted from the request when unchecked).

I changed the default value in the right file (`data_types.py`).

PR #5330, opened 2026-06-17
[VRP] Mark testcases as untrusted if generated by untrusted fuzzers

PauloVLB:
### Problem
Testcases generated by untrusted fuzzers should also be treated as untrusted. Currently, they inherit default trust settings, which could lead to them being executed in insecure environments or synced incorrectly.

### Proposed Solution
Propagate the trust status from the `Fuzzer` to the generated `Testcase`. 
In `fuzz_task.py`, we now check the `trusted` field of the generating fuzzer. If the fuzzer is untrusted (`trusted=False`), the resulting testcase is also marked as `trusted=False`.

### Chaining Note
Chained on top of `vrp-untrusted-fuzzer-safety` (which uses `trusted` field instead of `untrusted`).

### Testing
Added unit tests in `fuzz_task_test.py` to verify both trusted and untrusted fuzzer scenarios.

### Validation
Due to the difficulty of testing this in the dev environment (which would require a fuzzer to generate a testcase) and the lack of real untrusted fuzzers in production, this will be tested in the production environment.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
@@ -1932,6 +1932,7 @@ class FuzzingSession:
       time.sleep(failure_wait_interval)
       return uworker_msg_pb2.Output(  # pylint: disable=no-member
           error_type=uworker_msg_pb2.ErrorType.FUZZ_NO_FUZZER)  # pylint: disable=no-member
+    uworker_io.check_running_fuzzer_safe(self.fuzzer)
 
     # Update the session's test_timeout to use the fuzzer's timeout (if any).
     # When the fuzzer has a specified timeout, `update_fuzzer_and_data_bundles`
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
@@ -1049,6 +1049,14 @@ def create_testcase(group: uworker_msg_pb2.FuzzTaskCrashGroup,
   comment = (f'Fuzzer {fully_qualified_fuzzer_name} generated testcase crashed '
              f'in {crash.crash_time} seconds '
              f'(r{fuzz_task_output.crash_revision})')
+
+  is_trusted = True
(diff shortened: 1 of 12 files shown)


src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -1194,7 +1201,7 @@ def create_testcase(group: uworker_msg_pb2.FuzzTaskCrashGroup,
       minimized_arguments=crash.arguments,
       # TODO(https://github.com/google/clusterfuzz/issues/4175): Before enabling
```
dylanjew: is this still relevant?

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -1169,6 +1169,13 @@ def create_testcase(group: uworker_msg_pb2.FuzzTaskCrashGroup,
   comment = (f'Fuzzer {fully_qualified_fuzzer_name} generated testcase crashed '
              f'in {crash.crash_time} seconds '
              f'(r{fuzz_task_output.crash_revision})')
+
+  is_trusted = True
+  if uworker_input.setup_input.HasField('fuzzer'):
```
dylanjew: in which case would the `uworker_input` not have a `fuzzer`?  Also, do you need a `HasField` check or can you just check if `setup_input.fuzzer` is `None`?

dylanjew (approved):

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
PauloVLB: At first, I thought it would be nice to have the check just in case, but looking at how other parts of the code handle `setup_input` (like in `utask_main`), they don't check it either because the field is guaranteed to be set during preprocessing. Removed the check.

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
PauloVLB: I don't think so. Removed the comment.

ViniciustCosta (approved):

PR #5332, opened 2026-06-18
Adds support for --no-streaming flag in Android ASan jobs

IvanBM18:
## Overview: 
This PR ensures that Android APKs are installed using the `--no-streaming` flag during ASan fuzzing jobs. Modern Android versions default to streaming installation, which prevents the bundled wrap.sh script (responsible for setting up the ASan library environment) from triggering correctly. This change dynamically detects ASan environments and applies the `--no-streaming` workaround. We still left intact the legacy `asan_device_setup.sh` logic in case any other `haiku` fuzzer job depends on it

For more details for the `wrap.sh` script see the [NDK doc](https://developer.android.com/ndk/guides/wrap-script) and the [ASan docs](https://developer.android.com/ndk/guides/asan)

## Changes:

- `app.py`: Updated the install method signature to accept and forward additional flags trough `**kwargs`.
- `device.py`: Added a private helper to identify ASan jobs or device environments options, and to then pass the `--no-streaming` flag during installation.
- Added new unit tests for this

### DIFF
--- a/src/clusterfuzz/_internal/platforms/android/device.py
+++ b/src/clusterfuzz/_internal/platforms/android/device.py
@@ -352,6 +352,32 @@ def initialize_environment():
   environment.set_value('LOG_TASK_TIMES', True)
 
 
+def _get_no_streaming_flag():
+  """Returns a list containing the '--no-streaming' flag if the job or device
+  requires ASan. Returns an empty list otherwise.
+
+  This is required because streaming installation prevents wrap.sh from
+  functioning correctly on modern Android versions.
+  """
+  additional_flags = []
+
+  is_asan_job = False
+  job_name = environment.get_value('JOB_NAME')
+  if job_name:
+    memory_tool = environment.get_memory_tool_name(job_name)
+    if memory_tool == 'ASAN':
+      is_asan_job = True
+
+  is_asan_device = (
+      environment.get_value('ASAN_DEVICE_SETUP') or
+      settings.get_sanitizer_tool_name() == 'asan')
+
+  if is_asan_job or is_asan_device:
+    additional_flags.append('--no-streaming')
+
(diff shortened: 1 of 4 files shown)


src/clusterfuzz/_internal/platforms/android/device.py
```diff
@@ -352,6 +352,32 @@ def initialize_environment():
   environment.set_value('LOG_TASK_TIMES', True)
 
 
+def _get_no_streaming_flag():
```
Xeicker: I think it would look cleaner if this was a boolean that just returns if ASAN setup is required. And in the `install_application_if_needed` function it adds the flag depending on that boolean

Xeicker (approved):
Other than the nit LGTM

src/clusterfuzz/_internal/platforms/android/device.py
IvanBM18: You are correct, as it is right now its cleaner  if this returned a boolean and then another function could add the flag depending on that.

But if we ever need to add any other additional flag, then we would likely need to refactor the code to something like what we have right now (a method that sets up flags and calls them down into `install_app_if_needed()`), i will make this a little bit cleaner by returning a str instead of a list[str]. And ill keep this comment in case another reviewer also likes your idea better. Thanks!

javanlacerda (approved):

src/clusterfuzz/_internal/platforms/android/device.py
IvanBM18: Discussed offline:
Per changes requested from @fernandofloresg result in the implement of the changes suggested by @Xeicker 

Thanks :D

fernandofloresg (approved):

PR #5338, opened 2026-06-24
Make chromium testcases public when bug is made public

jonathanmetzman:
This is similar to how OSS-Fuzz behaves with a different implementation due to differences in how they use the issue tracker.
Tested and confirmed working on staging appengine.


ViniciustCosta (approved):
Lgtm

This was tested in our staging appenging and verified as working as intended.

PR #5294, opened 2026-05-26
Fix legacy Datastore access in data_handler.py for UWORKERs

jardondiego:
### Background

As part of the migration of ClusterFuzz to a scalable, isolated LUCI Swarming architecture, untrusted execution logic is handled by the UWORKER. The UWORKER operates within a highly sandboxed Docker container using a restricted IAM service account (untrusted-worker@...) that explicitly lacks Google Cloud Datastore read/write permissions.

During the teardown and log-upload phases of native Android fuzzing tasks (and potentially others), the UWORKER frequently crashed with google.api_core.exceptions.PermissionDenied (403) exceptions. Investigation revealed that several legacy monolithic utility functions in data_handler.py were still attempting raw Datastore queries (e.g.,
data_types.Job.query(...).get()) to fetch metadata about the fuzzing environment.

### Proposal

This PR comprehensively patches data_handler.py to enforce the UWORKER security boundary. We implement short-circuit checks (if environment.is_uworker():) at the top of the following functions:
  1. get_value_from_job_definition
  2. get_component_name
  3. get_repository_for_component
  4. get_value_from_fuzzer_environment_string

Instead of hitting the database, these functions now pull the requested metadata directly from the UWORKER's local environment variables (which are securely pre-injected by the Trusted Worker (TWORKER) before the container boots).

### Testing
  * Unit Tests: Added tests to explicitly mock is_uworker and assert that Datastore queries (ndb.query) are never invoked when executing in an untrusted context
  * End-to-End Validation: Fully verified on a live chrome-development instance. Successfully executed an end-to-end Swarming task (android_content_shell_native_job), confirming the UWORKER correctly detects a timeout, packages its logs, bypasses the Datastore traps, and exits cleanly with Code: 0.

PR #5339, opened 2026-06-24
Revert "Fix legacy Datastore access in data_handler.py for UWORKERs"

hunsche:
Reverts google/clusterfuzz#5294


hunsche:
This is generating a massive spike in 404s when fetching the revision, leading to instability. We are reverting it to resolve the issue.

cc @jardondiego

decoNR (approved):

PR #5342, opened 2026-06-30
Fix issue with double-opening temp corpus files

aakallam:
It is fine for Linux and Mac bots to create multiple file objects to the same on-disk file but Windows does not allow this and throws a PermissionError for any attempts to open a file that is already opened.

This slightly refactors `src/clusterfuzz/_internal/google_cloud_utils/storage.py` (and its callers) to allow passing a filepath or file object when downloading from signed URLs and changes `src/clusterfuzz/_internal/fuzzing/corpus_manager.py` to use the file-based method.

Bug: crbug.com/529346593

### DIFF
--- a/src/clusterfuzz/_internal/fuzzing/corpus_manager.py
+++ b/src/clusterfuzz/_internal/fuzzing/corpus_manager.py
@@ -480,17 +480,21 @@ class ProtoFuzzTargetCorpus(FuzzTargetCorpus):
     shell.create_directory(directory, create_intermediates=True)
     if corpus.backup_url:
       tmpdir = environment.get_value('FUZZ_INPUTS')
-      with tempfile.NamedTemporaryFile(
-          dir=tmpdir, suffix='.zip') as temp_zipfile:
-        try:
-          storage.download_signed_url_to_file(corpus.backup_url,
-                                              temp_zipfile.name)
-          with archive.open(temp_zipfile.name) as reader:
-            reader.extract_all(directory)
-            for member in reader.list_members():
-              self._filenames_to_delete_urls_mapping[member.name] = None
-        except RuntimeError:
-          logs.warning('Couldn\'t download corpus backup')
+      # To avoid opening the file multiple times on Windows, manually close the
+      # file after write and delete it after extraction.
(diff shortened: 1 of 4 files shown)


dylanjew (approved):

src/clusterfuzz/_internal/fuzzing/corpus_manager.py
```diff
@@ -483,8 +483,7 @@ def _sync_corpus_to_disk(self, corpus, directory):
       with tempfile.NamedTemporaryFile(
           dir=tmpdir, suffix='.zip') as temp_zipfile:
         try:
-          storage.download_signed_url_to_file(corpus.backup_url,
-                                              temp_zipfile.name)
+          storage.download_signed_url_to_file(corpus.backup_url, temp_zipfile)
           with archive.open(temp_zipfile.name) as reader:
```
PauloVLB: Not sure, but since `temp_zipfile` is still held open, wouldn't this line try to reopen it by name and potentially face the same error?

aakallam: No because I have changed `storage.download_signed_url_to_file()` to take a file object and simply write to it.

PauloVLB: I see, but I'm wondering about the very next line: `with archive.open(temp_zipfile.name)`. Won't this try to open the file by name again?

aakallam: Ah sorry, I read through your comment in my email, missed which line you were talking about and completely misunderstood!

You are right that trying to read the zip would throw a similar error so I have refactored that bit as well. There wasn't a super clean way to fix that but I'm hoping my solution is the least error-prone way to handle it.

dylanjew (approved):

src/clusterfuzz/_internal/fuzzing/corpus_manager.py
dylanjew: Good catch!

PauloVLB (approved):
LGTM

This will be deployed in today’s release, so let’s keep track of the results.

PR #5344, opened 2026-07-01
Hide Chromium trusted job warning on non-Chromium instances

ViniciustCosta:
This PR resolves a bug where the "untrusted workloads" warning banner was showing up unconditionally on non-Chromium instances of ClusterFuzz (such as OSS-Fuzz), despite the fact that the actual "trusted agreement" input field is correctly hidden.

### Changes
* **`src/appengine/private/components/upload-testcase/upload-form.html`**:  Wrapped the warning block in a `<template is="dom-if" if="[[fieldValues.isChromium]]">` block so that it is only rendered on Chromium instances. 
* **`src/appengine/private/components/upload-testcase/upload-form-simplified.html`**: Similarly wrapped the warning block in a `<template is="dom-if" if="[[fieldValues.isChromium]]">` block for consistency.

Related: b/390151007

### DIFF
--- a/src/appengine/private/components/upload-testcase/upload-form-simplified.html
+++ b/src/appengine/private/components/upload-testcase/upload-form-simplified.html
@@ -145,13 +145,15 @@
           </div>
         </template>
       </div>
-      <div class="warning">
-      <span><strong>WARNING</strong>: ClusterFuzz only partially supports untrusted workloads.<br></span>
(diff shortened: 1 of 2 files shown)


decoNR (approved):
As discussed offline, it was verified that this works in a Chrome environment (since our development environment reflects Chrome).

PR #5346, opened 2026-07-01
Ensure files are always closed after use

aakallam:
This is a follow up to #5342 that refactors a bunch of file I/O to use `with` statements to ensure that the file handles are always closed. Closing file handles is especially important on Windows because only one file object can point at a file at one time.

To make a few complex file opening cases work, this PR adds a couple new path variables to simplify the refactoring but otherwise does not introduce or change any logic.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/setup.py
+++ b/src/clusterfuzz/_internal/bot/tasks/setup.py
@@ -846,11 +846,14 @@ def archive_testcase_and_dependencies_in_gcs(resource_list, testcase_path: str,
 
   logs.info(f'Testcase and related files :\n{resource_list}')
 
+  upload_file_path = testcase_path
   if len(resource_list) <= 1:
     # If this does not have any resources, just save the testcase.
     # TODO(flowerhack): Update this when we teach CF how to download testcases.
     try:
-      file_handle = open(testcase_path, 'rb')
+      with open(testcase_path, 'rb') as _:
+        # Verify file can be opened.
+        pass
     except OSError:
       logs.error(f'Unable to open testcase {testcase_path}.')
       return None, None, None
@@ -884,15 +887,17 @@ def archive_testcase_and_dependencies_in_gcs(resource_list, testcase_path: str,
 
     # Create the archive.
     zip_path = os.path.join(environment.get_value('INPUT_DIR'), zip_filename)
-    zip_file = zipfile.ZipFile(zip_path, 'w')
-    for file_name in resource_list:
-      if os.path.exists(file_name):
-        relative_filename = file_name[base_len:]
-        zip_file.write(file_name, relative_filename, zipfile.ZIP_DEFLATED)
-    zip_file.close()
+    with zipfile.ZipFile(zip_path, 'w') as zip_file:
+      for file_name in resource_list:
+        if os.path.exists(file_name):
+          relative_filename = file_name[base_len:]
+          zip_file.write(file_name, relative_filename, zipfile.ZIP_DEFLATED)
 
     try:
-      file_handle = open(zip_path, 'rb')
+      with open(zip_path, 'rb') as _:
+        # Verify file can be opened.
+        pass
+      upload_file_path = zip_path
     except OSError:
       logs.error(f'Unable to open testcase archive {zip_path}.')
       return None, None, None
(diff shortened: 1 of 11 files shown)


src/clusterfuzz/_internal/bot/tasks/setup.py
```diff
@@ -879,31 +882,37 @@ def archive_testcase_and_dependencies_in_gcs(resource_list, testcase_path: str,
...
+        logs.error('Failed to upload testcase.')
+        return None, None, None
 
-  # Don't need the archive after writing testcase to blobstore.
-  if zip_path:
-    shell.remove_file(zip_path)
+  finally:
+    # Don't need the archive after writing testcase to blobstore.
+    if zip_path:
+      shell.remove_file(zip_path)
```
dylanjew, line 917: Is there a similar way we can handle file clean up?

dylanjew (approved):
thanks!

src/clusterfuzz/_internal/bot/tasks/setup.py
aakallam: Do you mean for the testcase files or something else?

dylanjew: I thought it was strange that we're manually calling `remove_file(zip_path)` in a couple places but I'm not sure of a way around that.

either way cleaning up the file usage LGTM

dylanjew (approved):

ViniciustCosta (commented):
Looks good!

Since it touches multiple files, could you get some evidence running on dev that the change is not adding unexpected regressions?

aakallam:
I deployed to dev yesterday and have confirmed that no regressions have been found

src/clusterfuzz/_internal/bot/tasks/setup.py
```diff
@@ -879,31 +882,37 @@ def archive_testcase_and_dependencies_in_gcs(resource_list, testcase_path: str,
...
-      file_handle = open(zip_path, 'rb')
+      with open(zip_path, 'rb') as _:
+        # Verify file can be opened.
+        pass
+      upload_file_path = zip_path
     except OSError:
       logs.error(f'Unable to open testcase archive {zip_path}.')
       return None, None, None
+    finally:
+      if zip_path:
```
ViniciustCosta: Wouldn't this remove the `zip_path` file before it is used in `with open(upload_file_path)`, since `upload_file_path` is just a reference to `zip_path`?

src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
```diff
@@ -1028,67 +1028,64 @@ def store_minimized_testcase(
...
-          file_handle = open(zip_path, 'rb')
+          testcase_filepath = zip_path
         else:
           if not filtered_file_list:
             # We minimized everything. The only thing needed to reproduce is the
             # interaction gesture.
-            file_path = file_list[0]
-            file_handle = open(file_path, 'wb')
-            file_handle.close()
+            testcase_filepath = file_list[0]
```
ViniciustCosta: In the original code, if we minimized everything, opening the file in `wb` mode and immediately closing it effectively truncated the file to 0 bytes. I think this step was removed here, which mean instead of uploading an empty testcase, it will now upload the original, un-minimized file contents, right?

src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
```diff
@@ -1028,67 +1028,64 @@ def store_minimized_testcase(
...
+  if not testcase_filepath:
+    logs.error('Unable to find file to read')
+    return
+
   # Store the testcase.
-  data = file_handle.read()
-  storage.upload_signed_url(data, minimize_task_input.testcase_upload_url)
-  minimized_keys = minimize_task_input.testcase_blob_name
-  file_handle.close()
+  with open(testcase_filepath, 'rb') as file_handle:
```
ViniciustCosta: Should we wrap this in a try-catch similar to what we had in the original code with the `logs.error('Unable to open archive for blobstore write.')`?

src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
aakallam: Oh, you are totally right. I missed it was `wb` mode and misinterpreted this as another place where the file was only opened to verify it was present.

src/clusterfuzz/_internal/bot/tasks/utasks/minimize_task.py
aakallam: I meant to add a check that the path exists in the new `if not testcase_filepath:` block above but have added this try-catch here to cover that and any other OSError.

src/clusterfuzz/_internal/bot/tasks/setup.py
aakallam: Good catch! Not sure why I ended up with that `finally` block duplicated there.

ViniciustCosta (approved):

PR #5347, opened 2026-07-02
[Swarming] Decouples linux from android_emulator fuzzing

IvanBM18:
## Overview
This small PR focuses on re enabling linux fuzzing in swarming by decoupling the linux jobs from automatically being treated as android jobs.

Previously we always booted up an al emulator on every single job that reached swarming(as per the config's specs) and we always used the `OS_OVERRIDE:ANDROID` env var, so that clusterfuzz followed the setup/logic decisions of a android device. This had 2 issues:
- Our android jobs have had to set `PLATFORM=LINUX` so that CF could send a swarming request targeting a Linux bot(which contained our emulated devices)
- Thanks to that, our actual Linux Jobs were always overriden to android

## Changes
- When the `job.platform` is android_emulator we override is target swarming bot OS to linux in the swarming request, which is the one that contains the emulator setup logic.
  - When we construct the task request we will still gather the new config specs(see other pr) from the ANDROID_EMULATOR platform
- Added the ANDROID_EMULATOR to the list of jobs that we want to schedule in swarming queues.
- Added the is_android_emulator() method to the battery check.

#### Note
This Pr is dependant of this other config pr
- https://clusterfuzz-config-472119376969.us-central1.sourcemanager.dev/clusterfuzz-testing/clusterfuzz-config/pulls/352/files
Which adds the config specs for android emulator jobs 

## Tests performed 
Since this type of change is more focused on swarming.yaml config changes i tested this by creating a new job using the `ANDROID_EMULATOR` platform and i verified if it was correctly scheduled in a linux swarming bot, and that said bot launched the emulator and also that the fuzzing task executed successfully
See [the logs](https://cloudlogging.app.goo.gl/goynt3MZBS49VTpEA)

### DIFF
--- a/configs/test/swarming/swarming.yaml
+++ b/configs/test/swarming/swarming.yaml
@@ -30,6 +30,26 @@ mapping:
     expiration_secs: 86400
     execution_timeout_secs: 86400
     docker_image: 'gcr.io/clusterfuzz-images/base:a2f4dd6-202202070654'
+  ANDROID_EMULATOR:
+    priority: 1
+    command: 
+      - 'luci-auth'
+      - 'context'
+      - '--'
+      - './linux_entry_point.sh'
+    cas_input_root:
+      cas_instance: 'projects/server-name/instances/instance_name'
+      digest:
+        hash: 'linux_entry_point_archive_hash'
+        size_bytes: 1234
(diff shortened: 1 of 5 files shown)


Xeicker (approved):
LGTM

javanlacerda (approved):

PR #5351, opened 2026-07-06
Use clusterfuzz_manifest.json to set ChromeBuildArchive

notvictorl:
Chrome build archives using archive schema v1 don't have `args.gn` at the root of the zip, so they are run as `DefaultBuildArchive` instead of `ChromeBuildArchive`. This leads to them not utilizing `clusterfuzz_manifest.json` later in the run, and in turn, waste time running `run_*_fuzzer` binaries. 

Adding `clusterfuzz_manifest.json` to the heuristic used to categorize `ChromeBuildArchive` in `open_with_reader()` since only Chrome archives create and include this file. Adding unit tests to help test this section.

Bug: https://crbug.com/507833074

### DIFF
--- a/src/clusterfuzz/_internal/build_management/build_archive.py
+++ b/src/clusterfuzz/_internal/build_management/build_archive.py
@@ -54,6 +54,10 @@ FUZZ_TARGET_ALLOWLISTED_PREFIXES = [
     'src_root',
 ]
 
+# Manifest file present in Chrome build archives. Specifies the archive
+# schema version along with optional build metadata (e.g., fuzz targets).
+MANIFEST_FILENAME = 'clusterfuzz_manifest.json'
+
 
 class BuildArchive(archive.ArchiveReader):
   """Abstract class for representing a build archive. This is mostly an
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -469,12 +473,13 @@ def open_with_reader(reader: archive.ArchiveReader) -> BuildArchive:
   # archive implementation to use.
   # Hopefully, we can search in the archive whether some files are present and
   # give us some hints.
-  # For instance, chrome build archives are embedding `gn.args`. Let's use
-  # this for now.
+  # For instance, chrome build archives are embedding `gn.args` and
```
letitz: While we're at it let's fix the grammar and file name here:


```suggestion
  # For instance, chrome build archives contain `args.gn` and/or
```

letitz (approved):
LGTM, thanks!

notvictorl:
@javanlacerda friendly nudge for review, thanks!

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -54,6 +54,10 @@
     'src_root',
 ]
 
+# Manifest file present in Chrome build archives. Specifies the archive
+# schema version along with optional build metadata (e.g., fuzz targets).
+MANIFEST_FILENAME = 'clusterfuzz_manifest.json'
```
ViniciustCosta: nit: since this is chrome-specific, could you change the var name to `CHROME_MANIFEST_FILE` ?

ViniciustCosta (approved):

PR #5352, opened 2026-07-07
[Corpus] Use threads for datastore operations under targets feature flag

PauloVLB:
b/530149063

### Possible Problem 
`corpus_pruning` tasks are failing silently with OOM because of how we do GCS operations (uploads, deletions). We are using a process pool with 16 workers, which calls `fork()`. If the parent process memory is large, doing this multiplies the memory and kills the bot instantly.

### Proposed Solution
Switch from processes to threads for these I/O bound GCS operations in `storage.py`. Threads share memory, avoiding the `fork()` memory multiplication. 
To roll this out safely, it is guarded by the `STORAGE_THREADED_OPS_FUZZ_TARGETS` feature flag.

### Feature Flag Setup
image

### Testing
We are going to test this with the `iperf` project, which is a known project where we can test on. Specifically, we will add `libFuzzer_iperf_cjson_fuzz` to the feature flag allowlist.

### Validation
I validated this in the `dev` environment and verified that the code path is triggered correctly and works as expected.

image

### DIFF
--- a/src/clusterfuzz/_internal/base/concurrency.py
+++ b/src/clusterfuzz/_internal/base/concurrency.py
@@ -22,13 +22,13 @@ POOL_SIZE = multiprocessing.cpu_count()
 
 
 @contextlib.contextmanager
-def make_pool(pool_size=POOL_SIZE, max_pool_size=None):
+def make_pool(pool_size=POOL_SIZE, max_pool_size=None, use_threads=False):
   """Returns a pool that can (usually) execute tasks concurrently."""
   if max_pool_size is not None:
     pool_size = min(pool_size, max_pool_size)
 
   # Don't use processes on Windows and unittests to avoid hangs.
-  if (environment.get_value('PY_UNITTESTS') or
+  if (use_threads or environment.get_value('PY_UNITTESTS') or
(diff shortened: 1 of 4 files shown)


javanlacerda (approved):
LGTM! Thanks for catching and fixing this. Nice job!

src/clusterfuzz/_internal/base/concurrency.py
```diff
@@ -22,13 +22,13 @@
 
 
 @contextlib.contextmanager
-def make_pool(pool_size=POOL_SIZE, max_pool_size=None):
+def make_pool(pool_size=POOL_SIZE, max_pool_size=None, use_threads=False):
   """Returns a pool that can (usually) execute tasks concurrently."""
   if max_pool_size is not None:
     pool_size = min(pool_size, max_pool_size)
 
   # Don't use processes on Windows and unittests to avoid hangs.
```
ViniciustCosta, line 30: Maybe update this comment in the future since there is this "new" case to use threads instead of processes.

PauloVLB: Done in https://github.com/google/clusterfuzz/pull/5421.

PR #5353, opened 2026-07-07
Correct execution permission check `100` -> `0o100`

notvictorl:
The bitwise check for `needs_execute_permission` uses decimal 100 (`0o144`) instead of octal 100 to determine if execution permissions would be needed. This results in giving execution permissions to files that had read permissions in group or others (`---r--r--`).

Corrected this bug by switching to `0o100` and added unit tests to better check which files receive execution permissions. 

Bug: https://crbug.com/531741750

### DIFF
--- a/src/clusterfuzz/_internal/tests/core/system/archive_test.py
+++ b/src/clusterfuzz/_internal/tests/core/system/archive_test.py
@@ -44,14 +44,13 @@ class UnpackTest(unittest.TestCase):
     """Test unpack with trusted=False passes with file having './' prefix."""
     tgz_path = os.path.join(TESTDATA_PATH, 'cwd-prefix.tgz')
     output_directory = tempfile.mkdtemp(prefix='cwd-prefix')
+    self.addCleanup(shell.remove_directory, output_directory)
     with archive.open(tgz_path) as reader:
       reader.extract_all(output_directory, trusted=False)
 
     test_file_path = os.path.join(output_directory, 'test')
     self.assertTrue(os.path.exists(test_file_path))
     self.assertEqual(open(test_file_path).read(), 'abc\n')
-
-    shell.remove_directory(output_directory)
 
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/tests/core/system/archive_test.py
```diff
@@ -76,6 +88,71 @@ def test_unpack_absolute_path_traversal(self):
...
+        reader.extract_all(output_directory, trusted=True)
+
+      expected_calls = [
+          mock.call(os.path.join(output_directory, 'exe.sh'), 0o750),
+          mock.call(os.path.join(output_directory, 'reg_600.txt'), 0o640),
+      ]
+      self.mock.chmod.assert_has_calls(expected_calls, any_order=True)
+      self.assertEqual(self.mock.chmod.call_count, 2)
+
+      shell.remove_directory(output_directory)
```
dylanjew: nit: is this remove_directory required or is there a way to clean it up automatically ?

dylanjew (approved):

src/clusterfuzz/_internal/tests/core/system/archive_test.py
notvictorl: good catch! we can use [`addCleanup`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.addCleanup) to more safely remove the temp directory even if the test encounters an exception. went ahead and changed it

dylanjew: Still optional, but I was thinking more along the lines of `with tempfile.NamedTemporaryFile() as output_dir:` and then having that clean up the file when out of scope.

dylanjew (approved):

ViniciustCosta (approved):
Seems like a nice catch to me!

PR #5354, opened 2026-07-07
symbolizer: add logging for symbolizer errors

g-ortuno:
Spawns a separate thread to observer stderr and include its output when logging symbolization errors. We use a separate thread to avoid potential deadlock situations if llvm-symbolizer writes more than 64KB.

This will help diagnose symbolizer issues like https://crbug.com/532093354 more easily in the future.

### DIFF
--- a/src/clusterfuzz/_internal/crash_analysis/stack_parsing/stack_symbolizer.py
+++ b/src/clusterfuzz/_internal/crash_analysis/stack_parsing/stack_symbolizer.py
@@ -28,10 +28,12 @@
 # Disable all pylint warnings/errors as this is based on external code.
 # pylint: disable-all
 
+import collections
 import os
 import re
 import subprocess
 import sys
+import threading
 
 from clusterfuzz._internal.base import utils
 from clusterfuzz._internal.google_cloud_utils import storage
@@ -240,6 +242,7 @@ class LLVMSymbolizer(Symbolizer):
     self.default_arch = default_arch
     self.system = system
     self.dsym_hints = dsym_hints
+    self.stderr_buffer = collections.deque(maxlen=100)
     self.pipe = self.open_llvm_symbolizer()
 
   def open_llvm_symbolizer(self):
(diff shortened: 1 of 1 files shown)


dylanjew (approved):

g-ortuno:
@dylanjew mind taking a look? Progression tasks haven't run since 8am for some reason[[1]](http://shortn/_DolwgEGkXf). Unfortunately no progression or regression tasks were spawned while the changes were in dev[[2]](http://shortn/_HYWqImAvyw). A few fuzzing tasks were spawned but I couldn't find one that found a crash. There are a lot of regression tasks that successfully symbolized stack traces during that window, but I'm not sure they were running the new code.

The symbolization errors are only happening on macOS, so we won't see any of the new logs on dev.

dylanjew (approved):

g-ortuno:
@javanlacerda PTAL! Would be good to include this in the next release so we can understand the macOS failures: https://crbug.com/532093354. Since this is in the symbolizer code, it shouldn't affect OSS, right?

dylanjew:
> @dylanjew mind taking a look? Progression tasks haven't run since 8am for some reason[[1]](http://shortn/_DolwgEGkXf). Unfortunately no progression or regression tasks were spawned while the changes were in dev[[2]](http://shortn/_HYWqImAvyw). A few fuzzing tasks were spawned but I couldn't find one that found a crash. There are a lot of regression tasks that successfully symbolized stack traces during that window, but I'm not sure they were running the new code.
> 
> The symbolization errors are only happening on macOS, so we won't see any of the new logs on dev.

The schedule progression cron runs at 7 AM every day. Maybe there's something with the dev/prod sync mechanism or with releases that clears the progression tasks queue.

g-ortuno:
With @dylanjew 's help, I confirmed that symbolization is still working http://shortn/_v1dlN1tmsO on dev so this should be safe to land.

g-ortuno:
Abandoning in favor of #5356

PR #5357, opened 2026-07-09
[Corpus] Fix stale mapping and respect delete in rsync_from_disk

PauloVLB:
b/530149063

### Problem
`ProtoFuzzTargetCorpus.rsync_from_disk` keeps a stale internal mapping of files to delete. Since the `corpus_pruning` task calls this method twice (once after initial pruning and again after cross-pollination), the second pass attempts to delete the exact same files again, causing a flood of 404 warnings. It was also completely ignoring the `delete` argument.

[Link for logs](https://cloudlogging.app.goo.gl/77M36EVPLVwFMCaE7)

image

### Proposed Solution
Update the internal mapping after deletion to remove the files that were just deleted. Also, wrap the deletion logic to respect the `delete` flag.

### Validation
Added a new unit test `test_rsync_multiple_calls` in `corpus_manager_test.py` to confirm that multiple calls don't try to delete the same files.

Also validated in dev environment: those are logs from a `corpus_pruning` task successfully executing without the mentioned warnings.

[Link](cloudlogging.app.goo.gl/xZxPAfDvuom4uFKk6)

image

### Why weren't these warnings visible before?

The bug was always there but silent. Before switching to threads, we used processes. On Linux, the background thread responsible for flushing logs to Cloud Logging dies in forked child processes. Thus, warnings from child processes (like these 404s) were swallowed. Threads keep execution in the same process, making the warnings visible.

### DIFF
--- a/src/clusterfuzz/_internal/fuzzing/corpus_manager.py
+++ b/src/clusterfuzz/_internal/fuzzing/corpus_manager.py
@@ -450,9 +450,14 @@ class ProtoFuzzTargetCorpus(FuzzTargetCorpus):
         self._filenames_to_delete_urls_mapping)) or
             len(filenames_to_delete) < 1_000)
 
-    logs.info('Deleting files.')
-    storage.delete_signed_urls(filenames_to_delete)
-    logs.info('Done files.')
+    if delete:
+      logs.info('Deleting files.')
+      storage.delete_signed_urls(filenames_to_delete)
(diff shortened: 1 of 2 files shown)


ViniciustCosta (approved):
Seems like a nice catch to me! (though I'm not an expert on the corpus pruning task behavior, so I'm not sure how would we verify that the recent changes are working as intended).

PR #5358, opened 2026-07-10
Gracefully shut down Centipede fuzzers

aakallam:
Currently, Centipede fuzzing sessions are sent a `SIGKILL` after the allotted time has passed, unlike libFuzzer sessions which send a `SIGTERM` and then a `SIGKILL` after 10s if the fuzzing process has not exited. Sending a `SIGKILL` to Centipede fuzzers does not let them update the corpus with the latest batch's data and could have other unintended side effects.

Centipede supports fuzzing durations, so this change passes the duration to the `centipede` binary and also instructs the bot to send `SIGTERM` (which causes Centipede to immediately begin wrapping up the session) before trying to kill the process later if it hasn't already stopped. The only exception to this is Windows where it has to just [jump](https://github.com/google/clusterfuzz/blob/04d37b351397bcd7a62c2e34a0b91c0ec4e5961e/src/clusterfuzz/_internal/system/new_process.py#L87) to process termination.

Bug: 533039107

### DIFF
--- a/src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
@@ -320,9 +320,21 @@ class Engine(engine.Engine):
 
     old_corpus_len = shell.get_directory_file_count(options.corpus_dir)
     logs.info(f'Corpus length before fuzzing: {old_corpus_len}')
-    logs.info(f'Launching a fuzzer run with arguments: {options.arguments}')
+
+    # Centipede is the only fuzzing engine to support fuzzing durations so set
+    # that here using the max_time.
+    # Note: Sending a SIGTERM is enough for Centipede to cleanly shut down but
+    # Windows bots can only do the equivalent of SIGKILL and will benefit from
+    # the fuzzing duration flag.
+    args = list(
+        options.arguments) + [f'--{constants.STOP_AFTER_FLAGNAME}={max_time}s']
+
+    logs.info(f'Launching a fuzzer run with arguments: {args}')
     fuzz_result = runner.run_and_wait(
-        additional_args=options.arguments, timeout=timeout)
+        additional_args=args,
+        timeout=timeout,
+        terminate_before_kill=True,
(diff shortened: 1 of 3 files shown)


src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
```diff
@@ -320,9 +320,14 @@ def fuzz(self, target_path, options, reproducers_dir, max_time):  # pylint: disa
 
     old_corpus_len = shell.get_directory_file_count(options.corpus_dir)
     logs.info(f'Corpus length before fuzzing: {old_corpus_len}')
-    logs.info(f'Launching a fuzzer run with arguments: {options.arguments}')
+    args = list(
```
dylanjew, line 329: should this be added in [prepare()](https://github.com/google/clusterfuzz/blob/76882597f27206e6ddfa89e6b5bf4828545b85be/src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py#L210) alongside the rest of the arguments as part of FuzzOptions?

aakallam: `prepare()` is part of an interface used by all engines so I didn't want to touch that just to get this added for Centipede. It's also a bit tricky to pass the timeout to `prepare()` because the timeout can depend on the [built fuzzer options](https://github.com/google/clusterfuzz/blob/76882597f27206e6ddfa89e6b5bf4828545b85be/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py#L1377) (for the other engines).

I will add some inline comments to make it clear why it can/should be handled there.

dylanjew (approved):

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
dylanjew: got it, thanks for looking into it!

dylanjew (approved):

ViniciustCosta (approved):
lgtm

aakallam:
Thanks for the reviews!

PR #5363, opened 2026-07-15
Upload Fuzzer Logs to GCS only if not in a uworker

IvanBM18:
Bug: b/473624472 

## Overview
After reverting [this PR](https://github.com/google/clusterfuzz/pull/5339/changes) `PERMISSION_DENIED` errors started to reappear in CF. Its clear that we can't follow the same approach of doing a short-circuit in the DB calls.
image

This instance of `PERMISSION_DENIED` errors occurs on `uworkers` that perform the `fuzz` task, in this workflow theres a logic to upload the Fuzzer logs to GCS at the `utask_main` stage. As we know `uworkers` cant access DB nor GCS, so when a `uworker` reaches this code it fails. Now instead of always uploading the Fuzzer logs to GCS, we only upload them if not in a `uworker`, this allows regular bots that perform the fuzz task(pre->main->pos) locally to continue to work. 

## Tests performed
Changes have been in `dev` since Jul-13. Since then error logs related to this instance of `PERMISSION_DENIED` issues have been resolved.
image
Note how this error dramatically decreases. Other instances of this error like this one:
image
Are unrelated to this changes and seem intermittent errors due to external reasons.

## Note:
This PR is the first of 3(maybe 4) PRs to tackle the PERMISSION_DENIED issues

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
@@ -1839,7 +1839,7 @@ class FuzzingSession:
         thread = process_handler.get_process()(
             target=testcase_manager.run_testcase_and_return_result_in_queue,
             args=(temp_queue, thread_index, testcase_file_path, gestures,
-                  env_copy, True))
+                  env_copy, not environment.is_uworker()))
 
         try:
           thread.start()


src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -1839,7 +1839,7 @@ def do_blackbox_fuzzing(self, fuzzer, fuzzer_directory, job_type):
         thread = process_handler.get_process()(
             target=testcase_manager.run_testcase_and_return_result_in_queue,
             args=(temp_queue, thread_index, testcase_file_path, gestures,
-                  env_copy, True))
+                  env_copy, not environment.is_uworker()))
```
dylanjew, line 1842: should this just be `environment.is_tworker`?  what about swarming and off swarming bots (like macs)?

It would be nice if there was a single function we could call with confidence that determined whether we have access to GCS/metadata and cleanly handled all environments :)

dylanjew (commented):
drive by, for context I was looking into a similar issue in crbug.com/534369736

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
IvanBM18: Tworkers only pick up pre/post tasks, hence they won't end up executing this workload. 

Heres where it gets tricky, there are 2 kind of bots that can reach this code path in the fuzzing task workflow:
- uworkers &
- pre-emptible bots, these are bots that only execute fuzzing tasks, and they execute them locally(see how fuzz [is a local Utask](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/tasks/task_types.py#L248)) using a trusted service account

Pre-emptibles don't have either the uworker nor the tworker flags in place. 

So to make this work for both types of bots, we only check, if its its a uworker bot or not. There are upcoming PRs to change the main PERMISSION_DENIED issue(see b/473624472 for more context)

IvanBM18: This explains why although this issue appears on all uworkers, it doesnt completely hindered our fuzzing hours metric.

dylanjew: Got it, thanks for the explanation!

dylanjew (approved):

Xeicker (approved):
LGTM

javanlacerda (approved):

PR #5365, opened 2026-07-15
Reject TAR links from untrusted archives

### LINKED ISSUE #5364: Reject symbolic and hard links in untrusted TAR archives
TarArchiveReader validates the member pathname before extraction, but an untrusted TAR member can be a symbolic link or hard link whose target resolves outside the extraction directory.

This is unsafe for callers that later read extracted member paths as ordinary files, such as bundled testcase processing.

Reject TAR symbolic and hard-link members when trusted=False. Keep existing trusted-archive behavior unchanged, and add regression tests covering both unsafe link types.

M0nd0R:
## Summary

- Reject symbolic-link and hard-link TAR members when extracting an untrusted archive.
- Preserve link extraction for callers that explicitly mark an archive trusted.
- Add regression tests for untrusted symbolic and hard links.

## Root cause

The path-traversal guard checked a member name but did not examine the TAR member type. A safe-looking member name could therefore create a link whose target resolved outside the extraction directory.

## Impact

This prevents untrusted archive consumers, including bundled testcase processing, from later following an extracted TAR link when reading a member path.

Closes #5364

## Testing

- PYTHONPATH=src python3 focused unittest run for the two new TAR-link regressions (pass).
- The full archive test module was not runnable in this host environment because it uses Python 3.14 without the project Google dependencies; existing tests that emit error logs fail before reaching the assertions.

### DIFF
--- a/src/clusterfuzz/_internal/system/archive.py
+++ b/src/clusterfuzz/_internal/system/archive.py
@@ -295,6 +295,13 @@ class TarArchiveReader(ArchiveReader):
     output_directory = os.path.realpath(path)
 
     if not trusted:
+      tar_member = self._archive.getmember(member)
+      if tar_member.issym() or tar_member.islnk():
+        logs.error('Link member attempted while unpacking archive %s '
+                   '(member=%s, link target=%s). Aborting.' %
+                   (self._archive_path, member, tar_member.linkname))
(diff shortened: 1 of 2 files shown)


M0nd0R:
Closing.

PR #5367, opened 2026-07-15
process: Adds logs to the timeout timers that terminate processes

g-ortuno:
In https://crbug.com/529865766, we are seeing that corpus files are being deleted after the fuzzing round finishes and during corpus merging. We suspect the FuzzTest wrapper is being sent a SIGTERM signal which doesn't properly propagate to the underlying binary, which continue to fuzz.

This PR changes new_process.py to log when we send SIGTERM and SIGKILL after a timeout.

### DIFF
--- a/src/clusterfuzz/_internal/system/new_process.py
+++ b/src/clusterfuzz/_internal/system/new_process.py
@@ -84,20 +84,38 @@ def wait_process(process,
   result = ProcessResult()
   is_windows = environment.platform() == 'WINDOWS'
 
+  def create_timeout_timer(interval, function, args):
+    """Helper to create a timer that logs timeout info."""
+    def target():
+      binary_name = os.path.basename(process.command[0])
+      action_fn = args[0]
+      logs.info(
+          f"TIMEOUT: Terminating {binary_name} after {interval} seconds "
+          f"(using {action_fn.__name__})."
+      )
+      function(*args)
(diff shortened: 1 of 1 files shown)


g-ortuno:
dylanj: PTAL!

src/clusterfuzz/_internal/system/new_process.py
```diff
@@ -84,20 +84,38 @@ def wait_process(process,
   result = ProcessResult()
   is_windows = environment.platform() == 'WINDOWS'
 
+  def create_timeout_timer(interval, function, args):
```
dylanjew, line 87: nit: is this the standard API for Timers? It might be helpful to add the list of Args to the `create_timeout_timer` docstring

dylanjew (approved):

src/clusterfuzz/_internal/system/new_process.py
g-ortuno: Yeah, it takes the same arguments as Timers. Added the args to the docstring and mentioned that it's the same args as Timers.

g-ortuno:
@ViniciustCosta PTAL! I tested on dev and I can see the logs and there are no issues.

dylanjew (approved):

ViniciustCosta (approved):

PR #5368, opened 2026-07-15
[VRP] Harden testcase safety check in long-lived bots

PauloVLB:
### Problem
Long-lived bots shouldn't be handling untrusted testcases locally (like during minimization, analyze, regression or variant tasks). The old check used the `UNTRUSTED_UTASK` env var with just a warning, which wasn't reliable.

### Proposed Solution
Harden `check_handling_testcase_safe` in `uworker_io.py` to check `testcase.trusted` and `environment.is_uworker()`. If a long-lived bot (`is_uworker()` is False) tries to handle an untrusted testcase, it will immediately exit with a fatal error. Also aligned its behavior and return values with `check_running_fuzzer_safe`.

### Notes on TODOs
- Removed the TODO in `task_creation.py` (`create_variant_tasks_if_needed`) because untrusted testcase handling is already covered.
- Updated the TODO in `corpus_pruning_task.py` to specify "engine fuzzing": the untrusted testcases we are supporting are strictly for blackbox fuzzing, which do not reach the corpus prunning task.

### Testing
- Added unit tests in `uworker_io_test.py` covering trusted testcases, untrusted testcases on uworkers, and fatal exits on long-lived bots.
- Updated test helpers (`test_utils.py`, `utask_helpers.py`, `minimize_task_test.py`) to default testcases to `trusted=True` so unrelated tests don't trigger the fatal safety check.


src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
dylanjew, line 235: nit: update this docstring to reflect that it now returns True or False

src/clusterfuzz/_internal/bot/tasks/task_creation.py
```diff
@@ -149,8 +149,6 @@ def create_regression_task_if_needed(testcase):
 
 def create_variant_tasks_if_needed(testcase):
   """Creates a variant task if needed."""
-  # TODO(https://b.corp.google.com/issues/328691756): Allow untrusted
```
dylanjew, line 152: is there work required here? Should we not schedule variants which require long lived, trusted bots for untrusted testcases? e.g. windows/mac?

PauloVLB: This is already handled in [here](https://github.com/google/clusterfuzz/pull/5368/changes#diff-9f085b3e9198cda2a4e18e57b460286070ccccbdfa50e152f37688c79be1375dL194-L197). 

If the testcase is untrusted, we skip scheduling variant tasks for jobs that are not considered "no privilege workloads" (i.e., long-lived bots like Windows/Mac) when remote execution is enabled. Long-lived bots are privileged, so `is_no_privilege_workload` returns `False`, causing the loop to `continue` and skip scheduling them.

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
PauloVLB: Done.

src/clusterfuzz/_internal/bot/tasks/task_creation.py
dylanjew: ah, I didn't realize you already handled it further below. thanks for clarifying!

dylanjew (approved):

src/clusterfuzz/_internal/bot/tasks/utasks/corpus_pruning_task.py
```diff
@@ -887,8 +887,8 @@ def _process_corpus_crashes(output: uworker_msg_pb2.Output):  # pylint: disable=
       absolute_testcase_path = os.path.join(
           environment.get_value('FUZZ_INPUTS'), 'testcase')
 
-      # TODO(https://b.corp.google.com/issues/328691756): Set trusted based on
-      # the job when we start doing untrusted fuzzing.
+      # TODO: Set trusted based on
```
ViniciustCosta: nit: this is not enforced on google style for open source, but wdyt about using go/todo-style and keeping an associated bug :) (just try to link with b/ instead of the b.corp...)

ViniciustCosta (approved):

src/clusterfuzz/_internal/bot/tasks/utasks/corpus_pruning_task.py
PauloVLB: Done. I’ve filed b/556173460 and added it to the comment. Thanks for the tip!

PR #5372, opened 2026-07-21
Remove .exe on targets before appending .runtime_deps

notvictorl:
`ChromeBuildArchive`s look at `.runtime_deps` files for respective fuzzers to unpack the paths necessary to run. For Windows, fuzzers have `.exe` file extensions (`fuzzer.exe`), however, the associated `.runtime_deps` file is `fuzzer.runtime_deps`, not `fuzzer.exe.runtime_deps`. Thus, simply appending `.runtime_deps` means the associated `.runtime_deps` file won't be found and we can't reliably extract all the needed runtime dependencies.

Removing `.exe` suffix if it exists before appending `.runtime_deps` to find the correct deps file. Adding unit tests that verify we correctly access `.runtime_deps` file and extract the expected dependencies.

Bug: https://crbug.com/536913420

### DIFF
--- a/src/clusterfuzz/_internal/build_management/build_archive.py
+++ b/src/clusterfuzz/_internal/build_management/build_archive.py
@@ -432,7 +432,12 @@ class ChromeBuildArchive(DefaultBuildArchive):
   def get_target_dependencies(
       self, fuzz_target: str) -> List[archive.ArchiveMemberInfo]:
     target_path = self.get_path_for_target(fuzz_target)
-    deps_file = f'{target_path}.runtime_deps'
+    if not target_path:
+      logs.warning(f'Target path not found for {fuzz_target}')
+      return super().get_target_dependencies(fuzz_target)
+
+    target_name = target_path.removesuffix('.exe')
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -432,7 +432,12 @@ def _get_common_files(self) -> List[str]:
   def get_target_dependencies(
       self, fuzz_target: str) -> List[archive.ArchiveMemberInfo]:
     target_path = self.get_path_for_target(fuzz_target)
-    deps_file = f'{target_path}.runtime_deps'
+    if not target_path:
+      logs.warning(f'Target path not found for {fuzz_target}')
+      return super().get_target_dependencies(fuzz_target)
+
+    target_name = target_path.removesuffix('.exe')
```
dylanjew, line 444: nit: add a comment explaining why windows has a special case

dylanjew (approved):

dylanjew (approved):

vitaliset (approved):

PR #5371, opened 2026-07-21
[Android v4 API] Migrate fetch_artifact to use new V4 API

IvanBM18:
## Overview

Bug: b/422775458

Our previous implementation of the android builv v4 api was wrong, it failed always because the api no longer worked on discovery service url. In other [PR](https://github.com/google/clusterfuzz/pull/5370) we focused on creating the new api to interact with the v4 requirements. In this PR we focus on using this api class in the `fetch_artifact.py` module,

## Changes
- Made private all the methods that were only used internally in `fetch_artifact.py`
- Refactors previous usage of the v4 api to now use the current version with minimal changes in the `fetch_artifact.py` script to avoid breaking anything

## Tests performed
- Created unit tests for `feth_artifact.py`
- Tested this script using [a custom script](https://github.com/google/clusterfuzz/compare/android-build-v4-testing) (test details in [this PR description](https://github.com/google/clusterfuzz/pull/5370) )

#### PR review
1. [**#5370** Add Android Build API V4 REST client and tests ](https://github.com/google/clusterfuzz/pull/5370)
2. 👉 #5371 Migrate fetch_artifact to use new V4 API *(This PR)*
3. [#5373 Creates Feature Flag for Android V4 API](https://github.com/google/clusterfuzz/pull/5373)

Please only approve, I'll merge this PR's to avoid any possible merge conflict that can appear in the process :D

PR #5373, opened 2026-07-21
[Android V4 api Migration] Creates Feature Flag

IvanBM18:
## Overview

As per requested, we created a new feature flag to disable definitely the API build use cases(used trough `fetch_artifact.py`), this feature flag is enabled by default, because that's the current API & ClusterFuzz workflow.

But can be disabled by adding the feature flag to the DB and marking it as enabled=false.

## Changes
- Creates new feature_flag, and enables it by default in `fetch_artifact.py`
  - Logs warnings when the feature flag is disabled in `fetch_artifact.py`
- Safely exists the related methods in `symbols_downloader` and `flash.py` which consume the `fetch_artifact.py`
- Moved up the `adb.bad_state_reach()` method from the `fetch_artifact.py` module to its callers, this way we can now handle better the None values and raised errors from the API being turned off by the feature flag.

## Tests
After the initial tests in the PR parents, i tested this changes once more in dev(including all 3 prs), looking specifically for `utils.fetch_url` or errors inherent to the return value from this method. Basically since my changes landed in dev the following error groups where seen:
image

Discarded the errors following this logic:
- 403 Errors are expected to appear as per the changes that were previously there to avoid them are no longer in dev
- Some error groups only appeared on previous clusterfuzz versions 
- Other errors are expected(like the queue size limit or bad revision errors)
- There are errors from corpus pruning & fuzzer not found, this have been seen for a while now, not related to my changes
- Lastly there are some errors in a cron_job, specifically in [this file ](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/cron/load_bigquery_stats.py)but this one doesn't make usage of `fetch_url` and nor does the method that triggered the error

#### PR review
1. [**#5370** Add Android Build API V4 REST client and tests ](https://github.com/google/clusterfuzz/pull/5370)
2. [#5371 Migrate fetch_artifact to use new V4 API](https://github.com/google/clusterfuzz/pull/5371) 
3.  👉 [#5373 Creates Feature Flag for Android V4 API](https://github.com/google/clusterfuzz/pull/5373) *(This PR)*

Please only approve, I'll merge this PR's to avoid any possible merge conflict that can appear in the process :D

### DIFF
--- a/src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
+++ b/src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
@@ -25,6 +25,7 @@ from typing import Optional
 import apiclient
 from oauth2client.service_account import ServiceAccountCredentials
 
+from clusterfuzz._internal.base import feature_flags
 from clusterfuzz._internal.config import db_config
 from clusterfuzz._internal.google_cloud_utils import storage
 from clusterfuzz._internal.metrics import logs
@@ -62,6 +63,14 @@ def _use_v4():
         'Defaulting to False.',
         error=str(e))
     return False
+
+
+def _call_android_api_enabled():
+  """Return True if we should call the Android Build API. Enabled by default."""
+  flag = feature_flags.FeatureFlags.CALL_ANDROID_API.flag
+  if flag is None:
+    return True
+  return flag.enabled
 
 
 def execute_request_with_retries(request):
@@ -332,6 +341,11 @@ def get_stable_build_info():
 
 def get_latest_artifact_info(branch, target, signed=False, stable_build=False):
   """Return latest artifact for a branch and target."""
+  if not _call_android_api_enabled():
+    logs.warning(
+        'Android build API is disabled by feature flag call_android_api.')
+    return None
+
   client = get_client()
   if not client:
     return None
@@ -405,6 +419,11 @@ def get_latest_artifact_info(branch, target, signed=False, stable_build=False):
 
 def get(bid, target, regex, output_directory, output_filename=None):
   """Return artifact for a given build id, target and file regex."""
+  if not _call_android_api_enabled():
+    logs.warning(
+        'Android build API is disabled by feature flag call_android_api.')
+    return None
+
   client = get_client()
   if not client:
     return None
--- a/src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
+++ b/src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
@@ -31,8 +31,6 @@ from clusterfuzz._internal.google_cloud_utils import storage
 from clusterfuzz._internal.metrics import logs
 from clusterfuzz._internal.system import environment
 
-from . import adb
-
 # 20 MB default chunk size.
 DEFAULT_CHUNK_SIZE = 20 * 1024 * 1024
 
@@ -281,7 +279,6 @@ def get_artifacts_for_build(client,
   if not artifacts:
     logs.error(f'No artifact found for target {target}, build id {bid}.\n'
                f'request {request_str}, results {results}')
-    adb.bad_state_reached()
 
   return artifacts
 
--- a/src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
+++ b/src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
@@ -288,7 +288,7 @@ def _get_artifacts_for_build(client,
 
   if not artifacts:
     logs.error(f'No artifact found for target {target}, build id {bid}.\n'
-               f'request {request_str}, results {results}')
+               f'results {results}')
 
   return artifacts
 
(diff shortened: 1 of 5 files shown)


src/clusterfuzz/_internal/platforms/android/flash.py
```diff
@@ -97,6 +97,9 @@ def boot_stable_build_cuttlefish(branch, target, image_directory):
   """Boot cuttlefish instance using stable build id fetched from gcs."""
   build_info = fetch_artifact.get_latest_artifact_info(
       branch, target, stable_build=True)
+  if not build_info:
```
dylanjew, line 100: what happens if this returns without the build? Does the device get stuck in a flash cycle?

dylanjew (commented):
thanks! just one question

src/clusterfuzz/_internal/platforms/android/flash.py
IvanBM18: Trueeeee forgot about what i told you in [this doc's appendix](https://docs.google.com/document/d/1NB4tRKorL3QXSU3ctmqad3IipPnI2CQn2Uh4KqW4mdo/edit?tab=t.0).

I performed an extensive investigation with jetski(TMI to digest by myself), and heres what i found out:
- All non uworker bots can update their codebase themselves trough [update_task](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/tasks/update_task.py)
- When a PHYSICAL android bot needs to flash and for some reason, the artifact/build is not found, we reach this [bad_state](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/platforms/android/adb.py#L67) method, which causes the android device to sleep for 1 hour.
- After that the bot wakes up and since we are still in a loop it updates itself and tries to work again, so if lets say we introduce a defect in the code then the bot will just update itself each hour till we eventually publish the fix
- If there's a none code issue(like a failing api) or something is inherently wrong with the bot then the bot will just continue to operate till it hits [another timeout](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/datastore/data_handler.py#L1190)

So the way i see it here is to ensure that if a bot fails it calls the [bad_state](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/platforms/android/adb.py#L67) function to trigger the whole self-healing process, which we already do(see changes in flash.py)


For emulated uworker devices(like in swarming), we are 100% sure that they wont try to flash themselves, so theres no worry in here. The other place where we could run into an android api v4 issue is at [symbols_downloader.py](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/platforms/android/symbols_downloader.py) where it tries to get artifacts to process a stack-trace of errors, but i also investigated if this could cause the bot to loop, and the validations that we have in place avoid the uworker bot from crashing and just ignores the stack_trace symbolization.

dylanjew: Thanks! This flag is fine as long as the bot is in the retry loop but isn't actually making a bunch of calls that we know will fail.

src/clusterfuzz/_internal/platforms/android/flash.py
```diff
@@ -179,12 +182,21 @@ def flash_to_latest_build_if_needed():
...
                'branch %s and target %s.' % (branch, target))
-    return
+
+  has_local_images = os.path.exists(image_directory) and bool(
+      os.listdir(image_directory))
+  if not build_info and not has_local_images:
+    logs.error(
+        'No valid build info available (API may be disabled) and no local '
+        'images found in %s. Bot cannot recover.' % image_directory)
+    adb.bad_state_reached()
```
dylanjew: why did you remove the return and add handling for `build_info` being `None`? 

Wouldn't we just want to short circuit at this point? Or does `adb.bad_state_reached()` act as a short circuit from here

IvanBM18: `adb.bad_state_reached()` acts as shortcircuit as it calls `sys.exit(-1)`  so thats why i removed the return.

I added handling for None as now we can return more Nones when the `CALL_ANDROID_API` is off

dylanjew: Thanks for the explanation! I guess the second part of my question was whether we should short circuit rather than handle the case when `build_info` is `None`

dylanjew (approved):

dylanjew:
LGTM! last question is more of a nit

src/clusterfuzz/_internal/platforms/android/flash.py
IvanBM18: Oh i misunderstood your question, yes preferably its best that we return early, so adding back the early return!

src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
```diff
@@ -72,6 +71,20 @@ def _use_v4():
...
+  Disabled always if invoked in a uworker
+  """
+  if environment.is_uworker():
+    logs.info('AndroidBuildAPI access disabled for uworker.')
+    return False
+
+  flag = feature_flags.FeatureFlags.CALL_ANDROID_API.flag
+  if flag is None:
+    return True
+  return flag.enabled
```
javanlacerda, line 85:
```suggestion
  return flag.enabled if flag else True
```

javanlacerda (approved):

PR #5374, opened 2026-07-22
symbolizer: Temporarily increase llvm-symbolizer cache size on macOS

g-ortuno:
llvm-symbolizer has been crashing for a few weeks: https://crbug.com/532093354.

We suspect it's because the size of Chromium or dSYM has increased past the cache size and is triggering a bug in LLVM.

This PR increases the cache size for llvm-symbolizer on macOS to get around the bug and avoid crashing.


g-ortuno:
@dylanjew PTAL!

vitaliset (approved):

dylanjew (approved):

PR #5375, opened 2026-07-22
[NDB PERMISSION_DENIED error] Fix trusty stacktrace on ANDROID_EMULATOR

IvanBM18:
This PR comes as a need after the revert of this https://github.com/google/clusterfuzz/pull/5339 PR which introduced short circuit mechanisms to avoid DB queries, this however proved to provoke issues, so short circuit is no longer an option to fix any PERMISION_DENIED erros.

When we run jobs with an `ANDROID_EMULATOR` jobs, they all fail after executing a testcase. This happens because `ANDROID_EMULATOR` platform goes trough some legacy code(before the pre/main/pos separation of tasks) which expects the current bot to have access to DB or GCS.

This error only appear for `uworkers` and doesn't happen for `pre-emptible` bots. Because pre-emptibles do contain a trusted service account, where as uworkers dont.

## Changes
- When entering the process in which we analyze the stacktrace we now check if we can process this stacktrace or not, validating the kind of bot we are at.
- Creates Unit tests for this new edge cases

## Test performed
Changes initially present in dev since July 13th, and it didn't seem to non brake anything under ANDROID_EMULATOR jobs. Now for `ANDROID_EMULATOR` jobs.

Before:
image

After:
image

### DIFF
--- a/src/clusterfuzz/_internal/crash_analysis/stack_parsing/stack_symbolizer.py
+++ b/src/clusterfuzz/_internal/crash_analysis/stack_parsing/stack_symbolizer.py
@@ -596,6 +596,19 @@ class SymbolizationLoop:
         pass
 
   def _can_process_trusty_stack_trace(self, trusty_app, trusty_bid) -> bool:
+    """Checks whether the trusty stacktrace can be processed.
+
+    Uworker bots can't access DB nor GCS, which are required to process the
+    trusty stacktrace. This checks if the current bot context 
+    is able to process the stacktrace.
+
+    Args:
+      trusty_app: The name of the crashed Trusted App.
+      trusty_bid: The build ID of the crashed Trusted App.
+
+    Returns:
+      True if the trusty stacktrace can be processed, False otherwise.
+    """
     symbols_dir = environment.get_value('SYMBOLS_DIR')
     return bool(
         trusty_app and trusty_bid and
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/crash_analysis/stack_parsing/stack_symbolizer.py
```diff
@@ -611,11 +595,22 @@ def _close_pipes(self):
       except ProcessLookupError:
         pass
 
+  def _can_process_trusty_stack_trace(self, trusty_app, trusty_bid) -> bool:
+    symbols_dir = environment.get_value('SYMBOLS_DIR')
+    return bool(
```
dylanjew, line 613: can you add a comment similar to your description here to explain?

dylanjew (approved):

src/clusterfuzz/_internal/crash_analysis/stack_parsing/stack_symbolizer.py
IvanBM18: Done! i added them trough a pydoc instead so it looks cleaner :D

dylanjew: thank you, I appreciate it!

dylanjew (approved):

javanlacerda (approved):

PR #5377, opened 2026-07-23
[NDB PERMISSION_DENIED] Refactor _to_fuzzer_run_output to use FuzzerRunOutputData

IvanBM18:
Bug: b/532610906

## Context
- This PR comes as a need after the revert of this https://github.com/google/clusterfuzz/pull/5339 PR which introduced short circuit mechanisms to avoid DB queries, this however proved to provoke issues, so short circuit is no longer an option to fix any PERMISION_DENIED erros.
- There were 2 instances of PERMISSION_DENIED issues, the first one is already solved trough this https://github.com/google/clusterfuzz/pull/5375

## Overview
This is the second PR(Of 4) to tackle an instance of PERMISSION_DENIED issues when executing a BlackBox test case in a `uworker` bot.

This PR refactors the method that encapsulates the fuzzer output data into the results proto to use the new `FuzzerRunOutput` class, this way we can reuse the `to_fuzzer_run_output()` for both BlackBox and Coverage guided fuzzing tasks

## Changes
- `to_fuzzer_run_output` now accepts only a data class, instead of various parameters

## PR stack
PR #5376: Introduce FuzzerRunOutputData and utils (Base: master)
Current -> PR #5377: Refactor _to_fuzzer_run_output to use FuzzerRunOutputData (Base: PR 1)
PR #5388: refactor: Add utils.create_temp_file helper (Base: PR 2)
PR #5378: Defer Blackbox Fuzzer Log Uploads (Base: PR 3)

PR #5376, opened 2026-07-23
[NDB PERMISSION_DENIED] Introduce FuzzerRunOutputData and utils

IvanBM18:
Bug: b/532610906

## Context
- This PR comes as a need after the revert of this https://github.com/google/clusterfuzz/pull/5339 PR which introduced short circuit mechanisms to avoid DB queries, this however proved to provoke issues, so short circuit is no longer an option to fix any PERMISION_DENIED erros.
- There were 2 instances of PERMISSION_DENIED issues, the first one is already solved trough this https://github.com/google/clusterfuzz/pull/5375

## Overview
This First PR(Of 4) to tackle an instance of PERMISSION_DENIED issues when executing a BlackBox test case in a `uworker` bot.

This PR creates 2 new classes that will help us handle the processing of a test case execution in the next prs

## Changes
- Introduces two new classes to encapsulate all the info obtained from a BlackBox test case execution
  - The fuzzer output can be received either trough bits, a string, or a log file
- Adds a new utils logic to read and delete a file after reading

## PR stack
Current -> PR #5376: Introduce FuzzerRunOutputData and utils (Base: master)
PR #5377: Refactor _to_fuzzer_run_output to use FuzzerRunOutputData (Base: PR 1)
PR #5388: refactor: Add utils.create_temp_file helper (Base: PR 2)
PR #5378: Defer Blackbox Fuzzer Log Uploads (Base: PR 3)

Please only approve I'll merge the PRs to handle any merge conflicts :D

PR #5378, opened 2026-07-23
[NDB PERMISSION_DENIED] Defer Blackbox Fuzzer Log Uploads

IvanBM18:
Bug: b/532610906

## Context
- This PR comes as a need after the revert of this https://github.com/google/clusterfuzz/pull/5339 PR which introduced short circuit mechanisms to avoid DB queries, this however proved to provoke issues, so short circuit is no longer an option to fix any PERMISION_DENIED errors.
- There were 2 instances of PERMISSION_DENIED issues, the first one is already solved trough this https://github.com/google/clusterfuzz/pull/5375

## Overview
This is the last PR to tackle an instance of PERMISSION_DENIED issues when executing a BlackBox test case in a `uworker` bot.

This PR refactors uses the utils, classes and refactors to previous PRs to defer the blackbox fuzzer log upload to the postprocess step. Since the blackbox fuzzer logs can be larger than regular Coverage guided engine fuzzer logs, we cant just simply save its content into the crash queue, instead we save the fuzzer output into a temp file and we save that file in the crashes queue.

Then when we finish executing all the test cases(which can be multiple concurrent ones), we drain the crash queue, retrieving the log file paths and transforming them into an instance of `FuzzerRunOutputData` class, which is then placed into the fuzzer session output to be handled & uploaded at the postprocess step

## Changes
- Fuzzer log for BB fuzzer are now saved into temp files instead of the actual string
  - This is done because we use a special type of queue to handle multiple concurrent fuzzing outputs, this queue has a hard limit on the amount of data that it can handle, 
    - So if a single fuzzer uploads to much info then the queue will  block operations until someone pulls data from the queue
    - if we have a job that runs multiple test cases simultaneously & a fuzzer that logs too much info we can end up in a livelock multithreading problem
- Creates a new method to drain the queue for multiple crash results
  - Encapsulates the fuzzer logs from the queue using the FuzzerRunOutputData
- Since we no longer upload the logs at the utask_main portion, we deleted the `upload` parameter for the run_tetcase method

## Tests performed
Changes in dev since July 23rd, from which we can see how the errors dramatically went down after the changes were introduced:

image

The other 403 issues that still appear in dev are, from custom_binary tests being performed by teammates in dev, if we remove those logs, we can see how the errors go to 0
image

Now when we execute test cases we no longer see the PERMISION_DENIED errors and at postprocess we correctly upload all logs to GCS:
image

## PR stack
PR #5376: Introduce FuzzerRunOutputData and utils (Base: master)
PR #5377: Refactor _to_fuzzer_run_output to use FuzzerRunOutputData (Base: PR 1)
PR #5388: refactor: Add utils.create_temp_file helper (Base: PR 2)
Current ->  PR #5378: Defer Blackbox Fuzzer Log Uploads (Base: PR 3)

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
@@ -18,6 +18,7 @@ import datetime
 import itertools
 import json
 import os
+import queue
 import random
 import re
 import time
@@ -1739,6 +1740,19 @@ class FuzzingSession:
               'platform': environment.platform(),
           })
 
+  def _drain_temp_queue(self, temp_queue: queue.Queue,
+                        crashes: list[testcase_manager.Crash]) -> None:
+    """Pops items from temp_queue and processes crashes and run outputs."""
+    while not temp_queue.empty():
+      result: testcase_manager.TestcaseRunResult = temp_queue.get()
+
+      if result.crash:
+        crashes.append(result.crash)
+      if result.fuzzer_run_output_data:
+        fuzzer_run_output = _to_fuzzer_run_output(result.fuzzer_run_output_data)
+        if fuzzer_run_output:
+          self.fuzz_task_output.fuzzer_run_outputs.append(fuzzer_run_output)
+
   def do_blackbox_fuzzing(self, fuzzer, fuzzer_directory, job_type):
     """Run blackbox fuzzing. Currently also used for engine fuzzing."""
     # Set the thread timeout values.
@@ -1848,7 +1862,7 @@ class FuzzingSession:
         thread = process_handler.get_process()(
             target=testcase_manager.run_testcase_and_return_result_in_queue,
             args=(temp_queue, thread_index, testcase_file_path, gestures,
-                  env_copy, not environment.is_uworker()))
+                  env_copy))
 
         try:
           thread.start()
@@ -1881,9 +1895,7 @@ class FuzzingSession:
         process_handler.terminate_stale_application_instances()
         needs_stale_process_cleanup = False
 
-      while not temp_queue.empty():
-        crashes.append(temp_queue.get())
-
+      self._drain_temp_queue(temp_queue, crashes)
       process_handler.close_queue(temp_queue)
       logs.info(f'Upto {test_number}')
       if thread_error_occurred:
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
@@ -1740,8 +1740,9 @@ class FuzzingSession:
               'platform': environment.platform(),
           })
 
-  def _drain_temp_queue(self, temp_queue: queue.Queue,
-                        crashes: list[testcase_manager.Crash]) -> None:
+  def _add_crashes_from_temp_queue(
+      self, temp_queue: queue.Queue,
+      crashes: list[testcase_manager.Crash]) -> None:
     """Pops items from temp_queue and processes crashes and run outputs."""
     while not temp_queue.empty():
       result: testcase_manager.TestcaseRunResult = temp_queue.get()
@@ -1895,7 +1896,7 @@ class FuzzingSession:
         process_handler.terminate_stale_application_instances()
         needs_stale_process_cleanup = False
 
-      self._drain_temp_queue(temp_queue, crashes)
+      self._add_crashes_from_temp_queue(temp_queue, crashes)
       process_handler.close_queue(temp_queue)
       logs.info(f'Upto {test_number}')
       if thread_error_occurred:
--- a/src/clusterfuzz/_internal/bot/testcase_manager.py
+++ b/src/clusterfuzz/_internal/bot/testcase_manager.py
@@ -18,6 +18,7 @@ import collections
 import dataclasses
 import datetime
 import os
+import queue
 import re
 import zlib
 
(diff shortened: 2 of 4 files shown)


src/clusterfuzz/_internal/bot/testcase_manager.py
```diff
@@ -586,28 +576,36 @@ def _do_run_testcase_and_return_result_in_queue(crash_queue,
...
+          return_code=return_code,
+          resource_list=resource_list,
+          gestures=gestures,
+          stack_file_path=stack_file_path)
+
+    # Save raw, unsymbolized execution output for deferred postprocessing.
+    crash_result_full = CrashResult(return_code, crash_time, output)
+    unsymbolized_output = crash_result_full.get_stacktrace(symbolized=False)
+    crash_path = file_path if crash else None
+    log_file_path = utils.create_temp_file(
```
dylanjew, line 596: FYI @aakallam had a clean up PR https://github.com/google/clusterfuzz/pull/5346 that improves how we clean up file handles to avoid a bug on Windows.

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -1738,6 +1739,19 @@ def _emit_testcase_generation_time_metric(self, start_time, testcase_count,
               'platform': environment.platform(),
           })
 
+  def _drain_temp_queue(self, temp_queue: queue.Queue,
```
dylanjew: nit: I would name this something more like `add_crashes_from_temp_queue`

src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/fuzz_task_test.py
```diff
@@ -1332,6 +1332,39 @@ def test_trials(self):
...
+    self.mock.get_process.return_value = mock_thread_cls
+
+    fuzzer = data_types.Fuzzer()
+    fuzzer.name = 'fantasy_fuzz'
+
+    session.do_blackbox_fuzzing(fuzzer, '/fake-fuzz-dir', 'asan_test')
+
+    self.assertEqual(2, mock_thread_cls.call_count)
+    for _, kwargs in mock_thread_cls.call_args_list:
+      self.assertEqual(5, len(kwargs['args']))
```
dylanjew: what is this checking for? it seems frail to check the number of args = 5

src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/fuzz_task_test.py
```diff
@@ -1372,6 +1405,155 @@ def test_utask_main_blackbox_fuzzer(self, mock_generate_blackbox_testcases):
...
+    mock_thread_cls = mock.MagicMock(wraps=threading.Thread)
+    self.mock.get_process.return_value = mock_thread_cls
+
+    session.do_blackbox_fuzzing(fuzzer, '/fake-fuzz-dir', 'asan_test')
+
+    self.assertEqual(2, mock_thread_cls.call_count)
+    for _, kwargs in mock_thread_cls.call_args_list:
+      self.assertEqual(
+          5, len(kwargs['args']),
+          'run_testcase_and_return_result_in_queue must take 5 positional args')
```
dylanjew: same here, I think you can rely on assertions for the `fuzzer_run_output` instead

src/clusterfuzz/_internal/tests/core/bot/testcase_manager_test.py
```diff
@@ -864,6 +865,102 @@ def test_test_for_reproducibility_greybox_succeed_after_multiple_tries(self):
...
+    crash_queue = queue.Queue()
+    self.mock.run_testcase.return_value = (0, 1.5, 'raw_output_trace')
+    self.mock.is_crash.return_value = False
+    self.mock._get_testcase_time.return_value = datetime.datetime(2026, 7, 10)
+    mock_unsymbolized_info = mock.MagicMock()
+    mock_unsymbolized_info.crash_stacktrace = 'unsymbolized_raw_trace'
+    self.mock.get_crash_data.return_value = mock_unsymbolized_info
+
+    testcase_manager.run_testcase_and_return_result_in_queue(
+        crash_queue, 0, '/testcase', [], {})
```
dylanjew: nit: add the param names so its clear what you're passing in

src/clusterfuzz/_internal/bot/testcase_manager.py
aakallam: As far as Windows I/O goes, this is fine because `create_temp_file()` is generating it with `tempfile.mkstemp()` and then closing the file handle before returning the path.

However, does this file get manually cleaned up somewhere?

IvanBM18: Yes, its deleted [here](https://github.com/google/clusterfuzz/pull/5376/changes#diff-b825d675f21145ba1281303c6b56231da8052c7977fe45ca80372569761a9812R680), basically we delete the file content reading its content

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
IvanBM18: Done!

src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/fuzz_task_test.py
IvanBM18: True...  now we check for specific info instead, thanks!

src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/fuzz_task_test.py
IvanBM18: Done!

src/clusterfuzz/_internal/tests/core/bot/testcase_manager_test.py
IvanBM18: Done!

dylanjew (approved):

javanlacerda (approved):

PR #5380, opened 2026-07-23
Gracefully handle invalid ghost fuzzer configurations

dylanjew:
When a job's fuzzer gets detached, it leaves a FuzzerJob mapping with an empty string for the fuzzer.  

I'm not sure whether there were new tasks being scheduled or it was the existing ones blocking up the queue, but this PR focuses on clearing out the queue.

When the workers picked up these ghost tasks, setup.py would fetch an invalid Fuzzer entity that lacked a blobstore_key and pass it to blobs.py in get_gcs_path, which triggered a TypeError.

This change raises an InvalidFuzzerError up front instead, gracefully clearing the bad tasks without crashing.

This doesn't fix the issue where we're scheduling invalid fuzz tasks, but it should clear the queue by ACKing these bad tasks.

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
@@ -29,6 +29,7 @@ from typing import Optional
 from google.cloud import ndb
 
 from clusterfuzz._internal.base import dates
+from clusterfuzz._internal.base import errors
 from clusterfuzz._internal.base import utils
 from clusterfuzz._internal.bot import testcase_manager
 from clusterfuzz._internal.bot.fuzzers import builtin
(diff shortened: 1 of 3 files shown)


src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -1216,15 +1217,15 @@ def write_crashes_to_big_query(group, newly_created_testcase, existing_testcase,
       # Happens in case the big query function is disabled (local development).
       return
 
-    errors = result.get('insertErrors', [])
-    failed_count = len(errors)
+    insert_errors = result.get('insertErrors', [])
```
dylanjew, line 1220: name conflict with the package

vitaliset (approved):

dylanjew:
@vitaliset Fixed the test

PR #5384, opened 2026-07-23
Fix MinimizeTaskTestUntrusted.test_minimize test

dylanjew:
Broken in https://github.com/google/clusterfuzz/pull/5380


dylanjew:
It turns out this was an important test! 😞

PR #5385, opened 2026-07-23
[DO NOT SUBMIT] Clean up orphaned FuzzerJob mapping entities

dylanjew:
Previously, `update_mappings_for_job` iterates over valid Fuzzer parent entities and adds a `FuzzerJob`. 

This means that any invalid `FuzzerJob` mappings lacking a parent Fuzzer are permanently orphaned in Datastore.

This PR adds a clean up script that looks for `FuzzerJobs` and deletes any entities that are no longer linked to a valid Fuzzer.

One thing to note: the reason Clusterfuzz thinks that FuzzerJobs with `fuzzer=""` is valid is because we have a Fuzzer without a name in the datastore, which we didn't delete because it's unsafe to do so.

There aren't any other entities we need to clean up besides the FuzzerJob. 

We query the `FuzzerJob`s [here](https://github.com/google/clusterfuzz/blob/05360c5aa9db354f6a9d83a105583eae7a230551/src/clusterfuzz/_internal/cron/schedule_fuzz.py#L276) and that creates the invalid task to schedule. Batch is done [here](https://github.com/google/clusterfuzz/blob/05360c5aa9db354f6a9d83a105583eae7a230551/src/clusterfuzz/_internal/cron/batch_fuzzer_jobs.py#L32)

Since the invalid `FuzzerJob` without a matching `Fuzzer` isn't validated before we create the task, we schedule the invalid task and flood our queue if we can't process those.

## Roll out

I don't plan to submit this PR. I'm just going to run this against our prod DB after verifying in dev and in dry run mode

### Dev

I ran this in dev and verified it deleted some invalid mappings for a blackbox fuzzer `ochang_js_fuzzer_untrusted`

#### Dry run mode results
```
python butler.py run -c   ../clusterfuzz-config/configs/internal  clean_invalid_fuzzer_jobs

/usr/local/google/home/dylanj/clusterfuzz/src/third_party/google/cloud/ndb/__init__.py:24: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  from pkg_resources import get_distribution
Running in dry-run mode, no datastore writes are committed. For permanent modifications, re-run with --non-dry-run.
Starting invalid FuzzerJob cleanup pass.
Found invalid mapping to delete: ID=5114699494490112, fuzzer="", job="centipede_chrome_asan_dchecks_highend"
Found invalid mapping to delete: ID=5319049844129792, fuzzer="", job="centipede_chrome_asan_highend"
Found invalid mapping to delete: ID=5534634553147392, fuzzer="", job="libfuzzer_chrome_asan_highend"
Found invalid mapping to delete: ID=6357342853562368, fuzzer="", job="libfuzzer_chrome_asan_debug_highend"
Total invalid mappings found: 4
Run with --non-dry-run to actually delete the mappings.
```

### DIFF
--- a/src/local/butler/scripts/clean_invalid_fuzzer_jobs.py
+++ b/src/local/butler/scripts/clean_invalid_fuzzer_jobs.py
@@ -1,0 +1,47 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Cleans up orphaned FuzzerJob entities that have no corresponding Fuzzer."""
+
+from clusterfuzz._internal.datastore import data_types
+from clusterfuzz._internal.datastore import ndb_utils
+
+
+def execute(args):
+  """Query all FuzzerJobs and delete any mapped to a non-existent fuzzer."""
+  print('Starting invalid FuzzerJob cleanup pass.')
+
+  valid_fuzzers = {fuzzer.name for fuzzer in data_types.Fuzzer.query()}
+
+  fuzzer_jobs = data_types.FuzzerJob.query()
+
+  fuzzer_jobs_to_delete = []
+  for mapping in fuzzer_jobs:
+    # A invalid FuzzerJob has an empty fuzzer name or a fuzzer name
+    # that no longer maps to an active Fuzzer entity.
+    if not mapping.fuzzer or mapping.fuzzer not in valid_fuzzers:
+      fuzzer_jobs_to_delete.append(mapping.key)
+      print(f'Found invalid mapping to delete: ID={mapping.key.id()}, '
+            f'fuzzer="{mapping.fuzzer}", job="{mapping.job}"')
+
+  print(f'Total invalid mappings found: {len(fuzzer_jobs_to_delete)}')
+
+  if not fuzzer_jobs_to_delete:
+    print('No mappings to delete.')
+
+  if args.non_dry_run:
+    print('Executing datastore deletion...')
+    ndb_utils.delete_multi(fuzzer_jobs_to_delete)
+    print('Deletion complete.')
+  else:
+    print('Run with --non-dry-run to actually delete the mappings.')


src/local/butler/scripts/clean_invalid_fuzzer_jobs.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def execute(args):
+  """Query all FuzzerJobs and delete any mapped to a non-existent fuzzer."""
+  print('Starting invalid FuzzerJob cleanup pass.')
+
+  valid_fuzzers = {fuzzer.name for fuzzer in data_types.Fuzzer.query()}
+
+  query = data_types.FuzzerJob.query()
+  fuzzer_jobs = ndb_utils.get_all_from_query(query)
```
dylanjew: oops, this isn't necessary. I'm going to remove it

dylanjew:
I ran this script and confirmed they are deleted in the datastore

src/local/butler/scripts/clean_invalid_fuzzer_jobs.py
```diff
@@ -0,0 +1,47 @@
...
+            f'fuzzer="{mapping.fuzzer}", job="{mapping.job}"')
+
+  print(f'Total invalid mappings found: {len(fuzzer_jobs_to_delete)}')
+
+  if not fuzzer_jobs_to_delete:
+    print('No mappings to delete.')
+
+  if args.non_dry_run:
+    print('Executing datastore deletion...')
+    ndb_utils.delete_multi(fuzzer_jobs_to_delete)
```
g-ortuno, line 44: Huh, i'm not quite sure how calling delete_multi with the keys of the FuzzerJobs above resulted in the right Jobs getting deleted haha

dylanjew: oh, we aren't trying to delete the `Jobs`, just the `FuzzerJob`. Sorry, this script could be much clearer. 

we want to keep the `Jobs` around because we think that deleting those may cause issues. 

I confirmed in the Datastore that the `Jobs` still exist but the invalid `FuzzerJobs` are deleted

g-ortuno: Do you have a link to the datastore with the `FuzzerJob`s? The one I'm looking at still has the high_end `FuzzerJob`s: http://shortn/_vtTIgcuINE

dylanjew: Maybe they're getting recreated now that the bad fuzzer is linked to them 😭

g-ortuno: Ah that sucks :/ well good thing we didn't revert the other change then!

dylanjew: it had a multiplier of 17.5 when I deleted it earlier, and now it has a multiplier of 1, which is the default before the cronjob updates the weights

https://github.com/google/clusterfuzz/blob/05360c5aa9db354f6a9d83a105583eae7a230551/src/clusterfuzz/_internal/datastore/data_types.py#L1575

g-ortuno: I see! Is it being recreated because it's in the `jobs` column of the libFuzzer `Fuzzer`? Do things break if we remove it from there?

dylanjew: Yes, it creates a new one here:

https://github.com/google/clusterfuzz/blob/d85e7f83839818c43a786788a529298c35b8be79/src/clusterfuzz/_internal/fuzzing/fuzzer_selection.py#L68-L70

I'm pretty sure that we can uncheck the jobs from the fuzzer. We've done that many times in the past. I'll go ahead and do it

PR #5386, opened 2026-07-23
Fix engine fuzzers blobstore downloads 

dylanjew:
https://github.com/google/clusterfuzz/pull/5380 fixed the issue where our queue was getting flooded with bad tasks but broke libfuzzer because engine fuzzers do not need a `blobstore_key`. Instead it skips the function that was causing an error for the misconfigured fuzzers

I queried the Datastore and confirmed that when builtin=True, the blobstore_key is null for the engine fuzzers: `syzkaller, googlefuzztest, centipede, libfuzzer`. Additionally, the invalid fuzzer  with `"Name/ID": "name=ochang_js_fuzzer_test"` is also there, which was what caused the initial errors

This fixed the invalid FuzzerJob handling, but caused an issue with other libfuzzer fuzzers because libfuzzer has no blobstore key, but skips this look up that would fail because it has not blobstore key since `builtin = False`
https://github.com/google/clusterfuzz/blob/05360c5aa9db354f6a9d83a105583eae7a230551/src/clusterfuzz/_internal/bot/tasks/setup.py#L606-L608

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/setup.py
+++ b/src/clusterfuzz/_internal/bot/tasks/setup.py
@@ -593,9 +593,6 @@ def preprocess_update_fuzzer_and_data_bundles(
   if not fuzzer:
     logs.error('No fuzzer exists with name %s.' % fuzzer_name)
     raise errors.InvalidFuzzerError
-  if not fuzzer.blobstore_key:
-    logs.error(f'Fuzzer {fuzzer_name} does not have a blobstore_key.')
-    raise errors.InvalidFuzzerError
 
   update_input = uworker_msg_pb2.SetupInput(  # pylint: disable=no-member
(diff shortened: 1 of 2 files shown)


g-ortuno (approved):

g-ortuno:
How hard would it be to add a test as a follow up? :eyes:

dylanjew:
> How hard would it be to add a test as a follow up? 👀

Good point. I'll add a test 😄

dylanjew:
> > How hard would it be to add a test as a follow up? 👀
> 
> Good point. I'll add a test 😄

Done!

g-ortuno:
lgtm still!

dylanjew:
Fixed formatting

vitaliset (approved):

PR #5387, opened 2026-07-23
Use fuzz_targets from clusterfuzz_manifest.json if available

notvictorl:
Chrome archives initially pick fuzz targets from a `fuzz_targets` property from `clusterfuzz_manifest.json`, but it doesn't use the same `fuzz_targets`  on future runs if the build on disk is reused. This leads to wasting time doing a directory scan for `_fuzzer` binaries to choose a fuzz target and can pick a binary that is not intended to be run (`run_*_fuzzer`). This change is largely meant for long running bots which are not used for Chrome but essentially closes the loop on using `fuzz_targets` when possible.

Modifying fuzzer utils to re-access `clusterfuzz_manifest.json` for `fuzz_targets` if available otherwise fall back to existing behavior. Adding unit tests for fuzzer utils. Refactoring `build_archive.py` and `build_manager.py` to use the new common utility functions.

Drive-by: Improve Chrome manifest path search by joining to `root_dir` for wrapped directories where ClusterFuzz's build directory is not the Chrome archives root directory with respective unit tests.

Bug: https://crbug.com/507833074

### DIFF
--- a/src/clusterfuzz/_internal/bot/fuzzers/utils.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/utils.py
@@ -32,6 +32,9 @@ FUZZ_TARGET_SEARCH_BYTES = [b'LLVMFuzzerTestOneInput', b'LLVMFuzzerRunDriver']
 VALID_TARGET_NAME_REGEX = re.compile(r'^[a-zA-Z0-9@_.-]+$')
 BLOCKLISTED_TARGET_NAME_REGEX = re.compile(r'^(jazzer_driver.*|jazzerjs)$')
 EXTRA_BUILD_DIR = '__extra_build'
+# Manifest file present in Chrome build archives. Specifies the archive
+# schema version along with optional build metadata (e.g., fuzz targets).
+CHROME_MANIFEST_FILENAME = 'clusterfuzz_manifest.json'
 
 
 def is_fuzz_target(file_path, file_opener: Optional[Callable] = None):
@@ -99,38 +102,63 @@ def is_fuzz_target(file_path, file_opener: Optional[Callable] = None):
     return False
 
 
-def _get_manifest_fuzz_targets(path):
-  """Get list of fuzz targets from clusterfuzz_manifest.json if present."""
-  manifest_path = os.path.join(path, 'clusterfuzz_manifest.json')
-  if not os.path.exists(manifest_path):
+def extract_fuzz_targets_from_manifest(
+    manifest_dict: dict,
+    manifest_filename: str = CHROME_MANIFEST_FILENAME) -> list[str] | None:
+  """Extract list of target paths from a loaded manifest dict if present."""
+  if not isinstance(manifest_dict, dict):
+    logs.error(f'{manifest_filename} is not a dict')
     return None
 
-  try:
-    with open(manifest_path, 'r', encoding='utf-8') as f:
-      manifest = json.load(f)
-  except Exception as e:
-    logs.error(f'Failed to read {manifest_path}: {e}')
-    return None
-
-  if not isinstance(manifest, dict):
-    logs.error(f'{manifest_path} is not a dict')
-    return None
-
-  manifest_targets = manifest.get('fuzz_targets')
+  manifest_targets = manifest_dict.get('fuzz_targets')
   if manifest_targets is None:
     return None
 
   if not isinstance(manifest_targets, list):
-    logs.error(f'fuzz_targets in {manifest_path} is not a list')
+    logs.error(f'fuzz_targets in {manifest_filename} is not a list')
     return None
 
   fuzz_target_paths = []
   for target_path in manifest_targets:
     if not isinstance(target_path, str):
-      logs.error(f'Entry in fuzz_targets ({manifest_path}) is not a string: '
-                 f'{target_path}')
+      logs.error(
+          f'Entry in fuzz_targets ({manifest_filename}) is not a string: '
+          f'{target_path}')
       continue
-    file_path = os.path.join(path, target_path)
+    fuzz_target_paths.append(target_path)
+
+  return fuzz_target_paths
+
+
+def read_chrome_manifest(build_dir: str) -> dict | None:
+  """Reads and parses clusterfuzz_manifest.json from build_dir if present."""
+  manifest_path = os.path.join(build_dir, CHROME_MANIFEST_FILENAME)
+  if not os.path.exists(manifest_path):
+    return None
+
+  try:
+    with open(manifest_path, 'r', encoding='utf-8') as f:
+      return json.load(f)
+  except Exception as e:
+    logs.error(f'Failed to read {manifest_path}: {e}')
+    return None
+
+
+def _get_manifest_fuzz_targets(build_dir: str) -> list[str] | None:
+  """Get list of fuzz targets from clusterfuzz_manifest.json if present."""
+  manifest = read_chrome_manifest(build_dir)
+  if manifest is None:
+    return None
+
+  manifest_path = os.path.join(build_dir, CHROME_MANIFEST_FILENAME)
+  manifest_targets = extract_fuzz_targets_from_manifest(
+      manifest, manifest_filename=manifest_path)
+  if manifest_targets is None:
+    return None
+
+  fuzz_target_paths = []
+  for target_path in manifest_targets:
+    file_path = os.path.join(build_dir, target_path)
     if os.path.exists(file_path):
       fuzz_target_paths.append(file_path)
     else:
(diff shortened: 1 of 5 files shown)


src/clusterfuzz/_internal/bot/fuzzers/utils.py
```diff
@@ -98,8 +99,52 @@ def is_fuzz_target(file_path, file_opener: Optional[Callable] = None):
     return False
 
 
+def _get_manifest_fuzz_targets(path):
```
dylanjew: should this be a shared util? https://github.com/google/clusterfuzz/blob/f547870e55c804c65212c9650af3e25736e59b23/src/clusterfuzz/_internal/build_management/build_archive.py#L317

src/clusterfuzz/_internal/bot/fuzzers/utils.py
```diff
@@ -98,8 +99,52 @@ def is_fuzz_target(file_path, file_opener: Optional[Callable] = None):
...
+    file_path = os.path.join(path, target_path)
+    if os.path.exists(file_path):
+      fuzz_target_paths.append(file_path)
+    else:
+      logs.warning(f'Fuzz target {target_path} in {manifest_path} not found.')
+
+  return fuzz_target_paths
+
+
 def get_fuzz_targets_local(path):
```
dylanjew, line 201: what is the behavior for the non local version of `get_fuzz_targets`?

src/clusterfuzz/_internal/bot/fuzzers/utils.py
```diff
@@ -98,8 +99,52 @@ def is_fuzz_target(file_path, file_opener: Optional[Callable] = None):
...
   """Get list of fuzz targets paths (local)."""
```
dylanjew: nit: update the docstring to explain the manifest and fallback behavior

src/clusterfuzz/_internal/tests/core/bot/fuzzers/utils_test.py
```diff
@@ -148,3 +149,74 @@ def test_par(self):
...
+    path = os.path.join(self.temp_dir, name)
+    os.makedirs(os.path.dirname(path), exist_ok=True)
+    with open(path, 'wb') as f:
+      f.write(contents)
+    return path
+
+  def test_manifest_targets_used(self):
+    """Test that clusterfuzz_manifest.json is used to find targets."""
+    target_a = self._create_file('target_a', contents=b'LLVMFuzzerTestOneInput')
+    self._create_file('run_target_a_fuzzer', contents=b'LLVMFuzzerTestOneInput')
```
dylanjew: nit: instead of `target_a` and `run_target_a_fuzzer` you could name these something like `manifest_target` and  `file_target` or `dir_target` and the assertions would be clearer

dylanjew (commented):
is this going to fix all of the `run_*_fuzzer` targets?

src/clusterfuzz/_internal/bot/fuzzers/utils.py
notvictorl: The non local version of `get_fuzz_targets` eventually ends up at `get_fuzz_targets_local` again. It just routes through an untrusted runner through a gRPC call first.

https://github.com/google/clusterfuzz/blob/f547870e55c804c65212c9650af3e25736e59b23/src/clusterfuzz/_internal/bot/untrusted_runner/file_impl.py#L103

src/clusterfuzz/_internal/bot/fuzzers/utils.py
notvictorl: good call. I extracted some of the shared parsing to a util function and refactored the code to use it.

src/clusterfuzz/_internal/bot/fuzzers/utils.py
```diff
@@ -99,24 +102,25 @@ def is_fuzz_target(file_path, file_opener: Optional[Callable] = None):
...
+  is in the expected format. For example, `manifest` might look like:
+  {
+    fuzz_targets: [
+      "fuzz_target_a",
+      "fuzz_target_b"
+    ]
+  }
+  """
+  if not isinstance(manifest_dict, dict):
     logs.error(f'{manifest_path} is not a dict')
```
dylanjew: nit: replace with `manifest_dict`, and same for the remaining logs.

src/clusterfuzz/_internal/bot/fuzzers/utils.py
```diff
@@ -99,24 +102,25 @@ def is_fuzz_target(file_path, file_opener: Optional[Callable] = None):
...
-    with open(manifest_path, 'r', encoding='utf-8') as f:
-      manifest = json.load(f)
-  except Exception as e:
-    logs.error(f'Failed to read {manifest_path}: {e}')
-    return None
-
-  if not isinstance(manifest, dict):
+def extract_fuzz_targets_from_manifest(
+    manifest_dict: dict,
+    manifest_path: str = CHROME_MANIFEST_FILENAME) -> list[str] | None:
```
dylanjew: do you still need to pass in a `manifest_path` here?

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -328,21 +328,10 @@ def __init__(self,
...
-      return
-
-    self._manifest_fuzz_targets = []
-    for target_path in fuzz_target_paths:
-      if isinstance(target_path, str):
-        self._manifest_fuzz_targets.append(target_path)
-      else:
-        logs.error('Entry in fuzz_targets (clusterfuzz_manifest.json) is not a '
-                   f'string: {target_path}')
+    from clusterfuzz._internal.bot.fuzzers import utils as fuzzer_utils
```
dylanjew: nit: does clusterfuzz typically do these function level imports?

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -328,21 +328,10 @@ def __init__(self,
...
+    self._manifest_fuzz_targets = (
+        fuzzer_utils.extract_fuzz_targets_from_manifest(manifest,
```
dylanjew: at what point do we log an error if there's no fuzz targets?

src/clusterfuzz/_internal/bot/fuzzers/utils.py
dylanjew: thanks!

dylanjew (approved):

src/clusterfuzz/_internal/build_management/build_archive.py
notvictorl: Not typically but in the `build_archive.py` file we were for `fuzzer_utils`. Since we were doing function level imports 3 times in the same file, I moved the import to the top of the file instead.

src/clusterfuzz/_internal/build_management/build_archive.py
notvictorl: We don't log an error if `fuzz_targets` doesn't exist in `clusterfuzz_manifest.json` because it is an optional property but instead of silently exiting `extract_fuzz_targets_from_manifest`, we could have an info log.

src/clusterfuzz/_internal/bot/fuzzers/utils.py
notvictorl: `manifest_path` was just for logging purposes to not create long error messages, but logging the actual manifest dict being worked on is more accurate.

src/clusterfuzz/_internal/build_management/build_archive.py
```diff
@@ -96,9 +97,6 @@ def list_fuzz_targets(self) -> List[str]:
         The list of fuzz targets.
     """
     if self._fuzz_targets is None:
-      # Import here as this path is not available in App Engine context.
```
dylanjew, line 99: is this comment relevant?

src/clusterfuzz/_internal/build_management/build_archive.py
dylanjew: is it temporarily optional as you start including it in all of our builds going forward?

I think adding an info log for the number of targets we find in the manifest makes sense

dylanjew (approved):

src/clusterfuzz/_internal/build_management/build_archive.py
notvictorl: I don't think so because it seems to be from a legacy architecture and has since been upgraded to not have this import issue.

dylanjew: thanks for checking

dylanjew (approved):

notvictorl:
@javanlacerda friendly nudge for review

src/clusterfuzz/_internal/build_management/build_archive.py
ViniciustCosta: How did you verify that?

ViniciustCosta (approved):

PR #5389, opened 2026-07-25
Pass audience to verify_oauth2_token in get_email_from_bearer_token

herdiyana256:
The Pub/Sub push handler authenticates the incoming OIDC token but never checks its audience:

```python
# handlers/.../get_email_from_bearer_token
claim = id_token.verify_oauth2_token(token, google_requests.Request())
```

`verify_oauth2_token` treats a missing `audience` as "do not check it". An ID token is minted for a specific destination, but any party that can mint tokens as the same service account (or that the service account itself calls out to) holds a token signed by that SA for a different audience. With no audience check here, such a token replays against this endpoint and passes.

```python
# request.url is the URL Pub/Sub actually pushed to, which a correctly
# configured OIDC push subscription sets as the token's audience.
claim = id_token.verify_oauth2_token(
    token, google_requests.Request(), audience=request.url)
```

Binding verification to `request.url` means only a token minted for this exact push endpoint is accepted, closing the cross-audience replay. Adds a regression test for the missing-audience case.

### DIFF
--- a/src/appengine/libs/auth.py
+++ b/src/appengine/libs/auth.py
@@ -144,7 +144,16 @@ def get_email_from_bearer_token(request):
 
   token = bearer_token.split(' ')[1]
   try:
-    claim = id_token.verify_oauth2_token(token, google_requests.Request())
+    # audience must be passed explicitly: verify_oauth2_token treats a
+    # missing audience as "don't check it", which would let an ID token
+    # minted for a completely different destination (but for the same
+    # service account, which can be requested by anything with permission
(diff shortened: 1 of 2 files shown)


herdiyana256:
Gentle bump, open ~3 weeks with no reviewer. Small fix: get_email_from_bearer_token verifies the bearer token without passing an audience, so the pubsub_push endpoint accepts any token bearing the right email claim. @PauloVLB @javanlacerda could one of you take a look or route it?

PR #5390, opened 2026-07-25
Require testcase access before disclosing crash_query duplicates

herdiyana256:
`crash_query` looks up a duplicate testcase by project name and crash signature and returns its identifiers, but only withheld the result when the duplicate was security-flagged:

```python
if (duplicate_testcase and duplicate_testcase.security_flag and
    not access.can_user_access_testcase(duplicate_testcase)):
  # only security-flagged duplicates were protected
```

`find_testcase` filters by project and crash signature only, never by whether the caller has any relationship to that project. So for a non-security-flagged testcase the branch above was skipped and the result was returned to anyone. Any authenticated user could pass an arbitrary project name and learn whether a matching testcase exists there, along with its `duplicate_id`/`bug_id`, regardless of project ownership.

```python
if duplicate_testcase and not access.can_user_access_testcase(
    duplicate_testcase):
  # gate every disclosure on caller access, not just security-flagged ones
```

Now every duplicate disclosure is gated on the caller actually having access to that testcase. Adds a regression test covering a non-security-flagged duplicate in a project the caller has no access to.

### DIFF
--- a/src/appengine/handlers/crash_query.py
+++ b/src/appengine/handlers/crash_query.py
@@ -56,10 +56,13 @@ class Handler(base_handler.Handler):
 
     duplicate_testcase = data_handler.find_testcase(
         project, state.crash_type, state.crash_state, security_flag)
-    if (duplicate_testcase and duplicate_testcase.security_flag and
-        not access.can_user_access_testcase(duplicate_testcase)):
-      # Do not disclose the existence or identifiers of a security-confidential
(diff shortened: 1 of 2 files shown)


herdiyana256:
Gentle bump, open ~3 weeks with no reviewer. Small access-control fix: crash_query discloses a duplicate testcase's data without checking access to it. @PauloVLB @javanlacerda could one of you take a look or route it?

PR #5391, opened 2026-07-25
Require testcase access in issue_redirector before redirecting

herdiyana256:
`issue_redirector` takes a testcase id and 302-redirects the caller to that testcase's bug URL, with no access check at all:

```python
  def get(self, testcase_id):
    testcase = ...
    self.redirect(get_issue_url(testcase))
```

There was no authorization on the testcase, so any caller could hit `/issue/<testcase_id>` and be redirected to the linked bug for any testcase, including security-restricted ones, disclosing the bug id and the existence of the testcase.

```python
from libs import access
...
if not access.can_user_access_testcase(testcase):
```

The redirect is now gated on the same testcase-access check the rest of the app uses. Adds a regression test asserting a caller without access is denied instead of redirected.

### DIFF
--- a/src/appengine/handlers/issue_redirector.py
+++ b/src/appengine/handlers/issue_redirector.py
@@ -16,6 +16,7 @@
 
 from clusterfuzz._internal.issue_management import issue_tracker_utils
 from handlers import base_handler
+from libs import access
 from libs import helpers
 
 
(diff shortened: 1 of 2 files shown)


herdiyana256:
Gentle bump, open ~3 weeks with no reviewer. Small access-control fix: /issue/<testcase_id> redirects to a testcase's bug tracker URL with no access check, reachable unauthenticated. @PauloVLB @javanlacerda could one of you take a look or route it?

PR #5394, opened 2026-07-28
[Metrics] Add metadata polling for detecting preemption

PauloVLB:
b/538558303

### Problem
Preemptible VMs often get shut down by GCP before they can finish a fuzzing batch and report their metrics. Because the standard OS signals (`SIGTERM`) are often swallowed by the nested process hierarchy (PID 1 barriers in containers), the bot dies silently, leading to lost fuzzing hours and inaccurate metrics.

### Proposed Solution
Instead of relying solely on OS signals, we now poll the GCP Metadata server for preemption events directly from a background thread. This bypasses the signal propagation barrier.

- Added `_PreemptionPoller` in `monitor.py` to check the `instance/preempted` metadata key every 15 seconds (only on Preemptible VMs).
- Centralized metadata logic in `compute_metadata.py` via `get_preempted_status()`.
- Added a `threading.Lock` to `_TrackFuzzTime` in `fuzz_task.py` to ensure thread safety when reading/writing active trackers concurrently from the main thread and the poller thread.

### Validation
Validated in dev environment. Preemption events are now successfully detected.

Preemption being detected by polling metadata [logs](https://cloudlogging.app.goo.gl/tekHj9UGrUAGBPBx5)

image

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
@@ -20,6 +20,7 @@ import json
 import os
 import random
 import re
+import threading
 import time
 from typing import Any
 from typing import Dict
@@ -61,6 +62,7 @@ from clusterfuzz._internal.metrics import events
 from clusterfuzz._internal.metrics import fuzzer_logs
 from clusterfuzz._internal.metrics import fuzzer_stats
 from clusterfuzz._internal.metrics import logs
+from clusterfuzz._internal.metrics import monitor
 from clusterfuzz._internal.metrics import monitoring_metrics
 from clusterfuzz._internal.platforms import android
 from clusterfuzz._internal.protos import uworker_msg_pb2
@@ -406,6 +408,23 @@ def _should_create_testcase(group: uworker_msg_pb2.FuzzTaskCrashGroup,
 
 class _TrackFuzzTime:
   """Track the actual fuzzing time (e.g. excluding preparing binary)."""
+  _active_trackers = set()
+  _lock = threading.Lock()
+  _callback_registered = False
+
+  @classmethod
+  def get_active_trackers(cls):
+    with cls._lock:
+      return list(cls._active_trackers)
+
+  @classmethod
+  def _preemption_callback(cls):
+    """Callback to record wasted fuzzing time on preemption."""
+    logs.info('Running preemption callback to record wasted time.')
+    for tracker in cls.get_active_trackers():
+      duration = tracker.time.time() - tracker.start_time
+      logs.info(f'Recording {int(duration)} seconds of preempted time for '
+                f'{tracker.fuzzer_name} on {tracker.job_type}.')
 
   def __init__(self, fuzzer_name, job_type, time_module=time):
     self.fuzzer_name = fuzzer_name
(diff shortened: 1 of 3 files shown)


dylanjew (approved):

javanlacerda (approved):
LGTM. I just wondering if is there the possibility to have the preemption metadata updated and not receive the SIGTERM. Because it would make us to have false positive metrics.

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -406,6 +408,23 @@ def _should_create_testcase(group: uworker_msg_pb2.FuzzTaskCrashGroup,
...
+  def get_active_trackers(cls):
+    with cls._lock:
+      return list(cls._active_trackers)
+
+  @classmethod
+  def _preemption_callback(cls):
+    """Callback to record wasted fuzzing time on preemption."""
+    logs.info('Running preemption callback to record wasted time.')
+    for tracker in cls.get_active_trackers():
+      duration = tracker.time.time() - tracker.start_time
```
ViniciustCosta, line 425: I might not have understood how this works, is this duration being used anywhere besides logging?

dylanjew:
> LGTM. I just wondering if is there the possibility to have the preemption metadata updated and not receive the SIGTERM. Because it would make us to have false positive metrics.

I think the second PR adds some robust handling for this case!

src/clusterfuzz/_internal/metrics/monitor.py
```diff
@@ -181,6 +198,47 @@ def stop(self):
...
+      should_stop = self._poller_thread_stop_event.wait(
+          timeout=self._check_interval)
+      if should_stop:
+        break
+
+      try:
+        status = compute_metadata.get_preempted_status()
+        if status and status.strip().upper() == 'TRUE':
+          logs.info('Preemption detected via metadata!')
+          handle_sigterm(None, None)
```
ViniciustCosta: Not sure if we should call the sigterm handler ourselves, it should be used when actual sigterm signal is sent.

In this case, there is an assumption that the preempted metadata should be treated as a sigterm, which seems right, but maybe it is safer to just run the callback, flush the metric, and let the execution continue?

ViniciustCosta (approved):

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
PauloVLB: Yes, exactly. In this branch, the callback is just logging the duration for testing purposes. The actual emission of this metric happens in https://github.com/google/clusterfuzz/pull/5396

src/clusterfuzz/_internal/metrics/monitor.py
PauloVLB: Makes total sense. I refactored the code to extract the cleanup logic into `_trigger_shutdown_cleanup`. Now both the `SIGTERM` handler and the metadata poller call this directly, avoiding the artificial signal handling.

ViniciustCosta: Still wondering if we should call `stop()` and stop all monitoring when the poller gets the metadata or only flush the metric and let it be stopped by the machine being killed.

Not a blocker, feel free to merge :)

PR #5396, opened 2026-07-28
[Metrics] Add preemption metrics and accumulation logic

PauloVLB:
b/538558303

_Note: this PR is chained with https://github.com/google/clusterfuzz/pull/5394_

### Problem
When a preemption occurs, we were only recording the duration of the active fuzzing round as wasted time. If a task is preempted mid-way, all previous successful rounds in that same task are also completely wasted since the task never reaches postprocess. This leads to a significant underestimation of wasted time.

### Proposed Solution
Enhanced time tracking to capture the waste across the entire fuzzing session while maintaining metric accuracy.

- Added `FUZZER_PREEMPTED_TOTAL_FUZZ_TIME` and `JOB_PREEMPTED_TOTAL_FUZZ_TIME` metrics to `monitoring_metrics.py`.
- Enhanced `_TrackFuzzTime` in `fuzz_task.py` to track accumulated wasted time across all rounds of a session (`_accumulated_wasted_time`).
- Upon preemption, the callback now emits the Total Accumulated Session Time (Current Round + Previous Rounds) to `preempted_total_time`.
- To fix subtraction math in dashboards, the callback also emits the duration of the interrupted round only to `total_time`, ensuring it reflects the attempted time without gaps.
- Added `tracker.preempted` flag to skip metric emission in `__exit__` for interrupted rounds and avoid double counting.

### Validation
Validated in dev environment.

image

image

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
@@ -418,6 +418,23 @@ class _TrackFuzzTime:
   _active_trackers = set()
   _lock = threading.Lock()
   _callback_registered = False
+  _accumulated_wasted_time = 0
+  _session_fuzzer = None
+  _session_job = None
+
+  @classmethod
+  def reset_session(cls, fuzzer_name, job_type):
+    """Resets the accumulated time at the start of a session."""
+    with cls._lock:
+      cls._accumulated_wasted_time = 0
+      cls._session_fuzzer = fuzzer_name
+      cls._session_job = job_type
+
+  @classmethod
+  def add_wasted_time(cls, duration):
+    """Accumulates time from finished rounds."""
+    with cls._lock:
+      cls._accumulated_wasted_time += duration
 
   @classmethod
   def get_active_trackers(cls):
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/metrics/monitoring_metrics.py
```diff
@@ -158,6 +158,30 @@
     ],
 )
 
+FUZZER_PREEMPTED_TOTAL_FUZZ_TIME = monitor.CounterMetric(
```
dylanjew, line 161: This doesn't include build set up and corpus sync time, which is also effectively wasted time from a fuzzing perspective. But our total fuzz time metrics also don't include that, so I understand that we may want it to match.

I don't know how long that set up takes, but it might be worth tracking that as well.

dylanjew: Also, not blocking at all because any metrics here will help but definitely worth discussing/thinking about

PauloVLB: Yes, it was done that way to match the existing metric, so the subtraction (`total_time` - `preempted_time`) makes sense.

Measuring the duration of the full main stage (including build set up and corpus sync) is the next step I'm working on. Thanks for pointing that out!

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -411,27 +411,109 @@ class _TrackFuzzTime:
...
+          'timeout': timeout
+      }
+      job_labels = {**common_labels, 'job': job_type, 'timeout': timeout}
+
+    fuzzer_metric.increment_by(int(duration), fuzzer_labels)
+    job_metric.increment_by(int(duration), job_labels)
+
   @classmethod
   def _preemption_callback(cls):
     """Callback to record wasted fuzzing time on preemption."""
```
dylanjew: I would explain what this is doing more. It's not obvious that we need to track the `_accumulated_wasted_time` because if the task is preempted at any point, then the entire fuzzing session is lost since we don't process the `fuzz_task_output`, even though previous fuzzing rounds were "successfull" within the context of engine/blackbox fuzzing.

dylanjew (approved):

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
PauloVLB: Done. I have updated the docstring for the callback to explain this more clearly.

dylanjew (approved):

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
dylanjew: thanks!

javanlacerda (approved):
LGTM. Awesome job.

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -411,27 +411,116 @@ class _TrackFuzzTime:
   _active_trackers = set()
   _lock = threading.Lock()
   _callback_registered = False
+  _accumulated_wasted_time = 0
```
ViniciustCosta, line 421: Maybe we should call it only "_accumulated_session_time" as it is not wasted until it is preempeted?

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -411,27 +411,116 @@ class _TrackFuzzTime:
...
+  @classmethod
+  def reset_session(cls, fuzzer_name, job_type):
+    """Resets the accumulated time at the start of a session."""
+    with cls._lock:
+      cls._accumulated_wasted_time = 0
+      cls._session_fuzzer = fuzzer_name
+      cls._session_job = job_type
+
+  @classmethod
+  def add_wasted_time(cls, duration):
```
ViniciustCosta, line 434: Ditto. Maybe "add_round_time()"?

ViniciustCosta (approved):

PR #5407, opened 2026-07-31
Stop executing testcases when proto limit exceeded

IvanBM18:
Bug:b/473624472

Enforces protobuf message size limits (PROTOBUF_MSG_LIMIT) during blackbox testcase execution in fuzz_task.py.

## Context
In the previous PR in the stack, we changed so that in utask_main we now save the fuzzer logs in a protobuf message, which is consumed  at `postprocess`.

**Previously:**
For each test case in BlackBox fuzzing, we immediately uploaded its fuzz & output logs  into GCS. This proved basically no limitations.

**Now:**
For each test case in BlackBox fuzzing, we save its fuzz & output logs into a protobuf.

The protobuf has a hard limit of 2GB(see[ the docs](https://protobuf.dev/programming-guides/proto-limits/)).

## Overview
To prevent a possible edge case in which logs + output from a big fuzzing session surpasses the protobuf size limit, we now limit the test cases executed in the session once they trespass the protobuf limit size.

## Changes
* **Early Loop Termination:** Checks if `fuzz_task_output` size exceeds 1.8Gb after each batch of worker threads and terminates the execution loop early in case it does.
* **Output Limit Guard in Queue Processing:** Prevents further `fuzzer_run_output `entries from being appended once the 1.8Gb is exceeded, while still preserving all discovered crashes.
* **Clean Unexecuted Testcase Handling:** Deletes unexecuted testcase files from disk and slices returned testcase paths and metadata to ensure metrics (testcases_executed) report accurately.
* **Unit Tests:** Adds unit tests for a bunch of edge cases(theres actually more test cases than actual prod code).

### DIFF
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
@@ -1973,11 +1973,14 @@ class FuzzingSession:
       crashes = [
           Crash.from_testcase_manager_crash(crash) for crash in crashes if crash
       ]
-    executed_testcase_file_paths, executed_testcases_metadata = (
-        self._handle_unexecuted_testcases(
-            testcase_file_paths=testcase_file_paths,
-            testcases_metadata=testcases_metadata,
-            testcases_executed=test_number))
+    executed_testcase_file_paths = testcase_file_paths
+    executed_testcases_metadata = testcases_metadata
+    if test_number < len(testcase_file_paths):
+      executed_testcase_file_paths, executed_testcases_metadata = (
+          self._handle_unexecuted_testcases(
+              testcase_file_paths=testcase_file_paths,
+              testcases_metadata=testcases_metadata,
+              testcases_executed=test_number))
     return (generate_result.fuzzer_metadata, executed_testcase_file_paths,
             executed_testcases_metadata, crashes)
 
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
@@ -75,6 +75,9 @@ from clusterfuzz.fuzz import engine
 # 10 MB limit for proto testcase payload field (see FuzzerRunOutput testcase in
 # src/clusterfuzz/_internal/protos/uworker_msg.proto:L248).
 MAX_TESTCASE_PAYLOAD_SIZE = 10 * 1024 * 1024
+# PROTO msg limit is 2 GB, reserving 200 MB for safety
+# https://protobuf.dev/programming-guides/proto-limits/
+PROTOBUF_MSG_LIMIT = 1.8 * 1024 * 1024 * 1024
 FUZZER_METADATA_REGEX = re.compile(r'metadata::(\w+):\s*(.*)')
 FUZZER_FAILURE_THRESHOLD = 0.33
 MAX_NEW_CORPUS_FILES = 500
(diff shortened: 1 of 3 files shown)


src/clusterfuzz/_internal/bot/testcase_manager.py
```diff
@@ -586,7 +586,7 @@ def _do_run_testcase_and_return_result_in_queue(
 
     # Save raw, unsymbolized execution output for deferred postprocessing.
     crash_result_full = CrashResult(return_code, crash_time, output)
-    unsymbolized_output = crash_result_full.get_stacktrace(symbolized=False)
+    unsymbolized_output = crash_result_full.get_stacktrace()
```
IvanBM18, line 594: I added the `symbolized=False` for some tests on other PR, but i forgot to turn it back to the way it was: `crash_result_full.get_stacktrace()`
So i'm just adding it back here.

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
dylanjew, line 1857: here, this won't be accurate. Although, i think it wasn't always accurate to begin with if there was an error at some point too.

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
```diff
@@ -1928,8 +1973,13 @@ def do_blackbox_fuzzing(self, fuzzer, fuzzer_directory, job_type):
       crashes = [
           Crash.from_testcase_manager_crash(crash) for crash in crashes if crash
       ]
-    return (generate_result.fuzzer_metadata, testcase_file_paths,
-            testcases_metadata, crashes)
+    executed_testcase_file_paths, executed_testcases_metadata = (
```
dylanjew: it might be clearer to have this only clean up unexecuted testcases if `test_number < testcases_generated`

src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/fuzz_task_test.py
```diff
@@ -1552,6 +1590,134 @@ def test_add_crashes_from_temp_queue(self):
...
+        fuzz_task_input=uworker_msg_pb2.FuzzTaskInput())
+    return fuzz_task.FuzzingSession(uworker_input, 10)
+
+  def test_is_fuzz_task_output_limit_exceeded_below_limit(self):
+    """Verify _is_fuzz_task_output_limit_exceeded returns False below PROTOBUF_MSG_LIMIT."""
+    session = self._create_session()
+    self.assertFalse(session._is_fuzz_task_output_limit_exceeded())
+
+  @mock.patch(
+      'clusterfuzz._internal.bot.tasks.utasks.fuzz_task.PROTOBUF_MSG_LIMIT', 0)
```
dylanjew: nit: it's probably more realistic to set or mock `self.fuzz_task_output.ByteSize()` rather than just setting the `PROTOBUF_MSG_LIMIT` to 0

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
IvanBM18: Good point, thanks!

src/clusterfuzz/_internal/tests/core/bot/tasks/utasks/fuzz_task_test.py
IvanBM18: Done!

src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py
IvanBM18: You mean that we should update `self.fuzz_task_output` at the end of the fuzzing session so that it has the total of executed test cases instead?
Or WDYM by it being innacuarate?

dylanjew: Oh, sorry I was confused. I thought this was the number of testcases executed that we were reporting, and that it would be inaccurate because we are now stopping early.

We do that correctly here: https://github.com/google/clusterfuzz/blob/b26420660bf6da662227a5df1c30a4fd06044c12/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py#L2061

dylanjew, line 1834: this was what I got confused with. It's a known bug that we don't actually emit the correct testcase count when `generate_blackbox_testcases` doesn't actually generate the max amount of testcases for whatever reason.

dylanjew (approved):

javanlacerda (approved):

PR #5409, opened 2026-08-01
Clean up invalid FuzzerJob entities in cleanup task

javanlacerda:
fix b/538683766

### Problem
When `Job` or `Fuzzer` entities are deleted, orphaned `FuzzerJob` mappings can linger in Datastore because the cleanup cron task does not detect and prune them.

### Solution
- Added `cleanup_invalid_fuzzer_jobs()` in `src/clusterfuzz/_internal/cron/cleanup.py` to delete `FuzzerJob` entities referencing non-existent `Fuzzer` or `Job` entities.
- Called `cleanup_invalid_fuzzer_jobs()` in `cleanup.py`'s `main()`.
- Added unit tests in `src/clusterfuzz/_internal/tests/appengine/handlers/cron/cleanup_test.py`.

### Testing
```bash
python butler.py py_unittest -t appengine -p cleanup_test.py
python butler.py lint
```

### DIFF
--- a/src/clusterfuzz/_internal/cron/cleanup.py
+++ b/src/clusterfuzz/_internal/cron/cleanup.py
@@ -232,6 +232,30 @@ def cleanup_unused_fuzz_targets_and_jobs():
       f'{len(valid_target_jobs)} valid FuzzTargetJob entities remain.')
 
 
+def cleanup_invalid_fuzzer_jobs():
+  """Clean up FuzzerJob entities that reference non-existent Fuzzers or Jobs."""
+  valid_fuzzers = {
+      fuzzer.name
+      for fuzzer in ndb_utils.get_all_from_model(data_types.Fuzzer)
+      if fuzzer.name
+  }
+  valid_jobs = {
+      job.name
+      for job in ndb_utils.get_all_from_model(data_types.Job)
+      if job.name
+  }
+
+  to_delete = []
+  for fuzzer_job in ndb_utils.get_all_from_model(data_types.FuzzerJob):
+    if (fuzzer_job.fuzzer not in valid_fuzzers or
+        fuzzer_job.job not in valid_jobs):
+      to_delete.append(fuzzer_job.key)
+
+  if to_delete:
+    ndb_utils.delete_multi(to_delete)
+  logs.info(f'Deleted {len(to_delete)} invalid FuzzerJob entities.')
+
+
 def get_jobs_and_platforms_for_project():
   """Return a map of projects to jobs and platforms map to use for picking top
   crashes."""
@@ -1473,6 +1497,7 @@ def main():
   cleanup_reports_metadata()
   leak_blacklist.cleanup_global_blacklist()
   cleanup_unused_fuzz_targets_and_jobs()
+  cleanup_invalid_fuzzer_jobs()
   cleanup_unused_heartbeats()
   logs.info('Cleanup task finished successfully.')
   return True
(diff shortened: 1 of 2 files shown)


dylanjew (approved):

ViniciustCosta (commented):
It should be ok to add this in cleanup, but I think this deletion is already done in `fuzzer_selection.py` here: https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/fuzzing/fuzzer_selection.py#L32,

which should be called by the update/delete handlers (ex: https://github.com/google/clusterfuzz/blob/master/src/appengine/handlers/fuzzers.py#L303)

dylanjew:
> It should be ok to add this in cleanup, but I think this deletion is already done in `fuzzer_selection.py` here: https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/fuzzing/fuzzer_selection.py#L32,
> 
> which should be called by the update/delete handlers (ex: https://github.com/google/clusterfuzz/blob/master/src/appengine/handlers/fuzzers.py#L303)


I think there's still a possibility that we're left with dangling `FuzzerJob`s in the Datastore. If you remove a fuzzer from a job via the `/jobs` page, we only look for existing fuzzers with the `Job.name in Fuzzer.jobs` when we go to delete the associated FuzzerJobs [here](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/fuzzing/fuzzer_selection.py#L83)

This means that if there exists a `FuzzerJob` that doesn't have a corresponding entry in `Fuzzer.jobs`, we will never delete it. It's not supposed to happen, but it's possible there's a bug somewhere that introduces the invalid config.

We don't validate the Fuzzer actually exists before we schedule `FuzzerJobs`, so having this repair cron would allow us to recover from these scenarios.

dylanjew:
> > It should be ok to add this in cleanup, but I think this deletion is already done in `fuzzer_selection.py` here: https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/fuzzing/fuzzer_selection.py#L32,
> > which should be called by the update/delete handlers (ex: https://github.com/google/clusterfuzz/blob/master/src/appengine/handlers/fuzzers.py#L303)
> 
> I think there's still a possibility that we're left with dangling `FuzzerJob`s in the Datastore. If you remove a fuzzer from a job via the `/jobs` page, we only look for existing fuzzers with the `Job.name in Fuzzer.jobs` when we go to delete the associated FuzzerJobs [here](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/fuzzing/fuzzer_selection.py#L83)
> 
> This means that if there exists a `FuzzerJob` that doesn't have a corresponding entry in `Fuzzer.jobs`, we will never delete it. It's not supposed to happen, but it's possible there's a bug somewhere that introduces the invalid config.
> 
> We don't validate the Fuzzer actually exists before we schedule `FuzzerJobs`, so having this repair cron would allow us to recover from these scenarios.

I wonder if we should expand this repair cron to just keep the `Fuzzer.jobs` list in sync with the `FuzzerJobs`, rather than only deleting dangling/invalid FuzzerJobs?

ViniciustCosta:
> I wonder if we should expand this repair cron to just keep the `Fuzzer.jobs` list in sync with the `FuzzerJobs`, rather than only deleting dangling/invalid FuzzerJobs?

I agree that this might be a cleaner solution

ViniciustCosta (approved):

PR #5413, opened 2026-08-05
[Metrics] Check provision model by consulting metadata

PauloVLB:
b/538558303

### Problem
Batch VMs don't have the `PREEMPTIBLE` env var set, so the preemption poller never starts on them. This means we lose track of wasted fuzzing time when they get preempted.

### Proposed Solution
Consult the GCE Metadata Server directly to check if the machine is preemptible or a Spot VM. 
Added `is_preemptible()` to `compute_metadata.py` to check both `instance/scheduling/preemptible` and `instance/scheduling/provisioning-model`, falling back to the env var for local testing. Updated `monitor.py` to use this helper.

### Validation
Verified that existing tests pass and the logic correctly falls back to env var when not on GCE.

We can now view preemption detection for batch VMs in dev:

image

[GCP Logs](https://cloudlogging.app.goo.gl/9TSJ69A7ThdL58kR6)

### DIFF
--- a/src/clusterfuzz/_internal/google_cloud_utils/compute_metadata.py
+++ b/src/clusterfuzz/_internal/google_cloud_utils/compute_metadata.py
@@ -64,3 +64,24 @@ def get_preempted_status():
   response = requests.get(attribute_url, headers=headers, timeout=5)
   response.raise_for_status()
   return response.text
+
+
+def is_preemptible():
+  """Returns True if the instance is preemptible (or Spot)."""
+  if is_gce():
+    # We ignore exceptions (mainly HTTPError if a key doesn't exist) to avoid
+    # log noise for standard VMs where these keys might not be present.
+    try:
+      if get('instance/scheduling/preemptible').strip().upper() == 'TRUE':
+        return True
+    except Exception:
+      pass
+
+    try:
+      if get(
+          'instance/scheduling/provisioning-model').strip().upper() == 'SPOT':
+        return True
+    except Exception:
+      pass
+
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/google_cloud_utils/compute_metadata.py
```diff
@@ -64,3 +64,22 @@ def get_preempted_status():
...
   return response.text
+
+
+def is_preemptible():
+  """Returns True if the instance is preemptible (or Spot)."""
+  if is_gce():
+    try:
+      if get('instance/scheduling/preemptible').strip().upper() == 'TRUE':
+        return True
+    except Exception:
```
dylanjew, line 77: should we log these exceptions? what types of `Exception` do you expect here?

dylanjew (approved):

src/clusterfuzz/_internal/google_cloud_utils/compute_metadata.py
PauloVLB: The expected exceptions here are mainly `HTTPError` if one of the metadata keys doesn't exist on that specific VM version (e.g. provisioning-model on older legacy VMs).

I'm not logging them to avoid log noise for standard VMs where these keys might not be relevant or present.

dylanjew: Thanks! I would add a comment explaining that :)

javanlacerda (approved):

PR #5414, opened 2026-08-06
[Metrics] Refactor metadata consumption and avoid unecessary calls

PauloVLB:
### Problem
The `is_preemptible()` check was causing high latency and log noise in some environments. This happened because it used a shared metadata helper that retries on failure, meaning missing keys (404s) caused multi-second delays. Additionally, the code in `compute_metadata.py` was repetitive.

### Proposed Solution
- Refactored `compute_metadata.py` to extract a fast, non-retrying internal helper `_get_raw()`.
- Optimized `is_preemptible()` to skip metadata queries entirely when running on App Engine or Kubernetes, avoiding unnecessary 404s.
- Ensured heuristics use short timeouts and fail fast without retrying on expected 404s.

### Validation

The not found error disappeared after the deploy:

image

[Link for GCP](https://cloudlogging.app.goo.gl/sksprTGgExq6mXzs6)

### DIFF
--- a/src/clusterfuzz/_internal/google_cloud_utils/compute_metadata.py
+++ b/src/clusterfuzz/_internal/google_cloud_utils/compute_metadata.py
@@ -21,9 +21,22 @@ from clusterfuzz._internal.base import retry
 from clusterfuzz._internal.system import environment
 
 _METADATA_SERVER = 'metadata.google.internal'
+_METADATA_URL = 'http://{}/computeMetadata/v1/'.format(_METADATA_SERVER)
 
 _RETRIES = 3
 _DELAY = 1
+
+
+def _get_raw(path, timeout=None):
+  """Internal helper to get metadata without retries."""
+  attribute_url = _METADATA_URL + path
+  headers = {'Metadata-Flavor': 'Google'}
+  if timeout is None:
+    timeout = environment.get_value('URL_BLOCKING_OPERATIONS_TIMEOUT')
+
+  response = requests.get(attribute_url, headers=headers, timeout=timeout)
+  response.raise_for_status()
+  return response.text
 
 
 @retry.wrap(
@@ -32,15 +45,7 @@ _DELAY = 1
     function='python.google_cloud_utils.compute_metadata.get')
 def get(path):
   """Get GCE metadata value."""
-  attribute_url = (
-      'http://{}/computeMetadata/v1/'.format(_METADATA_SERVER) + path)
-  headers = {'Metadata-Flavor': 'Google'}
-  operations_timeout = environment.get_value('URL_BLOCKING_OPERATIONS_TIMEOUT')
-
-  response = requests.get(
-      attribute_url, headers=headers, timeout=operations_timeout)
-  response.raise_for_status()
-  return response.text
+  return _get_raw(path)
 
 
 def is_gce():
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/google_cloud_utils/compute_metadata.py
```diff
@@ -26,21 +26,26 @@
 _DELAY = 1
 
 
+def _get_raw(path, timeout=None):
+  """Internal helper to get metadata without retries."""
+  attribute_url = 'http://{}/computeMetadata/v1/{}'.format(
```
decoNR: nit: Maybe make `http://{_METADATA_SERVER}/computeMetadata/v1/` a global variable given that it is static.

src/clusterfuzz/_internal/google_cloud_utils/compute_metadata.py
```diff
@@ -56,32 +61,32 @@ def is_gce():
...
   # in a background loop and should fail fast.
-  response = requests.get(attribute_url, headers=headers, timeout=5)
-  response.raise_for_status()
-  return response.text
+  return _get_raw('instance/preempted', timeout=5)
 
 
 def is_preemptible():
   """Returns True if the instance is preemptible (or Spot)."""
+  # Skip metadata queries on App Engine or K8s to avoid errors and latency.
```
decoNR: nit: I would add a small explanation on why it would cause errors (and which errors) just to make it clear. Maybe this is obvious and not needed. Feel free to decide to add it or not.

src/clusterfuzz/_internal/google_cloud_utils/compute_metadata.py
```diff
@@ -56,32 +61,32 @@ def is_gce():
...
-        return True
-    except Exception:
-      pass
+    for path in [
+        'instance/scheduling/preemptible',
+        'instance/scheduling/provisioning-model'
+    ]:
+      try:
+        value = _get_raw(path, timeout=2).strip().upper()
+        if value in ['TRUE', 'SPOT']:
```
decoNR, line 86: I do not think this will cause any errors, but just to flag it, this change slightly modifies the behavior. If the first request returns 'SPOT', it will now return true, whereas previously it would not. I assume this is not a possible scenario, correct?

decoNR (approved):

src/clusterfuzz/_internal/google_cloud_utils/compute_metadata.py
PauloVLB: Correct. According to the [GCP docs](https://docs.cloud.google.com/compute/docs/metadata/predefined-metadata-keys), the `/preemptible` key returns a boolean:

image

Additionally, `/provisioning-model` returns either `SPOT` or `STANDARD`. I had difficulty locating the documentation for this since it is a deprecated endpoint, though we still have some VMs using it in Clusterfuzz.

src/clusterfuzz/_internal/google_cloud_utils/compute_metadata.py
PauloVLB: Sure! Added some more explanation.

src/clusterfuzz/_internal/google_cloud_utils/compute_metadata.py
PauloVLB: Done.

src/clusterfuzz/_internal/google_cloud_utils/compute_metadata.py
decoNR: Nice, thanks for validating it!

PR #5415, opened 2026-08-06
[Protos][Swarming] Update swarming new task reponse proto

IvanBM18:
## Overview
Yesterday the `utask-main-schedueler` started to fail because of a wrong proto parsing, after consulting the proto definition we noticed that we were missing a proto, so basically this PR is full of proto generated changes from adding the TaskRequestMetadataResponse proto & its dependent proto definitions.

## Changes
- Updates the `swarming.proto` to include the missing protos
- Runs `generate_protos.sh` which updated a bunch of .py files
- The SwarmingAPI will still continue to return the same TaskRequestResponse proto, but we needed to perform a small change there, and in a unit test consequentially

## `dev` tests
Here are [some logs](https://cloudlogging.app.goo.gl/L8MVmbeDbYaSKV9r6) of when the error was happening in `dev`, & here are [logs of the `dev` env in the last 12hrs](https://cloudlogging.app.goo.gl/kMCecXuAvsL1QvJ76) note how we are now scheduling swarming tasks back again :D

Also heres a swarming task scheduled automatically today: https://chrome-swarming.appspot.com/task?id=79f17f5fdcc6cf10

(Is expected that this task fails at 3 min because the current job definition for linux has no release bucket on purpose)


javanlacerda (approved):

dylanjew (approved):
It would be nice if this was caught in a presubmit check or done automatically :(

PR #5422, opened 2026-08-11
[Butler] Make local App Engine instance accessible from local network

JuanMBriones:
When developing locally or interacting with the ClusterFuzz web server from other devices on the local network (such as another workstation or test devices), the development App Engine server was bound only to `127.0.0.1`, preventing external access on the LAN.

- Added `DEV_APPSERVER_ADDR = '0.0.0.0'` in `constants.py` and updated `DEV_APPSERVER_HOST` to use this address.
- Updated `run_server.py` to bind gunicorn to `DEV_APPSERVER_HOST` instead of hardcoded `127.0.0.1`.

- Ran `python butler.py lint` (passed).
- Ran `python butler.py integration_tests` (passed).
- Ran `python butler.py py_unittest -t core -m` (passed).
- Ran `python butler.py py_unittest -t appengine -m` (passed).


dylanjew (approved):
did this work with the developer containers? 

https://github.com/google/clusterfuzz/blob/6b713cf90a3504571ef0b8dc0597dcf8158d1470/docs/getting-started/local_instance.md?plain=1#L24

JuanMBriones:
> did this work with the developer containers?
> 
> https://github.com/google/clusterfuzz/blob/6b713cf90a3504571ef0b8dc0597dcf8158d1470/docs/getting-started/local_instance.md?plain=1#L24

Hiii Dylan, thanks for the review!!!:D

When working with devcontainers VSCode does the binding automatically (it makes available the running instance on the local network as well). That's why this is only focused when running the service through butler.

ViniciustCosta (approved):

JuanMBriones:
Thanks for the review guys! :D

PR #5424, opened 2026-08-13
[Typing] Add Python type annotations across codebase (Part 1: Mechanical)

letitz:
## Overview

This is **Part 1** of the static typing rollout for ClusterFuzz. It introduces comprehensive Python type annotations across all files in `src/appengine` and `src/clusterfuzz` with **zero runtime behavioral changes**.

Part 2 (stacked on top of this PR) contains targeted runtime fixes and defensive checks identified during the typing process.

## Summary of Changes

- **Codebase-Wide Type Annotations**: Added type annotations across 313 Python modules in `src/appengine`, `src/clusterfuzz`, and `src/local`.
- **NDB Type Stubs**: Implemented comprehensive type stub definitions for `google-cloud-ndb` in `typings/google/cloud/ndb/`, including:
  - Strongly typed `Model`, `Expando`, `Key`, and `Query` interfaces.
  - Keyword-only `repeated=True` overloads across property descriptors (`StringProperty`, `IntegerProperty`, `LocalStructuredProperty`, etc.) ensuring repeated properties resolve directly to `list[T]`.
  - Descriptor `__get__` overloads supporting instance vs class-level attribute access and Datastore comparison operators (`<=`, `>=`, `==`, `.IN()`).
- **Configuration & Strict Dual Type Checking**: Enabled type checking in `pyproject.toml` and `pyrightconfig.json` covering all codebase modules.
- **Obsolete Cast Cleanup**: Removed legacy `cast(Any, ...)` calls on NDB queries and model entities that predated custom stubs.

## Verification & Safety

- **Type Checkers**:
  - `uv run ty check`: **0 errors** (100% clean across all modules)
  - `uv run butler.py lint --type-check`: **0 errors, 0 warnings**
- **Unit Test Suite**:
  - `python butler.py py_unittest -t core -m`: **2,221 tests passed (0 errors, 0 failures)**
  - `python butler.py py_unittest -t appengine -m`: **742 tests passed (0 errors, 0 failures)**
- **Runtime Invariance**: All functional code logic, assertions, and execution paths remain strictly identical to previous runtime behavior (with type ignores/casts used where runtime changes are deferred to Part 2).

PR #5426, opened 2026-08-14
Require job access in coverage_report before redirecting

herdiyana256:
`coverage_report` resolves and redirects to the coverage report URL for a job taken from the request, decorated only with `@handler.oauth`:

```python
def get_report_url(report_type, argument, date):
  ...
  job = argument
  return _get_project_report_url(job, date)   # no access check
```

Coverage reports are per-job data, but `@handler.oauth` authenticates the caller without authorizing access to any particular job (the same gap the other job-scoped handlers like `fuzzer_stats` already close). So any caller could resolve the coverage report location for an arbitrary job, and by extension its project.

```python
from libs import access
...
if not access.has_access(job_type=job):
```

The resolution is now gated on job access, matching the sibling job-scoped handlers. Adds a regression test for the access-denied path.

### DIFF
--- a/src/appengine/handlers/coverage_report.py
+++ b/src/appengine/handlers/coverage_report.py
@@ -20,6 +20,7 @@ from clusterfuzz._internal.datastore import data_handler
 from clusterfuzz._internal.datastore import data_types
 from clusterfuzz._internal.metrics import fuzzer_stats
 from handlers import base_handler
+from libs import access
 from libs import handler
 from libs import helpers
 
@@ -59,6 +60,13 @@ def get_report_url(report_type, argument, date):
 
   if not data_types.Job.VALID_NAME_REGEX.match(job):
     raise helpers.EarlyExitError('Invalid job name.', 400)
+
+  # Coverage reports are per-job data; gate on job access like the other
+  # job-scoped handlers (e.g. fuzzer_stats). @handler.oauth does not require
+  # authentication, so without this any caller could resolve the report URL
+  # for an arbitrary job.
+  if not access.has_access(job_type=job):
+    raise helpers.AccessDeniedError()
 
   if not date or not VALID_DATE_REGEX.match(date):
     raise helpers.EarlyExitError('Invalid date.', 400)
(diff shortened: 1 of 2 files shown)


herdiyana256:
Friendly bump on this one, and a question about CI.

### What the change does

`coverage_report.py`'s `get_report_url` resolved a job's coverage report URL and redirected to it based on a caller-supplied `job` name, with no authorization check. The only decorator was `@handler.oauth`, which does not gate: the wrapper records `_oauth_email` when an `Authorization` header is present and calls the handler either way, so the route was reachable unauthenticated. The `job` value was validated for name format only.

The patch adds `access.has_access(job_type=job)` after the existing name validation and raises `AccessDeniedError` when it fails, plus a `test_no_access` case.

### Why this shape is the right one

The parity argument is inside this repo. `fuzzer_stats.py:456` gates the same category of per-job data with `access.has_access(fuzzer_name=fuzzer, job_type=job)`, so coverage data is already treated as job-scoped elsewhere. This handler was the outlier.

On public OSS-Fuzz the coverage reports are public anyway, so the impact case is private deployments where jobs and projects are access controlled. There, an unauthenticated caller could enumerate job names and learn the coverage report location and project mapping for jobs they cannot otherwise see.

### CI question

Two Cloud Build checks report failure here, `All-Tests (clusterfuzz-testing-public)` and `Build-Base-Images-PR (clusterfuzz-images)`. Both report a duration of 0, which reads like they never actually executed rather than failing on the diff. Their logs live in the Cloud Build console, which I cannot see from outside, so I do not want to guess at the cause.

For what it is worth, the pattern across currently open PRs is mixed rather than uniformly fork related: #5424 and #5425 are also from forks and their `All-Tests` passes, while #5426 and #5429 fail. If these need a maintainer to kick off, please do, and if the logs show something genuinely caused by this diff I will fix it right away.

The change touches two files, `coverage_report.py` (+8) and `coverage_report_test.py` (+13), and the branch still merges cleanly.

### One note for whoever picks this up

#5389, #5390 and #5391 are the same class of missing access check on other handlers, and all four have been sitting without a review since 14 August. If it is easier to look at them as a batch, or if you would rather I consolidate or close and resubmit them differently, just say which you prefer and I will do that.

PR #5433, opened 2026-08-21
[Android] Remove legacy Android Build API v3

IvanBM18:
Bug: b/542587003, b/422775458

We previously migrated the android build api to call endpoints in v4, while keeping v3 open for instances of clusterfuzz that still used v3.  But now that the V3 has 100% throttle, its makes no difference, calls to v3 will fail, and calls to v4 will fail if they have not valid credentials.

So this PR removes all of the stale v3 api code, both from unit tests & its consumers.

### DIFF
--- a/src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
+++ b/src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
@@ -48,6 +48,8 @@ STABLE_CUTTLEFISH_BUILD = {
 DEFAULT_STABLE_CUTTLEFISH_BUILD_INFO = (
     "gs://android-haiku/target-cuttlefish/stable_build_info.json")
 
+API_VERSION = 'V4'
+
 
 def _call_android_api_enabled():
   """Return True if we should call the Android Build API, enabled by default,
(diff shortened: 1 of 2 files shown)


dylanjew (commented):
Did the android team finish the migration yet?

dylanjew (approved):
otherwise LGTM

IvanBM18:
> otherwise LGTM

Yes! they were missing CL approval, but it just got merged see internal: cl/969979278

src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
```diff
@@ -104,28 +69,22 @@ def _download_artifact(client, bid, target, attempt_id, name, output_directory,
   logs.info('output_directory: %s' % output_directory)
   logs.info('output_filename: %s' % output_filename)
 
-  version_tag = 'V4' if _use_v4() else 'V3'
   logs.info(
       'AndroidBuildAPI download_artifact started.',
-      api_version=version_tag,
+      api_version='V4',
```
ViniciustCosta: nit: If all logs are set to the same value, maybe we can remove this field? Or at least put it into a top-level var (`_API_VERSION`)?

ViniciustCosta (approved):

src/clusterfuzz/_internal/platforms/android/fetch_artifact.py
IvanBM18: Done!

PR #5434, opened 2026-08-24
[Android Bad Build] Add liveness check

PauloVLB:
b/546076952

### Problem
On Android, `check_for_bad_build` launches the APK via `am start`, which exits with return code 0 as soon as the launch intent is delivered regardless of whether the app process survives. If the APK crashes or dies immediately on startup without generating a recognized ASan/security crash pattern in logcat, `crash_result.is_crash()` returns `False`. ClusterFuzz treats the build as healthy, falsely recording fuzzing hours for dead builds.

### Proposed Solution
Add an Android process liveness check in `check_for_bad_build`. After launching the app, verify if the package process is actually running via `android.adb.get_process_and_child_pids(package_name)`. If no active PIDs exist for the package, flag `is_bad_build = True` to halt execution before recording fuzzing hours.

### Testing & Validation
- Added unit tests in `testcase_manager_test.py` covering both dead APK and running APK scenarios.
- Verified in `dev` using the `aluminium_asan_chrome_public` job on `ANDROID_EMULATOR`, confirming that invalid builds are now caught and stopped at `check_for_bad_build`. 
- (See screenshot and logs link in b/546076952#comment7)

### DIFF
--- a/src/clusterfuzz/_internal/bot/testcase_manager.py
+++ b/src/clusterfuzz/_internal/bot/testcase_manager.py
@@ -1296,6 +1296,19 @@ def check_for_bad_build(job_type: str,
         f'return code = {return_code}, crash type = {crash_result.get_type()}',
         raw_output=output,
         output=build_run_console_output)
+  elif environment.is_android():
+    package_name = android.app.get_package_name()
+    if (package_name and
+        not android.adb.get_process_and_child_pids(package_name)):
+      is_bad_build = True
+      build_run_console_output = utils.get_crash_stacktrace_output(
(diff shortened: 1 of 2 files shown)


IvanBM18 (approved):
Thanks for looking into this, LGTM!

ViniciustCosta (approved):
Lgtm! Let's try to also get a review from Haiku folks for google3 Android? Or is this only executed for Chrome?

PauloVLB:
We had to merge this for the deployment window. Please feel free to leave comments so we can address them in a follow-up PR.

PR #5437, opened 2026-08-26
Reject tar hard link targets that escape the output directory

### LINKED ISSUE #5436: system/archive.py: tar hard link linknames with '..' can overwrite files outside the extraction directory
## Summary

`TarArchiveReader.extract` (`src/clusterfuzz/_internal/system/archive.py`) validates archive member **names** against path traversal (`_is_attempting_path_traversal`), but never validates the **linkname** of hard link members. A tar containing a hard link whose relative `linkname` includes `..` causes CPython's `tarfile` to `os.link()` an existing file **outside** the extraction directory into it; a subsequent write through that alias (e.g. a regular member that resolves to the same path) overwrites the outside file with attacker-controlled content.

The existing realpath containment check cannot detect this: `os.path.realpath` does not resolve hard links, so the aliased path still appears to be inside the output directory.

## Reproduction (stdlib only)

```python
import io, os, tarfile

os.makedirs("/tmp/poc/out", exist_ok=True)
open("/tmp/poc/victim-file", "w").write("ORIGINAL")

with tarfile.open("/tmp/poc/evil.tar", "w") as tf:
    link = tarfile.TarInfo("innocent.txt")   # clean member name
    link.type = tarfile.LNKTYPE
    link.linkname = "../victim-file"          # relative .. escapes the output dir
    tf.addfile(link)

    data = b"PWNED"
    payload = tarfile.TarInfo("innocent.txt") # same clean name
    payload.size = len(data)
    tf.addfile(payload, io.BytesIO(data))

with tarfile.open("/tmp/poc/evil.tar") as tf:
    for m in tf.getmembers():
        # _is_attempting_path_traversal(archive, out, m.name) -> False for both
        tf.extract(m, "/tmp/poc/out")   # no filter=, as in archive.py

print(open("/tmp/poc/victim-file").read())     # -> PWNED (outside the output dir)
```

Notes:
- An **absolute** linkname does not work (CPython strips the leading `/` and joins inside the output dir); the escape requires a **relative** linkname with `..` of the correct depth.
- The member names involved are clean, so the name-based guard passes; only the linkname escapes.
- Reachable through the untrusted archive upload/unpack flow (`TarArchiveReader.extract` at `archive.py`, looped by `extract_all` from the unpack task).

## Suggested fix

Apply the same containment check to the resolved hard-link target that `tarfile` itself will use (mirroring its absolute-target re-rooting). Implemented in the attached PR, along with regression tests:
- `test_tar_hardlink_traversal` — escaping relative linkname is rejected, victim file untouched;
- `test_tar_hardlink_within_directory` — legitimate in-archive hard links still extract (archive built with real on-disk hard links so the linkname is in canonical member-rooted form).

An alternative is passing `filter="data"` to `tarfile.extract` on Python ≥ 3.12; the guard approach in the PR is version-independent and mirrors the existing check style.

This issue was reported to the Google Bug Hunters team, who classified it as valid but below their security-tracking threshold and explicitly permitted public disclosure.

shaggyinsomniac:
Fixes #5436.

## What

- `TarArchiveReader.extract` now resolves a hard link member's target the same way `tarfile` will (absolute linknames re-rooted at the output directory, relative ones joined against the member's directory) and applies the existing `_is_attempting_path_traversal` containment check to it.
- Adds two regression tests: an escaping relative linkname is rejected with the victim file untouched, and legitimate in-archive hard links (canonical member-rooted linknames, built from real on-disk hard links) still extract.

## Why

The member-name guard could not catch this class: `os.path.realpath` does not resolve hard links, so an alias created by `os.link` pointing outside the output directory still appears contained. Details and a standalone stdlib PoC in #5436.

## Testing

`pytest src/clusterfuzz/_internal/tests/core/system/archive_test.py` — both new tests pass; no other tests changed behavior (one unrelated pre-existing env-sensitive zip-chmod mock test fails identically with and without this change).

### DIFF
--- a/src/clusterfuzz/_internal/system/archive.py
+++ b/src/clusterfuzz/_internal/system/archive.py
@@ -299,6 +299,24 @@ class TarArchiveReader(ArchiveReader):
                                         member)):
         return None
 
+      tarinfo = self._archive.getmember(member)
+      if tarinfo.islnk():
+        # Mirror how tarfile resolves hard link targets (absolute targets are
+        # re-rooted at the output directory), then apply the same containment
+        # check as for member names. Without this, a relative linkname
+        # containing '..' makes tarfile's os.link() alias a file outside the
+        # output directory, and the realpath-based check cannot detect it
+        # because it does not resolve hard links.
+        linkname = tarinfo.linkname
+        if linkname.startswith('/'):
+          linkname = linkname[1:]
+        link_target = os.path.join(
+            output_directory, os.path.dirname(member), linkname)
+        if _is_attempting_path_traversal(
+            self._archive_path, output_directory,
+            os.path.relpath(link_target, output_directory)):
+          return None
+
     self._archive.extract(member=member, path=output_directory)
     return os.path.realpath(os.path.join(output_directory, member))
 
--- a/src/clusterfuzz/_internal/tests/core/system/archive_test.py
+++ b/src/clusterfuzz/_internal/tests/core/system/archive_test.py
@@ -209,6 +209,59 @@ class ArchiveReaderTest(unittest.TestCase):
 
       self.assertEqual(expected_results, actual_results)
 
+  def test_tar_hardlink_traversal(self):
+    """Hard link linknames with '..' must not escape the output directory."""
+    with tempfile.TemporaryDirectory() as outer:
+      victim_path = os.path.join(outer, 'victim-file')
+      with open(victim_path, 'wb') as f:
+        f.write(b'original')
+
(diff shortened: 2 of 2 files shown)


shaggyinsomniac:
Both new tests pass locally on Python 3.10:
- `test_tar_hardlink_traversal` — verifies escaping linkname is blocked
- `test_tar_hardlink_within_directory` — verifies legitimate hard links still extract

The `All-Tests` and `Build-Base-Images` CI failures may be environment-specific (CI uses Python 3.11 in Docker). Could you share the failing test output from the Cloud Build logs so I can investigate?

PR #5438, opened 2026-08-26
Harden DEPS parsing and sanitize revision metadata handling

vitaliset:
- Replace dynamic evaluation in DEPS parsing with a restricted AST-based parser.
- Add revision string format validation and template-based URL formatting.
- Sanitize issue metadata across uworker post-processing tasks before persistence.
- Escape revision identifiers in testcase detail fallback rendering.

b/534004999

TAG=agy
CONV=6fa9c8c8-6088-4c02-b27d-ab873dda019f


vitaliset:
### Verification
- **Development Deployment**:
  - [Deployed](https://chat.google.com/room/AAQAy8bE1NA/nx1dZnRH0zI/cY_IKeLF58c?cls=10) to the development environment and validated against live App Engine traffic and active worker bot fleets.
  - Monitored real-time [telemetry](http://goto.google.com/cf-chrome-dev-health) and error logs across fuzzing workloads with zero parsing exceptions or regressions observed.

### Compatibility & Risk Assessment
- **Safe DEPS Parsing**: Fully supports all standard DEPS structures (string assignments, `Var()` lookups, lists, dictionaries, tuples) without altering expected dependency resolution behavior.
- **Revision Validation**: The validation pattern (`^[a-zA-Z0-9._-]+$`) covers all standard git commit hashes, SVN numbers, semantic versions, and release tags.
- **Metadata Sanitization & Escaping**: Operates purely as defensive validation on untrusted inputs without affecting valid pipeline data.

src/clusterfuzz/_internal/build_management/revisions.py
```diff
@@ -359,10 +478,14 @@ def _get_revision_vars_url_format(job_type, platform_id=None):
 @memoize.wrap(memoize.Memcache(60 * 60 * 24 * 30))  # 30 day TTL
 def get_component_revisions_dict(revision, job_type, platform_id=None):
   """Retrieve revision vars dict."""
-  if revision == 0 or revision == '0' or revision is None:
+  if revision in (0, '0', None, ''):
```
ViniciustCosta: It seems you are adding the empty string as a zero start revision, is this expected?

src/clusterfuzz/_internal/build_management/revisions.py
```diff
@@ -270,32 +277,144 @@ def _to_dict(contents):
...
-  exec(content, global_context, local_context)
+def is_valid_revision(revision: str | int | None) -> bool:
+  """Returns True if the revision is a safe and valid identifier.
 
-  revisions_dict = {}
+  Valid revisions contain only alphanumeric characters, dots, underscores and
+  hyphens. Rejects path traversal, query strings and other characters that
+  would let a poisoned revision reshape the url it is interpolated into.
+  """
+  if revision in (0, '0', None, ''):
```
ViniciustCosta: Not sure why these need to be an early return False.

Also considering that this function is applied to a revision after it already passed this check earlier.

src/clusterfuzz/_internal/datastore/data_handler.py
```diff
@@ -59,6 +59,20 @@
 NUM_TESTCASE_QUALITY_BITS = 3
 MAX_TESTCASE_QUALITY = 2**NUM_TESTCASE_QUALITY_BITS - 1
 
+# Testcase metadata keys that must hold an integer revision. These end up
+# interpolated into the admin-configured REVISION_VARS_URL template (see
+# revisions.get_component_revisions_dict) and rendered on the testcase detail
+# page, so untrusted workers must never be able to store arbitrary strings in
+# them.
+NUMERIC_METADATA_KEYS = frozenset({
```
ViniciustCosta, line 67: nit: my only concern is that if we add new numeric metadata keys, likely we won't add it here. I don't think there is a way to improve this though

ViniciustCosta (approved):
The sanitize DEPS logic is difficult for me to review, as I don't have context about this. Seems good to me given your tests on dev :)

src/clusterfuzz/_internal/build_management/revisions.py
vitaliset: Removed '' from the zero-start check in `get_component_revisions_dict` and restored `(0, '0', None)`. Empty strings will be caught and rejected by `is_valid_revision` directly.

src/clusterfuzz/_internal/build_management/revisions.py
vitaliset: Simplified `is_valid_revision` to focus strictly on format validation against `REVISION_PATTERN` (returning False for None or invalid characters/empty strings), without special-casing 0 or '0'.

src/clusterfuzz/_internal/datastore/data_handler.py
vitaliset: Having this centralized set ensures at least these known sensitive revision fields that get interpolated into URLs are strictly validated, and the docstring explains the rationale for future maintainers.

PR #5442, opened 2026-08-28
[Mac] Add support for butler.py to generate Mac ARM64 deployment zip

JuanMBriones:
[Mac] Add support for macOS ARM64 deployment packages and bot execution
### Problem
ClusterFuzz lacked support for generating macOS ARM64 deployment zip
packages and running bots on Apple Silicon. Additionally, packaging
failed due to mismatched macOS wheel SDK tags in pip download, and
is_supported_cpu_arch_for_job incorrectly parsed string CPU_ARCH values.

### Changes
* butler.py & constants.py: Add macos_arm64 platform target, fallback
  SDK tags, and Python 3.7–3.11 ABIs.
* common.py: Download platform pip requirements individually to resolve
  differing wheel tags, and support ARM64 ChromeDriver.
* platform_requirements.txt: Add native C-extension packages (bcrypt,
  cffi, cryptography, google-crc32c, PyYAML, wrapt).
* environment.py: Add get_cpu_arch() and support arch-specific tool paths
  in get_default_tool_path().
* update_task.py: Map macOS ARM64 bots to macos_arm64 deployment packages.
* commands.py: Fix CPU_ARCH parsing in is_supported_cpu_arch_for_job().
* Tests: Add unit tests for environment arch detection, tool path resolution,
  deployment URLs, and job arch validation.

### DIFF
--- a/butler.py
+++ b/butler.py
@@ -33,6 +33,8 @@ from local.butler import guard
 
 guard.check()
 
+from local.butler import constants
+
 
 class _ArgumentParser(argparse.ArgumentParser):
   """Custom ArgumentParser."""
@@ -125,7 +127,10 @@ def _add_package_subparser(toplevel_subparsers):
   parser_package = toplevel_subparsers.add_parser(
       'package', help='Package clusterfuzz with a staging revision')
   parser_package.add_argument(
-      '-p', '--platform', choices=['linux', 'macos', 'windows', 'all'])
+      '-p',
+      '--platform',
+      choices=list(constants.PLATFORMS.keys()) + ['all'],
+      default='all')
   parser_package.add_argument(
       '-r',
       '--release',
--- a/butler.py
+++ b/butler.py
@@ -129,8 +129,7 @@ def _add_package_subparser(toplevel_subparsers):
   parser_package.add_argument(
       '-p',
       '--platform',
-      choices=list(constants.PLATFORMS.keys()) + ['all'],
-      default='all')
+      choices=list(constants.DEPLOYMENT_TARGETS.keys()) + ['all'])
   parser_package.add_argument(
       '-r',
       '--release',
--- a/src/clusterfuzz/_internal/base/utils.py
+++ b/src/clusterfuzz/_internal/base/utils.py
@@ -315,12 +315,10 @@ def _get_deployment_zip_release_suffix(release):
   return suffix
 
 
-def get_platform_deployment_filename(platform, release):
-  """Return the platform deployment filename."""
-  # Expects linux, macos or windows.
-  base_filename = platform
+def get_deployment_target_filename(deployment_target, release):
+  """Return the deployment filename for the given target."""
   release_filename_suffix = _get_deployment_zip_release_suffix(release)
-  return f'{base_filename}{release_filename_suffix}.zip'
+  return f'{deployment_target}{release_filename_suffix}.zip'
 
 
 def get_remote_manifest_filename(release):
--- a/src/clusterfuzz/_internal/bot/tasks/commands.py
+++ b/src/clusterfuzz/_internal/bot/tasks/commands.py
@@ -112,8 +112,11 @@ def is_supported_cpu_arch_for_job():
     # No specific cpu architecture requirement specified in job, bail out.
     return True
 
-  # Convert to list just in case anyone specifies value as a single string.
-  supported_cpu_arch_list = list(supported_cpu_arch)
+  if isinstance(supported_cpu_arch, list):
+    supported_cpu_arch_list = supported_cpu_arch
+  else:
+    supported_cpu_arch_list = utils.parse_delimited(
+        str(supported_cpu_arch), delimiter=',', strip=True, remove_empty=True)
 
   return cpu_arch in supported_cpu_arch_list
 
--- a/src/clusterfuzz/_internal/bot/tasks/commands.py
+++ b/src/clusterfuzz/_internal/bot/tasks/commands.py
@@ -112,6 +112,8 @@ def is_supported_cpu_arch_for_job():
     # No specific cpu architecture requirement specified in job, bail out.
     return True
 
+  # We support specifying CPU_ARCH as a list e.g. ['armeabi', 'armeabi-v7a'] or
+  # as a string e.g. 'x86_64, arm64'.
   if isinstance(supported_cpu_arch, list):
     supported_cpu_arch_list = supported_cpu_arch
   else:
--- a/src/clusterfuzz/_internal/bot/tasks/update_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/update_task.py
@@ -72,7 +72,10 @@ def get_source_url():
       'Darwin': 'macos'
   }
   platform_name = platform_mappings[platform_name]
-  if platform_name == 'macos' and environment.get_cpu_arch() == 'arm64':
+
+  # macOS bots run on both x86_64 and arm64 (Apple Silicon), requiring distinct
+  # deployment bundles. Other desktop platforms currently share a single bundle.
+  if platform_name == 'macos' and platform.machine().lower() == 'arm64':
     platform_name = 'macos_arm64'
 
   return _deployment_file_url(
--- a/src/clusterfuzz/_internal/bot/tasks/update_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/update_task.py
@@ -66,20 +66,20 @@ def get_source_url():
   """Return the source URL."""
   release = utils.get_clusterfuzz_release()
   platform_name = platform.system()
-  platform_mappings = {
+  deployment_target_mappings = {
       'Linux': 'linux',
       'Windows': 'windows',
       'Darwin': 'macos'
   }
-  platform_name = platform_mappings[platform_name]
+  deployment_target = deployment_target_mappings[platform_name]
 
   # macOS bots run on both x86_64 and arm64 (Apple Silicon), requiring distinct
   # deployment bundles. Other desktop platforms currently share a single bundle.
-  if platform_name == 'macos' and platform.machine().lower() == 'arm64':
-    platform_name = 'macos_arm64'
+  if deployment_target == 'macos' and platform.machine().lower() == 'arm64':
+    deployment_target = 'macos_arm64'
 
   return _deployment_file_url(
-      utils.get_platform_deployment_filename(platform_name, release))
+      utils.get_deployment_target_filename(deployment_target, release))
 
 
 def get_source_manifest_url():
--- a/src/clusterfuzz/_internal/bot/tasks/update_task.py
+++ b/src/clusterfuzz/_internal/bot/tasks/update_task.py
@@ -72,6 +72,9 @@ def get_source_url():
       'Darwin': 'macos'
   }
   platform_name = platform_mappings[platform_name]
+  if platform_name == 'macos' and environment.get_cpu_arch() == 'arm64':
+    platform_name = 'macos_arm64'
+
   return _deployment_file_url(
       utils.get_platform_deployment_filename(platform_name, release))
 
--- a/src/clusterfuzz/_internal/system/environment.py
+++ b/src/clusterfuzz/_internal/system/environment.py
@@ -17,6 +17,7 @@ import ast
 import enum
 import functools
 import os
+import platform as platform_util
 import re
 import socket
 import subprocess
@@ -227,8 +228,19 @@ def get_cpu_arch():
     from clusterfuzz._internal.platforms import android
     return android.settings.get_cpu_arch()
 
-  # FIXME: Add support for desktop architectures as needed.
-  return None
+  # Allow environment override if set (e.g. for testing / emulation).
+  cpu_arch_override = get_value('BOT_CPU_ARCH')
+  if cpu_arch_override:
+    return cpu_arch_override
+
+  machine = platform_util.machine().lower()
+  if machine in ('arm64', 'aarch64'):
+    return 'arm64'
+  if machine in ('x86_64', 'amd64'):
+    return 'x86_64'
+  if machine in ('i386', 'i686', 'x86'):
+    return 'x86'
+  return machine or None
 
 
 def get_current_memory_tool_var():
@@ -271,18 +283,28 @@ def get_instrumented_libraries_paths():
 
 
 def get_default_tool_path(tool_name):
-  """Get the default tool for this platform (from scripts/ dir)."""
+  """Get the default tool for this platform (from resources/ directory)."""
   if is_android():
     # For android devices, we do symbolization on the host machine, which is
     # linux. So, we use the linux version of llvm-symbolizer.
     platform_override = 'linux'
+    cpu_arch = None
   else:
     # No override needed, use default.
     platform_override = None
+    cpu_arch = get_cpu_arch()
 
   tool_filename = get_executable_filename(tool_name)
-  tool_path = os.path.join(
-      get_platform_resources_directory(platform_override), tool_filename)
+  platform_res_dir = get_platform_resources_directory(platform_override)
+
+  # Check for architecture-specific tool path
+  # (e.g. resources/platform/mac/arm64/tool).
+  if cpu_arch:
+    arch_tool_path = os.path.join(platform_res_dir, cpu_arch, tool_filename)
+    if os.path.exists(arch_tool_path):
+      return arch_tool_path
+
+  tool_path = os.path.join(platform_res_dir, tool_filename)
   return tool_path
 
 
--- a/src/clusterfuzz/_internal/system/environment.py
+++ b/src/clusterfuzz/_internal/system/environment.py
@@ -228,11 +228,6 @@ def get_cpu_arch():
     from clusterfuzz._internal.platforms import android
     return android.settings.get_cpu_arch()
 
-  # Allow environment override if set (e.g. for testing / emulation).
-  cpu_arch_override = get_value('BOT_CPU_ARCH')
-  if cpu_arch_override:
-    return cpu_arch_override
-
   machine = platform_util.machine().lower()
   if machine in ('arm64', 'aarch64'):
     return 'arm64'
@@ -288,23 +283,13 @@ def get_default_tool_path(tool_name):
     # For android devices, we do symbolization on the host machine, which is
     # linux. So, we use the linux version of llvm-symbolizer.
     platform_override = 'linux'
-    cpu_arch = None
   else:
     # No override needed, use default.
     platform_override = None
-    cpu_arch = get_cpu_arch()
 
   tool_filename = get_executable_filename(tool_name)
-  platform_res_dir = get_platform_resources_directory(platform_override)
-
-  # Check for architecture-specific tool path
-  # (e.g. resources/platform/mac/arm64/tool).
-  if cpu_arch:
-    arch_tool_path = os.path.join(platform_res_dir, cpu_arch, tool_filename)
-    if os.path.exists(arch_tool_path):
-      return arch_tool_path
-
-  tool_path = os.path.join(platform_res_dir, tool_filename)
+  tool_path = os.path.join(
+      get_platform_resources_directory(platform_override), tool_filename)
   return tool_path
 
 
--- a/src/clusterfuzz/_internal/tests/core/system/environment_test.py
+++ b/src/clusterfuzz/_internal/tests/core/system/environment_test.py
@@ -479,8 +479,15 @@ class GetCpuArchTest(unittest.TestCase):
     test_helpers.patch(self, [
         'clusterfuzz._internal.system.environment.is_android',
         'platform.machine',
+        'clusterfuzz._internal.platforms.android.settings.get_cpu_arch',
     ])
     self.mock.is_android.return_value = False
+
+  def test_android(self):
+    """Test Android architecture delegation."""
+    self.mock.is_android.return_value = True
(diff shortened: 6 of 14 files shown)


g-ortuno:
Looks like All-Tests and Run basic tests are failing. Can you fix those?

src/clusterfuzz/_internal/system/environment.py
```diff
@@ -227,8 +228,19 @@ def get_cpu_arch():
     from clusterfuzz._internal.platforms import android
     return android.settings.get_cpu_arch()
 
-  # FIXME: Add support for desktop architectures as needed.
-  return None
+  # Allow environment override if set (e.g. for testing / emulation).
+  cpu_arch_override = get_value('BOT_CPU_ARCH')
```
g-ortuno: Tests can mock platform_util directly; what testing/emulation is setting this variable?

src/clusterfuzz/_internal/bot/tasks/commands.py
```diff
@@ -112,8 +112,11 @@ def is_supported_cpu_arch_for_job():
     # No specific cpu architecture requirement specified in job, bail out.
     return True
 
-  # Convert to list just in case anyone specifies value as a single string.
```
g-ortuno, line 115: Let's keep this comment, but make it more deliberate i.e.: "We support specifying CPU_ARCH as a list e.g. `['armeabi', 'armeabi-v7a']` or as a string e.g. `x86_64, arm64`.

src/clusterfuzz/_internal/bot/tasks/update_task.py
```diff
@@ -72,6 +72,9 @@ def get_source_url():
       'Darwin': 'macos'
   }
   platform_name = platform_mappings[platform_name]
+  if platform_name == 'macos' and environment.get_cpu_arch() == 'arm64':
```
g-ortuno: Why do we only do this for macOS? Can you add a comment to give some context?

g-ortuno: Related to the comment in get_cpu_arch() but I don't think we should use get_cpu_arch() since that represents the target CPU architecture but this is the host architecture. For example, platform.system() above returns Linux for Android jobs, but get_cpu_arch() would return the architecture of the android devices e.g. arm64.
```
if platform_name == 'macos' and platform.machine().lower() == 'arm64':
  platform_name = 'macos_arm64'
```

src/clusterfuzz/_internal/system/environment.py
```diff
@@ -227,8 +228,19 @@ def get_cpu_arch():
...
+    return cpu_arch_override
+
+  machine = platform_util.machine().lower()
+  if machine in ('arm64', 'aarch64'):
+    return 'arm64'
+  if machine in ('x86_64', 'amd64'):
+    return 'x86_64'
+  if machine in ('i386', 'i686', 'x86'):
+    return 'x86'
+  return machine or None
```
g-ortuno, line 238: Are we changing the return value for all architectures? Is that not going to affect all platforms? I did the analysis and this is what I found:

It seems there's three callers of this function:

1. commands.is_supported_cpu_arch_for_job()
2. update_task.get_source_url()
3. environment.get_default_tool_path(tool_name)

#### 1. is_supported_cpu_arch_for_job()

Before this PR, since get_cpu_arch() returned None for desktop, we would hit the first if and we would return True.

With this PR, we don't hit that `if` so we go into the next `if` and check the value of CPU_ARCH. None of the desktop jobs set it, so we return True.

It seems like our change doesn't affect this method except if the job sets CPU_ARCH.

#### 2. update_task.get_source_url()

Here is where the trouble starts. It looks like get_cpu_arch() returns the **target** CPU architecture which doesn't necessarily match the host CPU architecture. For example, for Android, the host CPU architecture is amd64 since we run on linux bots, but the target architecture is arm64, the architecture of the Android device.

We should rename this method to get_target_cpu_arch() to clarify what this is.

#### 3. environment.get_default_tool_path()

This one is also problematic because of the target/host issue. I left a comment there.

src/clusterfuzz/_internal/system/environment.py
```diff
@@ -281,8 +293,16 @@ def get_default_tool_path(tool_name):
...
 
   tool_filename = get_executable_filename(tool_name)
-  tool_path = os.path.join(
-      get_platform_resources_directory(platform_override), tool_filename)
+  platform_res_dir = get_platform_resources_directory(platform_override)
+
+  # Check for architecture-specific tool path (e.g. resources/platform/mac/arm64/tool).
+  cpu_arch = get_cpu_arch()
+  if cpu_arch:
+    arch_tool_path = os.path.join(platform_res_dir, cpu_arch, tool_filename)
```
g-ortuno: I don't think we should be checking if the tool exists in the path. We should have an invariant that all tools for the (platform, arch) are in the respective platform/[platform]/[arch?] path. Otherwise, we silently fall back to the non-arch-specific tool which might not be the correct tool for the architecture.

Also note that since we're using `get_cpu_arch()` on Android, we're checking if the tool is in "resources/platform/linux/arm64" which will never be true. 

I also don't think `get_default_tool_path` is the right function to change the directory. It looks like `get_platform_resources_directory` would be a better primitive to change.

src/local/butler/common.py
g-ortuno, line 212: Looks like we only use this for tests that no longer work 💀 . No need to change anything.

src/local/butler/constants.py
```diff
@@ -38,22 +38,66 @@
...
+        'macosx_11_0_arm64',
+        'macosx_10_9_universal2',
+    )),
+    ('darwin_arm64', (
+        'macosx_14_0_arm64',
+        'macosx_13_0_arm64',
+        'macosx_12_0_arm64',
+        'macosx_11_0_arm64',
+        'macosx_10_9_universal2',
+    )),
```
g-ortuno: Why do we have both `macos_arm64` and `darwin_arm64`?

src/local/butler/constants.py
```diff
@@ -38,22 +38,66 @@
 PLATFORMS = collections.OrderedDict([
     ('windows', 'win_amd64'),
     ('macos', ('macosx_10_14_x86_64', 'macosx_10_12_x86_64')),
+    ('macos_arm64', (
+        'macosx_14_0_arm64',
```
g-ortuno: Do you know what version of macOS our bots run? My understanding is that we can't run 14 wheels on macOS 13. Please add a comment mentioning that the max version should match our bots' version.

src/local/butler/common.py
```diff
@@ -299,25 +302,36 @@ def _install_platform_pip(requirements_path, target_path, platform_name):
...
-    raise RuntimeError(
-        f'Failed to find package in supported platforms: {pip_platforms}')
+  with open(requirements_path, 'r') as f:
+    requirements = [
+        line.strip()
+        for line in f
+        if line.strip() and not line.startswith('#')
+    ]
+
+  for req in requirements:
```
g-ortuno, line 314: Please add a comment explaining why we do it this way. It's not obvious why we have to iterate over the requirements instead of just passing the requirements file.

src/platform_requirements.txt
```diff
@@ -1,4 +1,10 @@
+bcrypt==4.1.2
```
g-ortuno, line 11: Can we add a comment to this? Something along the lines of:

>This file contains dependencies from src/Pipfile that contain native/compiled C/C++/Rust extensions.
>
>When packaging bot deployment bundles for non-Linux platforms (e.g. Windows, macOS, macOS ARM64) on a Linux build machine, butler uses this list to download pre-compiled binary wheels for the target platform and extract them into src/third_party.
>
>If you add or bump a package in Pipfile that has native extensions, make sure to update it here as well.

Can you also add a similar comment to Pipfile so people remember to change this file too?

butler.py
```diff
@@ -125,7 +126,10 @@ def _add_package_subparser(toplevel_subparsers):
   parser_package = toplevel_subparsers.add_parser(
       'package', help='Package clusterfuzz with a staging revision')
   parser_package.add_argument(
-      '-p', '--platform', choices=['linux', 'macos', 'windows', 'all'])
+      '-p',
+      '--platform',
+      choices=list(constants.PLATFORMS.keys()) + ['all'],
+      default='all')
```
g-ortuno: Let's not add a default when there was none before.

src/local/butler/common.py
```diff
@@ -299,25 +302,36 @@ def _install_platform_pip(requirements_path, target_path, platform_name):
...
+    for pip_platform_entry in pip_platforms:
+      temp_dir = tempfile.mkdtemp()
+      return_code, _ = execute(
+          f'{_pip()} download --no-deps --only-binary=:all: '
+          f'--platform={pip_platform_entry} --abi={pip_abi} "{req}" -d '
+          f'{temp_dir}',
+          exit_on_error=False)
+
+      if return_code != 0:
+        shutil.rmtree(temp_dir, ignore_errors=True)
```
g-ortuno: Why do we delete files when we didn't before? If we have to delete the tmp files could we use `tempfile.TemporaryDirectory()` instead?

g-ortuno (changes requested):
Please add a description to the PR! It should summarize all the changes we are making and why.

src/local/butler/constants.py
```diff
@@ -38,22 +38,66 @@
 PLATFORMS = collections.OrderedDict([
```
g-ortuno: We're overloading the "platforms" term and it's making things confusing. Inside environment, `platform()` returns the Operating system; we have a separate function, `get_cpu_arch()` to get the architecture. But in butler, `PLATFORMS` combines both OS and architecture e.g. macos_arm64. Could we rename this variable to "DEPLOYMENT_TARGETS" to make it clearer what we use this for?

JuanMBriones:
> Please add a description to the PR! It should summarize all the changes we are making and why.

Thanks for the review:D I just modified the commit message. Sorry I forgot about this!
Left comments inline and resolved them on a follow-up commit.

JuanMBriones:
> Looks like All-Tests and Run basic tests are failing. Can you fix those?

I already fixed failing tests. I saw an error on https://github.com/google/clusterfuzz/pull/5442/checks?check_run_id=99666204114 but it seems theres no correlation with the changes

```
Step #2 - "Build Ubuntu 24.04": E: Package 'google-cloud-sdk' has no installation candidate
Step #2 - "Build Ubuntu 24.04": The command '/bin/sh -c ... apt-get install -y google-cloud-sdk' returned a non-zero code: 100
ERROR: build step 2 "gcr.io/cloud-builders/docker" failed: step exited with non-zero status: 100
```

src/local/butler/constants.py
JuanMBriones: We are running MacOS 13.6.7, MacOS 14.5 and MacOS 15.7. Since macOS 14 bots can execute `macosx_14_0_arm64` wheels down to earlier fallback tags, `macosx_14_0_arm64` is the appropriate maximum deployment target. Added a comment documenting that this tag aligns with our fleet's macOS version.

src/local/butler/constants.py
JuanMBriones: Thanks for the tip! Renamed `PLATFORMS` to `DEPLOYMENT_TARGETS` and kept `PLATFORMS = DEPLOYMENT_TARGETS` as an alias for backwards compatibility. Is on the new commit https://github.com/google/clusterfuzz/pull/5442/changes/3e406ac1a37792dd32ce0afba36914454571ab8a

src/clusterfuzz/_internal/system/environment.py
JuanMBriones: Removed `BOT_CPU_ARCH` and updated tests to mock `platform_util.machine()` / `platform.machine()` directly.

src/local/butler/constants.py
JuanMBriones: Removed `darwin_arm64` to avoid duplication since `macos_arm64` is the standard target identifier. I don't think it will be useful to have "darwin" defined.

src/local/butler/common.py
JuanMBriones: I didn't knew that, thanks!! I just refactored to use `tempfile.TemporaryDirectory()` context manager for clean and automatic temporary directory cleanup.

src/local/butler/common.py
JuanMBriones: I put an explanation on why we are using this approach detailing that requirements are downloaded one by one so `pip download` can evaluate fallback platform tags individually per wheel.

src/local/butler/common.py
JuanMBriones: lol 💀💀 So, should I delete all the changes on this commit related to chrome driver?

src/platform_requirements.txt
JuanMBriones: Thanks for tohe suggestion:)

src/clusterfuzz/_internal/bot/tasks/update_task.py
JuanMBriones: I added a comment before this line
```
  # macOS bots run on both x86_64 and arm64 (Apple Silicon), requiring distinct
  # deployment bundles. Other desktop platforms currently share a single bundle.
```
https://github.com/google/clusterfuzz/pull/5442/changes/3e406ac1a37792dd32ce0afba36914454571ab8a#diff-80a5a8c67ccee4970fdf4eea7407768f75d7a29e0374c882eaf49a60d03ad5a9

JuanMBriones: That makes sense. Switched to `platform.machine().lower() == 'arm64'` to check host architecture rather than bot/target configuration

src/clusterfuzz/_internal/system/environment.py
JuanMBriones: Ohh I get! I switched to use `get_default_tool_path()` without fallback in path searching

src/clusterfuzz/_internal/system/environment.py
JuanMBriones: Thanks for the help with this!! I solved this by separating host vs target architecture:
- In `update_task.get_source_url()`, we switched to `platform.machine().lower() == 'arm64'` to check the host machine architecture directly when fetching the deployment bundle.
- Reverted `get_default_tool_path()` to preserve directory invariants without fallback logic.
- Removed `BOT_CPU_ARCH` and kept `get_cpu_arch()` clean.

src/clusterfuzz/_internal/system/environment.py
g-ortuno, line 224: After this PR lands, can you make a new PR to rename this function to "get_target_cpu_arch" and add a new "get_host_cpu_arch()" function? We can call get_host_cpu_arch() function in the places where we use `platform.machine()`.

src/local/butler/common.py
g-ortuno: It's alright.

src/local/butler/constants.py
g-ortuno: Sorry, to clarify, I was asking, what macOS version the swarming bots that will run our fuzzers use? If they are running macOS 13, then downloading packages for macOS 14 would cause issues.

src/local/butler/constants.py
```diff
@@ -34,26 +34,62 @@
...
+        'macosx_14_0_arm64',
+        'macosx_13_0_arm64',
+        'macosx_12_0_arm64',
+        'macosx_11_0_arm64',
+        'macosx_10_9_universal2',
+    )),
+    ('linux', ('manylinux2014_x86_64',)),
 ])
 
+# Alias for backward compatibility.
```
g-ortuno: Why do we need this? Can we change all call sites to use the new DEPLOYMENT_TARGETS variable?

g-ortuno (changes requested):
Just had a few small comments. Still need to take a look at the tests; I'll do those tomorrow morning.

src/clusterfuzz/_internal/bot/tasks/update_task.py
```diff
@@ -72,6 +72,12 @@ def get_source_url():
       'Darwin': 'macos'
   }
   platform_name = platform_mappings[platform_name]
```
g-ortuno: To keep cleaning up the platform/deployment target mess, can we rename this variable to deployment_target?

src/clusterfuzz/_internal/bot/tasks/update_task.py
```diff
@@ -72,6 +72,12 @@ def get_source_url():
...
+
+  # macOS bots run on both x86_64 and arm64 (Apple Silicon), requiring distinct
+  # deployment bundles. Other desktop platforms currently share a single bundle.
+  if platform_name == 'macos' and platform.machine().lower() == 'arm64':
+    platform_name = 'macos_arm64'
+
   return _deployment_file_url(
       utils.get_platform_deployment_filename(platform_name, release))
```
g-ortuno: Can we change get_platform_deployment_filename to use "deployment_target" instead of "platform"?

src/clusterfuzz/_internal/tests/core/system/environment_test.py
```diff
@@ -469,3 +469,65 @@ def test_function():
...
     self.assertEqual(None, test_function())
+
+
+class GetCpuArchTest(unittest.TestCase):
+  """Tests for get_cpu_arch."""
+
+  def setUp(self):
+    test_helpers.patch_environ(self)
+    test_helpers.patch(self, [
+        'clusterfuzz._internal.system.environment.is_android',
```
g-ortuno, line 480: how hard would it be to add tests for android?

g-ortuno (changes requested):

src/clusterfuzz/_internal/system/environment.py
JuanMBriones: Sure, I'll start working on this PR to land it ASAP once this lands

src/local/butler/constants.py
JuanMBriones: The Apple Silicon bots in the fleet run macOS 15 (Sequoia, e.g., 15.7) as well as macOS 13/14. We set the maximum version tag to macosx_15_0_arm64 and keep the older tags (14_0, 13__0, 12_0, 11_0, 10_9_universal2) in descending order.

When pip download resolves wheels, it matches against these tags from newest to oldest. Since macOS ABI is backward compatible (binaries built against older SDKs run on newer OS releases), this guarantees we download packages compatible with the bot's OS version without pulling packages built against an SDK newer than what the host supports

JuanMBriones: In this change I removed the `PLATFORMS = DEPLOYMENT_TARGETS` alias and updated all call sites across `src/local/butler/deploy.py` and `src/local/butler/package.py` to use `constants.DEPLOYMENT_TARGETS` directly.

src/clusterfuzz/_internal/bot/tasks/update_task.py
JuanMBriones: Renamed platform_name to deployment_target (and updated platform_mappings to deployment_target_mappings as well)

src/clusterfuzz/_internal/bot/tasks/update_task.py
JuanMBriones: Done! :D

src/clusterfuzz/_internal/tests/core/system/environment_test.py
JuanMBriones: Not too complicated. I added the following test `test_android` to `GetCpuArchTest` to verify it `get_cpou_arch` delegates correctly its settings.

src/clusterfuzz/_internal/base/utils.py
```diff
@@ -315,12 +315,14 @@ def _get_deployment_zip_release_suffix(release):
...
-  # Expects linux, macos or windows.
-  base_filename = platform
+def get_deployment_target_filename(deployment_target, release):
+  """Return the deployment filename for the given target."""
   release_filename_suffix = _get_deployment_zip_release_suffix(release)
-  return f'{base_filename}{release_filename_suffix}.zip'
+  return f'{deployment_target}{release_filename_suffix}.zip'
+
+
+# Alias for backward compatibility.
```
g-ortuno: We shouldn't add aliases for backwards compatibility, they are just code debt. We should rename the callsites.

src/local/butler/constants.py
g-ortuno: Discussed offline, we should set the max version to 13 to  match the min version of our bots.

src/clusterfuzz/_internal/base/utils.py
JuanMBriones: fixed it!

src/local/butler/constants.py
```diff
@@ -34,26 +34,54 @@
...
 
-# Supported Platforms and ABIS (newer to older order).
-PLATFORMS = collections.OrderedDict([
+# Supported Deployment Targets and ABIs (newer to older order).
+DEPLOYMENT_TARGETS = collections.OrderedDict([
     ('windows', 'win_amd64'),
     ('macos', ('macosx_10_14_x86_64', 'macosx_10_12_x86_64')),
-    ('linux', ('manylinux2014_x86_64')),
+    # The max macOS version tag should match our bots' OS version
+    # (currently macOS 15.x).
```
g-ortuno: I think this comment is out of date.

JuanMBriones: That's correct!! Changing it rn. TYSM:)

g-ortuno (approved):

hunsche (approved):

JuanMBriones:
Thank you so much for the review and the help on this one guys!! @g-ortuno @hunsche :D

PR #5443, opened 2026-08-31
Switch to offline stack trace symbolization on Windows

notvictorl:
Previously, `stack_symbolizer` bypassed symbolization on Windows, assuming Windows Clang ASan would provide provide symbolized stack traces through an online symbolizer. However, Windows Clang ASan's online symbolizer is flaky and could fail to restart the symbolizer. 

Chrome archives provide a `llvm-symbolizer.exe` for use but it's ignored due to favoring online symbolizer on windows on ClusterFuzz and its windows extraction was not supported. ClusterFuzz also includes an `llvm-symbolizer.exe` but this acts as a fallback if one is not included or `USE_DEFAULT_LLVM_SYMBOLIZER = True`.

Switches to offline symbolization for windows and updates `llvm-symbolizer.exe` tooling for Windows to LLVM 24. Adds support for extracting `llvm-symbolizer.exe` as a common file on `ChromeBuildArchive`s. Adds documentation for updating symbolizer binaries and adds unit testing that symbolization is performed on Windows.

Tested that the changes don't crash on dev by uploading a testcase: https://gosst.development.clusterfuzz.com/testcase-detail/5136503223648256

b/552035861


notvictorl:
I'll also try testing this on dev

resources/platform/README.md
```diff
@@ -0,0 +1,46 @@
+# Updating Platform Binaries
+
+This directory contains prebuilt binaries (such as `llvm-symbolizer`) used for crash stack trace symbolization across platforms.
+
+## Updating `llvm-symbolizer`
+
+Platform symbolizer binaries are extracted from the Clang toolchain distributed via [CIPD](https://chrome-infra-packages.appspot.com/).
```
dylanjew, line 7: how often would we need to update this? would it be worth automating/doing in CD?

resources/platform/README.md
```diff
@@ -0,0 +1,46 @@
...
+# 3. Copy binary and set permissions
+# Note: add .exe to llvm-symbolizer if windows (llvm-symbolizer.exe)
+cp "/tmp/clang/bin/llvm-symbolizer" "resources/platform/${OS}/llvm-symbolizer"
+chmod 0755 "resources/platform/${OS}/llvm-symbolizer"
+
+# 4. Clean up temporary files
+rm -rf /tmp/clang
+
+# 5. Verify stack symbolizer tests
+python3 butler.py py_unittest -t core -p stack_symbolizer_test.py
```
dylanjew: does this test actually use the local llvm-symbolizer? e.g. does a test fail when you delete this file?

this is good practice but I don't want to give myself false confidence 😆

dylanjew (commented):
@ViniciustCosta Do you know if OSSFuzz does any Windows fuzzing? I'm wondering if we should mark this under a `chrome` directory?

resources/platform/README.md
notvictorl: good question, I'll try deleting to see if it doesn't symbolize

notvictorl: Okay forgot the unit tests mocked the outputs and are running on non-windows usually so running them here wouldn't do much :/

I'll delete this line since it doesn't test anything. I think all we can do is check the hash on review since this is not easily testable especially on windows.

resources/platform/README.md
notvictorl: We shouldn't need to update this that frequently (probably every 2+ years) so its not worth automating. We should only need to update if we start encountering issues with `llvm-symbolizer`. 

For Chrome, this is rather a fallback since we package `llvm-symbolizer.exe` in archives and the removal of the bypass is what's more important.

notvictorl:
> I'll also try testing this on dev

Need to clarify on this PR that Chrome has a primary `llvm-symbolizer.exe` included for windows so the updated `llvm-symbolizer.exe` is a fallback. Also, can't upload testcases on windows so I don't think we can easily verify the symbolizer works as intended until it hits a crash to output :/

will make some corrections

resources/platform/README.md
```diff
@@ -0,0 +1,44 @@
...
+chmod 0755 "resources/platform/${OS}/llvm-symbolizer"
+
+# 4. Clean up temporary files
+rm -rf /tmp/clang
+
+```
+
+To list available package instances:
+```bash
+cipd instances "fuchsia/third_party/clang/${PLATFORM}" -limit 5
```
g-ortuno, line 42: Why fuchsia? 😅

g-ortuno:
A couple of questions:

1. Does ClusterFuzz actually extract the llvm-symbolizer? Does the "llvm-symbolizer" matcher in [get_target_dependencies](https://github.com/google/clusterfuzz/blob/d937566111ea06c27a9ec05d1fb6e1b6da86db25/src/clusterfuzz/_internal/build_management/build_archive.py#L454) find `llvm-symbolizer.exe`? Can we add a test?
2. My understanding from [this comment](https://github.com/google/clusterfuzz/blob/d937566111ea06c27a9ec05d1fb6e1b6da86db25/src/clusterfuzz/_internal/system/environment.py#L201) was that we disabled the external symbolizer because it didn't work. I guess we think it works now and we want to match linux and mac? If that's the case, we should probably change that code to not enable online symbolization.

> However, Windows does not automatically have symbolized stack traces so we should proceed with symbolizing the addresses.

Also this is not quite correct, I think, please correct me if I'm wrong! AFAICT, we do enable symbolization for Windows, see [this testcase's](https://clusterfuzz.com/testcase-detail/5415447180640256) original stack trace where it says we failed to restart the symbolizer. This is where my understanding gets a bit iffy, but I think since we're not extracting `llvm-symbolzer.exe`, we use the old one in the repo which fails and leaves the stack trace unsymbolized. Anyway, I think what you're doing by enabling symbolization is the right fix, but we should make sure we disable online symbolization.

letitz:
> Also, can't upload testcases on windows so I don't think we can easily verify the symbolizer works as intended until it hits a crash to output :/

You should be able to upload testcases to windows jobs, is there a known issue with that path?

notvictorl:
> You should be able to upload testcases to windows jobs, is there a known issue with that path?

What I receive when uploading to clusterfuzz is: 
```
ERROR: Job "windows_libfuzzer_chrome_asan" does not support running untrusted workloads. Untrusted workloads on Chrome are only supported on Linux.
```

ViniciustCosta:
> @ViniciustCosta Do you know if OSSFuzz does any Windows fuzzing? I'm wondering if we should mark this under a `chrome` directory?

Yeah I do think that only Chrome is currently fuzzing on Windows

letitz:
> > You should be able to upload testcases to windows jobs, is there a known issue with that path?
> 
> What I receive when uploading to clusterfuzz is:
> 
> ```
> ERROR: Job "windows_libfuzzer_chrome_asan" does not support running untrusted workloads. Untrusted workloads on Chrome are only supported on Linux.
> ```

You need to type in "This testcase is safe to run." or whatever the exact string is in the text box at the bottom of the upload form.

resources/platform/README.md
```diff
@@ -0,0 +1,44 @@
...
+### Steps
+
+```bash
+# 1. Set platform variables
+PLATFORM="linux-amd64" # linux-amd64 | mac-amd64 | windows-amd64
+OS="linux"             # linux | mac | windows
+
+# 2. Export Clang toolchain from CIPD
+mkdir -p /tmp/clang
+cipd export -root /tmp/clang -ensure-file - <<CIPD_EOF
```
g-ortuno, line 26: Isn't cipd part of depot_tools? Chrome publishes clang to `https://commondatastorage.googleapis.com/chromium-browser-clang/{PLATFORM}/clang-{VERSION}.tar.xz`

We could use curl to download it from there without needing cipd.

Might be overkill but we could add a butler script to do it as well.

notvictorl: good catch! CIPD is part of depot_tools (and fuchsia is the only llvm-symbolizer in CIPD but these are not specific to fuchsia so that should be fine).

I think adding a curl command to the instructions instead makes the most sense and in the future add a butler script for updating all the binary dependencies.

notvictorl:
> 1. Does ClusterFuzz actually extract the llvm-symbolizer? 

thanks for pointing this out, ClusterFuzz does extract the llvm-symbolizer for `DefaultBuildArchive`s but not `ChomeBuildArchive`s, so I'll need to modify that.

For the other points, we should use offline symbolization for windows because we can see its brittle and we know its consistently packaged in our archives and in ClusterFuzz as a fallback.

resources/platform/README.md
```diff
@@ -4,41 +4,50 @@ This directory contains prebuilt binaries (such as `llvm-symbolizer`) used for c
...
+### Target Platforms
 
 | Platform | `$PLATFORM` | Target Path |
 | :--- | :--- | :--- |
-| Linux | `linux-amd64` | `resources/platform/linux/llvm-symbolizer` |
-| macOS | `mac-amd64` | `resources/platform/mac/llvm-symbolizer` |
-| Windows | `windows-amd64` | `resources/platform/windows/llvm-symbolizer.exe` |
+| Linux | `Linux_x64` | `resources/platform/linux/llvm-symbolizer` |
+| macOS (Intel) | `Mac` | `resources/platform/mac/llvm-symbolizer` |
+| macOS (ARM64) | `Mac_arm64` | `resources/platform/mac/llvm-symbolizer` |
```
g-ortuno: Hmmm there's a bug in our mac arm64 code for this 🙃 . Unrelated to this CL though.

resources/platform/README.md
```diff
@@ -4,41 +4,50 @@ This directory contains prebuilt binaries (such as `llvm-symbolizer`) used for c
...
+| Windows | `Win` | `resources/platform/windows/llvm-symbolizer.exe` |
+
+### Finding the Clang Version
```
g-ortuno, line 20: I wonder if we should include a note here about what version we're currently using. The next person that updates llvm-symbolizer would update the note.

g-ortuno (approved):
Just one small thing otherwise LGTM!

resources/platform/README.md
notvictorl: oop I added a TODO to support mac arm64 in the README whenever its added. i think manuel is the contact for this (?)

resources/platform/README.md
notvictorl: that's a good idea. went ahead and documented the version for each llvm-symbolizer

resources/platform/README.md
dylanjew: is this captured in a bug? We aren't tracking TODOs in the clusterfuzz repo :)

resources/platform/README.md
dylanjew, line 1: thanks, this is helpful!

dylanjew (approved):
Can you update the description with the results from testing in dev?

Otherwise LGTM and approving since Windows is Chrome only.

resources/platform/README.md
dylanjew: Maybe add a child bug on b/537850811?

notvictorl: a bug to track does make more sense. created bug b/559164517 and changed it in the readme

PR #5444, opened 2026-09-01
Reverts task short circuit for duplicate test cases

dylanjew:
This was added in #5405, but in b/555138064 we found that this is a use case of Clusterfuzz to upload a better repro of an existing testcase and reupload it manually.


aakallam (approved):

ViniciustCosta (approved):
Thanks

dylanjew:
I pushed this change to dev and haven't seen any related issues. Going to merge this to unblock the v8 team

PR #5446, opened 2026-09-01
Update google-cloud-sdk to google-cloud-cli and fix apt-key deprecation

ViniciustCosta:
Fix base image Cloud Build failures when installing gcloud CLI/SDK dependencies on newer Ubuntu environments.

*  Replaced all references to `google-cloud-sdk` (and its addons like `app-engine-python`) with `google-cloud-cli`. Google formally stopped publishing updates under the legacy  SDK package name, causing a "no installation candidate" failure on newer deployments. 

* Replaced deprecated `apt-key add` commands with direct keyring configurations. Ubuntu 24.04 strictly enforces this deprecation, so GPG keys are now piped securely to `/usr/share/keyrings/cloud.google.asc` and referenced via the `[signed-by=...]` directive.                                                                                                           
                                                                                                                                                                                                       
References:
    * https://cloud.google.com/sdk/docs/install#deb
    * https://wiki.debian.org/DebianRepository/UseThirdParty

### DIFF
--- a/docker/base/Dockerfile
+++ b/docker/base/Dockerfile
@@ -139,12 +139,11 @@ RUN apt update && apt install libc6:i386 -y
 
 ENV CLOUDSDK_PYTHON=python3.11
 
-RUN echo "deb https://packages.cloud.google.com/apt cloud-sdk main" \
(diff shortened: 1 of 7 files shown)


PauloVLB (approved):
LGTM. Thanks for fixing this! :D

PR #5448, opened 2026-09-02
Change FUZZ_TARGET_CWD_IS_BUILD_DIR to FUZZ_TARGET_CWD_IS_TARGET_DIR

notvictorl:
Setting cwd to BUILD_DIR caused errors on `libfuzzer_chrome_msan` when performing fuzzing due to a misalignment between the fuzz target binary (e.g., `xml_parser_fuzzer`) and the `ld-linux.so` dynamic linker. `FUZZ_TARGET_CWD_IS_BUILD_DIR` is currently an environment string used only for Libfuzzer MSan jobs since it is important to align the cwd of where the fuzz target binary is run to find the dynamic linker through relative paths otherwise the fuzzer won't work.

Aligning the cwd to the fuzz target directory should resolve the errors, so switching cwd to use the fuzz target's directory when environment string is set to true. Renaming the `FUZZ_TARGET_CWD_IS_BUILD_DIR` to `FUZZ_TARGET_CWD_IS_TARGET_DIR` to reflect the change in behavior.

Tested on dev by uploading a testcase and watching for fuzzing round errors. ClusterFuzz previously threw an error after 'Fuzzing round 0', but does not after the change.

crbug.com/536875721

### DIFF
--- a/src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py
+++ b/src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py
@@ -1135,9 +1135,10 @@ def get_runner(fuzzer_path, temp_dir=None, use_minijail=None):
     temp_dir = fuzzer_utils.get_temp_dir()
 
   build_dir = environment.get_value('BUILD_DIR')
+  target_dir = os.path.dirname(fuzzer_path)
 
-  cwd = build_dir if environment.get_value(
-      'FUZZ_TARGET_CWD_IS_BUILD_DIR') else None
+  cwd = target_dir if environment.get_value(
+      'FUZZ_TARGET_CWD_IS_TARGET_DIR') else None
 
   is_android = environment.is_android()
   is_fuchsia = environment.platform() == 'FUCHSIA'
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py
```diff
@@ -1135,9 +1135,10 @@ def get_runner(fuzzer_path, temp_dir=None, use_minijail=None):
     temp_dir = fuzzer_utils.get_temp_dir()
 
   build_dir = environment.get_value('BUILD_DIR')
+  target_dir = os.path.dirname(fuzzer_path)
 
-  cwd = build_dir if environment.get_value(
```
dylanjew, line 1139: just for my understanding, we thought the build_dir was the right spot to execute msan, but it turns out that the fuzz target dir is actually what we want.

so there's no reason to keep the old `FUZZ_TARGET_CWD_IS_BUILD_DIR` around.

Is that correct?

dylanjew (commented):
LGTM, but could you add a section to the PR description about how you tested this in dev?

src/clusterfuzz/_internal/bot/fuzzers/libfuzzer.py
notvictorl: Yes, that's correct.

BUILD_DIR worked previously when we were running the wrong instrumented libraries but now is the wrong directory.

notvictorl:
> LGTM, but could you add a section to the PR description about how you tested this in dev?

Updated description with testing. Doesn't seem to encounter errors after Fuzzing round 0 after the change

dylanjew (approved):

dylanjew:
FYI: @vitaliset we merged this but the flag `FUZZ_TARGET_CWD_IS_TARGET_DIR` and related msan jobs are only used by chrome.

@notvictorl  verified in dev that the msan job is now making it past the first round of fuzzing https://cloudlogging.app.goo.gl/AdVkDm4LcNmmLupP6

Previously it was failing as described in the PR description

PR #5450, opened 2026-09-03
[VRP] Add constant for checking allowed platforms

PauloVLB:
b/556677120

### Problem
Long-lived bots currently exit with a fatal error whenever they encounter an untrusted testcase (`trusted=False`). However, there are existing testcases in Datastore marked as untrusted that belong to non-Linux jobs (e.g. Windows, Mac). Since uworker execution only runs in Linux containers, non-Linux tasks have no other way to execute and are completely blocked.

This was not caught during dev testing due to the lack of non-Linux VMs in the dev environment. This change does not affect the VRP security model because untrusted external fuzzers are strictly restricted to Linux jobs.

### Proposed Solution
- Defined `PLATFORMS_SUPPORTING_UNTRUSTED_WORKLOADS = {'linux'}` in `uworker_io.py`.
- Added helper `is_untrusted_testcase_allowed_on_bot(testcase)` in `uworker_io.py` to check if the testcase's job platform is outside `PLATFORMS_SUPPORTING_UNTRUSTED_WORKLOADS`.
- Updated `check_handling_testcase_safe` in `uworker_io.py` to allow execution on long-lived bots for non-Linux jobs.

### Testing
- Added unit tests in `uworker_io_test.py` verifying that untrusted testcases on non-Linux platforms (Windows, Mac, Android) pass on long-lived bots, while Linux jobs raise `SystemExit`.
- Verified unit tests pass: `python butler.py py_unittest -t core -p uworker_io_test.py`
- Verified code passes linting: `python butler.py lint`

### DIFF
--- a/src/clusterfuzz/_internal/base/feature_flags.py
+++ b/src/clusterfuzz/_internal/base/feature_flags.py
@@ -43,7 +43,6 @@ class FeatureFlags(Enum):
   SWARMING_MAX_PENDING_TASKS = 'swarming_max_pending_tasks'
 
   ENABLE_FUZZ_FOR_BOTS = 'enable_fuzz_for_bots'
-  ENABLE_UNTRUSTED_TESTCASES_FOR_BOTS = 'enable_untrusted_testcases_for_bots'
   STORAGE_THREADED_OPS_FUZZ_TARGETS = 'storage_threaded_ops_fuzz_targets'
   CALL_ANDROID_API = 'call_android_api'
 
--- a/src/clusterfuzz/_internal/base/feature_flags.py
+++ b/src/clusterfuzz/_internal/base/feature_flags.py
@@ -43,6 +43,7 @@ class FeatureFlags(Enum):
   SWARMING_MAX_PENDING_TASKS = 'swarming_max_pending_tasks'
 
   ENABLE_FUZZ_FOR_BOTS = 'enable_fuzz_for_bots'
+  ENABLE_UNTRUSTED_TESTCASES_FOR_BOTS = 'enable_untrusted_testcases_for_bots'
   STORAGE_THREADED_OPS_FUZZ_TARGETS = 'storage_threaded_ops_fuzz_targets'
   CALL_ANDROID_API = 'call_android_api'
 
--- a/src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
+++ b/src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
@@ -26,7 +26,9 @@ from google.protobuf import any_pb2
 from google.protobuf import timestamp_pb2
 import google.protobuf.message
 
+from clusterfuzz._internal.base import feature_flags
 from clusterfuzz._internal.base.tasks import task_utils
+from clusterfuzz._internal.datastore import data_types
 from clusterfuzz._internal.google_cloud_utils import storage
 from clusterfuzz._internal.metrics import logs
 from clusterfuzz._internal.protos import uworker_msg_pb2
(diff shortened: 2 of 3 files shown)


src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
```diff
@@ -231,14 +233,37 @@ def entity_from_protobuf(entity_proto: any_pb2.Any, model_type: Type[T]) -> T:
...
+    return False
+
+  job = data_types.Job.query(data_types.Job.name == testcase.job_type).get()
+  if not job or not job.platform:
+    return False
+
+  allowed_platforms = [
+      p.strip().lower()
+      for p in enable_untrusted_flag.string_value.split(',')
+      if p.strip()
```
dylanjew: does this handle if there is just one platform? do you stil need to `lower()` the input `allowed_platforms`?

dylanjew (commented):
Is this only a temporary problem? Will all of the Testcases scheduled for Mac/windows  eventually be `trusted=True`?

dylanjew:
> Is this only a temporary problem? Will all of the Testcases scheduled for Mac/windows eventually be `trusted=True`?

I guess what I'm getting at is that the point of having the `is trusted` check is to prevent these untrusted testcases from appearing on these bots.  Is our only solution instead to ensure that untrusted fuzzers are never scheduled on non-linux bots? Is that enforced in any way when trying to configure the fuzzer jobs?

Do we need to go through and backfill the Testcases.trusted field from fuzzers that are enabled on windows and mac?

ViniciustCosta:
> I guess what I'm getting at is that the point of having the `is trusted` check is to prevent these untrusted testcases from appearing on these bots. Is our only solution instead to ensure that untrusted fuzzers are never scheduled on non-linux bots? Is that enforced in any way when trying to configure the fuzzer jobs?
> 
> Do we need to go through and backfill the Testcases.trusted field from fuzzers that are enabled on windows and mac?

Just to highlight that the issue is that there is no untrusted worker logic for windows/mac/android platforms. All testcases and fuzzers for these platforms must be "trusted" in the sense of being able to run on long-lived privileged bots (that's the current state). 

I think testcases uploaded to clusterfuzz are marked as untrusted. And there are codepaths that are blocked by the testcase not being marked as trusted which are not really related to bot execution, so we should keep the trusted=false even though they can still run on bots for that platform. If we remove this, we should understand the impact for most use cases. @PauloVLB may add more details about that

src/clusterfuzz/_internal/base/feature_flags.py
```diff
@@ -43,6 +43,7 @@ class FeatureFlags(Enum):
   SWARMING_MAX_PENDING_TASKS = 'swarming_max_pending_tasks'
 
   ENABLE_FUZZ_FOR_BOTS = 'enable_fuzz_for_bots'
+  ENABLE_UNTRUSTED_TESTCASES_FOR_BOTS = 'enable_untrusted_testcases_for_bots'
```
ViniciustCosta: I think this name is not very clear, without context it is difficult to understand. It should have something related to platforms that support the trusted/untrusted model.

ViniciustCosta: Also, I think we should set it as a global var instead of a feature flag. This will be true for android/mac/windows until there are significant architectural changes and not a feature that we can simply enable/disable. Wdyt?

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
PauloVLB: Yes, this handles cases where `string_value` is, for example, 'MAC', resulting in `allowed_platforms = ['mac']`. The `lower()` call is included as a safety measure in case the value was incorrectly set using uppercase letters.

PauloVLB:
> All testcases and fuzzers for these platforms must be "trusted" in the sense of being able to run on long-lived privileged bots (that's the current state).

Exactly. Since we are not yet able to run test cases for non-Linux platforms in ephemeral environments, we are obligated to run them on long-lived bots.

> I think testcases uploaded to clusterfuzz are marked as untrusted.

Yes. In `create_user_uploaded_testcase`, we set all user-uploaded testcases as untrusted: https://github.com/google/clusterfuzz/blob/d937566111ea06c27a9ec05d1fb6e1b6da86db25/src/clusterfuzz/_internal/datastore/data_handler.py#L1446

> And there are codepaths that are blocked by the testcase not being marked as trusted which are not really related to bot execution

Indeed: https://github.com/google/clusterfuzz/blob/d937566111ea06c27a9ec05d1fb6e1b6da86db25/src/clusterfuzz/_internal/bot/tasks/utasks/progression_task.py#L430-L431

I agree that we should not backfill non-Linux testcases as `trusted = True`, as that status is used for purposes beyond just bot execution eligibility.

src/clusterfuzz/_internal/base/feature_flags.py
PauloVLB: Agreed. I changed the name to `PLATFORMS_SUPPORTING_UNTRUSTED_WORKLOADS` and removed the feature flag.

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
```diff
@@ -231,14 +232,30 @@ def entity_from_protobuf(entity_proto: any_pb2.Any, model_type: Type[T]) -> T:
   return entity
 
 
+PLATFORMS_SUPPORTING_UNTRUSTED_WORKLOADS = {'linux'}
+
+
+def is_untrusted_testcase_allowed_on_bot(testcase) -> bool:
+  """Returns True if the untrusted testcase is on a platform that does not
```
ViniciustCosta, line 239: nit: go/pystyle#function-docs - although this is not nearly close to be enforced on clusterfuzz legacy code, usually it is better to have a one-liner docstring followed by a more detailed description after a blank line.

src/clusterfuzz/_internal/bot/tasks/utasks/uworker_io.py
```diff
@@ -231,14 +232,30 @@ def entity_from_protobuf(entity_proto: any_pb2.Any, model_type: Type[T]) -> T:
...
+  support the untrusted execution model (so it must run on a long-lived bot)."""
+  job = data_types.Job.query(data_types.Job.name == testcase.job_type).get()
```
ViniciustCosta, line 241: Why not use `testcase.platform()`? Can you check if this field is filled in the database?

PR #5453, opened 2026-09-03
[Android] Add exit code constants and ProcessExitInfo data model

IvanBM18:
Bug: b/553141628

## Overview
Since Android API level 30 (and all apps targeting Android 11+), Android introduced `ApplicationExitInfo` to track process exit reasons and historical diagnostic metadata via `dumpsys activity exit-info`.

Previously, testcase execution using `am start` almost 99% of the time returned success exit code 0 even when an application process crashed or failed silently. To accurately track process termination reasons and status metadata from `dumpsys activity exit-info`, this PR defines the structured exit code constants (`ExitReason`, `ExitStatus`) and the `ProcessExitInfo` data model.

The purpose of this PR stack is to perform the following validation for each test case execution:
- Get the current android jobs app's pacakge
- then we look for the pid for our pacakge in the android's activity manager logs.
- once we have the pid, we can match it against the dumpys exit_info <package> to find out if our latest test-case indeed finished.
- If it finished, we will check the reason & status of why the process finished, this is basically the exit code & details as to why the process ended.
- with the exit info we can determine if it failed due to a runtime issue or not.

Learn more:
• https://developer.android.com/reference/android/app/ApplicationExitInfo

## Changes
- `src/clusterfuzz/_internal/platforms/android/constants.py`: Added `ExitReason` and `ExitStatus` constant definitions.
- `src/clusterfuzz/_internal/platforms/android/util.py`: Defined `ProcessExitInfo` dataclass for `dumpsys activity exit-info` parsing.

## PR stack
- `master`
  - **#PR 2.1a** `feature/android-exit-code-constants` 👈

### DIFF
--- a/src/clusterfuzz/_internal/platforms/android/constants.py
+++ b/src/clusterfuzz/_internal/platforms/android/constants.py
@@ -91,6 +91,7 @@ class ExitReason(IntEnum):
   https://developer.android.com/reference/android/app/ApplicationExitInfo
   """
 
+  UNKNOWN = 0
   EXIT_SELF = 1
   SIGNALED = 2
   LOW_MEMORY = 3
--- a/src/clusterfuzz/_internal/platforms/android/constants.py
+++ b/src/clusterfuzz/_internal/platforms/android/constants.py
@@ -112,10 +112,12 @@ class ExitReason(IntEnum):
 
 
 class ExitStatus(IntEnum):
-  """Process exit signal statuses. Taken from:
-  https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/base/core/java/android/os/Process.java?q=content:SIGNAL_KILL
+  """Process exit signal statuses corresponding to POSIX signals.
+
+  See: https://developer.android.com/reference/android/os/Process
   """
 
+  SIGQUIT = 3
   SIGILL = 4
   SIGABRT = 6
   SIGKILL = 9
--- a/src/clusterfuzz/_internal/platforms/android/constants.py
+++ b/src/clusterfuzz/_internal/platforms/android/constants.py
@@ -106,11 +106,13 @@ class ExitReason:
   FREEZER = 14
   PACKAGE_STATE_CHANGE = 15
   PACKAGE_UPDATED = 16
+  REASON_MEMORY_LIMITER = 17
+  REASON_ANOMALY = 18
 
 
 class ExitStatus:
-  """Process exit signal statuses. Taken from 
-  https://developer.android.com/reference/android/app/ApplicationExitInfo
+  """Process exit signal statuses. Taken from:
+  https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/base/core/java/android/os/Process.java?q=content:SIGNAL_KILL
   """
 
   SIGILL = 4
--- a/src/clusterfuzz/_internal/platforms/android/constants.py
+++ b/src/clusterfuzz/_internal/platforms/android/constants.py
@@ -83,3 +83,33 @@ DEPRECATED_DEVICE_LIST = [
 # Restrict pixel6 from picking up generic Android jobs to avoid
 # Binary Mismatch: Hence, 'ANDROID:PIXEL6' is added to the list.
 DEVICES_WITH_NO_FALLBACK_QUEUE_LIST = ['ANDROID:PIXEL6']
+
+
+class ExitReason:
+  """Android process exit reasons from ApplicationExitInfo."""
+
+  EXIT_SELF = 1
+  SIGNALED = 2
+  LOW_MEMORY = 3
+  CRASH = 4
+  CRASH_NATIVE = 5
+  ANR = 6
+  INITIALIZATION_FAILURE = 7
+  PERMISSION_CHANGE = 8
+  EXCESSIVE_RESOURCE_USAGE = 9
+  USER_REQUESTED = 10
+  USER_STOPPED = 11
+  DEPENDENCY_DIED = 12
+  OTHER_KILLS_BY_SYSTEM = 13
+  FREEZER = 14
+  PACKAGE_STATE_CHANGE = 15
+  PACKAGE_UPDATED = 16
+
+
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/platforms/android/constants.py
```diff
@@ -86,3 +86,37 @@
...
+  INITIALIZATION_FAILURE = 7
+  PERMISSION_CHANGE = 8
+  EXCESSIVE_RESOURCE_USAGE = 9
+  USER_REQUESTED = 10
+  USER_STOPPED = 11
+  DEPENDENCY_DIED = 12
+  OTHER_KILLS_BY_SYSTEM = 13
+  FREEZER = 14
+  PACKAGE_STATE_CHANGE = 15
+  PACKAGE_UPDATED = 16
```
Xeicker, line 113: In the link you share, I see there is also 17 & 18

src/clusterfuzz/_internal/platforms/android/constants.py
```diff
@@ -86,3 +86,37 @@
...
+
+
+class ExitStatus:
```
Xeicker: I think this should be renamed to reflect that it only covers statuses when there is a signal.
on REASON_EXIT_SELF status number comes from exit function

src/clusterfuzz/_internal/platforms/android/util.py
```diff
@@ -13,13 +13,25 @@
...
 from clusterfuzz._internal.system import environment
 
 
+@dataclass(frozen=True)
+class ProcessExitInfo:
+  """DTO representing process exit metadata from dumpsys activity exit-info."""
+
+  reason: int
+  reason_name: str  # e.g., 'APP CRASH(NATIVE)', 'SIGNALED'
+  subreason: int
```
Xeicker, line 34: I don't see the definition of sub reason in the link you provided.

Don't we need a class in `constants.py`?

src/clusterfuzz/_internal/platforms/android/constants.py
IvanBM18: Done, Thanks!

src/clusterfuzz/_internal/platforms/android/util.py
IvanBM18: At the time when i created this PR I couldn't find info for this in the android docs, but i just find references in the [android code](https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/proto_logging/stats/enums/app_shared/app_enums.proto?q=content:subreason)

TBF we don't use them atm, hence i think is not required to define an enum that wont be used, or what do you think?

src/clusterfuzz/_internal/platforms/android/constants.py
IvanBM18: Its also used when REASON_ SIGNALED,
See:
https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/proto_logging/stats/enums/app_shared/app_enums.proto;l=172?q=content:subreason

src/clusterfuzz/_internal/platforms/android/util.py
Xeicker: I think this link should be added to some clusterfuzz docs in case we have something for debugging. It looks useful to triage errors from logs.
Not sure if we have that though

Xeicker (approved):

src/clusterfuzz/_internal/platforms/android/util.py
IvanBM18: Sure, since we have 0 public docs on android fuzzing, i just added it as a comment on the dataclass definition. thanks for your feedback

src/clusterfuzz/_internal/platforms/android/util.py
```diff
@@ -13,13 +13,29 @@
...
 
+@dataclass(frozen=True)
+class ProcessExitInfo:
+  """DTO representing process exit metadata from dumpsys activity exit-info.
+
+  For a full list of android exit info reasons and subreasons, see:
+  https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/proto_logging/stats/enums/app_shared/app_enums.proto;l=270?q=content:subreason
+  """
+
+  reason: int
```
dylanjew: do these map to ExitReason in constants.py? should you use those instead of int and string? then you'd be able to log any errors if the code doesn't match an expected `ExitReason` during conversion

src/clusterfuzz/_internal/platforms/android/util.py
dylanjew: it is a bit strange to have a codesearch link here. It's helpful to know but is there a better way to surface this? Does the GOSST team have ways they've done this in the past?

src/clusterfuzz/_internal/platforms/android/util.py
IvanBM18: Good observation!, i just updated the enum here, and i'll update the error parsing/logging in its corresponding pr, thanks!

src/clusterfuzz/_internal/platforms/android/util.py
IvanBM18: The other alternative is to use this one:
https://android.googlesource.com/platform/frameworks/proto_logging/+/refs/heads/main/stats/enums/app_shared/app_enums.proto
WDYT?

dylanjew: oh, it looks like cs.android.com is public, so what you already have is fine!

dylanjew (commented):
LGTM! please confirm with GOSST whether this will affect the android Clusterfuzz instance and get approval if necessary

decoNR (approved):

decoNR:
> LGTM! please confirm with GOSST whether this will affect the android Clusterfuzz instance and get approval if necessary

It will, @ViniciustCosta sent a message in the Haiku chat to provide visibility regarding these and the following changes (since they are the ones that monitor this part).

PR #5454, opened 2026-09-03
[Android] Implement exit info extraction and process crash parsing

IvanBM18:
Bug: b/553141628

## Overview
Since Android API level 30 (and all apps targeting Android 11+), Android introduced `ApplicationExitInfo` to report process exit reasons and historical diagnostic metadata via `dumpsys activity exit-info`.

Testcase execution using `am start` almost 90% of the time returns exit code 0 regardless of whether the activity crashed or failed. To overcome this, this PR introduces core helper utilities to fetch `dumpsys activity exit-info` via ADB, extract process PIDs from ActivityManager logcat output, parse termination metadata, and evaluate activity crashes.

Learn more:
• https://developer.android.com/reference/android/app/ApplicationExitInfo

## Changes
- `src/clusterfuzz/_internal/platforms/android/adb.py`: Added `get_activity_exit_info()` to fetch dumpsys exit metadata for a package.
- `src/clusterfuzz/_internal/platforms/android/logger.py`: Added `log_activity_manager_output()` helper to read ActivityManager logcat entries.
- `src/clusterfuzz/_internal/platforms/android/util.py`: Implemented `get_latest_pid_for_package()`, `get_exit_info_for_pid()`, `activity_crashed()`, and `activity_crashed_by_package()`.
- `src/clusterfuzz/_internal/tests/core/platforms/android/util_test.py`: Added comprehensive unit tests for PID retrieval, exit-info parsing, and crash evaluation.

## Tests performed
```
```

## PR stack
- `master`
  - **#PR 2.1b** `feature/android-exit-code-core` 👈

### DIFF
--- a/src/clusterfuzz/_internal/platforms/android/adb.py
+++ b/src/clusterfuzz/_internal/platforms/android/adb.py
@@ -915,3 +915,18 @@ def write_data_to_file(contents, file_path, should_reboot=True):
     else:
       # Manually revert /system to read-only since we aren't rebooting.
       run_shell_command('mount -o ro,remount /system', root=True)
+
+
+def get_activity_exit_info(app_package: str) -> str:
+  """Get dumpsys activity exit-info output for the given application package.
+
+  Args:
+    app_package: Name of the application package.
+
+  Returns:
+    Dumpsys output for the application package.
+  """
+  dumpsys_output = run_shell_command(
+      ['dumpsys', 'activity', 'exit-info', app_package])
+  return dumpsys_output
+
--- a/src/clusterfuzz/_internal/platforms/android/adb.py
+++ b/src/clusterfuzz/_internal/platforms/android/adb.py
@@ -929,4 +929,3 @@ def get_activity_exit_info(app_package: str) -> str:
   dumpsys_output = run_shell_command(
       ['dumpsys', 'activity', 'exit-info', app_package])
   return dumpsys_output
-
--- a/src/clusterfuzz/_internal/platforms/android/adb.py
+++ b/src/clusterfuzz/_internal/platforms/android/adb.py
@@ -919,6 +919,18 @@ def write_data_to_file(contents, file_path, should_reboot=True):
 
 def get_activity_exit_info(app_package: str) -> str:
   """Get dumpsys activity exit-info output for the given application package.
+  Example dumpsys output:
+  package: org.chromium.chrome
+  Historical Process Exit for uid=10154
+      ApplicationExitInfo #0:
+        timestamp=2026-09-03 11:19:03.480 pid=27098 realUid=10154
+        packageUid=10154 definingUid=10154 user=0
+        process=org.chromium.chrome reason=2 (SIGNALED)
+        subreason=0 (UNKNOWN) status=9
+        importance=100 pss=0.00 rss=0.00 state=empty trace=null
+        description=null
+        anrInfo=null
+
 
   Args:
     app_package: Name of the application package.
--- a/src/clusterfuzz/_internal/platforms/android/util.py
+++ b/src/clusterfuzz/_internal/platforms/android/util.py
@@ -151,9 +151,71 @@ def get_latest_pid_for_package(app_package: str) -> int | None:
   return None
 
 
+def _to_enum(enum_cls, raw_value: int | str):
+  """Converts raw_value to an Enum member, or returns None if invalid."""
+  try:
+    return enum_cls(int(raw_value))
+  except (ValueError, TypeError):
+    return None
+
+
+def _parse_exit_info_from_dumpsys(dumpsys_output: str,
+                                  target_pid: int) -> ProcessExitInfo | None:
+  """Parses dumpsys activity exit-info output for target_pid.
+
+  Args:
+    dumpsys_output: Output text from `dumpsys activity exit-info`.
+    target_pid: Process ID to extract exit metadata for.
+
+  Returns:
+    ProcessExitInfo object if metadata for target_pid is found and parsed,
+    None otherwise.
+  """
+  if not dumpsys_output or target_pid is None:
+    return None
+
+  current_pid = None
+  for line in dumpsys_output.splitlines():
+    pid_match = re.search(r"\bpid=(\d+)", line)
+    if pid_match:
+      current_pid = int(pid_match.group(1))
+
+    if current_pid is None or current_pid != target_pid:
+      continue
+
+    reason_match = re.search(_REASON_STATUS_REGEX, line)
+    if not reason_match:
+      continue
+
+    reason, reason_name, subreason, subreason_name, status = (
+        reason_match.groups())
+
+    parsed_reason = _to_enum(constants.ExitReason, reason)
+    if parsed_reason is None:
+      logs.warning(f'[Android] Unexpected process exit reason code {reason} '
+                   f'for PID {target_pid}.')
+      parsed_reason = constants.ExitReason.UNKNOWN
+
+    parsed_status = _to_enum(constants.ExitStatus, status)
+    if parsed_status is None:
+      logs.warning(f'[Android] Unexpected process exit status code {status} '
+                   f'for PID {target_pid}.')
+      parsed_status = int(status)
+
+    return ProcessExitInfo(
+        reason=parsed_reason,
+        reason_name=reason_name or '',
+        subreason=int(subreason),
+        subreason_name=subreason_name or '',
+        status=parsed_status,
+    )
+
+  return None
+
+
 def get_exit_info_for_pid(app_package: str,
                           target_pid: int) -> ProcessExitInfo | None:
-  """Parses dumpsys activity exit-info output for target_pid.
+  """Fetches and parses dumpsys activity exit-info output for target_pid.
 
   Args:
     app_package: Name of the application package.
@@ -167,31 +229,7 @@ def get_exit_info_for_pid(app_package: str,
     return None
 
   dumpsys_output = adb.get_activity_exit_info(app_package)
-  if not dumpsys_output:
-    return None
-
-  current_pid = None
-  for line in dumpsys_output.splitlines():
-    pid_match = re.search(r"\bpid=(\d+)", line)
-    if pid_match:
-      current_pid = int(pid_match.group(1))
-
-    if current_pid is None or current_pid != target_pid:
-      continue
-
-    reason_match = re.search(_REASON_STATUS_REGEX, line)
-    if reason_match:
-      reason, reason_name, subreason, subreason_name, status = (
-          reason_match.groups())
-      return ProcessExitInfo(
-          reason=int(reason),
-          reason_name=reason_name or '',
-          subreason=int(subreason),
-          subreason_name=subreason_name or '',
-          status=int(status),
-      )
-
-  return None
+  return _parse_exit_info_from_dumpsys(dumpsys_output, target_pid)
 
 
 def activity_crashed(exit_info: ProcessExitInfo | None) -> bool:
--- a/src/clusterfuzz/_internal/platforms/android/util.py
+++ b/src/clusterfuzz/_internal/platforms/android/util.py
@@ -15,10 +15,25 @@
 
 from dataclasses import dataclass
 import os
+import re
 
 from clusterfuzz._internal.metrics import logs
 from clusterfuzz._internal.platforms import android
 from clusterfuzz._internal.system import environment
+
+from . import adb
+from . import constants
+from . import logger
+
+# Matching: "Start proc <PID>:<PROCESS_NAME>/"
+_START_PROC_REGEX = r"Start proc (\d+):(\S+?)/"
+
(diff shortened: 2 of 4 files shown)


Xeicker (approved):

src/clusterfuzz/_internal/platforms/android/adb.py
```diff
@@ -915,3 +915,17 @@ def write_data_to_file(contents, file_path, should_reboot=True):
     else:
       # Manually revert /system to read-only since we aren't rebooting.
       run_shell_command('mount -o ro,remount /system', root=True)
+
+
+def get_activity_exit_info(app_package: str) -> str:
+  """Get dumpsys activity exit-info output for the given application package.
```
dylanjew, line 921: do you have any documentation for what dumpsys output looks like?

src/clusterfuzz/_internal/platforms/android/util.py
```diff
@@ -102,3 +117,120 @@ def can_testcase_run_on_platform(testcase_platform_id, current_platform_id):
...
+  """Gets the latest PID for an application package from logcat.
+
+  Args:
+    app_package: Name of the target application package.
+
+  Returns:
+    PID of the package's latest process if found, None otherwise.
+  """
+  if not app_package:
+    return None
```
dylanjew: it might be helpful to log these cases. When would you expect `app_package` or `logcat_output` to be `None`?

src/clusterfuzz/_internal/platforms/android/util.py
```diff
@@ -102,3 +117,120 @@ def can_testcase_run_on_platform(testcase_platform_id, current_platform_id):
...
+
+  Args:
+    app_package: Name of the application package.
+    target_pid: Process ID to extract exit metadata for.
+
+  Returns:
+    ProcessExitInfo object if metadata for target_pid is found and parsed,
+    None otherwise.
+  """
+  if not app_package or target_pid is None:
```
dylanjew: same question here

src/clusterfuzz/_internal/platforms/android/util.py
```diff
@@ -102,3 +117,120 @@ def can_testcase_run_on_platform(testcase_platform_id, current_platform_id):
...
+    return None
+
+  dumpsys_output = adb.get_activity_exit_info(app_package)
+  if not dumpsys_output:
```
dylanjew: also probably worth logging

src/clusterfuzz/_internal/platforms/android/util.py
```diff
@@ -102,3 +117,120 @@ def can_testcase_run_on_platform(testcase_platform_id, current_platform_id):
...
+    if reason_match:
+      reason, reason_name, subreason, subreason_name, status = (
+          reason_match.groups())
+      return ProcessExitInfo(
+          reason=int(reason),
+          reason_name=reason_name or '',
+          subreason=int(subreason),
+          subreason_name=subreason_name or '',
+          status=int(status),
+      )
```
dylanjew: this would be good to pull into a separate function to unit test the regex parsing

src/clusterfuzz/_internal/platforms/android/util.py
```diff
@@ -102,3 +117,120 @@ def can_testcase_run_on_platform(testcase_platform_id, current_platform_id):
...
+
+
+  Args:
+    app_package: Name of the application package to check.
+
+  Returns:
+    True if the package's latest process crashed, False otherwise.
+  """
+  if not app_package:
+    return False
```
dylanjew, line 272: if you're checking for `app_package` here, you can make it required for the downstream functions and not have those defensive checks

src/clusterfuzz/_internal/platforms/android/adb.py
```diff
@@ -915,3 +915,17 @@ def write_data_to_file(contents, file_path, should_reboot=True):
...
+
+  Args:
+    app_package: Name of the application package.
+
+  Returns:
+    Dumpsys output for the application package.
+  """
+  dumpsys_output = run_shell_command(
```
dylanjew, line 941: how long is the dumpsys activity kept around? And is it written to immediately following a crash? 

This approach seems like it would work but it's a bit indirect

src/clusterfuzz/_internal/platforms/android/util.py
dylanjew: Your unit test is effectively doing this test, but it has to mock `get_activity_exit_info` to do so. Ideally your tests can mock as little as possible, which makes them less brittle.

but after looking through the tests and seeing you've covered testing this regex, this point is more of a nit

src/clusterfuzz/_internal/platforms/android/adb.py
IvanBM18: since dumpsys can have tons of information the documentation on its output(with examples*) is very scarce:
https://developer.android.com/tools/dumpsys
Official docs instead point you to the Java Class definition:
https://developer.android.com/reference/android/app/ActivityManager#getHistoricalProcessExitReasons(java.lang.String,%20int,%20int)
Which is not that useful if you ask me.. i can also add an example here, WDYT?

src/clusterfuzz/_internal/platforms/android/adb.py
IvanBM18: Is persistent in the disk, but it can be deleted using:
`adb shell am clear-exit-info [<PACKAGE>]`
`adb uninstall <app>`

src/clusterfuzz/_internal/platforms/android/util.py
IvanBM18: Sure, adding log!

the app's package should never be empty in a blackbox android fuzzing session, its catched very early into the session execution(top of my mind, in bad build check)

But i added this validations if someone ever calls this method in a CG fuzzing session.

IvanBM18: on the contrary logcat output shouldn't be empty if we are launching activities... but again in CG fuzzing this may not be the case, would have to monitor this changes for CG once they land

src/clusterfuzz/_internal/platforms/android/util.py
IvanBM18: Dito ^.

BUT worth noting that if a pkg doesn't have exit info it means that is still running

src/clusterfuzz/_internal/platforms/android/util.py
IvanBM18: Sure adding log!

src/clusterfuzz/_internal/platforms/android/util.py
IvanBM18: Done!, i also included logic for uknown exit codes as per one of your comments if the previous PR

src/clusterfuzz/_internal/platforms/android/util.py
IvanBM18: Sure thing, done!

src/clusterfuzz/_internal/platforms/android/adb.py
dylanjew: thanks, that example is helpful

src/clusterfuzz/_internal/platforms/android/util.py
```diff
@@ -36,6 +51,68 @@ class ProcessExitInfo:
   status: android.constants.ExitStatus | int
 
 
+def _to_enum(enum_cls, raw_value: int | str):
+  """Converts raw_value to an Enum member, or returns None if invalid."""
+  try:
+    return enum_cls(int(raw_value))
+  except (ValueError, TypeError):
+    return None
```
dylanjew, line 59: nit: add a log so that this doesn't silently lose the input

src/clusterfuzz/_internal/tests/core/platforms/android/util_test.py
```diff
@@ -0,0 +1,282 @@
...
+    exit_info = util.ProcessExitInfo(
+        reason=constants.ExitReason.SIGNALED,
+        reason_name='SIGNALED',
+        subreason=0,
+        subreason_name='',
+        status=constants.ExitStatus.SIGKILL,
+    )
+    self.assertTrue(util.activity_crashed(exit_info))
+
+  def test_crash_anr(self):
```
dylanjew: what does `test_crash_anr` mean?

dylanjew (commented):
LGTM, same question about checking with GOSST for the Android instance

src/clusterfuzz/_internal/platforms/android/util.py
IvanBM18: Adding it in: https://github.com/google/clusterfuzz/pull/5461 under the new debug level

src/clusterfuzz/_internal/tests/core/platforms/android/util_test.py
IvanBM18: A typo i made when renaming the tests, fixed!

src/clusterfuzz/_internal/platforms/android/util.py
```diff
@@ -36,6 +51,68 @@ class ProcessExitInfo:
...
+  Returns:
+    ProcessExitInfo object if metadata for target_pid is found and parsed,
+    None otherwise.
+  """
+  if not dumpsys_output or target_pid is None:
+    return None
+
+  current_pid = None
+  for line in dumpsys_output.splitlines():
+    pid_match = re.search(r"\bpid=(\d+)", line)
```
decoNR, line 79: nit: I was wondering if we could have a more robust way to parse the result instead of depending on regex. I did a quick research and did not find a simple way to do that, did you investigate other options?

src/clusterfuzz/_internal/platforms/android/util.py
```diff
@@ -36,6 +51,68 @@ class ProcessExitInfo:
...
+
+  Args:
+    dumpsys_output: Output text from `dumpsys activity exit-info`.
+    target_pid: Process ID to extract exit metadata for.
+
+  Returns:
+    ProcessExitInfo object if metadata for target_pid is found and parsed,
+    None otherwise.
+  """
+  if not dumpsys_output or target_pid is None:
```
decoNR, line 74: nit: Should we add a log here since this seems to not be expected?

decoNR (approved):
Please resolve the comments before merging.

I do not have access to the swarming test but since you tested it LGTM. This seems to be a nice improvement :)

decoNR:
> LGTM, same question about checking with GOSST for the Android instance

Just referencing the comment on the last PR here: https://github.com/google/clusterfuzz/pull/5453#issuecomment-5626325020

src/clusterfuzz/_internal/platforms/android/util.py
IvanBM18: Me neither, I investigated if adb could give us the output in a format thats more easily read/parsed(i.e json) but only found that `adb` can give you protobuf output for only this four sections:  
- activities
- broadcasts
- services
- processes.
But we use `exit-info` so no, we can have it on a parsed output.

src/clusterfuzz/_internal/platforms/android/util.py
IvanBM18: Sure thing!
Adding it under this PR that i have for adding debug logs: https://github.com/google/clusterfuzz/pull/5461

src/clusterfuzz/_internal/platforms/android/util.py
decoNR: I see, thank you!

AashutoshMurthy:
### Missing `if app_package:` guard causes log spam and unnecessary ADB overhead

In `run_process()`, every process execution on an Android bot now runs:
```python
app_package = android.app.get_package_name()
process_pid = android.util.get_latest_pid_for_package(app_package)
exit_info = android.util.get_exit_info_for_pid(app_package, process_pid)

if android.util.activity_crashed(exit_info):
  ...
```

When `app_package` is `None` (which happens on all non-APK Android jobs and during general setup/warmup checks):
1. `get_latest_pid_for_package(None)` executes `adb logcat -d -s ActivityManager:I`, dumping the entire system ActivityManager log over ADB and iterating through every line matching against literal strings like `process_name.startswith('None:')`.
2. `get_exit_info_for_pid(None, None)` and `activity_crashed(None)` emit spurious warnings on every run:
   ```
   [Android][None] Exit info not found, PID not given
   [Android] Exit info empty, not checking for crashes.
   ```
3. This adds unnecessary ADB round-trip latency and log noise to every process invocation when no package is configured.

#### Suggested Fix
Add a guard so exit info is only queried when `app_package` is present:

```python
    app_package = android.app.get_package_name()
    exit_info = None
    if app_package:
      process_pid = android.util.get_latest_pid_for_package(app_package)
      exit_info = android.util.get_exit_info_for_pid(app_package, process_pid)

    if exit_info and android.util.activity_crashed(exit_info):
      logs.warning(f'Activity Crashed with: {exit_info}')
      return_code = exit_info.reason
```

PR #5455, opened 2026-09-03
[Android] Integrate exit code handling into process_handler

IvanBM18:
Bug: b/553141628

## Overview
Since Android API level 30 (and apps targeting Android 11+), apps have scoped storage access. `am start` process execution does not report exit codes on activity crashes, returning 0 even when native crashes occur. 

This PR integrates Android activity crash detection and reboot tracking into `process_handler.run_process()`, ensuring process return codes accurately reflect native crashes (e.g. SIGSEGV, SIGABRT) or device reboots at test case executions.  We also correct an intermittent issue in which the time since the last reboot vs time since activity launched was calculated incorrectly.

## Changes
- `src/clusterfuzz/_internal/system/process_handler.py`: Integrated `activity_crashed_by_package()` check and reboot validation in `run_process()` for Android.
- `src/clusterfuzz/_internal/tests/core/system/process_handler_test.py`: Added unit tests for Android activity crash and clean exit handling in `run_process()`.

```
```

## PR stack
- `master`
  - **#PR 2.2** `feature/android-exit-code-process-handler` 👈

### DIFF
--- a/src/clusterfuzz/_internal/system/process_handler.py
+++ b/src/clusterfuzz/_internal/system/process_handler.py
@@ -174,6 +174,7 @@ def run_process(cmdline,
   if is_android:
     # Clear the log upfront.
     android.logger.clear_log()
+    initial_uptime = android.adb.time_since_last_reboot()
 
     # Run the app.
     adb_output = android.adb.run_command(
@@ -253,15 +254,26 @@ def run_process(cmdline,
     # waits for device to be online.
     time.sleep(ANDROID_CRASH_LOGCAT_WAIT_TIME)
     output = android.logger.log_output()
+    app_package = android.app.get_package_name()
+    process_pid = android.util.get_latest_pid_for_package(app_package)
+    exit_info = android.util.get_exit_info_for_pid(app_package, process_pid)
 
-    if android.constants.LOW_MEMORY_REGEX.search(output):
+    if android.util.activity_crashed(exit_info):
+      logs.warning(f'Activity Crashed with: {exit_info}')
+      return_code = exit_info.reason
+
+    elif android.constants.LOW_MEMORY_REGEX.search(output):
       # If the device is low on memory, we should force reboot and bail out to
       # prevent device from getting in a frozen state.
       logs.info('Device is low on memory, rebooting.', output=output)
       android.adb.hard_reset()
       android.adb.wait_for_device()
 
-    elif android.adb.time_since_last_reboot() < time.time() - start_time:
+    elif android.adb.time_since_last_reboot() < initial_uptime:
+      logs.info(
+          'Device rebooted mid-run',
+          output=f'initial uptime: {initial_uptime}, '
+          f'current uptime: {android.adb.time_since_last_reboot()}')
       # Check if a reboot has happened, if yes, append log output before reboot
       # and kernel logs content to output.
       log_before_last_reboot = android.logger.log_output_before_last_reboot()
--- a/src/clusterfuzz/_internal/system/process_handler.py
+++ b/src/clusterfuzz/_internal/system/process_handler.py
@@ -345,6 +345,11 @@ def run_process(cmdline,
 
   if testcase_run and (crash_analyzer.is_memory_tool_crash(output) or
                        crash_analyzer.is_check_failure_crash(output)):
+    if return_code:
+      logs.warning(
+          f'Process {cmdline!r} ended with exit code {return_code}, '
+          'but crash has been detected, overriding return code to 1.',
+          output=output)
     return_code = 1
 
   # If a crash is found, then we add the memory state as well.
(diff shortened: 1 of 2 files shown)


src/clusterfuzz/_internal/system/process_handler.py
```diff
@@ -253,15 +254,26 @@ def run_process(cmdline,
...
+    app_package = android.app.get_package_name()
+    process_pid = android.util.get_latest_pid_for_package(app_package)
+    exit_info = android.util.get_exit_info_for_pid(app_package, process_pid)
 
-    if android.constants.LOW_MEMORY_REGEX.search(output):
+    if android.util.activity_crashed(exit_info):
+      logs.warning(f'Activity Crashed with: {exit_info}')
+      return_code = exit_info.reason
+
+    elif android.constants.LOW_MEMORY_REGEX.search(output):
```
Xeicker, line 265: nit: I would use exit reason low memory. I'd consider that a bit more consistent

Xeicker (commented):
I don't quite get why it is an issue that the uptime is calculated incorrectly and how it is related to the returning codes issue.
Not that it shouldn't be fixed, just curious why to integrate it in the stacked PRs

Xeicker (approved):

src/clusterfuzz/_internal/tests/core/system/process_handler_test.py
```diff
@@ -149,3 +149,53 @@ def test_process_1_no_terminate_with_wrong_case(self):
...
+    self.mock.platform.return_value = 'ANDROID'
+    self.mock.get_package_name.return_value = 'com.example.app'
+    self.mock.time_since_last_reboot.return_value = 100.0
+    self.mock.log_output.return_value = ''
+    self.mock.run_command.return_value = ''
+    self.mock.get_ps_output.return_value = ''
+
+  def test_run_process_android_activity_crashed(self):
+    """Checks that run_process sets return_code from exit_info.reason and logs warning when an Android activity crash is detected."""
+    exit_info = mock.Mock(reason=5, status=11)
```
dylanjew: is there a reason to mock the `ProcessExitInfo` instead of just creating a real one?

dylanjew (commented):
LGTM

src/clusterfuzz/_internal/tests/core/system/process_handler_test.py
IvanBM18: aa not really... moving it to a process exit info type

src/clusterfuzz/_internal/system/process_handler.py
IvanBM18: Currently when memory on the device is low, CF performs a reset on the device and then continues the execution without marking the test case or the bad build check as a crash, returning a non 0 exit reason here would change that, so i don't think this one is feasable.

IvanBM18:
> I don't quite get why it is an issue that the uptime is calculated incorrectly and how it is related to the returning codes issue. Not that it shouldn't be fixed, just curious why to integrate it in the stacked PRs

Answering your 'uptime calculation question' When testing this changes i performed quite a ton of tests, and i saw that intermittently CF marked the test case as a failure because of this, even though i was seeing the test case execution directly on the avd's screen, and since the fix was a small pretty straight forward one i figured that it might fit into this small pr of changes in this stack

src/clusterfuzz/_internal/system/process_handler.py
```diff
@@ -253,15 +254,26 @@ def run_process(cmdline,
...
     time.sleep(ANDROID_CRASH_LOGCAT_WAIT_TIME)
     output = android.logger.log_output()
+    app_package = android.app.get_package_name()
+    process_pid = android.util.get_latest_pid_for_package(app_package)
+    exit_info = android.util.get_exit_info_for_pid(app_package, process_pid)
 
-    if android.constants.LOW_MEMORY_REGEX.search(output):
+    if android.util.activity_crashed(exit_info):
+      logs.warning(f'Activity Crashed with: {exit_info}')
+      return_code = exit_info.reason
```
decoNR, line 263: Just to confirm, it seems that return_code can be overwritten, could this cause any issues?

decoNR (approved):
Please resolve comments before merging.

src/clusterfuzz/_internal/system/process_handler.py
IvanBM18: Other than bringing more clarity it shouldn't bring any issues, actually in this method we usually change the return_code to tell callers if the invoked process crashed, see example:
- https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/system/process_handler.py#L276

decoNR: What I meant is that the value assigned to return_code here can be overwritten later in the function.

I just wanted to flag this to confirm if it is not a problem.

IvanBM18: You are right, i see it can be overwritten to
```
return_code = 1
```
If there was a memory crash... this seems very delicate so I'll add a log there to note the dev that the return_code was modified midflight.
Thanks for catching that!

PR #5456, opened 2026-09-03
[Android] Refine bad build check and testcase manager execution

IvanBM18:
Bug: b/553141628

## Overview
Since Android API level 30 (and apps targeting Android 11+), apps have scoped storage access. Previously, bad build checks for Android did not verify whether the application process actually survived startup because `am start` returns 0 even if the process dies immediately. 

This PR enhances `check_for_bad_build()` in `testcase_manager.py` by ensuring command line files are written prior to execution and flipping a conditional so in the bad build check the android specific validation proceeds other validations.

The later was required because thanks to [this PR](https://github.com/google/clusterfuzz/commit/733e28f67fc9997db15f8051d57a91f2ac281260?diff=unified) we added the possibility to check for false positives, but this added a problem in which valid builds were incorrectly flagged as crashes, so by flipping the conditional we fix this.

## Changes
- `src/clusterfuzz/_internal/bot/testcase_manager.py`: Updated `check_for_bad_build()` to pass `write_command_line_file=True` and verify target application process presence on Android post-launch.

```
```

## PR stack
- `master`
  - **#PR 2.3** `feature/android-bad-build-check` 👈
  
 ## Note:
  Adding additional debug logs in this other PR:
  - https://github.com/google/clusterfuzz/pull/5461

### DIFF
--- a/src/clusterfuzz/_internal/bot/testcase_manager.py
+++ b/src/clusterfuzz/_internal/bot/testcase_manager.py
@@ -1323,19 +1323,6 @@ def check_for_bad_build(job_type: str,
         f'return code = {return_code}, crash type = {crash_result.get_type()}',
         raw_output=output,
         output=build_run_console_output)
-  elif environment.is_android():
-    package_name = android.app.get_package_name()
-    if (package_name and
-        not android.adb.get_process_and_child_pids(package_name)):
-      is_bad_build = True
-      build_run_console_output = utils.get_crash_stacktrace_output(
-          command, output, output)
-      logs.info(
-          f'Bad build for {job_type} detected at r{crash_revision}: '
-          f'application process for {package_name} is not running after '
-          'startup.',
-          raw_output=output,
-          output=build_run_console_output)
 
   # Exit all running instances.
   process_handler.terminate_stale_application_instances()
--- a/src/clusterfuzz/_internal/bot/testcase_manager.py
+++ b/src/clusterfuzz/_internal/bot/testcase_manager.py
@@ -332,7 +332,12 @@ def run_testcase(thread_index, file_path, gestures, env_copy):
     app_directory = environment.get_value('APP_DIR')
     environment.set_value('PIDS', '[]')
 
+    logs.info(
+        f'Running testcase (thread {thread_index}): file_path={file_path}, '
+        f'needs_http={needs_http}, gestures={gestures}')
+
     # Get command line options.
+
     command = get_command_line_for_application(
         file_path, user_profile_index=thread_index, needs_http=needs_http)
 
@@ -1243,7 +1248,8 @@ def check_for_bad_build(job_type: str,
     os.environ['APP_ARGS'] = job_default_args
 
   try:
-    command = get_command_line_for_application(file_to_run='', needs_http=False)
+    command = get_command_line_for_application(
+        file_to_run='', needs_http=False, write_command_line_file=True)
   finally:
     if orig_app_args is not None:
       os.environ['APP_ARGS'] = orig_app_args
@@ -1273,19 +1279,40 @@ def check_for_bad_build(job_type: str,
   process_handler.terminate_stale_application_instances()
 
   # Check if the build is bad.
+  logs.info(
+      f'Starting bad build check for {job_type} at r{crash_revision} with '
+      f'command: {command} (timeout={fast_warmup_timeout})')
   return_code, crash_time, output = process_handler.run_process(
       command,
       timeout=fast_warmup_timeout,
       current_working_directory=app_directory)
   crash_result = CrashResult(return_code, crash_time, output)
+  logs.info(
+      f'Bad build check run_process completed: return_code={return_code}, '
+      f'is_crash={crash_result.is_crash(ignore_state=True)}, '
+      f'crash_type={crash_result.get_type()}')
 
   # 1. Need to account for startup crashes with no crash state. E.g. failed to
   #    load shared library. So, ignore state for comparison.
   # 2. Ignore leaks as they don't block a build from reporting regular crashes
   #    and also don't impact regression range calculations.
-  if (crash_result.is_crash(ignore_state=True) and
-      not crash_result.should_ignore() and
-      not crash_result.get_type() in ['Direct-leak', 'Indirect-leak']):
+  if environment.is_android():
+    package_name = android.app.get_package_name()
+    if (package_name and
+        not android.adb.get_process_and_child_pids(package_name)):
+      is_bad_build = True
+      build_run_console_output = utils.get_crash_stacktrace_output(
+          command, output, output)
+      logs.info(
+          f'Bad build for {job_type} detected at r{crash_revision}: '
+          f'application process for {package_name} is not running after '
+          'startup.',
+          raw_output=output,
+          output=build_run_console_output)
+  elif (crash_result.is_crash(ignore_state=True) and
+        not crash_result.should_ignore() and
+        not crash_result.get_type() in ['Direct-leak', 'Indirect-leak']):
+
     is_bad_build = True
     build_run_console_output = utils.get_crash_stacktrace_output(
         command,
--- a/src/clusterfuzz/_internal/bot/testcase_manager.py
+++ b/src/clusterfuzz/_internal/bot/testcase_manager.py
@@ -1296,6 +1296,8 @@ def check_for_bad_build(job_type: str,
   #    load shared library. So, ignore state for comparison.
   # 2. Ignore leaks as they don't block a build from reporting regular crashes
   #    and also don't impact regression range calculations.
+  # 3. On Android, if the application process is not running after startup,
+  #    the build is bad.
   if environment.is_android():
     package_name = android.app.get_package_name()
     if (package_name and
--- a/src/clusterfuzz/_internal/bot/testcase_manager.py
+++ b/src/clusterfuzz/_internal/bot/testcase_manager.py
@@ -1292,12 +1292,8 @@ def check_for_bad_build(job_type: str,
       f'is_crash={crash_result.is_crash(ignore_state=True)}, '
       f'crash_type={crash_result.get_type()}')
 
-  # 1. Need to account for startup crashes with no crash state. E.g. failed to
-  #    load shared library. So, ignore state for comparison.
-  # 2. Ignore leaks as they don't block a build from reporting regular crashes
-  #    and also don't impact regression range calculations.
-  # 3. On Android, if the application process is not running after startup,
-  #    the build is bad.
+  # On Android, if the application process is not running after startup, the
+  # build is bad.
   if environment.is_android():
     package_name = android.app.get_package_name()
     if (package_name and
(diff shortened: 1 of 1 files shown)


src/clusterfuzz/_internal/bot/testcase_manager.py
```diff
@@ -1273,19 +1279,40 @@ def check_for_bad_build(job_type: str,
...
+      f'crash_type={crash_result.get_type()}')
 
   # 1. Need to account for startup crashes with no crash state. E.g. failed to
   #    load shared library. So, ignore state for comparison.
   # 2. Ignore leaks as they don't block a build from reporting regular crashes
   #    and also don't impact regression range calculations.
-  if (crash_result.is_crash(ignore_state=True) and
-      not crash_result.should_ignore() and
-      not crash_result.get_type() in ['Direct-leak', 'Indirect-leak']):
+  if environment.is_android():
```
Xeicker, line 1301: The above comment doesn't seem to reflect the android logic. Could you add a comment on the android approach?
Also it seems like android is not able to detect the same crash types as other platforms?

Xeicker (changes requested):

src/clusterfuzz/_internal/bot/testcase_manager.py
IvanBM18: Sure, just added comments for this :D

Answering your question, yeah android pretty much used to rely on log regex matching to determine bad builds this proved to be very inconsistent and failed to flag some invalid builds when they were actually valid.

In this bug: b/546076952 paulo solved it by instead checking wether the app is running or not after the warm up command in line:1286.

Xeicker (approved):

src/clusterfuzz/_internal/bot/testcase_manager.py
```diff
@@ -332,7 +332,12 @@ def run_testcase(thread_index, file_path, gestures, env_copy):
     app_directory = environment.get_value('APP_DIR')
     environment.set_value('PIDS', '[]')
 
+    logs.info(
```
dylanjew, line 335: are these useful or are they more debug logs?

src/clusterfuzz/_internal/bot/testcase_manager.py
```diff
@@ -1273,30 +1279,26 @@ def check_for_bad_build(job_type: str,
...
   crash_result = CrashResult(return_code, crash_time, output)
+  logs.info(
+      f'Bad build check run_process completed: return_code={return_code}, '
+      f'is_crash={crash_result.is_crash(ignore_state=True)}, '
+      f'crash_type={crash_result.get_type()}')
 
   # 1. Need to account for startup crashes with no crash state. E.g. failed to
   #    load shared library. So, ignore state for comparison.
   # 2. Ignore leaks as they don't block a build from reporting regular crashes
   #    and also don't impact regression range calculations.
```
dylanjew: could you move this comment back to the code it describes?

dylanjew (commented):
LGTM. sorry, forgot to send my comments earlier

src/clusterfuzz/_internal/bot/testcase_manager.py
IvanBM18: This ones are debug logs in:
https://github.com/google/clusterfuzz/pull/5461

src/clusterfuzz/_internal/bot/testcase_manager.py
IvanBM18: Sure, done!

src/clusterfuzz/_internal/bot/testcase_manager.py
```diff
@@ -1243,7 +1248,8 @@ def check_for_bad_build(job_type: str,
     os.environ['APP_ARGS'] = job_default_args
 
   try:
-    command = get_command_line_for_application(file_to_run='', needs_http=False)
+    command = get_command_line_for_application(
```
decoNR, line 1251: What exactly does this modification change in the behavior, and why is it not harmful? Asking to get more context.

decoNR (approved):
Overall LGTM. Left a quick question, please answer it before merging.

Were you able to test it in dev to guarantee that this does not affect other platforms? I think this seems important since one of the changes is not specific to Android.

src/clusterfuzz/_internal/bot/testcase_manager.py
IvanBM18: Sure thing!

You can add in your job definition app arguments e.g. from chrome job:
```
APP_ARGS = --disable-fre --no-first-run --no-default-browser-check --enable-test-intents --disable-gpu-watchdog --disable-notifications --deny-permission-prompts --disable-popup-blocking --enable-logging=stderr --v=1 --no-experiments --allow-file-access-from-files --use-fake-device-for-media-stream --use-fake-ui-for-media-stream --autoplay-policy=no-user-gesture-required --disable-prompt-on-repost
```
Which are used at every single test case execution. Without this args the app fails to launch or our fuzzing fails, or we have unexpected results.

The thing is that  you would expect to see them like this:
```
$ command <APP_ARGS> <testcase>
```
But in android, you have to save this args in a special file, defined at the job:
```
COMMAND_LINE_PATH = /data/local/tmp/chrome-command-line
```
So by adding the `write_command_line_file=True` we ensure that the args are used at the bad build check.

Why are they required for the bad build check? well thats because the bad build check executes the same command that the fuzzing sessions execute but with an empty file `file_to_run=''`

IvanBM18:
> Overall LGTM. Left a quick question, please answer it before merging.
> 
> Were you able to test it in dev to guarantee that this does not affect other platforms? I think this seems important since one of the changes is not specific to Android.

Sure no worries, that specific argument in the
```
get_command_line_for_application()
```
Is only used on android, see:
- https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/testcase_manager.py#L1137

decoNR:
> > Overall LGTM. Left a quick question, please answer it before merging.
> > Were you able to test it in dev to guarantee that this does not affect other platforms? I think this seems important since one of the changes is not specific to Android.
> 
> Sure no worries, that specific argument in the
> 
> ```
> get_command_line_for_application()
> ```
> 
> Is only used on android, see:
> 
> * https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/testcase_manager.py#L1137

Perfect, thanks.

src/clusterfuzz/_internal/bot/testcase_manager.py
decoNR: Thank you for the explanation!

AashutoshMurthy:
### Control-flow issue: `elif` is unreachable on Android

Because `if environment.is_android():` is now placed first in the `if/elif` chain, Python unconditionally enters the first branch on any Android platform, which means the `elif (crash_result.is_crash(...)):` branch is never evaluated.

This causes two regressions on Android:
1. **Non-APK Android jobs (`package_name is None`):**
   The inner check `if (package_name and ...)` evaluates to `False`. Since the `elif` is skipped, `is_bad_build` is never set to `True`—even if the native binary or script crashed on launch with a SIGSEGV or ASan violation.
2. **Symbolized stack traces are bypassed for APKs:**
   Even when `package_name` is set and an app crashes on launch, bypassing the `elif` skips `crash_result.get_stacktrace(symbolized=True)`. The bot only checks if the PID is dead and logs raw, un-symbolized console output instead of the symbolized crash stack.

#### Suggested Fix
Evaluate the standard `crash_result.is_crash()` check first (so symbolized stacks are captured for any crash), and use the Android PID check as a fallback when no crash stack was caught:

```python
  if (crash_result.is_crash(ignore_state=True) and
      not crash_result.should_ignore() and
      not crash_result.get_type() in ['Direct-leak', 'Indirect-leak']):
    is_bad_build = True
    build_run_console_output = utils.get_crash_stacktrace_output(
        command,
        crash_result.get_stacktrace(symbolized=True),
        crash_result.get_stacktrace(symbolized=False))
    logs.info(...)
  elif environment.is_android():
    package_name = android.app.get_package_name()
    if (package_name and
        not android.adb.get_process_and_child_pids(package_name)):
      is_bad_build = True
      ...
```

AashutoshMurthy:
### Issues when `PKG_NAME` is unset & directory cleanup

#### 1. Malformed path (`/sdcard/Android/data//files`) when `PKG_NAME` is not set
When `PKG_NAME` is not set (e.g., non-APK jobs, or before an APK is identified), `get_package_name()` returns `None` and the fallback `''` produces `/sdcard/Android/data//files`. This causes two problems:

* **Host string replacement mismatch (`FileNotFoundError`):**
  In `testcase_manager.py:464` (`convert_dependency_url_to_local_path`), ClusterFuzz maps device paths back to the host using literal substring matching:
  ```python
  remote_dir = android.app.get_testcases_directory()  # '/sdcard/Android/data//files'
  local_path = local_path.replace(remote_dir, local_inputs_dir)
  ```
  If a browser or crash logger normalizes the path in logs to a single slash (`file:///sdcard/Android/data/files/test.html`), `"data//files"` fails to match `"data/files"`. `replace()` does nothing, leaving a device path on the host bot and causing a `FileNotFoundError`.
* **Breaks Android Scoped Storage permissions:**
  On device, the double slash resolves to a rogue folder named `files` directly under `/sdcard/Android/data/` (where Android expects `<package_name>` directories). Under Android 11+ Scoped Storage, the FUSE daemon restricts non-root apps to their own `/sdcard/Android/data/<package_name>` directory. Because no app owns `files`, non-root apps cannot access it (`EACCES`), breaking the exact permission model this PR was created to support.

#### 2. Incomplete cleanup with `rm -rf /*`
In `device.py:clear_testcase_directory()`, changing `adb.remove_directory(DIR, recreate=True)` to `rm -rf {app.get_testcases_directory()}/*` preserves the root directory and relies on shell wildcard expansion. Because `*` in POSIX shell does not match leading dots, any hidden files/directories (dotfiles) created during runs will be silently ignored, and orphaned folders (like `/sdcard/Android/data//files`) will persist across runs.

#### Suggested Fix
1. Fall back to a valid default directory when `package_name` is missing:
```python
  package_name = get_package_name()
  if not package_name:
    return '/sdcard/fuzzer-testcases'
  return f'/sdcard/Android/data/{package_name}/files'
```
2. For cleanup in `clear_testcase_directory()`, delete contents including hidden files (e.g., `find {dir} -mindepth 1 -delete`) so dotfiles are not left behind.

IvanBM18:
Hi @AashutoshMurthy 

Thanks for your comments, assumed all android fuzzing used apks... so i'll modify the code accordingly to this and your other PR comments.


Although, we can't fully revert the conditional because as [this line says](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/crash_analysis/crash_analyzer.py#L209) the crash analysis will mark a working chrome build as if it was crashing, this issue although old, is still happening.
 
An alternative to this is to is modify the bad build check, so that we only taking it into account if theres an apk present:
```python
  # On Android, if we have an APK package and the application process is not
  # running after startup, the build is bad.
  if (environment.is_android() and
      (package_name := android.app.get_package_name()) and
      not android.adb.get_process_and_child_pids(package_name)):
    is_bad_build = True
    build_run_console_output = utils.get_crash_stacktrace_output(
        command, output, output)
    logs.info(
        f'Bad build for {job_type} detected at r{crash_revision}: '
        f'application process for {package_name} is not running after '
        'startup.',
        raw_output=output,
        output=build_run_console_output)
  # 1. Need to account for startup crashes with no crash state. E.g. failed to
  #    load shared library. So, ignore state for comparison.
  # 2. Ignore leaks as they don't block a build from reporting regular crashes
  #    and also don't impact regression range calculations.
  elif (crash_result.is_crash(ignore_state=True) and
```
This way both apk and non apk android fuzzing should check for crashes accordingly at the bad build check.

PR #5458, opened 2026-09-04
Fix untrusted remote batch fuzz tasks to pass signed URLs

dylanjew:
Ensure tworkers create signed URLs for fuzz tasks by changing `is_remote_utask` to evaluate fuzz tasks specifically based on architecture.

Blackbox fuzzers on linux which require data bundles like `inferno_webbot` and `lokihardt_jshitter` are broken because they can't access their data bundles. Before the batch migration was complete, they fetched it as trusted linux bots through gsutil. Now, the tworker doesn't generate a signed URL because the fuzz task is evaluating to UtaskLocalExecutor and so it is skipped, leaving the untrusted linux batch bot without the signed URLs for the data bundle

This is somewhat hardcoded to fuzz tasks, but the execution of fuzz tasks is unique. 

Fixes b/556617562

## Explanation

1. tworker picks up fuzz task. [get_command_object()](https://github.com/google/clusterfuzz/blob/3a24291fbd47ba627f20d7703aaa89c23d808f1c/src/clusterfuzz/_internal/bot/tasks/commands.py#L201) -> resolves to UTask
2. Tworker preprocess checks whether it should create a signed URL. `setup.preprocess_get_data_bundles()` -> `corpus_manager.get_proto_data_bundle_corpus()` -> `task_main_runs_on_uworker()`-> `is_remote_utask('fuzz')`
    reads `COMMAND_TYPES= UTaskLocalExecutor` -> `False`
3. "Not getting signed data bundle URLs" and `corpus_urls = []`
4. Batch uworker runs utask_main and fetches data bundles: `sync_data_bundle_corpus_to_disk()` -> `is_uworker()` -> "signed-URL path"
5. `download_signed_urls([])`: empty data bundle dir -> `IndexError: Cannot choose from an empty sequence`

## Testing
Testeed in dev


src/clusterfuzz/_internal/tests/core/bot/tasks/task_types_test.py
```diff
@@ -62,6 +63,129 @@ def test_trusted(self):
...
+        'clusterfuzz._internal.batch.service.is_remote_task', return_value=True)
+    environment.set_value('TWORKER', True)
+
+    with mock_remotely_executing, mock_is_remote_task:
+      result = task_types.is_remote_utask('fuzz', 'linux_asan_chrome')
+
+    self.assertTrue(result)
+
+  def test_linux_fuzz_tworker_swarming(self):
+    """Tests that on an orchestration tworker, is_remote_utask returns False for
```
dylanjew, line 91: this confuses me but is the current behavior of the code. I guess swarming relies on different checks?

dylanjew:
Assigning to the sheriff, but if anyone on your team has more context here outside of Javan feel free to reassign. I'm not too familiar with the architecture here

dylanjew:
For testing in dev, do you know what types of bots we have? Are they all remote batch? I'm not sure whether we have any trusted long running bots in dev

dylanjew:
> For testing in dev, do you know what types of bots we have? Are they all remote batch? I'm not sure whether we have any trusted long running bots in dev

I don't see any batch vms in [clusterfuzz-development](https://pantheon.corp.google.com/compute/instanceTemplates/list?e=-13802955&mods=logs_tg_prod&project=clusterfuzz-development)

And then in the logs, I looked at the [logs](https://cloudlogging.app.goo.gl/e1tKhYN13nZR28r7A) for a`ttekett_dom_fuzzer` which ran in `clusterfuzz-linux-pre-regional-nmkg`. preprocess showed `Not getting signed data bundle URLs.` which I think is because we didn't migrate dev to batch to match prod.

There's also a bunch of idle `clusterfuzz-linux` bots in dev, unlike in prod where there aren't any and there are bots starting with `j-` which is batch

dylanjew:
Ok, the batch bots in dev were down due to the pubsub topics being down. So I can see batch bots in dev picking up the fuzz task.

however, there's still an issue: [logs](https://cloudlogging.app.goo.gl/xXDZbDmtiYeZSHU88) show that the signed URLs are still not getting passed into the uworker and the databundle is attempting to take the normal path and not untrusted path. 

I'm going to try to see what's wrong with this fix

dylanjew:
I found some [logs](https://cloudlogging.app.goo.gl/tYz8Styai31J6jqZ8) where a tworker did successfully pass the signed URL to a batch bot. I think the batch bots that weren't getting the URLs in dev are due to the inconsistent configuration

dylanjew:
I've updated b/556617562 with more notes. but the fuzzers which couldn't generate any testcases in dev that I was seeing was the result of a separate issue with the signed URL path where we rename the files.

decoNR:
> > For testing in dev, do you know what types of bots we have? Are they all remote batch? I'm not sure whether we have any trusted long running bots in dev
> 
> I don't see any batch vms in [clusterfuzz-development](https://pantheon.corp.google.com/compute/instanceTemplates/list?e=-13802955&mods=logs_tg_prod&project=clusterfuzz-development)
> 
> And then in the logs, I looked at the [logs](https://cloudlogging.app.goo.gl/e1tKhYN13nZR28r7A) for a`ttekett_dom_fuzzer` which ran in `clusterfuzz-linux-pre-regional-nmkg`. preprocess showed `Not getting signed data bundle URLs.` which I think is because we didn't migrate dev to batch to match prod.
> 
> There's also a bunch of idle `clusterfuzz-linux` bots in dev, unlike in prod where there aren't any and there are bots starting with `j-` which is batch

Hi , I just looked and there are some starting with j- , which have batch labels. Not sure if I am missing something.

decoNR:
> > > For testing in dev, do you know what types of bots we have? Are they all remote batch? I'm not sure whether we have any trusted long running bots in dev
> > 
> > 
> > I don't see any batch vms in [clusterfuzz-development](https://pantheon.corp.google.com/compute/instanceTemplates/list?e=-13802955&mods=logs_tg_prod&project=clusterfuzz-development)
> > And then in the logs, I looked at the [logs](https://cloudlogging.app.goo.gl/e1tKhYN13nZR28r7A) for a`ttekett_dom_fuzzer` which ran in `clusterfuzz-linux-pre-regional-nmkg`. preprocess showed `Not getting signed data bundle URLs.` which I think is because we didn't migrate dev to batch to match prod.
> > There's also a bunch of idle `clusterfuzz-linux` bots in dev, unlike in prod where there aren't any and there are bots starting with `j-` which is batch
> 
> Hi , I just looked and there are some starting with j- , which have batch labels. Not sure if I am missing something.

Just saw your other comment, ignore the above one :)

src/clusterfuzz/_internal/bot/tasks/task_types.py
```diff
@@ -105,20 +105,37 @@ def is_no_privilege_workload(command, job):
...
+  provision resources via signed URLs."""
   if environment.is_uworker():
     # Return True even if we can't query the db.
     return True
 
+  task_cls = COMMAND_TYPES[command]
+  if command == 'fuzz' and environment.is_tworker():
+    # Fuzz tasks run locally on pooled bots (as UTaskLocalExecutor), but on
+    # orchestration tworkers they are escalated to remote UTasks.
+    is_remote = UTask.is_execution_remote(command)
```
decoNR, line 121: Since this relies on the hardcoded `UTask` rather than environment data, it seems hacky and likely to get out of sync over time. Is there another way we could check this?

dylanjew: I agree, it is hacky :(

But it's also the reality of how fuzz task works and is currently out of sync with fuzz task execution. Fuzz task is broken on batch right now because it's defined as a `UtaskLocalExecutor` rather than a `UTask`, since it needs to support privileged long lived bots + untrusted remote Batch/Swarming. I made this failure clearer in the description.

I thought about trying to make it less hardcoded, but only `fuzz` operates in this way, so hardcoding it to match the only edge case seems like the most practical way to fix this and highlight the inconsistency

I think the long term solution is closer to https://github.com/google/clusterfuzz/pull/5412, where we classify `'fuzz': UTask`. Even then we still have a check on both sides of the utask boundary to determine whether we should generate and fetch signed URLs. Maybe the long term solution is for the long lived bots and remote bots to have separate queues.

src/clusterfuzz/_internal/tests/core/bot/tasks/task_types_test.py
decoNR: I am not sure if I understood your comment but am a bit confused because while the test name has "tworker_swarming", the test description states "if the task is NOT targeting remote batch or swarming." Is this correct?

src/clusterfuzz/_internal/bot/tasks/task_types.py
decoNR: I see, thanks for the explanation. Wdyt about creating an issue to track the long term solution which would include making this hardcoded part more robust? Not sure if you are already thinking of maintaining the current bug open due to the long term solution.

decoNR (approved):
Although I do not have much context, LGTM. The tests you conducted and described in the issue seem very comprehensive.

Just to confirm: comment #7 in the issue mentions that the previous solution broke the trusted bots. Were you able to verify that this change is safe in that regard?

src/clusterfuzz/_internal/tests/core/bot/tasks/task_types_test.py
dylanjew: I renamed this and added some comments! the previous name was wrong

src/clusterfuzz/_internal/bot/tasks/task_types.py
dylanjew: great idea! I created b/564495173 to track it, since I do plan to close out the existing bug once things are working.

dylanjew (commented):
thanks!

src/clusterfuzz/_internal/tests/core/bot/tasks/task_types_test.py
decoNR: Thank you!

src/clusterfuzz/_internal/bot/tasks/task_types.py
decoNR: Nice, thanks!

PR #5460, opened 2026-09-07
[Debug Logs] Enables debug level logs

IvanBM18:
## Overview

As time passes by we have struggled to keep our logs clean, this has been more clear in our issue with too many error logs, which.

As a first step to make logs cleaner im enabling debug level logging worlwide across all loggers(swarming, local, batch, k8s)

## Changes
- Adds the `debug` logging level to our `logs.py` module.
- Set the new base logging level from `logging.INFO `, think of it as the minimum supported logging level, without it our loggers would filter out DEBUG calls.
- Adds the ability for one to setup the log level with env vars

## Test performed
Added a simple log to the task loop in this commit:https://github.com/google/clusterfuzz/pull/5460/commits/e61947c9c54b5b3478299380fa551df87080b808, to confirm that its being logged with the correct severity:
image

https://cloudlogging.app.goo.gl/H1uBTcsR8oY14Vur8

### DIFF
--- a/src/clusterfuzz/_internal/metrics/logs.py
+++ b/src/clusterfuzz/_internal/metrics/logs.py
@@ -53,6 +53,8 @@ _logger = None
 _is_already_handling_uncaught = False
 _default_extras = {}
 
+BASE_LOGGING_LEVEL = logging.DEBUG
+
 
 def _increment_error_count():
   """"Increment the error count metric."""
@@ -115,7 +117,7 @@ def get_handler_config(filename, backup_count):
 
   return {
       'class': 'logging.handlers.RotatingFileHandler',
-      'level': logging.INFO,
+      'level': BASE_LOGGING_LEVEL,
       'formatter': 'simple',
       'filename': file_path,
       'maxBytes': max_bytes,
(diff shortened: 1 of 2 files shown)


PauloVLB (dismissed):

PauloVLB:
Thanks for working on this.

Because the root logger is set to `DEBUG` in App Engine and K8s, third-party libraries (like `urllib3`, `requests`, `google.auth`, and `google.cloud.logging`'s background transport dumping `"Submitted X logs"` on every flush) start dumping all their internal debug logs into Cloud Logging. In dev, this introduces a ton of third-party noise:


image

image

image

image



Should we silence those third-party loggers so they don't drown out our actual logs?

src/clusterfuzz/_internal/metrics/logs.py
```diff
@@ -53,6 +53,8 @@
 _is_already_handling_uncaught = False
 _default_extras = {}
 
+BASE_LOGGING_LEVEL = logging.DEBUG
```
dylanjew: could you make this an environment flag? Then you could enable and disable this in dev/prod as needed so we don't flood our logs by default

IvanBM18:
Hi @dylanjew @PauloVLB 

Took into consideration your comments and added both:
- Ability to control the logs shown by env var, defaults to INFO
- added a filter so that we only log CF logs(so that when level is DEBUG we wont show  external logs)

I'll create a following pr to make dev bots use the DEBUG log level

dylanjew (commented):
LGTM but will wait for GOSST approval

PauloVLB (approved):

PR #5463, opened 2026-09-09
Fix radamsa binary for ARM64

JuanMBriones:
When executing Mac ARM64 swarming tasks we execute radamsa binary,
hence having it as a ARM64 is needed.

- Ran `python butler.py py_unittest -t core -p engine_common_test.py` (passed).

Bug: 561690132


g-ortuno (changes requested):
It's not great that we have "resources/platform/[platform]" AND "fuzzers/bin/[platform]". Could we move RADAMSA to "resources/platform/" and then do something like:
```
def get_radamsa_path():
  """Return path to radamsa binary for current platform."""
  if environment.platform() not in ('LINUX', 'MAC'):
    return None
  return environment.get_default_tool_path('radamsa')
```

JuanMBriones:
> It's not great that we have "resources/platform/[platform]" AND "fuzzers/bin/[platform]". Could we move RADAMSA to "resources/platform/" and then do something like:
> 
> ```
> def get_radamsa_path():
>   """Return path to radamsa binary for current platform."""
>   if environment.platform() not in ('LINUX', 'MAC'):
>     return None
>   return environment.get_default_tool_path('radamsa')
> ```

Pls excuse the delay. I think I got a better solution for the platform + arch binaries.

For x86 binaries they'll be located on `resources/platform/[platform]` and for other archs (different than x86) they will be located on `resources/platform/[platform]/arch/`

Pls let me know what you think:D

g-ortuno:
So you're saying we would have:
```
resources/
  platform/
    macos/
      llvm-symbolizer
      radamsa
      arm64/
        llvm-symbolizer
        radamsa
```
Instead of:
```
resources/
  platform/
    macos/
      llvm-symbolizer
      radamsa
    macos_arm64/
      llvm-symbolizer
      radamsa
```

Was there an issue with the latter option? I don't have a super strong opinion with a small preference for the second one since it's less nested dirs and matches the format of deployment targets i.e. macos and macos_arm64.

src/clusterfuzz/_internal/system/environment.py
```diff
@@ -436,6 +436,12 @@ def get_platform_resources_directory(platform_override=None):
   if is_android(plt):
     plt = 'ANDROID'
 
+  arch = get_cpu_arch()
```
g-ortuno, line 439: This would be a good opportunity to use get_target_cpu_arch.

JuanMBriones:
> So you're saying we would have:
> 
> ```
> resources/
>   platform/
>     macos/
>       llvm-symbolizer
>       radamsa
>       arm64/
>         llvm-symbolizer
>         radamsa
> ```
> 
> Instead of:
> 
> ```
> resources/
>   platform/
>     macos/
>       llvm-symbolizer
>       radamsa
>     macos_arm64/
>       llvm-symbolizer
>       radamsa
> ```
> 
> Was there an issue with the latter option? I don't have a super strong opinion with a small preference for the second one since it's less nested dirs and matches the format of deployment targets i.e. macos and macos_arm64.

Yeah I opted for the first option. But, tbh I would love feedback on it. I'm open to choose what will suit the best:D

AFAIK platform is the operating system itself and I think is logical to have a subsection for the archtecture and not blend it together with the platform.

On the other hand, I support your arguments and I think having the same naming as the deployment target is a better practice as well.

There's another thing I've been looking into the last couple of days... having a Universal executable at least for Mac [1],[2] it basically have a header that is used by Mac to determine whether a executable might me compatible with the current architecture (the same as when we use `file` util). Meaning that we can avoid this at least for Mac.

[1]: https://en.wikipedia.org/wiki/Universal_binary
[2]: https://hacktricks.wiki/en/macos-hardening/macos-security-and-privilege-escalation/macos-files-folders-and-binaries/universal-binaries-and-mach-o-format.html

src/clusterfuzz/_internal/system/environment.py
```diff
@@ -436,6 +436,12 @@ def get_platform_resources_directory(platform_override=None):
...
+  if (not platform_override and plt != 'ANDROID' and arch and
+      not arch.startswith('x86')):
+    return os.path.join(get_resources_directory(), 'platform', plt.lower(),
+                        arch)
+
   return os.path.join(get_resources_directory(), 'platform', plt.lower())
```
g-ortuno, line 439: What do you think of:
```suggestion
plt = platform_override or platform()
platform_directory = os.path.join(get_resources_directory(), 'platform')

# All Android variants share the same android directory. It holds host-side
# tools (adb, aapt) and device-side payloads (.apk), neither of which is
# keyed by host arch, so it is never arch-split.
if is_android(plt.upper()):
  return os.path.join(platform_directory, 'android')

arch = get_host_cpu_arch()
if not arch or arch in _UNSUFFIXED_ARCHS:
  return os.path.join(platform_directory, plt.lower())

return os.path.join(platform_directory, plt.lower(), arch)
```

src/clusterfuzz/_internal/tests/core/system/environment_test.py
```diff
@@ -547,3 +547,44 @@ def test_android(self):
...
+                     environment.get_platform_resources_directory())
+
+  def test_linux(self):
+    """Test Linux x86_64 platform resources directory."""
+    self.mock.platform.return_value = 'LINUX'
+    self.mock.get_cpu_arch.return_value = 'x86_64'
+    self.assertEqual('/resources/platform/linux',
+                     environment.get_platform_resources_directory())
+
+  def test_android(self):
```
g-ortuno, line 585: We should test that `ANDROID_X86`, `ANDROID_EMULATOR`, `ANDROID_AUTO`, etc. all use /resources/platform/android?

src/clusterfuzz/_internal/tests/core/system/environment_test.py
```diff
@@ -547,3 +547,44 @@ def test_android(self):
     self.assertEqual('/resources/platform/linux/llvm-symbolizer',
                      environment.get_default_tool_path('llvm-symbolizer'))
     self.mock.get_platform_resources_directory.assert_called_once_with('linux')
+
+
+class GetPlatformResourcesDirectoryTest(unittest.TestCase):
```
g-ortuno, line 552: We should also have a test that passes a platform override.

g-ortuno (changes requested):

g-ortuno:
Up to you which dir structure you prefer! RE: the universal binary, that seems cool, but I think it adds another layer of complexity e.g. we would have to manually package all of our binaries. That seems more complicate than just downloading the right binaries to the right directory.

src/clusterfuzz/_internal/system/environment.py
JuanMBriones: just stacked the rename CL. Should be covered by it :D

src/clusterfuzz/_internal/system/environment.py
JuanMBriones: QQ where is `_UNSUFFIXED_ARCHS` defined? I see the idea is to see what arch we dont support?

JuanMBriones: I just saw it saw a dumb question lol. I introduced the set on the first commit:D

src/clusterfuzz/_internal/tests/core/system/environment_test.py
JuanMBriones: Added on latest PS:)

src/clusterfuzz/_internal/tests/core/system/environment_test.py
JuanMBriones: Added on latest PS:)

src/clusterfuzz/_internal/system/environment.py
```diff
@@ -55,6 +55,11 @@
     'use_sigaltstack': 1,
 }
 
+_UNSUFFIXED_ARCHS = {
```
g-ortuno, line 61: Add a comment.

g-ortuno (approved):
Just one comment otherwise, lgtm

src/clusterfuzz/_internal/system/environment.py
JuanMBriones: Thanks for the suggestion:D just solved in latest PS

src/clusterfuzz/_internal/system/environment.py
```diff
@@ -55,6 +55,14 @@
     'use_sigaltstack': 1,
 }
 
+# CPU architectures whose binaries live in the default unsuffixed platform
+# resource directory (e.g., `resources/platform/mac`) rather than an
+# architecture-suffixed directory (e.g., `resources/platform/mac_arm64`).
```
JuanMBriones, line 60: I added this comment. Pls let me know what you think:) @g-ortuno

g-ortuno (approved):

g-ortuno:
@PauloVLB PTAL!

src/clusterfuzz/_internal/bot/fuzzers/engine_common.py
```diff
@@ -201,16 +201,12 @@ def generate_new_testcase_mutations_using_radamsa(
 
 def get_radamsa_path():
```
PauloVLB: I might be missing something, but I think this could break radamsa on Linux and Intel Mac.

`get_radamsa_path()` now resolves to `resources/platform/<os>/radamsa`, but only `mac_arm64/radamsa` was added. On Linux, `resources/platform/linux/radamsa` is an existing folder (the one with `libradamsa.so`), and on Intel Mac `resources/platform/mac/radamsa` doesn't exist. The real binaries still seem to be in `bot/fuzzers/bin/{linux,mac}/`, and I couldn't find anything in the stack moving them.

I wrote a quick test that checks the returned path is an actual file (without mocking `os.chmod`): https://paste.googleplex.com/4727108878860288

To try it, save it as `src/clusterfuzz/_internal/tests/core/bot/fuzzers/radamsa_path_check_test.py` and run:
`python butler.py py_unittest -t core -p radamsa_path_check_test.py`

It passes on master, but fails on this branch: Linux returns a directory, and Intel Mac raises `FileNotFoundError` since the file doesn't exist.

Is there a move I'm not seeing? If not, would it make sense to keep the `bin/` layout and just add `bin/mac_arm64/radamsa`? Maybe also return `None` if the file is missing, so it doesn't raise mid-task.

g-ortuno: Replying since I'm the one that suggested moving the file. I think it's confusing for us to have two directories with binaries: `bin/` and `resources/platform/`, so I suggested we move radamsa to resources/platform/. I see that we missed moving the existing ones, which we should fix, but do you think we could standardize on a single directory for binaries?

PauloVLB: Sounds good to me! One catch is that `resources/platform/linux/radamsa` is already a folder, so we can't just move the binary there. But it looks like #3131 removed the only thing using it, and nothing references it anymore (worth double-checking), so I think we can just delete it.

g-ortuno: How about we special-case macOS arm64 in this method and then we follow up with moving radamsa for mac or removing it for linux? That way we can revert that PR if it breaks things.

PR #5465, opened 2026-09-11
Add cscope and tags files to gitignore

JuanMBriones:
When using cscope and tags for the project it might come handy automatically
ignore those files.

Bug: None
Signed-off-by: Manuel Briones <manuelbriones@google.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,8 @@ bazel-*
 # Ignore temporary build files.
 docker/base/Pipfile
 docker/base/Pipfile.lock
+
+# Tags and Cscope files
+tags
+cscope.files
+cscope.out


g-ortuno:
It looks like you're adding ignores for ctags, cscope, and gtags. Do you use all three? Can we add only the ones that you really need?

JuanMBriones:
> It looks like you're adding ignores for ctags, cscope, and gtags. Do you use all three? Can we add only the ones that you really need?

I just included the files I usually generate. Thanks for the review:)

g-ortuno (approved):

decoNR (approved):

JuanMBriones:
Thanks for the review guys!

PR #5466, opened 2026-09-11
Ensure `rss_limit_mb` is set to 0 for all Chromium/Google tasks

tbantikyan:
Bug: crbug.com/555763447

In order to ensure that no memory limit is present for our fuzzing engines, the `rss_limit_mb` flag must be set to 0. This was not being done for corpus pruning tasks and for testcases that were manually uploaded.


src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/fuzzer.py
```diff
@@ -45,11 +46,11 @@ def get_arguments(fuzzer_path) -> options.FuzzerArguments:
   if timeout is None:
     arguments[constants.TIMEOUT_FLAGNAME] = constants.DEFAULT_TIMEOUT_LIMIT
 
-  if not rss_limit_mb and (utils.is_chromium() or
-                           utils.default_project_name() == 'google'):
+  if rss_limit_mb is None and (utils.is_chromium() or
+                               utils.default_project_name() == 'google'):
     # TODO(metzman/alhijazi): Monitor if we are crashing the bots.
```
g-ortuno, line 51: I think you can delete this TODO.

src/clusterfuzz/_internal/bot/tasks/utasks/corpus_pruning_task.py
```diff
@@ -400,37 +401,35 @@ class LibFuzzerRunner(BaseRunner):
 
   def get_fuzzer_flags(self):
     """Get default libFuzzer options for pruning."""
-    rss_limit = RSS_LIMIT
```
g-ortuno, line 403: Is this constant still used?

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
g-ortuno, line 633: I think this method is currently unused? Can we leave a FIXME comment about investigating if this can be wired up or if it needs to be deleted?

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
```diff
@@ -647,7 +647,9 @@ def minimize_testcase(self, target_path, arguments, input_path, output_path,
     runner = _get_runner(target_path)
     workdir = engine_common.create_temp_fuzzing_dir('workdir')
     timeout = max_time + _CLEAN_EXIT_SECS
-    args = [
+    minimize_arguments = self._get_arguments(target_path)
```
g-ortuno, line 650: IIUC, this gets the current arguments for the fuzzer. Does that mean this would return `timeout_per_input`? If so, we probably want to override it like we do in the [reproduce task](https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py#L448).

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
g-ortuno, line 638: optional, maybe mention that Centipede doesn't store arguments in the database, so this param is safe to ignore.

src/clusterfuzz/_internal/tests/core/bot/fuzzers/centipede/centipede_engine_test.py
```diff
@@ -220,6 +220,31 @@ def test_options_arguments(self):
     args = arguments.list()
     self.assertIn('-rss_limit_mb=1234', args)
 
+  @patch('clusterfuzz._internal.base.utils.is_chromium', return_value=True)
+  def test_options_arguments_chromium_default(self, _):
+    """Tests that rss_limit_mb=0 on Chromium when unspecified."""
+    testcase_path = setup_testcase('test_fuzzer', self.test_paths)
+    engine_impl = engine.Engine()
+    # pylint: disable=protected-access
+    arguments = engine_impl._get_arguments(str(testcase_path))
```
g-ortuno: Let's also test timeout_per_input hsa the right value.

src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -514,6 +523,8 @@ def minimize_corpus(self, target_path, arguments, input_dirs, output_dir,
     """
     runner = libfuzzer.get_runner(target_path)
     libfuzzer.set_sanitizer_options(target_path)
+    arguments = fuzzer_options.FuzzerArguments.from_list(arguments)
```
g-ortuno, line 526: Apparently, from_list can return None if a regex fails, can we assert arguments is not None?

src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -567,13 +578,16 @@ def minimize_testcase(self, target_path, arguments, input_path, output_path,
     runner = libfuzzer.get_runner(target_path)
     libfuzzer.set_sanitizer_options(target_path)
 
+    arguments = fuzzer_options.FuzzerArguments.from_list(arguments)
```
g-ortuno, line 581: Same here, can we assert it's not None?

src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -98,6 +98,14 @@ def fuzz_additional_processing_timeout(self, options):
     # get_fuzz_timeout returns a negative value.
     return -fuzz_timeout
 
+  def _add_rss_limit_if_missing(self, target_path, arguments):
```
g-ortuno, line 101: The changes in this file are defensive, right? We use `--rss_limit_mb=0` as the default value, which is stored in the DB; when we retrieve the arguments and pass them to the functions in these file as `arguments`, they will use the default value already.

g-ortuno, line 101: WDYT about changing `arguments` to `db_arguments`? Or maybe add a comment above fuzzer_arguments to clarify that it's the arguments in the fuzzer's options file? It's not super clear why we have two sets of arguments 😅

src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/fuzzer.py
```diff
@@ -29,13 +29,14 @@ def get_extra_env(fuzzer_path):
...
 
-def get_arguments(fuzzer_path) -> options.FuzzerArguments:
+def get_arguments(fuzzer_path, fuzzer_options=None) -> options.FuzzerArguments:
   """Get arguments for a given fuzz target."""
   arguments = options.FuzzerArguments()
   rss_limit_mb = None
   timeout = None
 
-  fuzzer_options = options.get_fuzz_target_options(fuzzer_path)
+  if fuzzer_options is None:
```
g-ortuno, line 38: It's a bit confusing to re-use this method for get_fuzzer_flags in corpus_pruning_task.py. This method is supposed to take a `fuzzer_path` and return the arguments in the options file in that path; but now it takes a `fuzzer_options` which causes it to ignore fuzzer_path and the options file.

It also looks like we only call it to get rss_limit_mb. What do you think about extracting the rss_limit_mb logic into its own `get_rss_limit_mb_for_options(fuzzer_options)` method and use that in the corpus pruning task?

g-ortuno (changes requested):
Phew that was hard. I didn't get to look at the libfuzzer tests yet!

src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/fuzzer.py
tbantikyan: Done.

src/clusterfuzz/_internal/bot/tasks/utasks/corpus_pruning_task.py
tbantikyan: No. Nor was `RSS_LIMIT_MB_FLAG`. Deleted both.

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
tbantikyan: Added FIXME.

tbantikyan: Added comment.

src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py
tbantikyan: In many cases, what you said is correct. The changes are important for uploaded testcases, though, as those will not already have `--rss_limit_mb=0`.

tbantikyan: Renamed parameter for clarity.

src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py
tbantikyan: Good catch. I moved this logic into `_add_rss_limit_if_missing`.

I made it return `arguments` with no changes if `from_list` returns `None`, as I don't want to introduce a failures to tasks that previously were not failing. I'm not sure how common it is that we'd get an invalid regex; if it's uncommon enough, I'm thinking it's fine to forgo setting `rss_limit_mb`.

src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py
tbantikyan: [See other comment](https://github.com/google/clusterfuzz/pull/5466/changes/BASE..be55e5d2541874ef076b013ed2a6e86bb9e0e812#r4017752139).

src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/fuzzer.py
tbantikyan: Good idea. Factored out the `rss_limit_mb` logic into its own method.

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
tbantikyan: Yes, it does return `timeout_per_input`. I'm thinking it makes more sense to delete the argument than to do what reproduce is doing. In deleting, we are preserving the functionality that existed before this CL, as the `timeout_per_input` value is not set in this method. Also, I think that changing environment variables is a little like working with global variables and makes the code harder to reason about, so I'd prefer to avoid it.

tbantikyan (commented):
PTAL.

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
g-ortuno: Do you know what libfuzzer does for the minimize task? Does it set a value for TIMEOUT_PER_INPUT? If so, we should probably copy that.

Ah, sorry for the confusion. I didn't mean to say we should use environment variables, I more meant we should replace with `TIMEOUT_PER_INPUT_REPR_DEFAULT` or `TEST_TIMEOUT` the same way `reproduce` does. IIUC, if we don't set the flag we would use the default value from centipede which might not be the correct one depending on TEST_TIMEOUT and TIMEOUT_PER_INPUT_REPR_DEFAULT.

IIUC, `reproduce` uses environment variables because it's not running `centipede --binary`; it's running the binary directly e.g. `foo_fuzzer` and the only way to pass arguments to centipede in that case is through environment variables.

src/clusterfuzz/_internal/bot/tasks/utasks/corpus_pruning_task.py
```diff
@@ -400,37 +397,33 @@ class LibFuzzerRunner(BaseRunner):
...
         max_len = custom_max_len
 
-      # Some targets might falsely report leaks all the time, so allow this to
-      # be disabled.
-      custom_detect_leaks = libfuzzer_arguments.get(
-          'detect_leaks', constructor=int)
+    # Some targets might falsely report leaks all the time, so allow this to
+    # be disabled.
+    detect_leaks = 1
+    if self.fuzzer_options:
```
g-ortuno: out of curiosity, why do we split this if statement into two? Is it to have the variables closer to where they are set?

src/clusterfuzz/_internal/tests/core/bot/fuzzers/libFuzzer/engine_test.py
```diff
@@ -142,6 +142,14 @@ def test_prepare_auto_add_dict(self):
         '-dict=/path/target.dict'
     ], options.arguments)
 
+  def test_prepare_chromium(self):
```
g-ortuno: How come we don't use:
```
@patch('clusterfuzz._internal.base.utils.is_chromium', return_value=False)
```

g-ortuno (changes requested):
Just a couple of small things.

src/clusterfuzz/_internal/bot/tasks/utasks/corpus_pruning_task.py
tbantikyan: I like the readability of having the variable setting like this more than having one large if block. I reverted it to minimize the diff and keep changes relevant to the bug.

src/clusterfuzz/_internal/tests/core/bot/fuzzers/libFuzzer/engine_test.py
tbantikyan: Good catch, updated to use it here.

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
tbantikyan: libFuzzer's `minimize_testcase` method just passes through the arguments passed to it, but IIUC a timeout is typically present from the testcase. I'm hesitant to handle timeout logic now since it's not entirely relevant to the PR. I could open a bug and look into, alongside the `FIXME`, in another PR if that would make sense.

g-ortuno: If it's passing through `arguments` then they are the arguments from the database. Not all arguments are stored in the DB as is, so it might be removing it or setting something different.

Anyway, I think leaving a FIXME is ok. This method is never called anyway :/

g-ortuno (approved):

tbantikyan:
Assigning to sheriff @decoNR. Please find someone to take a look

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
decoNR: Can we create a bug for this to avoid losing track?

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
```diff
@@ -647,7 +649,14 @@ def minimize_testcase(self, target_path, arguments, input_path, output_path,
     runner = _get_runner(target_path)
     workdir = engine_common.create_temp_fuzzing_dir('workdir')
     timeout = max_time + _CLEAN_EXIT_SECS
-    args = [
+
+    # Centipede does not store arguments in the database, so the `arguments`
+    # parameter is safe to ignore
+    minimize_arguments = self._get_arguments(target_path)
```
decoNR, line 655: Just for context, I saw that this function sets other arguments than rss_limit. What is the impact of adding those extra arguments here? (not sure if the next line filters out all of them)

src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py
```diff
@@ -98,6 +98,22 @@ def fuzz_additional_processing_timeout(self, options):
...
+    if db_arguments is None:
+      logs.warning(
+          'Failed to parse arguments. Returning without setting rss_limit_mb.',
+          arguments=arguments)
+      return arguments
+
+    if constants.RSS_LIMIT_FLAGNAME not in db_arguments:
+      db_arguments[constants.RSS_LIMIT_FLAGNAME] = fuzzer.get_rss_limit_mb(
+          fuzzer_options.get_fuzz_target_options(target_path))
+    return db_arguments.list()
```
decoNR, line 115: I might be of context here but if none of the two "if" statements are true the function receives `arguments` and returns `db_arguments`. Is it guaranteed that they are the same? If not, why is this safe? The same question applies for the second if.

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
tbantikyan: Only two arguments aren't stripped out: `symbolizer-path` and `address_space_limit`. This mirrors what ClusterFuzz does for `minimize_corpus` and the other Centipede jobs. So if we want to set an external symbolizer, this ensures that minimize_task picks it up as well. We also ensure that `address_space_limit` is set for all tasks.

If you'd prefer, I can factor out the logic for setting`rss_limit_mb` so that we avoid setting any other arguments.

src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py
tbantikyan: If none of the two `if` statements are true, we are converting the list to a dictionary and back to a list. The input and output will be semantically identical, so it should be safe. The one potentially difference will be that duplicate flags are de-duped. Also, this is already an established pattern in the codebase ([example](https://github.com/google/clusterfuzz/blob/852c50ea754c3746c9947ef6c0be35c4bd85bc4e/src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py#L417)).

I think that this can be more clearly written. I updated the logic to be a true no-op if `rss_limit_mb` is present. Let me know what you think.

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
tbantikyan: Done.

src/clusterfuzz/_internal/bot/fuzzers/centipede/engine.py
decoNR: As it seems to be a pattern to use it, I do not think you need to change it. Thanks for the explanation.

src/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py
decoNR: Since it is a pattern, I do not think you need to change it. Thanks, the change you made LGTM.

decoNR (commented):
Hi @tbantikyan , were you able to test these changes in the dev environment. Could you please add evidence?

tbantikyan:
> Hi @tbantikyan , were you able to test these changes in the dev environment. Could you please add evidence?

I had it for a full day in `dev` last week. I have uploaded it to `dev` now and will let it sit for a few more hours. Would that be sufficient? We are hoping to get this change in for this week's release.

decoNR:
> I had it for a full day in `dev` last week. I have uploaded it to `dev` now and will let it sit for a few more hours. Would that be sufficient? We are hoping to get this change in for this week's release.

Hi, yes. Please add evidence here to show that your change worked and that no new associated errors have appeared. Thank you!

tbantikyan:
Corpus pruning tasks with `rss_limit_mb=0`:
- libFuzzer: [success log with arguments](https://pantheon.corp.google.com/logs/query;cursorTimestamp=2026-09-24T11:32:07.092697Z;duration=PT12H;query=SEARCH%2528%22%60rss_limit_mb%60%20running%20merge%22%2529%0Atimestamp%3D%222026-09-24T11:32:07.092697Z%22%0AinsertId%3D%22b3ri6lf4vr5bo%22?project=clusterfuzz-development)
- Centipede: [arguments](https://pantheon.corp.google.com/logs/query;cursorTimestamp=2026-09-24T05:15:48.964556Z;duration=P1D;query=jsonPayload.extras.task_id%3D%22f188a26b-046f-402f-a0c0-b02a1040be14%22%0Atimestamp%3D%222026-09-24T05:15:48.964556Z%22%0AinsertId%3D%22uu43u0f29tvv7%22?project=clusterfuzz-development), [success](https://pantheon.corp.google.com/logs/query;cursorTimestamp=2026-09-24T05:15:49.155387Z;duration=P1D;query=jsonPayload.extras.task_id%3D%22f188a26b-046f-402f-a0c0-b02a1040be14%22%0Atimestamp%3D%222026-09-24T05:15:49.155387Z%22%0AinsertId%3D%22uu43u0f29tvva%22?project=clusterfuzz-development)

Upload task, uploaded without `rss_limit_mb`, showing that the flag was injected: [log](https://pantheon.corp.google.com/logs/query;cursorTimestamp=2026-09-23T21:21:48.994042Z;endTime=2026-09-24T17:28:35Z;query=jsonPayload.extras.task_id%3D%22cb1cf358-b155-47a5-817f-8e6533b2713c%22%20AND%20jsonPayload.extras.testcase_id%3D%224640945576902656%22%20AND%20jsonPayload.extras.task_name%3D%22analyze%22%20AND%20timestamp%20%3E%3D%20%222026-08-24T17:49:49.544780Z%22%0ASEARCH%2528%22%60rss_limit_mb%60%22%2529%0Atimestamp%3D%222026-09-23T21:21:48.994042Z%22%0AinsertId%3D%2215xdpwbfeeivbv%22;startTime=2026-09-24T16:21:36Z?project=clusterfuzz-development)

New errors:
- [linux_d8_dbg doesn't exist](https://pantheon.corp.google.com/errors/detail/CNGFmtOEx5yq9QE;time=P1D;locations=global?project=clusterfuzz-development&e=-13802955&mods=logs_tg_prod) -- seems to be a configuration error unrelated to this change.

decoNR (approved):

PR #5467, opened 2026-09-11
Fix schedule fuzz return

PauloVLB:
### Problem

After go/clusterfuzz-config-pr/369, `schedule_fuzz` is configured to run every 5 minutes, but the logs show it running 4 times per 5 minute window:

image

Turns out `schedule_fuzz.main()` doesn't return anything. `run_cron.py` does [`return 0 if task_module.main() else 1`](https://github.com/google/clusterfuzz/blob/master/src/python/bot/startup/run_cron.py#L68), so `None` becomes exit code 1. The cron has always been "failing" from Kubernetes' point of view.

The CronJob has `restartPolicy: OnFailure` and `backoffLimit: 3`, so every tick turns into 1 attempt + 3 retries with the usual 10s/20s/40s backoff. That matches the gaps above exactly.

The annoying part is that each retry re-runs the scheduling logic from scratch, so we can be publishing up to 4x the intended number of fuzz tasks per cycle.

### Solution
`main()` returns `True` now. 

So exit 1 means "something unexpected broke", not "nothing got scheduled".

### Validation
After this deploys, the log cadence should drop from 4 lines per 5 minutes to 1.

### DIFF
--- a/src/clusterfuzz/_internal/cron/schedule_fuzz.py
+++ b/src/clusterfuzz/_internal/cron/schedule_fuzz.py
@@ -316,3 +316,4 @@ def schedule_fuzz_tasks():
 
 def main():
   schedule_fuzz_tasks()
+  return True


src/clusterfuzz/_internal/cron/schedule_fuzz.py
```diff
@@ -316,3 +316,4 @@ def schedule_fuzz_tasks():
 
 def main():
   schedule_fuzz_tasks()
+  return True
```
decoNR, line 319: What about making schedule_fuzz_tasks() to return True (if it succeeds) and return schedule_fuzz_tasks() here. This way we would retry on failure (I imagine this is the expected).

ViniciustCosta (approved):

decoNR (approved):

src/clusterfuzz/_internal/cron/schedule_fuzz.py
PauloVLB: That's already handled. `run_cron` wraps the call in a try/except, so if anything actually fails the exception bubbles up and we exit 1, which makes k8s retry. `return True` is only reached when nothing raised.

PR #5468, opened 2026-09-11
Add pip req arm64

JuanMBriones:
When packaging bot deployment bundles on a Linux host via `butler package`,
`_install_pip()` installs all packages from `Pipfile.lock` (including
transitive dependencies with Linux x86_64 ELF shared objects) into
`src/third_party`. `_install_platform_pip()` then downloads pre-compiled
binary wheels for each package listed in `src/platform_requirements.txt`
to replace the Linux binaries with target platform native binaries.

Previously, `markupsafe` and `pynacl` were transitive dependencies containing
compiled C/CFFI extensions that were missing from
`src/platform_requirements.txt`:

* `markupsafe==3.0.3` (transitive dependency of `Jinja2==3.1.4` / `Flask`):
   contains the compiled C extension `markupsafe/_speedups.cpython-*.so`.
   Without it in `platform_requirements.txt`, the Linux ELF `.so` remained in
   the macOS ARM64 bundle.
* `pynacl==1.6.0` (transitive dependency of `pygithub==1.55`): contains the
   compiled libsodium CFFI native binary `nacl/_sodium.abi3.so`. Without it in
   `platform_requirements.txt`, importing `github` / `nacl` on macOS ARM64 fails
   with `ImportError: dlopen(.../nacl/_sodium.abi3.so): not a mach-o file`.

Adding both packages ensures `butler package -p macos_arm64` downloads the
native macOS ARM64/universal2 wheels (`markupsafe-3.0.3-cp311-cp311-macosx_11_0_arm64.whl`
and `pynacl-1.6.0-cp38-abi3-macosx_10_10_universal2.whl`) into `src/third_party`.

Bug: 555853673


src/platform_requirements.txt
g-ortuno: Can we mention here that it also includes transitive dependencies? We should change Pipfile to mention that transitive dependencies should added here as well.

g-ortuno (changes requested):

src/platform_requirements.txt
JuanMBriones: @g-ortuno I think this might be achieved by above's lines? Also, on `Pipfile` we've got these lines https://github.com/google/clusterfuzz/blob/dd68bf6511544caa1e457b1c8984286e637d56a1/Pipfile#L7-8

JuanMBriones: I left inline comments. But I think they mention it (but not at the files beginning)

g-ortuno: Right, but I don't think it covers the packages this PR is adding. From the PR description:
> markupsafe and pynacl were transitive dependencies containing compiled C/CFFI extensions that were missing from src/platform_requirements.txt:

Specifically the part about transitive dependencies.

JuanMBriones: I see. I just updated the CL. Pls let me know what you think:)

g-ortuno:
I think you pushed your radamsa changes to this branch :sweat_smile:

g-ortuno (approved):

g-ortuno:
@PauloVLB PTAL!

PauloVLB (approved):

PR #5470, opened 2026-09-15
[Android] Adds guardrails for non apk fuzzing

IvanBM18:
## Overview
In a previous stack of PRs  (See https://github.com/google/clusterfuzz/pull/5453) I made a ton of changes for improving android fuzzing overall error detection when executing test cases, but, as i got pointed out, my changes assumed that android fuzzing work exclusively under apks, so some of the checks/validations didn't take into account jobs where no APK nor PKG_NAME is provided.

This PR focus on tuning various checks/validations previously altered in the stack of Prs above, so that if in android fuzzing with a binary(with this i mean non-apk fuzzing) then we go back to the usual checks that we had in place.

## Changes
- Bad build check in `tescase_manager.py` now only checks for an apk process liveness if an apk package is found.
- The test cases directory in `app.py` defaults to the old (non accesible by apks) dir if no apk is found.
- When clearing the directory in `device.py` we now clean using the `find` command, as my previous command could ignore the hidden files(e.g: `.<filename>`)
- Exit info is only checked if fuzzing with an apk, else we ignore this

## Tests
- Introduces a ton of unit tests for  the non apk test cases
Fuzzing is correctly executed on valid APKS:
image

And it continues to fail on invalid apks.


src/clusterfuzz/_internal/platforms/android/app.py
```diff
@@ -94,7 +94,9 @@ def get_package_name(apk_path=None):
 
 def get_testcases_directory():
   """Returns the testcases directory."""
-  package_name = get_package_name() or ''
+  package_name = get_package_name()
+  if not package_name:
```
dylanjew, line 98: nit: add a comment explaining when we fallback

src/clusterfuzz/_internal/tests/core/platforms/android/device_test.py
```diff
@@ -120,3 +120,47 @@ def test_asan_device_flavor(self):
...
+  def test_clears_apk_scoped_directory_as_root(self):
+    """Test that, for an APK target, the contents of the package's scoped
+    storage directory are deleted as root. Expects a single `find ... -delete`
+    shell command; `-mindepth 1` guarantees only the contents are removed and
+    the directory itself is preserved for the next run."""
+    environment.set_value('PKG_NAME', 'com.google.chrome')
+
+    device.clear_testcase_directory()
+
+    self.mock.run_shell_command.assert_called_once_with(
```
dylanjew: nit: it would be better to assert against the results of the nuanced difference between `rm -rf` and `find ... -delete` rather than just asserting which was called

dylanjew (commented):
LGTM

src/clusterfuzz/_internal/tests/core/platforms/android/device_test.py
IvanBM18: uff that ended up being a big change, asked jetski and he refactored the whole code logic to use the devices internal fs as a replacement for an adb device fs, but because of this i marked the tests so that they only run on POSIX environments, which have a fs that functions the same.

src/clusterfuzz/_internal/platforms/android/app.py
IvanBM18: Done!

src/clusterfuzz/_internal/tests/core/platforms/android/device_test.py
dylanjew: thanks! certainly not a blocker depending on how hard the set up is

IvanBM18: Thanks! marking as resolved, but FYI the big changes it' using  this function that overrides adb calls in said tests:
```python
def _run_emitted_command(self):
    """Executes the shell command that was handed to adb against the local
    filesystem, mimicking the device shell, and returns its exit code."""
    self.mock.run_shell_command.assert_called_once()
    command = self.mock.run_shell_command.call_args[0][0]
    return subprocess.run(
        command, shell=True, check=False, capture_output=True).returncode   
```
But apart from that, the tests look the same

decoNR (approved):

decoNR:
As discussed offline, I will merge it to enable deployment (FYI @IvanBM18) .

PR #5471, opened 2026-09-15
Respect exclude patterns in reproducibility test

emaxx-google:
Filter out crashes that match any exclude pattern during the reproducibility test. This fixes the issue that a crash would be treated as allegedly reliably reproducible even if all reproducibility runs matched the exclude pattern.

Before this commit, the exclude patterns (ignore_stacktrace) were checked in a few scenarios (check_for_initial_crash, variant_task, reproduce_with_retries, check_for_bad_build), but not in test_reproduce_reliability.

b/551164285


src/clusterfuzz/_internal/bot/testcase_manager.py
```diff
@@ -819,6 +819,10 @@ def test_reproduce_reliability(self, retries, expected_state,
       if not expected_state:
         expected_state = state.crash_state
 
+      if crash_result.should_ignore():
+        logs.info('Crash stacktrace matched ignore signatures, ignored.')
+        continue
+
```
PauloVLB: Nit: shouldn't the `should_ignore()` check come before the `if not expected_state:` block? Fuzz task calls this function with `expected_state=None`:
https://github.com/google/clusterfuzz/blob/873277e23397266c99e284c7c20d142b03081862/src/clusterfuzz/_internal/bot/tasks/utasks/fuzz_task.py#L359-L368
So if the first crash is one that should be ignored, its state still becomes the reference state. Any real crash after that gets compared against it, gets dropped as "unrelated state", and the testcase ends up marked as not reproducible.

PauloVLB (approved):

PR #5475, opened 2026-09-17
[Better Tests] Adds new test dependencies

IvanBM18:
Bug: b/555367430

This is our first ever PR focused on adding better tests to CF. Starting by adding new dependencies:
- `pytest`: This is the python's modern alternative to `unittest` allows to simplify test setup, and overall added for reusable mock creation across multiple modules.
- `pytest-cov`: To create test coverage reports over pytest tests.
image

- `testcontainers`: Allow us to launch containers with real-like external dependencies like datastore., pub sub queue, or in our case, even other CF worker bots, etc.

I also added a sample test .py script to verify that testcontainers & pytest setup is correct, this script is not executed at CI, since we are not yet ready for this.

For more info read:
https://docs.google.com/document/d/1OcvLED_sEJS0Y8RcluL39qqW4v8-liFxiG9eyV20s8c/edit?usp=sharing&resourcekey=0-Qca3yUhT4VsnYnUe74T_Iw


src/clusterfuzz/tests/dev_env_testcontainers_test.py
```diff
@@ -0,0 +1,76 @@
+# Copyright 2025 Google LLC
```
JuanMBriones: Perhaps this can be changed to
```
# Copyright 2026 Google LLC
```

src/clusterfuzz/tests/dev_env_testcontainers_test.py
```diff
@@ -0,0 +1,76 @@
...
+# limitations under the License.
+"""Developer environment checks for the testcontainers dependency.
+
+These verify that a developer environment can actually launch containers
+through testcontainers. They are skipped automatically when no usable Docker
+daemon is reachable (e.g. on bots without Docker).
+
+Run with:
+  pipenv run python -m pytest \
+      src/clusterfuzz/tests/dev_env_testcontainers_test.py
```
JuanMBriones: Run with:
\```
pipenv run python -m pytest \
      src/clusterfuzz/tests/dev_env_testcontainers_test.py
\```

src/clusterfuzz/tests/dev_env_testcontainers_test.py
```diff
@@ -0,0 +1,76 @@
...
+
+or, from inside an activated virtualenv (`pipenv shell`):
+  python -m pytest src/clusterfuzz/tests/dev_env_testcontainers_test.py
```
JuanMBriones: \```python -m pytest src/clusterfuzz/tests/dev_env_testcontainers_test.py\```

JuanMBriones (commented):
LGTM!

src/clusterfuzz/tests/dev_env_testcontainers_test.py
IvanBM18: Already done

src/clusterfuzz/tests/test_dev_env_testcontainers.py
```diff
@@ -0,0 +1,83 @@
...
+pytest collection working while opting out of butler's lint rule requiring an
+__init__.py next to every *_test.py file, which exists to protect unittest
+discovery that this file does not rely on.
+"""
+
+import pytest
+from testcontainers.core.container import DockerContainer
+from testcontainers.core.wait_strategies import LogMessageWaitStrategy
+
+IMAGE = 'alpine:3.20'
```
dylanjew: how did you get this value? should we get this from another config file?

src/clusterfuzz/tests/test_dev_env_testcontainers.py
```diff
@@ -0,0 +1,83 @@
...
+
+def _docker_available():
+  """Return True if a Docker daemon is reachable."""
+  try:
+    # Imported lazily so that a missing docker SDK results in a skip, not an
+    # import error.
+    import docker  # pylint: disable=import-outside-toplevel
+    docker.from_env().ping()
+    return True
+  except Exception:  # pylint: disable=broad-except
```
dylanjew: is this when docker isn't set up? Should we log something here rather than failing silently? or is there another error message when this is run without docker?

src/clusterfuzz/tests/test_dev_env_testcontainers.py
dylanjew, line 1: nit: i would use `local` instead of `dev` to avoid confusion

src/clusterfuzz/tests/test_dev_env_testcontainers.py
```diff
@@ -0,0 +1,83 @@
...
+or, from inside an activated virtualenv (`pipenv shell`):
+```
+  python -m pytest src/clusterfuzz/tests/test_dev_env_testcontainers.py
+```
+
+This is deliberately not part of CI: `butler.py py_unittest` only discovers
+tests under src/clusterfuzz/_internal/tests/{core,appengine}, so this file is
+never collected by the `core` or `appengine` targets that CI runs. The
+`test_` prefix (rather than a `_test.py` suffix) is also intentional: it keeps
+pytest collection working while opting out of butler's lint rule requiring an
```
dylanjew: I agree that we should keep our existing tests but introducing a new prefix syntax is confusing and easily missed when adding new tests.

The issue is that if we use the existing `_test.py` syntax, the old lint rules would be applied to these new pytest files? Maybe there's a way to enforce this with a test decorator and changing the lint rules?

dylanjew (commented):
Could you add a md documentation for this environment and running these tests?

src/clusterfuzz/tests/test_dev_env_testcontainers.py
IvanBM18: As long as i remembered, the alpine image in general has been the most lightweight linux image i can think of , it has basically nothing.

So i googled Alpine image and saw this version as the sable version, though double checking the release history for this, seems like there are newer releases, still WDYT?

src/clusterfuzz/tests/test_dev_env_testcontainers.py
IvanBM18: Done!

src/clusterfuzz/tests/test_dev_env_testcontainers.py
IvanBM18: The linting rule is inforced not trough pylint, but trough a custom script:
https://github.com/google/clusterfuzz/blob/master/src/local/butler/lint.py#L139

src/clusterfuzz/tests/test_dev_env_testcontainers.py
IvanBM18: Yes, and yes, just added a print(on purpose because these file is not on butler, hence i can't import a log library here).

IvanBM18:
> Could you add a md documentation for this environment and running these tests?

Im seeing to much confusion regarding this file, this one is merely for demonstration purposes or for a dev to run manually, and is never intended to be part of our new test suite, so maybe its best that i delete it in that case, WDYT? @dylanjew

.gitignore
```diff
@@ -64,3 +64,6 @@ docker/base/Pipfile.lock
 tags
 cscope.files
 cscope.out
+
+# Test coverage files
+.coverage
```
decoNR: nit: please add an empty line at the end, I think this is the pattern.

dylanjew:
> > Could you add a md documentation for this environment and running these tests?
> 
> Im seeing to much confusion regarding this file, this one is merely for demonstration purposes or for a dev to run manually, and is never intended to be part of our new test suite, so maybe its best that i delete it in that case, WDYT? @dylanjew

This one is just a test file, but I'm wondering whether all tests going forward are going to use this new syntax, or if just this test file will be different. I'd like to see how you plan to run new tests and what they'd look like. So yes, maybe that would be more clear in the next PRs.

We can keep the test file if it's helpful for local testing. If we don't expect people to use it to debug their local env we can probably delete it and rely on the follow up tests instead.

IvanBM18:
> This one is just a test file, but I'm wondering whether all tests going forward are going to use this new syntax, or if just this test file will be different. I'd like to see how you plan to run new tests and what they'd look like. So yes, maybe that would be more clear in the next PRs.
> 
> We can keep the test file if it's helpful for local testing. If we don't expect people to use it to debug their local env we can probably delete it and rely on the follow up tests instead.

Deleted the file to avoid confusion, plus the new tests should inherently do the same, hence what this tested, will be  asserted this inherently.

We can discuss file name syntax, plus other naming conventions in a PR focused on adding the first tests(if curious that one is [a draft](https://github.com/google/clusterfuzz/pull/5486) right now), this one original purpose is just the dependencies, sorry for the confusions this might have caused

PR #5476, opened 2026-09-17
[Swarming] Reconsiders swarming tasks for remote execution

IvanBM18:
We laid the ground for swarming scheduling quite some time ago, but due to other blockers(i.e. android fuzzing issues), we  never really tested the scheduling in prod.

The issue here is that `batch_service.is_remote_task()` only returns true for tasks whose platform has a matching config in the `batch.yaml` file, i suspect we didn't saw this before because until recently, android emulator fuzzing, was scheduled as a `LINUX` platform, which is considered in batch config, were as `ANDROID_EMULATOR` is not

## Changes
- When finishing a utask's preprocess step we know check for both swarming & regular batch tasks
- Added Unit Tests to check that a task's execution is redirected or either executed locally based on the environment, the job & the current task
 
 ## Tests
 - In prod all of the assertion errors i could find in the last week were related to the assertion error thrown when a swarming job is not considered in the batch assertion, i queried to find if there were other cases of this happening and couldn't find any: 
image

Swarming job Before: 
image

After: With this small change swarming task are now successfully redirected to the utask_main queue:
image

So that they are processed by the utask_main scheduler


src/clusterfuzz/_internal/bot/tasks/task_types.py
```diff
@@ -175,7 +175,8 @@ def execute(self, task_argument, job_type, uworker_env):
     if download_url is None:
       return
 
-    assert batch_service.is_remote_task(command, job_type)
+    assert batch_service.is_remote_task(
+        command, job_type) or swarming.is_swarming_task(job_type)
```
IvanBM18: Other option here would be to use 
the `is_remote_utask(command, job)` method, but that one performs additional checks and i don't dare myself to risk the scheduling of every single utask command

IvanBM18: Also if i use said method, for fuzzing tasks, it would return `false` hence not scheduling them. That's because fuzzing tasks are by default a `UTaskLocalExecutor` command, but CF's spagetti code overwrites it to a normal `UTask` when picked up by tworkers.

Xeicker: nit: in line 160 there is a call to this same function `swarming.is_swarming_task(job_type)` would it be worth saving the result in a variable?

Xeicker (approved):

src/clusterfuzz/_internal/bot/tasks/task_types.py
IvanBM18: Done!

src/clusterfuzz/_internal/bot/tasks/task_types.py
```diff
@@ -175,7 +176,7 @@ def execute(self, task_argument, job_type, uworker_env):
     if download_url is None:
       return
 
-    assert batch_service.is_remote_task(command, job_type)
+    assert batch_service.is_remote_task(command, job_type) or is_swarming_job
```
dylanjew, line 180: why is `batch_service.is_remote_task` different than `is_swarming_job`?

do we ever use `batch_service.is_remote_task(command, job_type)` and specifically want batch but not swarming?

I'm worried that if this is unclear there are other places which do this check where we need to check for both.

IvanBM18: `is_remote_task()` checks that for the given job platfrom theres a spec defined in batch's yaml config. 
Same for is_swarming_task, but for swarming specific config + additional checks.

Honestly I haven't see a log indicating that this assertion is safeguarding us from executing unwanted jobs, so technically we could delete it as well, but i don't want to do that change here because if a rollback is performed we would continue to block swarming scheduling.

Ideally we would have a method that globally tells us that the given job can be executed remotely, we kind of have `is_remote_utask(command, job)` but that one, if used at this exact line, would return false(when it should be true!)

So by having the changes like these we indeed ignore the already existing technical debt.

src/clusterfuzz/_internal/bot/tasks/task_types.py
dylanjew: Oh, I see this comment here. I agree it's confusing, but this is going to make it more confusing.

Maybe to unblock the swarming launch let's keep this as is but we should file a bug.

src/clusterfuzz/_internal/bot/tasks/task_types.py
IvanBM18: Another idea could be to create a bug and a todo comment here to either:
- Delete the assertion or
- Refactor to have a global check that validates that the given job can be executed remotely + the target backend yaml config has a spec for the given job platform.

dylanjew: Thanks for looking into this. I'm good with filing a crbug and leaving a TODO to revisit this.

I agree that a global method for "is this running remote so for example we generate signed URLs" is the way to go. And then if we care about whether it's running on batch or swarming, we would have more granular methods.

src/clusterfuzz/_internal/tests/core/bot/tasks/task_types_test.py
```diff
@@ -192,3 +195,176 @@ def test_execute_proceeds_below_limit(self):
...
+  # off to batch or swarming.
+  UTASK_COMMANDS = sorted(
+      command for command, task_type in task_types.COMMAND_TYPES.items()
+      if issubclass(task_type, task_types.BaseUTask))
+
+  # Utasks that are only remotely executed when a tworker forces it, see
+  # commands.get_command_object().
+  TWORKER_ONLY_REMOTE_COMMANDS = sorted(
+      command for command, task_type in task_types.COMMAND_TYPES.items()
+      if task_type is task_types.UTaskLocalExecutor)
```
dylanjew, line 214: out of curiousity, is this just `fuzz` task in reality?

src/clusterfuzz/_internal/bot/tasks/task_types.py
dylanjew: does this also need to account for is_swarming_task?

src/clusterfuzz/_internal/tests/core/bot/tasks/task_types_test.py
```diff
@@ -192,3 +195,176 @@ def test_execute_proceeds_below_limit(self):
...
+    self.mock.is_remote_task.return_value = False
+    self.mock.is_swarming_task.return_value = True
+
+    for command in self.UTASK_COMMANDS:
+      with self.subTest(command=command):
+        self._execute(command)
+        self._assert_queued_on(
+            pub_sub_task_queue.SWARMING_UTASK_MAIN_QUEUE.name)
+
+  def test_trusted_tasks_always_execute_locally(self):
```
dylanjew: does this test name make sense? it seems like the comment and assertion are that nothing is executed locally

src/clusterfuzz/_internal/tests/core/bot/tasks/task_types_test.py
```diff
@@ -192,3 +195,176 @@ def test_execute_proceeds_below_limit(self):
...
+        'clusterfuzz._internal.base.tasks.pub_sub_task_queue.PubSubTaskQueue.get_max_target_size',
+        'clusterfuzz._internal.base.tasks.task_utils.is_remotely_executing_utasks',
+        'clusterfuzz._internal.batch.service.is_remote_task',
+        'clusterfuzz._internal.bot.tasks.task_types.BaseUTask.execute_locally',
+        'clusterfuzz._internal.bot.tasks.task_types.UTask.preprocess',
+        'clusterfuzz._internal.swarming.is_swarming_task',
+    ])
+
+    # Defaults: nothing is remote and the utask_main queue is empty.
+    self.mock.is_remotely_executing_utasks.return_value = False
```
dylanjew, line 235: I understand wanting to cover all scenarios with tests, but my understanding is that this will always be true after the batch migration.

src/clusterfuzz/_internal/tests/core/bot/tasks/task_types_test.py
```diff
@@ -192,3 +195,176 @@ def test_execute_proceeds_below_limit(self):
...
+    utask_main queue, even though it is not a batch remote task."""
+    self.mock.is_remotely_executing_utasks.return_value = True
+    self.mock.is_remote_task.return_value = False
+    self.mock.is_swarming_task.return_value = True
+
+    for command in self.UTASK_COMMANDS:
+      with self.subTest(command=command):
+        self._execute(command)
+        if command in self.TWORKER_ONLY_REMOTE_COMMANDS:
+          self._assert_executed_locally()
```
dylanjew: why are tasks still executed locally when `Is_swarming_task` is true? what does this configuration actually map to?

dylanjew (commented):
mostly LGTM. just a couple questions

src/clusterfuzz/_internal/tests/core/bot/tasks/task_types_test.py
IvanBM18: Yes haha, still open in case this changes

src/clusterfuzz/_internal/tests/core/bot/tasks/task_types_test.py
IvanBM18: oops my bad, this test class went trough multiple iterations, changing it!

src/clusterfuzz/_internal/tests/core/bot/tasks/task_types_test.py
IvanBM18: Spot on, this if is unnecessary in this test case, and actually without that line, this test case is duplicate, thanks for pointing it out

src/clusterfuzz/_internal/tests/core/bot/tasks/task_types_test.py
IvanBM18: SO IIUC, you are suggesting only cover test cases where this is always true?

dylanjew: no, i think we need the test coverage. But I'm struggling to understand this batch vs non-batch flow on swarming.

I think the non-batch flow is a dead code path if the task is a swarming task.

IvanBM18: Ok... No worries honestly scheduling's logic is very confusing with constant overwrites and changes to the flow were a task was supposed to land but because X var is on then you completely change the trajectory.

So heres the flow as i understand it and as needed in the modified `Utask.execute()`:
- If on a tworker the task always executes remotely, on whichever backend is decided later.
- Swarming tasks(thanks to the scheduler) are ALWAYS picked up by a tworker, hence causing the fuzz and other UTasks to be executed remotely.
- Other kinds of tasks & jobs it depends on the kind of bot that reaches here:
  - for instance the legacy bots were not tworkers but they still picked up the task as uworker and ended up executing locally thanks to:
https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/_internal/bot/tasks/task_types.py#L172

src/clusterfuzz/_internal/bot/tasks/task_types.py
IvanBM18: Done

dylanjew (commented):
LGTM!

PauloVLB (approved):
Just a heads up on something this unblocks.

When swarming can't take a task (queue full, API error), the gate hands it over to gcp_batch:

https://github.com/google/clusterfuzz/blob/d6d9c256fce6a5e9e5aba541f9fed3b22805de3f/src/clusterfuzz/_internal/remote_task/remote_task_gate.py#L192-L195

And batch resolves the specs for the whole list before creating anything, so one task whose platform isn't in the batch config kills the entire chunk:

https://github.com/google/clusterfuzz/blob/d6d9c256fce6a5e9e5aba541f9fed3b22805de3f/src/clusterfuzz/_internal/batch/service.py#L493

Saw it in dev with `ValueError: No mapping for ANDROID_EMULATOR-PREEMPTIBLE-UNPRIVILEGED`:

image

It doesn't hurt much in dev because `gcp_batch_jobs_frequency` is 0.2, so the slice is a prefix of the list and swarming tasks are appended last, meaning most cycles only send regular tasks to batch. In prod the frequency is 1.0, so the slice is always the whole list: one bad swarming task fails the whole cycle, and since nothing gets acked it comes back a few minutes later and fails again.

Not asking to change this PR, but I think we should make batch skip the tasks it can't map instead of raising for the whole list before we let this run in prod.

dylanjew:
> Just a heads up on something this unblocks.
> 
> When swarming can't take a task (queue full, API error), the gate hands it over to gcp_batch:
> 
> https://github.com/google/clusterfuzz/blob/d6d9c256fce6a5e9e5aba541f9fed3b22805de3f/src/clusterfuzz/_internal/remote_task/remote_task_gate.py#L192-L195
> 
> And batch resolves the specs for the whole list before creating anything, so one task whose platform isn't in the batch config kills the entire chunk:
> 
> https://github.com/google/clusterfuzz/blob/d6d9c256fce6a5e9e5aba541f9fed3b22805de3f/src/clusterfuzz/_internal/batch/service.py#L493
> 
> Saw it in dev with `ValueError: No mapping for ANDROID_EMULATOR-PREEMPTIBLE-UNPRIVILEGED`:
> 
> image
> It doesn't hurt much in dev because `gcp_batch_jobs_frequency` is 0.2, so the slice is a prefix of the list and swarming tasks are appended last, meaning most cycles only send regular tasks to batch. In prod the frequency is 1.0, so the slice is always the whole list: one bad swarming task fails the whole cycle, and since nothing gets acked it comes back a few minutes later and fails again.
> 
> Not asking to change this PR, but I think we should make batch skip the tasks it can't map instead of raising for the whole list before we let this run in prod.

This sounds like something we should fix, but are you thinking it's not a blocker because the frequency of swarming not being able to take the task and falling back to gcp_batch is relatively low? If it continously fails and doesn't ack that sounds concerning

We can discuss in a separate bug/PR, but I also would expect failed swarming tasks to get retried on swarming, and failed batch tasks to get retried on batch, rather than skipping any retries for swarming tasks that we can't map to batch.

PauloVLB:
> but are you thinking it's not a blocker because the frequency of swarming not being able to take the task and falling back to gcp_batch is relatively low?

I assumed that when looking into the queue full problem, that's why I didn't treat this as a blocker for the PR. But there are other reasons for which swarming may refuse the task, like the job being deleted or the config changing after the task was queued, and those go to batch. And they don't recover on their own, dev has had the same 28 tasks in utask_main-swarming for 2+ days hitting the fallback: 
image


And agree on the retries. I think the root of it is that _handle_swarming_jobs returns a single list that mixes "not a swarming task" with "swarming task I couldn't schedule", so the gate can't tell them apart and sends both to batch. If we split those, the declined swarming tasks just get nacked and retried on swarming on the next cycle, no cross backend at all. Which is better anyway, since something like ANDROID_EMULATOR can't run on batch even if we wanted to.

I'd still make batch skip the tasks it can't map, but that's a separate bug: any deleted job already does the same thing today, no swarming involved.

IvanBM18:
Thanks for your findings paulo, Will leave this PR as it is (approved but not merged) and will stack a new one top of it changes so that unscheduled swarming tasks doesn't overthrows the whole batch of tasks as invalid.

Regarding the 28 tasks, i am positive said tasks are because the job no longer exists in the db, because currently the job-exporter is broken and besides breaking issues(like missing spec or config) would show more errors.

